PR #75, opened 2021-05-20
Deleting line that led to component tests being deselected

adam-a-a:
## Fixes/Addresses:

Quick fix to pytest.ini

## Summary/Motivation:

Component tests are not running. I deleted one line from pytest.ini to fix this.

- Remove `-m unit` option

### Legal Acknowledgement

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,7 +1,6 @@
 [pytest]
 addopts = -W ignore
           --durations=100
-          -m unit
           --cov
           --cov-config=.coveragerc
 testpaths = proteuslib


adam-a-a:
Alternatively can use `-m "unit or component"` but just deleted that line all together in pytest.ini. Is there any reason why we'd want to include that line over deleting it?

TimBartholomew:
@adam-a-a we can run all the tests with pytest -m "unit or component"

@ksbeattie did you add the line to deselect component tests because of test failures? I believe the test failure you were experience was because the test require IDAES main, not the most recent release. This emphasizes that Proteuslib needs to be tested against its own fork of IDAES main.

bknueven (approved):
The entire test suite executes in under 15 seconds, so no reason to just run unit tests.

As soon as nawi-hub/idaes-pse#1 is merged all the tests will pass against our fork of IDAES as well.

PR #79, opened 2021-05-20
Created initial test case for seawater alkalinity

aladshaw3:
## Fixes/Addresses:

Integration testing for application of chemical framework to solve problem that emulates the pH, salinity, and carbonate alkalinity of seawater. 

## Summary/Motivation:

This is an early real-world test for the utilization of the IDAES chemical framework to solve realistic aqueous chemistry problems. The test sets up a simulated seawater made up of water, sodium bicarbonate, and some additional salt ions in solution at the typical concentrations found in real seawater. Chemical equilibrium is then solved using both the "inherent" reaction system and standard "equilibrium" reactions. This demonstrates that the chemical framework can solve relevant problems for NAWI/ProteusLib. 

- Added new test file for calculations of seawater pH, salinity, and carbonate alkalinity

### Legal Acknowledgement

### DIFF
--- a/proteuslib/tests/test_seawater_alkalinity.py
+++ b/proteuslib/tests/test_seawater_alkalinity.py
@@ -832,17 +832,39 @@ class TestSeawaterAlkalinity():
     def test_initialize_inherent(self, inherent_reactions_config):
         model = inherent_reactions_config
 
+        orig_fixed_vars = fixed_variables_set(model)
+        orig_act_consts = activated_constraints_set(model)
+
         solver.options['bound_push'] = 1e-10
         solver.options['mu_init'] = 1e-6
         model.fs.unit.initialize(optarg=solver.options)
+
+        fin_fixed_vars = fixed_variables_set(model)
+        fin_act_consts = activated_constraints_set(model)
+
+        assert degrees_of_freedom(model) == 0
+
+        assert len(fin_act_consts) == len(orig_act_consts)
+        assert len(fin_fixed_vars) == len(orig_fixed_vars)
 
     @pytest.mark.component
     def test_initialize_equilibrium(self, equilibrium_reactions_config):
         model = equilibrium_reactions_config
 
+        orig_fixed_vars = fixed_variables_set(model)
+        orig_act_consts = activated_constraints_set(model)
+
         solver.options['bound_push'] = 1e-10
         solver.options['mu_init'] = 1e-6
         model.fs.unit.initialize(optarg=solver.options)
+
+        fin_fixed_vars = fixed_variables_set(model)
+        fin_act_consts = activated_constraints_set(model)
+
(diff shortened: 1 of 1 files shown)


TimBartholomew:
@aladshaw3 did you test this with `pytest -m "unit or component"`, currently pytest is set up to skip tests marked with component

Andresj89:
just a couple of questions:
1) Do we need to assert `test_initialize_inherent` and `test_initialize_equilibrium` after calling the initialize method?
2) A more general question, right now the config files are being built in the same file with the tests. Do you guys think that eventually we can build the config files straight from the data base? I think that would make our tests cleaner.

andrewlee94:
To follow up on @Andresj89's comments - yes, every test needs to contain an assert of some sort. For cases like this where we call a solver (even if indirectly), we need to make sure the solver returned a feasible solution.

aladshaw3:
> @aladshaw3 did you test this with `pytest -m "unit or component"`, currently pytest is set up to skip tests marked with component

Yes, I ran pytest with those options. All tests passed on my machine.

aladshaw3:
> just a couple of questions:
> 
> 1. Do we need to assert `test_initialize_inherent` and `test_initialize_equilibrium` after calling the initialize method?
> 2. A more general question, right now the config files are being built in the same file with the tests. Do you guys think that eventually we can build the config files straight from the data base? I think that would make our tests cleaner.

@andrewlee94 

1. Is there a specific assertion I can put here? Does the initialize method in IDAES return any status information we can check?
2. Yes, eventually the config files should be built from the database. That will be in future unit tests that will live under the electrolytedb folder I assume. However, for the time being, we need these tests to ensure that the IDAES chemistry framework can solve relevant chemical problems of interest.

andrewlee94:
@aladshaw3 Unfortunately, there is no easy or standard answer to that question. I just had a look, and in IDAES different models do different things. Notably, of the ones I looked at, none of them checked for a feasible solution (`initialize` does not return a status, which it probably should).

I suppose the correct answer is that IDAES needs to make `initialize` return a status at the end so we can confirm.

TimBartholomew:
> @aladshaw3 Unfortunately, there is no easy or standard answer to that question. I just had a look, and in IDAES different models do different things. Notably, of the ones I looked at, none of them checked for a feasible solution (`initialize` does not return a status, which it probably should).
> 
> I suppose the correct answer is that IDAES needs to make `initialize` return a status at the end so we can confirm.

I second that!

aladshaw3:
@Andresj89 @andrewlee94 @TimBartholomew 

Committed additional changes to this PR. Added assertions in the initialize tests following the IDAES 'test_BTIdeal.py' example. Ran all tests (both unit and component) and checked against Pyomo 6 to ensure all tests passed.

Andresj89 (approved):
LGTM

Issue #62, opened 2021-05-11
Add details on spacer geometry for hydraulic diameter computation


adam-a-a:
In the RO model, spacers are placed within the feed channel to increase crossflow velocity and provide structural support. 

The geometry of those spacers (e.g., filament diameter) will affect the hydraulic diameter and related parameters such as frictional pressure losses, dimensionless numbers (e.g., Reynolds), and the mass transfer coefficient.  

Currently, we only account for spacer porosity in the 0D RO model (pending PR #61 ).

PR #80, opened 2021-05-20
Add spacer geometry and update pressure drop and mass transfer coeff calculations

adam-a-a:
## Fixes/Addresses:

Issue #62 

## Summary/Motivation:

This PR updates the calculation of dh, the hydraulic diameter, by accounting for spacer geometry; diameter of the spacer filament is introduced as a new variable and the hydraulic diameter incorporates channel height, spacer porosity, and filament diameter. 

Spacers are used to enhance mixing and increase the crossflow velocity and thus the mass transfer coefficient. Consequently, this will also increase the pressure drop across the feed channel.  

- Add spacer filament diameter and full equation for hydraulic diameter computation
- Adjust variable bounds as needed (reduce unnecessarily large bounds) to deal with infeasibilities
- Update testing
- Correct an error in the Reynolds number equation (previously iterated over wrong set in line comprehension) that led to incorrect pressure drop calculations

### Legal Acknowledgement


proteuslib/unit_models/reverse_osmosis_0D.py
```diff
@@ -622,16 +627,22 @@ def eq_N_Re_io(b, t, x):
...
-                        2 * (b.channel_height * b.width)
-                        / (b.channel_height + b.width))
+                        4 * b.spacer_porosity
+                        / (2 / b.channel_height
+                           + (1 - b.spacer_porosity) * 4 / b.df))
+
+            # Filament diameter cannot be larger than channel height
+            @self.Constraint(doc="filament diameter constraint")
+            def eq_df(b):
+                return b.df <= b.channel_height
```
TimBartholomew, line 645: This constraint should be removed. While it should be always satisfied. we typically don't add constraints like this. For example we don't have a constraint that the outlet flux should be less than inlet flux and you can imagine many additional constraints we can right that should always be true. 

This case is slightly different than the example since these are both supposed to specified by the user, but if the user fixed them to wrong values they would initially just get that the model is infeasible (albeit they could dig into it and figure out it is this constraint). However, in general, we don't provide checks and/or constraints on whether user provided parameters are feasible, that responsibility is on the user. 

Also thinking about this more, typically we don't optimize the spacer geometry. Usually that would require another CFD study, for example df = 1/2 * h is a condition for using Hoek's relationship, therefore we should include the calculation of df as a constraint rather than a user provided input. I'm unsure about the other relationships.

proteuslib/unit_models/tests/test_reverse_osmosis_0D.py
```diff
@@ -586,10 +587,9 @@ def test_Pdrop_calculation(self):
         feed_flow_mass = 1
         feed_mass_frac_NaCl = 0.035
         feed_mass_frac_H2O = 1 - feed_mass_frac_NaCl
-        feed_pressure = 50e5
+        feed_pressure = 85e5
```
TimBartholomew, line 590: Does this test require 85 bar? Does it not solve with 55 bar?

proteuslib/unit_models/tests/test_reverse_osmosis_0D.py
```diff
@@ -633,14 +638,15 @@ def test_Pdrop_calculation(self):
         assert len(unscaled_constraint_list) == 0
 
         # test initialization
-        initialization_tester(m)
+        # initialization_tester(m)
```
TimBartholomew, line 641: commented code

proteuslib/unit_models/tests/test_reverse_osmosis_0D.py
```diff
@@ -633,14 +638,15 @@ def test_Pdrop_calculation(self):
...
 
         # test variable scaling
         badly_scaled_var_lst = list(badly_scaled_var_generator(m))
+        [print(k, v) for k, v in badly_scaled_var_lst]
```
TimBartholomew, line 645: delete print statment

proteuslib/unit_models/tests/test_reverse_osmosis_0D.py
```diff
@@ -633,14 +638,15 @@ def test_Pdrop_calculation(self):
...
         assert badly_scaled_var_lst == []
 
         # test solve
-        solver.options = {'nlp_scaling_method': 'user-scaling'}
+        solver.options = {'nlp_scaling_method': 'user-scaling', 'halt_on_ampl_error': 'yes'}
```
TimBartholomew, line 649: remove halt_on_ampl_error

proteuslib/unit_models/tests/test_reverse_osmosis_0D.py
```diff
@@ -649,23 +655,37 @@ def test_Pdrop_calculation(self):
...
+        assert (pytest.approx(0.0914, rel=1e-3) == value(m.fs.unit.velocity_io[0, 'out']))
+        assert (pytest.approx(151.622, rel=1e-3) == value(m.fs.unit.velocity_io[0, 'in']
+                * m.fs.unit.feed_side.properties_in[0].dens_mass_phase['Liq']
+                * m.fs.unit.dh
+                / m.fs.unit.feed_side.properties_in[0].visc_d_phase['Liq']))
+        assert (pytest.approx(63.001, rel=1e-3) == value(m.fs.unit.velocity_io[0, 'out']
+                * m.fs.unit.feed_side.properties_out[0].dens_mass_phase['Liq']
+                * m.fs.unit.dh
+                / m.fs.unit.feed_side.properties_out[0].visc_d_phase['Liq']))
+        assert (pytest.approx(-2.786e5, rel=1e-3) == value(m.fs.unit.deltaP[0]))
```
TimBartholomew, line 670: This is a much higher pressure drop than I'd expect given my OARO paper. The average Reynolds number is ~90, but it has about 35kPa/m pressure drop. Nearly 10x more than the OARO paper, which was operating at a much higher Reynolds number ~600 .

proteuslib/unit_models/tests/test_reverse_osmosis_0D.py
```diff
@@ -649,23 +655,37 @@ def test_Pdrop_calculation(self):
...
+        assert (pytest.approx(63.001, rel=1e-3) == value(m.fs.unit.N_Re_io[0, 'out']))
+        assert (pytest.approx(0.0914, rel=1e-3) == value(m.fs.unit.velocity_io[0, 'out']))
+        assert (pytest.approx(151.622, rel=1e-3) == value(m.fs.unit.velocity_io[0, 'in']
+                * m.fs.unit.feed_side.properties_in[0].dens_mass_phase['Liq']
+                * m.fs.unit.dh
+                / m.fs.unit.feed_side.properties_in[0].visc_d_phase['Liq']))
+        assert (pytest.approx(63.001, rel=1e-3) == value(m.fs.unit.velocity_io[0, 'out']
+                * m.fs.unit.feed_side.properties_out[0].dens_mass_phase['Liq']
+                * m.fs.unit.dh
+                / m.fs.unit.feed_side.properties_out[0].visc_d_phase['Liq']))
```
TimBartholomew, line 669: This test looks like it was part of your debugging for Reynolds number.

proteuslib/unit_models/tests/test_reverse_osmosis_0D.py
```diff
@@ -649,23 +655,37 @@ def test_Pdrop_calculation(self):
...
-        assert (pytest.approx(53.561, rel=1e-3) ==
+        assert (pytest.approx(49.2250, rel=1e-3) ==
                 value(m.fs.unit.feed_side.properties_interface_in[0].conc_mass_phase_comp['Liq', 'NaCl']))
-        assert (pytest.approx(39.524, rel=1e-3) ==
+        assert (pytest.approx(81.4165, rel=1e-3) ==
                 value(m.fs.unit.feed_side.properties_out[0].conc_mass_phase_comp['Liq', 'NaCl']))
-        assert (pytest.approx(46.958, rel=1e-3) ==  # TODO: expected this value to be higher than interface concentration at inlet, but bypassing for now- has to do with pressure drop
+        assert (pytest.approx(87.7047, rel=1e-3) ==
                 value(m.fs.unit.feed_side.properties_interface_out[0].conc_mass_phase_comp['Liq', 'NaCl']))
+        assert value(m.fs.unit.df) <= value(m.fs.unit.channel_height)
```
TimBartholomew, line 688: Usually we don't check user inputs

proteuslib/unit_models/tests/test_reverse_osmosis_0D.py
TimBartholomew: We know that there was a mistake in the hydraulic diameter calculation of the OARO paper, but it shouldn't make this big of a difference. You can compare some of these results to my membrane module modeling paper https://www.sciencedirect.com/science/article/pii/S0376738818326115, which has the corrected hydraulic diameter calculation and uses a similar porosity to your test case.

TimBartholomew (changes requested):
Some comments on removing unnecessary things, and in looking at the test case it still seems that the pressure drop is too high compared to my previous work, since we are using the same relationships the results should match.

proteuslib/unit_models/tests/test_reverse_osmosis_0D.py
adam-a-a: This was really to check that the constraint is in place.

TimBartholomew: It doesn't check if the constraint is in place, these values are fixed in the test case. An isinstance(constraint_name, Constraint) would check if it exists, but as mentioned above I think we should drop the constraint anyways.

proteuslib/unit_models/tests/test_reverse_osmosis_0D.py
TimBartholomew: Yeah I understand, but I take that more as a debugging activity. We rarely have two tests for a variable and another way to calculate it. 

For example you could check that the average flux equals the permeate flow over the membrane area or that the velocity equals the total mass flowrate divided by density and cross sectional area (its currently being calculated by total flowrate divided by cross sectional area). The combinations of calculating variables in equivalent ways would be really high and its arbitrary that we are just doing this for the Reynolds number calculation, so I'd leave this just as a debugging/model development exercise rather than a test.

adam-a-a: Ok, makes sense.

proteuslib/unit_models/reverse_osmosis_0D.py
```diff
@@ -369,8 +374,8 @@ def build(self):
         if ((self.config.pressure_change_type != PressureChangeType.fixed_per_stage)
                 or (self.config.mass_transfer_coefficient == MassTransferCoefficient.calculated)):
             self.length = Var(
```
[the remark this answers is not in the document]
adam-a-a: I wrote this in such a way that length or width were only built when needed. Allow me to walk you through the logic as it stands now:
- In the baseline case, the user just fixes membrane area itself; i.e., there's no need to explicitly define length or width since neither are used anywhere else.
- the `eq_area` constraint is only active when the `mass_transfer_coefficient` config option is set to `MassTransferCoefficient.calculated`, and this is once case when length and width are both needed explicitly. 
- `width` is needed when either `self.config.mass_transfer_coefficient == MassTransferCoefficient.calculated` is set, `self.config.pressure_change_type == PressureChangeType.calculated` is set, or both are set. 
- `length` is needed when `width` is needed (previous bullet), but even if the options from the previous bullet were not active, `length` is also required when `self.config.pressure_change_type == PressureChangeType.fixed_per_unit_length`. In this case, `width` wouldn't be necessary. This is the exception.

That is why length and width were not built together.

proteuslib/unit_models/reverse_osmosis_0D.py
adam-a-a: Setting a constraint that includes the calculation of df would just be something like ` df == df_h_ratio * h'. In this case we'd build another variable (`df_h_ratio`) with bounds between some epsilon value and 1. In my view, that would be unnecessary since the current constraint essentially does the same thing without needing to set any ratio. I don't think that this case is comparable to the example of setting a constraint for outlet flux < inlet flux because outlet flux should never be greater than inlet flux, unless something was computed incorrectly in the model. 

However, the filament diameter is directly related to channel height, and in imagining some case where the user were not to fix _both_ the height and filament diameter simultaneously, I originally thought we might need this constraint. Although we aren't considering optimizing spacer geometry in the detailed way that CFD studies would, I still think we can leave this open for the user in case they'd want the model to yield the df/h ratio (for example). 

Overall, I have no issues with removing this constraint if we are absolutely certain that `df` will always be fixed by the user in every scenario; alternatively, would the user set a lower bound and upper bound on df instead?  On the other hand, if we were to keep this constraint, we should also have more equations for the Sherwood number and friction factor, which would be related to spacer dimensions (which would be another PR).

proteuslib/unit_models/tests/test_reverse_osmosis_0D.py
adam-a-a: It does solve with 50 bar.

proteuslib/unit_models/tests/test_reverse_osmosis_0D.py
adam-a-a: I would think that any major differences would be attributed to the selection of (not limited to) feed mass flow rate, membrane dimensions, and spacer geometry, and I don't think I matched up all specified variables with your original study in this particular test. 

I changed all fixed variables to match your study from the link you provided, and it seems like my results match up although I didn't see what your pressure drop result was. 

The results I get for pressure drop and dP/dx are 1.66 MPa and 0.104 MPa.

Additionally, although the solver gets a feasible solution, I get some badly scaled variables  when using the specifications in your study (had to disable assert to check results). Maybe we can discuss.

proteuslib/unit_models/reverse_osmosis_0D.py
```diff
@@ -310,40 +309,40 @@ def build(self):
...
                 domain=NonNegativeReals,
                 units=units_meta('length'),
                 doc='Feed-channel height')
+            self.dh = Var(
+                initialize=1e-3,
+                bounds=(1e-5, 1e-2),
+                domain=NonNegativeReals,
+                units=units_meta('length'),
+                doc='Hydraulic diameter of feed channel')
+            self.df = Var(
```
TimBartholomew, line 330: I thought we discussed making df either a calculated variable (based on a constraint) or substituting it out with df = 1/2 h

proteuslib/unit_models/tests/test_reverse_osmosis_0D.py
```diff
@@ -635,7 +636,7 @@ def test_Pdrop_calculation(self):
         # test initialization
         initialization_tester(m)
 
-        # test variable scaling
+        # # test variable scaling
```
TimBartholomew, line 639: double #

pytest.ini
```diff
@@ -12,4 +12,4 @@ log_file_level = INFO
 markers =
     unit: quick tests that do not require a solver, must run in <2s
     component: quick tests that may require a solver
-    integration: long duration tests
+    integration: long duration tests
```
TimBartholomew, line 15: add new line, currently git is saying you are changing the file because there is no newline at the end of the file

TimBartholomew (changes requested):
Minor comments

TimBartholomew (approved):
LGTM

PR #83, opened 2021-05-21
Add test for recarbonation process

srikanthallu:
## Fixes/Addresses:

This is a validation test for recarbonation model which is used as pre-treatment process.

## Summary/Motivation:

The test uses generic chemistry and properties packages from IDAES framework. The initial validation test is used to check appropriate equilibrium reactions are used to solve for concentration and pH of solution.

- Adds a validation test (test_recarbonation_process.py) for building components and equilibrium reactions.

### Legal Acknowledgement

### DIFF
--- a/proteuslib/tests/test_recarbonation_process.py
+++ b/proteuslib/tests/test_recarbonation_process.py
@@ -1,0 +1,541 @@
+###############################################################################
+# ProteusLib Copyright (c) 2021, The Regents of the University of California,
+# through Lawrence Berkeley National Laboratory, Oak Ridge National
+# Laboratory, National Renewable Energy Laboratory, and National Energy
+# Technology Laboratory (subject to receipt of any required approvals from
+# the U.S. Dept. of Energy). All rights reserved.
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and license
+# information, respectively. These files are also available online at the URL
+# "https://github.com/nawi-hub/proteuslib/"
+#
+###############################################################################
+
+'''
+    This file is to run the example of running a phase change for the dissolution
+    of CO2 into water. 
+'''
+# =================== Import Statements ==============================
+
+import unittest
+import pytest
+
+# Import specific pyomo objects
+from pyomo.environ import (Block,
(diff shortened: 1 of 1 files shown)


proteuslib/tests/test_recarbonation_process.py
```diff
@@ -0,0 +1,541 @@
...
+        assert pytest.approx(10, rel=1e-3) == value(model.fs.unit.outlet.flow_mol[0])
+        assert pytest.approx(101325, rel=1e-3) == value(model.fs.unit.outlet.pressure[0])
+
+        total_molar_density = \
+            value(model.fs.unit.control_volume.properties_out[0.0].dens_mol_phase['Liq'])/1000
+        assert pytest.approx(55.2336, rel=1e-3) == total_molar_density
+        pH = -value(log10(model.fs.unit.outlet.mole_frac_comp[0, "H_+"]*total_molar_density))
+        pOH = -value(log10(model.fs.unit.outlet.mole_frac_comp[0, "OH_-"]*total_molar_density))
+        assert pytest.approx(5.336, rel=1e-3) == pH
+        assert pytest.approx(8.598, rel=1e-3) == pOH
```
aladshaw3, line 541: I would add one more assertion here for the amount of CO2 absorbed into the liquid phase. We know from literature that our aqueous CO2 should be around 0.02754 mol/L at this given inlet molefraction, and since we are utilizing the built in 'fugacity' calculations of IDAES, we should check the solution here to make sure we get it correct. That way if anything changes downstream in IDAES, the test would catch it here. 

So, I would add the following lines...

CO2_sorbed = model.fs.unit.control_volume.properties_out[0.0].conc_mol_phase_comp[('Liq', 'CO2')].value
assert pytest.approx(27.54, rel=1e-3) == CO2_sorbed

(NOTE: The last decimal places may vary slightly)

srikanthallu: I have added the assertion for CO2 absorbed in liquid phase.

aladshaw3 (approved):
All looks good to me.

PR #84, opened 2021-05-25
Added variables and improved initialization for 0D RO model

### LINKED ISSUE #55: Add variables to RO model
It would be useful if the RO unit model had variables for recovery of the components (i.e. `recovery_mass_comp[t, j]`) and the volumetric recovery (i.e. `recovery_vol[t, j]`).

TimBartholomew:
## Fixes/Addresses:

Fixes #55 

## Summary/Motivation:
There are several system variables that would be useful to have on the RO model.

- Added recovery_vol_phase variable and constraint
- Added recovery_mass_phase_comp variables and constraints
- Added rejection_phase_comp variable and constraint
- Added scaling factors and transformed constraints

### Legal Acknowledgement


TimBartholomew:
@adam-a-athis PR does not implement the additional variables through a callable method because after discussions with @andrewlee94, it became clear that these system variables would be generally useful in most cases and including a variable and its associated constraint adds negligible computational demand. Also adding callable method to create variables would depart from the typical IDAES API.

proteuslib/unit_models/reverse_osmosis_0D.py
```diff
@@ -293,6 +292,29 @@ def build(self):
...
+            units=pyunits.dimensionless,
+            doc='Volumetric-based recovery')
+        self.rejection_phase_comp = Var(
+            self.flowsheet().config.time,
+            self.config.property_package.phase_list,
+            self.solute_list,
+            initialize=0.1,
+            bounds=(1e-8, 1),
+            units=pyunits.dimensionless,
+            doc='Bulk concentration based component rejection')
```
adam-a-a, line 317: This is usually referred to as "observed salt/solute rejection".

proteuslib/unit_models/reverse_osmosis_0D.py
```diff
@@ -726,6 +748,27 @@ def eq_equal_flow_vol_interface_io(b, t, x):
...
+            return (b.recovery_mass_phase_comp[t, 'Liq', j] ==
+                    b.properties_permeate[t].flow_mass_phase_comp['Liq', j] /
+                    b.feed_side.properties_in[t].flow_mass_phase_comp['Liq', j])
+
+        @self.Constraint(self.flowsheet().config.time,
+                         self.solute_list)
+        def eq_rejection_phase_comp(b, t, j):
+            return (b.rejection_phase_comp[t, 'Liq', j] ==
+                    1 - (b.properties_permeate[t].conc_mass_phase_comp['Liq', j] /
+                         b.feed_side.properties_in[t].conc_mass_phase_comp['Liq', j]))
```
adam-a-a, line 770: Just reconsidering whether we should add intrinsic rejection as well, which would be the same constraint, except you'd divide by the feed concentration at the membrane-interface. Experimentalists make use of intrinsic solute rejection, and the value gives a sense of the "real" rejection of a membrane. 

I think it would be nice to have observed_rejection and intrinsic_rejection.

proteuslib/unit_models/reverse_osmosis_0D.py
```diff
@@ -1061,4 +1115,14 @@ def calculate_scaling_factors(self):
...
+            sf = iscale.get_scaling_factor(self.recovery_vol_phase[t, 'Liq'])
+            iscale.constraint_scaling_transform(c, sf)
+
+        for (t, j), c in self.eq_recovery_mass_phase_comp.items():
+            sf = iscale.get_scaling_factor(self.recovery_mass_phase_comp[t, 'Liq', j])
+            iscale.constraint_scaling_transform(c, sf)
 
+        for (t, j), c in self.eq_rejection_phase_comp.items():
+            sf = iscale.get_scaling_factor(self.rejection_phase_comp[t, 'Liq', j])
+            iscale.constraint_scaling_transform(c, sf)
```
adam-a-a, line 1128: add new line at end

adam-a-a (changes requested):
I had only a couple of minor comments. My main suggestion is to include intrinsic salt rejection in addition to observed salt rejection. Also, since we just merged #80, you may need to pull the repo and adjust the tests (e.g., number of variables).

proteuslib/unit_models/reverse_osmosis_0D.py
TimBartholomew: I looked into this and intrinsic rejection is tricky and I'd like to know more about the use case before we add it. The observed rejection is 1 - concentration in permeate/concentration in feed, for the intrinsic what is the denominator? The concentration at the membrane interface at the inlet or at the outlet? 

Essentially, the disconnect is the observed rejection is a module metric while the intrinsic rejection is a local metric (i.e. at a specific point)

I'm leaving this off for now, but we can always add it.

proteuslib/unit_models/reverse_osmosis_0D.py
```diff
@@ -313,7 +313,13 @@ def build(self):
             initialize=0.1,
             bounds=(1e-8, 1),
             units=pyunits.dimensionless,
-            doc='Bulk concentration based component rejection')
+            doc='Observed solute rejection')
+        self.over_pressure = Var(
+            self.flowsheet().config.time,
+            initialize=0.1,
+            bounds=(1e-8, 100),
```
adam-a-a, line 320: Shouldn't upper bound be 1 if this is a fraction?

proteuslib/unit_models/reverse_osmosis_0D.py
adam-a-a: This [paper by Kim & Hoek ](https://doi.org/10.1016/j.desal.2005.05.017) gives an example of how intrinsic rejection can be used (check out equations 6-7 for a quick glance). Essentially, intrinsic rejection is the "real" rejection of the membrane, similar to how the membrane-interface osmotic pressure is the "real" osmotic pressure that must be overcome. I suppose it'd be preferable to make this a local variable, so perhaps we can reconsider in another PR (could also get the advisory board's take on use cases).

bknueven (approved):

proteuslib/unit_models/reverse_osmosis_0D.py
TimBartholomew: You could have an over pressure that is over 100%. I obviously put a very loose bound but it could be over 1.

adam-a-a: Oh, right- I was thinking it would be less than 1 since you're subtracting 1, but I see how it could be >1.

proteuslib/unit_models/reverse_osmosis_0D.py
```diff
@@ -275,29 +276,81 @@ def build(self):
...
-
         # Add unit variables
+        flux_mass_io_phase_comp_initialize = {}
+        for t in self.flowsheet().config.time:
+            for io in self.io_list:
+                for j in self.config.property_package.component_list:
+                    if j in self.solvent_list:
+                        flux_mass_io_phase_comp_initialize[(t, io, 'Liq', j)] = 1e-3
+                    elif j in self.solute_list:
+                        flux_mass_io_phase_comp_initialize[(t, io, 'Liq', j)] = 1e-6
```
bknueven, line 287: It shouldn't hold up this PR, but to me it's more in the Pyomo-style to pass a function to initialize, which unwinds these for loops behind the scenes:
```
def flux_mass_io_phase_comp_initialize(b, t, io, j):
    if j in self.solvent_list:
        return 1e-3
    elif j in self.solute_list:
        return 1e-6
```
There should maybe be an "else" clause in this function, depending on how defensively we're programming.

bknueven (approved):

adam-a-a (changes requested):
I think it would be best to separate all the initialization revisions and put them in another PR dedicated to improving initialization/model stability. 

I think once that's done, all the other changes related to new system variables can be merged.

TimBartholomew:
> I think it would be best to separate all the initialization revisions and put them in another PR dedicated to improving initialization/model stability.
> 
> I think once that's done, all the other changes related to new system variables can be merged.

Adding the new variables made the model more unstable without better initial values for them, so the initialization improvements are loosely linked to adding the variables. I can update the name of the PR, if the main point was to keep better track of the changes. Otherwise this PR is still relatively manageable in size to be a single PR.

proteuslib/unit_models/reverse_osmosis_0D.py
TimBartholomew: Yeah I'll update this. There is already a check for if component j is not a solvent or a solute and it raises an Exception, so the else clause is probably unnecessary.

proteuslib/unit_models/reverse_osmosis_0D.py
```diff
@@ -463,11 +507,17 @@ def build(self):
             self.deltaP = Reference(self.feed_side.deltaP)
 
         # mass transfer
+        mass_transfer_phase_comp_initialize = {}
+        for t in self.flowsheet().config.time:
+            for j in self.config.property_package.component_list:
+                    mass_transfer_phase_comp_initialize[(t, 'Liq', j)] = \
+                        value(self.feed_side.properties_in[t].get_material_flow_terms('Liq', j)
+                              * self.recovery_mass_phase_comp[t, 'Liq', j])
```
bknueven, line 515: This one could be functionalized as well, though it's obviously not as clean.

bknueven (approved):
I would like to see this merged, and we can sort out any lingering issues with the 0D RO model in a subsequent PR.

proteuslib/unit_models/reverse_osmosis_0D.py
TimBartholomew: Oops good catch, that variable was in another section.

bknueven (dismissed):

proteuslib/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
```diff
@@ -234,7 +234,7 @@ def calculate_operating_pressure(feed_state_block=None, over_pressure=0.15,
     return value(t.brine[0].pressure_osm) * (1 + over_pressure)
 
 
-def calculate_RO_area(unit=None, water_recovery=0.5, solver=None):
+def calculate_RO_area(unit=None, water_recovery=0.5, solver_dict=None):
```
bknueven, line 237: I would prefer if we maintained the `solver` keyword argument and add a `optarg` keyword argument. The `solver` could be a string which we use to create a solver object within this function using the options. This maintains consistency with IDAES.

I'm happy to make this change myself.

bknueven (approved):
My comment above needn't hold this up as it's simple to fix later.

proteuslib/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
TimBartholomew: I'll make an issue for it. There's a couple of other things I could add.

bknueven: Thanks! Part of #102.

adam-a-a (approved):
LGTM

PR #87, opened 2021-05-25
Making parameter sweep tests cleanup after themselves

### LINKED ISSUE #78: pytest_output and pytest_output_debug
The parameter sweep tool creates pytest_output and pytest_output_debug folders during testing. Afterwards they become untracked folders.

We should do the following:
1. clean up at the end of the parameter sweep test to delete the folders
2. add those folders to gitignore

I'm leaning towards 1. Also is the pytest_output_debug necessary or was that for debugging the tests when they were being created?

### LINKED ISSUE #90: Stand up mpi4py GHA
to run parameter sweep tests in parallel.

bknueven:
## Fixes/Addresses:
Fixes #78 and fixes #90

## Summary/Motivation:
Leaving extra directories lying about isn't nice, and can affect test reliability 

- Making parameter sweep tests cleanup after themselves
- Use formatter strings for local debug file names for readability
- Adds mpi4py GitHub Action workflow / tests

### Legal Acknowledgement


proteuslib/tools/parameter_sweep.py
```diff
@@ -224,7 +224,7 @@ def parameter_sweep(m, sweep_params, outputs, results_file, optimize_fct,
 
     if debugging_data_dir is not None:
         # Create the local filename and data
-        fname = os.path.join(debugging_data_dir, 'local_results_%03d.csv' % (rank))
+        fname = os.path.join(debugging_data_dir, f'local_results_{rank}.csv')
```
eyoung55, line 227: Is there an easy way to preserve the leading zeros using f-strings?  The sorting is a little bit nicer if they're there, and using the format modifiers gives us control over the output.

eyoung55 (changes requested):
Looks good!  I'm not particular to one way of string formatting vs another, but I do think it would be good to preserve leading zeros for easily working through the debugging files.

eyoung55 (dismissed):

bknueven:
Using pytest's built-in temp file manager causes the these tests to fail in parallel.

TimBartholomew (dismissed):

.github/workflows/mpi4py-test.yml
```diff
@@ -0,0 +1,45 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: conda-incubator/setup-miniconda@v2
+      with:
+        auto-update-conda: true
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        conda install mpi4py
```
ksbeattie, line 33: Could you make this `conda install --quiet --yes mpi4py`?  I think the `--yes` is implied here already and the `--quiet` will cut down on the output noise (which in this case is not much, but for other packages it can become overwhelming, so I'd like to encourage all this as "good practice".

ksbeattie (changes requested):
Just the `--quiet --yes`.

TimBartholomew (approved):
LGTM

eyoung55 (approved):
looks good!

.github/workflows/mpi4py-test.yml
bknueven: Done!

ksbeattie (approved):

Issue #31, opened 2021-03-04
github actions


ksbeattie:
Get github actions running tests against the new NAWI/idaes-pse fork

PR #89, opened 2021-05-26
First GHA workflow

ksbeattie:
An very initial set-up of GitHub Actions, to run pytest.

This installs idaes-pse from the nawi-hub/idaes-pse fork.  More needs to be done around all this, but this is at least a starting point.

## Fixes/Addresses:

Addresses #31 

### Legal Acknowledgement

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,8 +50,6 @@ coverage.xml
 *.py,cover
 .hypothesis/
 .pytest_cache/
-pytest_output/
-pytest_output_debug/
 
 # Translations
 *.mo
--- a/.gitignore
+++ b/.gitignore
@@ -50,6 +50,8 @@ coverage.xml
 *.py,cover
 .hypothesis/
 .pytest_cache/
+pytest_output/
+pytest_output_debug/
 
 # Translations
 *.mo
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -3,11 +3,7 @@
 
 name: Python package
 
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -50,6 +50,8 @@ coverage.xml
 *.py,cover
 .hypothesis/
 .pytest_cache/
+pytest_output/
+pytest_output_debug/
```
bknueven, line 54: Once PR #87 is merged we shouldn't need these lines.

bknueven (approved):
Seem reasonable. I'm not sure why this isn't firing now, but we should probably merge and see if it works.

I'll add an issue to get mpi4py tests running.

.gitignore
ksbeattie: I've removed them.  Thanks.

andrewlee94 (approved):
I don't know too much about this, but nothing jumps out as wrong.

We should set up our own IDAES fork at some point so we can control our version through tags.

ksbeattie:
> Seem reasonable. I'm not sure why this isn't firing now, but we should probably merge and see if it works

I think that's because this is the first one.  I tested it in my branch, with different settings.  But we'll see in a moment...

ksbeattie:
> I don't know too much about this, but nothing jumps out as wrong.
> 
> We should set up our own IDAES fork at some point so we can control our version through tags.

I've got the fork set up, and it's being used here, but finer control via tags is still to come.

Issue #91, opened 2021-05-26
Improve initialization for RO and property packages


TimBartholomew:
Currently the initialization routine for RO is not stable. In multiple flowsheets and cases, it fails to find the feasible solution. The initialization can be improved via the following:

- [x] initialize the interface state blocks, currently they are not being initialized before the unit model solve
- [x] update default values in the property packages to be more reasonable, i.e. current default values for component mass flowrate is 1 for solvents and solutes. The solutes should be 10 or 100 times smaller than the solvent
- [x] check the degrees of freedom in the initialization routine and raise an error if it is not 0, there are currently no checks for the degrees of freedom

adam-a-a:
@TimBartholomew do we want to add an assert dof == 0 _in_ the initialize() method for RO or the property model? Actually, do we want to add this in any initialize() method at all? I thought that the user could add the assert statement before the initialize method in a simulation file; nevertheless, I suppose it wouldn't hurt to have an assert statement carried out before AND within the initialize method.

andrewlee94:
@adam-a-a We have generally said that initialize functions should assert degrees of freedom. Initialization routines generally assume this, and if this is not true we cannot guarantee the behaviour of the routine or the final result. Hence, we have generally forced the user to fix the degrees of freedom and fail if this is not true - if the user really wants to initialize with degrees of freedom, they should write their own routine.

adam-a-a:
@TimBartholomew  Do you think the merging of #173 justifies closing this issue? In my opinion any overflow from this issue could be covered by issue #132.

PR #95, opened 2021-05-27
Property test harness

### LINKED ISSUE #94: Implement test harness to reduce duplication in property packages
Many of the tests for property packages are duplicated and result in a significant amount of copied code. Implementing a test harness will consolidate the repetitive sections and ensure all property packages include all required tests.

The test harness could be broken up in three components:
- Unit tests
- Component tests
- Regression tests

TimBartholomew:
## Fixes/Addresses:

Fixes #94 

## Summary/Motivation:

Property test harnesses will reduce duplicated code and enforce that all property packages past required tests. The proposed test harness requires developers to provide property package specific dictionaries and checks that they fully cover the package.

- Added PropertyUnitTestHarness class
- Added PropertyComponentTestHarness class
- Added property_regression_test function
- Updated seawater property package to use the test harnesses

### Legal Acknowledgement


proteuslib/property_models/tests/property_test_harness.py
```diff
@@ -0,0 +1,398 @@
...
+                                     unscaled_constraints_generator,
+                                     badly_scaled_var_generator)
+from idaes.core.util.exceptions import (PropertyPackageError,
+                                        PropertyNotSupportedError)
+from idaes.core.util import get_solver
+
+# -----------------------------------------------------------------------------
+# Get default solver for testing
+solver = get_solver()
+solver.options["nlp_scaling_method"] = "user-scaling"
```
andrewlee94, line 42: I would suggest that the test harness take arguments for these, so that it can be configured as needed when it is used in a specific test case.

proteuslib/property_models/tests/property_test_harness.py
```diff
@@ -0,0 +1,398 @@
...
+
+
+# -----------------------------------------------------------------------------
+class PropertyUnitTestHarness(object):
+    def configure_class(self, m):
```
andrewlee94, line 47: What is `m` here? Should there be an `__init__` method?

proteuslib/property_models/tests/property_test_harness.py
```diff
@@ -0,0 +1,398 @@
...
+        m.stateblock_obj_dict = self.stateblock_obj_dict
+        m.state_vars_list = self.state_vars_list
+
+        m.statistics_dict = self.statistics_dict
+        m.general_methods_dict = self.general_methods_dict
+        m.default_methods_dict = self.default_methods_dict
+
+
+    def configure(self):
+        # Placeholder method to allow user to setup test harness
```
andrewlee94, line 81: Need to document what is required here.

proteuslib/property_models/tests/property_test_harness.py
```diff
@@ -0,0 +1,398 @@
...
+        pass
+
+    @staticmethod
+    def check_block_obj(blk, obj_dict):
```
andrewlee94, line 85: This stems from above, but how is `obj_dict` provided and what form does it take?

proteuslib/property_models/tests/property_test_harness.py
```diff
@@ -0,0 +1,398 @@
...
+        """check that all objects are their expected type and have their expected units"""
```
andrewlee94, line 86: Typo: should start with capital,

andrewlee94 (commented):
My biggest comment with a test harness here is whether it actually makes things easier and clearer for the developer. I.e. is it any easier to input all the necessary data in the correct form for the harness than it would be to just write the tests manually.

This is especially true when it comes to having to test matches to data (where tolerance may need to be adjusted as well).

proteuslib/property_models/seawater_prop_pack.py
```diff
@@ -517,7 +518,7 @@ def build(self):
         self.flow_mass_phase_comp = Var(
             self.params.phase_list,
             self.params.component_list,
-            initialize=1,
+            initialize={('Liq', 'H2O'): 0.965, ('Liq', 'TDS'): 0.035},
```
adam-a-a, line 521: Nice, I see you started addressing our initialization problem here (#91). =)

proteuslib/property_models/tests/property_test_harness.py
TimBartholomew: This follows the IDAES API for the property test harness there. I tried a variety of approaches, but this turned out to be the best hack I found to actually get the tests to run and have pytest class frame work.

proteuslib/property_models/tests/property_test_harness.py
TimBartholomew: See comment above about following IDAES API for the property test harness.

proteuslib/property_models/tests/property_test_harness.py
TimBartholomew: this covers multiple dictionaries, the description of each developer provided dictionary is commented in the configure_class method

TimBartholomew:
> My biggest comment with a test harness here is whether it actually makes things easier and clearer for the developer. I.e. is it any easier to input all the necessary data in the correct form for the harness than it would be to just write the tests manually.
> 
> This is especially true when it comes to having to test matches to data (where tolerance may need to be adjusted as well).

@andrewlee94, I agree that the test harness doesn't significantly make things easier for the developer of a brand new model. Writing out all the required dictionaries for the test harness is fairly similar to copying, pasting, and editing another similar model (property packages in this case). I think it is somewhat clearer because the part that needs to be edited is the testing dictionary, whereas the copy and paste method, the developer has to identify what are the property specific parts.

The main benefit of the test harness is standardizing the required tests and having them located in one place rather than copy and pasted throughout multiple files. This has a couple of benefits: 
- First, as a reviewer, we know that if the test harness is used, the tests cover all the required tests. The reviewer does not need to look at some list of tests and confirm that they are all present.
- Second, if the required test should change in someway it is significantly easier to change them in one location rather than going to all the individual files to replace, edit, or add a method.

As for regression tests where tolerances may need to be adjusted, I think that is easily overcome by using the relative tolerance rather than the absolute tolerance, the relative tolerance value I've been using is 1e-3.

proteuslib/property_models/tests/property_test_harness.py
TimBartholomew: Yes. I'll add that as an input.

proteuslib/property_models/tests/property_test_harness.py
TimBartholomew: I added doc strings for all the requirements.

proteuslib/property_models/tests/property_test_harness.py
TimBartholomew: I added a doc string for each function.

proteuslib/property_models/tests/property_test_harness.py
```diff
@@ -0,0 +1,532 @@
...
+solver.options["nlp_scaling_method"] = "user-scaling"
+
+
+# -----------------------------------------------------------------------------
+class PropertyUnitTestHarness(object):
+    def configure_class(self, m):
+        self.configure()
+
+        # attaching objects to model to use in pytest frame
+        obj_list = ['component_dict', 'phase_dict', 'param_obj_dict',
```
andrewlee94, line 52: Do these actually need to be attached to the model object being tested? Would it not be possible to access these strings/dicts from the harness class when doing these tests?

There is the possibility of namespace clashes if we go attaching new attributes to the model like this.

proteuslib/property_models/tests/property_test_harness.py
```diff
@@ -0,0 +1,532 @@
...
+            assert isinstance(c, c_type)
+
+        # phases
+        assert isinstance(m.fs.properties.phase_list, Set)
+        assert len(m.fs.properties.phase_list) == len(m.phase_dict)
+        for p_str in m.phase_dict:
+            assert p_str in m.fs.properties.phase_list
+            p = getattr(m.fs.properties, p_str)
+            p_type = m.phase_dict[p_str]
+            assert isinstance(p, p_type)
```
andrewlee94, line 196: Should you also check the `phase_component_set`?

proteuslib/property_models/tests/property_test_harness.py
```diff
@@ -0,0 +1,532 @@
...
+
+    @pytest.mark.unit
+    def test_parameters(self, frame):
+        m = frame
+
+        self.check_block_obj(m.fs.properties, m.param_obj_dict)
+        self.check_block_obj_coverage(m.fs.properties, m.param_obj_dict)
+
+        # test that the parameter variables are fixed
+        for v in m.fs.properties.component_objects(Var):
```
andrewlee94, line 206: This presumes that all parameters will be Vars. Do we also need to allow for Params in some cases?

proteuslib/property_models/tests/property_test_harness.py
```diff
@@ -0,0 +1,532 @@
...
+        m = frame
+        assert hasattr(m.fs.properties, 'define_metadata')
+        metadata = m.fs.properties.get_metadata().properties
+
+        # check method naming convention
+        for v_str in metadata.keys():
+            if v_str in m.state_vars_list:
+                assert metadata[v_str]['method'] is None
+            else:
+                assert metadata[v_str]['method'] == '_' + v_str
```
andrewlee94, line 238: Is this test really necessary? Would it not be better to just test that all the expected properties can be constructed?

proteuslib/property_models/tests/property_test_harness.py
```diff
@@ -0,0 +1,532 @@
...
+        # check that the testing property list matches metadata
+        metadata = m.fs.properties.get_metadata().properties
+        assert len(metadata) == len(prop_str_list)
+        for v_str in metadata.keys():
+            assert v_str in prop_str_list
+
+        # check that properties are not built if not demanded
+        for obj_str in prop_str_list:
+            if obj_str not in m.state_vars_list:
+                assert not m.fs.stream[0].is_property_constructed(obj_str)
```
andrewlee94, line 259: This assumes that anything that is not a state variable is build-on-demand. This forces developers to use build on demand, and does not allow them to set certain properties as always present (e.g. enthalpy).

proteuslib/property_models/tests/property_test_harness.py
```diff
@@ -0,0 +1,532 @@
...
+
+        # check that properties are built on demand
+        for obj_str in prop_str_list:
+            assert hasattr(m.fs.stream[0], obj_str)
+
+        # add constraints for on demand variables to testing dictionary
+        for obj_str in prop_str_list:
+            if (m.stateblock_obj_dict[obj_str][0] is Var
+                    and obj_str not in m.state_vars_list):
+                m.stateblock_obj_dict['eq_' + obj_str] = (Constraint, None)
```
andrewlee94, line 269: This assumes a standard naming convention for constraints, which is not documented anywhere. It also assumes only one constraint per property which is not always the case.

proteuslib/property_models/tests/property_test_harness.py
```diff
@@ -0,0 +1,532 @@
...
+                                         'get_enthalpy_flow_terms']
+        tested_general_methods_list = []
+        for (method_name, ind) in m.general_methods_dict.keys():
+            if method_name not in tested_general_methods_list:
+                tested_general_methods_list.append(method_name)
+        for i in required_general_methods_list:
+            assert i in tested_general_methods_list
+
+        for (m_str, m_ind), (v_str, v_ind) in m.general_methods_dict.items():
+            assert hasattr(m.fs.stream[0], m_str)
```
andrewlee94, line 299: What does this test actually do? Also, the name `general_method` looks like it is a reserved name - I think you should choose something else.

proteuslib/property_models/tests/property_test_harness.py
```diff
@@ -0,0 +1,532 @@
...
+                getattr(m.fs.stream[0], v_str)
+            except PropertyPackageError or PropertyNotSupportedError:
+                pass  # errors in metadata entries are tested in unit test harness
+
+        # scale model
+        calculate_scaling_factors(m)
+
+        # initialize
+        m.fs.stream.initialize(solver=m.solver, optarg=m.optarg)
+
```
andrewlee94, line 433: There are a lot of other things that should be checked here as well, most significantly that the model "state" has not changed (i.e. the set of fixed variables and activated constraints).

proteuslib/property_models/tests/property_test_harness.py
TimBartholomew: There can be Params on the property block, this test just makes sure that any of the parameters that are Vars are fixed.

proteuslib/property_models/tests/property_test_harness.py
TimBartholomew: The test mostly just ensures that the on demand methods follow the standard naming convention, which is very minor. Do you think we shouldn't enforce this convention? 

The test_build checks that all expected properties can be constructed but I wanted there to be some test directly for the metadata, rather than an indirect test like if the properties can be constructed. 

I think I'll add a test that checks that the metadata includes the variables and expressions in the developer provided testing dictionary.

proteuslib/property_models/tests/property_test_harness.py
TimBartholomew: Ok I can update this to only do it for variables and expressions that have an on demand method in the metadata

proteuslib/property_models/tests/property_test_harness.py
TimBartholomew: Ok. I only did this to automate adding the constraints to the testing dictionary. I'll drop it and developers will write out the constraints in the test harness set up.

proteuslib/property_models/tests/property_test_harness.py
andrewlee94: The main problem here is that there is no documented naming convention for constraints. Making this a standard is not necessarily a problem, but it would need to be documented and conveyed to the developers.

I am not sure how much value there is to testing the metadata directly either - the metadata doesn't do a lot all up and the only important part is that build-on-demand works. You are also assuming that only state vars will have `"mehtod"=None` (and even assuming that as well).

proteuslib/property_models/tests/property_test_harness.py
andrewlee94: Rather, I think you should have two separate tests - one for things that should always exist, and one for things that should be build-on-demand.

proteuslib/property_models/tests/property_test_harness.py
```diff
@@ -0,0 +1,532 @@
...
+                    'default_scaling_factor_dict', 'stateblock_obj_dict', 'state_vars_list',
+                    'statistics_dict', 'general_methods_dict', 'default_methods_dict']
+        for obj_str in obj_list:
+            if not hasattr(self, obj_str):
+                raise ConfigurationError(
+                    "{obj_str} is a required attribute for the PropertyUnitTestHarness. "
+                    "Add {obj_str} to the configure function that sets up the test harness.".
+                        format(obj_str=obj_str))
+            obj = getattr(self, obj_str)
+            setattr(m, obj_str, obj)
```
bknueven, line 62: I agree with @andrewlee94, and it doesn't seem we need to do this anymore. Or maybe we should figure out how to re-write this harness so it isn't necessary.

Failing that, we could put these attributes on a single Block (named `_proteuslib_prop_test`?) which we could check to ensure doesn't clash with an existing name on the model.

proteuslib/property_models/tests/property_test_harness.py
```diff
@@ -0,0 +1,532 @@
...
+
+    @pytest.mark.unit
+    def test_scaling(self, frame):
+        m = frame
+
+        calculate_scaling_factors(m.fs)
+
+        # check that all variables have scaling factors
+        unscaled_var_list = list(unscaled_variables_generator(m))
+        assert len(unscaled_var_list) == 0
```
adam-a-a, line 337: I just want to point out that the argument `include_fixed` is set to `False` by default. When testing the RO unit, I remember that setting `unscaled_variables_generator(m.fs.unit, include_fixed=True)` allowed us to catch certain variables that should've been scaled but weren't. That is, we expected the test to catch certain errors, and when it didn't; setting `include_fixed=True` resolved the issue.

However, doing this for the property models will pull up all of the parameters used as coefficients in empirical equations which do not have scaling factors.

adam-a-a (commented):
I only had a minor comment related to one of the tests; it seems like Andrew and Ben already have enough comments to keep you busy. 

I also wanted to mention that you've applied the test harness to the seawater property model but not the NaCl model. I'm guessing that you only intended to get this running first with the seawater property model and would apply to the NaCl model later on.

proteuslib/property_models/tests/property_test_harness.py
```diff
@@ -0,0 +1,368 @@
...
+                    raise Exception(
+                        "Property {v_name} is an on-demand property, but was found "
+                        "on the stateblock without being demanded".format(v_name=v_name))
+
+        # check that properties are built if demanded
+        for v_name in metadata:
+            if metadata[v_name]['method'] is not None:
+                if not hasattr(m.fs.stream[0], v_name):
+                    raise Exception(
+                        "Property {v_name} is included in the metadata but is not found "
```
andrewlee94, line 152: It might be worth rewording this error message to say that the property was not built when called for (i.e. be more specific so that this failure isn't confused with a similar one).

proteuslib/property_models/tests/property_test_harness.py
```diff
@@ -0,0 +1,368 @@
...
+    @pytest.mark.component
+    def test_default_initialization(self, frame_stateblock):
+        m = frame_stateblock
+
+        # initialize
+        m.fs.stream.initialize()
+
+        # check results
+        for (v_name, ind), val in m._test_objs.default_solution.items():
+            var = getattr(m.fs.stream[0], v_name)
```
andrewlee94, line 220: Given that initialize has no way of returning solver status yet, it might also be worth adding a check for the value (i.e. residual) of all constraints.

proteuslib/property_models/tests/property_test_harness.py
```diff
@@ -0,0 +1,368 @@
...
+    @pytest.mark.component
+    def test_badly_scaled(self, frame_stateblock):
+        m = frame_stateblock
+        badly_scaled_var_list = list(badly_scaled_var_generator(m, large=1e2, small=1e-2))
+        if len(badly_scaled_var_list) != 0:
+            lst = []
+            for (var, val) in badly_scaled_var_list:
+                lst.append((var.name, val))
+            raise Exception(
+                "The following variable(s) are badly scaled: {lst}".format(lst=lst))
```
andrewlee94, line 237: I would use "poorly" instead of "badly", and do we want to allow users to define an optional list of know poorly-scaled variables that they are willing to accept?

proteuslib/property_models/tests/property_test_harness.py
```diff
@@ -0,0 +1,368 @@
...
+        m = frame_control_volume
+
+        # scale model
+        calculate_scaling_factors(m)
+
+        # initialize control volume
+        m.fs.cv.initialize()
+
+        # check results, properties_in and properties_out should be the same as default initialize
+        for (v_name, ind), val in m._test_objs.default_solution.items():
```
andrewlee94, line 274: As above, it might be worth checking residuals too.

proteuslib/property_models/tests/property_test_harness.py
```diff
@@ -0,0 +1,368 @@
...
+        for v_str in metadata.keys():
+            getattr(m.fs.stream[0], v_str)
+
+        # scale model
+        calculate_scaling_factors(m)
+
+        # solve model
+        opt = get_solver(self.solver, self.optarg)
+        results = opt.solve(m)
+        assert results.solver.status == SolverStatus.ok
```
andrewlee94, line 349: You should also check TerminationCondition here.

proteuslib/property_models/tests/property_test_harness.py
```diff
@@ -0,0 +1,368 @@
...
+
+        # check results
+        for (v_name, ind), val in m._test_objs.default_solution.items():
+            var = getattr(m.fs.stream[0], v_name)
+            if not pytest.approx(val, rel=1e-3) == value(var[ind]):
+                raise Exception(
+                    "Variable {v_name} with index {ind} is expected to have a value of {val} +/- 0.1%, "
+                    "but it has a value of {val_t}. \nUpdate default_solution dict in the "
+                    "configure function that sets up the PropertyTestHarness".format(
+                        v_name=v_name, ind=ind, val=val, val_t=value(var[ind])))
```
andrewlee94, line 226: Should there be a subsequent test to just solve the initialized block and check solver convergence?

andrewlee94 (changes requested):
A few small comments and suggestions.

proteuslib/property_models/tests/property_test_harness.py
TimBartholomew: Added it

proteuslib/property_models/tests/property_test_harness.py
TimBartholomew: At least for this test, I don't think we want developers to choose what would be poorly scaled here. The default initialization/solve of the property package should be able to be scaled well. Developers can provide scaling factors.

For the regression test, there could be a case to allow developers to have poorly scaled variables, but again also in that case they can choose a scaling factor to make the model well scaled.

proteuslib/property_models/tests/property_test_harness.py
TimBartholomew: Added a function based on the infeasibility tool in pyomo

proteuslib/property_models/tests/property_test_harness.py
andrewlee94: I ask however because there may be cases where it is difficult or impossible to get all variables well scaled. Without a way to flag that, then that model could never pass the integration harness.

Of course, this option should be a last resort, but it would at least be clearly shown in the config setup.

proteuslib/property_models/tests/test_NaCl_prop_pack.py
```diff
@@ -14,337 +14,139 @@
...
-
-        m.fs.stream.initialize()
-
-        results = solver.solve(m)
-
-        # Check for optimal solution
-        assert results.solver.termination_condition == \
-               TerminationCondition.optimal
-        assert results.solver.status == SolverStatus.ok
+class TestSeawaterProperty(PropertyTestHarness):
```
adam-a-a, line 31:
```suggestion
class TestNaClProperty(PropertyTestHarness):
```

proteuslib/property_models/tests/test_NaCl_prop_pack.py
```diff
@@ -14,337 +14,139 @@
...
+@pytest.mark.component
+class TestNaClPropertySolution_1(PropertyRegressionTest):
+    def configure(self):
+        self.prop_pack = props.NaClParameterBlock
+        self.param_args = {}
 
-    @pytest.mark.component
-    def test_solution(self, frame):
-        m = frame
+        self.solver = 'ipopt'
```
adam-a-a, line 64: Since 'ipopt' is the default when using `get_solver()`, is it necessary to make this assignment?

proteuslib/property_models/tests/property_test_harness.py
```diff
@@ -0,0 +1,447 @@
...
+from idaes.core.util.scaling import (calculate_scaling_factors,
+                                     unscaled_variables_generator,
+                                     unscaled_constraints_generator,
+                                     badly_scaled_var_generator)
+from idaes.core.util.initialization import fix_state_vars
+from idaes.core.util import get_solver
+
+
+# -----------------------------------------------------------------------------
+class PropertyTestHarness(object):
```
dangunter, line 41: You don't need to inherit from object in Python 3

proteuslib/property_models/tests/property_test_harness.py
```diff
@@ -0,0 +1,447 @@
...
+        assert isinstance(m.fs.properties.phase_list, Set)
+        assert isinstance(m.fs.properties._phase_component_set, Set)
+
+    @pytest.mark.unit
+    def test_parameters(self, frame_stateblock):
+        m = frame_stateblock
+        # test that the parameter variables are fixed
+        for v in m.fs.properties.component_data_objects(Var):
+            if not v.is_fixed():
+                raise Exception(
```
dangunter, line 102: You should create your own subclass of Exception, e.g. PropertyTestException. You can create multiple subclasses derived from that if you want to get even more specific.

proteuslib/property_models/tests/test_NaCl_prop_pack.py
TimBartholomew: Yeah its not necessary, I guess I included it to show how to change the solver.

andrewlee94 (approved):

bknueven (approved):

adam-a-a (approved):
LGTM

PR #97, opened 2021-05-27
Chlorination process tests

aladshaw3:
## Fixes/Addresses:

This PR is for a test of the reactions associated with the chlorination process. This includes acid reactions for hypochlorous acid and ammonium, as well as complexation of hypochlorous acid with ammonia in solution. These tests also demonstrate the ability of the generic chemistry framework to mix-and-match "inherent" reactions with "equilibrium" reactions in the same unit process. 

## Summary/Motivation:

An important aspect of chlorination chemistry is the scavenging of "free chlorine" by ammonia species in solution. When ammonia is present, it will consume most of the available chlorines to form "combined chlorines" in solution, unit a break-point is reached. At that point, any additional chlorine added will end up as "free chlorine". This is a fairly complex chemical problem for IDAES to deal with as the concentrations of ammonia beyond the "break-point" are extremely low (<10^-20). As such, to get this model to converge, scaling was necessary as was customization of the 'eps' factors in the 'log_power_law_equil' function in IDAES. 

- Added new test file for chlorination process

### Legal Acknowledgement


proteuslib/tests/test_chlorination.py
```diff
@@ -0,0 +1,823 @@
...
+        assert hasattr(model.fs.thermo_params, 'reaction_H2O_Kw')
+        assert hasattr(model.fs.thermo_params.reaction_H2O_Kw, 'eps')
+
+        assert hasattr(model.fs.rxn_params, 'reaction_NH2Cl_K')
+        assert hasattr(model.fs.rxn_params.reaction_NH2Cl_K, 'eps')
+
+        eps = 1e-30
+
+        # Inherent reactions have eps in the 'thermo_params'
+        model.fs.thermo_params.reaction_H2O_Kw.eps.value = eps
```
andrewlee94, line 735: I am not sure what you are testing for here; you set the value for `eps` and then check that it has the value you just assigned.

proteuslib/tests/test_chlorination.py
```diff
@@ -0,0 +1,823 @@
...
+
+        assert model.fs.thermo_params.reaction_H2O_Kw.eps.value == eps
+        assert model.fs.thermo_params.reaction_NH4_Ka.eps.value == eps
+        assert model.fs.thermo_params.reaction_HOCl_Ka.eps.value == eps
+
+        assert model.fs.rxn_params.reaction_NH2Cl_K.eps.value == eps
+        assert model.fs.rxn_params.reaction_NHCl2_K.eps.value == eps
+        assert model.fs.rxn_params.reaction_NCl3_K.eps.value == eps
+
+
```
andrewlee94, line 752: Should only have 1 line break here.

proteuslib/tests/test_chlorination.py
```diff
@@ -0,0 +1,823 @@
...
+        solver.options['max_iter'] = 2
+        results = solver.solve(model)
+        assert results.solver.termination_condition == TerminationCondition.optimal
+        assert results.solver.status == SolverStatus.ok
+
+    @pytest.mark.component
+    def test_solution(self, chlorination_obj):
+        model = chlorination_obj
+
+        assert pytest.approx(300.0016, rel=1e-3) == value(model.fs.unit.outlet.temperature[0])
```
andrewlee94, line 802: I would generally suggest a tolerance of `rel=1e-5` for most tests (solver tolerance is at least 1e-6), unless there is an `eps` involved.

andrewlee94 (changes requested):
A few comments ,mostly about cleaning up stuff you don't need.

Andresj89 (dismissed):
LGTM

proteuslib/tests/test_chlorination.py
aladshaw3: Is there a function I can call to set this value? Should I just not have any asserts after setting the value?

The model is very sensitive to eps and this argument needs to be set for the solve.

proteuslib/tests/test_chlorination.py
```diff
@@ -0,0 +1,823 @@
...
+
+    @pytest.mark.unit
+    def test_build_model(self, chlorination_obj):
+        model = chlorination_obj
+
+        assert hasattr(model.fs.thermo_params, 'component_list')
+        assert len(model.fs.thermo_params.component_list) == 10
+        assert 'H2O' in model.fs.thermo_params.component_list
+        assert hasattr(model.fs.thermo_params, 'H2O')
+        assert isinstance(model.fs.thermo_params.H2O, Solvent)
```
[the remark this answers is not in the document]
andrewlee94: 1) `isinstance` is better than `hasattr` as it both checks that the component exists, and that it has the expected `type`.
2) You need to be careful with both. In these specific cases you are fine, but when dealing with actual properties, you can inadvertently trigger construction.

E.g., consider a case where I do not expect `entr_mol_phase` to exist. If I were to write"

`assert not hasattr(m.entr_mol_phase)`

it will in fact trigger the construction of `entr_mol_phase` (`hasattr` calls `getattr` which builds the component) and thus this test would always fail. Instead, I need to test for a related component that would be built at the same time as `entr_mol_phase`. This can also cause problems if you want to check the value of a component. If the component did not exist before the check, it will be constructed using default values, meaning there is a good chance the value will not be what you expect.

proteuslib/tests/test_chlorination.py
andrewlee94: Ah, OK. Setting the value is fine, but do you really need to have the following `asserts`? I thought you were trying to test something specific.

Andresj89 (approved):
LGTM

andrewlee94 (approved):
Only a couple of minor comments about imports.

proteuslib/tests/test_chlorination.py
```diff
@@ -0,0 +1,775 @@
...
+        NH4 <---> H + NH3
+        HOCl <---> H + OCl
+
+    Equilibrium Reactions:
+        NH3 + HOCl <---> NH2Cl + H2O
+        NH2Cl + HOCl <---> NHCl2 + H2O
+        NHCl2 + HOCl <---> NCl3 + H2O
+'''
+# Importing testing libraries
+import unittest
```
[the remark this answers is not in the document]
aladshaw3: Doesn't look like it is needed. I will just go ahead and merge this, then create a issue to clean up testing files.

proteuslib/tests/test_chlorination.py
```diff
@@ -0,0 +1,775 @@
...
+import pytest
+
+# Importing the object for units from pyomo
+from pyomo.environ import units as pyunits
+
+# Imports from idaes core
+from idaes.core import AqueousPhase
```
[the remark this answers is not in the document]
aladshaw3: Will create an issue to clean up and re-organize testing files.

PR #108, opened 2021-06-14
Test of Remineralization Process

aladshaw3:
## Fixes/Addresses:

This is a unit test of the remineralization process that also covers convergence of "true" and "apparent" species, as well as a test of kinetics and inherent reactions in a CSTR unit model. 

## Summary/Motivation:

Remineralization is one of the major water treatment processes post-RO. This test establishes that the chemistry associated with that process can be adequately solved in the IDAES framework. Additionally, this test also performs a validation test of the "true" vs "apparent" species system in IDAES to confirm that IDAES produces a reasonable "apparent" species set that can produce the correct ion composition when the problem is resolved. Lastly, there are also included within some initial kinetics tests to verify that these modeling components work in other unit models within IDAES. 

- Adds <code> test_remineralization.py </code> unit test 

### Legal Acknowledgement


proteuslib/tests/test_remineralization.py
```diff
@@ -0,0 +1,1543 @@
...
+        pH = -value(log10(model.fs.unit.outlet.mole_frac_comp[0, "H_+"]*total_molar_density))
+        pOH = -value(log10(model.fs.unit.outlet.mole_frac_comp[0, "OH_-"]*total_molar_density))
+
+        assert pytest.approx(8.205408733919795, rel=1e-5) == pH
+        assert pytest.approx(5.795760555786688, rel=1e-5) == pOH
+
+        # Calculate total hardness
+        TH = 2*value(model.fs.unit.control_volume.properties_out[0.0].conc_mol_phase_comp[('Liq', 'Ca_2+')])/1000
+        TH = TH*50000
+        assert pytest.approx(59.57980910667752, rel=1e-5) == TH
```
andrewlee94, line 1231: Minor comment, but it you are only testing to 1e-5 relative tolerance, then you don't need more than 6 significant figures in the value you are testing against.

proteuslib/tests/test_remineralization.py
```diff
@@ -0,0 +1,1543 @@
...
+        check.fs.unit.inlet.flow_mol.fix(10)
+
+        #Custom eps factors
+        eps = 1e-30
+        check.fs.thermo_params.reaction_H2O_Kw.eps.value = eps
+        check.fs.thermo_params.reaction_H2CO3_Ka1.eps.value = eps
+        check.fs.thermo_params.reaction_H2CO3_Ka2.eps.value = eps
+        check.fs.thermo_params.reaction_CO2_to_H2CO3.eps.value = eps
+
+        solver.options['bound_push'] = 1e-20
```
andrewlee94, line 1323: You are using the same (global) solver object, so no need to set these again.

proteuslib/tests/test_remineralization.py
```diff
@@ -0,0 +1,1543 @@
...
+        assert hasattr(model.fs.unit.control_volume.properties_out[0.0], 'scaling_factor')
+        assert isinstance(model.fs.unit.control_volume.properties_out[0.0].scaling_factor, Suffix)
+
+        assert hasattr(model.fs.unit.control_volume.properties_in[0.0], 'scaling_factor')
+        assert isinstance(model.fs.unit.control_volume.properties_in[0.0].scaling_factor, Suffix)
+
+    @pytest.mark.component
+    def test_initialize_solver_cstr_kin(self, remineralization_cstr_kin):
+        model = remineralization_cstr_kin
+        solver.options['bound_push'] = 1e-20
```
andrewlee94, line 1500: As before.

proteuslib/tests/test_remineralization.py
```diff
@@ -0,0 +1,1543 @@
...
+
+        # Non-Carbonate Hardness only exists if there is excess hardness above alkalinity
+        if TH <= CarbAlk:
+            NCH = 0
+        else:
+            NCH = TH - CarbAlk
+        CH = TH - NCH
+        assert pytest.approx(TH, rel=1e-5) == CH
+
+    @pytest.fixture(scope="class")
```
andrewlee94, line 1360: Just to help organise things a little better, I would suggest making this a separate test class. I would also suggest moving the declaration of the config dicts for these test to just immediately before here.

The reason being it helps group information together better to help the reader understand - when I started reading I was wondering why there were two different sets of configs and how you intended to use each of them.

Andresj89 (approved):
LGTM

aladshaw3:
> Only a couple of minor comments on a few things you don't actually need and a suggestion on grouping tests and associated configs together.

Thank you for the suggestions. I will add these to my to do list on the issue of cleaning up tests and make these changes into a future PR. Merging for now to stay on task for June milestone.

PR #119, opened 2021-06-22
Documentation: How to scale a model

TimBartholomew:
## Fixes/Addresses:

None

## Summary/Motivation:
Adds a "how to" for scaling models.

- Creates a how to scale a model rst
- Adds the doc to the toc tree in how to

### Legal Acknowledgement


bknueven (approved):
This is definitely an "advanced" how-to, as it implicitly assumes the reader knows about things like blocks.

I think it should probably come after the two existing how-tos, but we can always reorganize once more documentation is in.

adam-a-a (approved):
Nice job!

PR #120, opened 2021-06-22
correct minor error in NaCl property package documentation

TimBartholomew:
## Fixes/Addresses:

None

## Summary/Motivation:

Fixes an error in setting the default scaling factors for flowrate in the NaCl property package

- Updates the variable and index for the set default scaling factor

### Legal Acknowledgement


bknueven (approved):
Do we have an issue tracking the need for testing the documentation code? If not we should...

adam-a-a:
Now that we're here, it looks like the seawater property model is called instead of the NaCl property model.

adam-a-a (changes requested):
Ready to approve upon revision of line 90

adam-a-a (approved):

bknueven (approved):

PR #122, opened 2021-06-22
Doctest

### LINKED ISSUE #121: Add testing to documentation code
#120 demonstrates the need to have tests for our examples in the documentation.

bknueven:
## Fixes/Addresses:

Fixes #121 

## Summary/Motivation:

We should be testing the code in our documentation

- Adds doctest to GitHub actions
- Set up testing for existing documentation

### Legal Acknowledgement

### DIFF
--- a/docs/technical_reference/property_models/NaCl.rst
+++ b/docs/technical_reference/property_models/NaCl.rst
@@ -78,11 +78,16 @@ This NaCl property package includes support for scaling, such as providing defau
 
 The user can specify the scaling factors for component mass flowrates with the following:
 
-.. code-block:: python
+.. testsetup::
+
+   from pyomo.environ import ConcreteModel
+   from idaes.core import FlowsheetBlock
+
+.. testcode::
    
    # relevant imports
    import proteuslib.property_models.NaCl_prop_pack as props
-   import idaes.core.util.scaling as calculate_scaling_factors
+   from idaes.core.util.scaling import calculate_scaling_factors
 
    # relevant assignments
    m = ConcreteModel()
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -30,7 +30,10 @@ jobs:
         python -m pip install pytest
         python -m pip install git+https://github.com/nawi-hub/idaes-pse@main
         idaes get-extensions --verbose
-        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+        pip install -r requirements-dev.txt
     - name: Test with pytest
       run: |
         pytest
+    - name: Test documentation
+      run: |
+        make -C docs doctest -d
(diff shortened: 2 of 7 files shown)


docs/how_to/how_to_setup_RO_config_options.rst
```diff
@@ -6,7 +6,7 @@ contains options that account for concentration polarization, the mass transfer
 Example: Configure the RO model to account for concentration polarization and pressure drop
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-.. code-block:: python
+.. testcode::
```
adam-a-a, line 9: @bknueven When should we use the `testcode` directive vs. the `testsetup` directive? I noticed that in `how_to_scale_a_model.rst`, `testsetup` was used for a portion of code similar (with a few more lines of code) to this code block.

bknueven: The `testsetup` directive runs, but doesn't display the code, for just the situation where you don't want the user to see the setup to focus on the task being described.

The `testcode` directive displays the code in the documentation, so I replaced `code-block:: python` everywhere I found it with `testcode` so the look of documentation didn't change with this PR. I don't mind the additional imports in the RO config, but they could be done away with as you suggest.

adam-a-a: Thanks for the explanation! 

I think in this particular case, it is useful to show the extra imports needed for the RO config to show the user how to configure RO to account for concentration polarization and pressure drop. So, I don't think you should do away with them here.

adam-a-a (approved):
Looks good to me! I wasn't sure how to go about testing documentation, so thanks for the lesson!

bknueven:
I'd like to bring in the parameter sweep how-to (#123) before this is merged

docs/technical_reference/property_models/NaCl.rst
```diff
@@ -78,11 +78,16 @@ This NaCl property package includes support for scaling, such as providing defau
...
+
+   from pyomo.environ import ConcreteModel
+   from idaes.core import FlowsheetBlock
+
+.. testcode::
    
    # relevant imports
    import proteuslib.property_models.NaCl_prop_pack as props
-   import idaes.core.util.scaling as calculate_scaling_factors
+   from idaes.core.util.scaling import calculate_scaling_factors
```
TimBartholomew, line 90: And this is why we need testing :)
Good catch.

TimBartholomew (dismissed):
@bknueven thanks for demonstrating how code should be tested in documentation!

adam-a-a:
@bknueven I assumed that you wanted to add testing to the how-to for parameter sweep, so I rescinded my approval to let you do that first. If you'd rather get this merged as is though, let me know and I'll approve again.

TimBartholomew (approved):
LGTM

adam-a-a (approved):

PR #130, opened 2021-06-23
Update permeate properties for 0D RO

### LINKED ISSUE #125: Update permeate properties for 0D RO
The RO model would be improved if the permeate concentration for the flux calculation was determined based on the instantaneous concentration from the water and salt flux, as opposed to the current single unit outlet concentration.

This change will involve creating 2 additional stateblocks for these instantaneous concentrations, one for the feed-side inlet and one for the outlet

TimBartholomew:
## Fixes/Addresses:

Fixes #125 

## Summary/Motivation:
The current 0D RO implementation models the permeate concentration with one permeate block. The idea was that since permeate is assumed to be crossflow and its effect on the water flux is low, that it was valid to use one mixed value for the permeate concentration for calculating the water flux at the feed-side inlet and outlet. However, this results in some unintended consequences for LSRRO analysis which has significant salt flux and therefore permeate concentration is impactful. 

A better modeling approach for crossflow permeate is to use the instantaneous permeate concentration calculated from the salt and water flux at both the inlet and outlet. This necessitates the addition of 2 new state blocks.

- Created a control volume for the permeate side that creates the 2 new stateblocks, but does not include mass and energy balances. Essentially the control volume is created to organize the 3 total permeate stateblocks.
- Moved the properties_permeate stateblock to the new permeate side control volume
- Updated tests

### Legal Acknowledgement


TimBartholomew:
@OOAmusat @eyoung55 @jefalon, this PR will have a minor impact your flowsheets once merged. The permeate stateblock is now located on a control volume called permeate_side. So you will need to update everywhere that has `RO.propeties_permeate` to `RO.permeate_side.properties_permeate`

proteuslib/unit_models/reverse_osmosis_0D.py
```diff
@@ -537,16 +547,15 @@ def recovery_mass_phase_comp_bounds(b, t, p, j):
             self.flowsheet().config.time,
             doc="Material properties of feed-side interface at outlet",
             default=tmp_dict)
-        # Permeate properties
-        self.properties_permeate = self.config.property_package.state_block_class(
+        self.permeate_side.properties_permeate = self.config.property_package.state_block_class(
             self.flowsheet().config.time,
             doc="Material properties of permeate",
```
adam-a-a, line 552: Is this the state block for the mixed (average) permeate?
```suggestion
            doc="Material properties of mixed permeate",
```

proteuslib/unit_models/reverse_osmosis_0D.py
```diff
@@ -537,16 +547,15 @@ def recovery_mass_phase_comp_bounds(b, t, p, j):
...
             default=tmp_dict)
 
         # Add Ports
         self.add_inlet_port(name='inlet', block=self.feed_side)
         self.add_outlet_port(name='retentate', block=self.feed_side)
-        self.add_port(name='permeate', block=self.properties_permeate)
+        self.add_port(name='permeate', block=self.permeate_side.properties_permeate)
```
adam-a-a, line 558: Is there any difference here between using the `add_port` method as opposed to the `add_outlet_port` method?

proteuslib/unit_models/reverse_osmosis_0D.py
```diff
@@ -619,20 +629,65 @@ def eq_flux_io(b, t, x, p, j):
...
         def eq_permeate_isothermal(b, t):
             return b.feed_side.properties_out[t].temperature == \
-                   b.properties_permeate[t].temperature
+                   b.permeate_side.properties_permeate[t].temperature
+
+        # # Permeate-side instantaneous stateblocks
+        @self.permeate_side.Constraint(self.flowsheet().config.time,
+                                   self.io_list,
+                                   self.solute_list,
+                                   doc="Permeate instantaneous mass fraction")
```
adam-a-a, line 651: To be consistent with other inlet/outlet variables, I'd remove the word "instantaneous" in all places, including the "instant" qualifier.
```suggestion
                                   doc="Permeate mass fraction at inlet and outlet")
```

proteuslib/unit_models/reverse_osmosis_0D.py
```diff
@@ -654,12 +709,9 @@ def eq_concentration_polarization_io(b, t, x, j):
                         prop_io.conc_mass_phase_comp['Liq', j]
                         * self.cp_modulus[t, j])
             elif self.config.concentration_polarization_type == ConcentrationPolarizationType.calculated:
-                jw = self.flux_mass_io_phase_comp[t, x, 'Liq', 'H2O'] / self.dens_solvent
+                jw = self.flux_mass_io_phase_comp[t, x, 'Liq', self.solvent_list[1]] / self.dens_solvent
```
adam-a-a, line 712: I think I had originally changed this to `self.solvent_list[1]` here like you did. 

Then, @andrewlee94 recommended the use of `self.solvent_list.first()` and then upon further reflection suggested using `H2O` directly here since RO is only setup for that. A small detail, but all that just to suggest that we should keep what was there.

proteuslib/unit_models/reverse_osmosis_0D.py
```diff
@@ -823,29 +875,29 @@ def eq_equal_flow_vol_interface_io(b, t, x):
...
         def eq_rejection_phase_comp(b, t, j):
             return (b.rejection_phase_comp[t, 'Liq', j] ==
-                    1 - (b.properties_permeate[t].conc_mass_phase_comp['Liq', j] /
+                    1 - (b.permeate_side.properties_permeate[t].conc_mass_phase_comp['Liq', j] /
                          b.feed_side.properties_in[t].conc_mass_phase_comp['Liq', j]))
 
         @self.Constraint(self.flowsheet().config.time)
         def eq_over_pressure_ratio(b, t):
             return (b.feed_side.properties_out[t].pressure ==
                     b.over_pressure_ratio[t]
```
adam-a-a, line 898: Technically, we can make `over_pressure_ratio` an inlet/outlet variable too.

proteuslib/unit_models/tests/test_reverse_osmosis_0D.py
```diff
@@ -313,6 +325,10 @@ def test_initialize(self, RO_frame):
     def test_var_scaling(self, RO_frame):
         m = RO_frame
         badly_scaled_var_lst = list(badly_scaled_var_generator(m))
+        for (var, v) in badly_scaled_var_lst:
```
adam-a-a, line 328: It looks like this was here for debugging purposes and should be removed (although I am not against leaving it in).

adam-a-a (changes requested):
Just a couple of minor suggestions/questions to address, but more importantly, several checks are failing.

adam-a-a:
I tried fixing the check errors that we're seeing but to no avail. When I pull the branch and run `pytest`, all tests pass. However, the failed builds seem to occur in the RO flowsheet, starting at `test_set_operating_conditions`

proteuslib/unit_models/reverse_osmosis_0D.py
adam-a-a: I think I realized the difference after the fact.

 `self.permeate_side.properties_permeate` is a state block, thus the use of `add_port()` (has no inlet/outlet). Meanwhile, `self.feed_side` is a control volume, hence the use of `add_outlet_port`.

proteuslib/unit_models/reverse_osmosis_0D.py
TimBartholomew: I included your suggested change and renamed `permeate_side.properties_permeate` to `permeate_side.properties_mixed`, I think that will be more clear and not so redundant.

proteuslib/unit_models/reverse_osmosis_0D.py
TimBartholomew: Yeah it could be calculated but generally when people talk about over pressure they are just talking about where the driving force is the lowest. So I think making it be just the outlet makes sense.

adam-a-a:
My only concern now is how this version of the RO model will effect the stability of the LSRRO model. However, I suppose that can be addressed in another PR in response to #132. At least this addresses the problem of assuming a single, average permeate concentration in LSRRO.

adam-a-a (approved):

TimBartholomew:
> My only concern now is how this version of the RO model will effect the stability of the LSRRO model. However, I suppose that can be addressed in another PR in response to #132. At least this addresses the problem of assuming a single, average permeate concentration in LSRRO.

I'm not sure if the model is necessarily less stable than the previous version. We should definitely have @eyoung55 regenerate the map for LSRRO though.

bknueven (approved):
Just wanted to be sure this passed with the new version of IDAES we're releasing against.

Issue #126, opened 2021-06-22
Master issue: 1D-RO implementation


adam-a-a:
We need a 1D implementation of the RO model. 

First approach: 

- Implement 1D control volumes for feed and permeate sides of membrane
- don't add material, energy, momentum balances for the permeate (assuming this will eliminate effects of flow directionality on the permeate side)
- Start with simplified RO case; i.e., no concentration polarization, no pressure drop.

**Task list:**

- [x] Add config options to unit model class
- [x] Add `build` method to unit model class
- [x] Add `_make_performance `method to unit model class (containing variables and constraints)
- [x] Add `initialize `method to unit model class
- [x] Add `calculate_scaling_factors `method to unit model class
- [x] Create tests for unit model
- [x] Reiterate all steps until working version complete
- [x] Add/revise `get_costing `method as needed
- [x] Add `_get_performance_contents `for `report()` method
- [x] Add `_get_stream_table `for `report()` method

EDIT: Per @andrewlee94's recommendation, and because I was having trouble with my initial attempt, I switched the approach from using 1d control volumes for both sides of the membrane to just having a 1d CV for the feed side and an indexed stateblock for the permeate side (with one more stateblock for the final permeate at the module outlet).

adam-a-a:
@andrewlee94 I might have to ping you soon to chat about developing 1D models with IDAES, when you have the time. For now, I'm taking a first shot at developing the 1DRO model by following some of the 1D models in the idaes-pse repo. 

To give an initial idea of the approach, I am setting up the feed-side of the membrane and the permeate-side of the membrane each as a 1D control volume, but the permeate side should have no directional flow (only mass transfer from feed side at each finite element). At the moment, one thing that I'm unsure about is whether I need to (and how to) multiply the `mass_transfer_term` by dx since it seems that the mass transfer term (and most other terms from the 1D CV) are per unit length. Looking at other 1D models, this wasn't clear to me.

andrewlee94:
First of all, I don't know if you need a 1-D control volume for the permeate flow.
Secondly, yes, the mass transfer will need to be multiplied by dx as it is per unit length.

adam-a-a:
> First of all, I don't know if you need a 1-D control volume for the permeate flow.
> 
> Secondly, yes, the mass transfer will need to be multiplied by dx as it is per unit length.

[...]

PR #137, opened 2021-06-24
First draft of 1DRO unit model

adam-a-a:
## Summary/Motivation:

This is a first attempt at developing the 1D-RO model. While more detail will be added subsequently, a simple version of the 1D-RO model is presented here.

- 1D control volume on feed side
- Indexed state block for permeate side
- State block for the final permeate exiting the module
- initial testing of the model
- initial technical spec document as a placeholder (some simple info was filled in, but will need more work in future PR)

### Legal Acknowledgement


proteuslib/unit_models/reverse_osmosis_1D.py
```diff
@@ -0,0 +1,824 @@
...
+            doc="""Discretization method to use for DAE transformation. See Pyomo
+    documentation for supported transformations."""))
+
+    CONFIG.declare("transformation_scheme", ConfigValue(
+            default=useDefault,
+            description="Discretization scheme to use for DAE transformation",
+            doc="""Discretization scheme to use when transforming domain. See
+    Pyomo documentation for supported schemes."""))
+
+
```
adam-a-a, line 136:
```suggestion
```

proteuslib/unit_models/reverse_osmosis_1D.py
```diff
@@ -0,0 +1,800 @@
...
+            doc="""Indicates what type of momentum balance should be constructed,
+    **default** - MomentumBalanceType.pressureTotal.
+    **Valid values:** {
+    **MomentumBalanceType.none** - exclude momentum balances,
+    **MomentumBalanceType.pressureTotal** - single pressure balance for material,
+    **MomentumBalanceType.pressurePhase** - pressure balances for each phase,
+    **MomentumBalanceType.momentumTotal** - single momentum balance for material,
+    **MomentumBalanceType.momentumPhase** - momentum balances for each phase.}"""))
+
+    CONFIG.declare("finite_elements", ConfigValue(
```
andrewlee94, line 179: Just as a matter of organization, I would keep all the DAE related arguments together. At the moment, the default balance type arguments are stuck in the middle of them.

proteuslib/unit_models/reverse_osmosis_1D.py
```diff
@@ -0,0 +1,800 @@
...
+
+        Returns:
+            None
+        """
+        # Call UnitModel.build to setup dynamics
+        super().build()
+
+        self.scaling_factor = Suffix(direction=Suffix.EXPORT)
+
+        self.solvent_list = Set()
```
andrewlee94, line 248: You should be able to get these from the property package, e.g. `self.config.property_package.solvent_set`

proteuslib/unit_models/reverse_osmosis_1D.py
```diff
@@ -0,0 +1,800 @@
...
+    **Valid values:** {
+    **MaterialBalanceType.useDefault - refer to property package for default
+    balance type
+    **MaterialBalanceType.none** - exclude material balances,
+    **MaterialBalanceType.componentPhase** - use phase component balances,
+    **MaterialBalanceType.componentTotal** - use total component balances,
+    **MaterialBalanceType.elementTotal** - use total element balances,
+    **MaterialBalanceType.total** - use total material balance.}"""))
+
+    CONFIG.declare("energy_balance_type", ConfigValue(
```
andrewlee94, line 151: Later on you set energy balance type to `None`. Either don't add it here as an option, or make use of this argument later (which would require supporting energy balances).

proteuslib/unit_models/reverse_osmosis_1D.py
```diff
@@ -0,0 +1,800 @@
...
+
+        feed_side = self.feed_side
+        # Add geometry to feed side
+        feed_side.add_geometry()
+        # Add state blocks to feed side
+        feed_side.add_state_blocks(has_phase_equilibrium=False)
+        # Populate feed side
+        feed_side.add_material_balances(balance_type=self.config.material_balance_type,
+                                        has_mass_transfer=True)
+        feed_side.add_energy_balances(balance_type=EnergyBalanceType.none)
```
andrewlee94, line 273: As noted above.

However, also note that the underlying bug in the ideal property package was fixed such that you should be able to do an energy balance now and not see excessive temperature change due to a change in pressure.

proteuslib/unit_models/reverse_osmosis_1D.py
```diff
@@ -0,0 +1,800 @@
...
+        feed_side.add_momentum_balances(balance_type=self.config.momentum_balance_type,
+                                        has_pressure_change=self.config.has_pressure_change)
+        # Apply transformation to feed side
+        feed_side.apply_transformation()
+        # Add inlet/outlet ports for feed side
+        self.add_inlet_port(name="feed_inlet", block=feed_side)
+        self.add_outlet_port(name="feed_outlet", block=feed_side)
+        #TODO: Make permeate_side.permeate_out a ProductBlock instead?
```
andrewlee94, line 281: No, you don't want a ProductBlock (although it would work). Feed and Product blocks are just wrappers around a StateBlock anyway, with an appropriately named Port (inlet for a Product block) and some extra code to allow users to do `Product.state_var.fix`.

proteuslib/unit_models/reverse_osmosis_1D.py
```diff
@@ -0,0 +1,800 @@
...
+        self.add_port(name="permeate_outlet", block=self.permeate_out)
+
+        # ==========================================================================
+        """ Add references to control volume geometry."""
+        add_object_reference(self, 'length', feed_side.length)
+        add_object_reference(self, 'feed_area_cross', feed_side.area)
+
+        # Add reference to pressure drop for feed side only
+        if (self.config.has_pressure_change is True and
+                self.config.momentum_balance_type != 'none'):
```
andrewlee94, line 307: Need to use the MomentumBlaanceType Enum here.

proteuslib/unit_models/reverse_osmosis_1D.py
```diff
@@ -0,0 +1,800 @@
...
+        self.area = Var(
+            initialize=10,
+            bounds=(1e-1, 1e3),
+            domain=NonNegativeReals,
+            units=units_meta('length')**2,
+            doc='Membrane area')
+
+        # TODO: Add initial value, bounds and real equation later. Just need for now since add_geometry() generates
+        #  cross-sectional area var that I am referencing
+        # self.feed_area_cross = Var(
```
andrewlee94, line 394: Is this code needed in the released version?

proteuslib/unit_models/reverse_osmosis_1D.py
```diff
@@ -0,0 +1,800 @@
...
+        #     doc='Cross-sectional area of feed channel')
+
+        self.width = Var(
+            initialize=1,
+            bounds=(1e-1, 1e3),
+            domain=NonNegativeReals,
+            units=units_meta('length'),
+            doc='Membrane width')
+
+        # @self.Constraint(doc="Cross-sectional area of feed channel")
```
andrewlee94, line 415: Again, commented code.

proteuslib/unit_models/reverse_osmosis_1D.py
```diff
@@ -0,0 +1,800 @@
...
+        #     return b.feed_area_cross == 1 * 1 * b.width  # TODO: add channel_height and spacer_porosity
+        #
+        # @self.Constraint(doc="Cross-sectional area of permeate channel")
+        # def eq_permeate_area_cross(b):
+        #     return b.permeate_area_cross == 1 * 1 * b.width  # TODO: add channel_height and spacer_porosity
+
+        # mass transfer
+        def mass_transfer_phase_comp_initialize(b, t, x, p, j):
+            return value(self.feed_side.properties[t, x].get_material_flow_terms('Liq', j)
+                         * 0.5) # self.recovery_mass_phase_comp[t, 'Liq', j])
```
andrewlee94, line 426: Trailing comment,

proteuslib/unit_models/reverse_osmosis_1D.py
```diff
@@ -0,0 +1,800 @@
...
+
+        self.mass_transfer_phase_comp = Var(
+            self.flowsheet().config.time,
+            self.feed_side.length_domain,
+            self.config.property_package.phase_list,
+            self.config.property_package.component_list,
+            initialize=0.1,  # mass_transfer_phase_comp_initialize,
```
andrewlee94, line 433: Trailing comment,

proteuslib/unit_models/reverse_osmosis_1D.py
```diff
@@ -0,0 +1,800 @@
...
+                         self.feed_side.length_domain,
+                         self.config.property_package.phase_list,
+                         self.config.property_package.component_list,
+                         doc="Mass transfer from feed to permeate")
+        def eq_connect_mass_transfer(b, t, x, p, j):
+            if x == b.feed_side.length_domain.first():
+                return Constraint.Skip
+            else:
+                return (b.permeate_side[t, x].get_material_flow_terms(p, j)
+                        == -b.feed_side.mass_transfer_term[t, x, p, j] * b.length / nfe)
```
andrewlee94, line 568: Be careful here - this is only the number of finite elements. If the user selected a collocation transformation there will be more points than this.

`len(self.feed_side.length_domain)-1` might be more useful.

proteuslib/unit_models/reverse_osmosis_1D.py
```diff
@@ -0,0 +1,800 @@
...
+            self.flowsheet().config.time,
+            self.feed_side.length_domain,
+            self.config.property_package.phase_list,
+            self.config.property_package.component_list,
+            initialize=0.1,  # mass_transfer_phase_comp_initialize,
+            bounds=(1e-8, 1e6),
+            domain=NonNegativeReals,
+            units=units_meta('mass') * units_meta('time')**-1 * units_meta('length')**-1,
+            doc='Mass transfer to permeate')
+        # ==========================================================================
```
andrewlee94, line 438: More commented code.

proteuslib/unit_models/reverse_osmosis_1D.py
```diff
@@ -0,0 +1,800 @@
...
+            outlvl=outlvl,
+            optarg=optarg,
+            solver=solver,
+            state_args=feed_side_args)
+
+        flags_permeate_side = blk.permeate_side.initialize(
+            outlvl=outlvl,
+            optarg=optarg,
+            solver=solver,
+            state_args=permeate_side_args)
```
andrewlee94, line 646: Extension for the future, but is there a way you could estimate some of these based on the feed state? Notably, once you have the permeate side, it is probably possible to directly calculate the permeate outlet state.

proteuslib/unit_models/reverse_osmosis_1D.py
```diff
@@ -0,0 +1,800 @@
...
+        # Step 2: Solve unit
+        with idaeslog.solver_log(solve_log, idaeslog.DEBUG) as slc:
+            results = solve_indexed_blocks(opt, [blk], tee=slc.tee)
+
+        with idaeslog.solver_log(solve_log, idaeslog.DEBUG) as slc:
+            res = opt.solve(blk, tee=slc.tee)
+
+    def _get_performance_contents(self, time_point=0):
+        pass
+        # var_dict = {}
```
andrewlee94, line 666: This should be cleaned up.

proteuslib/unit_models/reverse_osmosis_1D.py
```diff
@@ -0,0 +1,800 @@
...
+        # var_dict["Tube Outer Diameter"] = self.d_tube_outer
+        # var_dict["Tube Inner Diameter"] = self.d_tube_inner
+        # var_dict["Tube Length"] = self.tube.length
+        # var_dict["Number of Tubes"] = self.N_tubes
+        #
+        # return {"vars": var_dict}
+
+    def _get_stream_table_contents(self, time_point=0):
+        pass
+        # return create_stream_table_dataframe(
```
andrewlee94, line 680: As above.

proteuslib/unit_models/reverse_osmosis_1D.py
```diff
@@ -0,0 +1,800 @@
...
+# the U.S. Dept. of Energy). All rights reserved.
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and license
+# information, respectively. These files are also available online at the URL
+# "https://github.com/nawi-hub/proteuslib/"
+#
+###############################################################################
+
+
+from enum import Enum
```
andrewlee94, line 15: You do not appear to be using this import.

proteuslib/unit_models/reverse_osmosis_1D.py
```diff
@@ -0,0 +1,800 @@
...
+# Import Pyomo libraries
+from pyomo.environ import (Var,
```
andrewlee94, line 17: Check your imports, I am pretty certain there area few here you don't use.

E.g. I do not recall seeing `Reference` used (you used `add_object_reference` instead.

andrewlee94 (changes requested):
I haven't looked at the docs or the test yet, but here are some comments on the unit model.

proteuslib/unit_models/reverse_osmosis_1D.py
adam-a-a: I think that `solvent_set` and `solute_set` weren't defined in the NaCl property model, so would have to modify the property model for that to work.

andrewlee94: That is true. That said, from the IDAES standards perspective at least, it is the property package's job to define these sorts of things.

proteuslib/unit_models/reverse_osmosis_1D.py
adam-a-a: For now, I will delete the option and add support for energy balances after I get other parts working.

proteuslib/unit_models/reverse_osmosis_1D.py
adam-a-a: Yes, I definitely would like to improve this. This is my first time writing an initialization routine like this, so still figuring it out things as I go. Would you happen to have any recommendations for resources on developing (getting started with developing) initialization routines? So far, I've just looked at several unit models in IDAES and followed along with the logic.

proteuslib/unit_models/reverse_osmosis_1D.py
TimBartholomew: Yes the solvent_set and solute_set should be defined in the property package. It really should be automatically created from defining the components like the component_list and phase_list on IDAES-side. I was planning on making a PR for IDAES, but it got deprioritized.

proteuslib/unit_models/reverse_osmosis_1D.py
andrewlee94: Unfortunately, there are no good resources I am aware of. As far as examples go, the gas-solids contactors have some of the most complex initialisation routines, but they might also not be easy to follow.

proteuslib/unit_models/reverse_osmosis_1D.py
andrewlee94: The generic property framework already does this, so it is only the custom property packages which lack these, Making a general PR would need to keep in mind how this would impact the generic framework.

proteuslib/unit_models/reverse_osmosis_1D.py
adam-a-a: I was planning to use this when I add in options for concentration polarization effects and pressure drop assumptions. I could delete now but would be bringing it right back in a subsequent PR (or perhaps just build on this one?)

andrewlee94: It is generally good practice to only include imports you use. You add the import back once you add the new options.

proteuslib/unit_models/reverse_osmosis_1D.py
adam-a-a: For now, I will leave this be until we move forward on a PR for the property model. Also want to note that we'd have to revise this segment of code in the 0D RO model as well.

adam-a-a:
@TimBartholomew- before I go ahead with enhancements (adding CP, pressure drop), I am contemplating whether it still makes sense to have _all_ of the same options for the 1d model as we do for the 0D RO model.

In terms of consistency, it would be good to apply the same options to the 1D. Additionally, users could take results they may already have from their own calculations (where simplifying assumptions were previously applied) and compare them with results from the 1D model in Proteuslib.

On the other hand, while allowing the user to fix the CP modulus, for example, is meant to facilitate the use of simplifying assumptions, would such simplifications be necessary when the user has chosen to implement a 1-D model? In other words, we've already taken a step up in complexity by shifting to a 1-D model, so does it still make sense to support the same simplifying assumptions as before? What are your thoughts on this?

adam-a-a:
@andrewlee94 @TimBartholomew I made most of the requested changes (haven't modified the initialization method just yet). I also add content to the two methods tied to the report() method. Good thing I did this because I then observed that the temperature at the outlet of the feed channel was ~ 360 C when the inlet was set to 298 C. 

I suppose I didn't add any isothermal constraints for temperature along the feed channel but didn't realize I needed to. Also, I excluded the addition of energy balances, so I am guessing that adding that back in might resolve the issue. My only concern is getting the DOF around 0, which I had to fiddle with a bit.

proteuslib/unit_models/reverse_osmosis_1D.py
adam-a-a: Not until I add on additional variables and constraints for detailed calculations that account for concentration polarization (which will be in a subsequent PR).

proteuslib/unit_models/reverse_osmosis_1D.py
```diff
@@ -0,0 +1,726 @@
...
+
+        feed_side = self.feed_side
+        # Add geometry to feed side
+        feed_side.add_geometry()
+        # Add state blocks to feed side
+        feed_side.add_state_blocks(has_phase_equilibrium=False)
+        # Populate feed side
+        feed_side.add_material_balances(balance_type=self.config.material_balance_type,
+                                        has_mass_transfer=True)
+        # feed_side.add_energy_balances(balance_type=self.config.energy_balance_type,
```
adam-a-a, line 265:
```suggestion
```

proteuslib/unit_models/reverse_osmosis_1D.py
```diff
@@ -0,0 +1,726 @@
...
+        #                               has_enthalpy_transfer=True)
```
adam-a-a, line 265:
```suggestion
```

docs/technical_reference/unit_models/reverse_osmosis_1D.rst
```diff
@@ -0,0 +1,169 @@
...
+   "Components", ":math:`j`", "['H2O', 'NaCl']*"
+
+\*Solute depends on the imported property model; example shown here is for the NaCl property model.
+
+.. note::
+
+    Everything below this section is pending since the 1D-RO model is currently under development.
+
+Variables
+----------
```
TimBartholomew, line 51: Instead of duplicating everything, I think it would be best to just refer readers to the 0D RO model and just note the differences between the 0D and 1D model here. That way we only have one place to change the documentation when we make a change in some variable or equation.

proteuslib/unit_models/reverse_osmosis_1D.py
```diff
@@ -0,0 +1,724 @@
...
+        tmp_dict = dict(**self.config.property_package_args)
+        tmp_dict["has_phase_equilibrium"] = False
+        tmp_dict["parameters"] = self.config.property_package
+        tmp_dict["defined_state"] = False  # these blocks are not inlets
+        self.permeate_side = self.config.property_package.state_block_class(
+            self.flowsheet().config.time,
+            self.feed_side.length_domain,
+            doc="Material properties of permeate along permeate channel",
+            default=tmp_dict)
+        self.permeate_out = self.config.property_package.state_block_class(
```
TimBartholomew, line 282: We may want to call this permeate_mixed for consistency with the 0D RO model. Although I'm up for changing names.

proteuslib/unit_models/reverse_osmosis_1D.py
```diff
@@ -0,0 +1,724 @@
...
+        feed_side.apply_transformation()
+        # Add inlet/outlet ports for feed side
+        self.add_inlet_port(name="feed_inlet", block=feed_side)
+        self.add_outlet_port(name="feed_outlet", block=feed_side)
+        # Make indexed stateblock and separate stateblock for permeate-side and permeate outlet, respectively.
+        tmp_dict = dict(**self.config.property_package_args)
+        tmp_dict["has_phase_equilibrium"] = False
+        tmp_dict["parameters"] = self.config.property_package
+        tmp_dict["defined_state"] = False  # these blocks are not inlets
+        self.permeate_side = self.config.property_package.state_block_class(
```
TimBartholomew, line 277: I think calling these stateblocks "permeate_side" could be confusing because "permeate_side" is a control volume in the 0D model. We should definitely decide whether we want the permeate side to have control volume without mass and energy balances or not and be consistent across the 1D and 0D model.

proteuslib/unit_models/reverse_osmosis_1D.py
```diff
@@ -0,0 +1,724 @@
...
+        add_object_reference(self, 'feed_area_cross', feed_side.area)
+
+        # Add reference to pressure drop for feed side only
+        if (self.config.has_pressure_change is True and
+                self.config.momentum_balance_type != MomentumBalanceType.none):
+            add_object_reference(self, 'deltaP', feed_side.deltaP)
+
+        self._make_performance()
+
+    def _make_performance(self):
```
TimBartholomew, line 301: We don't have a similar internal function for variable and constraint creation in the 0D RO model, is this something that was implemented in another 1D model or is necessary for 1D models?

proteuslib/unit_models/reverse_osmosis_1D.py
```diff
@@ -0,0 +1,724 @@
...
+    def _get_performance_contents(self, time_point=0):
+        var_dict = {}
+        var_dict["Membrane Area"] = self.area
+        var_dict["Membrane Length"] = self.length
+        var_dict["Membrane Width"] = self.width
+        #TODO: add more vars
+        return {"vars": var_dict}
+
+    def _get_stream_table_contents(self, time_point=0):
+        pass
```
TimBartholomew, line 606: I think you need to remove pass here

TimBartholomew (commented):
Some comments. I think the main thing is we should make sure that 0D and 1D model are consistent, we may plan to change the 0D model to match what is here.

proteuslib/unit_models/reverse_osmosis_1D.py
adam-a-a: It was something that I came across in some IDAES models and seems to be a matter of preference. I implemented that approach here for my own readability, but we can revert this back to the way vars and constraints are implemented in 0DRO (or modify 0DRO to match this). However, I'd prefer to get the 1DRO model working fully before making such changes.

andrewlee94: Opinion is somewhat divided in IDAES. Most models don't do it, but they are fairly simple and were all written by me. Others seem to prefer having these sub-methods, and they do have some advantages for testing (if you unit test each method separately, which no-one does).

docs/technical_reference/unit_models/reverse_osmosis_1D.rst
adam-a-a: That works for me. This was meant to be more of a placeholder until the 1dRO model was finished. I'd address this in a subsequent PR.

proteuslib/unit_models/reverse_osmosis_1D.py
adam-a-a: During my first attempt to implement the 1DRO model, I tried setting the permeate_side as a 1D control volume, but it made the implementation trickier. Furthermore, after some discussion in issue #126, I decided to pivot and set the permeate_side as an indexed stateblock. I think this is something we can revisit in a subsequent PR, once the 1DRO model is fully functional. The same applies for any name changing or consistency with the 0DRO model; let's reserve such changes for when functionality of the 1DRO in place. Then we can have a PR that cleans up discrepancies between the 0D and 1D models where possible.

andrewlee94: I personally wouldn't be too concerned. If someone is digging in deep enough to find this, then they had better know enough to understand what is happening anyway.

proteuslib/unit_models/reverse_osmosis_1D.py
adam-a-a: Similar to my other response, I think we should have a separate PR (once 1DRO is fully functional) that addresses naming consistency (and things of that nature) between 0D and 1D.

proteuslib/unit_models/reverse_osmosis_1D.py
adam-a-a, line 606:
```suggestion
```

docs/technical_reference/unit_models/reverse_osmosis_1D.rst
adam-a-a: Actually, just added the links to the relevant 0DRO doc sections until all vars/constraints are settled on (shouldn't be much difference, if any).

adam-a-a:
@TimBartholomew @andrewlee94 I think I addressed all of your comments. With regard to naming consistency and similar issues between 0DRO and 1DRO, that would be settled in a later PR. Right now, I'd like to get this PR through since I have another PR for 1DRO ready to go. Do you have any other  other comments or suggested changes?

TimBartholomew (approved):
This looks fine and we can address the consistency across 0D and 1D RO when we decide on how we are refactoring the 0D model.

andrewlee94 (approved):

Issue #113, opened 2021-06-17
Documentation for June release


ksbeattie:
- [x] Welcome/overview page (@TimBartholomew) (#127)
- [x] #145
- [x] Background (@TimBartholomew) (#124)
- [x] How-to on Chemistry (@aladshaw3) (#129)
- [x] Parameter Sweep tool (@eyoung55) (#123)
- [x] Electrolyte DB (@dangunter) (#77)
- [x] Jupyter Notebook with simple RO model (@bknueven, @lbianchi-lbl, @adam-a-a) (#117)

PR #144, opened 2021-06-30
Add first draft of installation guide

lbianchi-lbl:
## Fixes/Addresses:
 
Last remaining item of #113 (which apparently got converted to #145 when I clicked on it?)

## Summary/Motivation

This PR contains the first draft for the installation guide for ProteusLib. The current structure is the following:

- *Introduction/Using Conda environments*
  - Covers what is a Conda environment, why it is the recommended way to install ProteusLib, how to install Conda
- *General installation*
  - Covers how to install `proteuslib` in the general/default case of read-only, or *user-mode* installation, i.e. without needing to clone the ProteusLib repository
- *For ProteusLib developers*
  - Covers how to install `proteuslib` in *developer mode*, i.e. from a clone of the ProteusLib repo that will then be used to contribute back to the ProteusLib codebase

### Motivation/thoughts for reviewers

**TL;DR/Summary: A good installation guide is hard to write well and (partially) project-specific, so feel free to be honest in the review regarding both form and substance in this guide**

- Installation guides are in my experience very tricky to get right, as there is a number of prerequisites to install and/or configure before the user is able to install the actual software
- Following the Diataxis nomenclature for docs, they're arguably closest to a Tutorial in their goals (especially installation guides in a "Getting started" section)
- On the other hand, I find that the sheer number of unrelated steps and components involved can be confusing for the user if they're provided without any context of what they're used for
  - See e.g. a (perfectly valid) question of "do I need to use Conda/virtual environments to install ProteusLib?"
- Precisely because it's tricky, we thought the most reasonable thing to do would be to have a draft of the installation guide to get the conversation started (which is useable, meaning that the steps in it should work), and then iterate during the review process
- **Therefore, I encourage everyone to comment and provide feedback, questions, or corrections to this, and to consider any of the choices described there up for discussion**
- The basic structure (i.e. what's currently in the PR) is informed by our experience and general practises for Python projects, but every project has slightly different requirements and "developer audience", and it makes sense that the general installation procedures to be adapted to match that

- Add installation guide to the docs as dedicated page in the "Getting Started" section

### Legal Acknowledgement

### DIFF
--- a/docs/getting_started/install.rst
+++ b/docs/getting_started/install.rst
@@ -53,11 +53,12 @@ If you are going to use ProteusLib's functionality, but *do not* plan to contrib
 
 		pip install https://github.com/nawi-hub/proteuslib/archive/main.zip
 
-#. To verify that the installation was successful, try running the ProteusLib test suite using ``pytest``:
+#. To verify that the installation was successful, open a Python interpreter and try importing some of ProteusLib's modules, e.g.:
 
 	.. code-block:: shell
 
-		pytest
+		python
+		>>> from proteuslib.unit_models import *
 
 For ProteusLib developers
 -------------------------
--- a/docs/getting_started/install.rst
+++ b/docs/getting_started/install.rst
@@ -1,0 +1,101 @@
+Installing ProteusLib
+=====================
+
+Introduction
+------------
+
+.. _about-conda:
+
+Using Conda environments
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Conda environments are a way to create and manage multiple sets of packages and/or Python versions on the same system without incurring into conflicts.
+Each Conda environment is a dedicated directory, separate from other Conda environments and the operating system's own directories, containing its own collection of packages, executables, and Python installation, including the Python interpreter.
+Once a Conda environment is *activated*, using the ``conda activate`` command in a terminal/console, the environment's own version of Python will be used to run commands or interactive sessions for the remainder of the session.
+
+For these reasons, Conda environments are especially useful to install and manage multiple projects (and/or multiple *versions* of the same project) on the same computer with minimal effort,
+as they provide a way to seamlessly switch between different projects without conflicts.
+
+Using Conda environments is not mandatory to be able to install and use ProteusLib; however, it is strongly recommended.
+
+To use Conda environments, the ``conda`` package manager is required. Refer to the `Conda installation guide <https://conda.io/projects/conda/en/latest/user-guide/install/index.html>`_ for detailed steps on how to install Conda for your operating system.
+
+General installation
+--------------------
+
+If you are going to use ProteusLib's functionality, but *do not* plan to contribute to ProteusLib's codebase, choose this option.
+
+#. Create a Conda environment (in this example, named ``proteuslib``) where ProteusLib and its runtime dependencies will be installed:
+
+	.. code-block:: shell
+
+		conda create --name proteuslib --yes python=3.8
+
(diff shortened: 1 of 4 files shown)


lbianchi-lbl:
A reminder: at any point, you can click on the link provided by the ReadTheDocs check to view a built version of the docs including the changes being proposed in this PR: https://proteuslib--144.org.readthedocs.build/en/144/

docs/getting_started/install.rst
```diff
@@ -0,0 +1,101 @@
...
+
+	.. code-block:: shell
+
+		git clone https://github.com/nawi-hub/proteuslib && cd proteuslib
+
+#. Install ProteusLib and the development dependencies using ``pip`` and the ``requirements-dev.txt`` file:
+
+	.. code-block:: shell
+
+		pip install -r requirements-dev.txt
```
bknueven, line 87: Shouldn't we also tell developers to use the `-e` flag?

bknueven (approved):

docs/getting_started/install.rst
```diff
@@ -0,0 +1,101 @@
...
+
+Introduction
+------------
+
+.. _about-conda:
+
+Using Conda environments
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Conda environments are a way to create and manage multiple sets of packages and/or Python versions on the same system without incurring into conflicts.
```
adam-a-a, line 12:
```suggestion
Conda environments are a way to create and manage multiple sets of packages and/or Python versions on the same system without incurring conflicts.
```

docs/getting_started/install.rst
lbianchi-lbl: That's a good point. The `-e/--editable` flag is, by design, built into `requirements-dev.txt`, with the underlying idea that `pip install -r requirements-dev.txt` is the single, consistent command to run to set up the developer environment.

It could be worth to note this here. Perhaps using a `.. note::` directive pointing to the (currently empty) developer how-to guide?

docs/getting_started/install.rst
lbianchi-lbl: Good catch, thanks!

adam-a-a (approved):
Thanks for the thorough explanation! I think this draft does the job nicely.

docs/how_to_guides/how_to_contribute_development.rst
```diff
@@ -0,0 +1,4 @@
+.. _developer-guide:
+
+How to contribute to ProteusLib's development
```
adam-a-a, line 3: By the way, should we add an issue for this how-to to keep it in mind?

docs/getting_started/install.rst
bknueven: I had forgotten (and didn't check) that it was built in to `requirements-dev.txt`. I'm happy with this as-it-is then.

bknueven (dismissed):

lbianchi-lbl:
@adam-a-a @bknueven we've just realized with @ksbeattie that for the "General installation" it's not possible to use `pytest` since that's a developer dependency (and it's not clear whether the unit tests themselves would be installed when doing `pip install` not in editable mode). So we changed the last step in the "General installation" to something that doesn't require pytest.

Let me know if what we have now makes sense, or if there's a better way (keeping in mind that this is just as a "smoke test", i.e. to have some way of detecting early if something went wrong, rather than a comprehensive check).

ksbeattie (approved):
LGTM

adam-a-a (approved):

PR #146, opened 2021-06-30
updated parameter_sweep tool to allow for random sampling

jefalon:
## Fixes/Addresses:

None

## Summary/Motivation:

This PR add the ability to perform random sampling within the parameter sweep tool using uniform and/or normal distributions.

- Adds Sample base class to keep the parameter sampling standardized
- Adds LinearSample, UniformSample, NormalSample classes, which are used to perform the respective sampling

### Legal Acknowledgement


proteuslib/tools/parameter_sweep.py
```diff
@@ -36,48 +90,76 @@ def _init_mpi(mpi_comm=None):
...
-        # Fix the single value to values[k]
         if param.is_variable_type():
+            # Fix the single value to values[k]
             param.fix(values[k])
 
         elif param.is_parameter_type():
             # Fix the single value to values[k]
             param.set_value(values[k])
 
+        elif param.is_indexed():
```
bknueven, line 158: Is there a use-case for this I'm missing? I took out explicit handling of indexed variables or parameters because it seemed like handing the parameter sweep tool an indexed pyomo object was likely to be an error. And if not, we need to handle this case in the `Sample` validation logic above.

proteuslib/tools/parameter_sweep.py
```diff
@@ -129,9 +211,29 @@ def _default_optimize(model, options=None, tee=False):
...
+        # Convert to using Sample class
+        if isinstance(sweep_params[key],(list,tuple)):
+            sweep_params[key] = LinearSample(*sweep_params[key])
+
+        # Get the type of sampling
+        if i == 0:
+            sampling_type = getattr(sweep_params[key],"type",None)
+        else:
+            if sampling_type != getattr(sweep_params[key],"type",None):
+                raise ValueError("Cannot mix sampling types")
```
bknueven, line 228: Here's where we could just check `isinstance(sweep_param[key], RandomSample)` or such if we used stronger typing.

proteuslib/tools/parameter_sweep.py
```diff
@@ -278,5 +397,12 @@ def parameter_sweep(model, sweep_params, outputs, results_file, optimize_functio
 
         # Save the global data
         np.savetxt(results_file, save_data, header=data_header, delimiter=', ', fmt='%.6e')
+    else:
+        save_data = None
+
+    # Broadcast the results to all processors
+    save_data = comm.bcast(save_data, root=0)
```
bknueven, line 404: Similar to before, we should use `Bcast`; the non-zero ranks should know the size of `save_data` in advance.

bknueven (commented):
Looks like a good start; I left lots of nits in the comments.

eyoung55:
Still to do:
- [ ] Set consistent random seed
- [ ] Verify Gather preserves correct ordering
- [ ] Set type of buffer correctly
- [ ] Verify `num_samples` can be set correctly for all cases, including parsing tuple

proteuslib/tools/parameter_sweep.py
```diff
@@ -16,16 +16,70 @@
...
+
+# ================================================================
+
+class LinearSample(Sample):
+
+    def sample(self, num_samples): 
+        return np.linspace(self.lower_limit, self.upper_limit, self.num_samples)
+
+    def setup(self, lower_limit, upper_limit, num_samples):
+        self.type = "linear"
```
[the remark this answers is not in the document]
jefalon: Agreed, this will make the typing more explicit. I've updated the classes

proteuslib/tools/parameter_sweep.py
```diff
@@ -16,16 +16,70 @@
...
 
+from abc import abstractmethod, ABC 
 from idaes.core.util import get_solver
 
 # ================================================================
 
+class Sample(ABC): 
+
+    def __init__(self, pyomo_object, *args, **kwargs): 
+        assert pyomo_object.is_parameter_type() or pyomo_object.is_variable_type() or pyomo_object.is_indexed()
```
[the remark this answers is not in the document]
jefalon: updated this error message

proteuslib/tools/parameter_sweep.py
```diff
@@ -36,48 +90,76 @@ def _init_mpi(mpi_comm=None):
...
+            p = v.sample(num_samples)
+            param_values.append(p)
+
+        num_var_params = len(param_values)
+
+        if sampling_type == "linear":
+            # Form an array with every possible combination of parameter values
+            global_combo_array = np.array(np.meshgrid(*param_values,indexing="ij"))
+            global_combo_array = global_combo_array.T.reshape(-1, num_var_params,order="F")
+        else:
```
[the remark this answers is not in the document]
jefalon: yeah good call, done.

proteuslib/tools/parameter_sweep.py
```diff
@@ -36,48 +90,76 @@ def _init_mpi(mpi_comm=None):
...
+    return local_combo_array
 
 # ================================================================
 
 def _update_model_values(m, param_dict, values):
 
     for k, item in enumerate(param_dict.values()):
 
-        param = item[0]
+        if isinstance(item,(tuple,list)):
```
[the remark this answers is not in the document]
jefalon: good point, this was just legacy code. I removed it

proteuslib/tools/parameter_sweep.py
jefalon: Yeah I had to add the indexed logic to work with the 0D RO model. For some reason RO.A_comp is an indexed Var() and wasn't being updated. How do you want to go about treating this?

proteuslib/tools/parameter_sweep.py
jefalon: I updated this logic to take advantage of the new FixedSample and RandomSample classes

proteuslib/tools/parameter_sweep.py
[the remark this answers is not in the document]
jefalon: also, this is now handled in the _process_sweep_params() function which happen before this logic making this error message redundant. Should we keep it?

proteuslib/tools/parameter_sweep.py
bknueven: In this case, the user would unroll the index themselves. E.g., RO.A_comp[0] would work, I believe.

proteuslib/tools/parameter_sweep.py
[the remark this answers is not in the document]
bknueven: Yes, it doesn't hurt and makes the "case" analysis more explicit in this function. E.g., if we add another sampling method this function wouldn't silently accept it.

proteuslib/tools/parameter_sweep.py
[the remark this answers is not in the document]
bknueven: I think the answer for the parameter_sweep tool and this other code is to break out the logic that does pre-processing to check for tuples/lists into its own function.

proteuslib/tools/parameter_sweep.py
adam-a-a: FYI - A_comp is indexed by time and solvent component. I think this was originally done with dynamic modeling in mind (A_comp could change with time due to fouling although there are other ways we could account for membrane fouling over time). 

As for having the second index for solvent, this doesn't make much sense for RO since it supports only one solvent in the first place; I _think_ this was done for consistency with IDAES modeling and how things were normally indexed. On the other hand, for B_comp (salt permeability factor of membrane) this has indices for time and solute, but the RO model could technically take on more than one solute.

bknueven: This is exactly why I'm nervous about setting indexed things -- it could be the source of silent bugs (setting indices you didn't mean to set), which might be very hard to detect. Further, given the structure of IDAES models the only "proper" use case I could imagine is fixing all the time indices of an IndexedVar or IndexedParam to the same value. Fixing all the non-time indices, like for `flow_mass_phase_comp`, would almost surely be an error.

proteuslib/tools/parameter_sweep.py
```diff
@@ -129,9 +246,38 @@ def _default_optimize(model, options=None, tee=False):
...
+    # Check the list of parameters to make sure they are valid
+    for i, key in enumerate(sweep_params.keys()):
+
+        # Convert to using Sample class
+        if isinstance(sweep_params[key], (list, tuple)):
+            sweep_params[key] = LinearSample(*sweep_params[key])
+
+        # Get the type of sampling
+        if isinstance(sweep_params[key],FixedSample):
+            current_sampling_type = "fixed"
```
bknueven, line 262: We could set `current_sample_type` to `FixedSample` as opposed to a string.

proteuslib/tools/parameter_sweep.py
bknueven: I could be convinced that this could be a useful "typing aid" for the case of `A_comp`, e.g., and indexed var/param with exactly one index (in most use cases now). That said, we could check for that when we create the `Sample` object and replace the indexed component with the appropriate single var/param.

jefalon: I tried to take the indexed A_comp and get our the actual Var() I wanted to change, but I couldn't get it to work. Perhaps I was doing it wrong

proteuslib/tools/parameter_sweep.py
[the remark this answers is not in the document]
jefalon: removed

proteuslib/tools/parameter_sweep.py
jefalon: added the check for single valued indexed sets

proteuslib/tools/parameter_sweep.py
jefalon: added an Enum class

proteuslib/tools/parameter_sweep.py
```diff
@@ -15,18 +15,96 @@
...
+class SamplingType(IntEnum):
+    FIXED = auto()
+    RANDOM = auto()
+
+# ================================================================
+
+class _Sample(ABC): 
+
+    def __init__(self, pyomo_object, *args, **kwargs):
+        if pyomo_object is not None:
```
bknueven, line 35: Is there a reason to allow the pyomo_object to be None? If is it None, the current code will crash somewhere...

eyoung55: I think I added this to simplify some of the unit tests, e.g., create a sample like `UniformSample(None, range_A[0], range_A[1])` to test only that the random pieces are working.  Admittedly, just passing in a dummy `pyomo_object` isn't much harder and would remove this extra check.

proteuslib/tools/parameter_sweep.py
```diff
@@ -15,18 +15,96 @@
...
+            # Check for indexed with single value
+            if pyomo_object.is_indexed() and len(pyomo_object.values()) == 1:
+                print(pyomo_object.values())
+                exit()
```
bknueven, line 39:
```suggestion
            if pyomo_object.is_indexed() and len(pyomo_object) == 1:
```

jefalon: Whoops, looks like I left a bit of debugging in there. Thanks for cleaning it up!

bknueven: It's tedious, but we should add a simple case for the unit tests. Note something as simple as 
```
m=pyo.ConcreteModel()
m.x = pyo.Var(['a'])
```
would trigger this logic when `m.x` is passed in as the `pyomo_object`

proteuslib/tools/parameter_sweep.py
```diff
@@ -15,18 +15,92 @@
...
 import itertools
+import warnings
 
+from enum import IntEnum, auto
+from abc import abstractmethod, ABC 
 from idaes.core.util import get_solver
 
 # ================================================================
 
+class SamplingType(IntEnum):
```
adam-a-a, line 26: Just out of curiosity, can we use `Enum` here or is `IntEnum` necessary here? If `IntEnum` _is_ necessary, is it because `auto()` is being used?

bknueven (approved):
The tests have some redundant code, but that is a minor issue. Additional documentation for these changes is part of #161.

proteuslib/tools/parameter_sweep.py
bknueven: This is a good catch -- I looked it up and `IntEnum` will silently cast its members to `int` for comparison purposes. E.g.
```python
class FooType(Enum):
    BAR = 1

class BarType(IntEnum):
    FOO = 1

# False
FooType.BAR == 1

# True
BarType.FOO == 1
```

The `auto()` function is just for convenience.

bknueven (approved):
@jefalon I'm happy with this as-is. See if you're okay with the changes Ethan and I have made.

jefalon:
It looks great to me! Thanks everyone for helping get this finished. As soon as we get that second approval, I think we are ready to merge.

eyoung55 (approved):
Looks good to me.

jefalon:
@bknueven I don't think I have the right to merge. Do you?

PR #159, opened 2021-07-09
eliminate apidoc warnings 

adam-a-a:
## Summary/Motivation:

Fix warnings that occur when developing documentation and running `make html`

- add :noindex: directive to relevant files
- add modules.rst file to toctree
- remove apidoc folder from .gitignore

### Legal Acknowledgement


bknueven (changes requested):
The directory `docs/apidoc` should not be part of this PR

adam-a-a:
> The directory `docs/apidoc` should not be part of this PR

I wasn't planning to have this as part of the PR, but it was the only way I found to avoid warnings locally while also passing all checks on GH. I also added `docs/apidoc` to .gitignore (but wasn't entirely sure if it'd still be ignored after this)

PR #161, opened 2021-07-09
Updating parameter sweep docs

bknueven:
## Changes proposed in this PR
- [x] Break out optimize in RO_with_energy_recovery flowsheet & update parameter sweep how-to
- ~~Update RO_with_energy_recovery flowsheet & parameter sweep how-to to correctly set concentration~~ (Will wait for updates to NaCl Property Package)
- [x] Updates for parameter sweep docs to keep them up-to-date after #146.

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

Issue #133, opened 2021-06-24
Update flowsheet for parameter sweep docs


TimBartholomew:
The parameter sweep tool uses the RO with energy recovery flowsheet as the example. However, the functions don't necessarily align with the use. We should break up the optimize into an optimize_set_up and a solve call.

Also the docs should be updated to make it clear that users don't necessarily have to have the same functions as the flowsheet, instead they just need a fully initialized model, with bounds around the decision variables, and a function for the solve.

TimBartholomew:
Also the parameter sweep has a wrong label, it states flow_mass_phase_comp is concentration. We should make it so that the mass fraction is varied instead, but this will also require unfixing both flow_mass_phase_comps for NaCl and H2O and fix the volumetric flowrate.

PR #166, opened 2021-07-15
Example of Monte Carlo sampling using RO ERD flowsheet

eyoung55:
## Fixes/Addresses:

Adds a script demonstrating how to perform Monte Carlo sampling of an existing flowsheet using RO with energy recovery as an example.

## Summary/Motivation:

It's useful to see a working example of how to use the random sample classes and how to interface with an actual flowsheet.

- Add an example Monte Carlo analysis of an existing flowsheet.

### Legal Acknowledgement


adam-a-a:
This looks good to me and could probably be transformed directly into a how-to for our documentation (with some minor additions). What do you think, @eyoung55 @bknueven @jefalon ?

proteuslib/flowsheets/RO_with_energy_recovery/monte_carlo_sampling_RO_ERD.py
```diff
@@ -0,0 +1,35 @@
...
+initialize_system(m, solver=solver)
+
+# Simulate once outside the parameter sweep to ensure everything is appropriately initialized 
+solve(m, solver=solver)
+
+# Define the sampling type and ranges for three different variables
+sweep_params = {}
+sweep_params['A_comp'] = NormalSample(m.fs.RO.A_comp, 4.0e-12, 0.5e-12)
+sweep_params['B_comp'] = NormalSample(m.fs.RO.B_comp, 3.5e-8, 0.5e-8)
+sweep_params['Porosity'] = UniformSample(m.fs.RO.spacer_porosity, 0.95, 0.99)
```
adam-a-a, line 25:
```suggestion
sweep_params['Spacer_porosity'] = UniformSample(m.fs.RO.spacer_porosity, 0.95, 0.99)
```
I know it's just an alias but would want users with desal expertise to have no confusion about which porosity we are talking about (even though one could easily observe `spacer_porosity` later in the line).

eyoung55: Makes sense to me! Fixed.

bknueven:
@adam-a-a I think I agree -- perhaps this file broken out could replace what's there, which is already somewhat broken anyways (see #133 and #161).

adam-a-a:
@bknueven That could work. Just to clarify, are you referring to replacing the how-to for parameter sweep with a slightly beefed-up version of this?  

I think it wouldn't hurt to have two separate how-to docs: one that introduces the use of parameter sweep (w/LinearSample), and a second follow-on how-to that shows how to perform Monte Carlo sampling via this example. On the one hand, it could seem a little repetitive, but on the other hand, I think it would really drive the teaching points home for the user.

bknueven:
@adam-a-a Given that `LinearSample` and the random sampling types cannot be combined, I think you are correct that there should be two separate how-to's:
1. How to systematically sample flowsheet parameters
2. How to Monte Carlo sample flowsheet parameters.

adam-a-a:
@bknueven That could work. Just to clarify, are you referring to replacing the how-to for parameter sweep with a slightly beefed-up version of this?  

I think it wouldn't hurt to have two separate how-to docs: one that introduces the use of parameter sweep, and a second follow-on how-to that shows how to perform Monte Carlo sampling 

> @adam-a-a Given that `LinearSample` and the random sampling types cannot be combined, I think you are correct that there should be two separate how-to's:
> 
> 1. How to systematically sample flowsheet parameters
> 2. How to Monte Carlo sample flowsheet parameters.

@bknueven Sounds good to me.

bknueven:
The comments about documentation aside, I think there is value in having this example. @eyoung55 could you add a test for this in `proteuslib/flowsheets/RO_with_energy_recovery/tests` just so this code doesn't accidentally go stale?

eyoung55:
Just getting up to speed on Friday's conversation.  @bknueven, yes I can add a test with this code.  Is the idea that this script should effectively live as a test only, with some additional explanation being added to the existing how-to guide on random sampling?  Or is there still value in having this as a standalone script `proteuslib/flowsheets/RO_with_energy_recovery/monte_carlo_sampling_RO_ERD.py `?  I'm thinking the former, but let me know if this idea has changed.

bknueven:
I think there is some value in having this whole Python modules as example. This one is so small the entire code block could just be displayed verbatim in a how-to guide, and we could also give an example of running it in parallel.

adam-a-a (approved):
LGTM. As the purpose of this PR was to show an example only, my previous points on directly building how-to docs from this can be resolved in a subsequent PR. 
Perhaps we should create an issue for creating those how-to docs in case one doesn't exist already.

bknueven (approved):

Issue #106, opened 2021-06-11
Documentation on Apparent vs True Chemical Species


aladshaw3:
Need to create "How-to-do" and "Explanation" documentation for apparent species and true species conventions that we will use in the chemistry modules. Some of this may eventually live in IDAES, but we can go ahead and start with it here. Technical references for this should just be a link to IDAES documentation for Generic Properties. 

- [x] "How-to-do" for specifying a species as Apparent or True (#169)
- [x] "Explanation" for what the difference is between Apparent and True species and why it is important (#169)
- [x] "Technical Reference" link to appropriate IDAES documentation (where appropriate)
- [x] "How-to-do" for creating a basic chemistry module with IDAES objects

Issue #140, opened 2021-06-28
Need how-to guide that illustrates utilization of EDB in model development


adam-a-a:
Quoting the guide on [how to setup simple chemistry](https://watertap.readthedocs.io/en/latest/how_to_guides/how_to_setup_simple_chemistry.html):
> Much of the difficulties and complications with setting up a proper thermo-properties configuration dictionary can be handled by the Electrolyte Database system in ProteusLib (Documentation pending)

A how-to guide should be created not only to elaborate on this quote, but also to show the reader how to implement the EDB for developing models.

dangunter:
@adam-a-a the link above is broken now

adam-a-a:
@dangunter I updated the link (changed "proteuslib" to "watertap" in the url)

PR #169, opened 2021-07-16
Initial documentation on Apparent and True species

aladshaw3:
## Fixes/Addresses:

#106 

- Adds documentation for apparent species 

### Legal Acknowledgement

### DIFF
--- a/docs/how_to_guides/how_to_use_apparent_and_true_chemical_species.rst
+++ b/docs/how_to_guides/how_to_use_apparent_and_true_chemical_species.rst
@@ -1,11 +1,14 @@
 How to use apparent and true chemical species
 =============================================
 
-:doc:`how_to_setup_simple_chemistry`
-
 In ProteusLib, most all chemical processes simulated will be considered "true"
 species, i.e., species that actually exist in an aqueous solution. However, their
 may be times when you want to have your system report the "apparent" species
 between your inlet and/or outlet ports of a flowsheet. To do this, you need
 to understand how to define a species in a given configuration file as "apparent",
 which is a special component type in the **thermo-properties** configuration dictionary.
+
+For more information on creating a **thermo-properties** configuration dictionary,
+see 'How to setup simple chemistry'.
+
+:ref:`rst_tutorial`
--- a/docs/how_to_guides/how_to_use_apparent_and_true_chemical_species.rst
+++ b/docs/how_to_guides/how_to_use_apparent_and_true_chemical_species.rst
@@ -37,91 +37,59 @@ an ``"Apparent"`` species.
 
 .. testcode::
 
-# Importing the object for units from pyomo
-from pyomo.environ import units as pyunits
+  # Importing the object for units from pyomo
+  from pyomo.environ import units as pyunits
 
-# Imports from idaes core
-from idaes.core import AqueousPhase
-from idaes.core.components import Solvent, Cation, Anion
+  # Imports from idaes core
+  from idaes.core import AqueousPhase
+  from idaes.core.components import Solvent, Cation, Anion
 
-# Imports from idaes generic models
-import idaes.generic_models.properties.core.pure.Perrys as Perrys
-from idaes.generic_models.properties.core.pure.ConstantProperties import Constant
-from idaes.generic_models.properties.core.state_definitions import FTPx
-from idaes.generic_models.properties.core.eos.ideal import Ideal
+  # Imports from idaes generic models
+  import idaes.generic_models.properties.core.pure.Perrys as Perrys
+  from idaes.generic_models.properties.core.pure.ConstantProperties import Constant
+  from idaes.generic_models.properties.core.state_definitions import FTPx
+  from idaes.generic_models.properties.core.eos.ideal import Ideal
 
-# Configuration dictionary
-thermo_config = {
-    "components": {
-        'H2O': {"type": Solvent,
-              # Define the methods used to calculate the following properties
-              "dens_mol_liq_comp": Perrys,
-              "enth_mol_liq_comp": Perrys,
-              "cp_mol_liq_comp": Perrys,
-              "entr_mol_liq_comp": Perrys,
-              # Parameter data is always associated with the methods defined above
-              "parameter_data": {
-                    "mw": (18.0153, pyunits.g/pyunits.mol),
-                    # Parameters here come from Perry's Handbook:  p. 2-98
-                    "dens_mol_liq_comp_coeff": {
-                        '1': (5.459, pyunits.kmol*pyunits.m**-3),
-                        '2': (0.30542, pyunits.dimensionless),
-                        '3': (647.13, pyunits.K),
-                        '4': (0.081, pyunits.dimensionless)},
-                    "enth_mol_form_liq_comp_ref": (-285.830, pyunits.kJ/pyunits.mol),
-                    "enth_mol_form_vap_comp_ref": (0, pyunits.kJ/pyunits.mol),
-                    # Parameters here come Perry's Handbook:  p. 2-174
-                    "cp_mol_liq_comp_coeff": {
-                        '1': (2.7637E5, pyunits.J/pyunits.kmol/pyunits.K),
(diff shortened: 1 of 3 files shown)


aladshaw3:
@TimBartholomew  Part of what I put in issue #106 was for "Explanation" on the difference between "Apparent" and "True" chemical species. Will that live in a different folder of the docs?

docs/how_to_guides/how_to_use_apparent_and_true_chemical_species.rst
```diff
@@ -0,0 +1,213 @@
...
+                # Define the methods used to calculate the following properties
+                "dens_mol_liq_comp": Perrys,
+                "enth_mol_liq_comp": Perrys,
+                "cp_mol_liq_comp": Perrys,
+                "entr_mol_liq_comp": Perrys,
+                # Parameter data is always associated with the methods defined above
+                "parameter_data": {
+                      "mw": (18.0153, pyunits.g/pyunits.mol),
+                      # Parameters here come from Perry's Handbook:  p. 2-98
+                      "dens_mol_liq_comp_coeff": {
```
adam-a-a, line 67: Now that I think about it, is there a way for the user to avoid having to specify these parameter data from Perry's Handbook in this example? Is that where the EDB will come into play? I was just thinking that if a user didn't have these parameter values readily available, it could be painful... but I am assuming that the database (or some other means) would be used to avoid manually entering such values (at least for common configurations).

docs/how_to_guides/how_to_use_apparent_and_true_chemical_species.rst
```diff
@@ -0,0 +1,213 @@
+How to use apparent and true chemical species
+=============================================
+
+In ProteusLib, most all chemical processes simulated will be considered "true"
+species, i.e., species that actually exist in an aqueous solution. However, their
```
adam-a-a, line 5:
```suggestion
species, i.e., species that actually exist in an aqueous solution (e.g., Na+ and Cl- for NaCl). However, their
```
Already make sense and is clear, but I think it would be nice to spell this out as much as possible.

docs/how_to_guides/how_to_use_apparent_and_true_chemical_species.rst
adam-a-a: On the other hand, this is useful because it shows the user how to manually enter such parameters.

aladshaw3: @adam-a-a  Yes, eventually much of this will be done by EDB. EDB is in parallel development with the rest of the chem team, so for the time being I am doing this stuff manually. Future documentation for how to do this with EDB is going to come (#140).

andrewlee94: We need to be careful with this however - we need to make sure we have the right to reproduce data like this. Also, the IDAES approach has generally be to provide a way for users to input data, but to leave finding the data to the user.

docs/how_to_guides/how_to_use_apparent_and_true_chemical_species.rst
```diff
@@ -0,0 +1,213 @@
...
+
+For more information on creating a **thermo-properties** configuration dictionary,
+see 'How to setup simple chemistry'.
+
+
+What you need to update in the thermo-properties configuration dictionary
+-------------------------------------------------------------------------
+
+1. Add the ``"Apparent"`` and ``"StateIndex"`` objects under your import statements
+2. Specify specific species as ``"Apparent"`` within its component definition
```
adam-a-a, line 19:
```suggestion
2. Specify certain species as ``"Apparent"`` within its component definition
```

docs/how_to_guides/how_to_use_apparent_and_true_chemical_species.rst
```diff
@@ -0,0 +1,213 @@
...
+3. Give a list of "true" species it will dissociate into
+4. Define a "state_components" argument in the config and specify ``"StateIndex.true"``
+
+
+Example Configuration Dictionary
+--------------------------------
+
+This is best understood by going through an example. In this case, we will consider
+building off of the 'How to setup simple chemistry' guide and add ``"NaCl"`` as
```
adam-a-a, line 28: could add a hyperlink to this how-to (even though it's the doc right before this one in the toc tree)

docs/how_to_guides/how_to_use_apparent_and_true_chemical_species.rst
```diff
@@ -0,0 +1,213 @@
...
+                         "mass": pyunits.kg,
+                         "amount": pyunits.mol,
+                         "temperature": pyunits.K},
+      }
+      # End thermo_config definition
+
+
+.. note::
+
+    When you define a species as ``"Apparent"``, and specify ``"state_components": StateIndex.true``,
```
adam-a-a, line 209:
```suggestion
    When you define a species as ``"Apparent"`` and specify ``"state_components": StateIndex.true``,
```

adam-a-a (changes requested):
Aside from a couple of my piddly suggested changes, this looks good to me. I suppose the only other issue to consider is whether we should make changes to the parameter data before merging.

docs/how_to_guides/how_to_use_apparent_and_true_chemical_species.rst
aladshaw3: I wanted to do that, but trouble figuring it out. Is there a resource/guide for doing this?

docs/how_to_guides/how_to_use_apparent_and_true_chemical_species.rst
aladshaw3: Sounds to me that, at least for the purpose of this example, we should just use any open source parameters or methods. The exact values of parameters are not important for any of these docs.

docs/how_to_guides/how_to_use_apparent_and_true_chemical_species.rst
adam-a-a: I also had trouble figuring this out. It ends up that doing the following would work:
```suggestion
building off of the :ref:`How to setup simple chemistry<how_to_setup_simple_chemistry>` guide and add ``"NaCl"`` as
```
So the syntax is :ref: `Label title<file_name>` in this case. You can also find more helpful info on Sphinx links [here](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#hyperlinks) and [here](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/references.html).

TimBartholomew (approved):
LGTM

docs/how_to_guides/how_to_use_apparent_and_true_chemical_species.rst
aladshaw3: @adam-a-a It looks like after I made this change the building of the docs failed.

adam-a-a: Strange- I tested this locally before suggesting and didn't get any error; maybe I forgot to run `make clean` to start fresh before running `make html` again and missed the warning. Now when I try it locally, I also get the same error.

In any case, adding the following line to the top of the "how to setup simple chemistry" doc should (hopefully) resolve the issue:
`.. _how_to_setup_simple_chemistry:
`

docs/how_to_guides/how_to_use_apparent_and_true_chemical_species.rst
```diff
@@ -0,0 +1,213 @@
+How to use apparent and true chemical species
+=============================================
+
+In ProteusLib, most all chemical processes simulated will be considered "true"
+species, i.e., species that actually exist in an aqueous solution (e.g., Na+ and Cl- for NaCl). However, their
```
adam-a-a, line 5:
```suggestion
species, i.e., species that actually exist in an aqueous solution (e.g., Na+ and Cl- for NaCl). However, there
```
Sorry to drag this out but just noticed this as I was giving one more pass.

adam-a-a (approved):

TimBartholomew (approved):

Issue #93, opened 2021-05-27
Identify properties that aren't built in RO model to avoid misreporting


adam-a-a:
While testing one of the LSRRO simulation files, I noticed that the osmotic pressure of the final brine discharge was being displayed, and the value was erroneously low. After further investigation, I realized that the osmotic pressure of the _bulk_ solution at the RO outlet was not being called, and properties are set up to be built on demand. Since we modified the RO model to include state blocks for the inlet/outlet of the membrane interface (#11), the bulk osmotic pressure is no longer called in the RO model; instead, the osmotic pressure at the membrane interface is called. Thus, if the user wanted to display the bulk osmotic pressure at the RO outlet, he or she would get the initialized value from that property model. For example, the default value for osmotic pressure in the NaCl property model is 10 bar, and this is the value that was displayed for the bulk osmotic pressure of a high-salinity brine.

This is the only example I stumbled upon, but we should probe further to identify any properties that are not called in the RO model in a given state block and ensure that none of these would be of interest to a user. Otherwise, he or she will try to report said value expecting to get a calculated value, and this will either result in (i) misreporting, (ii) cascading error if any post-processing/calculations are conducted, or (iii) the user thinking that model calculations are incorrect when the actual problem is that the model just didn't perform the calculation.

Furthermore, we should provide a remedy for this, which may entail modifications in the RO model.

TimBartholomew:
Generally, this is not a RO model issue as this type of user error is true of all IDAES models. We can provide better documentation (actually we currently have none) that lists for the user all the variables that are part of the model.

Whether or not that the bulk feed and retentate osmotic pressure is a property is up for discussion, I'm leaning away from that, but really it should come down to what we think is best for the user in the report method of the RO model.

adam-a-a:
After the addition of the `over_pressure_ratio` variable in #84, the bulk osmotic pressure at the RO outlet is no longer unused. Thus, the value would be reported correctly if called after a solve. For the time being, it seems like this issue is longer a problem for the RO model.

However, I should note that the `unused_variables_set` and `number_unused_variables` functions don't seem to descend into the model, so in our tests, the true number of unused variables might be missed unless we call each block--or at least that is my interpretation; this might merit further discussion but is currently low priority.

andrewlee94:
@adam-a-a I didn't bother checking to confirm, but you should be able to provide the argument `descend_into=True` to the model statistics methods to tell them to descend into child blocks.

[...]

Issue #56, opened 2021-05-06
Add reports to unit models


TimBartholomew:
Currently the unit models do not include the methods to support reports. These should be added.

adam-a-a:
**Note:** `process_base.py` contains the `report()` method, which calls both `_get_performance_contents()` and `_get_stream_table_contents()` methods. There is also the `serialize_contents()` method which returns both performance contents and stream table but can possibly result in a ConfigurationError if inlets/outlets of unit model are atypical.

- [ ] Add `_get_performance_contents()` method to unit models in the following format, based on IDAES unit model documentation: 
````
def _get_performance_contents(self, time_point=0):
    var_dict = {"display name": self.var[time_point]}
    expr_dict = {"display name": self.expr[time_point]}
    param_dict = {"display name": self.param[time_point]}

    return {"vars": var_dict, "exprs": expr_dict, "params": param_dict}
````
- [ ] Add `_get_stream_table_contents()` method
- [ ] Include `report()` when updating RO (or future) tutorial

adam-a-a:
Running TO-DO: 

- [ ] Add report() call to RO tutorial
- [ ] Add reporting to NF model
- [ ] Add reporting to ERD
- [ ] Add reporting to Pump

adam-a-a:
Reopening this issue since on item on the task list has not been merged yet.

PR #170, opened 2021-07-16
Add reporting to 0DRO

adam-a-a:
## Fixes/Addresses:

This PR partially addresses  #56 and #93. 

## Summary/Motivation:

This PR adds vars to be reported by calling report() method after a solve:
e.g.,: `m.fs.unit.report()`

The user will see a report of 0DRO variables and a stream table showing state variables. The list of variables currently shown is not exhaustive, but this PR is meant to get the ball rolling on some form of reporting via the RO model. 

**Note:** check out the TODO note and let me know what you think. Furthermore, feel free to propose other key variables to add to the output.

- adds variables to display in `_get_performance_contents()` method
- adds ports to display in `_get_stream_table_contents()` method

### Legal Acknowledgement

### DIFF
--- a/proteuslib/unit_models/reverse_osmosis_0D.py
+++ b/proteuslib/unit_models/reverse_osmosis_0D.py
@@ -1055,12 +1055,48 @@ class ReverseOsmosisData(UnitModelBlockData):
         if hasattr(self, "width"):
             var_dict["Membrane Width"] = self.width
         if hasattr(self, "deltaP"):
-            var_dict["Pressure Change"] = self.deltaP[time_point]
+            var_dict["Pressure Drop"] = self.deltaP[time_point]
         if hasattr(self, "N_Re_io"):
-            var_dict["Inlet Reynolds Number"] = self.N_Re_io[time_point, 'in']
-        if hasattr(self, "N_Re_io"):
-            var_dict["Outlet Reynolds Number"] = self.N_Re_io[time_point, 'out']
-        # TODO: add more vars
+            var_dict["Reynolds Number @Inlet"] = self.N_Re_io[time_point, 'in']
+            var_dict["Reynolds Number @Outlet"] = self.N_Re_io[time_point, 'out']
+        if hasattr(self, "velocity_io"):
+            var_dict["Velocity @Inlet"] = self.velocity_io[time_point, 'in']
+            var_dict["Velocity @Outlet"] = self.velocity_io[time_point, 'out']
+        var_dict['Concentration @Inlet,Membrane-Interface '] = (
+                    self.feed_side.properties_interface_in[time_point].conc_mass_phase_comp['Liq', 'NaCl'])
+        var_dict['Concentration @Outlet,Membrane-Interface '] = (
+            self.feed_side.properties_interface_out[time_point].conc_mass_phase_comp['Liq','NaCl'])
+        var_dict['Concentration @Inlet,Bulk'] = (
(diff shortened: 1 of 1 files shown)


proteuslib/unit_models/reverse_osmosis_0D.py
```diff
@@ -1045,13 +1046,69 @@ def initialize(blk, initialize_guess=None, state_args=None, outlvl=idaeslog.NOTS
...
+        var_dict["Volumetric Recovery Rate"] = self.recovery_vol_phase[time_point, 'Liq']
+        var_dict["Solvent Mass Recovery Rate"] = self.recovery_mass_phase_comp[time_point, 'Liq', 'H2O']
+        var_dict["Membrane Area"] = self.area
+        if hasattr(self, "length"):
+            var_dict["Membrane Length"] = self.length
+        if hasattr(self, "width"):
+            var_dict["Membrane Width"] = self.width
         if hasattr(self, "deltaP"):
-            var_dict["Pressure Change"] = self.deltaP[time_point]
+            var_dict["Pressure Drop"] = self.deltaP[time_point]
```
bknueven, line 1058: This is very pedantic -- but isn't "Pressure Change" more correct, as this reports a negative quantity. That is, "Pressure Drop" would be the negative of `self.deltaP[time_point]`.

bknueven (approved):
This is better than not having it, and it quite good at that. The TODO should be left in so it can be addressed in other PRs.

proteuslib/unit_models/reverse_osmosis_0D.py
adam-a-a: I agree -- I'll change it back.

proteuslib/unit_models/reverse_osmosis_0D.py
```diff
@@ -1045,13 +1046,69 @@ def initialize(blk, initialize_guess=None, state_args=None, outlvl=idaeslog.NOTS
...
+            var_dict['Volumetric Flowrate @Outlet'] = (
+                self.feed_side.properties_in[time_point].flow_vol_phase['Liq'])
+
+        # TODO: (1) add more vars, (2) would be nice to add units to output, and (3) should be able to report output of
+        #  "NaN" or "Not Reported", mainly for properties that exist but are not necessarily constructed within model
+        #  constraints. One example in this case is the osmotic pressure of the bulk feed, which would certainly be of
+        #  interest to users with a background in desalination (it is currently not reported because it is not directly
+        #  used in any model constraints). Furthermore, the report() method seems to be limited to Pyomo Var objects for
+        #  which the pyomo value() method is applied to. That is, a Pyomo Var object must be used; e.g., providing a
+        #  list as output would yield an error.
```
bknueven, line 1098: It seems like you can provide Expressions as `"exprs" : exprs_dict` and Params as `"params" : params_dict` in the return dictionary below:
https://github.com/IDAES/idaes-pse/blob/b878aea16c073e813779283d8aa997de429fbe75/idaes/core/process_base.py#L226-L323

Expressions might be useful for some of these derived properties here and elsewhere.

bknueven (approved):

proteuslib/unit_models/reverse_osmosis_0D.py
adam-a-a: Nice - thanks for checking this. It also looks like we'd want to modify one of the args in `tabular_writer()` to add another column that displays units; this is not a necessity but I think it would be useful and would provide more transparency for the user's sake.

TimBartholomew (approved):
LGTM

PR #175, opened 2021-07-22
Update idaes-pse to latest tag on nawi-hub fork

lbianchi-lbl:
## Summary/Motivation:

Update prerelease dependency for `idaes-pse` in `setup.py` after fast-forwarding and tagging our `nawi-hub/idaes-pse` fork.

### Legal Acknowledgement

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -38,6 +38,9 @@ jobs:
             os-version: windows-2019
           # - os: macos
           #   os-version: macos-10.15
+          # limit uploading coverage report for a single Python version in the matrix
+          - python-version: '3.8'
+            cov_report: true
     steps:
     - uses: actions/checkout@v2
     - name: Set up Python ${{ matrix.python-version }}
@@ -59,9 +62,15 @@ jobs:
         echo '::group::Output of "idaes get-extensions" command'
         idaes get-extensions --verbose
         echo '::endgroup::'
+    - name: Add coverage report options
+      if: matrix.cov_report
+      run: echo 'PYTEST_ADDOPTS="--cov-report=xml"' >> $GITHUB_ENV
     - name: Test with pytest
       run: |
         pytest
+    - name: Upload coverage report to Codecov
+      if: matrix.cov_report
+      uses: codecov/codecov-action@v2
     - name: Test documentation code
       run: |
         make -C docs doctest -d
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -50,8 +50,10 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
+        echo '::group::Output of "conda install" commands'
+        conda install --quiet --yes pip=21.1 wheel setuptools
+        echo '::endgroup::'
         echo '::group::Output of "pip install" commands'
-        python -m pip install --progress-bar off --upgrade pip wheel setuptools
         pip install --progress-bar off -r requirements-dev.txt
         echo '::endgroup::'
         echo '::group::Display installed packages'
--- a/docs/getting_started/install.rst
+++ b/docs/getting_started/install.rst
@@ -29,7 +29,7 @@ If you are going to use ProteusLib's functionality, but *do not* plan to contrib
 
 	.. code-block:: shell
 
-		conda create --name proteuslib --yes python=3.8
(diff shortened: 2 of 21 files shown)


lbianchi-lbl:
It seems most of the tests are failing with the most recent tag. For a comparison between the two tags: https://github.com/nawi-hub/idaes-pse/compare/1.10.1.proteuslib.2021.07.01...nawi-hub:1.10.1.proteuslib.2021.07.22

aladshaw3:
@andrewlee94 Looks like all test failures are linked to bad scaling of the aqueous chemistry. Most failures are solver or initailizer failures. These are indicative of poor scaling. What was changed in IDAES exactly that now makes our prior scaling work defunct?

aladshaw3:
@andrewlee94 @srikanthallu @TimBartholomew This should not be merged in until the entire chem team can test this new IDAES code with the core chemistry modules. This new IDAES completely breaks everything we have worked on for the past few months and completely changes what our modeling and scaling strategies will be going forward. 

Unfortunately, I will have very limited time to do a deep dive on this. I am already over committed on time for the next two weeks working on proposals and other projects. This needs to wait for the time being.

lbianchi-lbl:
Just to provide a bit more information (unfortunately purely on the software engineering side, but hopefully it can still be helpful): I ran `git bisect` to try to figure out what's the first commit in `idaes-pse` that causes one or more tests in `proteuslib` to fail, and the results seems to be [this one](https://github.com/nawi-hub/idaes-pse/commit/f9da921328a8b4f99507ba4270c453037c93b015).

The complete log is attached here: [git-bisect-2.log](https://github.com/nawi-hub/proteuslib/files/6866152/git-bisect-2.log)

andrewlee94:
Looking through those failures, I see three different causes, two of which are more-or-less expected:

1. The calculation of density for mixtures in IDAES was corrected, which results in small changes in the calculated density. The new results should be more accurate, but do mean the tests need to be updated.
2. I saw at least one test where the difference was in the 6th significant figure of the result when testing to a relative error of 1e-5. However, the `assert` statement was only written with 4-5 significant figures, so  that small change was enough to break the tests. I would suggest always writing `assert` statements to one more significant figure than you are asserting.
3. I saw a couple of tests where the solver failed to converge for some reason, and it might be due to scaling. However, I am not sure what would have caused that behaviour - the only significant change I know of is the density calculation and that should be peripheral to the model.

@lbianchi-lbl I cannot see how that specific PR/commit would have caused any of these problems. None of our models are using those features yet.

adam-a-a:
@aladshaw3 I was able to eliminate a couple of the failures, but there is one that I can't get through yet. Apparently, while trying to solve in `test_solve_cstr_kin`, I find the following error related to taking the log of a negative value:

`Error evaluating constraint fs.unit.control_volume.properties_out[0.0].inherent_equilibrium_constraint[H2CO3_Ka2]: can't evaluate log(-2.21051e-20).`

I should add that I tried including another test to check for poorly scaled variables and attempted to repair those poorly scaled vars. Right now, that code is commented out, but if you were to run the test with that uncommented, you'd see a similar error message, except you'd see it for a different reaction (e.g., `H20_Kw` instead of `H2CO3_Ka2`). 

Therefore, it seems that adjusting variable scaling further might resolve the error, but I am not sure to get there. @andrewlee94 tried working through a solution with me; maybe he or somebody else can make some headway with this.

aladshaw3:
I cannot recreate the same set of errors (test errors on linux) that are being reported on my own machine, so it will be very difficult to fix this.

aladshaw3:
@adam-a-a I was able to recreate some of the issues by creating a manual IDAES env and running the tests in their. Then, I am able to correct some of the issues by changing how the scaling is done (at least on the linux side). I have a few more issues to work out, but you will start to see the linux checks passing soon (I would imagine). 

Additional updates may be needed for Windows. The major problem with all the chemistry is that the problems are all very poorly scaled and the scaling changes from unit to unit, computer to computer. I don't think we will ever be able to do any realistic chemistry reliably unless we massively simplify what reactions we consider, how we make the equilibrium statements, and find a universal way to scale the problems.

ksbeattie (approved):
LGTM

ksbeattie (approved):

aladshaw3 (approved):
LGTM, I opened a PR to deal with the skipped tests. Will get to that asap.

PR #178, opened 2021-07-23
Calculate state method

### LINKED ISSUE #57: Add calculate_state method to property packages
In many cases its useful to fix a property and calculate the associated state variables. This type of calculation is not supported by the stateblock initialization because it fixes the state variables in an attempt to calculate the properties, which if properties are fixed then the degrees of freedom is less than 0.

This new method could be called `calculate_state`, and the arguments would be similar to the initialize method including outlvl, solver, optarg, and a dictionary with the variables that should be temporarily fixed and their values (like the state_args).

TimBartholomew:
## Fixes/Addresses:

Fixes #57 

## Summary/Motivation:
Oftentimes users want to fix properties on a stateblock and while this is possible it frequently causes errors in initialization. This happens because stateblock initialization routines assume the state variables should be fixed before the solve call. This PR creates another method called `calculate_state`, where a stateblock is solved for user provided variables and values (that can be properties). 

- Adds calculate state method to the NaCl and Seawater property models
- Adds testing for the new method

### Legal Acknowledgement


andrewlee94:
@TimBartholomew What is the actual use case for this method, and how do you see it being used? I can see some potential issues with this, and it would help to know exactly what you intend this for. I will also note that as it stands this would only affect the ProteusLib specific property packages, so would hardly be general, which might be confusing to some users.

Also, will ask the question of how this would differ from initializing the properties at some known state (hopefully close to the final requirement), and then fixing/specifying the required properties after initialization. For example, this is how the isentropic pressure changer works (and actually even the energy balances). We start by initializng the isentropic state at a condition close to the final solution, and then add a constraints that fix pressure and entropy for the actual solve.

proteuslib/property_models/tests/property_test_harness.py
```diff
@@ -304,6 +304,16 @@ def test_badly_scaled(self, frame_stateblock):
             raise PropertyValueError(
                 "The following variable(s) are poorly scaled: {lst}".format(lst=lst))
 
+    @pytest.mark.component
+    def test_calculate_state(self, frame_stateblock):
+        m = frame_stateblock
+
+        # check if calculate_state runs and solves rather than how its used
```
adam-a-a, line 311: Can you add a test that demonstrates how supplying volumetric feed flow rate and mass concentration to `calculate_state` would lead to some expected set of results?

TimBartholomew: Yes, I haven't finished the PR yet. I'm planning on adding an individual test class in the property test harness for calculate_state cases. I'm also going to use it in the RO with energy recovery flowsheet.

TimBartholomew:
> @TimBartholomew What is the actual use case for this method, and how do you see it being used? I can see some potential issues with this, and it would help to know exactly what you intend this for. I will also note that as it stands this would only affect the ProteusLib specific property packages, so would hardly be general, which might be confusing to some users.
> 
> Also, will ask the question of how this would differ from initializing the properties at some known state (hopefully close to the final requirement), and then fixing/specifying the required properties after initialization. For example, this is how the isentropic pressure changer works (and actually even the energy balances). We start by initializng the isentropic state at a condition close to the final solution, and then add a constraints that fix pressure and entropy for the actual solve.

@andrewlee94 I see this method being used before initialization routines in a variety of cases.

1. Determining and setting inlet state variables when the user prefers to provide properties. For example, multiple proteuslib developers prefer to set up the inlet conditions with a flowrate and concentration rather than component mass flowrates. With this method then they can calculate the state with hold_state as true, and then use the initialization method and propagate states as is.
2. Determining state variables within system-wide initialization routines. Similar to 1, there are points within the initialization of a system that uses a pressure exchanger where I want to base a split fraction based on volumetric flowrate rather than state variables. With how our initialization tools are built though I need to calculate the state variables in order to propagate the state. In that routine, I fix the properties on a state block, check the degrees of freedom, solve the block, check the solver status, and unfix the properties. This method does that all in one method call and includes descriptive exceptions for the checks.
3. Determining state variables within a unit initialization method. There may be cases where initializations could be improved if state blocks were solved for some property.

Cases 1 and 2 will be demonstrated in this PR soon for the RO with energy recovery flowsheet, this was still a work in progress.

andrewlee94:
@TimBartholomew So, the first thing I can see is that a tool like this is going to be very property-package specific - i.e. I do not think it will be possible to build a generalised version of this. This means that any implementation is going to be case-specific, and it does mean that users might be confused when only certain property packages support this behaviour (and only for certain properties).

There may still be value in doing this, but I do see a lot of caveats and edge cases that will arise from this (and I do mean will). To put somethings in writing and allow some thought/discussion (note that these are written from a general stand point, many of these could be trivial for specific examples):

1. This will only work if the properties the user desired to fix are `Vars` - it will not work if they are written as `Expressions` (e.g. the generic property framework). So, what properties this can be used on will depend on the property package in question and how it is written.
2. Try to use this on an uninitialized property package could be (will be?) hit-or-miss, as intermediate variables may not be available. E.g. using concentration in place of mole fractions requires solving for density. For simpler property packages this might work, but it could easily become challenging for more complex property packages. The practical definition of state variables is that all other properties are explicit in terms of these. The reverse means that state variables are generally implicit in terms of other properties (if you have only one "other" property then you can back-solve for the missing state, but this is a very limited use case, and is still not necessarily simple).
3. Degrees of freedom could be challenging, and users could easily specify a degenerate set of conditions if they are not careful. This is especially true for inlets where there is one less degree of freedom due to sum of mole fraction constraints (or equivalent).
4. You are starting to move towards a combinatorial problem. The number of potential property specifications a user could "know" is huge, and each set of combinations require a different approach to resolve to a solvable state. Obviously, we cannot support a truly arbitrary set of specifications, so any implementation begins to become very case-specific.

In terms of the specific case studies you mentioned, I would suggest that in all cases you would get more reliable results by initializing the state block first at your best current guess (even if that is the inlet conditions), and then switching to your preferred definition of the system. Yes, it is one more step, but it avoids a lot of potential problems as the model constraints are generally written explicitly in terms of the state variables (that is what makes them state variables). Otherwise, you are now trying to solve a model based on implicit relationships which is much harder.

As I mentioned before, this is how the isentropic pressure changer model works. The isentropic state should be solved based on the inlet flowrate and composition, inlet specific entropy and outlet pressure. However, specific entropy is very rarely a state variable, so instead we initialize at the inlet flow, composition and temperature and outlet pressure. This is not correct, but is generally very close enough such that we can then just solve the entire unit with the isentropic constraint. A similar argument holds for the actual outlet condition (where the outlet specific enthalpy, pressure, flow and composition is known).

Of your listed use cases, case 1 is probably the most important as it is often the case that the user knows the inlet in one set of properties, but the calculations are better performed using a different set of state variables. However, even here my gut feeling is that it would be better to initialize the state block using the best guesses of the actual state variables, and to only apply the desired specification once the state is initialised. In terms of case 2 and 3, I very much think that is the case - just initialize using simple guesses based on the state variables and then apply the actual spec. Unless you have some really weird system, I cannot see your best guess being too far off the mark.

TLDR: It is almost always easier to initialize at a set of initial guesses for the state variable first and get a feasible solution which is close to your desired conditions, and to then move to the actual state you want rather than to try to directly solve for the condition you want by specifying properties derived from the state variables.

TimBartholomew:
@adam-a-a @bknueven @lbianchi-lbl have we seen these test failures before? 

1. py3.9/linux is failing in doc link checking, but I haven't touched any doc files.

2. py3.7/linux passed when I did my latest commit, but after I reran the tests trying to see if the above failure was a fluke, then this failed at installing the dependencies.

proteuslib/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
```diff
@@ -146,14 +157,15 @@ def set_operating_conditions(m, water_recovery=0.5, over_pressure=0.3, solver=No
...
-    m.fs.feed.flow_mass_phase_comp[0, 'Liq', 'H2O'].fix(feed_flow_mass * feed_mass_frac_H2O)
-    m.fs.feed.pressure.fix(101325)  # atmospheric pressure [Pa]
-    m.fs.feed.temperature.fix(273.15 + 25)  # room temperature [K]
+    # state variables
+    m.fs.feed.properties[0].pressure.fix(101325)  # feed pressure [Pa]
+    m.fs.feed.properties[0].temperature.fix(273.15 + 25)  # feed temperature [K]
+    # properties (cannot be fixed for initialization routines, must calculate the state variables)
+    m.fs.feed.properties.calculate_state(
+        var_args={('flow_vol_phase', 'Liq'): 1e-3,  # feed volumetric flow rate [m3/s]
+                  ('mass_frac_phase_comp', ('Liq', 'NaCl')): 0.035},  # feed NaCl mass fraction [-]
```
adam-a-a, line 166: Just to understand better--would users technically be able to enter any variables they want here? 

It would be interesting to see if @bknueven could repurpose his approach for assessing stability with parameter_sweep to this `calculate_state` method. The intention would be to probe common variable inputs and their effect on stability, which we could then advise the user about in documentation (and raise exceptions for combos we just don't want to allow).

adam-a-a:
> @adam-a-a @bknueven @lbianchi-lbl have we seen these test failures before?
> 
> 1. py3.9/linux is failing in doc link checking, but I haven't touched any doc files.
> 2. py3.7/linux passed when I did my latest commit, but after I reran the tests trying to see if the above failure was a fluke, then this failed at installing the dependencies.

@TimBartholomew in reply to #1, we just noticed the link-check failing in #188 which should now be resolved once we get one more approval to merge.

proteuslib/property_models/NaCl_prop_pack.py
```diff
@@ -283,6 +288,92 @@ def release_state(self, flags, outlvl=idaeslog.NOTSET):
...
+                                   "of freedom were {dof}, but 0 is required. Check var_args and ensure "
+                                   "the correct fixed variables are provided."
+                                   "".format(sb_name=sb.name, dof=degrees_of_freedom(sb)))
+
+        # Solve
+        with idaeslog.solver_log(solve_log, idaeslog.DEBUG) as slc:
+            results = solve_indexed_blocks(opt, [self], tee=slc.tee)
+            solve_log.info_high("Calculate state: {}.".format(idaeslog.condition(results)))
+
+        if not check_optimal_termination(results):
```
andrewlee94, line 359: A thought here (for the future): the IDAES Homotopy metasolver could be useful here for more challenging problems, and it would be worth thinking about adding support for using it.

andrewlee94 (commented):
A couple of quick observations.

proteuslib/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
bknueven: I wonder if this is the "right" interface. An alternative is:
```python
var_args={'flow_vol_phase[Liq]': 1e-3,  # feed volumetric flow rate [m3/s]
          'mass_frac_phase_comp[Liq, NaCl]': 0.035},  # feed NaCl mass fraction [-]
```
and using `find_component` or `component` in the `calculate_state` method implementation as oppose to `getattr`.

As for parameter_sweep, I think it would need some changes to know to look for a `calculate_state` method and know to do the appropriate thing. It's certainly possible, but I imagine it getting challenging for the user to communicate what gets set where, in additional to the effects of this method on the warm-start initialization provided by the prior solve.

bknueven: Regarding combinations we wish to exclude, I think that is relatively straightforward based on some work I did last week. We can define polyhedral regions in the unit model's variables and test for inclusion using ipopt or a linear programming solver.

adam-a-a: > I wonder if this is the "right" interface. An alternative is:
> 
> ```python
> var_args={'flow_vol_phase[Liq]': 1e-3,  # feed volumetric flow rate [m3/s]
>           'mass_frac_phase_comp[Liq, NaCl]': 0.035},  # feed NaCl mass fraction [-]
> ```
I think I agree with @bknueven on this one, since it seems the format he is proposing is more in line with that of IDAES and other parts of Proteuslib.

proteuslib/flowsheets/RO_with_energy_recovery/monte_carlo_sampling_RO_ERD.py
```diff
@@ -55,3 +55,4 @@ def run_parameter_sweep(results_file, seed=None):
 
 if __name__ == '__main__':
     run_parameter_sweep('output/monte_carlo_results.csv')
+    # run_parameter_sweep('output/monte_carlo_results.csv', seed=1)  # for running testing case
```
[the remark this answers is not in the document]
TimBartholomew: I put it there because to run the case for testing you need seed=1, it took me a few minutes to figure it out so it'd help anyone who changes the flowsheet and needs to update the monte carlo testing.

proteuslib/property_models/NaCl_prop_pack.py
TimBartholomew: Oh I didn't even know that existed

proteuslib/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
TimBartholomew: @adam-a-a and @bknueven ok you guys have pointed out two issues: 1) how does the parameter sweep tool know to use calculate_state, 2) API for inputting variables and an attribute should be consistent. 

After thinking about it, I don't think 1 is an issue because the parameter sweep likely doesn't need to use it. This calculate_state method would primarily be used for determining state variables from properties before initialization. Since users have to initialize the model before using the parameter sweep tool, they can just fix the properties (and unfix the state variables for the appropriate degrees of freedom) in subsequent simulation/optimization.

I think 2 is a fair point, but its hard to address because the API isn't consistent across IDAES and ProteusLib. I also want to loop @andrewlee94 for this API discussion. For example there are equivalent things like the following:

1. The form used in IDAES scaling methods, ProteusLib property test harness, and this PR where the dictionary key is a tuple with the variable name string and its index:
```python
 var_dict = {('temperature', None): 300,
             ('flow_mass_phase_comp', ('Liq', 'NaCl')): 1,
             ('flow_mass_phase_comp', ('Liq', 'H2O')): 2,
            }
```
2. The form used in IDAES initialization routines where the dictionary key is the variable name string then indices are keys in another nested dictionary:
```python
 var_dict = {'temperature': 300,
             'flow_mass_phase_comp': {('Liq', 'NaCl'): 1,
                                      ('Liq', 'H2O'): 2},
            }
```
3. The equivalent form used in parameter sweep where the dictionary key is the variable itself:
```python
 var_dict = {blk.temperature: 300,
             blk.flow_mass_phase_comp['Liq', 'NaCl']: 1,
             blk.flow_mass_phase_comp['Liq', 'H2O']: 2,
            }
```
4. Then this suggestion where the dictionary key is the variable name and index in a single string:
```python
 var_dict = {'temperature': 300,
             'flow_mass_phase_comp[Liq, NaCl]': 1,
             'flow_mass_phase_comp[Liq, H2O]': 2,
            }
```
I'm not sure what other ways IDAES or ProteusLib has used to input variables and an attribute as an argument, but they might be floating out there. 

I liked method 1 better than method 2 because I think it is cleaner than the nested dictionary. Method 3 makes a lot of sense too, but sometimes these dictionaries need to be created before the model is made, like in the property test harness where the solution values need to be provided to the test which later creates the model. Method 4 is interesting, I had not used find_component or component before.

proteuslib/flowsheets/RO_with_energy_recovery/monte_carlo_sampling_RO_ERD.py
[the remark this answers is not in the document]
andrewlee94: I would suggest a meaningful comment then, and not just a commented line of code. Something like:

```
# For testing cases, a seed needs to be provided as an additional argument, e.g. seed=1
```

Otherwise this just looks like a left over piece of code that will be ignored

proteuslib/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
andrewlee94: So, API discussions are open-ended, and as noted there is not a standard defined for this yet.

Regarding Method 2 - this was used for the initialization as it is a very specific case and that structure fits well with the structure. I.e., in initialization we have a pre-defined list of state variables which we can iterate over, so it is just a matter of matching names from the list with the keys in the `state_args` dict.

andrewlee94: Also, I just added this topic to the IDAES Core Developer meeting agenda for next week.

proteuslib/property_models/seawater_prop_pack.py
```diff
@@ -493,6 +493,91 @@ def release_state(self, flags, outlvl=idaeslog.NOTSET):
...
+                            "fixed to a value {val}, but it was already fixed to value {val_2}. "
+                            "Unfix the variable before calling the calculate_state "
+                            "method or update var_args."
+                            "".format(sb_name=sb.name, v_name=var.name, val=val, val_2=value(var[ind])))
+                else:
+                    flags[(k, v_name, ind)] = False
+                    var[ind].fix(val)
+
+            if degrees_of_freedom(sb) != 0:
+                raise RuntimeError("While using the calculate_state method on {sb_name}, the degrees "
```
adam-a-a, line 554: There's a new function in `proteuslib.util.initialization` called `check_dof` which can be used to handle these checks now. `check_dof` will check for DOF=0 and you can specify whether you want a warning or an error to show up (e.g., `check_dof(sb, fail_flag=True)` would raise an error when DOF!=0.)

However, observing your message here, `check_dof` may need an optional arg for a custom message and possibly some other tweaking (which is outside the scope of this PR).

proteuslib/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
TimBartholomew: @adam-a-a and @bknueven which forms do you guys prefer and why? I'm leaning towards form 1 because it is at least consistent with the property test harness that I added to ProteusLib. I agree full consistency should be the goal, but I won't be able to do that in this PR.

proteuslib/property_models/seawater_prop_pack.py
TimBartholomew: Yeah I think in this case because of the custom message its best to leave it as is. Plus at this moment I'm not sure what other functionality I'd gain by using it even if an input allowed a custom message, i.e. to check whether the degrees of freedom is 0 is only one line of code.

proteuslib/property_models/seawater_prop_pack.py
```diff
@@ -493,6 +493,91 @@ def release_state(self, flags, outlvl=idaeslog.NOTSET):
...
+                                   "of freedom were {dof}, but 0 is required. Check var_args and ensure "
+                                   "the correct fixed variables are provided."
+                                   "".format(sb_name=sb.name, dof=degrees_of_freedom(sb)))
+
+        # Solve
+        with idaeslog.solver_log(solve_log, idaeslog.DEBUG) as slc:
+            results = solve_indexed_blocks(opt, [self], tee=slc.tee)
+            solve_log.info_high("Calculate state: {}.".format(idaeslog.condition(results)))
+
+        if not check_optimal_termination(results):
```
adam-a-a, line 564: Note, similar to my previous comment, there is a new function called `check_solve` in `proteuslib.util.initialization` that is meant to check for optimal termination. When this condition is not satisfied, either a warning or an error will be thrown (based on your specification of the `fail_flag`). For example, your conditional here could be replaced by something like
` check_solve(results, logger=init_log, fail_flag=False, checkpoint=f'Calculate_state method on {sb_name}')`

proteuslib/property_models/seawater_prop_pack.py
adam-a-a: Sure- my comment was more of an FYI for you and anyone else who might read through these comments. I think what you have is fine. The same applies to my comment below about the new `check_solve` function.

andrewlee94: I would leave it as is as well. The `check_dof` tool is rather specific to initialization use-cases where the user is one step removed from the code.

proteuslib/property_models/seawater_prop_pack.py
andrewlee94: On that note, there is a Pyomo method that does the same thing that could be used directly here. The `check_solve` method is again specific to initialization where things are held at arms length, and we have specific steps (checkpoints) to record.

adam-a-a: @andrewlee94 What's the name of that Pyomo method?

andrewlee94: Sorry - you are already using it ('check_optimal_termination`).

proteuslib/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
bknueven: The `parameter_sweep` version (form 3) is subtly different -- the *keys* to the dictionary are just names, an the var and values are store in a value tuple, e.g.
```python
var_dict = { 'temp' : (blk.temperature, 300),
             'flow_mass_phase_comp_liq_NaCl' : (blk.flow_mass_phase_comp['Liq', 'NaCl'], 1),
             'flow_mass_phase_comp_liq_H2O': (blk.flow_mass_phase_comp['Liq', 'H2O'], 2),
            }
 ```
 
 In order to use python objects as "keys" you need to use pyomo.environ.ComponentMap, e.g.
 ```python
import pyomo.environ as pyo

var_map = ComponentMap()
var_map[blk.temperature] =  300
var_map[blk.flow_mass_phase_comp['Liq', 'NaCl']] = 1
var_map[blk.flow_mass_phase_comp['Liq', 'H2O']] = 2
 ```

**Review by adam-a-a — DISMISSED** [review]
Despite our collective comments related to the API discussion, I think all of that can be saved for another issue/PR. This looks good to me.

File: proteuslib/property_models/tests/test_seawater_prop_pack.py [hunk 7]
```diff
@@ -130,3 +130,60 @@ def configure(self):
...
+@pytest.mark.component
+class TestSeawaterCalculateState_1(PropertyCalculateStateTest):
+    def configure(self):
+        self.prop_pack = props.SeawaterParameterBlock
+        self.param_args = {}
+
+        self.solver = 'ipopt'
+        self.optarg = {'nlp_scaling_method': 'user-scaling'}
+
+        self.scaling_args = {('flow_mass_phase_comp', ('Liq', 'H2O')): 1e-1,
```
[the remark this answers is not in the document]
TimBartholomew: @adam-a-a so you are advocating for form 4 described above, right?

proteuslib/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
TimBartholomew: @bknueven yeah oops I should of been more clear that it wasn't exactly the form the parameter_sweep tool uses. I knew it used the short name, but I was trying to provide an option of a different key such as the actual variable itself.

I would actually prefer form 3 generally, but in a lot of cases these dictionaries are inputs into functions that will later create the model, so form 3 can't be used in those cases.

adam-a-a (dismissed):

proteuslib/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
bknueven: Right -- I agree that the ComponentMap version is the best when the model already exists. I has lots of advantages, such as being able to write:
```python
for var, val in var_map.items():
    # for example
    var.fix(val)
```
which is much cleaner than any of the other alternatives. I'm not necessarily a big fan for user-facing interfaces you also have to *tell* the user to use `ComponentMap` in place of `dict`, which may be confusing for new python/Pyomo users.

bknueven (approved):

adam-a-a (approved):

bknueven (approved):

PR #179, opened 2021-07-27
Fail on warning functionality in initialize method

adam-a-a:
## Summary/Motivation:
EDIT: Thanks to @bknueven for the thoughts and suggestions.

To fail or not to fail -- that is the question. This PR allows the `initialize()` method to throw either (i) a warning upon an unsuccessful solve or (ii) a failure upon an unsuccessful solve. 

All that the user must do is specify the `fail_on_warning` argument to be `True` in `initialize()`; otherwise, the default is set to `False` and only a warning is thrown. For example:

`my_RO_unit.initialize(fail_on_warning=True)`

- add `fail_on_warning` argument and functionality to `initialize()` method in 0d and 1d RO
- includes adding a `_raise_if_infeasible()` function which will throw either the warning or error after a failed solve
- includes a wrapper function, `_check_solve()`, which checks for optimal termination after a solve. If all is well, successful message given; if solve has failed, call the `_raise_if_infeasible()` function to determine whether warning or error should be raised.

### Legal Acknowledgement

### DIFF
--- a/proteuslib/unit_models/reverse_osmosis_0D.py
+++ b/proteuslib/unit_models/reverse_osmosis_0D.py
@@ -936,21 +936,23 @@ class ReverseOsmosisData(UnitModelBlockData):
         Returns:
             None
         """
-        def _raise_if_infeasible(fail_flag, fail_point=None):
-            if fail_point is None:
-                msg = 'The solver failed to converge to an optimal solution.' \
-                      'This suggests that the user provided infeasible inputs or ' \
-                      'that the model is poorly scaled.'
-            else:
-                msg = f"{fail_point} failed. The solver failed to converge to an optimal solution. " \
+        def _check_solve(results, checkpoint=None):
+            def _raise_if_infeasible(fail_flag):
+                msg = f"{checkpoint} failed. The solver failed to converge to an optimal solution. " \
                       f"This suggests that the user provided infeasible inputs or that the model is poorly scaled."
-            if fail_flag is True:
-                raise RuntimeError(msg)
-            elif fail_flag is False:
-                init_log.warning(msg)
+                if fail_flag is True:
+                    raise RuntimeError(msg)
+                elif fail_flag is False:
+                    init_log.warning(msg)
+                else:
+                    raise Exception(f'The fail_on_warning argument in the initialize method was set to {fail_flag}. '
+                                    f'fail_on_warning is a boolean argument. Set fail_on_warning to True or False.')
+            if checkpoint is None:
+                checkpoint = 'Initialization step'
+            if check_optimal_termination(results):
+                init_log.info(f'{checkpoint} successful.')
             else:
-                raise Exception(f'The fail_on_warning argument in the initialize method was set to {fail_flag}. '
-                                f'fail_on_warning is a boolean argument. Set fail_on_warning to True or False.')
+                _raise_if_infeasible(fail_flag=fail_on_warning)
 
 
         init_log = idaeslog.getInitLogger(blk.name, outlvl, tag="unit")
@@ -1068,11 +1070,7 @@ class ReverseOsmosisData(UnitModelBlockData):
         # Solve unit
         with idaeslog.solver_log(solve_log, idaeslog.DEBUG) as slc:
             res = opt.solve(blk, tee=slc.tee)
-        if check_optimal_termination(res):
-            pass
-        else:
-            _raise_if_infeasible(fail_flag=fail_on_warning,
-                                 fail_point='Initialization Step 3')
+        _check_solve(res, checkpoint='Initialization Step 3')
         # ---------------------------------------------------------------------
         # Release Inlet state
         blk.feed_side.release_state(flags, outlvl)
--- a/proteuslib/unit_models/reverse_osmosis_0D.py
+++ b/proteuslib/unit_models/reverse_osmosis_0D.py
@@ -25,8 +25,7 @@ from pyomo.environ import (Var,
                            Block,
                            units as pyunits,
                            exp,
-                           value,
-                           check_optimal_termination)
+                           value)
 from pyomo.common.config import ConfigBlock, ConfigValue, In
 # Import IDAES cores
 from idaes.core import (ControlVolume0DBlock,
@@ -38,7 +37,6 @@ from idaes.core import (ControlVolume0DBlock,
                         useDefault)
 from idaes.core.util.config import is_physical_parameter_block
 from idaes.core.util.exceptions import ConfigurationError
-from idaes.core.util.model_statistics import degrees_of_freedom
 from idaes.core.util import get_solver
 from idaes.core.util.tables import create_stream_table_dataframe
 import idaes.core.util.scaling as iscale
@@ -985,7 +983,7 @@ class ReverseOsmosisData(UnitModelBlockData):
             hold_state=True)
 
         init_log.info_high("Initialization Step 1 Complete.")
-        if ignore_dof is False:
+        if not ignore_dof:
             check_dof(blk, fail_flag=fail_on_warning, logger=init_log)
         # ---------------------------------------------------------------------
         # Initialize other state blocks
--- a/proteuslib/unit_models/reverse_osmosis_0D.py
+++ b/proteuslib/unit_models/reverse_osmosis_0D.py
@@ -893,8 +893,7 @@ class ReverseOsmosisData(UnitModelBlockData):
     def initialize(blk,
                    initialize_guess=None,
                    state_args=None,
-                   outlvl_init=idaeslog.INFO,
-                   outlvl_solve=idaeslog.NOTSET,
+                   outlvl=idaeslog.INFO,
                    solver=None,
                    optarg=None,
                    fail_on_warning=False,
@@ -917,8 +916,7 @@ class ReverseOsmosisData(UnitModelBlockData):
                          package(s) to provide an initial state for the inlet
                          feed side state block (see documentation of the specific
                          property package) (default = None).
-            outlvl_init : sets output level of InitLogger, init_log (default= idaeslog.INFO)
-            outlvl_solve sets output level of SolveLogger, solve_log (default= idaeslog.NOTSET)
+            outlvl_init : sets output level of initialization routine
             optarg : solver options dictionary object (default=None)
             solver : solver object or string indicating which solver to use during
                      initialization, if None provided the default solver will be used
@@ -929,8 +927,8 @@ class ReverseOsmosisData(UnitModelBlockData):
             None
         """
 
-        init_log = idaeslog.getInitLogger(blk.name, outlvl_init, tag="unit")
-        solve_log = idaeslog.getSolveLogger(blk.name, outlvl_solve, tag="unit")
+        init_log = idaeslog.getInitLogger(blk.name, outlvl, tag="unit")
+        solve_log = idaeslog.getSolveLogger(blk.name, outlvl, tag="unit")
         # Set solver and options
         if optarg is None:
             optarg = {'nlp_scaling_method': 'user-scaling'}
@@ -966,7 +964,7 @@ class ReverseOsmosisData(UnitModelBlockData):
 
         # Initialize feed inlet state block
         flags = blk.feed_side.properties_in.initialize(
-            outlvl=outlvl_init,
+            outlvl=outlvl,
             optarg=optarg,
             solver=solver,
             state_args=state_args,
@@ -1007,32 +1005,32 @@ class ReverseOsmosisData(UnitModelBlockData):
             state_args_interface_out['flow_mass_phase_comp'][('Liq', j)] *= initialize_guess['cp_modulus']
 
         blk.feed_side.properties_out.initialize(
-            outlvl=outlvl_init,
+            outlvl=outlvl,
             optarg=optarg,
             solver=solver,
             state_args=state_args_retentate,)
         blk.feed_side.properties_interface_in.initialize(
-                outlvl=outlvl_init,
+                outlvl=outlvl,
                 optarg=optarg,
                 solver=solver,
                 state_args=state_args_interface_in,)
         blk.feed_side.properties_interface_out.initialize(
-                outlvl=outlvl_init,
+                outlvl=outlvl,
                 optarg=optarg,
                 solver=solver,
                 state_args=state_args_interface_out,)
         blk.permeate_side.properties_mixed.initialize(
-            outlvl=outlvl_init,
+            outlvl=outlvl,
             optarg=optarg,
             solver=solver,
             state_args=state_args_permeate,)
         blk.permeate_side.properties_in.initialize(
-            outlvl=outlvl_init,
+            outlvl=outlvl,
             optarg=optarg,
             solver=solver,
             state_args=state_args_permeate,)
         blk.permeate_side.properties_out.initialize(
-            outlvl=outlvl_init,
+            outlvl=outlvl,
             optarg=optarg,
             solver=solver,
             state_args=state_args_permeate,)
@@ -1040,12 +1038,12 @@ class ReverseOsmosisData(UnitModelBlockData):
 
         # ---------------------------------------------------------------------
         # Solve unit
-        with idaeslog.solver_log(solve_log, outlvl_solve) as slc:
+        with idaeslog.solver_log(solve_log, idaeslog.DEBUG) as slc:
             res = opt.solve(blk, tee=slc.tee)
         check_solve(res, checkpoint='Initialization Step 3', logger=init_log, fail_flag=fail_on_warning)
         # ---------------------------------------------------------------------
         # Release Inlet state
-        blk.feed_side.release_state(flags, outlvl_init)
+        blk.feed_side.release_state(flags, outlvl)
         init_log.info(
             "Initialization Complete: {}".format(idaeslog.condition(res))
         )
--- a/proteuslib/unit_models/reverse_osmosis_0D.py
+++ b/proteuslib/unit_models/reverse_osmosis_0D.py
@@ -893,7 +893,7 @@ class ReverseOsmosisData(UnitModelBlockData):
     def initialize(blk,
                    initialize_guess=None,
                    state_args=None,
-                   outlvl=idaeslog.INFO,
+                   outlvl=idaeslog.NOTSET,
                    solver=None,
                    optarg=None,
                    fail_on_warning=False,
@@ -916,7 +916,7 @@ class ReverseOsmosisData(UnitModelBlockData):
                          package(s) to provide an initial state for the inlet
                          feed side state block (see documentation of the specific
                          property package) (default = None).
-            outlvl_init : sets output level of initialization routine
+            outlvl : sets output level of initialization routine
             optarg : solver options dictionary object (default=None)
             solver : solver object or string indicating which solver to use during
                      initialization, if None provided the default solver will be used
--- a/proteuslib/unit_models/reverse_osmosis_0D.py
+++ b/proteuslib/unit_models/reverse_osmosis_0D.py
@@ -43,6 +43,10 @@ import idaes.core.util.scaling as iscale
 from proteuslib.util.initialization import check_solve, check_dof
 import idaes.logger as idaeslog
 
+
+__author__ = "Tim Bartholomew, Adam Atia"
+
+# Set up logger
 _log = idaeslog.getLogger(__name__)
 
 
--- a/proteuslib/unit_models/reverse_osmosis_0D.py
+++ b/proteuslib/unit_models/reverse_osmosis_0D.py
@@ -42,6 +42,7 @@ from idaes.core.util.model_statistics import degrees_of_freedom
 from idaes.core.util import get_solver
 from idaes.core.util.tables import create_stream_table_dataframe
 import idaes.core.util.scaling as iscale
+from proteuslib.util.initialization import check_solve, check_dof
 import idaes.logger as idaeslog
 
 _log = idaeslog.getLogger(__name__)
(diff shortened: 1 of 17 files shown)


adam-a-a:
@bknueven @andrewlee94 @TimBartholomew- the PR should work and I don't think the errors from GHA are related to the PR, so feel free to proceed with review. 

@lbianchi-lbl @dangunter @ksbeattie - Something funky seems to be happening with GHA. Seeing this error for the windows runs: `ValueError: path is on mount 'D:', start on mount 'c:' `

bknueven (commented):
We can add coverage for this functionality if we set the `fail_on_warning=True` in some test.

We may also want to further generalize to the degrees of freedom check.

lbianchi-lbl:
@adam-a-a I agree with you that the failing CI checks are likely not related to the PR, but I can't understand what's going on.

The only thing I can think of is some mismatch with the location of the Python distributions between pip and Conda. I remember seeing a pip warning about an upcoming change in its behavior related to installing temporary files, which might or might not be related if e.g. an update of either pip, Conda, or the default configuration for the GHA runner happened recently.

I'll keep an eye on this: if the problem is not specific to Proteuslib, we should start seeing similar errors elsewhere as well.

lbianchi-lbl:
It looks like the version of pip installed by default could have something to do with this: compare e.g.:

- https://github.com/nawi-hub/proteuslib/runs/3177393586?check_suite_focus=true#step:4:583 (from this PR, failing)
- OK: https://github.com/nawi-hub/proteuslib/runs/3146937707?check_suite_focus=true#step:4:577 (from a check run in another PR in the last days, passing)

One way to address this would be to pin the pip version, at least in the GHA Windows environment. I'll open another PR to try this or other possible fixes.

proteuslib/unit_models/reverse_osmosis_0D.py
```diff
@@ -925,9 +931,29 @@ def initialize(blk, initialize_guess=None, state_args=None, outlvl=idaeslog.NOTS
...
+        def _check_solve(results, checkpoint=None):
+            def _raise_if_infeasible(fail_flag):
+                msg = f"{checkpoint} failed. The solver failed to converge to an optimal solution. " \
+                      f"This suggests that the user provided infeasible inputs or that the model is poorly scaled."
+                if fail_flag is True:
+                    raise RuntimeError(msg)
+                elif fail_flag is False:
+                    init_log.warning(msg)
+                else:
+                    raise Exception(f'The fail_on_warning argument in the initialize method was set to {fail_flag}. '
```
andrewlee94, line 948: Use a `ValueError` here rather than a bare `Exception`.

proteuslib/unit_models/reverse_osmosis_1D.py
```diff
@@ -1021,9 +1022,28 @@ def initialize(blk,
                      initialization (default = None, use default solver)
             optarg : solver options dictionary object (default=None, use default solver options)
 
+            fail_on_warning : boolean argument to fail or only produce  warning upon unsuccessful solve (default=False)
+
         Returns:
             None
         """
+        def _check_solve(results, checkpoint=None):
```
andrewlee94, line 1030: This looks like it should be a general utility method so that you do not need to re-implement it in each property package.

proteuslib/unit_models/reverse_osmosis_0D.py
```diff
@@ -925,9 +931,29 @@ def initialize(blk, initialize_guess=None, state_args=None, outlvl=idaeslog.NOTS
...
+                                    f'fail_on_warning is a boolean argument. Set fail_on_warning to True or False.')
+            if checkpoint is None:
+                checkpoint = 'Initialization step'
+            if check_optimal_termination(results):
```
andrewlee94, line 952: What exactly does this method test for, and what conditions pass/fail? Notably, how does it handle IPOPTs "Solved to acceptable level" result?

My guess is that it is the same as what we would test, but we need to make sure.

proteuslib/unit_models/reverse_osmosis_0D.py
```diff
@@ -925,9 +931,29 @@ def initialize(blk, initialize_guess=None, state_args=None, outlvl=idaeslog.NOTS
...
+                init_log.info(f'{checkpoint} successful.')
```
andrewlee94, line 953: Your logger level here differs from that of the original property package, and assumes `info` is always the desired level.

It might be worth making the logger level and optional argument as well.

andrewlee94 (changes requested):
My biggest comment here is that I think this would be better as a general IDAES utility tool that can be used in any property package, rather than something that has to be re-written for every property package.

I will also note that IDAES is trying to work on a way for initialization to return the final solver status of an initialization run, but that it will require some work.

proteuslib/unit_models/reverse_osmosis_0D.py
adam-a-a: > This function returns True if the termination condition for the solver is 'optimal', 'locallyOptimal', or 'globallyOptimal', and the status is 'ok"

The source is [here](https://github.com/Pyomo/pyomo/blob/4f6f523483ffbf0eade463b1872885a140b607f1/pyomo/opt/results/solver.py).

proteuslib/unit_models/reverse_osmosis_1D.py
adam-a-a: I think that would be good. Right now we only have a "Tools" directory, which contains the parameter_sweep tool. Do you think we should have a separate "util" folder, like IDAES, or could this live in "Tools"? 

Better yet--this might just belong in `idaes.core.util.initialization`, assuming this isn't redundant compared with what's already there. Should we reserve this for a subsequent PR (either here or on IDAES)?

proteuslib/unit_models/reverse_osmosis_0D.py
adam-a-a: Admittedly, I don't know too much about the different logger levels. I'll look into this.

proteuslib/unit_models/reverse_osmosis_1D.py
andrewlee94: I was suggesting a separate PR to IDAES (as you can't combine it with this one).

adam-a-a:
> It looks like the version of pip installed by default could have something to do with this: compare e.g.:
> 
> * https://github.com/nawi-hub/proteuslib/runs/3177393586?check_suite_focus=true#step:4:583 (from this PR, failing)
> * https://github.com/nawi-hub/proteuslib/runs/3146937707?check_suite_focus=true#step:4:577 (from a check run in another PR in the last days, passing)
> 
> One way to address this would be to pin the pip version, at least in the GHA Windows environment. I'll open another PR to try this or other possible fixes.

Thanks, @lbianchi-lbl ! So it seems the difference is between version 2.278.0 and 2.279.0, correct?

proteuslib/unit_models/reverse_osmosis_0D.py
andrewlee94: So, my only comment here is that we might be willing to accept an "Acceptable level" result from IPOPT, but I am not sure whether that would pass this test or not. However, that is also an edge case that we don't need to deal with immediately.

adam-a-a:
> We can add coverage for this functionality if we set the `fail_on_warning=True` in some test.
> 
> We may also want to further generalize to the degrees of freedom check.

@bknueven Will do! So basically add the some warning/error functionality for the DOF check, correct? I'll also add a test or two.

proteuslib/unit_models/reverse_osmosis_0D.py
andrewlee94: First, note that this only really applies if we are making this a general utility tool. For custom property packages, do whatever you want.

All you really need is a new argument at the top `loglvl=idaeslog.info` and then use `loglvl` wherever you are logging a message.

This will default to the `info` level, but allow users of the method to select a different logger level if desired to match their needs.

adam-a-a:
> My biggest comment here is that I think this would be better as a general IDAES utility tool that can be used in any property package, rather than something that has to be re-written for every property package.
> 
> I will also note that IDAES is trying to work on a way for initialization to return the final solver status of an initialization run, but that it will require some work.

I'm know that I am probably unaware of the complexities that can arise behind this, but would the following approach be problematic? If so, why?

- add an boolean argument to return solver status and termination condition
- if true, return `results.solver.status` and `results.solver.termination_condition`
- if false, pass

andrewlee94:
@adam-a-a The problem is not returning the solver status; it is stetting up everything else to expect the `initialize` routine to return a solver status. At the moment, everything is built around `initialize` returning nothing (or more accurately `None`) - we have to change other things to handle the fact that a solver status would be returned instead.

adam-a-a:
> @adam-a-a The problem is not returning the solver status; it is stetting up everything else to expect the `initialize` routine to return a solver status. At the moment, everything is built around `initialize` returning nothing (or more accurately `None`) - we have to change other things to handle the fact that a solver status would be returned instead.

That makes sense. Proteuslib might be a good place to test this out.

adam-a-a:
@andrewlee94 @bknueven I think I addressed your comments and adding some more features w/testing for 1DRO. If you are good with these changes, I'll apply the same for ODRO.

bknueven:
I think the 1D-RO looks good, though I do agree with @andrewlee94 that the `_check_solve` function common to both RO initialization routine should be put in a "common" location. I would suggest either `proteuslib/utils/initialization.py` or `proteuslib/common/initialization.py` (the former follows an IDAES convention, the later a Pyomo convention). If we accumulate other initialization helper functions they can be put in this file. Further "utility" or "common" modules could be defined for similar cases in the codebase.

adam-a-a:
> I think the 1D-RO looks good, though I do agree with @andrewlee94 that the `_check_solve` function common to both RO initialization routine should be put in a "common" location. I would suggest either `proteuslib/utils/initialization.py` or `proteuslib/common/initialization.py` (the former follows an IDAES convention, the later a Pyomo convention). If we accumulate other initialization helper functions they can be put in this file. Further "utility" or "common" modules could be defined for similar cases in the codebase.

@bknueven Do you think I should do that in this PR? My last impression from @andrewlee94 was that we should hold off and include this for the util directory in IDAES. However, if you think we should have them in both places (i.e., `proteuslib/utils/initialization.py` as well as `idaes/core/util/initialization.py`), I can add that in.

bknueven:
> @bknueven Do you think I should do that in this PR? My last impression from @andrewlee94 was that we should hold off and include this for the util directory in IDAES. However, if you think we should have them in both places (i.e., `proteuslib/utils/initialization.py` as well as `idaes/core/util/initialization.py`), I can add that in.

I think my answer depends on how soon this will be done in IDAES. From the strict perspective of ProteusLib, I would rather not have identical code in two places. If IDAES eventually adds similar/identical functionality we can always remove the version in ProteusLib or reference it in `proteuslib/utils/initialization.py`.

adam-a-a:
> > @bknueven Do you think I should do that in this PR? My last impression from @andrewlee94 was that we should hold off and include this for the util directory in IDAES. However, if you think we should have them in both places (i.e., `proteuslib/utils/initialization.py` as well as `idaes/core/util/initialization.py`), I can add that in.
> 
> I think my answer depends on how soon this will be done in IDAES. From the strict perspective of ProteusLib, I would rather not have identical code in two places. If IDAES eventually adds similar/identical functionality we can always remove the version in ProteusLib or reference it in `proteuslib/utils/initialization.py`.

Sounds reasonable to me!

lbianchi-lbl:
> > It looks like the version of pip installed by default could have something to do with this: compare e.g.:
> > 
> > * https://github.com/nawi-hub/proteuslib/runs/3177393586?check_suite_focus=true#step:4:583 (from this PR, failing)
> > * https://github.com/nawi-hub/proteuslib/runs/3146937707?check_suite_focus=true#step:4:577 (from a check run in another PR in the last days, passing)
> > 
> > One way to address this would be to pin the pip version, at least in the GHA Windows environment. I'll open another PR to try this or other possible fixes.
> 
> Thanks, @lbianchi-lbl ! So it seems the difference is between version 2.278.0 and 2.279.0, correct?

I was referring to the difference between the pip versions, i.e. `21.1.3` for the passing checks and `21.2.1` for the failing checks. What packages you were referring to with versions `2.278.0` and `2.279.0`?

adam-a-a:
> > > It looks like the version of pip installed by default could have something to do with this: compare e.g.:
> > > 
> > > * https://github.com/nawi-hub/proteuslib/runs/3177393586?check_suite_focus=true#step:4:583 (from this PR, failing)
> > > * https://github.com/nawi-hub/proteuslib/runs/3146937707?check_suite_focus=true#step:4:577 (from a check run in another PR in the last days, passing)
> > > 
> > > One way to address this would be to pin the pip version, at least in the GHA Windows environment. I'll open another PR to try this or other possible fixes.
> > 
> > 
> > Thanks, @lbianchi-lbl ! So it seems the difference is between version 2.278.0 and 2.279.0, correct?
> 
> I was referring to the difference between the pip versions, i.e. `21.1.3` for the passing checks and `21.2.1` for the failing checks. What packages you were referring to with versions `2.278.0` and `2.279.0`?

@lbianchi-lbl  I suppose it's not important anymore, but I was referring to `Current runner version: '2.279.0'` and `Current runner version: '2.278.0'` (line 1 of two links you provided) which was probably incorrect. Thanks for taking care of this!

bknueven (dismissed):

proteuslib/unit_models/reverse_osmosis_0D.py
```diff
@@ -956,17 +968,15 @@ def initialize(blk, initialize_guess=None, state_args=None, outlvl=idaeslog.NOTS
...
             solver=solver,
             state_args=state_args,
             hold_state=True)
 
         init_log.info_high("Initialization Step 1 Complete.")
-        if degrees_of_freedom(blk) != 0:
-            raise Exception(f"Initialization was called on {blk} "
-                            f"but it had {degrees_of_freedom(blk)} degree(s) of freedom "
-                            f"when 0 was expected. Check that the appropriate variables are fixed.")
+        if ignore_dof is False:
```
andrewlee94, line 978: Minor note for the future: it is considered more Pythonic to write `if not ignore_dof:` in cases like this.

proteuslib/unit_models/reverse_osmosis_0D.py
```diff
@@ -910,17 +919,20 @@ def initialize(blk, initialize_guess=None, state_args=None, outlvl=idaeslog.NOTS
                          package(s) to provide an initial state for the inlet
                          feed side state block (see documentation of the specific
                          property package) (default = None).
-            outlvl : sets output level of initialization routine
+            outlvl_init : sets output level of InitLogger, init_log (default= idaeslog.INFO)
```
andrewlee94, line 922: This may be a problem in some cases, depending on how the core IDAES code was written. Some unit models (or control volumes) may assume that `outlvl` is defined in all `initialize` methods, which will cause failures here.

Why did want to separate these two logger levels?

proteuslib/unit_models/reverse_osmosis_0D.py
```diff
@@ -26,7 +26,7 @@
                            units as pyunits,
                            exp,
                            value,
-                           assert_optimal_termination)
+                           check_optimal_termination)
```
andrewlee94, line 29: Is this import still needed here?

proteuslib/unit_models/reverse_osmosis_1D.py
```diff
@@ -25,7 +25,7 @@
                            Constraint,
                            Expression,
                            Block,
-                           assert_optimal_termination)
+                           check_optimal_termination)
```
andrewlee94, line 28: As before.

proteuslib/unit_models/tests/test_reverse_osmosis_0D.py
```diff
@@ -175,6 +175,74 @@ def test_option_pressure_change_calculated():
...
+
+    @pytest.mark.component
+    def test_initialize_ignore_dof(self, m):
+         # Unfix a variable and check that ignore_dof arg works
+        temp = m.fs.unit.N_Re_io[0, 'in'].value
+        m.fs.unit.N_Re_io[0, 'in'].unfix()
+        # Next line should pass since we are ignoring non-zero DOF
+        m.fs.unit.initialize(ignore_dof=True, fail_on_warning=False)
+        # Check that ValueError is thrown when ignore_dof and fail_on_warning are set to do so
+        with pytest.raises(ValueError):
```
andrewlee94, line 231: It is good practice to check the error message here to.

```
with pytest.raises(ValueError, match="expected message"):
```

proteuslib/util/initialization.py
```diff
@@ -0,0 +1,86 @@
...
+"""
+
+
+__author__ = "Adam Atia"
+
+from pyomo.environ import check_optimal_termination
+from idaes.core.util.model_statistics import degrees_of_freedom
+
+
+def check_solve(results, checkpoint=None, logger=None, fail_flag=False):
```
andrewlee94, line 24: I won't insist on it here (especially seeing as they will probably migrate to IDAES fairly soon), but it would be good to have unit tests for these methods.

andrewlee94 (commented):
A couple of minor observations, but my main question here is why you separated `outlvl` into two parts. This is likely to cause problems as `outlvl` is more-or-less a standard argument.

proteuslib/unit_models/reverse_osmosis_0D.py
adam-a-a: I set a second one specifically for the solver, in case someone wanted to set only that to DEBUG while setting whatever they want for the `init_log`. I suppose I was just riffing, thinking it would add further flexibility for troubleshooting initialization issues, but if you think it'll cause problems, I can just go back to using the one logger level.

andrewlee94: Actually - it generally won't be a problem as this is a unit model itself (I am too used to thinking from the bottom up). However, it could be an issue for someone using the automatic sequential decomposition tool on a flowsheet who wants to set the `outlvl` option. Most core models only have `outlvl`, so these models will be an exception which would mean you need to catch and real with this exception when calling `initialize`.

The reason we only had one `outlvl` was that we figured that if you wanted to see the full solver output, you probably also needed most of the debug information to put it in context. With the two separated, you could in theory set the solver output to debug and the general output to none and only see the solver output, but without any context to tell you what each solve was doing.

adam-a-a: I think I recall the reason why I did this now; I noticed that several initialization routines don't use `outlvl` for the solver_log. Instead, the logger level typically seems to be hard-coded to DEBUG as follows:
`with idaeslog.solver_log(solve_log, idaeslog.DEBUG) as slc:` 

So I figured--why not give the user the option to set this logger level rather than having it be hard-coded (even though setting to DEBEUG is perfectly fine and logical)? 

Additionally, could the potential issue with flowsheets be resolved by 

1. renaming `outlvl_int` to `outlvl` and then setting `outlvl_solve=None` and 
2. include a conditional within the `initialize()` function that sets `outlvl_solve` to some default logger level if it is `None`? 

Overall, it seems less clean to do this, and I'm planning to change things back to how they were, as you're suggesting. I just wanted to think more about the hypothetical remedy before doing so.

proteuslib/unit_models/tests/test_reverse_osmosis_0D.py
adam-a-a: Thanks- I was trying to recall how to do this from memory but skipped over it.

proteuslib/unit_models/reverse_osmosis_0D.py
andrewlee94: You could do that, but I would be inclined to just have the single `outlvl`. I don't see too much need for having separate output levels for the solver and the general logger. There generally isn't too much debug information anyways, and if you are looking at solver outputs during initialization you are probably trying to work out why it didn't work, which means that any debug messages are probably useful to you too.

I noted the issue with hard coding the solver output level whilst reviewing this, and my response would have been that they should have used `outlvl` rathe than hard coding it.

proteuslib/unit_models/reverse_osmosis_0D.py
```diff
@@ -891,7 +890,14 @@ def eq_over_pressure_ratio(b, t):
                     * (b.feed_side.properties_out[t].pressure_osm
                     - b.permeate_side.properties_out[t].pressure_osm))
 
-    def initialize(blk, initialize_guess=None, state_args=None, outlvl=idaeslog.NOTSET, solver=None, optarg=None):
+    def initialize(blk,
+                   initialize_guess=None,
+                   state_args=None,
+                   outlvl=idaeslog.INFO,
```
andrewlee94, line 896: We generally use NOTSET here, as that tell IDAES to use whatever default the user has specified in their global configuration.

proteuslib/unit_models/reverse_osmosis_0D.py
```diff
@@ -910,11 +916,13 @@ def initialize(blk, initialize_guess=None, state_args=None, outlvl=idaeslog.NOTS
                          package(s) to provide an initial state for the inlet
                          feed side state block (see documentation of the specific
                          property package) (default = None).
-            outlvl : sets output level of initialization routine
+            outlvl_init : sets output level of initialization routine
```
andrewlee94, line 919: Should be `outlvl`

proteuslib/unit_models/reverse_osmosis_1D.py
```diff
@@ -988,9 +986,11 @@ def initialize(blk,
                    feed_side_args=None,
                    permeate_side_args=None,
                    permeate_block_args=None,
-                   outlvl=idaeslog.NOTSET,
+                   outlvl=idaeslog.INFO,
```
andrewlee94, line 989: As above.

andrewlee94 (changes requested):
A few minor comments.

proteuslib/util/initialization.py
adam-a-a: @andrewlee94 OK, I added unit tests for those methods despite the potential migration.

andrewlee94 (approved):
Looks good.

bknueven (approved):

PR #188, opened 2021-07-30
Add docs section on using pip to update existing dev environments

lbianchi-lbl:
## Fixes/Addresses:

Adding these steps in the docs should help address present and future issues (as experienced by @aladshaw3 and @srikanthallu) when updating `proteuslib` and its dependencies in an existing development environment using pip.

### Legal Acknowledgement

### DIFF
--- a/docs/getting_started/install.rst
+++ b/docs/getting_started/install.rst
@@ -18,7 +18,7 @@ as they provide a way to seamlessly switch between different projects without co
 
 Using Conda environments is not mandatory to be able to install and use ProteusLib; however, it is strongly recommended.
 
-To use Conda environments, the ``conda`` package manager is required. Refer to the `Conda installation guide <https://conda.io/projects/conda/en/latest/user-guide/install/index.html>`_ for detailed steps on how to install Conda for your operating system.
+To use Conda environments, the ``conda`` package manager is required. Refer to the `Conda installation guide <https://docs.conda.io/projects/conda/en/latest/user-guide/install/>`_ for detailed steps on how to install Conda for your operating system.
 
 General installation
 --------------------
--- a/docs/getting_started/install.rst
+++ b/docs/getting_started/install.rst
@@ -93,6 +93,36 @@ If you plan to contribute to ProteusLib's codebase, choose this option.
 
 		pytest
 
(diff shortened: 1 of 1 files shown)


adam-a-a (approved):
Nice, @lbianchi-lbl ! I was just having issues myself with setting up a Linux VM and trying to get proteus going there. This is extremely helpful to have. I'm not sure (yet) if my problem will be completely fixed by running `pip --no-cache-dir install --force-reinstall -r requirements-dev.txt`, but either way, we need this in the documentation.

andrewlee94 (dismissed):

adam-a-a:
@lbianchi-lbl I'm not sure if you wanted to add anything to this PR. If not, we're ready to merge.

lbianchi-lbl:
@adam-a-a As far as I'm concerned this is good to go, however it looks like the link-checking step is failing, curiously enough only for one Python version but not the others.

adam-a-a:
> @adam-a-a As far as I'm concerned this is good to go, however it looks like the link-checking step is failing, curiously enough only for one Python version but not the others.

@lbianchi-lbl Interesting- it seems that link-checking steps were disabled for all Python versions except one; I think this was done to save time. So there's a good chance the check would fail for the other versions. The error seems to be related to a link for the Conda Installation guide. I found this link that we can try if we want to replace the old one that's resulting in the error: https://docs.conda.io/projects/conda/en/latest/user-guide/install/

adam-a-a:
To be more precise, this is the error we are seeing:

`(getting_started/install: line   21) broken    https://conda.io/projects/conda/en/latest/user-guide/install/index.html - HTTPSConnectionPool(host='conda.io', port=443): Max retries exceeded with url: /projects/conda/en/latest/user-guide/install/index.html (Caused by SSLError(SSLCertVerificationError("hostname 'conda.io' doesn't match either of '*.readthedocs-hosted.com', 'readthedocs-hosted.com'")))`

adam-a-a (approved):

lbianchi-lbl:
@adam-a-a oh, this makes sense, thanks for the clarification, and for fixing the link.

TimBartholomew (approved):

PR #190, opened 2021-08-02
Minor tweak of check_solve 

adam-a-a:
## Summary/Motivation:
This PR fixes the output from check_solve when a solve is successful. Previously, if a solve was successful, the printed/logged message was "None successful." 

- minor fix to change message when solve is successful, but no string provided for `checkpoint`

### Legal Acknowledgement


andrewlee94 (approved):
Looks fine. 

You don't have to change anything, but an alternative would have been to have the following earlier in the code:

```
if checkpoint is None:
    checkpoint = Solve.
```

TimBartholomew (approved):

Issue #168, opened 2021-07-16
Need documentation for scaling chemical processes


aladshaw3:
Scaling for chemical processes is a consistent issue since we have to solve systems involving extremely low concentrations. As a result, the scaling is a critical component of running a chemical process model. This needs to be added to documentation. 

- [x] Create "How-to" guide for scaling chemical reactions
- [x] Create "How-to" guide for scaling variables chemical processes
- [x] Create "How-to" guide for inherent reactions vs other reactions

aladshaw3:
@andrewlee94 I have you tagged on this issue primarily for discussion purposes. I can do the writing for the docs on this.

PR #198, opened 2021-08-12
Chemistry Scaling Docs

aladshaw3:
## Fixes/Addresses:

#168 

## Summary/Motivation:

This PR adds documentation on scaling for chemical processes. Scaling of these models is a consistent and re-occurring issue, so these documents will be continuously updated as our scaling process changes. These docs represent the current scaling methods employed in our testing. I also updated the how to guide for basic chemistry modules to reflect our change over from molarity to molefraction in testing. 

- Added how to for scaling reactions
- Added how to for scaling chemical species 
- Added how to for using inherent reactions
- Updated how to for basic chemistry 
- Updated indexing for how to guides 

### Legal Acknowledgement

### DIFF
--- a/docs/how_to_guides/how_to_scale_chemical_reactions.rst
+++ b/docs/how_to_guides/how_to_scale_chemical_reactions.rst
@@ -4,7 +4,7 @@ How to scale chemical reactions
 ===============================
 
 ..warning::
-    DO NOT use 'user-scaling' as a solver option when scaling any chemistry module. 
+    DO NOT use 'user-scaling' as a solver option when scaling any chemistry module.
 
 No model in ProteusLib can solve without proper scaling of the constraints
 and variables within that model. This can be a difficult task for aqueous
@@ -38,7 +38,8 @@ different scenarios will be discussed and demonstrated below.
 
     IMPORTANT: Scaling of just the chemical reactions is insufficient for solving
     a chemistry module. User's MUST also scale the chemical species in the system
-    due to the dilute nature of aqueous chemistry. See TBD.
+    due to the dilute nature of aqueous chemistry. See
+    :ref:`How to scale chemical species<how_to_scale_chemical_species>`.
 
 
 Types of Reactions
--- a/docs/how_to_guides/how_to_scale_chemical_reactions.rst
+++ b/docs/how_to_guides/how_to_scale_chemical_reactions.rst
@@ -3,7 +3,7 @@
 How to scale chemical reactions
 ===============================
 
-..warning::
+.. warning::
     DO NOT use 'user-scaling' as a solver option when scaling any chemistry module.
 
 No model in ProteusLib can solve without proper scaling of the constraints
--- a/docs/how_to_guides/how_to_scale_chemical_reactions.rst
+++ b/docs/how_to_guides/how_to_scale_chemical_reactions.rst
@@ -79,3 +79,41 @@ of that calculated **k_eq** value.
 Lastly, we perform a scaling transformation of the **inherent_equilibrium_constraint**. In our case,
 we scale this constraint down by a factor of 0.1 because the **log_power_law_equil** constraint
 form is already scaled up reasonably well.
+
+
+Equilibrium Reactions using log_power_law_equil form
+----------------------------------------------------
+
+Equilibrium reactions (see :ref:`How to setup simple chemistry<how_to_setup_simple_chemistry>`)
+are placed into the **GenericReactionParameterBlock**. As such, the scaling methods employed
+must be applied to the **rxn_params** for this set of reactions. The code segment below
+shows a demonstration of scaling factors that generally work well for these types of reactions.
+
+
+Equilibrium Scaling Demonstration
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. code-block::
+
+    # Equilibrium reactions have eps in the 'rxn_params'
+    factor = 1e-4
+    for rid in model.fs.rxn_params.equilibrium_reaction_idx:
+        scale = value(model.fs.unit.control_volume.reactions[0.0].k_eq[rid].expr)
+        # Want to set eps in some fashion similar to this
+        if scale < 1e-16:
+            model.fs.rxn_params.component("reaction_"+rid).eps.value = scale*factor
+        else:
+            model.fs.rxn_params.component("reaction_"+rid).eps.value = 1e-16*factor
+
+    for i in model.fs.unit.control_volume.equilibrium_reaction_extent_index:
+        scale = value(model.fs.unit.control_volume.reactions[0.0].k_eq[i[1]].expr)
+        iscale.set_scaling_factor(model.fs.unit.control_volume.equilibrium_reaction_extent[0.0,i[1]], 10/scale)
+        iscale.constraint_scaling_transform(model.fs.unit.control_volume.reactions[0.0].
+                equilibrium_constraint[i[1]], 0.1)
+
+
+.. note::
+
+    These scaling arguments are identical to the **Inherent Reaction** scaling methods,
+    however, because these reactions exist in a different location of the model, we
+    showed this here for completeness. All reactions, regardless of location, need scaling.
--- a/docs/how_to_guides/how_to_scale_chemical_reactions.rst
+++ b/docs/how_to_guides/how_to_scale_chemical_reactions.rst
@@ -22,6 +22,13 @@ different scenarios will be discussed and demonstrated below.
     see :ref:`How to setup simple chemistry<how_to_setup_simple_chemistry>`.
 
 
+.. note::
+
+    The scaling factor constants used in these demonstrations are the best found
+    for these types of problems based on numerical testing of a variety of problems.
+    These values can and may need to change for your particular problem. 
+
+
 Types of Reactions
 ------------------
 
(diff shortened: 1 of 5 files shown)


aladshaw3:
I am not using the 'testcode' blocks here because that would require significantly more overhead. Only basic code segments for specifically applying scaling are provided for simplicity.

docs/how_to_guides/how_to_scale_chemical_reactions.rst
```diff
@@ -0,0 +1,158 @@
...
+    # Scaling for kinetic reactions
+    for i in model.fs.rxn_params.rate_reaction_idx:
+        scale = value(model.fs.unit.control_volume.reactions[0.0].reaction_rate[i].expr)
+        iscale.set_scaling_factor(model.fs.unit.control_volume.rate_reaction_extent[0.0,i], 10/scale)
+
+
+Stoichiometric Reactions
+------------------------
+
+TBD
```
adam-a-a, line 158:
```suggestion
.. note::

    Documentation under development.
```

docs/how_to_guides/how_to_scale_chemical_reactions.rst
```diff
@@ -0,0 +1,158 @@
...
+            model.fs.thermo_params.component("reaction_"+rid).eps.value = 1e-16*factor
+
+    for i in model.fs.unit.control_volume.inherent_reaction_extent_index:
+        scale = value(model.fs.unit.control_volume.properties_out[0.0].k_eq[i[1]].expr)
+        iscale.set_scaling_factor(model.fs.unit.control_volume.inherent_reaction_extent[0.0,i[1]], 10/scale)
+        iscale.constraint_scaling_transform(model.fs.unit.control_volume.properties_out[0.0].
+                inherent_equilibrium_constraint[i[1]], 0.1)
+
+
+What this code segment is doing is to first assign an **eps** value for the **log_power_law_equil**
```
adam-a-a, line 86: I think this part could benefit from a little more clarification, perhaps through an example showing how `eps` is directly used in the `log_power_law_equil`.

docs/how_to_guides/how_to_scale_chemical_reactions.rst
```diff
@@ -0,0 +1,158 @@
...
+function. The default value is generally insufficient, so we change that value to always be at least
+4 orders of magnitude less than the **k_eq** value calculated for each reaction.
```
adam-a-a, line 88: Since you apply 10/scale in line 81, should it technically be 5 orders of magnitude?

docs/how_to_guides/how_to_scale_chemical_reactions.rst
```diff
@@ -0,0 +1,158 @@
...
+    showed this here for completeness. All reactions, regardless of location, need scaling.
+
+
+Rate Reactions using power_law_rate form
+----------------------------------------
+
+Rate reactions only exist in the **GenericReactionParameterBlock** and so these scaling
+arguments apply to **rxn_params** for these types of reactions. These are much simpler to
+scale than both the **Inherent** and **Equilibrium** reactions, but are just as important
+to apply scaling for. Below is a demonstration off applying scaling.
```
adam-a-a, line 142:
```suggestion
to apply scaling for. Below is a demonstration of applying scaling.
```

docs/how_to_guides/how_to_scale_chemical_reactions.rst
```diff
@@ -0,0 +1,158 @@
+.. _how_to_scale_chemical_reactions:
+
+How to scale chemical reactions
+===============================
+
+.. warning::
+    DO NOT use 'user-scaling' as a solver option when scaling any chemistry module.
```
adam-a-a, line 7: Should we rephrase this to sound less absolute? Perhaps something along the lines of 'user-scaling' not currently supported, recommending not to use this, and for users to try at their own risk.

Furthermore, a note on what the default solver option is, which is sufficient for chemistry modules (for those who might read this without knowing anything about the solver options).

docs/how_to_guides/how_to_scale_chemical_species.rst
```diff
@@ -0,0 +1,52 @@
+.. _how_to_scale_chemical_species:
+
+How to scale chemical species
+=============================
+
+.. warning::
+    DO NOT use 'user-scaling' as a solver option when scaling any chemistry module.
```
adam-a-a, line 7: same as comment above for scaling chem rxns

docs/how_to_guides/how_to_scale_chemical_species.rst
```diff
@@ -0,0 +1,52 @@
...
+
+.. note::
+
+    The following example assumes that your unit model is named **unit** and
+    your **FlowsheetBlock** is named **fs**. For information on setup of the
+    basic chemistry modules,
+    see :ref:`How to setup simple chemistry<how_to_setup_simple_chemistry>`.
+
+For this example, our **GenericParameterBlock** (named **thermo_params**) is using
+the "state_definition" of **FTPx**. This may change how what variables the scaling
```
adam-a-a, line 33:
```suggestion
the "state_definition" of **FTPx**. This may change what variables the scaling
```

docs/how_to_guides/how_to_scale_chemical_reactions.rst
aladshaw3: I could add a link to the IDAES docs on this method (if those documents exist), however, based on conversations with Andrew recently about future IDAES changes, it sounds like this method is going to be replaced and then there is no eps value to set. I'm not sure how much effort we want to put into talking about something that won't be around long.

docs/how_to_guides/index.rst
```diff
@@ -8,6 +8,9 @@ How To Guides
    how_to_setup_RO_config_options
    how_to_setup_simple_chemistry
    how_to_use_apparent_and_true_chemical_species
+   how_to_use_inherent_reactions
+   how_to_scale_chemical_reactions
+   how_to_scale_chemical_species
    how_to_scale_a_model
```
adam-a-a, line 14: I think it would make more sense for `how_to_scale_a_model` to come before any of the other scaling docs because it gives a general intro to scaling in general. 

However, if you look at the last section of `how_to_scale_a_model`, it recommends using `user-scaling`, which was meant to be targeted toward the RO model. So we should probably specify that since the chemistry scaling docs recommend not using `user-scaling`. This contradiction between the two could confuse readers.

docs/how_to_guides/how_to_scale_chemical_reactions.rst
aladshaw3: No. The eps value does not impact the scaling factors (significantly) because of how it is used in the function. It is a smoothing parameter for the log function. However, it is important and convenient to set this value here. 

Also, there would not be much need to go into details on the eps value, because it will go away with a future iteration of IDAES.

docs/how_to_guides/how_to_scale_chemical_reactions.rst
aladshaw3: I feel like this would need to be its own documentation. It would be aligned with the initialization and solve steps of a chemistry module, since that is where you actually invoke the solver and set solver options. I put the warning here because this is directly related to scaling, but solver options should be their own documentation, but maybe link those documents to each other?

docs/how_to_guides/how_to_scale_chemical_reactions.rst
adam-a-a: Fair enough. However, since we are going to have this up and people could technically read it before those changes come about, maybe you could just add in a segment of what you explained in your comment below, e.g.:
"... assign an **eps** value (which is a smoothing parameter) for the **log_power_law_equil** function."

docs/how_to_guides/index.rst
aladshaw3: At this moment, I am not sure we can do much about this contradiction. The chemistry modules generally don't solve with user-scaling and the RO modules generally don't solve with gradient-based scaling. Once this issue is ultimately resolved, after we get a working flowsheet, then we can update all documentation to reflect those changes. 

Also, it may make more sense for <code>how_to_scale_a_model</code> to be changed to <code>how_to_scale_the_RO_model</code> since the current title implies it is universally applicable.

docs/how_to_guides/how_to_scale_chemical_reactions.rst
adam-a-a: A link between those docs would be great, but I think for now, just removing the words "DO NOT" to something more flexible like "It is not recommended to" would suffice for this PR. If you go beyond that (e.g., providing a link to documents related to solver options, etc. ) even better.

adam-a-a (changes requested):
This documentation looks very good and will be helpful for working with chemistry modules. I just have a couple of suggestions before merging.

docs/how_to_guides/index.rst
adam-a-a: I totally agree with regard to the contradiction and resolving the discrepancy between scaling chem and RO modules. I think we can provide a temporary remedy by mentioning something like the following in your initial warning about user-scaling: "While "user-scaling" is currently the most stable option for unit models such as RO, we do not recommend using this option for chemistry modules at this time." 

Another note: that last section in `how_to_scale_a_model` is actually outdated in the way that it calls for the solver, so it seems that little section will need updating (not for this PR). I think we should keep the generic title of 'how_to_scale_a_model' though.

docs/how_to_guides/how_to_scale_chemical_reactions.rst
```diff
@@ -0,0 +1,164 @@
+.. _how_to_scale_chemical_reactions:
+
+How to scale chemical reactions
+===============================
+
+.. warning::
+    This scaling methods are meant to be used with the 'gradient-based' scaling
```
adam-a-a, line 7:
```suggestion
    These scaling methods are meant to be used with the 'gradient-based' scaling
```

docs/how_to_guides/index.rst
adam-a-a, line 14:
```suggestion
   how_to_use_inherent_reactions
   how_to_scale_a_model
   how_to_scale_chemical_reactions
   how_to_scale_chemical_species
```

adam-a-a (approved):
LGTM

srikanthallu (approved):
LGTM

Issue #163, opened 2021-07-15
EDB: Fix how ions are defined in the schema and database


dangunter:
- [x] The "anion" and "cation" types are required to be associated with a "charge"
- [x] Defined "anion" and "cation" as "liquidPhase", but ions should be "aqueousPhase"

All issues are the ions-- if I remove the ions from the list of species I ask for, it builds fine

aladshaw3:
For consistency, whenever I have ions in solution, I define all 'liquids' as 'aqeuousPhase'. Definitely the ions should be 'aqueousPhase', but I also define things like 'H2O' and 'H2CO3' as 'aqueousPhase' as well. Not sure how much that matters. It depends on how IDAES differentiates between 'liquidPhase' and 'aqueousPhase'.

andrewlee94:
For the purposes of this, all electrolyte phases (i.e. phases containing ions) should be `aqueousPhases`. I believe IDAES will raise an Exception if you try to declare `Ions` in anything other than an `AqueousPhase`.

dangunter:
I will change the word "liquid" to "aqueous"

PR #204, opened 2021-08-19
Fix EDB handling of ions

dangunter:
## Fixes/Addresses: #163 

- [x] Fix data to include proper type (Solute, Solvent, Anion, Cation) and charge for cation and anions.
- [x] Fix component, reaction, base data as needed to match generated configs used in tests
- [x] Add methods as needed to modify configs after fetch from DB
- [x] Add methods to fetch from DB by additional criteria
- [x] Test with simple water reactions

 Test with simple water reactions with eNRTL -- next PR(s)!

### Legal Acknowledgement


dangunter:
@aladshaw3 (maybe @andrewlee94 ), in the test for the eNRTL water dissociation (`test_enrtl_water_pH.py`), the parameter data in the test file for the H2CO3_Ka1 and _Ka2 is different from what I have in the database (from earlier data). Should I update the database files to match, or are these alternate formulations that should both be preserved?

dangunter:
It's not pretty, and arguably not even that good -- but it's a step forward, but I swear I will next document and test what's there, so others can maybe someday use it for real scientific/engineering tasks.

aladshaw3 (approved):
LGTM - I believe the handling of ions is correct now. I can dig more into the skipped tests later, which I think are unrelated to the handling of the ions.

andrewlee94 (approved):
I am going to assume this is good.

PR #216, opened 2021-09-03
Update requirements for IDAES to 1.11.0 release

lbianchi-lbl:
## Summary/Motivation:

- Update requirements for IDAES to latest released version (1.11.0)

- Set requirements for `idaes-pse` in `setup.py` to `>=1.11.0` for both `RELEASE` and `PRERELEASE` variants 

### Legal Acknowledgement

### DIFF
--- a/proteuslib/tests/test_enrtl_water_pH.py
+++ b/proteuslib/tests/test_enrtl_water_pH.py
@@ -848,8 +848,8 @@ class TestENRTLcarbonicAcid():
         pH = -log10(value(model.fs.unit.control_volume.properties_out[0.0].act_phase_comp["Liq","H_+"])*55.2)
         pOH = -log10(value(model.fs.unit.control_volume.properties_out[0.0].act_phase_comp["Liq","OH_-"])*55.2)
 
-        assert pytest.approx(8.29618, rel=1e-4) == pH
-        assert pytest.approx(5.72455, rel=1e-4) == pOH
+        assert pytest.approx(8.28, rel=1e-2) == pH
+        assert pytest.approx(5.72, rel=1e-2) == pOH
         assert pytest.approx(14.00, rel=1e-2) == pH + pOH
 
         gamma = {}
--- a/proteuslib/tests/test_enrtl_water_pH.py
+++ b/proteuslib/tests/test_enrtl_water_pH.py
@@ -690,7 +690,8 @@ carbonic_thermo_config = {
                        #        HOWEVER, the typical Ka1 dissociation constant of
                        #        10**-10.35 is defined on a molar basis. Thus, we must
                        #        divide by the total (~55.2 M) concentration raised to the
-                       #        net reaction order (i.e., 1 in this case).                       "dh_rxn_ref": (0, pyunits.kJ/pyunits.mol),
+                       #        net reaction order (i.e., 1 in this case).                       
+                       "dh_rxn_ref": (0, pyunits.kJ/pyunits.mol),
                        "k_eq_ref": (10**-10.35/55.2,pyunits.dimensionless),
                        "T_eq_ref": (300, pyunits.K),
 
--- a/proteuslib/tests/test_enrtl_water_pH.py
+++ b/proteuslib/tests/test_enrtl_water_pH.py
@@ -631,12 +631,9 @@ carbonic_thermo_config = {
                        #        HOWEVER, the typical Kw dissociation constant of
                        #        1e-14 is defined on a molar basis. Thus, we must
                        #        divide by the total (~55.2 M) concentration raised to the
-                       #        net reaction order (i.e., 2 in this case). Furthermore,
-                       #        IDAES internally does molar basis as mol/m**3, thus,
-                       #        we divide here by 1000**2 to perform both necessary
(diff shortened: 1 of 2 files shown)


andrewlee94 (approved):
Looks OK to me.

lbianchi-lbl:
> Looks OK to me.

The tests seem to think otherwise ;) Seriously though, are these failures expected?

andrewlee94:
@lbianchi-lbl I think at least some of these are due to a bug fix I had for IDAES that probably got merged yesterday. This corrected the units of some equilibrium and rate constants, and I am guessing some of our examples had inherited these incorrect units and are thus now failing.

@aladshaw3 @srikanthallu should probably be made aware of this.

lbianchi-lbl:
@andrewlee94 thanks for the clarification. If I'm understanding this correctly, this is something that should be fixed on the Proteuslib side (i.e. the current behavior in IDAES is expected and considered correct going forward), right?

andrewlee94:
@lbianchi-lbl I think so, although I didn't have much time to look at the issue closely.

aladshaw3:
@lbianchi-lbl @andrewlee94 The issue is caused by the units in the 'test_enrtl_water_ph.py' test. In that file there are 2 configuration, dictionaries where all the units for the 'k_eq_ref' need to be made dimensionless. 

Additionally, you need to remove each factor of '1000' in the value of 'k_eq_ref'. This value was acting to undo what the error in units was doing before. Thus, the expected behavior is by changing the units to dimensionless all tests will pass, except for when the solution is checked. Checking the solution should pass by removing the factors of '1000' from the values of 'k_eq_ref'.

lbianchi-lbl:
@aladshaw3 Thanks for the explanation! It was clear enough that I managed to bumble my way through the test and make the changes.

There seems to be only one test failing now; it looks like a difference in the expected value, so I'll avoid any kind of guessing and defer to you for possible fixes:

```txt
______________ TestENRTLcarbonicAcid.test_solution_carbonic_acid _______________

self = <proteuslib.tests.test_enrtl_water_pH.TestENRTLcarbonicAcid object at 0x7f93ccd28210>
carbonic_acid_model = <pyomo.core.base.PyomoModel.ConcreteModel object at 0x7f93dd99c6e0>

    @pytest.mark.component
    def test_solution_carbonic_acid(self, carbonic_acid_model):
        model = carbonic_acid_model
    
        pH = -log10(value(model.fs.unit.control_volume.properties_out[0.0].act_phase_comp["Liq","H_+"])*55.2)
        pOH = -log10(value(model.fs.unit.control_volume.properties_out[0.0].act_phase_comp["Liq","OH_-"])*55.2)
    
>       assert pytest.approx(8.27544, rel=1e-4) == pH
E       assert 8.27544 ± 8.3e-04 == 8.29618401972289
E        +  where 8.27544 ± 8.3e-04 = <function approx at 0x7f93fea974d0>(8.27544, rel=0.0001)
E        +    where <function approx at 0x7f93fea974d0> = pytest.approx

proteuslib/tests/test_enrtl_water_pH.py:854: AssertionError
```

aladshaw3:
@lbianchi-lbl In this particular case, I would just update the old value to the new value. The difference between a pH of 8.275 and 8.29618 is about 2.5E-10 in activity units (this is one reason why doing the detailed chemistry is so finicky). So basically, they are the same number for our purpose. 

Same can be said of pOH. Just make sure the sum of pOH and pH is near 14. 

The other values being check 'should' be close enough... but the tolerances on the asserts may be too tight in general for this type of problem.

proteuslib/tests/test_enrtl_water_pH.py
```diff
@@ -639,7 +636,7 @@ def test_solution_water(self, water_model):
                        #        we divide here by 1000**2 to perform both necessary
                        #        unit conversions in line.
                        "dh_rxn_ref": (0, pyunits.kJ/pyunits.mol),
-                       "k_eq_ref": (10**-14/1000/1000/55.2/55.2,pyunits.mol**2/pyunits.L**2),
+                       "k_eq_ref": (10**-14/1000/1000/55.2/55.2,pyunits.dimensionless),
                        "T_eq_ref": (298, pyunits.K),
```
aladshaw3, line 640: Remove all factors of 1000 in all k_eq_ref

aladshaw3 (commented):
Check over all k_eq_ref values. None should have any factors of 1000. 

Verify that pH + pOH =~ 14

pH should be around 8.3 for the tests with carbonic acid

proteuslib/tests/test_enrtl_water_pH.py
lbianchi-lbl: Whoops, that one got away, thanks for catching it.

lbianchi-lbl:
Thanks again @aladshaw3 for the suggestions. I set all the relative tolerances to `1e-2`, let me know if that's OK for this particular case or it's too loose.

Otherwise, we should probably go ahead and merge this since there are some other PRs that depend on it.

aladshaw3 (approved):
LGTM - those tolerances are fine for pH calculations.

bknueven (approved):

PR #224, opened 2021-09-17
Adding option to fill infeasible points with interpolated values

eyoung55:
## Fixes/Addresses:

This adds an optional argument to the parameter sweep which allows the user to save a version of the results in which "interior" points with an output value `nan` will be filled with an interpolated value derived from their neighbors. 

## Summary/Motivation:

When performing parameter sweeps it is sometimes the case that combinations of inputs will lead to an infeasible problem.  In some cases, it is possible to estimate a value for these points by using an interpolation that draws on nearby feasible solutions.

- Add function `_interp_nan_values` to parameter sweep
- Add unit test for `_interp_nan_values`

### Legal Acknowledgement


proteuslib/tools/parameter_sweep.py
```diff
@@ -17,6 +17,7 @@
 import itertools
 import warnings
 
+from scipy.interpolate import griddata
```
bknueven, line 20: Do we already pull in `scipy` as an IDAES dependency?

proteuslib/tools/parameter_sweep.py
```diff
@@ -438,6 +469,12 @@ def parameter_sweep(model, sweep_params, outputs, results_file=None, optimize_fu
     if rank == 0 and results_file is not None:
         # Save the global data
         np.savetxt(results_file, global_save_data, header=data_header, delimiter=',', fmt='%.6e')
+
+        if interpolate_nan_outputs:
+            global_results_clean = _interp_nan_values(global_values, global_results)
+            global_save_data_clean = np.hstack((global_values, global_results_clean))
+
+            np.savetxt('%s_clean' % (results_file), global_save_data_clean, header=data_header, delimiter=',', fmt='%.6e')
```
bknueven, line 477:
```suggestion
            np.savetxt('interpolated_%s' % (results_file), global_save_data_clean, header=data_header, delimiter=',', fmt='%.6e')
```
We should be specific with this name

proteuslib/tools/parameter_sweep.py
```diff
@@ -329,6 +354,12 @@ def parameter_sweep(model, sweep_params, outputs, results_file=None, optimize_fu
         debugging_data_dir (optional) : Save results on a per-process basis for parallel debugging
                                         purposes. If None no `debugging` data will be saved.
 
+        interpolate_nan_outputs (optional) : When the parameter sweep has finished, interior values
+                                             of np.nan will be replaced with a value obtained via
+                                             a linear interpolation of their surrounding valid neighbors.
+                                             If true, a second output file with the extension "_clean"
+                                             will be saved alongside the raw (un-interpolated) values. 
+
```
bknueven, line 362: For another PR: we should make parameter_sweep a class, and use a Pyomo config object to manage these ever-growing options. That said, I would like to wait until we are feature-stable to do this.

bknueven (commented):
Looks good. I would like the output file name changed, and I have a question about `scipy`. If we're not already pulling it in as part of IDAES we should handle the import differently. (Pyomo has utilities do delay imports of modules until they're needed.)

proteuslib/tools/parameter_sweep.py
eyoung55: yep, interpolated is definitely the clearer name here, good catch

jefalon (approved):
This is a very useful addition, good call formally adding it to the repo!

proteuslib/tools/parameter_sweep.py
eyoung55: yes, `scipy` is included as an IDAES dependency

proteuslib/tools/parameter_sweep.py
eyoung55: Agreed, the argument list is starting to feel a little cumbersome, so it's good to have this idea as part of the record.

bknueven (approved):

jefalon (approved):

PR #237, opened 2021-10-11
Gradient-based scaling for RO unit models

bknueven:
## Fixes/Addresses:
Makes various tweaks to both the RO unit models and NaCl Property Package to enable gradient-based scaling to be used with these models. The scaling changes we made such that Jacobians had reasonable condition numbers (`< 1e6`) when applying gradient-scaling.

## Summary/Motivation:
Ensure WaterTAP unit models are broadly compatible with the larger universe of IDAES models, including chemistry models.

1. Constraint scaling tweaks to 0D & 1D RO models
2. Constraint scaling tweaks to NaCl property package
3. Replace 0DRO `Var` `over_pressure_ratio` with `Expression`
4. Creating new standard ipopt options in unit models, their tests, and basic flowsheet examples for `bound_push = 1e-08`.
5. Changes to flowsheet unit tests due 1 & 4

## Open Questions / For Discussion:
1. Since WaterTAP models can have small bounds, should we be updating `idaes.cfg["ipopt"]["options"]["bound_push"]` when WaterTAP is imported? A value of `1e-08` is needed to initialize the 1DRO model, but a larger global value could be chosen. We can effectively disable IPOPT's bound's pushing by setting `bound_push` small enough, which ensures a model initialized through standard IDAES methods will not have its starting point modified by IPOPT.
2.  In particular, we run into issues with variables with small (`< 1e-6`) lower bounds and no or much larger upper bounds. The bounds push in IPOPT, [described in the paper on p. 20](http://cepac.cheme.cmu.edu/pasilectures/biegler/ipopt.pdf), will push small initial values on variable with a small LB no or large UB to ~`1e-2` with default settings, sometimes destroying a "good" initialization. In tests I observed that pushing a small variable up a single order of magnitude (e.g., from `1e-6` to `1e-5`)  would cause issues for the RO unit model.
3. For an interior point method (like IPOPT) it is important to have an interior starting point. Perhaps the correct thing to do is to write an initializer for WaterTAP / IDAES models which performs bounds pushing with more finesse than IPOPT's `bound_push`/`bound_frac` options, and then disable IPOPT's bounds pushing. 
4. We could avoid the issues in (2) by doing one or several of the following: 
    1. ensuring our models have variable with bounds that are within an order of magnitude
    1. are unbounded if bounds are not really necessary
    1. manually scaling some problematic variables in WaterTAP models to avoid small bounds
    1. or maintaining `user-scaling`.

### Legal Acknowledgement

### DIFF
--- a/watertap/unit_models/reverse_osmosis_0D.py
+++ b/watertap/unit_models/reverse_osmosis_0D.py
@@ -930,7 +930,7 @@ class ReverseOsmosisData(UnitModelBlockData):
         solve_log = idaeslog.getSolveLogger(blk.name, outlvl, tag="unit")
         # Set solver and options
         if optarg is None:
-            optarg = {'bound_push': 1e-6}
+            optarg = {'bound_push': 1e-8}
         opt = get_solver(solver, optarg)
 
         # assumptions
--- a/watertap/unit_models/reverse_osmosis_0D.py
+++ b/watertap/unit_models/reverse_osmosis_0D.py
@@ -1290,7 +1290,7 @@ class ReverseOsmosisData(UnitModelBlockData):
 
         for ind, c in self.eq_permeate_production.items():
             sf = iscale.get_scaling_factor(self.mass_transfer_phase_comp[ind])
-            iscale.constraint_scaling_transform(c, sf)
+            iscale.constraint_scaling_transform(c, sf/10.)
 
         for ind, c in self.eq_flux_io.items():
             sf = iscale.get_scaling_factor(self.flux_mass_io_phase_comp[ind])
@@ -1298,7 +1298,7 @@ class ReverseOsmosisData(UnitModelBlockData):
 
         for ind, c in self.eq_connect_mass_transfer.items():
             sf = iscale.get_scaling_factor(self.mass_transfer_phase_comp[ind])
-            iscale.constraint_scaling_transform(c, sf)
+            iscale.constraint_scaling_transform(c, sf/10.)
 
         for ind, c in self.eq_connect_enthalpy_transfer.items():
             sf = iscale.get_scaling_factor(self.feed_side.enthalpy_transfer[ind])
@@ -1356,7 +1356,7 @@ class ReverseOsmosisData(UnitModelBlockData):
         if hasattr(self, 'eq_N_Re_io'):
             for ind, c in self.eq_N_Re_io.items():
                 sf = iscale.get_scaling_factor(self.N_Re_io[ind])
-                iscale.constraint_scaling_transform(c, sf*1e3)
+                iscale.constraint_scaling_transform(c, sf*1e4)
 
         if hasattr(self, 'eq_N_Sc_io'):
             for ind, c in self.eq_N_Sc_io.items():
--- a/watertap/unit_models/reverse_osmosis_0D.py
+++ b/watertap/unit_models/reverse_osmosis_0D.py
@@ -376,13 +376,6 @@ class ReverseOsmosisData(UnitModelBlockData):
             units=pyunits.dimensionless,
             doc='Observed solute rejection')
 
-        #self.over_pressure_ratio = Var(
-        #    self.flowsheet().config.time,
-        #    initialize=1.1,
-        #    bounds=(0.5, 5),
-        #    units=pyunits.dimensionless,
-        #    doc='Over pressure ratio')
-
         if self.config.concentration_polarization_type == ConcentrationPolarizationType.fixed:
             self.cp_modulus = Var(
                 self.flowsheet().config.time,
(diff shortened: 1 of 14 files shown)


watertap/unit_models/reverse_osmosis_0D.py
```diff
@@ -889,12 +882,12 @@ def eq_rejection_phase_comp(b, t, j):
                     1 - (b.permeate_side.properties_mixed[t].conc_mass_phase_comp['Liq', j] /
                          b.feed_side.properties_in[t].conc_mass_phase_comp['Liq', j]))
 
-        @self.Constraint(self.flowsheet().config.time)
-        def eq_over_pressure_ratio(b, t):
-            return (b.feed_side.properties_out[t].pressure ==
-                    b.over_pressure_ratio[t]
-                    * (b.feed_side.properties_out[t].pressure_osm
-                    - b.permeate_side.properties_out[t].pressure_osm))
+        @self.Expression(self.flowsheet().config.time,
```
adam-a-a, line 885: I think it's better to have this as an expression but curious what triggered you to do this.

adam-a-a (commented):
Everything seems good to me, as long as these changes allow us to maintain/improve convergence while sweeping across a range of feed concentrations and recovery rates. Have you tried using parameter sweep to see how these changes affect convergence across said range? Also, many of the new scaling factors don't make immediate sense -- what was your approach for choosing these values? My guess is that you used some of the Jacobian scaling/probing tools to check extreme Jacobian entries and adjusted factors until these were resolved (but there is probably more to it that I'm unaware of).

bknueven:
> Everything seems good to me, as long as these changes allow us to maintain/improve convergence while sweeping across a range of feed concentrations and recovery rates. Have you tried using parameter sweep to see how these changes affect convergence across said range?

No, but that should probably be done before this is merged. I suspect there will be a benefit, but it's impossible to know without running.

> Also, many of the new scaling factors don't make immediate sense -- what was your approach for choosing these values? My guess is that you used some of the Jacobian scaling/probing tools to check extreme Jacobian entries and adjusted factors until these were resolved (but there is probably more to it that I'm unaware of).

This is exactly the approach, nothing more, nothing less. Usually checking (i) after initialization, (ii) after simulation, and (iii) after optimization. Primarily the goal was to reduce large entries and increase small entries both towards 1, and improve the condition number of the Jacobian.

watertap/unit_models/reverse_osmosis_0D.py
bknueven: I was having a hard time making this constraint well-scaled. I asked @TimBartholomew and he agreed with you that making this an Expression was a good idea.

adam-a-a:
@bknueven You probably already caught this, but it seems that the tests are failing (or at least the one that I looked at) due to the following assert in test_lssro:
`        assert model.statistics.number_of_objectives == 0
`

bknueven:
>Everything seems good to me, as long as these changes allow us to maintain/improve convergence while sweeping across a range of feed concentrations and recovery rates. Have you tried using parameter sweep to see how these changes affect convergence across said range?

I had to make some adjustments to the LSRRO flowsheet, but now it performs better with these changes. Sweeping across recovery and feed concentration for 1 through 5 stages, 1360 runs total:
| source: | main | main w/ new init | scale_0DRO |
| --- | --- | --- | --- |
| Number NaN: | 431 | 430 | 407 |

bknueven:
I think we should consider merging this as-is, and resolve some of the larger issues in a subsequent PR.

adam-a-a (approved):
LGTM

watertap/unit_models/reverse_osmosis_1D.py
```diff
@@ -1156,21 +1156,26 @@ def calculate_scaling_factors(self):
         super().calculate_scaling_factors()
         # setting scaling factors for variables
         for j in self.config.property_package.component_list:
-            iscale.set_scaling_factor(self.permeate_side[0, 0].flow_mass_phase_comp['Liq', j], 0)
+            iscale.set_scaling_factor(self.permeate_side[0, 0].flow_mass_phase_comp['Liq', j], 1e+5)
```
TimBartholomew, line 1159: This scaling factor used to be 0? That seems like that should have been a big issue, I'm surprised it worked before.

TimBartholomew (approved):
LGTM. The one scaling factor I pointed out should probably be set to some value based on the property function, but we can have this merged as in and address that later with the eventual planned refactoring.

watertap/unit_models/reverse_osmosis_1D.py
adam-a-a: I think this was because the inlet was skipped.

PR #244, opened 2021-10-22
EDB documentation and examples

dangunter:
## Fixes/Addresses:

Issue #140 

## Summary/Motivation:

The EDB isn't very well documented

- Add an example notebook
- Add some documentation

### Legal Acknowledgement


lbianchi-lbl:
@dangunter re: pandoc, we should check if it possible to install it using Conda instead of using an OS-specific method. I remember doing something like this for the IDAES examples repo (or, wishing I did it because I ran into OS compatibility issues).

lbianchi-lbl:
@dangunter @ksbeattie It looks like the RTD hook is working again!

And, I think the EDB examples look really nice! I also love how simple it is to just drop a notebook in the `docs` dir and have it built by RTD without fuss.

dangunter:
All tests finally pass! Please note that the example notebook is being run in a test in `watertap/tests/test_notebooks.py`, so at least in Linux where we have the MongoDB container in the Github Actions CI, tests passing means that notebook does not have any failures.

dangunter:
@lbianchi-lbl I have done a one-off test for mongodb presence/correctness in the notebook test (`watertap/tests/test_notebooks.py`), I think you may have something more elegant that would also fail if mongodb was loaded incorrectly (instead of just skipping the test in that case).

aladshaw3 (approved):
LGTM

dangunter:
@lbianchi-lbl can you do a quick review of this? I don't have red-button power to merge in this organization

watertap/tests/test_notebooks.py
```diff
@@ -0,0 +1,98 @@
...
+
+@pytest.fixture(scope="session")
+def docs_root(dirname="docs"):
+    p = Path(__file__).parent
+    while not (p / dirname).exists():
+        pp = p.parent
+        if pp == p:
+            raise RuntimeError(f"Could not find '{dirname}' dir")
+        p = pp
+    yield p / dirname
```
lbianchi-lbl, line 74: How is test_notebooks.py behaving if the test suite is being run in "package mode", i.e. when running `pip install watertap[testing] && pytest --pyargs watertap`, and the `docs` directory is not present? In that case `__file__` would point to the installation directory and this would fail (or I guess return an unrelated `docs` directory, if one is found during the search).

Eventually I'd like to introduce a common `docs_rootdir` fixture and/or a mark for tests that should only run in "repo clone mode", but for the scope of this PR, we could use `pytest.skip()` instead of raising an exception if the `docs` dir is not found. If `pytest.skip()` is called inside a fixture, any test that depend on it (directly or indirectly, i.e. through a derived fixture) will be skipped.

watertap/tests/test_notebooks.py
```diff
@@ -0,0 +1,98 @@
+"""
+Test Jupyter notebooks in the docs.
+"""
+# stdlib
+import logging
+from pathlib import Path
+import time
+
+# deps
+import nbformat
```
lbianchi-lbl, line 10: To avoid `ImportError`s at collect/import time, we could use [`pytest.importorskip`](https://docs.pytest.org/en/latest/reference/reference.html#pytest-importorskip-ref):

```py
nbformat = pytest.importorskip("nbformat")
```

watertap/tests/test_notebooks.py
```diff
@@ -0,0 +1,98 @@
...
+        _log.warning(f"Could not connect to MongoDB: {err}")
+    return False
+
+# Tests
+# =====
+
+
+@pytest.mark.skipif(not mongodb(), reason="MongoDB is required")
+@pytest.mark.component
+def test_edb_notebooks(docs_root):
```
lbianchi-lbl, line 91: This works, but the check is performed at import/collection time.

One alternative to perform the check at runtime instead is to turn this into a fixture:

```suggestion
@pytest.fixture(scope="session")
def mongodb_check():
    try:
        edb = ElectrolyteDB()  # will fail if no DB
        edb.get_base()  # will fail if DB is not loaded
    except Exception as err:
        pytest.skip(f"Could not connect to MongoDB: {err}")

# Tests
# =====


@pytest.mark.usefixtures('mongodb_check')
@pytest.mark.component
def test_edb_notebooks(docs_root):
```

.github/workflows/checks.yml
```diff
@@ -58,7 +58,7 @@ jobs:
     - name: Install dependencies
       run: |
         echo '::group::Output of "conda install" commands'
-        conda install --quiet --yes pip=21.1 wheel setuptools
+        conda install --quiet --yes pip=21.1 wheel setuptools pandoc
```
lbianchi-lbl, line 61: I think it is fine to add `pandoc` here for the time being, but eventually we should have a more explicit way to collect all non-`pip`-installable dev and/or optional dependencies (e.g. an `environment-dev.yml` file for Conda that internally calls `requirements-dev.txt` in the `pip` section).

watertap/tests/test_notebooks.py
```diff
@@ -0,0 +1,98 @@
...
+    print("\n")
+    for nb_path in find_notebooks(docs_root / "examples" / "edb"):
+        if ".ipynb_checkpoints" in nb_path.parts:
+            continue
+        print(f"run notebook: {nb_path}")
+        run_notebook(nb_path)
```
lbianchi-lbl, line 97: I think it would be nice to "lift up" this loop as a parametrized `edb_notebook_path` parametrized fixture. I think it might be not worth at the moment since we don't have that many notebooks to run yet, but something to keep in mind for the future.

lbianchi-lbl (dismissed):
In general everything looks good.

- My main question is what happens when the tests are run in "package mode", i.e. when the `docs` directory (and possibly the extra `nbXXX` packages) are not available. We should make sure that this case is taken into account and dealt with gracefully to avoid unexpected failures
- There were a few ideas for future "nice-to-have"s that popped up while reviewing, but I just added them for future reference and don't think they should be addressed now.

adam-a-a (commented):
The EDB_example notebook looks good, but I think we might want to put it in either the tutorials section or the how-to section, rather than having a separate examples directory (just considering our efforts to conform to the Diataxis framework).

dangunter:
in the docs it's linked out of a how-to page. I think examples is a clearer directory name.

watertap/tests/test_notebooks.py
dangunter: Yes, seems overkill for the 1 notebook, but agree in principle. Need to leave something for future work..

.github/workflows/checks.yml
dangunter: If that's the "right" way to do it, then sure. Should go with some README developer docs since this is pretty obscure.

watertap/tests/test_notebooks.py
dangunter: If I don't perform this at the import time, then I can't tell whether the test was skipped or not. I prefer it this way, in the case where a real DB is needed. We can assure it runs once if necessary.

watertap/tests/test_notebooks.py
dangunter: Yeah skip is a good idea, though that's a bit trickier than desired. Is there just a flag we can use to tell if the package is an archive file instead of a real directory tree?

watertap/tests/test_notebooks.py
lbianchi-lbl: > If I don't perform this at the import time, then I can't tell whether the test was skipped or not

Wait, I'm not sure I understand. What do you mean with this? I think in general performing the DB connection at runtime (i.e. using `pytest.skip()` from within the fixture function body) rather than at collect time is preferable, but maybe I'm missing some drawbacks of this approach.

dangunter: sorry I didn't actually read your code. No, that would work, essentially doing the skip test later.

watertap/tests/test_notebooks.py
lbianchi-lbl: Sorry for the confusion, what I meant with "package mode" is "non-editable installation mode", i.e. where the Python package has its files stored under `.../python3.X/site-packages/watertap/...` (where `docs` is not present) rather than a local clone of the repo (where `docs` is present).

In this case probably checking for `if "site-packages" in Path(__file__).parts` should be enough, although, as I said, this is not specific to the notebook tests and should be implemented as a globally-available utility skipmark/fixture.

aladshaw3 (approved):
LGTM

adam-a-a (approved):
LGTM

dangunter:
@lbianchi-lbl can you approve, it won't let me merge until "requested changes are addressed" and afaict that means you need to give the thumbs-up

PR #245, opened 2021-10-22
Membrane base

bknueven:
## Fixes/Addresses:
N/A

## Summary/Motivation:
Initial pass at membrane base class. Further integration is possible, but would necessitate breaking changes to one or both of the 0D and 1D RO unit models.

- Combine shared code between the 0D and 1D RO unit models into a common base class

### Legal Acknowledgement

### DIFF
--- a/watertap/unit_models/_reverse_osmosis_base.py
+++ b/watertap/unit_models/_reverse_osmosis_base.py
@@ -46,7 +46,7 @@ class PressureChangeType(Enum):
 
 class _ReverseOsmosisBaseData(UnitModelBlockData):
     """
-    Reverse-Osmosis filtration base class
+    Reverse Osmosis base class
     """
 
     CONFIG = ConfigBlock()
--- a/watertap/unit_models/_reverse_osmosis_base.py
+++ b/watertap/unit_models/_reverse_osmosis_base.py
@@ -18,6 +18,7 @@ from idaes.core import UnitModelBlockData, useDefault, MaterialBalanceType,\
         EnergyBalanceType, MomentumBalanceType
 from idaes.core.util import scaling as iscale
 from idaes.core.util.config import is_physical_parameter_block
+from idaes.core.util.exceptions import ConfigurationError
 from idaes.core.util.tables import create_stream_table_dataframe
 import idaes.logger as idaeslog
 
@@ -43,9 +44,9 @@ class PressureChangeType(Enum):
     calculated = auto()              # pressure drop across membrane channel is calculated
 
 
-class _MembraneBaseData(UnitModelBlockData):
+class _ReverseOsmosisBaseData(UnitModelBlockData):
     """
-    Membrane-based filtration base class
+    Reverse-Osmosis filtration base class
     """
 
     CONFIG = ConfigBlock()
@@ -129,13 +130,13 @@ class _MembraneBaseData(UnitModelBlockData):
     **MomentumBalanceType.momentumPhase** - momentum balances for each phase.}"""))
 
     CONFIG.declare("concentration_polarization_type", ConfigValue(
-        default=ConcentrationPolarizationType.none,
+        default=ConcentrationPolarizationType.calculated,
         domain=In(ConcentrationPolarizationType),
         description="External concentration polarization effect in RO",
         doc="""
             Options to account for concentration polarization.
 
-            **default** - ``ConcentrationPolarizationType.none``
+            **default** - ``ConcentrationPolarizationType.calculated``
 
         .. csv-table::
             :header: "Configuration Options", "Description"
@@ -146,13 +147,13 @@ class _MembraneBaseData(UnitModelBlockData):
         """))
 
     CONFIG.declare("mass_transfer_coefficient", ConfigValue(
-        default=MassTransferCoefficient.none,
+        default=MassTransferCoefficient.calculated,
         domain=In(MassTransferCoefficient),
         description="Mass transfer coefficient in RO feed channel",
         doc="""
             Options to account for mass transfer coefficient.
 
-            **default** - ``MassTransferCoefficient.none``
+            **default** - ``MassTransferCoefficient.calculated``
 
         .. csv-table::
             :header: "Configuration Options", "Description"
--- a/watertap/unit_models/membrane_base.py
+++ b/watertap/unit_models/membrane_base.py
@@ -1,0 +1,140 @@
+###############################################################################
+# WaterTAP Copyright (c) 2021, The Regents of the University of California,
+# through Lawrence Berkeley National Laboratory, Oak Ridge National
+# Laboratory, National Renewable Energy Laboratory, and National Energy
+# Technology Laboratory (subject to receipt of any required approvals from
+# the U.S. Dept. of Energy). All rights reserved.
+#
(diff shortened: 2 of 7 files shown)


TimBartholomew (dismissed):
LGTM. It's a good start. We should have a meeting to discuss how we can get more of the variables and constraints into the membrane base. And ensure this is lined up well with the other thing we wanted from the refactoring, which is to make it easy to make the counterflow membrane unit models.

watertap/unit_models/membrane_base.py
```diff
@@ -0,0 +1,393 @@
...
+    **default** - MomentumBalanceType.pressureTotal.
+    **Valid values:** {
+    **MomentumBalanceType.none** - exclude momentum balances,
+    **MomentumBalanceType.pressureTotal** - single pressure balance for material,
+    **MomentumBalanceType.pressurePhase** - pressure balances for each phase,
+    **MomentumBalanceType.momentumTotal** - single momentum balance for material,
+    **MomentumBalanceType.momentumPhase** - momentum balances for each phase.}"""))
+
+    CONFIG.declare("concentration_polarization_type", ConfigValue(
+        default=ConcentrationPolarizationType.none,
```
adam-a-a: I think we can set this to `ConcentrationPolarizationType.calculated` which should resolve the issue you bring up later in the 1DRO config. The reason the 1D version has this default already is because, logically, if you are going to go so far as to model RO as 1D, you might as well be calculating concentration polarization effects too. I think it would be good to have this as the default for 0D as well; if the user wanted to show comparisons with neglecting concentration polarization or assuming a fixed concentration polarization modulus, he/she could opt to do so.

watertap/unit_models/membrane_base.py
```diff
@@ -0,0 +1,393 @@
...
+        .. csv-table::
+            :header: "Configuration Options", "Description"
+
+            "``ConcentrationPolarizationType.none``", "Simplifying assumption to ignore concentration polarization"
+            "``ConcentrationPolarizationType.fixed``", "Specify an estimated value for the concentration polarization modulus"
+            "``ConcentrationPolarizationType.calculated``", "Allow model to perform calculation of membrane-interface concentration"
+        """))
+
+    CONFIG.declare("mass_transfer_coefficient", ConfigValue(
+        default=MassTransferCoefficient.none,
```
adam-a-a: Similarly, we can change this to `MassTransferCoefficient.calculated` if we are going to use the calculated option for concentration polarization.

watertap/unit_models/reverse_osmosis_1D.py
```diff
@@ -337,12 +183,11 @@ def build(self):
...
         super().build()
 
-        self.scaling_factor = Suffix(direction=Suffix.EXPORT)
-
+        # Check configuration errors
         self._process_config()
 
         # Build 1D Control volume for feed side
-        self.feed_side = ControlVolume1DBlock(default={
+        self.feed_side = feed_side = ControlVolume1DBlock(default={
```
adam-a-a, line 146: I could be wrong, but I thought I heard that this would not be supported anymore (i.e., x = y = z).

watertap/unit_models/membrane_base.py
```diff
@@ -0,0 +1,393 @@
...
+"""))
+
+    def build(self):
+        """
+        Common variables and constraints for an RO unit model
+        """
+        super().build()
+
+        if len(self.config.property_package.solvent_set) > 1:
+            raise ConfigurationError("Membrane models only supports one solvent component,"
```
adam-a-a:
```suggestion
            raise ConfigurationError("Membrane models only support one solvent component,"
```

watertap/unit_models/membrane_base.py
```diff
@@ -0,0 +1,393 @@
...
+                                     "the provided property package has specified {} solvent components"
+                                     .format(len(self.config.property_package.solvent_set)))
+
+        if (len(self.config.property_package.phase_list) > 1
+                or 'Liq' not in [p for p in self.config.property_package.phase_list]):
+            raise ConfigurationError(
+                "Membrane models only supports one liquid phase ['Liq'],"
```
adam-a-a:
```suggestion
                "Membrane models only support one liquid phase ['Liq'],"
```

adam-a-a (changes requested):
Looks good, but just a couple of small suggested changes. 

Also, I should mention that I've been working on a modified 0DRO model that would be compatible with the generic property framework (and the eNRTL model). This includes setting the set of solutes to either be solute_set or ion_set, as well as some changes in property names and material flow basis. I haven't pushed yet since I cannot get it to solve yet with the generic seawater ion package. With that said, we _might_ want to hold off on merging this until those changes are in. _Or_ we can merge this and then I'd incorporate my changes with what you have here.

watertap/unit_models/reverse_osmosis_1D.py
andrewlee94: Not sure where you heard that - can you recall the context?

bknueven: This is a feature of the Python language, so I doubt it's going away before Python 4...
```python
>>> x = y = 42
>>> x, y
(42, 42)
```

I couldn't find a more authoritative source, but at least according to [this StackExchange post](https://stackoverflow.com/a/51147431) this change does exactly what was done before, e.g., sets `self.feed_side` to the result of the `ControlVolume1DBlock` constructor and sets `feed_side` to `self.feed_side`.

adam-a-a: OK, then I probably confused this with something else. I don't remember exactly, but I thought someone mentioned this on a dev call, referring to things that Pyomo would no longer allow. Thought I'd double-check anyway!

bknueven:
@adam-a-a I don't have strong feelings about whether this should be merged. It's probably worth it for the common configuration (especially with your suggested changes), but the subtle differences between the 0D and 1D variables and constraints (esp. as it relates to in/out), make it difficult to combine more without breaking things (i.e., renaming variables/constraints).

Perhaps I should explore other possible re-organizations of our unit models before we go forward with this particular iteration.

watertap/unit_models/reverse_osmosis_1D.py
andrewlee94: That is a separate issue - Pyomo does not support a `Component` being assigned to two different `Blocks/Models`, but it has no problems with aliases.

bknueven:
The test failures are unrelated to this PR; see #246.

watertap/unit_models/_reverse_osmosis_base.py
```diff
@@ -0,0 +1,394 @@
...
+
+class PressureChangeType(Enum):
+    fixed_per_stage = auto()         # pressure drop across membrane channel is a user-specified value
+    fixed_per_unit_length = auto()   # pressure drop per unit length across membrane channel is a user-specified value
+    calculated = auto()              # pressure drop across membrane channel is calculated
+
+
+class _ReverseOsmosisBaseData(UnitModelBlockData):
+    """
+    Reverse-Osmosis filtration base class
```
adam-a-a:
```suggestion
    Reverse Osmosis base class
```

adam-a-a (approved):
LGTM

TimBartholomew (approved):
LGTM

PR #248, opened 2021-10-28
Add pytest fixtures for EDB and parametrization

lbianchi-lbl:
### Adding an `edb` fixture

- Several of our tests are semantically parametrizable in terms of "run the same tests with or without EDB"
- The current strategy is to instantiate the test data by setting class attributes directly on pytest `Test...` classes
- This in turn requires, for the case where the EDB is used, to run the EDB instantiation logic at import time
- Running code at import time should be avoided if possible, especially if it involves performing highly non-local actions like connecting to an external DB instance, and especially if it's done within pytest (since pytest will import modules in way that's not necessarily what one might expect, e.g. re: the order in which modules are imported)
- Using fixtures (that are managed and dispatched by pytest at runtime) instead of setting class attributes at import time on a pytest test class (something that pytest is not "expecting") seems to me like a better choice, (again, especially when an external database is involved)
- The `edb` fixture implemented in this PR not only eliminates the need for import-time connection attempts, but also (by default) transparently falls back to using a mocked MongoDB client (courtesy of `mongomock`) if the connection fails
- This means that tests can be written without having to care whether a real MongoDB instance is available or not, and therefore don't need to be skipped if that is not the case
- Additionally, the `edb` fixture also takes care of data loading/bootstrap and cleanup, reducing the chances of hidden dependencies between tests

### Sample of a test module refactored using more parametrization

- Using fixtures to implement the EDB logic helps paving the way for streamlining and extending test parametrization
- As an example ("test test-case"?) , I refactored the `pure_water_pH` test module using the `edb` fixture
- Additionally, I added another fixture, `variant`, that is specific to that module and parametrized with the possible values `equilibrium` and `inherent`, to help "factor out" the common functionality that's currently distributed within the `XXX_equilibrium_YYY` and `XXX_inherent_YYY` test functions and fixtures
- This is **just a proposal**:
  - If we find that the test module benefits from the refactoring, we can extend the same general logic to other test modules as well
  - Otherwise, we can select which parts of this PR to include and roll back the rest of the changes

### Example of pytest output (verbose) with parametrization

```txt
============================= test session starts =============================
platform win32 -- Python 3.9.7, pytest-6.2.5, py-1.10.0, pluggy-1.0.0 -- c:\miniconda\envs\watertap-dev\python.exe
cachedir: .pytest_cache
rootdir: D:\a\watertap\watertap, configfile: pytest.ini, testpaths: watertap
plugins: cov-3.0.0
collecting ... collected 456 items / 420 deselected / 36 selected

watertap/tests/test_pure_water_pH.py::TestPureWater::test_build_model[Variant.equilibrium] PASSED [  2%]
watertap/tests/test_pure_water_pH.py::TestPureWater::test_units_inherent[Variant.equilibrium] PASSED [  5%]
watertap/tests/test_pure_water_pH.py::TestPureWater::test_dof[Variant.equilibrium] PASSED [  8%]
watertap/tests/test_pure_water_pH.py::TestPureWater::test_stats[Variant.equilibrium] PASSED [ 11%]
watertap/tests/test_pure_water_pH.py::TestPureWater::test_scaling[Variant.equilibrium] PASSED [ 13%]
watertap/tests/test_pure_water_pH.py::TestPureWater::test_equilibrium_extra_scale_factors[Variant.equilibrium] PASSED [ 16%]
watertap/tests/test_pure_water_pH.py::TestPureWater::test_initialize[Variant.equilibrium] PASSED [ 19%]
watertap/tests/test_pure_water_pH.py::TestPureWater::test_solve[Variant.equilibrium] PASSED [ 22%]
watertap/tests/test_pure_water_pH.py::TestPureWater::test_solution[Variant.equilibrium] PASSED [ 25%]
watertap/tests/test_pure_water_pH.py::TestPureWaterEDB::test_build_model[Variant.equilibrium] PASSED [ 27%]
watertap/tests/test_pure_water_pH.py::TestPureWaterEDB::test_units_inherent[Variant.equilibrium] PASSED [ 30%]
watertap/tests/test_pure_water_pH.py::TestPureWaterEDB::test_dof[Variant.equilibrium] PASSED [ 33%]
watertap/tests/test_pure_water_pH.py::TestPureWaterEDB::test_stats[Variant.equilibrium] PASSED [ 36%]
watertap/tests/test_pure_water_pH.py::TestPureWaterEDB::test_scaling[Variant.equilibrium] PASSED [ 38%]
watertap/tests/test_pure_water_pH.py::TestPureWaterEDB::test_equilibrium_extra_scale_factors[Variant.equilibrium] PASSED [ 41%]
watertap/tests/test_pure_water_pH.py::TestPureWaterEDB::test_initialize[Variant.equilibrium] PASSED [ 44%]
watertap/tests/test_pure_water_pH.py::TestPureWaterEDB::test_solve[Variant.equilibrium] PASSED [ 47%]
watertap/tests/test_pure_water_pH.py::TestPureWaterEDB::test_solution[Variant.equilibrium] PASSED [ 50%]
watertap/tests/test_pure_water_pH.py::TestPureWater::test_build_model[Variant.inherent] PASSED [ 52%]
watertap/tests/test_pure_water_pH.py::TestPureWater::test_units_inherent[Variant.inherent] PASSED [ 55%]
watertap/tests/test_pure_water_pH.py::TestPureWater::test_dof[Variant.inherent] PASSED [ 58%]
watertap/tests/test_pure_water_pH.py::TestPureWater::test_stats[Variant.inherent] PASSED [ 61%]
watertap/tests/test_pure_water_pH.py::TestPureWater::test_scaling[Variant.inherent] PASSED [ 63%]
watertap/tests/test_pure_water_pH.py::TestPureWater::test_equilibrium_extra_scale_factors[Variant.inherent] SKIPPED [ 66%]
watertap/tests/test_pure_water_pH.py::TestPureWater::test_initialize[Variant.inherent] PASSED [ 69%]
watertap/tests/test_pure_water_pH.py::TestPureWater::test_solve[Variant.inherent] PASSED [ 72%]
watertap/tests/test_pure_water_pH.py::TestPureWater::test_solution[Variant.inherent] PASSED [ 75%]
watertap/tests/test_pure_water_pH.py::TestPureWaterEDB::test_build_model[Variant.inherent] PASSED [ 77%]
watertap/tests/test_pure_water_pH.py::TestPureWaterEDB::test_units_inherent[Variant.inherent] PASSED [ 80%]
watertap/tests/test_pure_water_pH.py::TestPureWaterEDB::test_dof[Variant.inherent] PASSED [ 83%]
watertap/tests/test_pure_water_pH.py::TestPureWaterEDB::test_stats[Variant.inherent] PASSED [ 86%]
watertap/tests/test_pure_water_pH.py::TestPureWaterEDB::test_scaling[Variant.inherent] PASSED [ 88%]
watertap/tests/test_pure_water_pH.py::TestPureWaterEDB::test_equilibrium_extra_scale_factors[Variant.inherent] SKIPPED [ 91%]
watertap/tests/test_pure_water_pH.py::TestPureWaterEDB::test_initialize[Variant.inherent] PASSED [ 94%]
watertap/tests/test_pure_water_pH.py::TestPureWaterEDB::test_solve[Variant.inherent] PASSED [ 97%]
watertap/tests/test_pure_water_pH.py::TestPureWaterEDB::test_solution[Variant.inherent] PASSED [100%]
```

- Add globally available `edb` fixture
- Parametrize `test_pure_water_pH` based on `variant` (inherent or equilibrium)
- Use subclass with overridden fixtures instead of a separate module when using EDB as the data source
- Add an option (as a pytest CLI flag) to cause EDB tests to fail (instead of using a `mongomock`ed EDB instance) if the DB connection fails, which helps when we do want to test the behavior with a real DB

## For reviewers

- [ ] Implementation of the EDB fixture:
  - [ ] Any bugs or shortcomings of the current functionality or implementation
  - [ ] Would having an `edb` fixture available for the entire test suite be useful?
  - [ ] Any contraindication to using "transparent mocking" for when a running MongoDB instance is not available?
- [ ] Parametrization of a sample test module
  - [ ] In general: would this type of refactoring be valuable?
    - [ ] Is the test code easier to read?
    - [ ] Is it easier to write test code?
    - [ ] Is it easier to navigate the test output? (see above)
  - [ ] In particular, for the sample test module: **is the test still correct?** I did the refactoring using code-based (and domain-ignorant) "pattern matching", so please check for any blunders or questionable changes I introduced

### Legal Acknowledgement


dangunter:
Overall looks good, but one thing is missing. We need a function and a fixture, so we can also do `pytest.skipif` logic when we need an actual mongodb server (i.e. the mock won't be sufficient). I think that's just the front-end of the connection logic.

By the way, I'm fiddling with the connection logic [here](https://github.com/dangunter/proteuslib/blob/edb_cert/watertap/edb/db_api.py) because there needs to be some extra parameters at least on my Windows box to get the connection to a remote AWS-hosted MongoDB server to work.

lbianchi-lbl:
> Overall looks good, but one thing is missing. We need a function and a fixture, so we can also do `pytest.skipif` logic when we need an actual mongodb server (i.e. the mock won't be sufficient). I think that's just the front-end of the connection logic.

This makes sense. I can think of a couple of ways to deal with this case:

- In this PR, something along these lines already exists with the `--edb-no-mock` CLI flag/option: by default, mocks are allowed in case the client connection fails. If the `--edb-no-mock` is used, then when the EDB client connection test fails, the test fails as well
- If a "stricter" `edb` instance is needed for a particular test function or module, it's straightforward to override the `edb` fixture which will be used within that testing scope instead of the general (mock-enabled) one

> By the way, I'm fiddling with the connection logic [here](https://github.com/dangunter/proteuslib/blob/edb_cert/watertap/edb/db_api.py) because there needs to be some extra parameters at least on my Windows box to get the connection to a remote AWS-hosted MongoDB server to work.

I don't know much about the EDB internals, but would it make sense to have a way to customize these options when an EDB instance is created (e.g. `**kwargs` that are adapted and then passed to the `MongoClient`, or a way to plug in another `MongoClient` instance)? If so, this could be a way for setting these options for the global `edb` fixture, i.e. for tests that don't need to care what DB backend is being used.

watertap/tests/test_pure_water_pH.py
```diff
@@ -94,274 +96,298 @@
 
 __author__ = "Austin Ladshaw"
```
aladshaw3: You should probably add your name here since so much of this file has changed since I worked on it.

Also, maybe Dan's name should be here to since this file integrates EDB tests with the chemistry test.

watertap/tests/test_pure_water_pH.py
```diff
@@ -569,15 +540,21 @@ def test_scaling_equilibrium(self, equilibrium_reactions_config):
...
         )
 
-        # When using equilibrium reactions, there are another set of scaling factors calculated
+    def test_equilibrium_extra_scale_factors(self, model_scaling, variant: Variant):
+        """
+        When using equilibrium reactions, there are another set of scaling factors calculated
+        """
+        if not variant.is_equilibrium:
+            pytest.skip("Only applicable for equilibrium reaction")
+
```
aladshaw3: Are we ok with doing this in this way? This will show up in the pytest run as skipped tests (and will always show up as skipped tests).

watertap/tests/test_pure_water_pH.py
```diff
@@ -598,67 +582,35 @@ def test_initialize_inherent(self, inherent_reactions_config):
...
 
     @pytest.mark.component
-    def test_solution_inherent(self, inherent_reactions_config):
-        model = inherent_reactions_config
+    def test_solution(self, model_solve):
+        model, _ = model_solve
 
-        assert pytest.approx(298, rel=1e-5) == value(
+        assert pytest.approx(298, rel=1e-4) == value(
             model.fs.unit.outlet.temperature[0]
```
aladshaw3: Did this tolerance need to be changed? I noticed you did not update any of the other tolerances.

aladshaw3 (changes requested):
It looks basically ready to go. I just had a few minor comments.

watertap/tests/test_pure_water_pH.py
dangunter: what do people think of doing authors like this:

    __authors__ = ["Author 1", "Author 2", etc.]
    __author__ = __authors__[0]

I feel like the single author is too restrictive, but still reasonable to expect as a first-author.

watertap/tests/test_pure_water_pH.py
dangunter: For me the semantics of skipping tests should be that some external condition is not present, but if the conditions are all controlled by the test code itself a simple if/then block would suffice.

lbianchi-lbl: I guess I have a more "liberal" view of `pytest.skip()`, as in it can be used to express the concept of "this test function does not apply to this context", where the context includes both external conditions (`if sys.platform == "win32": pytest.skip("This test cannot run on Windows")`) and internal data, i.e. fixture values (such as in this case). 

Consequently, in this view, having skipped tests in the output is normal: in some cases, factoring tests for ease of parametrization might result in test-parameter value combinations that don't make sense domain-wise (such as here), and using `pytest.skip()` instead of a "silent" `if` can be a way of making this more explicit in the output.

However, of course, there are many cases where the line is blurred and both options make sense implementation-wise, in which case I think we should go with the one that makes the most sense domain-wise. In this case, if the assertion(s) to perform for equilibrium reactions is more of an additional check than a separate test altogether, it makes sense to have a single test function and use a simple `if` statement to bypass those `assert`s for the inherent reaction case. **EDIT**: corrected equilibrium/inherent

watertap/tests/test_pure_water_pH.py
lbianchi-lbl: Thanks for spotting that, I think you're right. Let me try to change it back to `1e-5` and see if it does make a difference.

watertap/tests/test_pure_water_pH.py
lbianchi-lbl: My overengineering shoulder devil is whispering:

```py
__authors__ = {
    "Austin Ladshaw": "Concept and original implementation",
    "Ludovico Bianchi": "pytest parametrization and refactoring",
    "Dan Gunter": "EDB functionality"
}
__author__ = list(__authors__)[0]
```

, but I'm not giving in to the temptation and go with what Dan is suggesting.

aladshaw3: It is kind of nice to see individual contributions credited in the author list, but maybe it is overkill.

dangunter: but what about the simpler list version? I feel like we should either list the actual (multiple) authors or skip the __author__ thing altogether (borg-style).

lbianchi-lbl: No, I like the `__authors__` idea, I guess more as a more human-curated `git blame` than for keeping track of individual contributions. I've added it following Dan's proposed format.

aladshaw3 (approved):
LGTM

aladshaw3:
@lbianchi-lbl @dangunter 

Looks like the update to the EDB api has caused the current failures in this PR. I think it is an easy fix. The 'get_one_base' function in EDB object no longer exists. I believe it was replaced with 'get_base' function.

watertap/tests/test_pure_water_pH.py
```diff
@@ -685,33 +638,47 @@ def test_solution_inherent(self, inherent_reactions_config):
...
-        assert pytest.approx(6.9997414, rel=1e-5) == pOH
-        assert pytest.approx(0.99999, rel=1e-5) == value(
-            model.fs.unit.outlet.mole_frac_comp[0.0, "H2O"]
-        )
+class TestPureWaterEDB(TestPureWater):
+
+    @pytest.fixture
+    def thermo_config(self, edb):
+        base = edb.get_one_base("water_reaction")
+        elements = ["H", "O"]
```
aladshaw3: Change 'get_one_base' to 'get_base'

watertap/tests/test_pure_water_pH.py
```diff
@@ -685,33 +638,47 @@ def test_solution_inherent(self, inherent_reactions_config):
...
+            base.add(r)
+        return base.idaes_config
+
+
+    @pytest.fixture
+    def water_reaction_config(self, edb):
+        elements = ["H", "O"]
+        components = [c.name for c in edb.get_components(element_names=elements)]
+        base = edb.get_one_base("water_reaction")
+        # Need to remove these to avoid errors when using the generated config
```
aladshaw3: Change 'get_one_base' to 'get_base'

aladshaw3 (changes requested):
Update to new EDB API calls.

dangunter:
weird, I just put an alias get_one_base -> get_base that prevented that, maybe it didn't get pushed

aladshaw3 (approved):
LGTM

ksbeattie (approved):
LGTM

PR #251, opened 2021-11-07
Reorganize repo

TimBartholomew:
## Fixes/Addresses:
None

## Summary/Motivation:
The repository should be reorganized to integrate WaterTAP3 content.

- Reorganized repository

### Legal Acknowledgement


docs/apidoc/watertap.rst
```diff
@@ -8,7 +8,7 @@ Subpackages
    :maxdepth: 4
 
    watertap.edb
-   watertap.flowsheets
+   watertap.examples.flowsheets
```
andrewlee94, line 11: Given that we are expecting more things under examples, would it be better to have a new `examples.rst` file here which for now is just a TOC pointing onwards to `flowsheets`?

TimBartholomew:
Hey guys, I'm not sure why it is saying I deleted all the lines in the chemistry tests and then added all the same lines back. Do we think that is ok? Or should I attempt to revert just those changes and try again?

andrewlee94 (approved):
Only two things I noted going through, one which is captured as a comment. The other was the observation that the `edb` folder hasn't been moved yet, but I recommend that be a separate PR (if that wasn't already the intention).

andrewlee94:
@TimBartholomew As long as it works, stick with what you have.

watertap/examples/chemistry/tests/test_tutorials.py
```diff
@@ -22,7 +22,7 @@
 import os.path
 from nbconvert.preprocessors import ExecutePreprocessor
 
-_tutorials_dir = os.path.join(os.path.dirname(os.path.abspath(__file__)),'..','..','tutorials')
+_tutorials_dir = os.path.join(os.path.dirname(os.path.abspath(__file__)), '../../..', '..', 'tutorials')
```
bknueven, line 25:
```suggestion
_tutorials_dir = os.path.join(os.path.dirname(os.path.abspath(__file__)), '..','..','..', '..', 'tutorials')
```
That said, this will work on modern Windows systems as-is.

adam-a-a (approved):
LGTM - Any necessary adjustments to the repo tree can be made in subsequent PRs.

bknueven (approved):
One minor comment. Otherwise looks good.

bknueven:
> Hey guys, I'm not sure why it is saying I deleted all the lines in the chemistry tests and then added all the same lines back. Do we think that is ok? Or should I attempt to revert just those changes and try again?

`git` does this sometimes. I wouldn't worry about it.

docs/apidoc/watertap.rst
TimBartholomew: I'm honestly not familiar with the apidoc. I can't find it on the webpage, but yes the TOC should be for examples. Since I don't know how to view the change right now, I'll leave it for this PR.

watertap/examples/chemistry/tests/test_tutorials.py
TimBartholomew: Let's open another PR for this, just so we don't need everyone to approve and go through another round of testing.

PR #254, opened 2021-11-10
Adding WaterTAP3 data

bknueven:
...with a first pass at organization into techno-economic data and case studies.

## Summary/Motivation:
WaterTAP3 Merger

- Bring over a bunch of csv files from WaterTAP3

### Legal Acknowledgement


kurbansitterley (commented):
I will leave it to someone else to make a decision, so this is more just providing context on some of these files.

These files are needed to build a model in WT3: 

- case_study_basis
- case_study_water_sources
- constituent_removal
- electricity_costs
- plant_cost_indices
- treatment_train_setup
- water_recovery

I don't think these files _need_ to be brought over to watertap:

- an_ex_cost_eqns
- basic_unit
- basic_units_solaire
- gac_cost_eqns
- gac_cost_curves
- cost_curves
- chlorine_dost_cost
- uv_cost_interp
- baseline_cases_runs
- excel_to_python_names

All of these besides the last two contain data used to create cost curves for a few units in WT3. There is not real problem with bringing them over, but it might clutter the folder and be confusing. There is good costing data in here but there is probably a better approach to using it in the model than the way we had it implemented in WT3.

adam-a-a:
@bknueven maybe we can start by only including the necessary files that @kurbansitterley pointed out and figure out what to do with the rest later on. However, if you feel we should just merge now and weed things out later, happy to approve once you confirm.

bknueven:
Given the complexity involved, I'm happy to leave things out we don't think are necessary. They can be brought over on an as-needed basis.

bknueven:
With those changes, most of the WaterTAP3 files are now in `examples/case_studies`. The only files left in `data/techno_economic` have to do with chemistry. To my untrained eye `data/techno_economic/chemical_addition.csv` perhaps has some overlap with the EDB, but that could be sorted later.

adam-a-a:
> With those changes, most of the WaterTAP3 files are now in `examples/case_studies`. The only files left in `data/techno_economic` have to do with chemistry. To my untrained eye `data/techno_economic/chemical_addition.csv` perhaps has some overlap with the EDB, but that could be sorted later.

@bknueven I think `data/techno_economic/chemical_addition.csv` provides coefficients for associated cost equations, while EDB does not include anything related to costs so should be OK as is.

adam-a-a (approved):
LGTM

kurbansitterley:
> > With those changes, most of the WaterTAP3 files are now in `examples/case_studies`. The only files left in `data/techno_economic` have to do with chemistry. To my untrained eye `data/techno_economic/chemical_addition.csv` perhaps has some overlap with the EDB, but that could be sorted later.
> 
> @bknueven I think `data/techno_economic/chemical_addition.csv` provides coefficients for associated cost equations, while EDB does not include anything related to costs so should be OK as is.

Yes you are correct, `chemical_addition.csv` contains coefficients for cost equations so could be left out also. `catalysts_chemicals.csv` however is $/kg of chemical so is probably ok here (but should be updated at some point).

kurbansitterley (approved):
`chemical_addition.csv` could probably be excluded also but could also be sorted out later.

PR #260, opened 2021-11-12
Changes for connecting to cloud mongodb server

dangunter:
## Fixes/Addresses:

You can connect to cloud instances of MongoDB by just changing the URL, but one problem that can occur is a failure to verify certificates (over SSL), so this adds some code to retry using the `certifi.where()` module, if the certifi package is installed.

In the process of testing and documentation, some Jupyter Notebooks were added to the repo. These are kept in sync by running them automatically.

### Legal Acknowledgement


watertap/edb/db_api.py
```diff
@@ -265,6 +323,9 @@ def get_base(self, name: str = None) -> Union[Result, Base]:
         else:
             return result
 
+    # older method name
+    get_one_base = get_base
+
```
aladshaw3, line 341: Do we plan on phasing this older method out? If so, should we have some kind of depreciation warning (does that even matter of us at this point) ?

watertap/tests/test_notebooks.py
```diff
@@ -0,0 +1,77 @@
...
+# =====
+
+
+@pytest.mark.component
+def test_edb_notebooks(docs_root, electrolytedb):
+    for nb_path in find_notebooks(docs_root / "examples" / "edb"):
+        if ".ipynb_checkpoints" in nb_path.parts:
+            continue
+        print(f"run notebook: {nb_path}")
+        run_notebook(nb_path)
```
aladshaw3, line 76: Quick question: This is the directory ('docs/examples/edb') in which I should put my other edb examples I plan on making, correct? And then this code should automatically check those, right?

aladshaw3 (approved):
LGTM. Just a few brief questions in my review.

watertap/edb/db_api.py
aladshaw3: This function is now unneeded

watertap/tests/test_notebooks.py
aladshaw3: yes

adam-a-a (approved):
LGTM

PR #261, opened 2021-11-16
Adding base classes for WT3 functionality

andrewlee94:
## Summary/Motivation:
This PR adds base classes for WT3 type property packages and zero-order unit models along with associated tests.

-
-

### Legal Acknowledgement


watertap/core/zero_order_properties.py
```diff
@@ -0,0 +1,304 @@
...
+
+    # -------------------------------------------------------------------------
+    # Other properties
+    def _cp_mass(self):
+        add_object_reference(self, "cp_mass", self.params.cp_mass)
+
+    def _dens_mass(self):
+        add_object_reference(self, "dens_mass", self.params.dens_mass)
+
+    def get_material_flow_terms(blk, p, j):
```
adam-a-a, line 278: Do we need the `p` here?

andrewlee94: Yes, we do. It is a part of the framework API, so when this method is called it will always be passed `p` (even though we don't actually need it in this case).

watertap/core/zero_order_unit.py
```diff
@@ -0,0 +1,264 @@
...
+            doc="Material properties at waste outlet",
+            default=tmp_dict_2)
+
+        # Create Ports
+        self.add_port("inlet", self.properties_in, doc="Inlet port")
+        self.add_port("outlet", self.properties_out, doc="Outlet port")
+        self.add_port("waste", self.properties_waste, doc="Waste port")
+
+        # Add performance variables
+        self.water_recovery = Var(
```
adam-a-a: In the NF and RO unit models, we've used `recovery_vol_phase` for volumetric recovery (the typical form) and `recovery_mass_phase_comp` for mass recovery of each component. In this case, `recovery_vol_phase` is the equivalent. Perhaps we want to keep this as is at the moment for the WT3 migration to prevent potential issues with renaming across files that make use of `water_recovery`, but just wanted to make note of that so that we are all aware.

watertap/core/zero_order_unit.py
```diff
@@ -0,0 +1,264 @@
...
+            doc='Water recovery fraction')
+        self.removal_fraction = Var(
+            self.flowsheet().time,
+            self.config.property_package.solute_set,
+            domain=NonNegativeReals,
+            initialize=0.01,
+            units=pyunits.dimensionless,
+            doc='Solute removal fraction')
+
+        self.deltaP_outlet = Var(
```
adam-a-a: I'm not sure that `deltaP_outlet` makes sense to have in all cases. For example, with RO, `deltaP_outlet` would be the pressure at the permeate outlet (typically atmospheric or slightly higher and on one side of the membrane if considering a flat sheet element) minus the feed pressure at the inlet. @kurbansitterley would probably have some better insight on when/where it makes sense to have both `deltaP_outlet` and `deltaP_waste` for WT3 models, but from my perspective, it seems like we would only be concerned with `deltaP_waste`. If that is true, we could eliminate `deltaP_outlet` and eventually rename `deltaP_waste` as `deltaP`.

watertap/core/zero_order_unit.py
```diff
@@ -0,0 +1,264 @@
...
+        def outlet_pressure_constraint(b, t):
+            return (b.properties_in[t].pressure + b.deltaP_outlet[t] ==
+                    b.properties_out[t].pressure)
+
+        @self.Constraint(self.flowsheet().time, doc='Waste pressure equation')
+        def waste_pressure_constraint(b, t):
+            return (b.properties_in[t].pressure + b.deltaP_waste[t] ==
+                    b.properties_waste[t].pressure)
+
+        # TEmperature equality
```
adam-a-a:
```suggestion
        # Temperature equality
```

watertap/core/zero_order_unit.py
```diff
@@ -0,0 +1,264 @@
...
+                         self.config.property_package.solute_set,
+                         doc='Solute mass balance equations')
+        def solute_mass_balances(b, t, j):
+            return (b.properties_in[t].get_material_flow_terms("Liq", j) ==
+                    b.properties_out[t].get_material_flow_terms("Liq", j) +
+                    b.properties_waste[t].get_material_flow_terms("Liq", j))
+
+        # Pressure drop
+        @self.Constraint(self.flowsheet().time, doc='Outlet pressure equation')
+        def outlet_pressure_constraint(b, t):
```
adam-a-a, line 202: For the case of NF/RO, I don't think we would need this.

watertap/core/zero_order_unit.py
andrewlee94: I am not sure that is a good idea - there could be cases where we need both (or even cases where we want `deltaP_outlet` and not `deltaP_waste`). However, there are ways we can make these terms optional, and let the user or the derived classes decide which terms to include.

Two ideas I have has so far:

1. Have `has_pressure_change` config arguments for each outlet, and rely on the user to set them. This has some minor issues for cases where a term is not optional however, as the config argument will still exist (but only have one valid value).
2. Have a private attribute on the unit model which the derived class sets before calling `super().build()`. This hides the option from the user and lets the unit model developer make the decision, which is probably a better option as the unit model would also then know whether it needs to look up a value for `deltaP` from the database. The downside is that the unit model developer needs to be aware of this and set the private attribute correctly.

watertap/core/zero_order_unit.py
andrewlee94: This ties into the comment above on whether and how `deltaP` should be an optional term in these types of units.

watertap/core/zero_order_unit.py
andrewlee94: For now, I went with the WT3 naming convention (for the most part), but we should think about standardizing naming.

watertap/core/zero_order_unit.py
kurbansitterley: Yes, in most cases the pressure on the permeate side will be atmospheric, but I can think of two scenarios where you might want a `deltaP_outlet` variable:

- In a multi-pass system where the permeate from the first RO unit goes to another RO unit. This is done for example when the Boron concentration in the permeate requires further reduction to meet a regulation. Something like this could require a booster pump in between the first and second passes, so you might want to know the pressure differential coming out of the permeate just to be consistent in the momentum balance.
- If you want to use "permeate throttling" as a way to direct/control flow between stages. This is where you would intentionally throttle pressure on the permeate side, thus reducing flux for that module and sending more flow to the next stage. I am honestly not very familiar with this configuration/operation and only became acquainted with it because it was used at Kay Bailey (one of the baseline studies; we did not include permeate throttling in our model) so I looked it up. Attached figure and more information on permeate throttling from [here](https://membranes.com/docs/trc/flowcon.pdf).

Screen Shot 2021-11-16 at 12 51 08 PM

My vote would be to at least have the ability to have both, but regardless to keep the `_waste` in the variable name (i.e. `deltaP_waste` rather than `deltaP`) just in the interest of being explicit. I can't really comment on the implementation for watertap but I hope this is helpful.

watertap/core/zero_order_unit.py
andrewlee94: I changes these to `recovery_vol` and `removal_mass_solute` to better fit the naming convention. Seeing as these models only support 1 phase, I did not include the phase index.

watertap/core/zero_order_unit.py
adam-a-a: After reading your comments, I think it is best to keep the two options, and it looks like @andrewlee94 implemented his second option. With regard to RO, @kurbansitterley, I think these examples would be resolved by specifying the permeate outlet pressure explicitly rather than the pressure change between feed inlet and permeate outlet. Nevertheless, keeping both deltaP variables seems like the way to go.

docs/technical_reference/core/zero_order_unit.rst
```diff
@@ -0,0 +1,130 @@
+Zero Order Unit Model Base Class
+================================
+
+.. index::
+   pair: watertap.core.zero_order_unit;SITOBaseData
+
+.. currentmodule:: watertap.core.zero_order_unit
+
+The zero-order unit model base class is intended to form the basis for a library of generic single inlet-two outlet (SITO) units models using recovery and removal fractions to partition the incoming flow between the two outlets. This base class is not intended to be used by itself, and requires the user to define a derived class before it can be used in a flowsheet.
```
adam-a-a:
```suggestion
The zero-order unit model base class is intended to form the basis for a library of generic single inlet-two outlet (SITO) unit models using recovery and removal fractions to partition the incoming flow between the two outlets. This base class is not intended to be used by itself and requires the user to define a derived class before it can be used in a flowsheet.
```

docs/technical_reference/core/zero_order_unit.rst
```diff
@@ -0,0 +1,130 @@
...
+:math:`\Delta P_{outlet,t}`  deltaP_outlet        time           Pressure difference between inlet and outlet stream. Optional.
+:math:`\Delta P_{waste,t}`   deltaP_waste         time           Pressure difference between inlet and waste stream. Optional.
+============================ ==================== ============== ================================================================
+
+The :math:`\Delta P_{outlet,t}` and :math:`\Delta P_{waste,t}` terms are optional and are only created if the `self._has_deltaP_outlet` and `self._has_deltaP_waste` attributes are `True` respectively.
+
+Constraints
+-----------
+
+The SITO base class writes the following constraints which relate the inlet state to those in the outlet and waste streams. As mentioned previously, these constraint are formulated such that they will be linear if `flow_vol`, `conc_mass_comp`, `pressure` and `temperature` are the state variables used by the property package and the recovery and removal fractions are fixed.
```
adam-a-a:
```suggestion
The SITO base class writes the following constraints which relate the inlet state to those in the outlet and waste streams. As mentioned previously, these constraints are formulated such that they will be linear if `flow_vol`, `conc_mass_comp`, `pressure` and `temperature` are the state variables used by the property package and the recovery and removal fractions are fixed.
```

watertap/core/tests/test_zero_order_unit.py
```diff
@@ -0,0 +1,677 @@
...
+            self._has_deltaP_waste = True
+            super().build()
+
+    @pytest.mark.unit
+    def test_phase_list(self, model):
+        model.fs.params.phase_list = ["foo"]
+
+        with pytest.raises(ConfigurationError,
+                           match="fs.unit configured with invalid property "
+                           "package. Zero-order models only support proeprty "
```
adam-a-a:
```suggestion
                           "package. Zero-order models only support property "
```

watertap/core/tests/test_zero_order_unit.py
```diff
@@ -0,0 +1,677 @@
...
+
+    m = ConcreteModel()
+
+    m.fs = FlowsheetBlock(default={"dynamic": False})
+
+    m.fs.water_props = WaterParameterBlock(
+        default={"solute_list": ["A", "B", "C"]})
+
+    with pytest.raises(NotImplementedError,
+                       match="fs.unit derived class class has not implemented "
```
adam-a-a:
```suggestion
                       match="fs.unit derived class has not been implemented "
```

watertap/core/tests/test_zero_order_unit.py
```diff
@@ -0,0 +1,677 @@
...
```
adam-a-a:
```suggestion
                       match="fs.unit derived class has not been implemented "
```

watertap/core/tests/test_zero_order_unit.py
```diff
@@ -0,0 +1,677 @@
...
+    @pytest.mark.component
+    def test_initialization(self, model):
+        initialization_tester(model)
+
+    @pytest.mark.component
+    def test_solve(self, model):
+        results = solver.solve(model)
+
+        # Check for optimal solution
+        assert results.solver.termination_condition == \
```
adam-a-a, line 292: Can use `assert_optimal_termination(results)` method from `pyomo.environ` instead

watertap/core/tests/test_zero_order_unit.py
```diff
@@ -0,0 +1,677 @@
...
```
adam-a-a, line 630: Same as above (although this works and no dire need to change if you don't want to)

watertap/core/zero_order_unit.py
```diff
@@ -0,0 +1,380 @@
...
+# Some more inforation about this module
+__author__ = "Andrew Lee"
+
+
+class SITOBaseData(UnitModelBlockData):
+    """
+    Standard base class for single inlet-two outlet unit models.
+
+    This class is intended to be used for creating derived model classes and
+    cannot be instantiated as by itself. When creating derived classes,
```
adam-a-a:
```suggestion
    cannot be instantiated by itself. When creating derived classes,
```

watertap/core/zero_order_unit.py
```diff
@@ -0,0 +1,380 @@
...
+    developers must set the '_has_deltaP_outlet` and `_has_deltaP_waste`
+    attributes on the model before calling `super().build()`. These attributes
+    determine whetehr the deltaP terms for the outlet and waste streams are
```
adam-a-a:
```suggestion
    determine whether the deltaP terms for the outlet and waste streams are
```

watertap/core/zero_order_unit.py
```diff
@@ -0,0 +1,380 @@
...
+        and used when constructing these, **default** - None.
+        **Valid values:** {see property package for documentation.}'''))
+
+    def build(self):
+        super().build()
+
+        # Check that derived class has implemented flags for pressure change
+        if not hasattr(self, "_has_deltaP_outlet"):
+            raise NotImplementedError(
+                f"{self.name} derived class class has not implemented "
```
adam-a-a:
```suggestion
                f"{self.name} derived class has not been implemented "
```

watertap/core/zero_order_unit.py
```diff
@@ -0,0 +1,380 @@
...
+                f"_has_deltaP_outlet.")
+        if not hasattr(self, "_has_deltaP_waste"):
+            raise NotImplementedError(
+                f"{self.name} derived class class has not implemented "
```
adam-a-a:
```suggestion
                f"{self.name} derived class has not been implemented "
```

watertap/core/zero_order_unit.py
```diff
@@ -0,0 +1,380 @@
...
+                f"_has_deltaP_waste.")
+
+        # Check that property package meets requirements
+        if self.config.property_package.phase_list != ["Liq"]:
+            raise ConfigurationError(
+                f"{self.name} configured with invalid property package. "
+                f"Zero-order models only support proeprty packages with a "
```
adam-a-a:
```suggestion
                f"Zero-order models only support property packages with a "
```

adam-a-a (commented):
@andrewlee94 Just a few typos that I fixed and a couple quick, optional changes suggested

adam-a-a:
If the penultimate commit (c8cd71f) fails, then it is simply due to fixing typos in a test and missing the typo in either the property or unit model base class

adam-a-a (approved):
LGTM - up to @andrewlee94 if you care to throw in the `assert_optimal_termination` method in tests, but it shouldn't hold up the PR.

kurbansitterley (approved):
I am not familiar with the testing files so I was mostly looking at the `zero_order_unit` ... I would wait for another reviewer before a merge.

PR #263, opened 2021-11-17
Documentation for why we want EDB

### LINKED ISSUE #141: Background on why we need/want the EDB
A brief doc on why we want the EDB and how it simplifies the setup of chemistry configurations for model development would be beneficial. 

At the surface-level, it may seem obvious that storing data in a database for simple queries and loading would be more efficient than manually entering thermo config dictionaries (for example), but it would be nice if we could really spell out the purpose and benefits of having the EDB and the role that it plays specifically in Proteuslib. 

From a more holistic perspective, perhaps we should come up with a schema (if a currently relevant one doesn't already exist) that illustrates the main components of a complete Proteuslib workflow and where the EDB fits in.

aladshaw3:
## Fixes/Addresses:

#141 

- Adding documentation on why we want edb
-

### Legal Acknowledgement


adam-a-a (dismissed):
LGTM

docs/background/why_use_edb.rst
```diff
@@ -0,0 +1,56 @@
...
+code duplication, as each of the configuration dictionaries would have significant overlap with
+each other.
+
+The EDB will also greatly simplify how WaterTAP users can define their own system chemistry. As an
+example, take a look at the difference between the amount of code and knowledge that is necessary
+to create configuration dictionaries :ref:`manually<how_to_setup_simple_chemistry>` versus how
+the EDB can generate those same configuration dictionaries :ref:`automatically<how_to_use_edb>`
+in the 'Simple EDB Example'.
+
+To start using EDB, have a look at the :ref:`installation instructions<install-edb>`
```
lbianchi-lbl:
```suggestion
To start using EDB, have a look at the :ref:`installation instructions<install-edb>`.
```

lbianchi-lbl (changes requested):
Looks good! I just spotted one missing period, which I think is the nitpicky-est non-whitespace correction one can make.

lbianchi-lbl (approved):

adam-a-a (approved):

PR #264, opened 2021-11-18
Adding leaky isobaric pressure exchanger to unit models

avdudchenko:
## Fixes/Addresses:
This adds a modified pressure exchanger unit model that include mass leakage from high pressure to low pressure as would be observed in real isobaric devices.  

## Summary/Motivation:
Real isobaric pressure exchangers have some leakage from high pressure side to low pressure side. 
In piston type pressure exchangers this is a result of imperfect seals allowing for equal passage of salt and water
In direct contact devices, the high pressure fluid directly contacts low pressure fluid in "contact volume cylinder" which results in some mixing of the two fluid, and can lead to unequal rate of salt and liquid passage.  

This model added two variables to configure solvent and slolute passage to existing model.
The transfer of mass from high pressure side to low pressure side is achieved by two newly added constraints eq_mass_transfer_from_high_to_low_pressure and eq_mass_high_pressure_outlet. 

In eq_mass_transfer_from_high_to_low_pressure constraint the low pressure (LP) outlet phase composition is forced to be equal to sum of LP inlet phase composition + high pressure (HP) inlet phase composition multiplied by solute or solvent leakage factors.  
eq_mass_high_pressure_outlet is a redundant constraint (I think....) but it computes out the HP outlet phase composition, by multiplying the HP inlet phase composition with solute or solvent leakage factors. 

This was tested in an isobaric ro flowsheet (not included in this pull) and does not cause issues.

There is also a test_leaky_pressure_exchanger script that includes a mass balance check (building on the test_pressure_exchanger), which passes the tests.

### Legal Acknowledgement

### DIFF
--- a/watertap/unit_models/leaky_pressure_exchanger.py
+++ b/watertap/unit_models/leaky_pressure_exchanger.py
@@ -147,7 +147,7 @@ class PressureExchangerData(UnitModelBlockData):
             bounds=(1e-6, 1),
             domain=NonNegativeReals,
             units=pyunits.dimensionless,
-            doc='Pressure exchanger salt leakage')
+            doc='Pressure exchanger solute leakage')
 
         self.solvent_leakage_fraction = Var(
             self.flowsheet().config.time,
@@ -155,7 +155,7 @@ class PressureExchangerData(UnitModelBlockData):
             bounds=(1e-6, 1),
             domain=NonNegativeReals,
             units=pyunits.dimensionless,
-            doc='Pressure exchanger salt leakage')
+            doc='Pressure exchanger solvent leakage')
         # Build control volume for high pressure side
         self.high_pressure_side = ControlVolume0DBlock(default={
             "dynamic": False,
--- a/watertap/unit_models/leaky_pressure_exchanger.py
+++ b/watertap/unit_models/leaky_pressure_exchanger.py
@@ -182,8 +182,6 @@ class PressureExchangerData(UnitModelBlockData):
         def work(b, t):
             return b.properties_in[t].flow_vol * b.deltaP[t]
 
-
-
         @self.high_pressure_side.Expression(
             self.flowsheet().config.time,
             self.config.property_package.phase_list,
@@ -191,7 +189,7 @@ class PressureExchangerData(UnitModelBlockData):
             doc='mass transferred to low pressure side')
         def mass_transfer(b, t, p, j):
             return b.properties_in[t].flow_mass_phase_comp[p,j]-b.properties_out[t].flow_mass_phase_comp[p,j]
-            
+
         #Build control volume for low pressure side
         self.low_pressure_side = ControlVolume0DBlock(default={
             "dynamic": False,
@@ -217,6 +215,14 @@ class PressureExchangerData(UnitModelBlockData):
             doc='Work transferred to low pressure side fluid')
         def work(b, t):
             return b.properties_in[t].flow_vol * b.deltaP[t]
+
+        @self.low_pressure_side.Expression(
+            self.flowsheet().config.time,
+            self.config.property_package.phase_list,
+            self.config.property_package.component_list,
+            doc='mass transferred to low pressure side (will be negative, gained mass)')
+        def mass_transfer(b, t, p, j):
+            return b.properties_in[t].flow_mass_phase_comp[p,j]-b.properties_out[t].flow_mass_phase_comp[p,j]
 
         #Add Ports
         self.add_inlet_port(name='high_pressure_inlet', block=self.high_pressure_side)
@@ -255,13 +261,13 @@ class PressureExchangerData(UnitModelBlockData):
                 b.high_pressure_side.properties_in[t].flow_mass_phase_comp[p,j]*b.solute_leakage_fraction[t]+
                 b.low_pressure_side.properties_in[t].flow_mass_phase_comp[p,j])
 
-        ### probably not necessary ###
+        ### this constraint (eq_mass_high_pressure_outlet) might not be needed###
         @self.Constraint(
             self.flowsheet().config.time,
             self.config.property_package.phase_list,
             self.config.property_package.component_list,
             doc="mass loss from HP inlet propagation to HP outlet")
-        def eq_mass_high_pressure_outlet(b, t,p, j):
+        def eq_mass_high_pressure_outlet(b, t, p, j):
             comp = self.config.property_package.get_component(j)
             if comp.is_solvent():
                 return (b.high_pressure_side.properties_out[t].flow_mass_phase_comp[p,j]==\
@@ -402,7 +408,12 @@ class PressureExchangerData(UnitModelBlockData):
         if iscale.get_scaling_factor(self.efficiency_pressure_exchanger) is None:
             # efficiency should always be between 0.1-1
             iscale.set_scaling_factor(self.efficiency_pressure_exchanger, 1)
-
+        if iscale.get_scaling_factor(self.solute_leakage_fraction) is None:
+            # efficiency should always be between 0.1-1
+            iscale.set_scaling_factor(self.solute_leakage_fraction, 1)
+        if iscale.get_scaling_factor(self.solvent_leakage_fraction) is None:
+            # efficiency should always be between 0.1-1
+            iscale.set_scaling_factor(self.solvent_leakage_fraction, 1)
         # scale expressions
         if iscale.get_scaling_factor(self.low_pressure_side.work) is None:
             sf = iscale.get_scaling_factor(self.low_pressure_side.properties_in[0].flow_vol)
@@ -414,6 +425,16 @@ class PressureExchangerData(UnitModelBlockData):
             sf = sf * iscale.get_scaling_factor(self.high_pressure_side.deltaP[0])
             iscale.set_scaling_factor(self.high_pressure_side.work, sf)
 
+        if iscale.get_scaling_factor(self.high_pressure_side.mass_transfer) is None:
+            for p,j in self.high_pressure_side.properties_in[0].flow_mass_phase_comp:
+                sf = iscale.get_scaling_factor(self.high_pressure_side.properties_in[0].flow_mass_phase_comp[p,j])
+                sf = sf * iscale.get_scaling_factor(self.high_pressure_side.properties_out[0].flow_mass_phase_comp[p,j])
+                iscale.set_scaling_factor(self.high_pressure_side.mass_transfer, sf)
+        if iscale.get_scaling_factor(self.low_pressure_side.mass_transfer) is None:
+            for p,j in self.low_pressure_side.properties_in[0].flow_mass_phase_comp:
+                sf = iscale.get_scaling_factor(self.low_pressure_side.properties_in[0].flow_mass_phase_comp[p,j])
+                sf = sf * iscale.get_scaling_factor(self.low_pressure_side.properties_out[0].flow_mass_phase_comp[p,j])
+                iscale.set_scaling_factor(self.low_pressure_side.mass_transfer, sf)
         # transform constraints
         for t, c in self.low_pressure_side.eq_isothermal_temperature.items():
             sf = iscale.get_scaling_factor(self.low_pressure_side.properties_in[t].temperature)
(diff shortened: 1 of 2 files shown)


TimBartholomew (changes requested):
This really should be incorporated with the pressure exchanger model (i.e. leaky should be a configuration option).

watertap/unit_models/leaky_pressure_exchanger.py
```diff
@@ -0,0 +1,491 @@
...
+        def work(b, t):
+            return b.properties_in[t].flow_vol * b.deltaP[t]
+
+        @self.high_pressure_side.Expression(
+            self.flowsheet().config.time,
+            self.config.property_package.phase_list,
+            self.config.property_package.component_list,
+            doc='Mass transfered on high pressure side')
+        def mass_transfer(b, t, p, j):
+            return b.properties_in[t].flow_mass_phase_comp[p,j]-b.properties_out[t].flow_mass_phase_comp[p,j]
```
andrewlee94, line 191: As above, the control volume is already creating this for you, so you are now replacing the existing Var.

watertap/unit_models/leaky_pressure_exchanger.py
```diff
@@ -0,0 +1,491 @@
...
+        self.high_pressure_side.add_momentum_balances(
+            balance_type=self.config.momentum_balance_type,
+            has_pressure_change=True)
+
+        self.high_pressure_side.deltaP.setub(0)
+
+        @self.high_pressure_side.Expression(
+            self.flowsheet().config.time,
+            doc='Work transferred to high pressure side fluid (should be negative)')
+        def work(b, t):
```
andrewlee94, line 182: The `work` term is created automatically by the control volume when `has_pressure_change=True`, so this is not needed. Worse, it is implicitly replacing the existing `work` Var which is likely to cease issues in future.

watertap/unit_models/leaky_pressure_exchanger.py
```diff
@@ -0,0 +1,491 @@
...
+        @self.Constraint(
+            self.flowsheet().config.time,
+            self.config.property_package.phase_list,
+            self.config.property_package.component_list,
+            doc="mass leak from HP inlet to LP outlet")
+        def eq_mass_transfer_from_high_to_low_pressure_outlet(b, t, p, j):
+            comp = self.config.property_package.get_component(j)
+            if comp.is_solvent():
+                return (b.low_pressure_side.properties_out[t].flow_mass_phase_comp[p,j]==\
+                b.high_pressure_side.properties_in[t].flow_mass_phase_comp[p,j]*b.solvent_leakage_fraction[t]+
```
TimBartholomew, line 257: You can use general methods here so that the pressure exchanger model can work with property packages that use different state variables. For example, if you use get_material_flow_terms() like: https://github.com/watertap-org/watertap/blob/e7d0c000b7f1a3e23afde9e3820ec9ca0fdfed84/watertap/unit_models/reverse_osmosis_0D.py#L318-L320
Then any basis would work. Currently, WaterTAP property packages are primarily FpcTP (mass flow based on phase and component, temperature, and pressure), but in the future other property packages could be used like FTPx (mass flow, temperature, pressure, and mass fraction) or volumetric flow rate and concentration based.

watertap/unit_models/leaky_pressure_exchanger.py
```diff
@@ -0,0 +1,491 @@
...
+            balance_type=self.config.material_balance_type,
+            has_mass_transfer=True)
+
+        self.low_pressure_side.add_momentum_balances(
+            balance_type=self.config.momentum_balance_type,
+            has_pressure_change=True)
+
+        self.low_pressure_side.deltaP.setlb(0)
+
+        @self.low_pressure_side.Expression(
```
andrewlee94, line 213: As above.

watertap/unit_models/leaky_pressure_exchanger.py
```diff
@@ -0,0 +1,491 @@
...
+            self.flowsheet().config.time,
+            doc='Work transferred to low pressure side fluid')
+        def work(b, t):
+            return b.properties_in[t].flow_vol * b.deltaP[t]
+
+        @self.low_pressure_side.Expression(
```
andrewlee94, line 219: As above.

watertap/unit_models/leaky_pressure_exchanger.py
```diff
@@ -0,0 +1,491 @@
...
+        def eq_equal_flow_vol(b, t):
+            return (b.high_pressure_side.properties_in[t].flow_vol ==
+                    b.low_pressure_side.properties_in[t].flow_vol)
+
+        @self.Constraint(
+            self.flowsheet().config.time,
+            self.config.property_package.phase_list,
+            self.config.property_package.component_list,
+            doc="mass leak from HP inlet to LP outlet")
+        def eq_mass_transfer_from_high_to_low_pressure_outlet(b, t, p, j):
```
andrewlee94, line 253: It would be far more efficient to just write:

```
b.high_pressure_side.mass_transfer[p, j] == b.solvent_leakage_fraction[t]*b.high_pressure_side.properties_in[t].flow_mass_phase_comp[p,j]
```

watertap/unit_models/leaky_pressure_exchanger.py
```diff
@@ -0,0 +1,491 @@
...
+            units=pyunits.dimensionless,
+            doc='Pressure exchanger solute leakage')
+
+        self.solvent_leakage_fraction = Var(
+            self.flowsheet().config.time,
+            initialize=0.05,
+            bounds=(1e-6, 1),
+            domain=NonNegativeReals,
+            units=pyunits.dimensionless,
+            doc='Pressure exchanger solvent leakage')
```
TimBartholomew, line 158: I would of thought that there should just be one leakage fraction because isn't it the whole solution that leaks? Or is there some preferential mechanism that has more solvent or solute leak like diffusion?

watertap/unit_models/leaky_pressure_exchanger.py
```diff
@@ -0,0 +1,491 @@
...
+            doc="mass leak from HP inlet to LP outlet")
+        def eq_mass_transfer_from_high_to_low_pressure_outlet(b, t, p, j):
+            comp = self.config.property_package.get_component(j)
+            if comp.is_solvent():
+                return (b.low_pressure_side.properties_out[t].flow_mass_phase_comp[p,j]==\
+                b.high_pressure_side.properties_in[t].flow_mass_phase_comp[p,j]*b.solvent_leakage_fraction[t]+
+                b.low_pressure_side.properties_in[t].flow_mass_phase_comp[p,j])
+            if comp.is_solute():
+                return (b.low_pressure_side.properties_out[t].flow_mass_phase_comp[p,j]==\
+                b.high_pressure_side.properties_in[t].flow_mass_phase_comp[p,j]*b.solute_leakage_fraction[t]+
```
andrewlee94, line 261: Should `solute_leackage_fraction` be indexed by solute, and if not then will it actually be different from the solvent leakage fraction?

watertap/unit_models/leaky_pressure_exchanger.py
```diff
@@ -0,0 +1,491 @@
...
+                b.low_pressure_side.properties_in[t].flow_mass_phase_comp[p,j])
+
+        ### this constraint (eq_mass_balance_high_pressure_outlet) might not be needed###
```
andrewlee94, line 264: This is definitely not needed - it is redundant when combined with the mass balances in the control volumes and the above constraint, If it is not removed, then the model will be degenerate and will not solve.

watertap/unit_models/leaky_pressure_exchanger.py
```diff
@@ -0,0 +1,491 @@
...
+
+        @self.high_pressure_side.Constraint(
+            self.flowsheet().config.time,
+            doc="Isothermal constraint")
+        def eq_isothermal_temperature(b, t):
+            return b.properties_in[t].temperature == b.properties_out[t].temperature
+
+    def initialize(
+            self,
+            state_args=None,
```
andrewlee94, line 300: We probably need to have two sets of state args - one each for the high and low pressure sides - as conditions in each side will be significantly different,

andrewlee94 (changes requested):
I had a number of comments on the model equations. Notably, there are a lot of places where the unit model is implicitly replacing things already created by the control volumes, and one redundant/degenerate constraint.

watertap/unit_models/leaky_pressure_exchanger.py
avdudchenko: Will remove then.

watertap/unit_models/leaky_pressure_exchanger.py
avdudchenko: This is actually in original pressureexchagner model (should this term be removed then as well)?

watertap/unit_models/leaky_pressure_exchanger.py
avdudchenko: Hmm, it does solve with this, but will remove.

watertap/unit_models/leaky_pressure_exchanger.py
avdudchenko: This depends on design of pressure exchanger. In some of them you might have simply diffusion or mixing happing in the exchange volume so you only have transfer of solute, but now liquid mass transfer to account for (or it might be very minimal). 
Check out this design (https://www.youtube.com/watch?v=mJe1nynkBGk) , the low pressure and high pressure fluid come in direct contact and there is no "piston" to separate them, the two fluids mix, and can exit evenly such that you have only solute transport but minimal or on water transport.

watertap/unit_models/leaky_pressure_exchanger.py
avdudchenko: Not sure that level of detail will be needed. This is to account for designs where the high pressure fluid and low pressure fluid mix in exchange volume so effectively only solute transfers.

watertap/unit_models/leaky_pressure_exchanger.py
avdudchenko: will update to this ( and merge in to pressure exchanger model.

watertap/unit_models/leaky_pressure_exchanger.py
avdudchenko: Explained for cases where there convective/diffusion mixing of high and low pressure fluid in pressure exchange model.

watertap/unit_models/leaky_pressure_exchanger.py
avdudchenko: will fix

watertap/unit_models/leaky_pressure_exchanger.py
avdudchenko: will fix

watertap/unit_models/leaky_pressure_exchanger.py
avdudchenko: will fix

watertap/unit_models/leaky_pressure_exchanger.py
avdudchenko: Thank you for suggestion. Will update.

watertap/unit_models/leaky_pressure_exchanger.py
adam-a-a: @avdudchenko you could also just have `self.leakage_fraction_comp` and index by the `component_list` which would include solvent and solute (we generally use the qualifier "comp" for anything that is indexed by component.) I didn't thoroughly review all of your code yet, but as @TimBartholomew mentioned in another comment, and assuming that these leakage fractions are the only new variables added to the pressure exchanger, you could declare a new boolean configuration option such as `has_leakage`; if `True`, create the leakage fraction vars (and later on, apply that `has_leakage` check to the associated constraints).

watertap/unit_models/leaky_pressure_exchanger.py
avdudchenko: It might be useful todo. The idea is to capture PX that have diffusion/convection happening in the pressure exchange volume. But I would assume convection dominates, so then it would be unnecessary. I will think about this one, might be useful for cases where we expect dilute species that strongly impact permeate quality (boron, or norms for example). Thanks for the suggestion.

watertap/unit_models/leaky_pressure_exchanger.py
avdudchenko: Surprisingly, it solves without errors in a full RO model. Should we be expecting errors to pop up, or was I just lucky? Will remove in next version.

andrewlee94: I would expect to see two things:

1. There will be a degree of freedom issue somewhere. This appears as an extra constraint in the problem, so when aiming to get 0 degrees of freedom this means that we will be able to fix one less variable.
2. The problem with degeneracies is that there still exist feasible solutions, and with good initial guesses you can end up with one of these. However, the Jacobian of the problem is singular, which means if you need to find a new feasible point, some of the information is missing. I.e. degeneracies can often go unnoticed as they can be solved under specific conditions, but can suddenly start failing when you try to do something else.

avdudchenko: That makes sense. Thanks for the info.

watertap/unit_models/leaky_pressure_exchanger.py
avdudchenko: Yes I think that is my plan at this point.

avdudchenko:
Closing this request as it was decided "leaks" will be added to normal pressure exchanger model. Once that is ready I will create a new pull request.

PR #265, opened 2021-11-18
Failing RO initialization fails tests

bknueven:
## Summary/Motivation:
Right now, the RO tests don't fail if initialization fails. When developing new capability sometimes this leads to false confidence, *or* a different test (like checking scaling) fails instead.

- Pass `fail_on_warnings=True` flag into `initialization_tester` in 0D and 1D RO tests
- Make the 1D RO initialization always ignore warnings on intermediate solves.

### Legal Acknowledgement


andrewlee94:
FYI: IDAES was discussing the idea of having `initialize` methods return the final results object so that we could check the status of the solve. However, nobody commented on my suggestion and I hadn't found a good way to make the change without breaking things.

adam-a-a (dismissed):
LGTM

adam-a-a (approved):

TimBartholomew (approved):

PR #268, opened 2021-11-24
Updating pressure exchanger model to include mass transfer option

avdudchenko:
## Fixes/Addresses:

Includes option to enable solute and mass transfer from high to low pressure side.

## Summary/Motivation:
Isobaric pressure exchangers do not perfectly seal of the high pressure from low pressure side.  This updated alowes one to capture two types of seal issues: 
(1) The fluid from high pressure side  through  the piston seal into low pressure side, leading to equal solvent and solute transfer
(2) The device does not use a piston to separate high and low pressure sides, and the high and low pressure fluids come in direct contact. This can result in high and low salinity fluid to mix, leading to diffusive transport, and non-equal transfer of solvent and solute. It is not expected that individual species of solute would have different transport rates, thus we do not include an indexed mass transfer term for the solute. 

-has_mass_transfer option added to original pressure exchanger
-mass transfer is enabled with the has_mass_tranfer option
-if has_mass_transfer ==True:
--two variables are added and define mass transfer fraction for solute and solvent (solvent_transfer_fraction and solute_transfer_fraction)
--a constraint (eq_mass_transfer_from_high_pressure_side)  is added that computes mass_transfer_term for solute and solvent 
--a constraint (eq_mass_transfer_term) is added that connects mass transfer from high pressure side to mass transfer on low pressure

## changes to orignal PX model
--Removed based on prior suggestion
self.low_pressure_side.deltaP.setlb(0) 
self.high_pressure_side.deltaP.setlb(0) 
### Legal Acknowledgement


watertap/unit_models/pressure_exchanger.py
```diff
@@ -140,6 +146,22 @@ def build(self):
             units=pyunits.dimensionless,
             doc='Pressure exchanger efficiency')
 
+        if self.config.has_mass_transfer:
+            self.solute_transfer_fraction = Var(
```
adam-a-a: In case there is more than one solute, perhaps you should add an index for `solute_list`. Or you could consider having `solution_transfer_fraction` instead of splitting it up by solvent and solute.

avdudchenko: Right, this will be more consistent with other modules as well. I updated the code to reflect this.

watertap/unit_models/pressure_exchanger.py
```diff
@@ -118,6 +118,12 @@ class PressureExchangerData(UnitModelBlockData):
         **Valid values:** {
         see property package for documentation.}"""))
 
+    CONFIG.declare("has_mass_transfer", ConfigValue(
+        default=False,
+        domain=In([True, False]),
+        description="Defines if there is mass trnasport between high and low pressure sides",
```
adam-a-a:
```suggestion
        description="Defines if there is mass transport between high- and low-pressure sides",
```

watertap/unit_models/pressure_exchanger.py
```diff
@@ -118,6 +118,12 @@ class PressureExchangerData(UnitModelBlockData):
...
+        doc="""Indicates whether pressure exchange solute and solvent mass transfer terms should be constructed or not.
```
adam-a-a:
```suggestion
        doc="""Indicates whether pressure exchanger solution mass transfer terms should be constructed or not.
```

watertap/unit_models/pressure_exchanger.py
```diff
@@ -140,6 +146,16 @@ def build(self):
             units=pyunits.dimensionless,
             doc='Pressure exchanger efficiency')
 
+        if self.config.has_mass_transfer:
+            self.solution_transfer_fraction = Var(
+                self.flowsheet().config.time,
+                self.config.property_package.component_list,
```
adam-a-a, line 196: Do you still need to index by component if this is now just a fraction for the solution?

watertap/unit_models/pressure_exchanger.py
```diff
@@ -230,6 +243,26 @@ def eq_isothermal_temperature(b, t):
         def eq_isothermal_temperature(b, t):
             return b.properties_in[t].temperature == b.properties_out[t].temperature
 
+        if self.config.has_mass_transfer:
+            @self.Constraint(
+                self.flowsheet().config.time,
+                self.config.property_package.phase_list,
+                self.config.property_package.component_list,
```
adam-a-a, line 308: Remember to remove `component_list` and corresponding indices here, if you find that you actually don't need `component_list`.

watertap/unit_models/pressure_exchanger.py
avdudchenko: We still want to have an option of index by component incase they can have different transfer fractions (e.g. H2O, TDS, NaCl, etc. ) Is there a better/cleaner way to do that? ( I figured doing it by solute/solvent, is not necessary here).

andrewlee94: The problem with doing that is that model structure is then dependent on whether you use component indexing or not. In these cases, I would generally go with always indexing by component - if they are all the same then we can just fix the entire indexed variable to the same value (which can be done in a single line of code).

adam-a-a: Ah,  I thought you decided to not consider different transfer fractions, hence renaming it to `solution_transfer_fraction`, but I see now that you just want to account for all components under one variable name. Maybe renaming it to something like `comp_transfer_fraction` might be clearer.

watertap/unit_models/tests/test_pressure_exchanger.py
```diff
@@ -137,17 +167,28 @@ def test_build():
...
-    assert number_variables(m.fs.unit) == 39
-    assert number_total_constraints(m.fs.unit) == 31
+    assert number_variables(m.fs.unit) == 39 + extra_variables
+    assert number_total_constraints(m.fs.unit) == 31 + extra_constraint
     assert number_unused_variables(m.fs.unit) == 0
 
+
+@pytest.mark.unit
+def test_build_with_mass_transfer():
+    # 2 variables for mass tranfer fractions (solute,solvent)
```
adam-a-a:
```suggestion
    # 2 variables for mass transfer fractions (solute,solvent)
```

watertap/unit_models/tests/test_pressure_exchanger.py
```diff
@@ -137,17 +167,28 @@ def test_build():
...
+    test_build(has_mass_transfer=True,\
+    extra_variables=6,\
+    extra_constraint=4)
+
 class TestPressureExchanger():
     @pytest.fixture(scope="class")
     def unit_frame(self):
         m = ConcreteModel()
         m.fs = FlowsheetBlock(default={'dynamic': False})
         m.fs.properties = props.SeawaterParameterBlock()
```
adam-a-a, line 210: This might cause some headaches and could be reserved for a subsequent PR, but wondering if we'd want another test that uses a property package with multiple ions.

adam-a-a: For example, you could try using the `seawater_ion_generic` property model, which is based on the generic property framework in IDAES; this is likely overkill and issues may arise when trying this, but it includes 5 ions representative of seawater.

adam-a-a (dismissed):
LGTM - I don't want to hold up the PR with my comment about adding a test for multiple ions in solution. I'll leave that decision to @avdudchenko and other reviewers (@TimBartholomew , @andrewlee94 ).

adam-a-a:
Just one more note -- if this doesn't happen in this PR, we should make an issue to update the documentation for the pressure exchanger to reflect the consideration of mixing within the ERD.

watertap/unit_models/pressure_exchanger.py
avdudchenko: That sounds reasonable I will update the name.

watertap/unit_models/tests/test_pressure_exchanger.py
avdudchenko: Will give it a go and probably should address any issues in this PR, as we will likely want to use this model with more complex property packages... Let me see what issues crop up.

adam-a-a: Revisiting `seawater_ion_generic`, I recalled that a molar flow basis is used, which might cause the first upset. You can give it a try (checkout `test_nanofiltration_ZO` to see how `seawater_ion_generic` gets imported and loaded into a `GenericParameterBlock`), but maybe there's an easier way: you can try @andrewlee94 's new `zero_order_properties` located in the `core` folder. This way, you can just declare a solute_list when you build the property model. However, things might still break if certain properties don't match between the unit model and property model.

EDIT: Maybe the simplest path of all would be to use the simple property model we made for full treatment train analysis: `watertap/examples/flowsheets/full_treatment_train/model_components/seawater_ion_prop_pack.py`

avdudchenko: So I tried adding the generic seawater ion generic property package, but it did not work, I would need to add a way to  handle mole fraction to mass fraction conversion, but looking at the generic property packge notes, it does not look like a standard approach was finalized.  

I then tried to add the seawater_ion_prop_pack test, and it solves (and estimates all the right pressure/mass transfers etc.) but is failing dof test and intiatilation test (probably due to dof errror). It suggestes I have -1 DOF, but I can't figure out how/where its being over specified. 

Is there something special I need to do to get it work that I am missing? Here is the unit_frame set up. all the inlets are fixed, and outlets and unfixed? Is there another parameter I need to unfix that is introduced by sewater_ion_prop_pack? 

`
def unit_frame(self):
        m = ConcreteModel()
        m.fs = FlowsheetBlock(default={'dynamic': False})
        m.fs.properties = property_seawater_ions.PropParameterBlock()
        m.fs.unit = PressureExchanger(default={'property_package': m.fs.properties,\
         'has_mass_transfer': True})

        # Specify inlet conditions
        temperature = 25 + 273.15

        lowP_pressure = 101325
        highP_mass_frac_TDS = 0.07
        highP_pressure = 50e5

        solvent_transfer=0.05
        solute_transfer={'Na': 0.1,
                          'Ca': 0.11,
                          'Mg': 0.12,
                          'SO4': 0.13,
                          'Cl': 0.14}

        feed_flow_mass = 1
        lowP_mass_frac = {'Na': 11122e-6,
                          'Ca': 382e-6,
                          'Mg': 1394e-6,
                          'SO4': 2136e-6,
                          'Cl': 20300e-6}

        highP_mass_frac = {'Na': 2*11122e-6,
                          'Ca': 2*382e-6,
                          'Mg': 2*1394e-6,
                          'SO4': 2*2136e-6,
                          'Cl':2*20300e-6}


        m.fs.unit.solution_transfer_fraction[0,'H2O'].fix(solvent_transfer)
        for s in solute_transfer:
            m.fs.unit.solution_transfer_fraction[0, s].fix(solute_transfer[s])

        m.fs.unit.low_pressure_side.properties_in[0].flow_mass_phase_comp['Liq', 'H2O'].fix(
            feed_flow_mass * (1 - sum(x for x in lowP_mass_frac.values())))
        for s in lowP_mass_frac:
            m.fs.unit.low_pressure_side.properties_in[0].flow_mass_phase_comp['Liq', s].fix(feed_flow_mass * lowP_mass_frac[s])
        m.fs.unit.low_pressure_side.properties_in[0].pressure.fix(lowP_pressure)
        m.fs.unit.low_pressure_side.properties_in[0].temperature.fix(temperature)
        m.fs.unit.low_pressure_side.properties_in[0].mass_frac_phase_comp
        m.fs.unit.low_pressure_side.properties_in[0].flow_mol_phase_comp
        m.fs.unit.low_pressure_side.properties_in[0].flow_vol

        m.fs.unit.high_pressure_side.properties_in[0].flow_mass_phase_comp['Liq', 'H2O'].fix(
            feed_flow_mass * (1 - sum(x for x in highP_mass_frac.values())))
        for s in highP_mass_frac:
            m.fs.unit.high_pressure_side.properties_in[0].flow_mass_phase_comp['Liq', s].fix(feed_flow_mass * highP_mass_frac[s])

        m.fs.unit.high_pressure_side.properties_in[0].pressure.fix(highP_pressure)
        m.fs.unit.high_pressure_side.properties_in[0].temperature.fix(temperature)
        m.fs.unit.high_pressure_side.properties_in[0].mass_frac_phase_comp
        m.fs.unit.high_pressure_side.properties_in[0].flow_mol_phase_comp
        m.fs.unit.high_pressure_side.properties_in[0].flow_vol


        efficiency = 0.95
        m.fs.unit.efficiency_pressure_exchanger.fix(efficiency)

        return m `

avdudchenko: Latest commit fixed the issue.

watertap/unit_models/tests/test_pressure_exchanger.py
```diff
@@ -246,6 +293,202 @@ def test_solution(self, unit_frame):
...
+        for s in lowP_mass_frac:
+            m.fs.unit.low_pressure_side.properties_in[0].flow_mass_phase_comp['Liq', s].fix(feed_flow_mass * lowP_mass_frac[s])
+        m.fs.unit.low_pressure_side.properties_in[0].pressure.fix(lowP_pressure)
+        m.fs.unit.low_pressure_side.properties_in[0].temperature.fix(temperature)
+        m.fs.unit.low_pressure_side.properties_in[0].mass_frac_phase_comp
+        m.fs.unit.low_pressure_side.properties_in[0].flow_mol_phase_comp
+        m.fs.unit.low_pressure_side.properties_in[0].flow_vol
+
+        ### NOTE THE PRESSURE EXHCANGER EXPECTS EQUAL VOLUMETRIC FLOW RATES so one of mass fractions on high pressure or low pressure sides
+        # should be left unfixed, as other wise you are fixing all the mass fractions, and forcing equal flow, which leaves no degrees of freedom
```
adam-a-a: @avdudchenko Which mass fraction is being unfixed here as compared to before when you got the -1 DOF?

watertap/unit_models/tests/test_pressure_exchanger.py
```diff
@@ -246,6 +293,202 @@ def test_solution(self, unit_frame):
...
+
+        m.fs.unit.low_pressure_side.properties_in[0].flow_mass_phase_comp['Liq', 'H2O'].fix(
+            feed_flow_mass * (1 - sum(x for x in lowP_mass_frac.values())))
+        for s in lowP_mass_frac:
+            m.fs.unit.low_pressure_side.properties_in[0].flow_mass_phase_comp['Liq', s].fix(feed_flow_mass * lowP_mass_frac[s])
+        m.fs.unit.low_pressure_side.properties_in[0].pressure.fix(lowP_pressure)
+        m.fs.unit.low_pressure_side.properties_in[0].temperature.fix(temperature)
+        m.fs.unit.low_pressure_side.properties_in[0].mass_frac_phase_comp
+        m.fs.unit.low_pressure_side.properties_in[0].flow_mol_phase_comp
+        m.fs.unit.low_pressure_side.properties_in[0].flow_vol
```
adam-a-a: Are these lines necessary?

watertap/unit_models/tests/test_pressure_exchanger.py
```diff
@@ -246,6 +293,202 @@ def test_solution(self, unit_frame):
...
+           feed_flow_mass * (1 - sum(x for x in highP_mass_frac.values()))
+
+        for s in highP_mass_frac:
+            m.fs.unit.high_pressure_side.properties_in[0].flow_mass_phase_comp['Liq', s].fix(feed_flow_mass * highP_mass_frac[s])
+
+        m.fs.unit.high_pressure_side.properties_in[0].pressure.fix(highP_pressure)
+        m.fs.unit.high_pressure_side.properties_in[0].temperature.fix(temperature)
+        m.fs.unit.high_pressure_side.properties_in[0].mass_frac_phase_comp
+        m.fs.unit.high_pressure_side.properties_in[0].flow_mol_phase_comp
+        m.fs.unit.high_pressure_side.properties_in[0].flow_vol
```
adam-a-a: Are these lines necessary?

ksbeattie:
@avdudchenko, any news on this?

watertap/unit_models/pressure_exchanger.py
TimBartholomew: The name still needs to be updated. I think a more IDAES name would be `mass_transfer_fraction_comp`

watertap/unit_models/pressure_exchanger.py
```diff
@@ -151,20 +167,18 @@ def build(self):
...
         self.high_pressure_side.add_material_balances(
-            balance_type=self.config.material_balance_type)
+            balance_type=self.config.material_balance_type,
+            has_mass_transfer=self.config.has_mass_transfer)
 
         self.high_pressure_side.add_momentum_balances(
             balance_type=self.config.momentum_balance_type,
             has_pressure_change=True)
 
-        self.high_pressure_side.deltaP.setub(0)
```
TimBartholomew, line 200: What was the reasoning to remove this bound? It should still be always true.

watertap/unit_models/pressure_exchanger.py
```diff
@@ -176,14 +190,13 @@ def work(b, t):
...
         self.low_pressure_side.add_material_balances(
-            balance_type=self.config.material_balance_type)
+            balance_type=self.config.material_balance_type,
+            has_mass_transfer=self.config.has_mass_transfer)
 
         self.low_pressure_side.add_momentum_balances(
             balance_type=self.config.momentum_balance_type,
             has_pressure_change=True)
 
-        self.low_pressure_side.deltaP.setlb(0)
```
TimBartholomew, line 229: Same as above.

watertap/unit_models/pressure_exchanger.py
```diff
@@ -208,8 +221,12 @@ def eq_pressure_transfer(b, t):
...
             doc="Equal volumetric flow rate")
         def eq_equal_flow_vol(b, t):
-            return (b.high_pressure_side.properties_in[t].flow_vol ==
-                    b.low_pressure_side.properties_in[t].flow_vol)
+            if self.config.has_mass_transfer:
+                return (b.high_pressure_side.properties_out[t].flow_vol ==
+                        b.low_pressure_side.properties_in[t].flow_vol)
+            else:
+                return (b.high_pressure_side.properties_in[t].flow_vol ==
+                        b.low_pressure_side.properties_in[t].flow_vol)
```
TimBartholomew: I think this conditional can be removed. The relationship for when there is mass transfer is still applicable for when there isn't mass transfer.

watertap/unit_models/pressure_exchanger.py
```diff
@@ -376,6 +420,19 @@ def calculate_scaling_factors(self):
...
+        if hasattr(self, 'eq_mass_transfer_from_high_pressure_side'):
+            for (t,p,j), c in self.eq_mass_transfer_from_high_pressure_side.items():
+                sf = iscale.get_scaling_factor(self.high_pressure_side.properties_in[t].get_material_flow_terms(p,j))
+                iscale.constraint_scaling_transform(c, sf)
+
+        if hasattr(self, 'eq_mass_transfer_term'):
+            for (t,p,j), c in self.eq_mass_transfer_term.items():
+                sf = iscale.get_scaling_factor(self.high_pressure_side.mass_transfer_term[t, p, j])
+                sf = sf*iscale.get_scaling_factor(self.low_pressure_side.mass_transfer_term[t, p, j])
+                iscale.constraint_scaling_transform(c, sf)
```
TimBartholomew, line 540: Line 431 should be deleted. This scaling factor should just be based on the mass_transfer_term, not the square of the mass_transfer_term (this is multiplying the scaling factor for the mass_transfer_term for the high and low pressure side together).

TimBartholomew (dismissed):
A few minor comments

watertap/unit_models/pressure_exchanger.py
avdudchenko: I think in an earlier pull (that I removed) it was suggested this was redundant.

watertap/unit_models/pressure_exchanger.py
avdudchenko: Fixed this.

watertap/unit_models/tests/test_pressure_exchanger.py
avdudchenko: Will remove. I guess they are not needed.

watertap/unit_models/tests/test_pressure_exchanger.py
avdudchenko: Will remove.

watertap/unit_models/tests/test_pressure_exchanger.py
avdudchenko: I am leaving the high-pressure inflow mass fractions unfixed, so it can satisfy low-pressure side flow-in== high-pressure flow-out constraint (If I fix this  I get a -1DOF, as it can not balance the flow).  In the version with no mass transfer the inflow mass fractions were fixed, as the constraint ensured that both inflows were the same.

watertap/unit_models/pressure_exchanger.py
avdudchenko: Will do. That make sense.

watertap/unit_models/pressure_exchanger.py
avdudchenko: Fixed this as well.

adam-a-a:
@avdudchenko It looks like there are some merge conflicts to resolve.

adam-a-a (approved):

bknueven (approved):

bknueven:
@TimBartholomew could you approve and/or remove your change request?

adam-a-a:
It looks like @TimBartholomew 's previously requested changes still stand even though the request was removed. Thus, we can't merge yet.

bknueven:
@ksbeattie @lbianchi-lbl could one of you just use administrator privileges to merge this? I'm not sure when Tim will have a chance to look at this again.

Issue #243, opened 2021-10-21
Resolve scaling and bounds push issues


bknueven:
I've collected some thoughts below from #237. While the perturbation helper in #242 will help WaterTAP developers identify when this affects their models, we should decide a standard way forward for WaterTAP models.

# Problem Statement
WaterTAP models often have small (`< 1e-6`) lower bounds and no or much larger upper bounds. The bounds push in IPOPT, [described in the paper on p. 20](http://cepac.cheme.cmu.edu/pasilectures/biegler/ipopt.pdf), will push small initial values on variable with a small LB no or large UB to ~`1e-2` with default settings, sometimes destroying a "good" initialization. In tests I observed that pushing a small variable up a single order of magnitude (e.g., from `1e-6` to `1e-5`)  would cause issues for the RO unit model.

# Potential Solutions

## 1. Selectively apply `bounds_push` when needed
This is essentially our current strategy. With #242 it may become easier to implement, but could become an issue eventually for end-users.

## 2. Change default IPOPT options when `watertap` is imported
We could update `idaes.cfg["ipopt"]["options"]["bound_push"]` when WaterTAP is imported. A value of `1e-08` is needed to initialize the 1DRO model, but a larger global value could be chosen. We can effectively disable IPOPT's bound's pushing by setting `bound_push` small enough, which ensures a model initialized through standard IDAES methods will not have its starting point modified by IPOPT.
#### Potential Pitfalls
Ensuring a strictly feasible primal starting point is necessary for IPOPT's algorithm to work correctly. Not starting at a strict interior point could negatively affect solver performance.

## 3. Write a custom bounds pusher for WaterTAP models
To ensure we start at a strict interior point, we could write a custom bounds pusher for WaterTAP models which is more aware of small values, and then disable IPOPT's bounds pushing.
#### Potential Pitfalls
It's not clear how to incorporate this into every solve without either monkey-patching Pyomo/IDAES when `watertap` is imported or requiring an additional function call before solving a model. Perhaps we could provide our own "solver" by copying the IDAES `get_solver` strategy.

## 4. Be more careful with bounds
- Ensure our models have variable with bounds that are within an order of magnitude
- Variables are unbounded if bounds are not really necessary

## 5. Scaling small values
- Manually scaling some problematic variables in WaterTAP models to avoid small bounds
- Maintaining `user-scaling`.

andrewlee94:
[...]

Issue #258, opened 2021-11-11
Update all chemistry tests to use FpcTP state vars


aladshaw3:
Most all chemistry tests (and sub-flowsheet objects) use the 'FTPx' state variables. We want to move towards using the 'FpcTP' state variables because this scheme is more numerically stable, especially for multi-phase flows. Each set of test files should be gone through and updated to use the newer state variable definition. **This will also require changing how each problem is scaled and initialized!!!** Thus, this is a pretty large task. 

Tests to update:
- [x] test_chlorination
- [x] ~test_enrtl_water_pH~
- [x] ~test_pure_water_pH~
- [x] test_recarbonation_process
- [x] test_remineralization 
- [x] ~test_seawater_alkalinity~
- [x] ~test_water_softening~
- [x] **REMOVE** test_phoshorus_removal (made obsolete by #240)
- [x] **REMOVE** test_water_softening_Ca (redundant) 
- [x] **REMOVE** test_water_softening_Mg (redundant)
- [x] **REMOVE** test_enrtl_water_pH_edb (does nothing)

aladshaw3:
This should probably be done after #270 (since that will change how these tests are scaled)

PR #270, opened 2021-11-29
Add automatic Jacobian scaling solver

bknueven:
## Addresses #243
Part 1/N. I've taken care to make the individual commits digestible, so it may be easiest to review a single commit at a time. 

## Summary/Motivation:
This PR adds the `ipopt-watertap` solver, which is a lightweight wrapper around Pyomo's ipopt solver. This solver enables the easy simultaneous use of variable scaling while enabling Jacobian constraint scaling for constraints without an explicit scaling factor.

This PR doesn't fully address #243, only making changes to the remainder of the code base necessary for existing tests to pass. In the case of the chemistry tests, they are somewhat simplified. In general, this should eliminate the need for bound pushing everywhere in the code base, but that will be reserved for subsequent PRs (see my working branch here: https://github.com/bknueven/watertap/tree/scaled_solver_2).

### Why add a new "solver"?
The main reason to create a new solver is to encapsulate the scaling done by the IDAES Jacobian scaling function. Because `constraint_autoscale_large_jac` *sets* the scaling factor on each constraint, subsequent calls to this function on the same block are a no-op. This is sub-optimal, because the Jacobian scaling should probably be re-applied upon subsequent solves, as is done with Ipopt's Jacobian scaling. We could also *ignore* the scaling factors set on every constraint on subsequent calls, but this would also eliminate constraint scaling factors set by the user, unit model, or property package.

Therefore the "ipopt-watertap" solver *caches* the constraint scaling factors initially on the model/block, computes and applies the Jacobian scaling factors, and then resets the constraint scaling factors to their original values *after* the solve. In this way the model/block is held harmless by the "solve" operation, while enabling this solver to be used seamlessly by existing IDAES models. 

- Add "ipopt-watertap" solver
- Make the ipopt-watertap solver the default IDAES solver when watertap is imported
- Update chemistry test scaling / bound push -- ~~in two cases relative tolerances needed to be loosened~~ the phosphorus removal test needed special handling
- Update LSRRO tests initialization baselines
- Set uninitialized value in enrtl_FTPx test
- Makes Pyomo 6.2 a requirement to get Pyomo/pyomo#2121

### Legal Acknowledgement


bknueven:
@ksbeattie @lbianchi-lbl can we update the version of IDAES were testing against to IDAES/idaes-pse@main? This PR needs Pyomo 6.2, which conflicts with the specific Pyomo requirement in our current pre-release version of IDAES.

lbianchi-lbl:
@bknueven I've just synced `watertap-org/idaes-pse` to upstream and created the `1.11.0.watertap.2021.11.29` tag. just FIY I updated `setup.py` directly in this branch to trigger the CI checks, so you might want to pull this change to your local branch.

andrewlee94 (dismissed):
Everything looks OK to me, although I don't know that much about how the solver registry works.

Someone familiar with it should check the changes to the results of LSSRO model however.

watertap/core/plugins/solvers.py
```diff
@@ -0,0 +1,135 @@
...
+        if len(args) > 1 or len(args) == 0:
+            raise TypeError(f"IpoptWaterTAP.solve takes 1 positional argument but {len(args)} were given")
+        if not isinstance(args[0], (_BlockData, IBlock)):
+            raise TypeError("IpoptWaterTAP.solve takes 1 positional argument: a Pyomo ConcreteModel or Block")
+
+        self._set_user_scaling()
+
+        self._tee = kwds.get("tee", False)
+        if self._tee:
+            print("ipopt-watertap: Ipopt with user variable scaling and IDAES jacobian constraint scaling")
```
eslickj, line 52: Probably should ignore tee and use a logger instead of prints throughout.

eslickj (dismissed):
Looks good as far as I can tell.  I'd recommend logging over prints in the solver wrapper, but that's the only thing I noticed.

bknueven:
> Someone familiar with it should check the changes to the results of LSSRO model however.

I wrote this test -- the changes are just within the results from the flowsheet initialization function. The simulation and optimization results are the same within the test tolerances.

watertap/core/plugins/solvers.py
bknueven: Thanks for having a look at this @eslickj. 

I settled on the current approach so there's a single place to set output from the `solve` method, which I think is easier on the user -- in particular reporting the options set by the user in `_get_option`. That said, I'm open to reporting this at a set log-level (or changing the log-level its reported at based on the setting of `tee`).

bknueven:
> I don't know that much about how the solver registry works.

I'm not going to pretend to have a deep understanding of it, but as a result of calling the registration (which is called whenever `watertap` is imported):
```python
import watertap
import pyomo.environ as pyo

solver = pyo.SolverFactory("ipopt-watertap")
```
the variable `solver` will be an instance `IpoptWaterTAP`.

watertap/core/plugins/solvers.py
eslickj: I think that's mostly fine.  I have one concern about what is probably a rare case, and I don't even know if it's problem or not.  If someone where to redirect the solver output, would the print statements get caught?  I guess it depends on if someone just starts grabbing all of stdout or just from the solver subprocess.  I do have a utility to direct solver output to a logger, but don't remember how it works.  I'll check on it.

eslickj: Okay, it uses ```pyomo.common.tee.capture_output()```, and as far as I can tell it would capture the prints.  So I have no reason to object to the print statements.

watertap/core/plugins/solvers.py
```diff
@@ -0,0 +1,139 @@
...
+        else:
+            if self._tee:
+                print(f"ipopt-watertap: {option_name}={option_value}")
+        return option_value
+
+    def _set_user_scaling(self):
+        if "nlp_scaling_method" in self.options and self.options["nlp_scaling_method"] != "user-scaling":
+            _log.warning("The ipopt-watertap solver is designed to be run with user-scaling. "
+                    f"Provided nlp_scaling_method {self.options['nlp_scaling_method']} will be ignored")
+        self.options["nlp_scaling_method"] = "user-scaling"
```
aladshaw3: So it looks like using this new solver will prevent a user from requesting 'gradient-based' scaling, which is what IDAES generally does. 

(1) Are we all ok with this behavior? 
(2) Will we still be able to update 'bound_push' and other args as needed (even in future PRs on this)?

andrewlee94: For cases where the user wants `gradient-based` scaling, they should probably use the standard IPOPT solver instead of this one. That said, for simplicity of workflows, it might make sense to allow users to specify gradient based scaling here and just skip the additional steps (just so we don't have to explain when to use one solver over the other).

watertap/examples/chemistry/tests/test_phosphorus_removal.py
```diff
@@ -1055,13 +1052,10 @@ def test_scaling_simple_phosphorus_removal(self, simple_phosphorus_removal):
     def test_initialize_solver(self, simple_phosphorus_removal):
         model = simple_phosphorus_removal
 
-        # These might be unnecessary now with new changes in IDAES
-        model.fs.unit.control_volume.properties_out[0.0].log_mole_frac_phase_comp_true.setlb(-50)
-        model.fs.unit.control_volume.properties_out[0.0].log_mole_frac_phase_comp_true.setub(0.001)
-        model.fs.unit.control_volume.properties_out[0.0].mole_frac_phase_comp.setub(1.001)
+        # if the initialization is sufficiently good, it may be
+        # a better idea to do this manually before the solve
+        iscale.constraint_autoscale_large_jac(model.fs.unit)
```
aladshaw3, line 1057: Was this step necessary for the phosphorus removal test case? Would it not solve without it?

watertap/core/plugins/solvers.py
bknueven: @andrewlee94 that's an interesting alternative. I think we should go with it.

watertap/examples/chemistry/tests/test_phosphorus_removal.py
bknueven: Yes -- phosphorus removal is the one case that didn't "just work". You might also notice the slightly larger variable scaling. The Jacobian scaling at the initial values is much better than the resulting Jacobian scaling after the control volume initialization but before the whole unit initialization.

That said, I did not explore deeply other options, e.g., setting `bound_push` lower. It's also possible the approach for the chemistry models needs to be somewhat adjusted if we adopt this solver.

watertap/core/plugins/solvers.py
bknueven: @aladshaw3 to answer your question (2): the only options that are not passed are those directly related to `gradient-based`, which have analogous in the IDAES Jacobian autoscaling tool (`nlp_scaling_max_gradient` and `nlp_scaling_min_value`). When Ipopt isn't using `gradient-based` scaling, these options are ignored anyways.

watertap/examples/chemistry/tests/test_phosphorus_removal.py
aladshaw3: In all likely hood, if this is the only chemistry test giving issues, then I think we are fine with the new solver. 

This test can actually be removed (it is on my to do list), as it is now an obsolete representation of the phosphorus removal process, which is being tested in 'test_pH_dependent_solubility.py'

aladshaw3: @bknueven For your reference, #258 is my issue containing the list of chemistry tests that are outdated and will be removed (or updated) at some point.

watertap/core/plugins/solvers.py
eslickj: Theoretically the the gradient based scaling provided by IDAES is exactly the same as the gradient scaling that Ipopt does, so it shouldn't make a difference if you use this or gradient scaling in Ipopt.  I did some testing when the IDAES Jacobian scaling was in early development and confirmed that it was exactly the same, but probably wouldn't hurt to reconfirm that and write some tests along those lines.

What using gradient scaling in ipopt does do, is make Ipopt ignore user-provided scale factors.

aladshaw3 (dismissed):
I think these changes are fine. Most of the the chemistry tests are showing better convergence behavior and appear less sensitive to the 'bound_push' args. The tests do take longer to run, but that is from the overhead of the jacobian scaling tool done before the solves.

watertap/core/plugins/solvers.py
```diff
@@ -0,0 +1,150 @@
...
+
+        if not self._is_user_scaling():
+            self._reset_needed = False
+            super()._presolve(*args, **kwds)
+            return
+
+        if self._tee:
+            print("ipopt-watertap: Ipopt with user variable scaling and IDAES jacobian constraint scaling")
+
+        # These options are typically available with grandient-scaling, and they
```
adam-a-a:
```suggestion
        # These options are typically available with gradient-scaling, and they
```

watertap/core/plugins/solvers.py
```diff
@@ -0,0 +1,150 @@
...
+        # have corresponding options in the IDAES constraint_autoscale_large_jac
+        # function. Here we use there Ipopt names and default values, see
```
adam-a-a:
```suggestion
        # function. Here we use their Ipopt names and default values, see
```

watertap/core/plugins/solvers.py
```diff
@@ -0,0 +1,150 @@
...
+                    ignore_variable_scaling=ignore_variable_scaling,
+                    max_grad=max_grad,
+                    min_scale=min_scale)
+        except AssertionError as err:
+            if str(err) == "Error in AMPL evaluation":
+                print("ipopt-watertap: Issue in AMPL function evaluation; Jacobian constraint scaling not applied.")
+                halt_on_ampl_error = self.options.get("halt_on_ampl_error", "yes")
+                if halt_on_ampl_error == "no" :
+                    print("ipopt-watertap: halt_on_ampl_error=no, so continuing with optimization.")
+                else:
```
adam-a-a, line 88: This question is just for my understanding because I'm not sure about how the logic works here. What gets assigned to `halt_on_ampl_error` from `self.options.get("halt_on_ampl_error", "yes")`? If halt_on_ampl_error was set to no, what would be assigned to `halt_on_ampl_error`? 

Also, in the `else` clause, the error message tells the user to run with `halt_on_ampl_error` set to yes. Isn't it already set to yes? Or is the purpose of the error message just to tell the user to _also_ set `symbolic_solver_labels=True` to more clearly diagnose the issue?

watertap/core/plugins/solvers.py
```diff
@@ -0,0 +1,150 @@
...
+
+        # this creates the NL file, among other things
+        super()._presolve(*args, **kwds)
+
+    def _postsolve(self):
+        if self._reset_needed:
+            self._reset_scaling_factors()
+            # remove our reference to the model
+            del self._model
+        del self._reset_needed
```
adam-a-a, line 103: What happens if `self._reset_needed` isn't deleted?

watertap/core/plugins/solvers.py
```diff
@@ -0,0 +1,150 @@
...
+                print("The ipopt-watertap solver is designed to be run with user-scaling. "
+                        f"Ipopt with nlp_scaling_method={self.options['nlp_scaling_method']} will be used instead")
+            return False
+        return True
+
+
+## reconfigure IDAES to use the ipopt-watertap solver
+import idaes
+_default_solver_config_value = idaes.cfg.get("default_solver")
+_idaes_default_solver = _default_solver_config_value._default
```
adam-a-a, line 146: How/when does this get used?

watertap/core/plugins/solvers.py
bknueven: `halt_on_ampl_error = self.options.get("halt_on_ampl_error", "yes")` is equivalent to
```
if "halt_on_ampl_error" not in self.options:
    halt_on_ampl_error = "yes"
else:
    halt_on_ampl_error = self.options["halt_on_ampl_error"]
```

bknueven: Regarding the second point -- the default Ipopt option is `halt_on_ampl_error=no`. But given what we're doing, I think it makes sense to have a more cautions default for this solver, and if you run Ipopt with these two options it will print out the (likely) problematic constraint in a way you can use to diagnose your Pyomo model.

watertap/core/plugins/solvers.py
bknueven: This code gets called when the module is imported, which because of the various imports above happens when you first import `watertap`.

bknueven: I suppose this particular variable is not used, however. I guess there's a question here about adding an additional function (which I didn't create) to undo the step done below (resetting the default solver). In any case you could do it "by hand" by accessing this variable.

watertap/core/plugins/solvers.py
bknueven: Nothing, really. Just cleaning up things here, I suppose.

watertap/core/plugins/tests/test_solvers.py
```diff
@@ -0,0 +1,148 @@
...
+class TestIpoptWaterTAP:
+    @pytest.fixture(scope="class")
+    def m(self):
+        m = pyo.ConcreteModel()
+        m.a = pyo.Var(initialize=1)
+        m.b = b = pyo.Block()
+        b.a = pyo.Var([1,2])
+
+        m.c = pyo.Constraint(expr=(0, 1./(m.a**2) ,1))
+        b.c = pyo.Constraint([1,2], rule= lambda b,i: (i-1, b.a[i], i+2))
```
adam-a-a, line 32: Just because I'm not used to writing constraints this way, I want to make sure I understand-- is this saying first that 
0 <= b.a[1] <= 3
and then
1 <= b.a[2] <= 4 ?

bknueven: Yep! That's a compact syntax for a range constraint.

watertap/core/plugins/tests/test_solvers.py
```diff
@@ -0,0 +1,148 @@
...
+        b.d = pyo.Constraint(expr=b.a[1]**4 + b.a[2]**4 <= 4)
+        set_scaling_factor(b.d, 1e6)
```
adam-a-a, line 34: What's the difference between using this and `constraint_scaling_transform` on a constraint? As I look at most of our `calculate_scaling_factors` methods in unit models, we usually used `constraint_scaling_transform` to apply scaling factors to constraints (or at least I thought). I was originally under the impression that `set_scaling_factor` only applied to variables which I'm now seeing was an incorrect interpretation.

watertap/core/plugins/tests/test_solvers.py
```diff
@@ -0,0 +1,148 @@
...
+        cons_with_sf = list(constraints_with_scale_factor_generator(m))
+        assert cons_with_sf == [(m.b.d, 1e6)]
+
+        assert not hasattr(s, '_model')
+
+    @pytest.mark.unit
+    def test_option_absorption(self, m, s):
+        s.options['ignore_variable_scaling'] = True
+        results = s.solve(m, tee=True)
+        pyo.assert_optimal_termination(results)
```
adam-a-a, line 91: Not sure what's being tested here

watertap/core/plugins/tests/test_solvers.py
```diff
@@ -0,0 +1,148 @@
...
+        del s.options['ignore_variable_scaling']
+
+    @pytest.mark.unit
+    def test_get_option(self, s):
+        s.options['ignore_constraint_scaling'] = True
+
+        assert s._get_option('ignore_constraint_scaling', False) is True
+        assert 'ignore_constraint_scaling' not in s.options
+        assert s._get_option('ignore_constraint_scaling', False) is False
```
adam-a-a, line 100: Not clear what's happening here in the first assert. It looks like `ignore_constraint_scaling` was removed from options (from the pop?) but why assert "s._get_option('ignore_constraint_scaling', **False**)"? What would happen if you had "s._get_option('ignore_constraint_scaling', **True**)"? I am guessing that the last assert is to double-check that the option no longer exists although you did so with the second assert.

watertap/core/plugins/tests/test_solvers.py
```diff
@@ -0,0 +1,148 @@
...
+        s.options["nlp_scaling_method"] = "gradient-based"
+        s._presolve(m, tee=True)
+        assert not hasattr(s, "_scaling_cache")
+        s.options["nlp_scaling_method"] = "user-scaling"
+
+    @pytest.mark.unit
+    def test_passthrough_positive(self, m, s):
+        s.options["nlp_scaling_method"] = "gradient-based"
+        pyo.assert_optimal_termination(s.solve(m, tee=True))
+        del s.options["nlp_scaling_method"]
```
adam-a-a, line 113: Why delete the option here but in line 107 reassign to `user-scaling`?

adam-a-a: Oh, because in line 107, you don't want to delete after a presolve. In this test you're doing the full solve. Is that the reason?

watertap/core/plugins/tests/test_solvers.py
```diff
@@ -0,0 +1,148 @@
...
+
+    @pytest.mark.unit
+    def test_passthrough_negative(self, m, s):
+        s.options["nlp_scaling_method"] = "gradient-based"
+        s.options["ignore_variable_scaling"] = True
+        with pytest.raises(ApplicationError):
+            pyo.assert_optimal_termination(s.solve(m, tee=True))
```
adam-a-a, line 120: Why do we expect this to fail? Will the solve fail when variable_scaling is ignored and scaling_method is gradient-based?

watertap/core/plugins/tests/test_solvers.py
bknueven: Ipopt will exit early if an invalid option is set. This tests that the option 'ignore_variable_scaling' is trapped and deleted by `IpoptWaterTAP._presolve`.

watertap/core/plugins/tests/test_solvers.py
bknueven: The `_get_option` helper both pops and fetches the named option in the first argument, with the default specified as the second argument if the option doesn't exist. Since the `'ignore_constraint_scaling'` option is **set**, this helper should return the set value, as opposed to the default (`False`) (the first `assert`), and it should **pop** the options (the second `assert`). Now that the option no longer exists, the third assert verifies that `_get_option` returns the specified default when the option does not exist.

watertap/core/plugins/tests/test_solvers.py
```diff
@@ -0,0 +1,148 @@
...
+        m.a.value = 0
+        with pytest.raises(RuntimeError):
+            s.solve(m)
+        m.a.value = 1
+
+    @pytest.mark.unit
+    def test_presolve_ignore_AMPL_evaluation_error(self, m, s):
+        m.a.value = 0
+        s.options["halt_on_ampl_error"] = "no"
+        s._presolve(m)
```
adam-a-a, line 147: I'm guessing the test passes because all that comes of this is a printed statement, correct?

watertap/core/plugins/tests/test_solvers.py
bknueven: Probably safer to do a delete on line 107.

watertap/core/plugins/tests/test_solvers.py
bknueven: In this case, `nlp_scaling_method=gradient-based` gets you regular Ipopt, which will fail because it doesn't recognize the `ignore_variable_scaling` option. Perhaps we should get those anyways, so that they're effectively ignored.

watertap/examples/chemistry/tests/test_water_softening.py
```diff
@@ -391,9 +391,6 @@ def test_stoich_inherent(self, water_stoich):
         orig_fixed_vars = fixed_variables_set(m)
         orig_act_consts = activated_constraints_set(m)
 
-        solver.options['bound_push'] = 1e-10
-        solver.options['mu_init'] = 1e-6
-        solver.options['nlp_scaling_method'] = 'user-scaling'
```
adam-a-a, line 396: Out of curiosity, did these tests fail when keeping these now deleted options in? Or did they still pass but pass more slowly than when they were cleared away?

watertap/core/plugins/tests/test_solvers.py
bknueven: Yeah, execution isn't interrupted, just a warning printed.

adam-a-a (dismissed):
LGTM. I just had a bunch of questions/comments to make sure I understood the things that I was unsure about, simply to improve my understanding of some of the details.

adam-a-a:
@aladshaw3 Please reapprove since your review went stale (only typo corrections committed).

watertap/core/plugins/tests/test_solvers.py
bknueven: These have an interesting interplay: `constraint_scaling_transform` modifies the Pyomo constraint directly, such that after the transform, this is what the "unscaled" constraint looks like to Ipopt. You can also apply a constraint scaling factor, which will be applied by Ipopt if `nlp_scaling_method=user-scaling` for its interior point method.

In either case, whether specifying a constraint scaling factor and using `user-scaling` **or** just using the `gradient-based` scaling, the transformed constraint will be potentially re-scaled again for the solve, but critically its unscaled (to Ipopt) tolerance will be adjusted by the `constraint_scaling_transform`.

adam-a-a (approved):
Reapproving

watertap/core/plugins/tests/test_solvers.py
adam-a-a: Ah! Makes perfect sense now =)

watertap/examples/chemistry/tests/test_water_softening.py
bknueven: At this point I would need to check. I probably removed them up-front because one of the points of this change is to not need these options anymore.

Note that all the various flowsheet-tests in `examples` still pass, even though they largely also use similar options.

aladshaw3 (approved):
LGTM

PR #238, opened 2021-10-14
Make NF zero-order model compatible with generic property model

adam-a-a:
## Summary/Motivation:
Unit models should be designed to be compatible for custom-made property models as well as the IDAES generic property model framework. Thus, a seawater ion property model based on the generic property framework was created, and the zero-order nanofiltration (NF-ZO) model was modified to be compatible with said property model.

## Changes proposed in this PR:
- made a seawater ion configuration file which is fed to the GenericParameterBlock
- modified the NF-ZO model to be compatible with the seawater_ion_generic (seawater composition using generic property framework)
- made tests for the new property model and modified NF ZO model

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #278, opened 2021-12-08
Adding outline of WT database class

andrewlee94:
## Fixes/Addresses:

This PR will add a database class for interacting with libraries of parameter data for zero-order models.

## Summary/Motivation:
The Database class will be used for automating to loading of zero-order parameters from yml data files and providing these to zero-order type unit models for setting parameter values.

-
-

### Legal Acknowledgement


andrewlee94:
This still needs an example of a parameter yml file to use for testing against, but the core functionality is hopefully here. This ended up being even easier than I first thought, and should actually provide a general use class that can be used for all parameter sets (default and case-study specific) that use a standard format.

watertap/core/wt_database.py
```diff
@@ -0,0 +1,111 @@
...
+import os
+import yaml
+
+
+class Database:
+    def __init__(self, dbpath=None):
+        self._cached_files = {}
+
+        if dbpath is None:
+            self._dbpath = os.path.join("watertap", "data", "techno_economic")
```
bknueven: I think `dbpath` should probably be private, like it is here, but not in the rest of the class.

watertap/core/wt_database.py
```diff
@@ -0,0 +1,111 @@
...
+        self._cached_files = {}
+
+    def _get_technology(self, technology):
+        if technology in self.cached_files:
+            # If data is already in cached files, return
+            return self.cached_files[technology]
+        else:
+            # Else load data from required file
+            try:
+                with open(technology+".yml", "r") as f:
```
bknueven: Will need something like `os.path.join(self._dbpath, technology+".yml")`

bknueven (commented):
I think the outline looks great. I found a few things that will be evident once it's used the first time.

avdudchenko (dismissed):
Looks good to me. I don't have indepth suggestions here.

watertap/core/wt_database.py
adam-a-a: So if I'm understanding @bknueven correctly (correct me if I'm wrong), he's suggesting that all instances of `self.dbpath` be changed to `self._dbpath`.

andrewlee94: Yes - I just missed the underscored when typing this up as I didn't have a data file to test against.

watertap/core/tests/test_wt_database.py
```diff
@@ -0,0 +1,156 @@
...
+        # Check data
+        for k, v in data.items():
+            if k == "recovery":
+                assert v == "overloaded_again"
+            elif k == "new_param":
+                assert v is True
+            elif k == "new_param_2":
+                assert v is False
+            else:
+                # All other entries should match defaults
```
adam-a-a, line 148: The test is fine, but do we want to keep parameters from default that aren't necessarily included in said subtype? In this example, what if `new_param` is not a relevant parameter in `subtype2`, for example, removal of a constituent which is not applicable to the subtype? 

I suppose one way to handle this would be to assign a NaN or something like that in the yaml file which would overload `new_param` under `subtype2`.

watertap/core/wt_database.py
```diff
@@ -0,0 +1,111 @@
...
+    def get_unit_operation_parameters(self, technology, subtype=None):
+        """
+        Method to retrieve parameters for a given technology by subtype.
+
+        Args:
+            technology - unit operation technology to look up and retrieve
+                         parameters for.
+            subtype - (optional) string or list-of-strings indicating specific
+                      sub-type of technology to return parameters for. If not
+                      provided, the default parameters are used instead.
```
adam-a-a, line 48: I commented on this above, but do we necessarily want the default parameters to be used instead?

adam-a-a: Two initial ways I see to do this are 
1) in the data file, include redundant keys under subtype which are assigned a value of NaN when data isn't relevant or available for that particular parameter
2) Add an optional argument which would allow either a) use of default parameters for missing parameters (as we have it now) or b) replace those missing parameters with NaNs (or remove them entirely?)

watertap/core/tests/test_wt_database.py
andrewlee94: Note that the unit model still gets to decide which parameters to use, so it doesn't matter if we return more parameters than it needs.

adam-a-a: In that case, I guess one would just need to be careful when loading those parameters into a unit model.

andrewlee94: I did this so that a subtype only needs to contain values for parameters that change, and to have a fall back in case something isn't there.

watertap/unit_models/zero_order/nanofiltration_zo.py
```diff
@@ -0,0 +1,73 @@
...
+    CONFIG = SITOBaseData.CONFIG()
+
+    def build(self):
+        self._has_deltaP_treated = True
+        self._has_deltaP_byproduct = True
+
+        super().build()
+
+        # Add electricity consumption to model
+        self.electricity = Var(self.flowsheet().time,
```
adam-a-a, line 42: Fairly subjective, but I'd rename to `electricity_consumption`

watertap/unit_models/zero_order/nanofiltration_zo.py
```diff
@@ -0,0 +1,73 @@
...
+                               units=pyunits.kW/pyunits.hour,
+                               doc="Electricity consumption of unit")
+        self.electricity_intensity = Var(
+            units=pyunits.kW/pyunits.m**3,
```
adam-a-a: This should be kWh/m**3

watertap/unit_models/zero_order/nanofiltration_zo.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def build(self):
+        self._has_deltaP_treated = True
+        self._has_deltaP_byproduct = True
+
+        super().build()
+
+        # Add electricity consumption to model
+        self.electricity = Var(self.flowsheet().time,
+                               units=pyunits.kW/pyunits.hour,
```
adam-a-a: This _can_ be in kW to represent _power_ consumption or kWh to represent _energy_ consumption. If doing the latter, you could just multiply by hour in the constraint. However, kW/hour should not be used.

watertap/core/wt_database.py
bknueven: If the user just want to change one or two things, forcing them to specify every other parameter is cumbersome. This is a perhaps a bit more dangerous, but is still explicit. The current implementation is a pattern I've seen used by other tools: it can be cumbersome if you layer too many sub-types on top of each other, but that's an anti-pattern we guard against.

watertap/unit_models/zero_order/nanofiltration_zo.py
```diff
@@ -0,0 +1,71 @@
...
+from idaes.core import declare_process_block_class
+
+from watertap.core.zero_order_sito import SITOBaseData
+
+# Some more inforation about this module
+__author__ = "Andrew Lee"
+
+
+@declare_process_block_class("NanofiltrationZO")
+class NanofiltrationZOData(SITOBaseData):
```
adam-a-a: We do have a zero-order NF model in `unit_models` already but I'm not sure whether you intended to focus on an NF model based on the `SITOBaseData` class. Since we eventually want WT3/Proteus models to be compatible on flowsheets, it might be good to try this with the NF model that we have now (as an early first step).

watertap/unit_models/zero_order/tests/test_nanofiltration_zo.py
```diff
@@ -0,0 +1,162 @@
...
+        assert isinstance(model.fs.unit.deltaP_treated, Var)
+        assert isinstance(model.fs.unit.deltaP_byproduct, Var)
+
+        assert isinstance(model.fs.unit.electricity_consumption, Constraint)
+
+    @pytest.mark.component
+    def test_load_parameters(self, model):
+        data = model.db.get_unit_operation_parameters("nanofiltration")
+
+        model.fs.unit.load_parameters_from_database()
```
adam-a-a, line 68: This is more of a minor suggestion for a future PR, but I think it would be useful to have a method like `list_parameters_from_database` to be able to see what is going to be "loaded" and fixed within the unit model (even though one could probably just see what is inside `data` manually).

adam-a-a:
It looks like the tests are failing because the nanofiltration.yml file can't be found. Guessing that something might be going wrong with the `os.path.join`.

watertap/core/wt_database.py
bknueven:
```suggestion
            self._dbpath = os.path.join(os.path.dirname(os.path.abspath(__file__)),"..", "data", "techno_economic")
```
I think something like this needs to be the default, so the default `_dbpath` will get set correctly no matter where `watertap` happens to be installed. Note that `os.path.dirname(os.path.abspath(__file__))` should return `watertap/core`.

watertap/unit_models/zero_order/nanofiltration_zo.py
andrewlee94: The purpose where is to demonstrate the standard workflow for all models that will come from WT3, so it needs to use `SITOBaseData`. All things considered, I would suggest it would be better to look to see if the existing zero-order NF model could be replaced by this than the other way around.

adam-a-a: The two main variables that the existing zero-order NF model includes (that the SITOBase NF doesn't) are membrane area and membrane water (solvent) flux.

adam-a-a: @TimBartholomew might want to jump in and comment on anything else that the existing zero-order NF model includes that should be incorporated into the SITOBase NF model.

andrewlee94: Is membrane area used to determine performance, or estimated based on the flow of water to be treated? If it is the former, then that is becoming a predictive model (even if very simple) rather than what we are calling zero-order models here. If it is the later, then we can look at adding it but we would need an area/flow parameter or similar.

adam-a-a: So the volumetric water flux, which is the m3/s/m2 of membrane area, would be fixed, and then either the membrane area is fixed, or the volumetric recovery rate is fixed. Thus, I think this would fall under the latter case that you mentioned.

Another thing to note is the existing NF_ZO model was modified a little bit to be compatible with the generic property framework (tested with `seawater_ion_generic`). Before those modifications, the model did not work with the generic property framework as written. Just another consideration before we think about replacing the existing zero-order NF.

andrewlee94: The problem with the first point is it now means that the structure of the NF model would be different to all the other, and require a different parameter for recovery - not a show stopper but more things that need to be documented. Also, the current database on has the overall recovery parameter, not the recovery per unit area.

For the second, the new model should work for any property package that defines volumetric flow and mass based concentrations - perhaps not the most generic of properties but that is the documented foundation of the WT3 approach. If the ProteusLib zero-order NF model needed special handling then that implies you added something unusual to that model.

adam-a-a: The overall (volumetric) recovery is also in the Proteus 0-order NF model--it just has membrane water flux, a typical and often used variable, and area too. We can add those two vars to this NF model without altering recovery. All we'd do is additionally fix the membrane water flux after building. @kurbansitterley - do you think it'd be reasonable to augment the NF model like this?

On the second point, yes- I would say I did something unusual for the Proteus 0-order NF because the way it was written before, it could only handle mass flow basis, not mole, and it only detected solutes, not ions (if I am recalling correctly). If you want to have a quick look, check out PR #238 or you could look directly at the [NF_ZO file](https://github.com/adam-a-a/watertap/blob/c9c3cfee95244e719d94926c311ed00f582eee4a/watertap/unit_models/nanofiltration_ZO.py).

adam-a-a: With all that said, I think we could initially keep both versions until we make a decision on how to merge them. Ultimately, since we will eventually try to make WT3/Plib models interchangeable in flow sheets, WT3 models will need added functionality anyway, so merging capabilities of the two models makes sense (even though we'd be jumping the gun at this point).

watertap/unit_models/zero_order/tests/test_nanofiltration_zo.py
andrewlee94: This is a little trick, because as noted above we are not down-selecting the parameters we retrieve from the database (in the current API at least). I.e. when we do `load_parameters_from_database` we get removals for all components the database supports (we leave it to the specfic instance of the unit model to work out which parameters it actually cares about). So if we display the raw parameter data we would see lots of things we wouldn't actually use.

watertap/unit_models/zero_order/nanofiltration_zo.py
```diff
@@ -0,0 +1,74 @@
...
+        self.set_param_from_data(self.electricity_intensity, pdict)
+
+        for t, j in self.removal_mass_solute:
+            self.set_param_from_data(
+                self.removal_mass_solute[t, j],
+                pdict,
+                index=j,
+                use_default_removal=use_default_removal)
+
+        # TODO: Do not have parameters for these yet - need to add to datafile
```
bknueven: If we merge without addressing this, we should probably open an issue.

watertap/unit_models/zero_order/tests/test_nanofiltration_zo.py
adam-a-a: In that case, I think it would be useful to have a method where the user could instantly check which unit model parameters were fixed by the database and to what values. Technically, they could just display the model to see what is fixed along with associated values, but it would be nice to have something that explicitly shows a concise list of parameters affected by the database. So, nice to have, but something to keep in mind for a subsequent PR.

bknueven (dismissed):
Besides the open TODO, this looks great.

watertap/unit_models/zero_order/tests/test_nanofiltration_zo.py
andrewlee94: The problem with that is that each unit model would be different, so the unit model developer would need to create that method. Also, you have the issue of keeping things up to date - the values taken from the database may not be the values currently in the model as something might have changed them since. I think it would be far more valuable to show the current values of the parameters, rather than the values loaded from the database.

So, the next question is how to do that?

1. `model.display()` would show all the necessary information (and already exists), but it shows everything so the important information might get obscured.
2. We could add a method to the model which will display a redefined set of variables (basically the IDAES `model.report(0` method), but that would need additional effort from the model developer and would be limited to the pre-defined variables. If a user needed to specify other parameters for their case (e.g. they fixed an outlet concentration rather than a removal fraction), that might not show up. 

So maybe the answer is a combination of the above - the `report()` method for common cases, and a fall back to `display()` if you need more information.

watertap/unit_models/zero_order/nanofiltration_zo.py
andrewlee94: My first response is that we should get parameter values for this in the `.yml` file before merging this (as this is going to be the example that we use for doing all the subsequent units).

watertap/unit_models/zero_order/nanofiltration_zo.py
kurbansitterley: @adam-a-a I don't see anything wrong with this augmentation to the NF model, especially considering your comment on needing to add functionality to WT3 models.

watertap/unit_models/zero_order/tests/test_nanofiltration_zo.py
adam-a-a: My main thought was that upon calling `load_parameters_from_database`, the user should be able to immediately see what was fixed at that moment. If they change fixed values afterwards, that'd be up to them to check via `display()` or a `report()` method (which we'd have to add in). 

However, I think it would be good to add one more optional argument to `load_parameters_from_database` that passes to `set_param_from_data`. The optional arg would be boolean and would either (1) print out which parameters were fixed and to what values/units if true or (2) display nothing if false.

watertap/unit_models/zero_order/tests/test_nanofiltration_zo.py
```diff
@@ -0,0 +1,301 @@
...
+
+        m.fs.unit = NanofiltrationZO(default={
+            "property_package": m.fs.params,
+            "database": m.db})
+
+        m.fs.unit.inlet.flow_vol.fix(10)
+        m.fs.unit.inlet.conc_mass_comp[0, "sulfur"].fix(1)
+        m.fs.unit.inlet.conc_mass_comp[0, "toc"].fix(2)
+        m.fs.unit.inlet.conc_mass_comp[0, "tss"].fix(3)
+        m.fs.unit.inlet.pressure.fix(1e5)
```
adam-a-a: There's a problem here: we're fixing volumetric flow rate in (pump flow) and pressure (pump pressure), and then we fix "electricity_intensity" after loading parameters from the database. "electricity_intensity" in this case should be equal to flow_vol * pressure / pump_efficiency / flow_vol multiplied by a conversion factor to convert the whole thing to kWh/m3. I see some issues:

(1) The fixed value for electricity_intensity doesn't equal the calculation mentioned above. The NF unit model should have a constraint linking electricity_intensity to flow_vol in and pressure in. 
(2) Pump efficiency should probably be added to the unit model as a parameter, unless we're just going to assume 100% pump efficiency all the time and don't want to allow flexibility to change that.
(3) Later on, when we want to be able to swap WT3 and Plib models, we'd have to be careful since WT3 models, more specifically, this NF model based on the WT3 approach, rolls the energy required for pumping into the unit model. Meanwhile, if we were using Plib's NF model, we'd add a pump unit model to the flowsheet before the NF model and compute the energy requirement for the pump separately. Thus, say the user wanted to swap out the WT3 version of NF from a flowsheet containing several unit models and replace it with a Plib NF model. They'd have to know/remember that energy consumption is computed differently between the two and would not be able to just swap NF models; they'd need to add the corresponding feed pump with the NF model. Looking at it the other way around, if they wanted to replace a Plib NF model on a flowsheet (which would also include the pump unit model upstream of that NF, if setup correctly), swapping the Plib NF for a WT3 NF without removing the pump from the flowsheet would lead to double-counting the energy requirement. This opens up more room for error, so we'll have to keep this in mind moving forward and decide on how to remedy this, whether through good documentation or by altering the way in which we are implementing these models.

andrewlee94: Remember that each unit can decide what parameters are important to it, and thus that the `yml` files do not all need to look the same. - that means that for pumps we would likely not have an energy intensity parameter, but rather an efficiency parameter instead. Additionally, a pump is a single inlet-single outlet unit, so it does not make sense to use the SITO base class for it or to define recovery and removal fractions.

adam-a-a: That makes sense, but we still need to address points 1 - 2 mentioned above. Furthermore, I still think we'll need to be careful when it comes time to swap these SITO unit models with Proteus unit models for the reasons mentioned. Right now, it looks like the user would exclude an IDAES pump (pressure exchanger) from the flowsheet and only include the SITO nanofiltration model; if they wanted to transition to a more detailed NF model, they'd need to make sure to add the IDAES pump unit explicitly in addition to replacing the SITO NF unit with the more detailed NF unit.

andrewlee94: I am not sure we need that, and going down that route might add more complexity than we want. For the NF case (and others like it), "energy intensity" is essentially an aggregate term for the efficiency of the pump(s) and the defined pressure drop across the unit - so points 1 and 2 are implicitly captured there already. Note that as energy intensity has units of power/flow, the flow effects are already accounted for.

The alternative would be to remove energy intensity entirely and instead require a pump efficiency and a pressure drop be defined for these types of units, and to then do a proper pump calculation for each. That is not hard to do, but it is starting to move us away from the original "zero order" concept where there were just aggregate parameters for everything.

andrewlee94: Additionally, this would mean that different types of units would need different energy calculations, adding more differences in model structure. E.g. in a stirred tank the energy requirement is not related to the pressure drop.

adam-a-a: Right- the energy intensity would just be pump pressure * pump flow / pump efficiency/ pump flow (assuming pump efficiency encapsulates the overall efficiency of motor and pump). However, when you are fixing energy intensity to some value, it needs to match up with the applied pressure, which isn't the case in the test shown here. So I think energy intensity should not be a loaded parameter since we will be fixing pressure to some value that may not match up with that loaded value. The nanofiltration model should probably have a constraint/expression for energy intensity.

watertap/unit_models/zero_order/nanofiltration_zo.py
TimBartholomew: @adam-a-a this base class is for zero-order models that are similar to each other (hopefully most models in WT3), I don't think we can make a class that covers all zero-order models and the ProteusLib zero-order nanofiltration model is a good example of that. That model has a variety of things that are specific to membrane units, like flux and membrane area. We do not want those membrane specific things to be in a generic base class that we want to apply to all different types of technologies. We could have the ProtuesLib zero-order nanofiltration model inherit from this base class and add the membrane specific parameters, a key point I think you made though is if we are locking ourselves into a specific basis (i.e. mass).

@andrewlee94 are there any issues with your current class to providing flexibility for the basis (e.g. mass vs molar)?

watertap/unit_models/zero_order/tests/test_nanofiltration_zo.py
adam-a-a: Another approach if we want to keep energy intensity in the yaml file-- add the default feed pressure associated with the energy intensity to the yaml file, which would then be fixed to the inlet pressure.

watertap/data/techno_economic/nanofiltration.yml
```diff
@@ -0,0 +1,92 @@
...
+      value: 0.97
+    ultraviolet_transmittance_uvt:
+      constituent_longform: Ultraviolet Transmittance (UVT)
+      units: '%'
+      value: 90.0
+  electricity_intensity:
+    units: kWh/m^3
+    value: 0.231344952
+  recovery_vol:
+    reference: Based on Vogtle case study data. No costs for WT3 treatment.
```
adam-a-a:
```suggestion
```

watertap/unit_models/zero_order/tests/test_nanofiltration_zo.py
andrewlee94: There is a catch there - the pump is included within the unit for these cases, so the inlet pressure is not equal to the pump pressure (otherwise the pump would have to be external to the unit). In fact, that raises question about exactly what "pressure drop" means in these types of models as most of the pressure changes occur within the bounds of the unit.

andrewlee94: However, yes, pressure drop (in terms of the difference between inlet and outlet state) needs to be part of the yml file.

watertap/data/techno_economic/nanofiltration.yml
```diff
@@ -0,0 +1,92 @@
...
+      value: 0.97
+    toc:
+      constituent_longform: Total Organic Carbon (TOC)
+      units: kg/kg
+      value: 0.75
+    tss:
+      constituent_longform: Total Suspended Solids (TSS)
+      units: kg/kg
+      value: 0.97
+    ultraviolet_transmittance_uvt:
```
adam-a-a: UV transmittance is more of a feedwater characteristic--not really something that should be under removal_mass_solute

watertap/unit_models/zero_order/tests/test_nanofiltration_zo.py
TimBartholomew: @adam-a-a @andrewlee94 - Just a thought on this, there could be an optional boolean argument that if true prints out the variable and its fixed value whenever it is fixed within `load_parameters_from_database`. I think this implementation would be easy, wherever the variable would be fixed it could use a function that has that boolean argument (like verbose or something) that prints the variable, its value, and fixes the variable.

watertap/unit_models/zero_order/nanofiltration_zo.py
andrewlee94: @TimBartholomew Whilst I think there are no fundamental issues with switching between mass and mole basis, we would need to be careful to ensure that the two do not get mixed up. Currently, I believe everything is done on a mass basis to keep things simple.

Note that the concentration basis is basically determined by the feedwater definition (a yet to be added yml file and Feed block), whilst the removal basis is defined by the unit-specific yml file. We are already allowing for people to mix-and-match databases if desired, so we would need to have checks for the flow basis if we want to allow for both mass and mole bases.

watertap/data/techno_economic/nanofiltration.yml
andrewlee94: Any reason why you think this should be removed? Note that I generally did not make changes to the contents of the yml file.

watertap/data/techno_economic/nanofiltration.yml
andrewlee94: There was a discussion about what to do with this, and I think the final conclusion was to remove it. However, the WT3 team should confirm that.

watertap/data/techno_economic/nanofiltration.yml
adam-a-a: It isn't really relevant. We're just assuming a recovery rate of 85%, so the note on "costs for WT3 treatment" doesn't need to be here. Unless @kurbansitterley says that the recovery rate value was taken from this Vogtle study which provides further insight on underlying assumptions for NF performance parameters included here in the yaml (even then, 85% is a valid recovery rate that can just be assumed for NF), it doesn't need to be here since we aren't providing references for the other parameters.

watertap/unit_models/zero_order/tests/test_nanofiltration_zo.py
TimBartholomew: @adam-a-a As you said, WT3 models generally did not separate out pumps, so their nanofiltration model includes the pump. We should keep it at that level. It would be more confusing if the user had to add a pump before nanofiltration, but they have no idea what the pressure should be fixed to (this is a zero order model so that should be specified, see paragraph below) and since the pump isn't a part of the nanofiltration model we don't have a path to specify that pump as a nanofiltration pump. I don't think we want to separate out the pumps just so that users know if they change models they need to add an IDAES pump, especially if the user isn't supposed to change how the pump operates.

I think WT3 NF model has to be for a specific pressure, otherwise it would be important to include how that changes the flux/membrane area/capital cost. So the pressure shouldn't be a degree of freedom for the user or the model and should just be unlinked to the energy consumption. If the nanofiltration model was standalone (no other units), then the pressures should be stale variables (i.e. uninvolved in any constraints). If the nanofiltration model is connected to other models, the inlet pressure should probably be fixed to atmospheric (otherwise the energy consumption data would be off), the outlet permeate pressure should probably be fixed to atmospheric (since that should be the default for filtration processes), the outlet brine pressure should be fixed to whatever value is specific to the data if we have it (it's likely WT3 doesn't have the information, so maybe atmospheric is the best value and our documentation would just need to warn users that the outlet pressure is not accurate for downstream analysis)

adam-a-a: The inlet pressure of the membrane module is equal to the pump discharge pressure (which is what I meant by "pump pressure", but really the "pump pressure" in the energy calculation is the added pump head, i.e., pump discharge head - pump inlet head). The pump inlet pressure is typically atmospheric while the discharge pressure is whatever pressure is needed to overcome osmotic pressure at all points along the membrane module.

The pressure drop in the case of NF is the drop across the membrane module, which is only important for ensuring that the applied pressure at the module outlet (after the pressure drop) is still high enough to overcome osmotic pressure at the outlet to maintain permeate flux and ensure that there is no flux reversal.

TimBartholomew:
@andrewlee94 this is a minor comment but I think we should rename SITO (single inlet two outlet) to either OITO (one inlet two outlet) or SIDO (single inlet double outlet) to be consistent with the name of the numbering. In a comment above you mentioned a pump would be single inlet single outlet (SISO), so maybe SIDO makes the most sense. This would also standardize the double inlet double outlet (DIDO) model name as well.

watertap/unit_models/zero_order/tests/test_nanofiltration_zo.py
adam-a-a: @TimBartholomew I agree- that was my thought as well. This way, users could be more informed about what's changing in the model when they apply `load_parameters_from_database`.

watertap/data/techno_economic/nanofiltration.yml
adam-a-a: Alternatively, if that Vogtle study was used for all or several of the parameter values in the yaml, we could just have a higher-level reference indicating that.

watertap/unit_models/zero_order/tests/test_nanofiltration_zo.py
andrewlee94: I have added functionality for this, but rather than add a new argument I have set it up so that all calls to `set_param_from_data` record the action in the logger at the `info_high` level. This way, users can control whether they see this information by setting the logger level.

andrewlee94:
@TimBartholomew I renamed the class as you suggested.

docs/technical_reference/core/zero_order_unit.rst
```diff
@@ -36,24 +36,24 @@ Usage
...
+The SIDO base class constructs a simple representation of unit operation with a single inlet (named `inlet`) and two outlets (named `treated` and `byproduct`). A `StateBlock` is constructed for the inlet and each outlet with a `Port` associated with each of these.
 
 Property Package Requirements
 -----------------------------
 
-The SITO base class makes a number of assumptions about the structure of the associated property package, and contains a number of checks to ensure the property package meets these requirements. The requirements for the property package are:
+The SIDO base class makes a number of assumptions about the structure of the associated property package, and contains a number of checks to ensure the property package meets these requirements. The requirements for the property package are:
 
 1. A single phase named `Liq`.
 2. A single `Solvent` species named `H2O` (package must define a `solvent_set` with length that contains "H2O").
```
bknueven:
```suggestion
2. A single `Solvent` species named `H2O` (package must define a `solvent_set` with length one that contains "H2O").
```

watertap/core/zero_order_sido.py
```diff
@@ -359,32 +312,67 @@ def calculate_scaling_factors(self):
...
+            {"Inlet": self.inlet,
+             "Treated": self.treated,
+             "Byproduct": self.byproduct},
+            time_point=time_point)
+
+    def _get_performance_contents(self, time_point=0):
+        var_dict = {"Water Recovery": self.recovery_vol[time_point]}
+        for (t, j), v in self.removal_mass_solute.items():
+            if t == time_point:
+                var_dict[f"Solute Removal [{j}]"] = v
```
bknueven: Alternatively, using slices:
```suggestion
        for j, v in self.removal_mass_solute[time_point,:].wildcard_items():
            var_dict[f"Solute Removal [{j}]"] = v
```

watertap/unit_models/zero_order/nanofiltration_zo.py
```diff
@@ -0,0 +1,75 @@
...
+
+        self.set_param_from_data(self.recovery_vol, pdict)
+        self.set_param_from_data(self.electricity_intensity, pdict)
+
+        for t, j in self.removal_mass_solute:
+            self.set_param_from_data(
+                self.removal_mass_solute[t, j],
+                pdict,
+                index=j,
+                use_default_removal=use_default_removal)
```
bknueven: `recovery_vol` and `removal_mass_solute` are defined in the base class. It makes me wonder if some of this couldn't be moved up a level, given that more than half these lines would seemingly need to appear in every implementation of this method.

bknueven (dismissed):
I found a typo in untouched documentation, and had a few optional code suggestions. Nothing to hold this up.

watertap/unit_models/zero_order/nanofiltration_zo.py
```diff
@@ -72,3 +72,11 @@ def load_parameters_from_database(self, use_default_removal=False):
         # TODO: Do not have parameters for these yet - need to add to datafile
         self.deltaP_treated.fix(0)
         self.deltaP_byproduct.fix(0)
+
+    def _get_performance_contents(self, time_point=0):
+        perf_dict = super()._get_performance_contents(time_point=0)
```
adam-a-a: Should this just be `time_point` instead of `time_point=0`?

andrewlee94: No, we want it to default to time=0 so that people doing steady-state models (most of them) don't need to worry about this argument. This is only there to support dynamic cases (and even then only allows you to pick a single time point).

adam-a-a: Right, setting the default for time_point to 0 supports steady-state models so users don't have to worry about setting this unless they're working with dynamic models--but I was referring to 
`perf_dict = super()._get_performance_contents(time_point=0)`

I figured since time_point is already set to 0 in the function call, you don't need this, and if the user wanted some other time_point, say, time_point=2, I would think we'd want to call super on that same time_point of 2, not 0. Or are you saying that regardless of the time_point of interest, `super()._get_performance_contents` should always use `time_point=0` for whatever reason?

andrewlee94: Ah - missed that. I saw the first line in the GitHub snippet and not the second.

watertap/unit_models/zero_order/nanofiltration_zo.py
```diff
@@ -72,3 +72,11 @@ def load_parameters_from_database(self, use_default_removal=False):
...
+
+        perf_dict["vars"]["Electricity Demand"] = self.electricity[time_point]
+        perf_dict["vars"]["Electricity Intensity"] = self.electricity_intensity
```
adam-a-a, line 78: There might be better alternatives for this, but should we consider adding units to the reported variable names for clarity? Perhaps there should eventually be another column for units in the report table. Just something that I wanted to point out for a future PR, and I think (not 100% sure) I saw an issue or PR somewhere on IDAES that mentions adding units to reports.

andrewlee94: That is a something we want to do in IDAES but have not yet had time for. Either way, that would be an IDAES issue and not WaterTap, as we are relying on IDAES' reporting functionality for this.

watertap/core/zero_order_properties.py
```diff
@@ -12,7 +12,8 @@
 ###############################################################################
 """
 This module contains the general purpose property package for zero-order
-unit models.
+unit models. Zero-order models do not track temeprature and pressure, or any
```
adam-a-a:
```suggestion
unit models. Zero-order models do not track temperature and pressure, or any
```

adam-a-a (approved):
LGTM

watertap/core/zero_order_sido.py
```diff
@@ -359,32 +312,114 @@ def calculate_scaling_factors(self):
...
+            units = getattr(pyunits, pdata["units"])
+        except KeyError:
+            raise KeyError(
+                f"{self.name} - no units provided for {pname} (index: "
+                f"{index}) in database.")
+
+        parameter.fix(val*units)
+        _log.info_high(f"{parameter.name} fixed to value {val} {str(units)}")
+
+    def set_recovery_and_removal(self, data, use_default_removal=False):
```
bknueven, line 390: Should this method be "private", since it's only to be called from derived classes?

bknueven (approved):
I left a question, but I think the answer is a matter of taste.

watertap/core/zero_order_sido.py
andrewlee94: I originally made it private, but made it public so as not to discourage developers of derived classes from using it (less-experienced users might avoid using any private methods).

bknueven: Good point -- I never let an underscore stop me, but that's not everyone

PR #283, opened 2021-12-14
Aligned RO Initialization and Scaling

bknueven:
## Fixes/Addresses:
Working towards a "membrane base" class to generalize our RO models.

## Summary/Motivation:
We would like to further combine/generalize the 0D and 1D RO models. This PR attempts to unify, to the extent possible, the initialization and scaling between the two models. This will enable moving more functionality into the `_ReverseOsmosisBaseData` class in a subsequent PR. In light of #270, this PR also undoes some the changes to the NaCl property package and RO unit models introduced by #237.

- Mostly reset NaCl scaling to before #237
- Lower `pressure` bound in NaCl property package (when `pressure` is scaled, 1 bar gets "bound push-ed")
- Remove ~most~ constraint scaling from 0D & 1D models
- Remove "magic constants" from 0D & 1D scaling
- Unify variable ~and remaining constraint~ scaling in 0D & 1D RO models
- 1DRO initialization utilizes 0DRO initialization logic -- improves performance even though mapping isn't perfect

### Legal Acknowledgement


watertap/unit_models/reverse_osmosis_0D.py
```diff
@@ -890,154 +853,75 @@ def calculate_scaling_factors(self):
...
-            for ind, c in self.eq_friction_factor_darcy_io.items():
-                sf = iscale.get_scaling_factor(self.friction_factor_darcy_io[ind])
-                iscale.constraint_scaling_transform(c, sf/10.)
+            for (t, x), c in self.eq_velocity_io.items():
+                if x == 'in':
+                    prop_io = self.feed_side.properties_in[t]
+                elif x == 'out':
+                    prop_io = self.feed_side.properties_out[t]
+                sf = iscale.get_scaling_factor(prop_io.flow_vol_phase['Liq'])
+                iscale.constraint_scaling_transform(c, sf)
```
adam-a-a: I was under the impression that with the new solver, we generally wouldn't need to apply `constraint_scaling_transform`. Was the inclusion necessary, or did you keep the transforms in since they were already there?

bknueven: So for the time being it seems necessary. When I eliminate the calls to `constraint_scaling_transform` from the RO models I get two test failures in the full_treatment_train folder:
```
========================================================================= short test summary info =========================================================================
FAILED watertap/examples/flowsheets/full_treatment_train/analysis/tests/test_flowsheets.py::test_flowsheet_two_stage - RuntimeError: The solver failed to converge to an...
FAILED watertap/examples/flowsheets/full_treatment_train/analysis/tests/test_flowsheets.py::test_flowsheet_1DRO - RuntimeError: The solver failed to converge to an opti...
FAILED watertap/property_models/tests/test_NaCl_prop_pack.py::TestNaClProperty_idaes::test_initialize_failure - Failed: DID NOT RAISE <class 'idaes.core.util.exceptions...
FAILED watertap/property_models/tests/test_seawater_prop_pack.py::TestSeawaterProperty_idaes::test_initialize_failure - Failed: DID NOT RAISE <class 'idaes.core.util.ex...
========================================================== 4 failed, 531 passed, 8 skipped in 236.68s (0:03:56) ===========================================================
```

Interestingly, as an experiment I merged #285 into this branch locally, I did not see these failures (and I think the four I see are not of much concern):
```
========================================================================= short test summary info =========================================================================
FAILED watertap/examples/flowsheets/lsrro/tests/test_lsrro.py::TestLSRRO_2Stage::test_optimize - assert 0.00045104115267494333 == 0.000451048 ± 4.5e-09
FAILED watertap/property_models/tests/test_NaCl_prop_pack.py::TestNaClProperty_idaes::test_initialize_failure - Failed: DID NOT RAISE <class 'idaes.core.util.exceptions...
FAILED watertap/property_models/tests/test_seawater_prop_pack.py::TestSeawaterProperty_idaes::test_initialize_failure - Failed: DID NOT RAISE <class 'idaes.core.util.ex...
FAILED watertap/unit_models/zero_order/tests/test_nanofiltration_zo.py::TestNFZO_w_default_removal::test_report - AssertionError: assert '\n============================...
========================================================== 4 failed, 531 passed, 8 skipped in 219.25s (0:03:39) ===========================================================
```

Note that we still may need to transform *some* constraints, though ideally those can be kept to the property packages. The new solver should make this the exception rather than the rule, however. I also don't think I spent much time finding a minimal set of constraints that needed transforming on the RO side.

watertap/unit_models/tests/test_reverse_osmosis_0D.py
```diff
@@ -311,12 +311,7 @@ def test_calculate_scaling(self, RO_frame):
         # check that all variables have scaling factors
         unscaled_var_list = list(unscaled_variables_generator(m))
         assert len(unscaled_var_list) == 0
-        # check that all constraints have been scaled
-        unscaled_constraint_list = list(unscaled_constraints_generator(m))
```
adam-a-a, line 315: @bknueven since some constraints are still being scaled, should we include those in the test to verify that they are being scaled? That way, if changes are made later on, we'd have an additional layer of tracking which constraints typically benefit from scaling and which can be left alone. It might be more troublesome than worthwhile though. What do you think?

adam-a-a (dismissed):
Some minor comments but LGTM

watertap/unit_models/tests/test_reverse_osmosis_0D.py
bknueven: I guess I'm still open to being more selective on scaling the constraints. If #285 goes before this PR, scaling the constraints isn't strictly necessary (at least for the tests to pass).

adam-a-a (dismissed):
LGTM

watertap/property_models/NaCl_prop_pack.py
```diff
@@ -741,49 +741,34 @@ def calculate_scaling_factors(self):
...
-                v = getattr(self, v_str)
-                sf = iscale.get_scaling_factor(v, default=1, warning=True)
-                c = getattr(self, 'eq_' + v_str)
-                iscale.constraint_scaling_transform(c, sf)
+        if self.is_property_constructed('pressure_osm'):
+            sf = iscale.get_scaling_factor(self.pressure_osm, default=1, warning=True)
+            iscale.constraint_scaling_transform(self.eq_pressure_osm, sf)
+        if self.is_property_constructed('osm_coeff'):
+            sf = iscale.get_scaling_factor(self.mass_frac_phase_comp['Liq', 'NaCl'], default=1, warning=True)
+            iscale.constraint_scaling_transform(self.eq_osm_coeff, sf**1.25)
```
TimBartholomew: This looks really odd to me. The osmotic coefficient should really just have a scaling factor of 1, the value ranges from 0.8 to 1.3. It shouldn't be tied to the scaling factor of the mass fraction of NaCl, and it shouldn't be raised to a power of 1.25.

watertap/property_models/NaCl_prop_pack.py
```diff
@@ -741,49 +741,34 @@ def calculate_scaling_factors(self):
...
 
-        for v_str in ('dens_mass_phase', 'enth_mass_phase'):
-            if self.is_property_constructed(v_str):
-                v = getattr(self, v_str)
-                sf = iscale.get_scaling_factor(v['Liq'], default=1, warning=True)
-                c = getattr(self, 'eq_' + v_str)
-                iscale.constraint_scaling_transform(c, sf*10.)
+        if self.is_property_constructed('dens_mass_phase'):
+            sf = iscale.get_scaling_factor(self.mass_frac_phase_comp['Liq', 'NaCl'])
+            iscale.constraint_scaling_transform(self.eq_dens_mass_phase, sf/self.params.dens_mass_param['1'])
```
TimBartholomew: This looks odd to me. I'm not sure why the density is tied to the scaling factor for the NaCl mass fraction. It should just be `1/self.params.dens_mass_param['0']` (note the parameter is indexed by 0 rather than 1). 

Oh wait its easier than that, we have a default scaling factor being set for the dens_mass_phase variable. So the scaling factor can be:
sf = iscale.get_scaling_factor(self.dens_mass_phase['Liq'])

watertap/property_models/NaCl_prop_pack.py
```diff
@@ -741,49 +741,34 @@ def calculate_scaling_factors(self):
...
 
         # property relationship indexed by component
-        v_str_lst_comp = ['molality_comp']
-        for v_str in v_str_lst_comp:
```
TimBartholomew, line 772: I created these blocks to transform similar constraints to make it easier on the developer. I know in this case there is only one property that is indexed by component, but possibly in the future a developer might add another and it would be easier for them to just add to this list rather than create the 4 lines of code. It's not a big deal though and we can get rid of it, but that was my intention.

TimBartholomew (changes requested):

watertap/property_models/NaCl_prop_pack.py
TimBartholomew: Oh since this is a constraint, I think we can just get rid of this because a scaling factor of 1 should be fine.

TimBartholomew (dismissed):

adam-a-a (dismissed):

TimBartholomew (approved):

adam-a-a (approved):

PR #282, opened 2021-12-10
Excise bound_push Part 1

bknueven:
## Addresses #243 
Part 2/3. Follow-up from #270.

## Summary/Motivation:
Removes `bound_push` Ipopt option from the codebase, *except* in the full_treatment_train directory.

## Changes proposed in this PR:
- Remove `bound_push` from unit model initialization and test
- Remove `bound_push` from example RO w/ ERD and LSRRO flowsheets

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #284, opened 2021-12-17
Updating the chemistry tests and docs

aladshaw3:
## Fixes/Addresses:

#258 & #272

Several files are touched in this PR. Most of this is changes to documentation or the removal of outdated/redundant examples.

## Summary/Motivation:

PR is to remove outdated tests and update some tests with FpcTP state vars. 

## Changes proposed in this PR:
- Removal of outdated/redundant tests
- Updating methods used in tests
- Updating documentation to reflect new chemistry scaling standards
- Cleaning up docs
- Removing multiple duplicate flowsheet examples under the chemistry folder

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #285, opened 2022-01-04
Excise bound push pt 2

### LINKED ISSUE #243: Resolve scaling and bounds push issues

bknueven:
## Fixes #243
Part 3/3. Follow-up from #270 and #282.

## Summary/Motivation:
Finish removing `bound_push` and manual Jacobian scaling from the codebase

- Replace `solve_with_user_scaling` in full_treatment_train folder with `solve_block`
- Update chemistry examples in full_treatment_train folder to align with those in the examples folder
- Tighten default `watertap-ipopt` options numerical tolerances. We were often "getting lucky" with the IDAES default `tol` of `1e-06`.
   -  `tol` to Ipopt default `1e-08` (_scaled_ NLP error)
   - `constr_viol_tol` to `1e-08` (_unscaled_ feasibility tolerance, Ipopt default `1e-04`) 

### Legal Acknowledgement


bknueven:
Converting this to draft for now: #284 should go first, and will create merge conflicts.

watertap/examples/chemistry/tests/test_pure_water_pH.py
```diff
@@ -142,6 +142,7 @@ def base_units():
 def solver():
     s = get_solver()
     s.options["max_iter"] = 200
+    s.options["tol"] = 1e-12
     return s
```
aladshaw3: Was it necessary to manually change tolerance for this test?

watertap/examples/flowsheets/full_treatment_train/chemical_flowsheet_util.py
```diff
@@ -58,26 +58,11 @@ def unfix_all_molefractions(port):
...
+    if tee:
         print(blk)
     results = solver.solve(blk, tee=tee)
 
-    # This is a temporary fix to get around some errors in RO with detailed level
-    try:
-        iscale.constraint_autoscale_large_jac(blk)
-    except:
-        pass
-
```
aladshaw3, line 69: Glad to see that this will not be necessary anymore.

watertap/examples/flowsheets/full_treatment_train/electrolyte_scaling_utils.py
```diff
@@ -190,31 +190,5 @@ def calculate_chemical_scaling_factors(unit, thermo_params, rxn_params, state_ar
...
-        print("Unscaled constraints:")
-        for c in iscale.unscaled_constraints_generator(unit):
-            print(f"    {c}")
-        print("Scaled constraints by factor:")
-        for c, s in iscale.constraints_with_scale_factor_generator(unit):
-            print(f"    {c}, {s}")
-        print("Badly scaled variables:")
-        for v, sv in iscale.badly_scaled_var_generator(unit, large=1e2, small=1e-2, zero=1e-12):
-            print(f"    {v} -- {sv} -- {iscale.get_scaling_factor(v)}")
-        print(f"Jacobian Condition Number: {iscale.jacobian_cond(jac=jac):.2e}")
```
aladshaw3: Much of this is removed already in the pending PR #284 (there will be many conflicts after 284 is merged)

watertap/examples/flowsheets/full_treatment_train/flowsheet_components/chemistry/PostTreatment_SimpleNaOCl_Chlorination.py
```diff
@@ -54,7 +54,7 @@
 from pyomo.util.check_units import assert_units_consistent
```
aladshaw3: This file (PostTreatment_SimpleNaOCl_Chlorination.py) is removed by PR #284 (redundant and unused in full flowsheet)

watertap/examples/flowsheets/full_treatment_train/flowsheet_components/chemistry/PreTreatment_Simple_Softening.py
```diff
@@ -46,7 +46,7 @@
 # Import pyomo methods to check the system units
```
aladshaw3: This file (PreTreatment_Simple_Softening.py) is removed by PR #284 (redundant and unused in full flowsheet)

watertap/examples/flowsheets/full_treatment_train/flowsheet_components/chemistry/SepRO_plus_Chlorination.py
```diff
@@ -70,7 +70,7 @@
 from pyomo.network import Arc
 
 from watertap.examples.flowsheets.full_treatment_train.util import (
-    solve_with_user_scaling,
```
aladshaw3: This file (SepRO_plus_Chlorination.py) is removed by PR #284 (redundant and unused in full flowsheet)

watertap/examples/flowsheets/full_treatment_train/flowsheet_components/chemistry/ZeroDRO_plus_Chlorination.py
```diff
@@ -71,7 +71,7 @@
 from idaes.generic_models.unit_models.translator import Translator
 from pyomo.network import Arc
 
-from watertap.examples.flowsheets.full_treatment_train.util import solve_with_user_scaling, check_dof
```
aladshaw3: This file (ZeroDRO_plus_Chlorination.py) is removed by PR #284 (redundant and unused in full flowsheet)

aladshaw3 (commented):
I will likely need to review this again because you will have a number of conflicts to resolve after PR #284. Everything looks good to me (as far as the chemistry stuff is concerned). Just a few comments/questions below.

watertap/examples/chemistry/tests/test_pure_water_pH.py
bknueven: It was, but I think I've found a global set of Ipopt options that work.

This one was strange -- a `tol` of `1e-06` "works" (the tests pass). However, smaller values (e.g., `1e-07`, `1e-09`, would cause regression test failures. Looking at the log, Ipopt happens to get lucky and with `1e-06` solves to a tighter tolerance than `1e-09` due to when the solver chooses to kick out of restoration.

bknueven:
@andrewlee94 would you might having a look at the single failing test to see if the baseline should just be updated or if the difference is significant?

aladshaw3:
@bknueven 

I think @adam-a-a or @TimBartholomew should be the ones to have a look at that assertion error, since it is in the NF model result.

adam-a-a:
> @bknueven
> 
> I think @adam-a-a or @TimBartholomew should be the ones to have a look at that assertion error, since it is in the NF model result.

Actually, @andrewlee94 would be best suited to check this out since the assertion error is attributed to one of the tests for the new zero-order (WT3) NF model; the error has to do with a mismatch in the captured output after calling the report() method.

andrewlee94:
It looks like changing the solver options changed to results of the zero-order models too, and significantly enough that the printed output in the `report` method changed. Seeing as the `report` method was looking for an expected string output (hard-coded in the test), this caused the failure.

The fix is to either update the hard coded string, or to write up something that will construct the expected string from the model output for testing.

bknueven:
Thanks @andrewlee94. After looking in the other tests, you compare the changed reported value to zero. With the new solver options, the report now shows this value as `1.0164e-20`, whereas before it was `9.9000e-07`.

watertap/examples/flowsheets/full_treatment_train/flowsheet_components/chemistry/posttreatment_ideal_naocl_chlorination_block.py
```diff
@@ -695,14 +680,10 @@ def run_ideal_naocl_mixer_example(fixed_dosage=False):
 
     check_dof(model)
 
-    # scale the mixer
-    scale_ideal_naocl_mixer(model.fs.ideal_naocl_mixer_unit)
```
adam-a-a, line 699: Why omit scaling of the mixer?

adam-a-a: Ah, because it just uses `iscale.constraint_autoscale_large_jac()` which is being eliminated from all code in this PR.

watertap/examples/flowsheets/full_treatment_train/flowsheet_components/examples/full_example.py
```diff
@@ -206,11 +205,11 @@ def set_up_optimization(m, system_recovery=0.7, **kwargs_flowsheet):
             >= m.fs.RO2.permeate_side.properties_mixed[0].flow_vol_phase['Liq'])
 
     check_dof(m, dof_expected=6 if is_twostage else 4)
-    # solve_with_user_scaling(m, tee=False, fail_flag=True)
+    # solve_block(m, tee=False, fail_flag=True)
```
adam-a-a, line 208: commented code

watertap/examples/flowsheets/full_treatment_train/util.py
```diff
@@ -12,40 +12,28 @@
...
-#       bound_push and mu_init as is!!! However, if your flowsheet or block
-#       contains ONLY chemistry, then setting bound_push = 1e-10 and mu_init = 1e-6
-#       works ver well. 
-def solve_with_user_scaling(blk, solver=None, tee=False, fail_flag=True, bound_push=0.1, mu_init=1e-1):
+def solve_block(blk, solver=None, tee=False, fail_flag=True):
     if solver is None:
-        solver = get_solver(options={'nlp_scaling_method': 'user-scaling',
-                                     'bound_push': bound_push,
-                                     'mu_init': mu_init})
+        solver = get_solver(options={'tol':1e-11})
```
adam-a-a: Can you remind me which tolerance is being adjusted here? I'm guessing solver convergence failed without this tolerance value but unsure why.

watertap/examples/flowsheets/full_treatment_train/util.py
```diff
@@ -12,40 +12,28 @@
...
     results = solver.solve(blk, tee=tee)
     if fail_flag:
         check_solve(results)
 
 
 def check_dof(blk, dof_expected=0):
```
adam-a-a, line 31: Do we need this if we already have the same or similar method in `core/util/initialization` ? Guessing we're keeping this to avoid having to replace imports everywhere else.

watertap/examples/flowsheets/full_treatment_train/util.py
```diff
@@ -12,40 +12,28 @@
...
-    if degrees_of_freedom(blk) != dof_expected:
-        raise RuntimeError("The degrees of freedom on {blk} were {dof} but {dof_e} "
-                           "were expected, check the fixed variables on that block".format(
-            blk=blk, dof=degrees_of_freedom(blk), dof_e=dof_expected))
+    assert_degrees_of_freedom(blk, dof_expected)
 
 
 def check_solve(results):
```
adam-a-a, line 35: Same as above - why not just use check_solve from `core/util/initialization`?

adam-a-a (dismissed):
LGTM- I had some minor comments that can be addressed but shouldn't hold up the merge.

watertap/examples/flowsheets/full_treatment_train/util.py
bknueven: It shouldn't be -- that is a carry-over from the tortured history of this branch. I probably need to do a full re-review myself to ensure nothing else like this slipped through.

bknueven:
```suggestion
        solver = get_solver()
```

watertap/examples/flowsheets/full_treatment_train/util.py
```diff
@@ -22,7 +22,7 @@
 
 def solve_block(blk, solver=None, tee=False, fail_flag=True):
     if solver is None:
-        solver = get_solver(options={'tol':1e-11})
+        solver = get_solver()
     results = solver.solve(blk, tee=tee)
```
aladshaw3: One of the chemistry post-treatment blocks is failing a solution check after the tolerance was changed. The model still reaches an optimal solution, but the value is very slightly off now. The failure in the solution check is probably a negligible difference between the values, but I will need to go into it deeper to be certain.

adam-a-a: Looking at the documentation for IPOPT options, I see the default for this convergence tolerance is 1e-8, so I suppose the 1e-11 tolerance yielded a solution with smaller NLP error. As long as we still get "optimal termination" and the solutions are negligibly different, I'd think we'd be better off sticking with the default value of 1e-8.

bknueven: I think we should find the value that will give us reliable results for the chemistry models. By default IDAES relaxes this tolerance to `1e-06`, because that makes the most sense for the typical power plant model. We can, and in my opinion, should make a similar choice.

I'll be working on this today to see if there's a set of global options that works across our WaterTAP models.

andrewlee94: I think there should be a way to overwrite the IDAES default scaling for WaterTap (i.e. set out own set of defaults). John Eslick should know how to do this.

watertap/core/plugins/solvers.py
```diff
@@ -39,9 +39,10 @@ def _presolve(self, *args, **kwds):
 
         self._tee = kwds.get("tee", False)
 
-        # Set the IDAES-default tol
+        # Set the default watertap options
         if "tol" not in self.options:
             self.options["tol"] = 1e-08
+        if "constr_viol_tol" not in self.options:
```
aladshaw3: Same test is failing for the same reason, but the rate of convergence has improved with these changes.  

However, there may be a new problem with this test now: The solver is producing different solutions depending on the operating system now. In windows, it reports a pH of 8.47 (which gives ion distribution of 11.7 % to HOCl and 88.3 % to OCl-), but on Linux it reports a pH of 8.76 (which gives ion distribution of 7.1 % to HOCl and 92.9 % to OCl-). 

To me, this is an unacceptable deviation. At the very least, each OS/build should give the same results.

aladshaw3: Also, the rate of convergence for this test on this branch is significantly worse than the rate of convergence on main (as it currently stands).

bknueven:
@aladshaw3 would you mind having another look at this?

aladshaw3 (approved):
LGTM

adam-a-a (approved):
LGTM

PR #288, opened 2022-01-06
Adding zero-order Feed Block and extending database

andrewlee94:
## Summary/Motivation:
This PR adds a Feed block for zero-order flowsheets with methods for loading initial flowrate and concentration data from a compatible database. This also includes the necessary extensions to the database class to support he new lookups.

- Adds FeedZO block with inherits from the IDAES Feed Block
- Adds example water_sources.yml file to core database
- Extends database functionality for feed water lookups

### Legal Acknowledgement


watertap/unit_models/zero_order/nanofiltration_zo.py
```diff
@@ -11,7 +11,7 @@
 #
 ###############################################################################
 """
-This module constians a ero-order representation of a nanofiltration unit
+This module constians a zero-order representation of a nanofiltration unit
```
adam-a-a:
```suggestion
This module contains a zero-order representation of a nanofiltration unit.
```

watertap/unit_models/zero_order/feed_zo.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+
+    CONFIG = FeedData.CONFIG()
+
+    def build(self):
+        super().build()
+
+    def load_feed_data_from_database(self, overwrite=False):
+        """
+        Method to load initial folwrate and concentrations from database.
```
adam-a-a:
```suggestion
        Method to load initial flowrate and concentrations from database.
```

watertap/data/techno_economic/water_sources.yml
```diff
@@ -0,0 +1,54 @@
+default: seawater
+seawater:
+  metadata: Based on Carlsbad casestudy for seawater
```
adam-a-a:
```suggestion
  metadata: Based on Carlsbad case study for seawater
```

TimBartholomew (dismissed):
LGTM

watertap/data/techno_economic/water_sources.yml
```diff
@@ -0,0 +1,54 @@
...
+      constituent_longform: Strontium
+      value: 0.0013
+      units: kg/m**3
+    sulfate:
+      constituent_longform: Sulfate
+      value: 2.68
+      units: kg/m**3
+    tds:
+      constituent_longform: Total Dissolved Solids (TDS)
+      value: 35
```
adam-a-a, line 43: The ion concentrations should add up and equal to the TDS. Do we want to add a test to check that and possibly adjust database values if needed? I didn't check but thought it'd be worth mentioning to keep in mind.  I also realize that we don't necessarily need the ionic components to add up to TDS if we simply want to account for specific components for targeted removal.

watertap/core/tests/test_wt_database.py
```diff
@@ -156,6 +156,32 @@ def test_get_unit_operation_parameters_subtype_argument(self, db):
...
+    def test_get_solute_set_specified(self, db):
+        comp_set = db.get_solute_set("seawater")
+
+        assert comp_set == [
+            "boron", "bromide", "calcium", "chloride", "magnesium",
+            "potassium", "sodium", "strontium", "sulfate", "tds", "tss"]
+
+    @pytest.mark.unit
+    def test_get_solute_set_no_default(self, db):
+        # First, delte default entry from database
```
adam-a-a:
```suggestion
        # First, delete default entry from database
```

watertap/core/zero_order_properties.py
```diff
@@ -53,9 +54,16 @@ class WaterParameterBlockData(PhysicalParameterBlock):
...
+        description='An instance of a WaterTAP Database to use for parameters.'
+        ))
+    CONFIG.declare('water_source', ConfigValue(
+        description=
+        'Water source to use when looking up parameters from database.'))
     CONFIG.declare("solute_list", ConfigValue(
         domain=list,
-        description="List of solute species names"))
+        description="List of solute species of interest. If None, will use "
+        "all species defined in the water_source provied."))
```
adam-a-a:
```suggestion
        "all species defined in the water_source provided."))
```

watertap/unit_models/zero_order/feed_zo.py
```diff
@@ -0,0 +1,82 @@
...
+
+        for t in self.flowsheet().time:
+            if overwrite or not self.outlet.flow_vol[t].fixed:
+                try:
+                    val = data["default_flow"]["value"]
+                    units = getattr(pyunits, data["default_flow"]["units"])
+                    self.outlet.flow_vol[t].fix(val*units)
+                except KeyError:
+                    _log.info(
+                        f"{self.name} no default flowrate was definined "
```
adam-a-a:
```suggestion
                        f"{self.name} no default flowrate was defined "
```

watertap/data/techno_economic/water_sources.yml
andrewlee94: @adam-a-a That ends up being problematic - all concentrations must be assumed to be independent if we want to write removal constraints for them. I.e. if TDS must be the sum of all ions, then you can either have TDS as a component, or you can have all the separate ions, but not both. This is important at the unit model level, as it means you are limited to either specifying a removal of the aggregate TDS, or you have to do each species separately.

Also - these numbers came from the existing data files, so the WT3 team will have to weigh in on this.

watertap/unit_models/zero_order/tests/test_feed_zo.py
```diff
@@ -0,0 +1,125 @@
...
+            model.fs.unit.outlet.flow_vol[0])
+        assert model.fs.unit.outlet.flow_vol[0].fixed
+
+        for j, v in data["solutes"].items():
+            assert pytest.approx(v["value"], rel=1e-12) == value(
+                model.fs.unit.outlet.conc_mass_comp[0, j])
+            assert model.fs.unit.outlet.conc_mass_comp[0, j].fixed
+
+    @pytest.mark.unit
+    def test_load_feed_data_from_database_no_overwirte(self, model):
```
adam-a-a:
```suggestion
    def test_load_feed_data_from_database_no_overwrite(self, model):
```

watertap/unit_models/zero_order/tests/test_feed_zo.py
```diff
@@ -0,0 +1,125 @@
...
+    @pytest.mark.unit
+    def test_load_feed_data_from_database_missing(self, model, caplog):
+        data = model.db.get_source_data()
+
+        del data["default_flow"]
+        del data["solutes"]["tds"]
+
+        model.fs.unit.load_feed_data_from_database(overwrite=True)
+
+        assert ("fs.unit no default flowrate was definined in database water "
```
adam-a-a:
```suggestion
        assert ("fs.unit no default flowrate was defined in database water "
```

watertap/data/techno_economic/water_sources.yml
kurbansitterley: We considered putting in such a constraint to WT3 and quickly determined it was pretty much impossible. 

From a water chemistry perspective, there is no reason the ion concentrations should add up to the TDS. When you are measuring TDS, you are literally drying out a sample of the water and weighing it (shortened version of the method). Naturally, the ionic species that you measured in an aquatic sample are not present in this dried sample. Ionic aquatic species form some sort of precipitate as a result of the drying process and this is what you are measuring as TDS... thus the molar mass of the compound would go up and would almost certainly not translate to a 1:1 comparison with the aquatic species concentration. 

Also, TDS is a bulk measurement that measures _all_ the dissolved components in a water sample. This includes ionic species as well as dissolved organics and any very small particulates that would get through the filter cutoff used to designate "dissolved" (I don't remember what it is off the top of my head). Further, to even try to compare the summation of dissolved ionic species to TDS (assuming you had a good bent on the water chemistry involved), you would have to assume that whoever measured the ionic species did a comprehensive sweep of all the possible ionic species in the water sample, had excellent sensitivities on the instrument used, and reported everything to you... This is almost certainly not the case for this dataset we have.

adam-a-a: @kurbansitterley Those are all good points. However, I think back to the times when I used available tools from membrane manufacturers, and usually when feedwater composition is entered by the user, (1) the ion concentrations are added up to yield some calculated TDS (not necessarily the precise measurement you'd get at the plant), and (2) there is a check to see if the composition is balanced in terms of eq/L. So from one perspective, we could just have things as they are with the intention of giving the user a set of parameter values to choose from. From another perspective, users might depend on these feedwater compositions as balanced and "complete" compositions that they could load into other more detailed models, which may account for electroneutrality of the feed solution, for example; other users might automatically assume that the sum of all ion concentrations listed equal the TDS value provided. Whatever we decide to do here, I suppose we just need to be clear about it in documentation.

adam-a-a: @andrewlee94 - with regard to removal rates of either ions or TDS, we probably should consider whether the current implementation makes sense if values can be loaded and fixed for removal rates for both ions and TDS. Should we add some sort of check to make sure? It's one thing to be off by a little bit in terms of the ion concentrations summing up to TDS, but we should make sure any offsets are not significant.

andrewlee94: @adam-a-a First, I think there might be a slight confusion here. I think you are reading TDS as total dissolved salts (which was my initial reading), whilst in WT3 I think it is total dissolved solids. if that is the case, then TDS is not necessarily the sum of all the ionic species (although that is part of it). Also, we could consider the case where the specific ions are only a subset of the total dissolved salts, but that the non-specific ions are all treated the same (i.e. only specified ions have unique removals).

adam-a-a: @andrewlee94 Total dissolved solids is also what I'm considering. The terms are used interchangeably even though technically, TDS would include more than just salts. However, I'm fairly confident that in desalination applications, the salts would be dominant and are usually used to represent TDS, whether defined as total dissolved salts or solids. 

At the end of the day, we can package this up however we like and provide sufficient documentation explaining such, but I'd lean towards setting this up in a way that the typical desal expert would expect things to be setup. This shouldn't hold up the PR, but maybe we should get group consensus on how we want these water source yaml files to be setup. Should they be setup with the intention of providing a "complete" picture of the feedwater composition (i.e., as complete as practically possible for modeling purposes), or will they be setup to provide subsets of interest? Since the code is written in a way that the user can define the subset they want from the DB already, I'd vote for these feedwater yaml files to be representative of a relatively "complete", balanced composition (without trying to go to the level of detail that @kurbansitterley mentioned with regard to conducting "a comprehensive sweep of all the possible ionic species"). At the very least, we should ensure that TDS values provided in data files exceed the sum of all ion concentrations (i.e., as long as TDS is not less than that sum).

adam-a-a: Maybe the short answer is to have "calculated TDS" and "actual TDS". Maybe actual isn't the best word... measured? That way, the property model could have `calculated_TDS`, and the user could choose between the two, assuming that they are different from each other.

adam-a-a (approved):
LGTM. Let's consider the points about TDS and ion concentrations provided in yaml files as a separate discussion and follow-up on that. Code functionality is what's important here, and this PR hits the mark on that.

watertap/data/techno_economic/water_sources.yml
kurbansitterley: Sorry, I didn't mean to be pedantic there, just wanted to point out that sum(ion species) != TDS...

A "balanced" around eq/L would make more sense and my gut tells me that it is probably more feasible, but I don't have any direct experience with that. Your point about in desalination applications is a good one. Salts would almost certainly dominate the TDS value, especially for seawater.

bknueven (approved):

watertap/data/techno_economic/water_sources.yml
andrewlee94: @adam-a-a Yes, I think this needs to be a part of a larger discussion - I just used the values that WT3 had in their datafile.

Whilst that makes sense from a realistic stand-point, making that work in an EO environment could be challenging. I.e. how and where do we check? Is this something which is just in the database definitions, but let the user override as they want, or do we need to enforce this check? If we do enforce the check, is it only at the inlet, or at every state in the model? Enforcing that check now means that we need to have some sort of additional constraint, probably an inequality, which will add to problem complexity and another potential point of failure.

PR #292, opened 2022-01-12
Expanding zero-order water source database

andrewlee94:
## Summary/Motivation:
This PR expands the zero-order database for water sources, and adds some basic tests for the data file.

- Adds source data for different seawaters, agricultural, municipal, brackish and produced waters.
-

### Legal Acknowledgement

### DIFF
--- a/watertap/data/techno_economic/water_sources.yml
+++ b/watertap/data/techno_economic/water_sources.yml
@@ -1,4 +1,312 @@
 default: seawater
+agricultural:
+  metadata: Based on San Luis case study
+  solutes:
+    arsenic:
+      constituent_longform: Arsenic
+      value: 1.3E-05
+      units: kg/m**3
+    boron:
+      constituent_longform: Boron
+      value: 0.00315
+      units: kg/m**3
+    calcium:
+      constituent_longform: Calcium
+      value: 0.000483
+      units: kg/m**3
+    chromium:
+      constituent_longform: Chromium
+      value: 1.45E-05
+      units: kg/m**3
+    iron:
+      constituent_longform: Iron
+      value: 0.000964
+      units: kg/m**3
+    manganese:
+      constituent_longform: Manganese
+      value: 2.8E-05
+      units: kg/m**3
+    mercury:
+      constituent_longform: Mercury
+      value: 1.5E-06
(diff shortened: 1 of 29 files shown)


watertap/data/techno_economic/water_sources.yml
```diff
@@ -1,4 +1,312 @@
...
+      constituent_longform: Total Dissolved Solids (TDS)
+      value: 2.5
+      units: kg/m**3
+  default_flow:
+    value: 1.336278
+    units: m**3/s
+  pH:
+    max: 7.6
+    min: 7.6
+brackish_high:
```
adam-a-a, line 117: Between `brackish_ave` and `brackish_high`, there isn't a significant difference in TDS. If we want a composition representative of highly brackish water, we should have a set of solutes that corresponds to TDS around 10,000 ppm. Even though the naming of these two probably just follow some specific case provided in the study mentioned in metadata, it isn't representative of highly brackish water. Just something to have in mind for users who may look through these water sources to choose a set representative of highly brackish groundwater.

adam-a-a: Note, the definition of high brackish vs. low/medium can vary by literature source, but I still don't think we'd regard this a high brackish.

adam-a-a (approved):
LGTM - just a minor comment on the "brackish_high" water source. We can add data for a highly brackish water source later on. I'd probably rename "brackish_high" since we'd want to reserve that name for a highly brackish source.

watertap/data/techno_economic/water_sources.yml
andrewlee94: Again - I only had the data from WT3 available and I don't know enough to change it (or judge it). I will also note they had a brackish_low dataset, but that was identical to the brackish_ave set.

kurbansitterley: A better brackish_high could be the san_luis data at ~12 g/L TDS, but we considered that an agriculture source.

kurbansitterley (commented):
LGTM, and I echo @adam-a-a comment on brackish_high.

kurbansitterley (approved):
LGTM, and I echo @adam-a-a comment on brackish_high.

Issue #295, opened 2022-01-13
Initial MVC Model


carson-tucker:
For NAWI, we need to implement an MVC model in WaterTAP. 

This will require the following: 

- [x] Pure water property package with vapor
- [x] Evaporator
- [x] Condenser
- [x] Compressor (maybe different than IDAES)
- [x] Addition of Chen's approximation for log-mean temperature difference to IDAES heat exchangers

PR #296, opened 2022-01-13
Add initial mvc

carson-tucker:
## Fixes/Addresses:

#295 

## Summary/Motivation:

We need to add an MVC model to WaterTAP. 

- 
-

### Legal Acknowledgement


TimBartholomew:
@lbianchi-lbl @bknueven I'm not sure how to address these test failures. 

The macOS setup was failing before I merged in main (I assumed I needed to merge in main without looking at the error), after I merged in main the py3.9/Linux failed to set up. 

The mac failure is for a test added in this PR, but I'm not sure how to see more details on the failure, all I see is `Error: Process completed with exit code 4.`

lbianchi-lbl:
@TimBartholomew I'm looking into this now. 

- I've checked the Linux failure and it looks like it could be a network error, so I restarted the job(s) to see if it goes away
- For the macOS job, it's less clear
  - Usually when pytest crashes like that, without running all the tests or generating a failure report, it might be indicative of an error with e.g. the stdout/stderr stream capture
  - According to https://docs.pytest.org/en/latest/reference/exit-codes.html, exit code 4 means an error with the pytest CLI invocation, which sounds a bit strange since the error seems to occur when the test run is already underway
- I'll take a look at the test being added by this PR to see if something catches the eye

TimBartholomew:
It is a string comparison :)

lbianchi-lbl:
@TimBartholomew :) lol, I'm not sure why, but somehow using `capsys` to extract that information might be causing the error. A wild guess would be that it's related to how Pyomo collects the output of the separate process where the solver is running, which might be different (and possibly less tested) because it's macOS and not using the IDAES Ipopt executable.

One thing to try would be to move the code where the model is created directly in `test_demo_heat_exchanger.py` as a fixture, and then perform the assertions in the test function without using `capsys`.

lbianchi-lbl:
Also: as far as I can tell from https://github.com/IDAES/idaes-pse/blob/f8b6ade5954ea8722777f28ad6d77dfae561d9ae/idaes/core/process_base.py#L226, it should be possible to pass any file-like object to the `report()` method. If you create a `io.StringIO` object, you can use it as the `ostream` kwarg in `report()`, and then use `getvalue()` to extract a string to be used for the comparison (which, one more time, is probably not what we want in the long term, but should hopefully be enough to avoid these type of issues with pytest).

andrewlee94:
@lbianchi-lbl Now that you mention it, I do recall having that capability - the benefits of directly copying what Pyomo did.

bknueven:
I can reproduce, but haven't been able to fix, the issues with macOS. That said, I think it would be preferable for these example `main` functions to return the model object. That would make testing easier (no more string comparisons), and the examples would be useful when called from Python directly -- something I've had to add in to other examples in the past.

bknueven:
For now, this test should be skipped on macOS. The issue is the IDAES heat exchanger needs a complied IDAES extension which isn't available. When I run the example directly:
```
(idaes) bknueven-35255s:components bknueven$ python demo_heat_exchanger.py 
2022-01-31 09:49:40 [INFO] idaes.init.fs.unit.hot.properties_in: Property initialization: optimal - Optimal Solution Found.
2022-01-31 09:49:40 [INFO] idaes.init.fs.unit.hot.properties_out: Property initialization: optimal - Optimal Solution Found.
2022-01-31 09:49:40 [INFO] idaes.init.fs.unit.hot.properties_out: fs.unit.hot.properties_out State Released.
2022-01-31 09:49:40 [INFO] idaes.init.fs.unit.hot: Initialization Complete
2022-01-31 09:49:40 [INFO] idaes.init.fs.unit.cold.properties_in: Property initialization: optimal - Optimal Solution Found.
2022-01-31 09:49:40 [INFO] idaes.init.fs.unit.cold.properties_out: Property initialization: optimal - Optimal Solution Found.
2022-01-31 09:49:40 [INFO] idaes.init.fs.unit.cold.properties_out: fs.unit.cold.properties_out State Released.
2022-01-31 09:49:40 [INFO] idaes.init.fs.unit.cold: Initialization Complete
Cannot find library "/Users/bknueven/.idaes/bin/functions.so".
function cbrt not available
```

I'm not sure why we aren't getting a more elegant failure from pytest -- I even see it when **not** capturing the output.

PR #297, opened 2022-01-13
WaterTAP3 zeroth order model database merge. 

avdudchenko:
Importing watertap3 database of performance metrics for zeroth-order models from their .csv files to .yaml format. 

## Fixes/Addresses:

Added the first version of basic zeroth-order model database, this does not include any cost numbers. Please check that the formatting is all good, and if we need to change any units/defaults and I can update. This includes removal factors for default and OCWD cases where available. I can update this to include all other cases.

## Summary/Motivation:
-data for WT3 merger into WT

-added new data folder in data/techno_economic/
-data extracted from WT3 .csv files located in zeroth_order_model_data folder 
-Following process have not been imported based on comments
--'irwin','waterreuse','onsite_storage_old','silica_solution','brine_ro','deep_well','fc_crystallizer'
-Following removal factors were not imported
--'ultraviolet_transmittance_uvt'

### Legal Acknowledgement


watertap/data/techno_economic/zeroth_order_model_data/aeration_basin.yaml
```diff
@@ -0,0 +1,41 @@
+default:
+  default_removal_mass_solute:
+    unit: kg/kg
+    value: 0
+  energy_electric_flow_vol_inlet:
+    unit: kWh/m^3
```
andrewlee94: It would be better to have powers in Python format (`**`) as it means we can directly parse the units into Pyomo objects. Otherwise, we need to first process the `^` into a `**`.

watertap/data/techno_economic/zeroth_order_model_data/aeration_basin.yaml
```diff
@@ -0,0 +1,41 @@
...
+    value: 0.41226735
+  recovery_vol:
+    reference: .nan
```
andrewlee94: Minor comment, but for a reference I would personally have used None (or omitted the reference line) rather than `.nan`. This is probably just semantics however (and this entry is just for information and not used currently).

watertap/data/techno_economic/zeroth_order_model_data/anaerobic_digestion_oxidation.yaml
```diff
@@ -0,0 +1,20 @@
...
+    value: 0.15
+  recovery_vol:
+    reference: 'Beef Processing case study. '
+    unit: m^3/m^3
+    value: 0.9999
+  removal_frac_mass_solute:
+    bod:
+      constituent_longform: Biological Oxygen Demand (BOD)
+      unit: kg/kg
+      value: 0.0009699999999999999
```
andrewlee94: Minor comment - is this just 0.00097 represented as a float (due to binary precision)? It probably doesn't matter in the end, but it would be nice to use the actual value if possible (rather than the nearest float).

watertap/data/techno_economic/zeroth_order_model_data/nano_filtration.yaml
```diff
@@ -0,0 +1,88 @@
+default:
```
andrewlee94: This is a duplicate of the yml file we already have in the `techno_economic` folder (which was used as the initial example). However, I am seeing some differences in the file, notably in the names of components (e.g. `Fe_3+` versus `iron`), the longform names and many values (likely due to precision).

We need to reconcile these two version and make sure we have standard names across all files (This also includes the water source definitions).

andrewlee94 (changes requested):
My main comment is that I don't think we want/need the `zeroth_order_model_data` sub-folder. The entire `techno_economic` folder is intended for zero order models, so I think we just want everything in that folder.

I had a couple of minor comments on things I saw from a quick skim. However, detailed checking of the data will be better done as we implement each unit model.

watertap/data/techno_economic/zeroth_order_model_data/aeration_basin.yaml
avdudchenko: Regarding folders, I think we gonna have a lot of data of different type in that folder, cost data, model performance data (as in this folder) For user navigation, it might make sense to have separate data locations so they are organized at least based on something.  (order of the model, data type?)

watertap/data/techno_economic/zeroth_order_model_data/aeration_basin.yaml
avdudchenko: Oh good point. I did not notice it was doing that. I will update that to None (Think this is a quirk from csv import)

watertap/data/techno_economic/zeroth_order_model_data/anaerobic_digestion_oxidation.yaml
avdudchenko: Hmm not sure. These were auto imported and converted. I did not put precision limits on these imports, I will look into adding that into next commit.

watertap/data/techno_economic/zeroth_order_model_data/nano_filtration.yaml
avdudchenko: Yes, I updated all the names based on standard defined names in the following document that was reviewed by everyone I believe https://app.box.com/s/kd19k78qgnfq2p8mvt3l617xa0tji0uw 

In there we wanted to define all elements via their element name (Fe, etc. including oxidation state) You can find my standard conversion list for constituents in the following file  
 https://app.box.com/s/iqdfxdem8wc8lmjfc3fgbk9mf8v9ze7d 

We can revert back to naïve names if we want to.

watertap/data/techno_economic/zeroth_order_model_data/aeration_basin.yaml
bknueven: It might make sense to have things organized based on data type. But I think for now this `techno_economic` folder will be for data for "WaterTAP3-style" models. As we discussed last time, the distinction between "0-order" and "0-dimensional" is not always clear, so I wouldn't pick that as the organization method.

bknueven: I'll also disagree with @andrewlee94 on the power representation. If users are editing this file, `^` is more typical in mathematical type-setting and elsewhere. And its easy enough to replace every instance of `^` with `**`.

watertap/data/techno_economic/zeroth_order_model_data/anaerobic_digestion_oxidation.yaml
bknueven: Skimming through it looks like the longest number has 9 significant figures (some values of `energy_electric_flow_vol_inlet`).

watertap/data/techno_economic/zeroth_order_model_data/aeration_basin.yaml
TimBartholomew: yeah I agree with @andrewlee94 and @bknueven, lets just drop these into the technoeconomic folder and we can reorganize into subfolders later

watertap/data/techno_economic/zeroth_order_model_data/decarbonator.yaml
```diff
@@ -0,0 +1,12 @@
+default:
```
adam-a-a: Example of unit model with no default performance data. If a particular case study was used to originally populate these parameters, I would eventually opt for choosing one of those case studies to provide default values here, with a metadata key pointing to that case study.

adam-a-a: In other words, I think it'd be better to include some values rather than provide no values and wait for the user to load data from case studies afterwards. If this case only has cost data, then we can leave as is. However, we should determine whether some of these yaml files are missing both performance and cost data and (a) mark them as placeholder files for future population or (b) remove them.

andrewlee94 (approved):
I think we will need to refine things more, but that can wait until we actually start trying these out,

bknueven (approved):
The only thing I find slightly weird is that some of the `reference` values are `.nan`.

But at this point I agree with @andrewlee94: we can refine this as-we-go.

PR #301, opened 2022-01-19
Parameter sweep refine

### LINKED ISSUE #300: Enable storing variables of WaterTAP models in h5 files for parameter sweep
This is a feature currently being developed where the variables of WaterTAP models can be stored in an h5 file when calling parameter sweep. Unit tests are currently being implemented in the branch [`bulk_output`](https://github.com/k1nshuk/watertap/tree/bulk_output/watertap). They will be pushed into the branch [`parameter_sweep_refine`](https://github.com/watertap-org/watertap/tree/parameter_sweep_refine). I will initiate a PR from the latter branch.

k1nshuk:
## Fixes/Addresses:

This PR enables the parameter sweep tool to now record detailed model values for every parameter sweep model evaluation. These values are first collected in a python dictionary which is then stored in an H5 file. A companion human readable text file is created with the same file name as the H5 file which contains the metadata of the information contained within the H5 file.

## Summary/Motivation:

Storing a detailed summary of model evaluations from a parameter sweep enables better post-processing of the results as it allows the user to revisit prior runs to analyze the performance of their models

- Storing the detailed solutions of a parameter sweep run in an H5 file.
- Storing the metadata of the H5 file in a separate text file of the same name.
- Ability to read in the parameter sweep variables from file.
- Ability to load a default configuration for a WaterTAP model from file.

### Legal Acknowledgement


docs/how_to_guides/how_to_use_parameter_sweep.rst
```diff
@@ -25,11 +25,13 @@ In general you would import your own flowsheet module.
...
 .. testsetup::
 
-    # quiet idaes logs
+    # quiet idaes and pyomo logs
     import idaes.logger as idaeslogger
     idaeslogger.getLogger('ideas.core').setLevel('CRITICAL')
     idaeslogger.getLogger('ideas.core.util.scaling').setLevel('CRITICAL')
     idaeslogger.getLogger('idaes.init').setLevel('CRITICAL')
+    import logging
+    logging.getLogger('pyomo.core').setLevel(logging.ERROR)
```
bknueven: What errors were we seeing from Pyomo?

docs/how_to_guides/how_to_use_parameter_sweep.rst
```diff
@@ -87,18 +89,19 @@ For this RO flowsheet we'll report the levelized cost of water, the optimized RO
...
 .. testcode::
 
-    parameter_sweep(m, sweep_params, outputs, 'results.csv')
+    parameter_sweep(m, sweep_params, outputs, csv_results_file='results.csv')
 
 .. testcleanup::
 
     import os
     os.remove('results.csv')
+    os.remove('output_dict.h5')
```
bknueven: Are we silently creating this file now?

setup.py
```diff
@@ -101,6 +101,8 @@
         "scipy",
         # https://www.python.org/dev/peps/pep-0508/#environment-markers
         'pywin32==225 ; platform_system=="Windows" and python_version>="3.8"',
+        # for parameter_sweep
+        "h5py",
```
bknueven, line 105: At some point we should think about using soft requirements. But maybe we'll wait on IDAES to sort this out...

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
```diff
@@ -257,7 +257,8 @@ def solve(blk, solver=None, tee=False):
     if solver is None:
         solver = get_solver()
     results = solver.solve(blk, tee=tee)
-    assert_optimal_termination(results)
+
+    return results
```
bknueven, line 262: Do we use the results object now? If not, why not `try/except` when this function is called?

watertap/examples/flowsheets/RO_with_energy_recovery/default_configuration.yaml
```diff
@@ -0,0 +1,258 @@
+m.fs.M1.P1_state[0.0].enth_mass_phase[Liq]: 99740.72571999984
```
bknueven: All these should *not* include the `m.`. The name of the top-level object should not matter. It could be called `model`, or `foo`. E.g.
```suggestion
fs.M1.P1_state[0.0].enth_mass_phase[Liq]: 99740.72571999984
```

watertap/examples/flowsheets/RO_with_energy_recovery/mc_sweep_params.yaml
```diff
@@ -0,0 +1,37 @@
+#================================================================
+A_comp:
+  type: NormalSample
+  model_value: m.fs.RO.A_comp
```
bknueven: Similar in this file
```suggestion
  model_value: fs.RO.A_comp
```

Also, I'm not sure why we wouldn't call this `model_param` or just `param`.

watertap/examples/flowsheets/RO_with_energy_recovery/monte_carlo_sampling_RO_ERD.py
```diff
@@ -61,9 +72,13 @@ def run_parameter_sweep(results_file, seed=None, use_LHS=False):
     # Run the parameter sweep study using num_samples randomly drawn from the above range
     num_samples = 10
 
+    # Filename of the h5 file that we want to write but without the extension
```
bknueven: It should be easy to support both with and without the extension.

watertap/examples/flowsheets/RO_with_energy_recovery/monte_carlo_sampling_RO_ERD.py
```diff
@@ -61,9 +72,13 @@ def run_parameter_sweep(results_file, seed=None, use_LHS=False):
...
+    h5_fname = "monte_carlo_results"
+
     # Run the parameter sweep
-    global_results = parameter_sweep(m, sweep_params, outputs, results_file=results_file, 
-        optimize_function=optimize, optimize_kwargs={'solver':solver}, num_samples=num_samples, seed=seed)
+    global_results = parameter_sweep(m, sweep_params, outputs, csv_results_file=results_file,
+        results_fname=h5_fname, optimize_function=optimize, optimize_kwargs={'solver':solver},
```
bknueven: Why is the name of the csv file `csv_results_file` and the h5 file `results_fname`. Is there a good reason to output both? Could we just have `results_file_name` and write *either* a csv *or* an h5 file? We could have an additional flag for it and/or infer it from the file extension.

watertap/examples/flowsheets/RO_with_energy_recovery/parameter_sweep_input_parser.py
```diff
@@ -0,0 +1,310 @@
...
+    for param, values in input_dict.items():
+
+
+        # Split the dot-separated object path into a list of
+        # separate objects and attributes
+        path = _split_pyomo_path(values['model_value'])
+
+        # Follow the attribute path to the correct nested
+        # property, should work with indexing!
+        child = _return_child_property(m, path)
```
bknueven: If you drop the `m.` in the name, you can replace these lines (and both functions) with
```
child = m.find_component(values['model_value'])
if child is None:
    raise ValueError(f"Could not acccess attribute {values['model_value']}")

File: watertap/examples/flowsheets/RO_with_energy_recovery/parameter_sweep_input_parser.py [hunk 10]
```diff
@@ -0,0 +1,310 @@
...
+    fail_ct = 0
+
+    for key, default_value in input_dict.items():
+        # Split the dot-separated object path into a list of
+        # separate objects and attributes
+        path = _split_pyomo_path(key)
+
+        # Follow the attribute path to the correct nested
+        # property, should work with indexing!
+        child = _return_child_property(m, path)
```
bknueven: Similar to above.

watertap/examples/flowsheets/RO_with_energy_recovery/parameter_sweep_input_parser.py
```diff
@@ -0,0 +1,310 @@
...
+            failed_to_set_value = True
+
+        fail_ct += failed_to_set_value
+
+        if failed_to_set_value:
+            warning_msg = 'WARNING: Could not set value of parameter %s' % (key)
+            warnings.warn(warning_msg)
+
+    print('Set %d of %d options to default values (%d failures, see warnings)' %
+        (len(input_dict)-fail_ct, len(input_dict), fail_ct))
```
bknueven: A nit: Python has supported f-strings for a long time now...

watertap/examples/flowsheets/RO_with_energy_recovery/parameter_sweep_input_parser.py
```diff
@@ -0,0 +1,310 @@
...
+
+        if do_nothing:
+            new_value = old_value
+
+        else:
+            new_value = default_value
+
+        if verbose:
+            print('Property: %s' % (key))
+            print('New Value: %e, Old Value: %e' % (new_value, old_value))
```
bknueven: Could we use a logger at the DEBUG level here?

watertap/tools/parameter_sweep.py
```diff
@@ -16,13 +16,68 @@
...
+                        # ERROR
+                        'solverFailure' : 17,              # Solver failed to terminate correctly
+                        'internalSolverError' : 18,        # Internal solver error
+                        'error' : 19,                      # Other errors
+                        # ABORTED
+                        'userInterrupt' : 20,              # Interrupt signal generated by user
+                        'resourceInterrupt' : 21,          # Interrupt signal in resources used by
+                                                           #    optimizer
+                        'licensingProblems' : 22,          # Problem accessing solver license
+        }
```
bknueven: I don't really understand this -- I would do:
```suggestion
from pyomo.opt import TerminationCondition as _TerminationCondition
_index_to_termination_condition = { idx : cond for idx, cond in enumerate(_TerminationCondition) }
_termination_condition_to_index = { cond : idx for idx, cond in _index_to_termination_condition.items() }
```
And just use the two dictionaries as needed.

watertap/tools/parameter_sweep.py
```diff
@@ -319,108 +375,288 @@ def _interp_nan_values(global_values, global_results):
...
-                                     ``optimize_function``. The first arg will always be ``model``,
-                                     e.g., ``optimize_function(model, **optimize_kwargs)``. The default
-                                     uses no kwargs.
+def _create_component_output_skeleton(component, num_samples):
+    # TODO: Revisit thie variable "component" name
+
+    comp_dict = {}
+    attr_list  = dir(component)
+    comp_dict["value"] = np.zeros(num_samples, dtype=np.float)
+    if 'lb' in attr_list:
```
bknueven: Why not just use `hasattr`?
```suggestion
    if hasattr(component, 'lb'):
```

watertap/tools/parameter_sweep.py
```diff
@@ -430,9 +666,17 @@ def parameter_sweep(model, sweep_params, outputs, results_file=None, optimize_fu
...
+            reinitialize_function(model, **reinitialize_kwargs)
+
+        # Attempt to actually solve the flowsheet
         try:
-            # Simulate/optimize with this set of parameters
-            optimize_function(model, **optimize_kwargs)
+            # Simulate/optimize with this set of parameter
+            results = optimize_function(model, **optimize_kwargs)
+            pyo.assert_optimal_termination(results)
+            # _force_exception(k) # DELETE_ME!!!
```
bknueven: Why have `try/except` block here at all, if we *expect* the `optimize_function` to return the result object? We can use `pyo.check_optimal_termination` instead.

watertap/tools/parameter_sweep.py
```diff
@@ -319,108 +375,288 @@ def _interp_nan_values(global_values, global_results):
...
+    for key in sweep_params.keys():
+        var = sweep_params[key].pyomo_object
+        var_str = sweep_params[key].pyomo_object.name
+        output_dict["sweep_params"][var_str] =  _create_component_output_skeleton(var, num_samples)
 
-        model : A Pyomo ConcreteModel containing a watertap flowsheet, for best 
-                results it should be initialized before being passed to this
-                function.
+    # 2. lets deal with the output variables from the pyomo model
+    for var in variables_in_activated_equalities_set(model.fs):
```
bknueven: Much better to do:
```suggestion
    for var in model.component_data_objects(Var, active=True):
```

watertap/tools/parameter_sweep.py
```diff
@@ -319,108 +375,288 @@ def _interp_nan_values(global_values, global_results):
...
+        var_str = var.name
+        output_dict["outputs"][var_str] = _create_component_output_skeleton(var, num_samples)
 
-        sweep_params: A dictionary containing the values to vary with the format
-                      ``sweep_params['Short/Pretty-print Name'] =
-                      (model.fs.variable_or_param[index], lower_limit, upper_limit, num_samples)``.
-                      A uniform number of samples ``num_samples`` will be take between
-                      the ``lower_limit`` and ``upper_limit``.
+    # 3. lets deal with expressions
+    for expr in expressions_set(model.fs):
```
```suggestion
    for expr in model.component_data_objects(Expression, active=True):
```

watertap/tools/parameter_sweep.py
```diff
@@ -319,108 +375,288 @@ def _interp_nan_values(global_values, global_results):
...
+        expr_str = expr.name
+        output_dict["outputs"][expr_str] = _create_component_output_skeleton(var, num_samples)
 
-        outputs : A dictionary containing "short names" as keys and and Pyomo objects
-                  on ``model`` whose values to report as values. E.g.,
-                  ``outputs['Short/Pretty-print Name'] = model.fs.variable_or_expression_to_report``.
+    # 4. Lets deal with objectives
+    for obj in total_objectives_set(model):
+        obj_str = obj.name
+        output_dict["outputs"][obj_str] = _create_component_output_skeleton(var, num_samples)
```
bknueven: The three `for` loops above could be replaced with
```
for pyo_obj in model.component_data_objects((Var, Expression, Objective), active=True):
   output_dict["outputs"][pyo_obj.name] = _create_component_output_skeleton(pyo_obj, num_samples)
```

watertap/tools/parameter_sweep.py
```diff
@@ -319,108 +375,288 @@ def _interp_nan_values(global_values, global_results):
...
-    '''
+        # 2. Expressions
+        for expr in expressions_set(model.fs):
+            output_dict["outputs"][expr.name]["value"][case_number] = np.nan
 
-    # Get an MPI communicator
-    comm, rank, num_procs = _init_mpi(mpi_comm)
+        # 3. Objectives
+        for obj in total_objectives_set(model):
+            output_dict["outputs"][obj.name]["value"][case_number] = np.nan
```
bknueven: Similar to above on the `for` loops.

watertap/tools/parameter_sweep.py
```diff
@@ -444,32 +688,74 @@ def parameter_sweep(model, sweep_params, outputs, results_file=None, optimize_fu
...
 
-    # ================================================================
-    # Save results
-    # ================================================================
+        elif reinitialize_before_sweep and previous_run_failed:
+            pass
+
+        # Update the loop based on the reinitialization
+        solver_termination_condition = results.solver.termination_condition.name
+        _update_local_output_dict(model, sweep_params, k, local_values[k, :], solver_termination_condition, local_output_dict)
```
```suggestion
        _update_local_output_dict(model, sweep_params, k, local_values[k, :], results.solver.termination_condition, local_output_dict)
```

If you take some of the suggestions above.

bknueven (changes requested):
I have a lot of suggestions. My main concern is we should not be replicating functionality in Pyomo: it will turn into a maintenance issue.

Please feel free to ask me personally about manipulating Pyomo models: chances are somebody has wanted to do it in the past, and it's already baked-in. There is lots of code in the PR as-it-is which can be replaced with a few method calls.

watertap/examples/flowsheets/RO_with_energy_recovery/monte_carlo_sampling_RO_ERD.py
k1nshuk: The ability to write csv files will go away in a later PR since the dictionary stored within the h5 file is much more exhaustive and makes the csv file redundant. The reason the two exist concurrently is because I wanted to make sure that the transition from one to another happens smoothly.

docs/how_to_guides/how_to_use_parameter_sweep.rst
k1nshuk: Currently, yes. I can change this behavior if its not desired.

watertap/examples/flowsheets/RO_with_energy_recovery/monte_carlo_sampling_RO_ERD.py
bknueven: I personally think we should maintain both capabilities. While h5 files have their uses, csv files have the advantage of being human-readable, can be opened in Excel, and are easily parsed. Further, you may not *want* to save all the data from a model -- for doing large sweeps this could become an issue.

watertap/examples/flowsheets/RO_with_energy_recovery/parameter_sweep_input_parser.py
eyoung55: The muscle memory to do %-formatting is strong, but I'll work on it! Fixed.

watertap/examples/flowsheets/RO_with_energy_recovery/parameter_sweep_input_parser.py
eyoung55: Didn't know about this, definitely much easier--thanks Ben.

watertap/examples/flowsheets/RO_with_energy_recovery/mc_sweep_params.yaml
eyoung55: `param` works for me, fixed.

watertap/examples/flowsheets/RO_with_energy_recovery/default_configuration.yaml
eyoung55: fixed.

watertap/tools/parameter_sweep.py
k1nshuk: Incorporated in a later commit.

watertap/tools/parameter_sweep.py
k1nshuk: The class `TerminationConditionMapping` still exists, but the dictionary is created by importing `TerminationCondition` from `pyomo.opt`. I agree, I was duplicating things present within Pyomo and wasn't a fan of it at creation, but I don't know why I didn't think of this implementation in the first place. I do still believe that `TerminationConditionMapping` should remain because, its an added layer on top of `pyomo.opt.TerminationCondition`.

bknueven: I maintain there's no need for a class -- an instance of this object exists to store static data. If we really want the forward and reverse map in a single "container", we could do:
```
class _TerminationConditionMap:
    int_to_termination_condition = { idx : cond for idx, cond in enumerate(_TerminationCondition) }
    termination_condition_to_int = { cond : idx for idx, cond in enumerate(_TerminationCondition) }
```
Then to use:
```
_TerminationConditionMap.int_to_termination_condition[i]
```

watertap/tools/parameter_sweep.py
```diff
@@ -319,19 +342,460 @@ def _interp_nan_values(global_values, global_results):
...
+        if termination_condition in pyomo_termination_condition.str_to_int_mapping:
+            ctr += 1
+            return pyomo_termination_condition.str_to_int_mapping[termination_condition]
+    elif isinstance(termination_condition, np.integer):
+        if termination_condition in pyomo_termination_condition.int_to_str_mapping:
+            ctr += 1
+            return pyomo_termination_condition.int_to_str_mapping[termination_condition]
+
+    if ctr == 0:
+        raise ValueError("Pyomo termination condition value not found in known list")
```
bknueven: I don't really understand why we wouldn't know which direction (from int to TerminationCondition or TerminationCondition to int).

If we know which direction we're going, we should just access the dictionary directly, as opposed to calling this function.

k1nshuk: Yeah, I agree. This can go away. I'll make the change.

watertap/tools/parameter_sweep.py
```diff
@@ -319,108 +342,290 @@ def _interp_nan_values(global_values, global_results):
...
+
+        # 1. Variables: We will set everything to NaNs unless they are an input
+        # for var in variables_in_activated_equalities_set(model.fs):
+        #     if var.name in op_ps_dict.keys():
+        #         # This conditional exists so that we capture the actual input
+        #         # values. This is a redundancy since we already capture the
+        #         # inputs above.
+        #         output_dict["outputs"][var.name]["value"][case_number] = var.value
+        #     else:
+        #         output_dict["outputs"][var.name]["value"][case_number] = np.nan
```
eyoung55: @k1nshuk do we need to save the sweep params in `output_dict["outputs"]` as well?

k1nshuk: @eyoung55 not necessarily, since they are already stored separately in the output_dict. I suppose we could implement a filter that excludes them. 

The only difference between the sweep params values in `output_dict["outputs"]` and `output_dict["sweep_params"]` is they way they are obtained. The former is from the Pyomo model while the latter is from the `sweep_params` input dictionary to the parameter_sweep. Although, now that I think about it, that distinction doesn't really matter if everything within Pyomo and idaes works properly.

eyoung55: I guess obtaining them from the Pyomo model is a good sanity check to confirm that you're changing model values as expected, but as far as storing them in the final output file, maybe not necessary.

bknueven:
We should xfail the macOS tests -- when get IDAES extensions for macOS these will be reactivated.
```python
from idaes.config import bin_directory as idaes_bin_directory
solver = get_solver()
is_solver_from_idaes_ext = idaes_bin_directory in solver.executable()

def bad_test(self):
    if not is_solver_from_idaes_ext:
        pytest.xfail(
                "This test is known to be failing with solver: "
                f"{solver}, {solver.executable()}"
       )
```

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
k1nshuk: We do use the results object within the parameter sweep. Assertion for optimal termination is now handled within `do_param_sweep` in parameter_sweep.py and is contained within a try/except block.

watertap/tools/parameter_sweep.py
k1nshuk: It has been simplified as seen here.
https://github.com/watertap-org/watertap/blob/666d10ab5414fd603a4003b9e34402b35acfa372/watertap/tools/parameter_sweep.py#L37

This container is also tested in the pytests.

watertap/tools/parameter_sweep.py
k1nshuk: Based on our discussion, it was decided to keep it as a try/except block.

watertap/tools/parameter_sweep.py
k1nshuk: incorporated in a later commit.

watertap/tools/parameter_sweep.py
k1nshuk: incorporated.

watertap/tools/parameter_sweep.py
k1nshuk: incorporated
https://github.com/watertap-org/watertap/blob/666d10ab5414fd603a4003b9e34402b35acfa372/watertap/tools/parameter_sweep.py#L681

watertap/tools/parameter_sweep.py
k1nshuk: This function was removed in commit cc8ee3ed823f1e41fbf99499f33672e1c17ff20a

bknueven:
@lbianchi-lbl, @ksbeattie this PR has sections of code which are only executed when done in parallel. What's the possibility of getting a coverage report out of the MPI tests? If that's not worth the trouble, we should probably be prepared to merge this while failing the `codecov/patch` test.

lbianchi-lbl:
@bknueven I've looked into this a bit and these are my thoughts:

- I'm assuming the lines that are only run with MPI are the blocks under `if num_blocks > 1:` in `parameter_sweep.py`
- Adding coverage for the `MPI/tests` job by itself it's pretty straightforward
- However, I'm a bit torn whether we want to have our coverage results (including the reports from `main` that are used as the baseline for PRs) depending on a test environment that's quite different from the rest of our tests, where
  - The `mpi4py` is available
  - The test file is run "through" `mpirun`, i.e. `mpirun -n 2 pytest ...` instead of `pytest ...`
- Another way of thinking about this is that the MPI capability is (although not strictly required) a primary use case for the Parameter Sweep tool, which would warrant building the CI/testing infrastructure around it; but that's not necessarily true for the WaterTAP project as a whole
  - I feel this is a particular aspect of a more general discussion about the long-term role of the Parameter Sweep tool within the project and the wider ecosystem 
- One alternative would be to simply exclude the MPI-specific blocks from being counted towards coverage, which in practice would mean adding a `#pragma: no cover` comment after the `if` clause

lbianchi-lbl:
Following the discussion today, the most pragmatic thing at this point is probably to annotate the MPI-specific code so that those lines are ignored when computing the test coverage.

@k1nshuk, to do so, it should be enough to add an inline comment `# pragma: no cover` on the same line as the `if` clause for all the blocks that should be skipped, e.g.:

```py
def _aggregate_results(local_results, global_values, comm, num_procs):

    if num_procs > 1:  # pragma: no cover
        local_results = local_results.astype(np.float64)

        global_results = np.zeros((np.shape(global_values)[0], np.shape(local_results)[1]), dtype=np.float64)

        # Collect the number of result values to be sent from each process
        send_counts = np.zeros(num_procs, dtype=np.int64)
        comm.Gather(np.int64(np.size(local_results)), send_counts, root=0)

        # Collect the global results results onto rank 0
        comm.Gatherv(local_results, (global_results, send_counts), root=0)

        # Broadcast the results to all ranks
        comm.Bcast(global_results, root=0)

    else:
        global_results = np.copy(local_results)

    return global_results
```

More details here: https://coverage.readthedocs.io/en/6.3.1/excluding.html#advanced-exclusion

docs/how_to_guides/how_to_use_parameter_sweep.rst
eyoung55: Seeing errors like:
```bash
Failed example:
    parameter_sweep(m, sweep_params, outputs, csv_results_file='results.csv')
Expected nothing
Got:
    WARNING: Loading a SolverResults object with a warning status into
        model.name="unknown";
          - termination condition: infeasible
          - message from solver: Ipopt 3.13.4\x3a Converged to a locally
```
And clamping down the Pyomo logger blocks this output. Is there an easier/better fix?

watertap/examples/flowsheets/RO_with_energy_recovery/parameter_sweep_input_parser.py
eyoung55: Would the suggested fix work, I haven't worked with the logger before so this is copied from a similar use case and may not be the best approach:
```python
import idaes.logger as idaeslog
...
...
if _log.isEnabledFor(idaeslog.DEBUG):
    _log.debug(f'Property: {key}')
    _log.debug(f'New Value: {new_value:.6e}, Old Value: {old_value:.6e}')
```

docs/how_to_guides/how_to_use_parameter_sweep.rst
bknueven: The `parameter_sweep` function should probably be capturing this specific warning from Pyomo. Otherwise we'll have potentially a lot of these in some output file somewhere. I don't thing we currently have a use case for parameter sweep where we'd want to save the Ipopt solver log.

Around our calls to `simulate` or `optimize` let's try something like:
```python
from pyomo.common.tee import capture_output
with capture_output():
    optimize_fct(...)
```

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
bknueven: Any reason not do just use the `_default_optimize` function and leave this file as-it-is?

watertap/tools/parameter_sweep.py
```diff
@@ -319,108 +338,239 @@ def _interp_nan_values(global_values, global_results):
...
+    if hasattr(component, 'lb'):
+        comp_dict["lower bound"] = component.lb
+    if hasattr(component, 'ub'):
+        comp_dict["upper bound"] = component.lb
+    if hasattr(component, 'get_units'):
+        unit_obj = component.get_units()
+        if unit_obj is not None:
+            comp_dict["units"] = component.get_units().name
+        else:
+            comp_dict["units"] = "non-dimensional"
```
bknueven: We might just say "None" (or `None`). All we know in this case is that the user didn't *specify* units, not that the component doesn't have a unit in actually (i.e., is *dimensionless*, like ratio).

docs/how_to_guides/how_to_use_parameter_sweep.rst
bknueven: If we default the solver tee to `True`, we could also think about optionally putting this output in a file or database somewhere.

watertap/tools/parameter_sweep.py
```diff
@@ -319,108 +338,239 @@ def _interp_nan_values(global_values, global_results):
...
-        sweep_params: A dictionary containing the values to vary with the format
-                      ``sweep_params['Short/Pretty-print Name'] =
-                      (model.fs.variable_or_param[index], lower_limit, upper_limit, num_samples)``.
-                      A uniform number of samples ``num_samples`` will be take between
-                      the ``lower_limit`` and ``upper_limit``.
+    if outputs is None:
+        # No outputs are specified, so every Var, Expression, and Objective on the model should be saved
+        for pyo_obj in model.component_data_objects((pyo.Var, pyo.Expression, pyo.Objective), active=True):
+            # Only need to save this variable if it isn't one of the value in sweep_params
+            if pyo_obj.name not in var_str_list:
```
bknueven: I don't think this worth the effort. Why not just report out everything?

If we want to maintain the current behavior, `var_str_list` needs to be made into a `set`. Otherwise this for loop is quadratic in the number of `Var`s / `Constraint`s / `Expression`s.

watertap/tools/parameter_sweep.py
```diff
@@ -319,108 +338,239 @@ def _interp_nan_values(global_values, global_results):
...
+    # Get the inputs
+    op_ps_dict = output_dict["sweep_params"]
+    for key, item in sweep_params.items():
+        var_name = item.pyomo_object.name
+        op_ps_dict[var_name]['value'][case_number] = item.pyomo_object.value
 
-        debugging_data_dir (optional) : Save results on a per-process basis for parallel debugging
-                                        purposes. If None no `debugging` data will be saved.
+    # Get the outputs from model
+    if solver_termination_condition == "optimal":
```
bknueven: Why not use `pyo.check_optimal_termination_condition`? It will validate for more than "optimal", and if somebody is using something besides `ipopt` it may return an acceptable termination condition that's not "optimal".

bknueven (commented):
I think this is nearly there.

I'm still confused about *why* we're saving the termination conditions. In the main loop, if `assert_optimal_termination` is raises an Exception we'll just have `nan`s reported for this instance, right? Seems like unneeded overhead.

I also wonder (perhaps for a follow-on PR) if we're saving *everything* if it also doesn't make sense to save the *inputs* for a solve as well. E.g., the `value` of every `Var`. That way a user could in theory re-run and debug a "bad" sweep value.

watertap/tools/parameter_sweep.py
bknueven: Or better still, we should just use a `ComponentSet` from `pyomo.common.collections`, which can just store the pyomo objects directly, as opposed to its string representation.

k1nshuk:
> I think this is nearly there.
> 
> I'm still confused about _why_ we're saving the termination conditions. In the main loop, if `assert_optimal_termination` is raises an Exception we'll just have `nan`s reported for this instance, right? Seems like unneeded overhead.
> 
> I also wonder (perhaps for a follow-on PR) if we're saving _everything_ if it also doesn't make sense to save the _inputs_ for a solve as well. E.g., the `value` of every `Var`. That way a user could in theory re-run and debug a "bad" sweep value.

There are 2 primary reasons we are saving the termination conditions which will be addressed in a subsequent PR.

1. We want to identify and store the parameters that fail to solve for further study.
2. A subsequent implementation of the `parameter_seep` function will have the ability to ensure that we have a guaranteed number of successful solves, i.e., if a the user wants a parameter sweep of 5000 samples, we want to guarantee 5000 successful solves.


I do agree that in the next PR, we will expand the storage of variables to include everything. The current implementation, in effect, reflects my limited knowledge of Pyomo.

watertap/tools/parameter_sweep.py
k1nshuk: Changed non-dimensional to "None" in commit 2e87f5633e2ccb6bc512089d5cbb6703cb3a16fd

docs/how_to_guides/how_to_use_parameter_sweep.rst
eyoung55: Wrapped the calls to optimize in `capture_output` as suggested which removes the need to set the pyomo logger in the tested documentation. Marking as resolved.

watertap/examples/flowsheets/RO_with_energy_recovery/parameter_sweep_input_parser.py
eyoung55: Added the behavior shown above.

watertap/tools/parameter_sweep.py
k1nshuk: I have updated it to use `pyo.check_optimal_termination` in the commit 878aa3a1a74aaf9b79ef1ac65249bf0591801d0a. I agree that its probably a better way to deal with optimal solutions, although, looking at its definition, I don't think the values will change when using other optimizers besides IPOPT.

https://github.com/Pyomo/pyomo/blob/fa26176776e0d3f4f341d8b93c13ae554f3e8da2/pyomo/opt/results/solver.py#L129-L147

watertap/tools/parameter_sweep.py
k1nshuk: I agree that there is a better way of doing this but can we push it to a subsequent PR?

eyoung55: Yes, being consistent about whether we primarily operate on strings/names or pyomo objects would be good. Currently, there's enough of a reliance on strings for creating dictionaries that the `var_str_list` as a `set` works well here, but we could revisit it in the future.

docs/how_to_guides/how_to_use_parameter_sweep.rst
k1nshuk: Silent generation of output files was removed in 0fbc45fb5bed6d95313c65d709e83e6501041349

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
k1nshuk: Not sure about this one. `_default_optimize` existed in `parameter_sweep.py` before I started making changes. @eyoung55 what are your thoughts?

bknueven:
> > I think this is nearly there.
> 
> > 
> 
> > I'm still confused about _why_ we're saving the termination conditions. In the main loop, if `assert_optimal_termination` is raises an Exception we'll just have `nan`s reported for this instance, right? Seems like unneeded overhead.
> 
> > 
> 
> > I also wonder (perhaps for a follow-on PR) if we're saving _everything_ if it also doesn't make sense to save the _inputs_ for a solve as well. E.g., the `value` of every `Var`. That way a user could in theory re-run and debug a "bad" sweep value.
> 
> 
> 
> There are 2 primary reasons we are saving the termination conditions which will be addressed in a subsequent PR.
> 
> 
> 
> 1. We want to identify and store the parameters that fail to solve for further study.
> 
> 2. A subsequent implementation of the `parameter_seep` function will have the ability to ensure that we have a guaranteed number of successful solves, i.e., if a the user wants a parameter sweep of 5000 samples, we want to guarantee 5000 successful solves.

This is precisely my point--we don't really care about the termination condition. We can map that to a simple success/failure bool through `check_optimal_termination` and store that instead, which would be more useful for the case you describe.

k1nshuk:
> This is precisely my point--we don't really care about the termination condition. We can map that to a simple success/failure bool through `check_optimal_termination` and store that instead, which would be more useful for the case you describe.

I was assuming that we cared about knowing why some cases within the parameter sweep fail and that it could help with further investigation into the cause. The only drawback I see for storing strings vs bool is the added memory space. @avdudchenko What are your thoughts on storing the exact string value of the termination condition vs a bool for successful/failed for the parameter_sweep run cases?

watertap/examples/flowsheets/RO_with_energy_recovery/monte_carlo_sampling_RO_ERD.py
k1nshuk: Incorporated the ability to handle without or without the extension using the function here 
https://github.com/watertap-org/watertap/blob/6899183f16af7c5e9e195a4c8de6b53ff52acbc3/watertap/tools/parameter_sweep.py#L153-L158

bknueven:
I explain more on #342 why keeping the TerminationCondition is troublesome -- under certain circumstances you may not even have one, which the code as-it-is does not handle. Keeping this to a `bool` is simpler, and as somebody who debugs these models from time to time, loses no information.

If we really wanted to keep enough information to diagnose failure, we would really need at least the inputs to the solve (the variable values when `solve` is called), and possibly the Ipopt trace.

watertap/tools/parameter_sweep.py
andrewlee94: @k1nshuk Not all solvers return "optimal" as their final state given a successful solution. For example, some solvers will return "locally_optimal" instead - it has the same general meaning but would fail the `assert`.

k1nshuk: Right, I was just trying to say that Pyomo seems to be standardizing the optimal termination conditions to `optimal`, `locallyOptimal`, and `globallyOptimal`. I do agree that simply checking for "optimal" (what I had there) is not enough.

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
k1nshuk: My bad, I misunderstood your comment previously. The current test suite does pass when using `_default_optimize` in `parameter_sweep.py`. So it can be left in its original state. I know you added the option to check for optional termination condition in https://github.com/watertap-org/watertap/pull/342#issue-1147740717. I am fine either way.

watertap/tools/tests/test_parameter_sweep.py
```diff
@@ -369,16 +627,21 @@ def test_parameter_sweep_recover(self, model, tmp_path):
...
-                   'performance':m.fs.performance,
-                   'objective':m.objective}
+        A = m.fs.input['a']
+        B = m.fs.input['b']
+        sweep_params = {A.name : (A, 0.1, 0.9, 3),
+                        B.name : (B, 0.0, 0.5, 3)}
+        # outputs = {'output_c':m.fs.output['c'],
+        #            'output_d':m.fs.output['d'],
+        #            'performance':m.fs.performance,
+        #            'objective':m.objective}
```
bknueven:
```suggestion
```

watertap/tools/tests/test_parameter_sweep.py
```diff
@@ -409,16 +707,21 @@ def test_parameter_sweep_bad_recover(self, model, tmp_path):
...
```
bknueven:
```suggestion
```

watertap/tools/tests/test_parameter_sweep.py
```diff
@@ -436,9 +739,158 @@ def test_parameter_sweep_bad_recover(self, model, tmp_path):
...
+        m.fs.slack.setub(0)
+
+        A = m.fs.input['a']
+        B = m.fs.input['b']
+        sweep_params = {A.name : (A, 0.1, 0.9, 3),
+                        B.name : (B, 0.0, 0.5, 3)}
+        # outputs = {'output_c':m.fs.output['c'],
+        #            'output_d':m.fs.output['d'],
+        #            'performance':m.fs.performance,
+        #            'objective':m.objective}
```
bknueven:
```suggestion
```

watertap/examples/flowsheets/RO_with_energy_recovery/mc_sweep_params.yaml
```diff
@@ -0,0 +1,37 @@
...
+#   upper_limit: 4.5e-8
+#   num_samples: 4
+#================================================================
+# Spacer_porosity:
+#   type: LatinHypercubeSample
+#   # param: m.fs.RO.feed_side.properties_interface_out[0.0].conc_mass_phase_comp['Liq','H2O']
+#   param: m.fs.RO.spacer_porosity
+#   lower_limit: 0.95
+#   upper_limit: 0.99
+#================================================================
```
bknueven:
```suggestion
```

Maybe there's some value in maintaining the commented params here? I'm not sure, we don't reference this in the documentation

bknueven (commented):
We should remove the commented code, and maybe the commented params in the yaml file

bknueven (dismissed):
Looks good -- the ability to load from `yaml` and write to `h5` is a great new feature.

We should still really reorganize `parameter_sweep` into a class at some point, utilizing Pyomo Config blocks to deal with all the options we now have.

k1nshuk:
> Looks good -- the ability to load from `yaml` and write to `h5` is a great new feature.
> 
> We should still really reorganize `parameter_sweep` into a class at some point, utilizing Pyomo Config blocks to deal with all the options we now have.

Agreed, I was thinking of converting it into a class for the next PR myself.

eyoung55 (approved):
These additions look good to me!

bknueven (approved):

PR #306, opened 2022-01-25
Unit test harness

### LINKED ISSUE #66: Implement test_harness to reduce repetition in 0DRO testing
Due to the additional configuration options being added to the 0D RO model, the test file has a lot of repeated code that could be reduced by implementing test_harness.

TimBartholomew:
## Fixes/Addresses:

#66 

## Summary/Motivation:
Many of the unit model tests are copied and pasted for each unit model and its options. This test harness centralizes and standardizes the tests, improving the maintainability of the code and providing more clarity to the unit model developers.

- Creates a unit test harness that tests the following
- test statistics (i.e. number of variables, constraints, unused variables)
- test that units of measurement are consistent
- test the degrees of freedom
- test the initialization
- test the solve and that the report is as expected

### Legal Acknowledgement


TimBartholomew:
@adam-a-a @bknueven @andrewlee94 this PR currently just adds the unit test harness and provides one test for the RO demonstrating its use. Eventually all the unit model tests will be updated with it, but I'd like comments on it first.

TimBartholomew:
The test that checks the report is failing for linux, my best guess is that windows and linux terminals capture multi-line printing differently. I noticed for windows, in order to make the report output match what pytest captures, I had to manually add in trailing spaces on some lines.

andrewlee94:
@TimBartholomew String output checking is touchy at best, and I wonder if it is something that really should be automated like that.

bknueven:
This looks like it could be an upstream issue. On Linux and Python 3.9 on Windows, this is the diff:
```
   ====================================================================================
- Unit : unit                                                                Time: 0.0
?            ---

+ Unit : fs.unit                                                             Time: 0.0
?        +++
```

Why is one reporter printing the full name and the other only printing the local name?

lbianchi-lbl:
@TimBartholomew I agree with @andrewlee94 and would recommend to avoid string comparison unless absolutely necessary. I'm not familiar with the code in question but off the top of my head these could be alternatives to consider:

- Compare the two lists of (non-empty) output lines instead of the two multi-line strings
- (preferable) Produce or convert the report information as structured data (e.g. JSON, or "plain old data" Python object made of native types only), and use it for the comparison instead of the human-readable text output

In both cases, pytest should be able to deal with comparing composite objects properly (e.g. performing the diff recursively for nested dict/lists) out of the box; in other words, the more structured the objects being compared, the more detailed and granular the comparison, without requiring us implementing the comparison logic every time, which in my opinion is very useful.

ksbeattie:
This is being changed by PR #528, so will be revisited after that's merged.

TimBartholomew:
This needs to be reworked - its main test is based on report checking, but we have decided to not do that for WaterTAP.

wigging:
@TimBartholomew Since you started this merge request, can you resolve the merge conflicts?

wigging:
@TimBartholomew I fixed the merge conflict so ignore my previous comment.

adam-a-a:
Closing this PR since it would need to be reworked and has been open for almost 2 years. We can reopen it and bring it back to life with a new PR when the time comes.

PR #307, opened 2022-01-25
Adding zero-order pumps from WT3

andrewlee94:
## Summary/Motivation:
Adds pump model and a lot of cleanup and generalization.

- Rework base classes on new understanding of WT3 models
- Move more things into base classes
- Convert SIDO class into a set of callable methods, and add equivalent methods for pass-through units
- Pump models and tests
- Combining all pump data into a single data file

### Legal Acknowledgement


watertap/core/tests/test_zero_order_electricity.py
```diff
@@ -0,0 +1,138 @@
...
+    @pytest.mark.component
+    def test_solution(self, model):
+        assert pytest.approx(42*10, rel=1e-5) == value(
+            model.fs.unit.electricity[0])
+
+    @pytest.mark.component
+    def test_report(self, model):
+        stream = StringIO()
+
+        model.fs.unit.report(ostream=stream)
```
TimBartholomew, line 116: Thanks for the example :)

watertap/core/tests/test_zero_order_electricity.py
```diff
@@ -0,0 +1,138 @@
...
+    def test_private_attributes(self, model):
+        assert model.fs.unit._tech_type is None
+        assert model.fs.unit._has_recovery_removal is False
+        assert model.fs.unit._fixed_perf_vars == [
+            model.fs.unit.energy_electric_flow_vol_inlet]
+        assert model.fs.unit._initialize is None
+        assert model.fs.unit._scaling is None
+        assert model.fs.unit._get_Q is get_Q
+        assert model.fs.unit._perf_var_dict == {
+            "Electricity Demand": model.fs.unit.electricity,
```
TimBartholomew, line 88: I think this is fine for the PR, but maybe we should consider this variable to be more descriptive like "electricity_demand" or "demand_electric" or even switch out demand for consumption or use. Has this been standardized in IDAES?

watertap/core/zero_order_base.py
```diff
@@ -0,0 +1,277 @@
...
+                             solver=None, optarg=None)
+
+    def calculate_scaling_factors(self):
+        '''
+        Placeholder scaling routine, should be overloaded by derived classes
+        '''
+        super().calculate_scaling_factors()
+
+        if callable(self._scaling):
+            self._scaling(self)
```
TimBartholomew, line 136: Is this the unit specific calculate_scaling_factors()? I'd probably name it `_calculate_scaling_factors()`

watertap/core/zero_order_base.py
```diff
@@ -0,0 +1,277 @@
...
+            None
+        """
+        self.set_param_from_data(self.recovery_vol, data)
+
+        for t, j in self.removal_frac_mass_solute:
+            self.set_param_from_data(
+                self.removal_frac_mass_solute[t, j],
+                data,
+                index=j,
+                use_default_removal=use_default_removal)
```
TimBartholomew, line 188: I was imagining that we could have these `removal_frac_mass_solutes` also include water. That way the models could be full specified for the mass balance by either having a `removal_frac_mass_comp` for all components (including water), or for all except for water and then a `recovery_vol`. This would let us get away from setting `recovery_vol` for those units that really just remove components and leave the water alone.

This could be saved for another PR, but I'd like your thoughts.

TimBartholomew (dismissed):
LGTM. Mostly my comments were about naming and could be addressed in future PRs.

One higher-level comment is that I think expanding the removal fraction to include water, so that we can either specify the water removal or the `recovery_vol` to close the mass balance. I think it'd be helpful for all of those units that have 0.999 or 0.9999 `recovery_vol` because really those are just processes that don't remove water.

watertap/core/zero_order_base.py
andrewlee94: It is a private attribute that contains a reference to the unit specific `calculate_scaling_factors`. As a private attribute that most users will never see (let alone touch), I am not sure that naming really matters much.

watertap/core/zero_order_base.py
andrewlee94: I would suggest not doing that, as it leaves an opening for over-specifying the problem (i.e. fixing both the removal and recovery of water). For cases like you mentioned, there is nothing stopping us from having a different subtype of the material balances where water is unaffected (e.g. an optional argument that says that water recovery is assumed to be 100% so don't include it).

This also touches on the numerical issues associated with stream with zero water flow.

docs/technical_reference/core/pt_methods.rst
```diff
@@ -0,0 +1,49 @@
...
+
+Variables
+---------
+
+The build_pt method creates no additional variables beyond to those created by the `StateBlocks`.
+
+Constraints
+-----------
+
+The build_pt method writes no additional constraints beyond to those created by the `StateBlocks`.
```
adam-a-a:
```suggestion
The build_pt method writes no additional constraints beyond those created by the `StateBlocks`.
```

docs/technical_reference/core/pt_methods.rst
```diff
@@ -0,0 +1,49 @@
...
+
+Model Structure
+---------------
+
+The build_pt method constructs a simple representation of unit operation with a single inlet (named `inlet`) and outlet (named `outlet`) which have the same material state. A single `StateBlock` is constructed which is used for both the inlet and outlet Ports.
+
+Variables
+---------
+
+The build_pt method creates no additional variables beyond to those created by the `StateBlocks`.
```
adam-a-a:
```suggestion
The build_pt method creates no additional variables beyond those created by the `StateBlocks`.
```

docs/technical_reference/core/sido_methods.rst
```diff
@@ -0,0 +1,112 @@
+Single Inlet - Double Outlet Helper Methods
+===========================================
+
+.. index::
+   pair: watertap.core.zero_order_sido;build_sido
+
+.. currentmodule:: watertap.core.zero_order_sido
+
+The `build_sido` method is intended to be used to rapidly construct a standard set of material balance equations for zero-order type models with a signle inlet and two outlets.
```
adam-a-a:
```suggestion
The `build_sido` method is intended to be used to rapidly construct a standard set of material balance equations for zero-order type models with a single inlet and two outlets.
```

docs/technical_reference/core/sido_methods.rst
```diff
@@ -0,0 +1,112 @@
...
+
+where :math:`C_{t,j}` is mass concentration of solute :math:`j` at time :math:`t`.
+
+A similar constraint is then written for each solute to relate the solute concentration in the treated stream to that in the inlet stream.
+
+`solute_treated_equation(t, j)`:
+
+.. math:: (1-f_{t, j}) \times C_{inlet,t, j} = r_t \times C_{treated,t,j}
+
+Pressure balances are then written for each steam, including the :math:`\Delta P` terms if present.
```
adam-a-a:
```suggestion
Pressure balances are then written for each stream, including the :math:`\Delta P` terms if present.
```

docs/technical_reference/core/zero_order_unit.rst
```diff
@@ -2,129 +2,50 @@ Zero Order Unit Model Base Class
...
 
-The zero-order unit model base class is intended to form the basis for a library of generic single inlet-double outlet (SIDO) unit models using recovery and removal fractions to partition the incoming flow between the two outlets. This base class is not intended to be used by itself and requires the user to define a derived class before it can be used in a flowsheet.
+The zero-order unit model base class is intended to form the basis for a library of generic zero-order type (fixed performance) unit models. It defines a standard API for these types of unit operations and a set of common helper methods for building and working with these.
 
 Usage
 -----
 
+Ther ZeroOrderBaseData class contains no variables or constraints of its own - rather it is intended to be used as the starting point for construction of zero-order type models for specific unit operations.
+
+The cod below shows an outline of how the ZeroOrderBaseData class is intended to be used to develop custom zero-order type models.
```
adam-a-a:
```suggestion
The code below shows an outline of how the ZeroOrderBaseData class is intended to be used to develop custom zero-order type models.
```

docs/technical_reference/core/zo_form_libraries.rst
```diff
@@ -0,0 +1,11 @@
+Zero-Order Model Form Libraries
+===============================
+
+WaterTap includes a library of helper methods for assembling common model forms for zero-order type models. This library includes methods for common forms of material balances for different numbers of inlets and outlets, methods for including chemical additions and different approaches for determining electricity consumption. A list of the available methods is given below.
```
adam-a-a, line 4: Are methods for chemical additions a part of this PR too? Or this that for a subsequent PR?

andrewlee94: That will be the next PR.

watertap/data/techno_economic/pump.yaml
```diff
@@ -0,0 +1,115 @@
+default:
```
adam-a-a, line 1: This can be for a subsequent PR, but @kurbansitterley pointed out that there are several other yaml files (without the word "pump" in them) that are actually supposed to contain parameters for different pumps. 

Those should also be removed if redundant or combined here.

watertap/unit_models/zero_order/tests/test_pump_zo.py
```diff
@@ -0,0 +1,159 @@
...
+        m.fs.params = WaterParameterBlock(
+            default={"solute_list": ["sulfur", "toc", "tss"]})
+
+        m.fs.unit = PumpZO(default={
+            "property_package": m.fs.params,
+            "database": db})
+
+        return m
+
+    @pytest.mark.parametrize("subtype", [params.keys()])
```
adam-a-a, line 151: I am sure this works just fine, but I don't understand how it works exactly. My understanding is that you somehow load in all pump subtypes from the yaml file, but I don't see the connection where that happens.

watertap/core/tests/test_zero_order_base.py
```diff
@@ -0,0 +1,341 @@
...
+# Technology Laboratory (subject to receipt of any required approvals from
+# the U.S. Dept. of Energy). All rights reserved.
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and license
+# information, respectively. These files are also available online at the URL
+# "https://github.com/watertap-org/watertap/"
+#
+###############################################################################
+"""
+Tests for general zero-order property package
```
adam-a-a:
```suggestion
Tests for general zero-order base class
```

watertap/core/tests/test_zero_order_base.py
```diff
@@ -0,0 +1,341 @@
...
+    m.fs.params = WaterParameterBlock(default={"solute_list": ["A", "B", "C"]})
+
+    m.fs.unit = DerivedZOBase(default={"property_package": m.fs.params})
+
+    assert m.fs.unit._tech_type is None
+    assert m.fs.unit._has_recovery_removal is False
+    assert m.fs.unit._fixed_perf_vars == []
+    assert m.fs.unit._initialize is None
+    assert m.fs.unit._scaling is None
+    assert m.fs.unit._get_Q is None
```
adam-a-a, line 46: Why do we need `_get_Q`? I noticed that the unit models will have their own `get_Q` method that just grabs the inlet flowrate, but I didn't see how it was necessary at this point. Was this done in preparation for shifting to a mass flow basis in the property model in a subsequent PR?

watertap/core/tests/test_zero_order_base.py
```diff
@@ -0,0 +1,341 @@
...
+        model.fs.unit.removal_frac_mass_solute = Var(
+            model.fs.time, model.fs.params.solute_set)
+
+        model.fs.unit.set_param_from_data(
+            model.fs.unit.removal_frac_mass_solute[0, "A"],
+            {"removal_frac_mass_solute": {"A": {
+                "value": 0.42, "units": "m^3/m^3"}},
+             "default_removal_frac_mass_solute": {
+                "value": 0.70, "units": "kg/kg"}},
+            index="D",
```
adam-a-a, line 312: I didn't catch why there is an `index="D"` here.

adam-a-a: Nor why it doesn't raise an error. 

If you added `assert model.fs.unit.removal_frac_mass_solute[0, "D"].value == 0.7`, I am guessing that would be True. Is the assignment of index D for "default_removal_frac_mass_solute"?

watertap/core/zero_order_pt.py
```diff
@@ -0,0 +1,141 @@
...
+# Set up logger
+_log = idaeslog.getLogger(__name__)
+
+
+def build_pt(self):
+    """
+    Helper method for constructing material balances for zero-order type models
+    with pass-through behavior.
+
+    One StateBlocks is added with two corresponding Ports:
```
adam-a-a:
```suggestion
    One StateBlock is added with two corresponding Ports:
```

watertap/core/zero_order_pt.py
```diff
@@ -0,0 +1,141 @@
...
+    tmp_dict["defined_state"] = True
+
+    self.properties = self.config.property_package.build_state_block(
+        self.flowsheet().time,
+        doc="Material properties in unit",
+        default=tmp_dict)
+
+    # Create Ports
+    self.add_port("inlet", self.properties, doc="Inlet port")
+    self.add_port("outlet", self.properties, doc="Tutlet port")
```
adam-a-a:
```suggestion
    self.add_port("outlet", self.properties, doc="Outlet port")
```

watertap/unit_models/zero_order/tests/test_pump_zo.py
andrewlee94: There is some code at the module level above that gets the parameter dict, from which we create a list of top-level keys (the sub-types). This then tells `pytest` to run the test multiple times, each time using a different value from the list for the `subtype` argument in the line below.

watertap/core/tests/test_zero_order_base.py
andrewlee94: This is needed because pass-through units only have one StateBlock named `properties` - this method is there to get around the different names.

watertap/core/tests/test_zero_order_base.py
andrewlee94: I think the point was to use an index that didn't actually exist - there is no index D, but I am telling it to get the value for it anyway and assign that to the variable for index A (just so that I can check the result). As index does not exist, it should get the default value, rather than the specific value assigned for component A.

watertap/core/tests/test_zero_order_electricity.py
andrewlee94: The closest we have to a standard comes from DISPATCHES, where they used `electricity`. I will note that we have `heat` and `work` terms already as well.

watertap/data/techno_economic/pump.yaml
andrewlee94: I would need a list of those files to know what to include. @kurbansitterley could you provide a list of any other data files for pumps that should be included here?

kurbansitterley: From a costing perspective, in addition to anything with "pump" in the name, these are all pumps:

- agriculture
- discharge
- industrial
- irrigation
- mining

andrewlee94: @kurbansitterley Do you have suggestions for better names for those sub-types? "agriculature" is not really a sub-type for a pump (hence why I missed it in the first pass).

andrewlee94: Even more so - all those subtypes use the same values so we should have a general name that covers all those cases.

kurbansitterley: These are all identical in costing to the separate water_pumping_station module, so you could just use that. Or, something like "use_pumping_station" since these were all the last units in the treatment train to pump the water to their end use (i.e. for agriculture or mining).

andrewlee94:
@adam-a-a @TimBartholomew This PR should be ready for review now.

watertap/core/tests/test_zero_order_electricity.py
adam-a-a: Personally, I prefer `electricity_consumption`, but in the end it's just a name.

watertap/core/zero_order_sido.py
```diff
@@ -11,18 +11,14 @@
 #
 ###############################################################################
 """
-This module contains the base class for all zero order single inlet-double
-outlet (SIDO) unit models.
+This module contains methods for constructung the materail balances for
```
adam-a-a:
```suggestion
This module contains methods for constructing the material balances for
```

adam-a-a (approved):
LGTM

TimBartholomew (approved):
LGTM

PR #308, opened 2022-01-26
Adding missing recovery and removal rates for non-basic units. 

avdudchenko:
This includes all units in the water recovery file that were not in the basic_units. Same structure as before. 

Please let me know if anything else is missing and I will add it into the commit. (probably good idea before doing full merge)

## Fixes/Addresses:

-
-

### Legal Acknowledgement


andrewlee94 (dismissed):
I think these will best be reviewed as we work through the unit models.

andrewlee94:
@avdudchenko We are seeing test failures because of the use of `unit` instead of `units` for the units of measurement. I added some basic validation tests for the data files which check for things we expect to be there, and whilst dong so I changed all instances of `unit` to `units` (as units are inherently plural). However, the new files are all using your original form of `unit`, and so are failing the validation tests.

andrewlee94:
@adam-a-a @bknueven @kurbansitterley I believe this is ready for review now.

watertap/data/techno_economic/agglom_stacking.yaml
```diff
@@ -0,0 +1,8 @@
+default:
+  recovery_vol:
+    value: 0.9999
+    units: dimensionless
+    reference: .nan
```
bknueven, line 5: This occurs a few places. Given this is a string field I'd be happier with nearly anything else.

bknueven (approved):
I have a minor comment, but it shouldn't hold this up.

watertap/data/techno_economic/agglom_stacking.yaml
andrewlee94: That is on the list of things to clean up, but I was thinking of doing that on a case-by-case basis as we did the unit models.

kurbansitterley (approved):
No reason to port these from WT3:

- cartridge_filtration_with_backflush
- chlorine_addition_cylinder
- ro_deep
- ro_deep_guo
- ro_deep_mcgivney
- ro_deep_scnd_pass
- gac_gravity_60_min
- gac_gravity_30_min
- injection_well
- nuclear_cooling_tower
- tri_media_filtration_with_backflush

These weren't on the list I sent of required units from WT3. I suspect were created from an old water_recovery.csv or constituent_removal.csv... we are still using WT3 and I have cleaned up these files recently. 

But LGTM otherwise. These files could be merged and removed later or removed now, doesn't matter to me.

PR #311, opened 2022-01-28
Ion property package

adam-a-a:
## Summary/Motivation:
This PR provides a property model to be used with the future DSPM-DE model for NF and, with some subsequent additions, the future electrodialysis model (and others that may depend upon the extended Nernst Planck equation).

- tests for ion prop model
- ion prop model that allows user to define a set of solutes and provide data for MW, ion diffusivity, Stokes radius of each ion, etc., as relevant for the DSPM-DE model.
- also added a method to `assert_electroneutrality`, where a certain tolerance can be provided.

### Legal Acknowledgement


watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -0,0 +1,838 @@
...
+"""
+
+# Import Python libraries
+import idaes.logger as idaeslog
+
+from enum import Enum, auto
+# Import Pyomo libraries
+from pyomo.environ import Constraint, Expression, Reals, NonNegativeReals, \
+    Var, Param, Suffix, value, check_optimal_termination
+from pyomo.environ import units as pyunits
```
andrewlee94: You can combine this line with the one above.

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -0,0 +1,838 @@
...
+        else: #TODO only works if no data provided at all, but should have conditional that handles one or more missing vals
+            _log.warning(f"Missing initial configuration values for {data}. All values being arbitrarily set to {default}. "
+                         f"Ensure correct data values are assigned before solving.")
+            return default
+
+    def build(self):
+        '''
+        Callable method for Block construction.
+        '''
+        super(DSPMDEParameterData, self).build()
```
andrewlee94, line 100: This can be shortened to `super().build()`

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -0,0 +1,838 @@
...
+
+        # reference
+        # Todo: enter any relevant references
+
+
+        # molecular weight
+        self.mw_comp = Param(
+            self.component_list,
+            mutable=True,
+            initialize=self._init_param_data('mw_data'),
```
andrewlee94: Do you need the `_init_param_data` method here? I Think you might be able to achieve all of what you want (including handling missing values using:

```
self.mw_comp = Param(
            self.component_list,
            mutable=True,
            initialize=self.config.mw_data,
            default=# add default value here,
            units=pyunits.kg/pyunits.mol,
            doc="Molecular weight kg/mol")
```

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -0,0 +1,838 @@
...
+            units=pyunits.kg/pyunits.mol,
+            doc="Molecular weight kg/mol")
```
andrewlee94: As a general rule I avoid putting units in the doc string now that the component itself has units. This just duplicates information and adds a potential point where data can get out of synch,

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -0,0 +1,838 @@
...
+        # Set solver and options
+        opt = get_solver(solver, optarg)
+
+        # Fix state variables
+        flags = fix_state_vars(self, state_args)
+        # Check when the state vars are fixed already result in dof 0
+        for k in self.keys():
+            dof = degrees_of_freedom(self[k])
+            if dof != 0:
+                raise PropertyPackageError("\nWhile initializing {sb_name}, the degrees of freedom "
```
andrewlee94: I suggest replacing this with the new `InitializationError`.

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -0,0 +1,838 @@
...
+        for k in self.keys():
+            if number_unfixed_variables(self[k]) != 0:
+                skip_solve = False
+
+        if not skip_solve:
+            # Initialize properties
+            with idaeslog.solver_log(solve_log, idaeslog.DEBUG) as slc:
+                results = solve_indexed_blocks(opt, [self], tee=slc.tee)
+            init_log.info_high("Property initialization: {}.".format(idaeslog.condition(results)))
+
```
andrewlee94, line 291: You should also add a check here that the final solve is optimal (or `None` if skipped), and raise an `InitializationError` if this is not the case.

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -0,0 +1,838 @@
...
+            bounds=(1e-8, None),
+            domain=NonNegativeReals,
+            units=pyunits.mol/pyunits.s,
+            doc='Mole flow rate')
+
+        self.temperature = Var(
+            initialize=298.15,
+            bounds=(273.15, 373.15),
+            domain=NonNegativeReals,
+            units=pyunits.degK,
```
andrewlee94: Minor point, but you can use just `K` here (and it is technically more correct - for whatever reason, Kelvins are not meant to be used with "degrees").

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -0,0 +1,838 @@
...
+            doc='State pressure')
+
+    # -----------------------------------------------------------------------------
+    # Property Methods
+    def _mass_frac_phase_comp(self):
+        self.mass_frac_phase_comp = Var(
+            self.params.phase_list,
+            self.params.component_list,
+            initialize=lambda b,p,j : 0.4037 if j == "H2O" else 0.0033, #todo: revisit
+            bounds=(1e-6, None),  # upper bound set to None because of stability benefits
```
andrewlee94: In IDAES, we have generally started making the upper bound 1.001 or something similar (i.e. have an upper bound, but give it a little slack).

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -0,0 +1,838 @@
...
+        def rule_molality_comp(b, j):
+            return (b.molality_comp[j] ==
+                    b.flow_mol_phase_comp['Liq', j]
+                    / b.flow_mol_phase_comp['Liq', 'H2O']
+                    / b.params.mw_comp['H2O'])
+
+        self.eq_molality_comp = Constraint(self.params.solute_set, rule=rule_molality_comp)
+
+    def _radius_stokes_comp(self):
+        add_object_reference(self, "radius_stokes_comp", self.params.radius_stokes_comp)
```
andrewlee94, line 559: This is the correct way to deal with these.

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -0,0 +1,838 @@
...
+    # def get_material_density_terms(self, p, j):
+    #     """Create material density terms."""
+
+    # def get_enthalpy_density_terms(self, p):
+    #     """Create enthalpy density terms."""
+
+    def default_material_balance_type(self):
+        return MaterialBalanceType.componentTotal
+
+    # def default_energy_balance_type(self):
```
andrewlee94, line 632: Commented code - you probably want to include this term, if only for future extended future use of this model.

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -0,0 +1,838 @@
...
+    # -----------------------------------------------------------------------------
+    # General Methods
+    # NOTE: For scaling in the control volume to work properly, these methods must
+    # return a pyomo Var or Expression
+
+    def get_material_flow_terms(self, p, j):
+        """Create material flow terms for control volume."""
+        return self.flow_mol_phase_comp[p, j]
+
+    # def get_enthalpy_flow_terms(self, p):
```
andrewlee94, line 617: Commented code - some of this probably should be included (e.g. enthalpy flow terms). You might not need it now, but who knows about the future,

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -0,0 +1,838 @@
...
+
+        # molecular weight
+        self.mw_comp = Param(
+            self.component_list,
+            mutable=True,
+            initialize=self._init_param_data('mw_data'),
+            units=pyunits.kg/pyunits.mol,
+            doc="Molecular weight kg/mol")
+        # Stokes radius
+        self.radius_stokes_comp = Param(
```
andrewlee94, line 126: Do you really want these to be `Params`? It might be fine for now, but it does limit you in terms of future parameter estimation/sensitivity studies,

andrewlee94 (commented):
A few comments, of which only one or two are really critical (`InitializationErrors`), but a few ideas on things that might work better.

watertap/property_models/ion_DSPMDE_prop_pack.py
adam-a-a: I thought about that but then figured that things like molecular weight should just be parameters. I _could_ change these to vars, but I'd prefer to leave them as is and add TODO notes to consider changing to vars in a subsequent PR. Right now, I am using this property model as is to finish implementing the nanofiltration DSPM-DE model, which is having issues, so I'd rather avoid making changes to params/variables until I resolve issues with DSPM-DE.

watertap/property_models/ion_DSPMDE_prop_pack.py
adam-a-a: I'll add a TODO note on this for a subsequent PR, but I think the main priority now is to get the DSPM-DE model working with this property model first.

watertap/property_models/ion_DSPMDE_prop_pack.py
adam-a-a: I'll add a TODO note on this for a subsequent PR, but I think the main priority now is to get the DSPM-DE model working with this property model first.

watertap/property_models/ion_DSPMDE_prop_pack.py
adam-a-a: OK, I'll add upper bounds instead of None.

watertap/property_models/ion_DSPMDE_prop_pack.py
adam-a-a: Will do.

watertap/property_models/ion_DSPMDE_prop_pack.py
adam-a-a: We added `fail_flag` to a couple utilities (`check_solve`) to provide the option of only providing a warning (when desired) as opposed to an exception. What do you think about adding `InitializationError` to the utility functions in util/initialization instead?

andrewlee94: The IDAES standard is now to raise an `InitializationError` here - the flowsheet can then choose to ignore it if it wants (with a `try/except`). The point was to make sure the end-user was aware of the issue and acknowledge it (by either fixing the problem or doing the `try/except`).

The problem with the utilities is it starts to bury this behaviour deeper so that it is hidden and less obvious what is happening.

adam-a-a: OK, I added the `InitializationError` in.

adam-a-a:
@andrewlee94 I think I addressed all of your comments. Let me know if you have any other suggestions.

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -0,0 +1,814 @@
...
+# Set up logger
+_log = idaeslog.getLogger(__name__)
+
+
+class ActivityCoefficientModel(Enum):
+    ideal = auto()                    # Ideal
+    davies = auto()                   # Davies
+
+
+@declare_process_block_class("DSPMDEParameterBlock") #TODO: replace w/ "PropParameterBlock"
```
aladshaw3:
```suggestion
@declare_process_block_class("PropParameterBlock")
```

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -0,0 +1,814 @@
...
+                var = getattr(self[k], v_name)
+                var[ind].unfix()
+
+        # fix state variables if hold_state
+        if hold_state:
+            fix_state_vars(self)
+
+        return results
+
+@declare_process_block_class("DSPMDEStateBlock",
```
aladshaw3, line 404:
```suggestion
@declare_process_block_class("PropStateBlock",
```

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -0,0 +1,814 @@
...
+           "``ActivityCoefficientModel.davies``", "Activity coefficients estimated via Davies model"
+       """))
+
+    def build(self):
+        '''
+        Callable method for Block construction.
+        '''
+        super(DSPMDEParameterData, self).build()
+
+        self._state_block_class = DSPMDEStateBlock
```
aladshaw3, line 102:
```suggestion
        self._state_block_class = PropStateBlock
```

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -0,0 +1,814 @@
...
+            fix_state_vars(self)
+
+        return results
+
+@declare_process_block_class("DSPMDEStateBlock",
+                             block_class=_DSPMDEStateBlock)
+class DSPMDEStateBlockData(StateBlockData):
+    def build(self):
+        """Callable method for Block construction."""
+        super().build()
```
aladshaw3, line 409:
```suggestion
        super(DSPMDEStateBlockData, self).build()
```

watertap/property_models/tests/test_ion_DSPMDE_prop_pack.py
```diff
@@ -0,0 +1,450 @@
...
+                        PhysicalParameterBlock,
+                        StateBlockData,
+                        StateBlock,
+                        MaterialBalanceType,
+                        EnergyBalanceType,
+                        )
+from idaes.core.util.scaling import calculate_scaling_factors, set_scaling_factor, get_scaling_factor
+
+from pyomo.util.check_units import assert_units_consistent
+from watertap.property_models.ion_DSPMDE_prop_pack import DSPMDEParameterBlock, DSPMDEStateBlock, ActivityCoefficientModel
```
aladshaw3, line 35:
```suggestion
from watertap.property_models.ion_DSPMDE_prop_pack import PropParameterBlock, PropStateBlock, ActivityCoefficientModel
```

watertap/property_models/tests/test_ion_DSPMDE_prop_pack.py
```diff
@@ -0,0 +1,450 @@
...
+
+solver = get_solver()
+# -----------------------------------------------------------------------------
+
+@pytest.fixture(scope="module")
+def model():
+    m = ConcreteModel()
+
+    m.fs = FlowsheetBlock(default={"dynamic": False})
+    m.fs.properties = DSPMDEParameterBlock(default={
```
aladshaw3, line 55:
```suggestion
    m.fs.properties = PropParameterBlock(default={
```

watertap/property_models/tests/test_ion_DSPMDE_prop_pack.py
```diff
@@ -0,0 +1,450 @@
...
+        assert j in ["H2O"]
+    assert isinstance(model.fs.properties.solute_set, Set)
+    for j in model.fs.properties.solute_set:
+        assert j in ["A", "B", "C", "D"]
+
+    assert isinstance(model.fs.properties.phase_list, Set)
+    for j in model.fs.properties.phase_list:
+        assert j in ["Liq"]
+
+    assert model.fs.properties._state_block_class is DSPMDEStateBlock
```
aladshaw3, line 94:
```suggestion
    assert model.fs.properties._state_block_class is PropStateBlock
```

watertap/property_models/tests/test_ion_DSPMDE_prop_pack.py
```diff
@@ -0,0 +1,450 @@
...
+    assert value(m.fs.stream[0].dens_mass_phase['Liq']) == pytest.approx(1001.76,  rel=1e-3)
+
+    assert value(m.fs.stream[0].act_coeff_phase_comp['Liq', 'A']) == 1
+
+@pytest.fixture(scope="module")
+def model2():
+    m = ConcreteModel()
+
+    m.fs = FlowsheetBlock(default={"dynamic": False})
+    m.fs.properties = DSPMDEParameterBlock(default={
```
aladshaw3, line 177:
```suggestion
    m.fs.properties = PropParameterBlock(default={
```

watertap/property_models/tests/test_ion_DSPMDE_prop_pack.py
```diff
@@ -0,0 +1,450 @@
...
+    assert_optimal_termination(results)
+
+    assert value(stream[0].flow_vol_phase['Liq']) == pytest.approx(1.91524e-5,  rel=1e-3)
+
+@pytest.fixture(scope="class")
+def model3():
+    m = ConcreteModel()
+    m.fs = FlowsheetBlock(default={"dynamic": False})
+
+    m.fs.properties = DSPMDEParameterBlock(default={
```
aladshaw3, line 246:
```suggestion
    m.fs.properties = PropParameterBlock(default={
```

watertap/property_models/tests/test_ion_DSPMDE_prop_pack.py
```diff
@@ -0,0 +1,450 @@
...
+        assert get_scaling_factor(m.fs.stream[0].diffus_phase_comp['Liq', j]) is not None
+        assert get_scaling_factor(m.fs.stream[0].act_coeff_phase_comp['Liq', j]) is not None
+    assert get_scaling_factor(m.fs.stream[0].dens_mass_phase['Liq']) is not None
+    assert get_scaling_factor(m.fs.stream[0].visc_d_phase['Liq']) is not None
+
+@pytest.mark.component
+def test_seawater_data():
+    m = ConcreteModel()
+    m.fs = FlowsheetBlock(default={'dynamic':False})
+    m.fs.properties = DSPMDEParameterBlock(default={
```
aladshaw3, line 371:
```suggestion
    m.fs.properties = PropParameterBlock(default={
```

aladshaw3 (changes requested):
My suggestions are based on your first TODO note and what Tim had mentioned before about standardizing the class names of property packages. It is probably better to go with the standardized names now (if that is what we are going to be doing), then have to make a bunch of changes in the code later.

watertap/property_models/ion_DSPMDE_prop_pack.py
andrewlee94: Do we really want to standardize these? Whilst it might sound like a good idea now, I can see it causing confusion later. Consider a case where I want to use two property packages in a flowsheet, and they both use standard names - now they need to be imported with aliases (so the user is effectively giving them unique names anyway).

adam-a-a: I had the same thought upon further reflection. Maybe it's best to keep this as is instead of trying to standardize.

watertap/property_models/tests/test_ion_DSPMDE_prop_pack.py
adam-a-a: > My suggestions are based on your first TODO note and what Tim had mentioned before about standardizing the class names of property packages. It is probably better to go with the standardized names now (if that is what we are going to be doing), then have to make a bunch of changes in the code later.

@aladshaw3 - given @andrewlee94 's feedback and thinking about it more, it might be best to leave things as they were instead of trying to standardize (i.e., do not shift to using `PropStateBlock` instead of `DSPMDEStateBlock`). If we decide that's best for now, I'd just delete the to do comment.

aladshaw3: Ok, that sounds fine with me.

aladshaw3 (dismissed):
This looks good enough to me for now. It would be good to get this merged so that others who will be working on unit models this depends on can get started as soon as possible. We can modify this more as time goes on.

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -0,0 +1,814 @@
...
+        skip_solve = True  # skip solve if only state variables are present
+        for k in self.keys():
+            if number_unfixed_variables(self[k]) != 0:
+                skip_solve = False
+
+        if not skip_solve:
+            # Initialize properties
+            with idaeslog.solver_log(solve_log, idaeslog.DEBUG) as slc:
+                results = solve_indexed_blocks(opt, [self], tee=slc.tee)
+                if not assert_optimal_termination:
```
andrewlee94: Two issues here:

1. This needs an argument of the `results` object.
2. You should use `check_optimal_convergence` instead of `assert_optimal_convergence` (otherwise you will get a Pyomo error here instead of the following `InitializationError`.

So: `if not check_optimal_convergence(results)`

andrewlee94 (dismissed):

watertap/property_models/ion_DSPMDE_prop_pack.py
adam-a-a: @andrewlee94 - Oops. I just committed the fix.

andrewlee94 (approved):

aladshaw3 (approved):
LGTM

PR #314, opened 2022-02-01
Update requirement for idaes-pse to latest watertap-org tag

lbianchi-lbl:
## Fixes/Addresses:

- Issues with Click 8 compatibility after IDAES/idaes-pse#683

### Legal Acknowledgement

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,7 +43,7 @@ SPECIAL_DEPENDENCIES_FOR_PRERELEASE = [
     # update with a tag from the nawi-hub/idaes-pse
     # when a version of IDAES newer than the latest stable release from PyPI
     # will become needed for the watertap development
-    "idaes-pse[prerelease] @ https://github.com/watertap-org/idaes-pse/archive/1.12.0.watertap.2022.01.06.zip",
+    "idaes-pse[prerelease] @ https://github.com/watertap-org/idaes-pse/archive/1.12.1.watertap.2022.01.31.zip",
 ]
 
 # Arguments marked as "Required" below must be included for upload to PyPI.


aladshaw3:
Looks like whatever was changed in the new IDAES version broke a couple of the chemistry unit tests.

andrewlee94:
@aladshaw3 It is not so much that a change in IDAES broke these models, but that we are now more active in reporting issues that were already there (and raising exceptions). These failures are appearing because initialization routines are now raising `InitializationErrors` if they fail to converge at the final step, which were passing unnoticed before.

It should be noted that this isn't always a terminal failure - these models worked in the past by ignoring the convergence failure and moving on, assuming that subsequent steps would resolve the issue. You can still do the same thing, but now you will need to explicitly acknowledge the initialization failure using a `try/except pass` statement. Alternatively, you can try to determine what is causing the failure in initialization and address that.

aladshaw3:
@andrewlee94 I will dig into this some today and see if I can resolve these issues. If I can fix this in some way, I will open another PR to merge before this gets merged.

bknueven:
What often happens here is Ipopt converges to a "pretty good" but infeasible point (e.g., NLP error ~1e-5), and reinitializing the dual values and re-scaling around that point is enough to get a few more orders of magnitude of feasibility. This is why these test pass -- the "solve" step is usually just solving the unit.

The RO unit models take advantage of this fact and will do a [second pass at solving the unit if the first is infeasible](https://github.com/watertap-org/watertap/blob/d80de3bb8642e2f0333a30b076754d8670a4e8c5/watertap/unit_models/reverse_osmosis_1D.py#L764-L770).

As for this PR, I would be in favor of just putting `try/except`  blocks around the affected units. We can open an issue and revisit.

aladshaw3:
@bknueven I was able to fix 2 of the initialization errors just by playing with the solver tolerance just before the solve (see PR #316 ). However, there is another error that I believe is mostly on the IDAES side where the initializer is reporting 'too few dof' for the stoichiometric reactor (which was never an issue before). I believe @andrewlee94 is looking into that now.

andrewlee94:
@aladshaw3 I found your problem - you are leaving one of the inlet concentrations unfixed when setting up the problem (`Ca(OH)2`), which violates the assumption we make when initializing the problem. The underlying assumption during initialization is that the model will be square if we fix all the inlet conditions and solve (as normally the upstream feed is known). Thus, during initialization we fix all the feed compositions, meaning in this case you loose a degree of freedom.

I am actually not sure how this actually worked previously, but the problem is in the specification of your problem, not in IDAES.

aladshaw3:
@lbianchi-lbl  I have the fix for the errors on PR #316. Should we just merge that PR instead?

lbianchi-lbl:
> @lbianchi-lbl I have the fix for the errors on PR #316. Should we just merge that PR instead?

Yeah, please go ahead and merge that instead. It's even possible that this PR will be closed automatically by GitHub once #316 is in `main` since its changes are a strict subset of #316 (or something).

aladshaw3:
#316 did not automatically close this PR, so I am closing it now.

PR #315, opened 2022-02-01
Add first pass of tests for EDB CLI

### LINKED ISSUE #304: EDB CLI are not currently tested
None of the functions in `commands.py` ([here](https://github.com/watertap-org/watertap/blob/main/watertap/edb/commands.py)) are currently unit tested. Is there a way to create unit tests for these command line functions?

lbianchi-lbl:
- Add tests for the EDB CLI

### Legal Acknowledgement


bknueven (approved):
These files were severely under-covered, so this should merged.

The diff coverage is clearly not an issue -- the project coverage goes up a whole percentage point with this PR.

aladshaw3 (approved):
LGTM

aladshaw3:
@lbianchi-lbl Looks like we need to temporarily disable the check on `codecov/patch` or loosen up the coverage target for this case.

lbianchi-lbl:
> @lbianchi-lbl Looks like we need to temporarily disable the check on `codecov/patch` or loosen up the coverage target for this case. 

Since this should be a one-off corner case, I've just red-button merged it in.

PR #320, opened 2022-02-03
Refactor of zero-order models to mass flow basis

andrewlee94:
## Summary/Motivation:
The current formulation of using volumetric flow and concentration as the state variables to define streams fails badly when flow goes to zero. This PR refactors most of the code to use a mass flow basis instead.

Note that there is a question about what to do with the old `recovery_vol` parameter - this PR proposes renaming it to `recovery_frac_mass_H2O` - it is wordy, but follows the naming convention for removal fractions. This would involve renaming the parameter in all the yaml files, so I am holding off on doing that until we have agreement.

- Lots

### Legal Acknowledgement


bknueven (approved):
Mostly a rubber stamp -- I don't see anything that should hold this up.

watertap/core/zero_order_properties.py
```diff
@@ -140,11 +129,10 @@ def define_metadata(cls, obj):
...
 
         obj.add_properties(
-            {'flow_vol': {'method': None},
-             'conc_mass_comp': {'method': None},
-             'cp_mass': {'method': '_cp_mass'},
-             'dens_mass': {'method': '_dens_mass'},
-             'flow_mass_comp': {'method': None}})
+            {'flow_mass_comp': {'method': None},
+             'flow_vol': {'method': '_flow_vol'},
+             'conc_mass_comp': {'method': '_conc_mass_comp'},
```
TimBartholomew, line 134: Not for this PR, but I think it makes sense to have an additional on-demand property for mass fraction. Especially since that is essentially how the conc_mass_comp is calculated as mass_frac_comp * density.

watertap/unit_models/zero_order/feed_zo.py
```diff
@@ -39,6 +42,37 @@ class FeedZOData(FeedData):
     def build(self):
         super().build()
 
+        units = self.config.property_package.get_metadata().get_derived_units
```
TimBartholomew, line 45: I always imagined that we would just use the IDAES feedblock and then use the calculate_state method that I added to WaterTAP property packages so that users explicitly call a calculation to fix the state variables given the flowrate and concentrations. But I see what you did here and that works too.

watertap/unit_models/zero_order/feed_zo.py
```diff
@@ -80,3 +114,59 @@ def load_feed_data_from_database(self, overwrite=False):
...
                               f"database water source. Value was not fixed.")
+
+        # Set initial values for mass flows in properties based on these
+        # Assuming density of 1000
+        for t in self.flowsheet().time:
+            for j in self.properties[t].params.solute_set:
+                self.properties[t].flow_mass_comp[j].set_value(
+                    value(self.flow_vol[t] * self.conc_mass_comp[t, j]))
+            self.properties[t].flow_mass_comp["H2O"].set_value(
+                value(self.flow_vol[t]*1000))
```
TimBartholomew, line 125: Why are we hard coding the 1000 here? Can't we use the default density in the property model? The initial value isn't a big deal and this is close enough, just was curious.

TimBartholomew (approved):
LGTM, some minor comments but I don't think they should hold up the PR

PR #322, opened 2022-02-09
Initial Coagulation/Flocculation Property Package

aladshaw3:
## Fixes/Addresses:

Creates the first iteration of a coagulation/flocculation property packaged needed for the unit models for coagulation/flocculation and settling tank. 

## Summary/Motivation:

Necessary to track state vars for TSS, TDS, and Sludge produced from a coagulation process. These will be used in the unit models for coagulation/flocculation and settling tanks. 

- Adds coagulation_prop_pack to our property models
- Adds unit testing for the package

### Legal Acknowledgement


watertap/property_models/coagulation_prop_pack.py
```diff
@@ -0,0 +1,748 @@
...
+
+        # phases
+        self.Liq = LiquidPhase()
+        self.Sol = SolidPhase()
+
+        # components
+        self.H2O = Component(default={"valid_phase_types": PhaseType.liquidPhase})
+        self.TSS = Component(default={"valid_phase_types": PhaseType.liquidPhase})
+        self.TDS = Component(default={"valid_phase_types": PhaseType.liquidPhase})
+        self.Sludge = Component(default={"valid_phase_types": PhaseType.solidPhase})
```
adam-a-a, line 72: Do we really need a sludge component in the solid phase? Couldn't we represent the sludge as a very concentrated liquid stream comprising H2O and TSS for the most part?

watertap/property_models/coagulation_prop_pack.py
```diff
@@ -0,0 +1,748 @@
...
+
+        # ====== parameters ========
+        mw_comp_data = {'H2O': 18.01528e-3,
+                        'TSS': 31.4038218e-3,
+                        'TDS': 31.4038218e-3,
+                        'Sludge': 31.4038218e-3}
```
adam-a-a: What are the estimates for molecular weight values of TSS, TDS, and Sludge based on?

watertap/property_models/tests/test_coag_prop_pack.py
```diff
@@ -0,0 +1,204 @@
...
+
+        # call the 'calculate_state', which will call initialize and return results
+        #       pass var_args as the fixed states from 'test_stats'
+        args = {('temperature', None): 298,
+                ('pressure', None): 101325,
+                ('flow_mass_phase_comp', ('Liq','H2O')): 1,
+                ('flow_mass_phase_comp', ('Liq','TDS')): 0.01,
+                ('flow_mass_phase_comp', ('Liq','TSS')): 0.01,
+                ('flow_mass_phase_comp', ('Sol','Sludge')): 0.001}
+        results = model.fs.stream.calculate_state(var_args=args)
```
adam-a-a, line 178: We usually just use `initialize` or `initialization_tester` for this test

watertap/property_models/tests/test_coag_prop_pack.py
```diff
@@ -0,0 +1,204 @@
...
+        assert results.solver.termination_condition == TerminationCondition.optimal
+        assert results.solver.status == SolverStatus.ok
+
+        # check to make sure DOF does not change
+        assert degrees_of_freedom(model) == 0
+
+    @pytest.mark.component
+    def test_solve(self, coag_obj):
```
adam-a-a, line 186: I think you should have another test to check the solution values.

adam-a-a (changes requested):
Things look good, but I have a couple of comments and suggested changes.

watertap/property_models/coagulation_prop_pack.py
adam-a-a: I vaguely remember us talking about this at a meeting but don't recall the consensus. From my perspective, I don't see why we _need_ sludge as a separate solid phase component, but maybe there's a good reason for this that I'm missing.

aladshaw3: Yes, I feel we need to have separate phases. This is thinking forward to the settling tank unit model where we settle out only the solids. In a general sense, the solids won't all necessarily be called 'Sludge', so it will be easier to settling out materials by just identifying all components in the solid phase as the things that get settled. 

For instance, if we want to use the settling tank unit with a Generic Property package that is calculating precipitates. There are many precipitates possible, but they will always be identified as solid phase. So we could using a generalized approach in the settling tank that would work with any solids just by looking at what components are identified as solid based on the phase it is declared in.

watertap/property_models/tests/test_coag_prop_pack.py
aladshaw3: I know, but doing this increases test coverage and accomplishes the same thing.

watertap/property_models/coagulation_prop_pack.py
aladshaw3: So, there is no such thing as a MW for TSS and Sludge, but I wanted to have something here just in case we needed molar flows (for whatever reason). In this case, I just assigned TSS and Sludge to have same MW as TDS (from the seawater prop pack). 

I can make the mw_comp Mutable so anyone can change them if needed.

adam-a-a: I wonder if it would make sense to only apply molar flows for H2O and TDS and omit for TSS and Sludge. Otherwise, if someone tries to get the molar flows for all components, they are going to see strange results for TSS and sludge that aren't really applicable.

watertap/property_models/coagulation_prop_pack.py
adam-a-a: If that's the case, maybe that component could be renamed to something more descriptive like `SettledSolids`.

aladshaw3: So, in my mind, that only makes sense if we assume all the solids are removed (i.e., for a perfect separator). In the 0D settling tank, we may have only a specific amount of solids settled, the rest may remain in the stream.

watertap/property_models/coagulation_prop_pack.py
aladshaw3: Will remove molar flows. It doesn't make sense to have in the context of coagulation and we can add it back later if needed.

watertap/property_models/tests/test_coag_prop_pack.py
aladshaw3: Will add in solution check

aladshaw3:
@lbianchi-lbl  What is this codecov/patch test and why is it giving me grief now?

The only lines not tested are those that would raise configuration or initialization errors. Literally every other line in this PR is unit tested. What gives?

lbianchi-lbl:
@aladshaw3 The `codecov/patch` check is enforcing that the test coverage for the files added or modified by a PR is greater or equal to the test coverage for the entire project. Since WaterTAP has pretty good test coverage already at around 90%, it can happen that a file with similar coverage fluctuates below that threshold, which seems to be the case here, where the difference between the observed coverage for this file is less that 1% of the target.

As with any metric, personally I wouldn't stress too much about the coverage numbers: clearly there isn't any significant difference if a file is 89% covered or 90% covered. There are ways to configure Codecov with a tolerance range to account for cases such as this, and we should discuss it tomorrow during the dev call.

More in general, 85-90% is usually where the "coverage plateau" occurs: most of the "happy paths" (i.e. code without invalid values) are tested, and the missing statements correspond to the "unhappy paths", i.e code used for validation and/or error handling. These should be tested as well, however, in practice it is more difficult since it requires creating multiple scenarios to account for all possible errors.

watertap/property_models/tests/test_coag_prop_pack.py
```diff
@@ -222,3 +222,53 @@ def test_solution(self, coag_obj):
...
+
+        # check if any variables are badly scaled
+        badly_scaled_var_list = list(iscale.badly_scaled_var_generator(model, large=1e2, small=1e-2))
+        if len(badly_scaled_var_list) != 0:
+            lst = []
+            for (var, val) in badly_scaled_var_list:
+                lst.append((var.name, val))
+                print(var.name, var.value)
+            print("The following variable(s) are poorly scaled: {lst}".format(lst=lst))
+        assert len(badly_scaled_var_list) == 0
```
lbianchi-lbl: By default pytest will automatically display values used for `assert`, so this could be simplified as e.g.:

```py
badly_scaled_var_values = {
    var.name: val
    for (var, val) in iscale.badly_scaled_var_generator(
        model, large=1e2, small=1e-2
    )
}
assert not badly_scaled_var_values
```

watertap/property_models/coagulation_prop_pack.py
```diff
@@ -0,0 +1,640 @@
...
+
+        def rule_dens_mass_phase(b, p):
+            if (p == 'Liq'):
+                return (b.dens_mass_phase[p] == (b.params.ref_dens_liq +
+                            b.params.dens_slope * b.mass_frac_phase_comp['Liq', 'TDS'] +
+                            b.params.dens_slope * b.mass_frac_phase_comp['Liq', 'TSS']) *
+                            (b.params.dens_param_A * b.temperature**2 + b.params.dens_param_B *
+                            b.temperature + b.params.dens_param_C) * (b.params.ref_pressure_correction +
+                            b.params.ref_pressure_slope * b.pressure) )
+            elif (p == 'Sol'):
```
adam-a-a, line 471: @aladshaw3  Can you add a quick comment on the source of this constraint? You may have it somewhere else already but I missed it.

adam-a-a (dismissed):
I think things look good to go. Any necessary refinement (if any) can probably be discovered and addressed once corresponding unit model dev proceeds.

watertap/property_models/coagulation_prop_pack.py
```diff
@@ -0,0 +1,640 @@
...
+
+    def _enth_flow(self):
+        # enthalpy flow expression for get_enthalpy_flow_terms method
+        temperature_ref = 273.15 * pyunits.K
+
+        def rule_enth_flow(b):  # enthalpy flow [J/s]
+            return (b.params.cp
+                    * sum(b.flow_mass_phase_comp[pair] for pair in b.flow_mass_phase_comp)
+                    * (b.temperature - temperature_ref))
+        self.enth_flow = Expression(rule=rule_enth_flow)
```
TimBartholomew: I recently learned that IDAES control volumes expect enthalpy flow to be indexed by phase. An example of this is the [water_prop_pack](https://github.com/watertap-org/watertap/blob/add_initial_mvc/watertap/property_models/water_prop_pack.py) on the add_initial_mvc branch. I don't think this should hold up the PR, but if you use a control volume for your unit model and use the add_energy_balances, you'll need to make sure whatever is returned by the `get_enthalpy_flow_terms` method below is indexed by p.

watertap/property_models/coagulation_prop_pack.py
```diff
@@ -0,0 +1,640 @@
...
+    # NOTE: For scaling in the control volume to work properly, these methods must
+    # return a pyomo Var or Expression
+
+    def get_material_flow_terms(self, p, j):
+        """Create material flow terms for control volume."""
+        return self.flow_mass_phase_comp[p, j]
+
+    def get_enthalpy_flow_terms(self, p):
+        """Create enthalpy flow terms."""
+        return self.enth_flow
```
TimBartholomew: As I mentioned in the above comment this should be indexed by phase.

watertap/property_models/coagulation_prop_pack.py
```diff
@@ -0,0 +1,640 @@
...
+                    else:
+                        sf = (iscale.get_scaling_factor(self.flow_mass_phase_comp[pair])
+                              / iscale.get_scaling_factor(self.flow_mass_phase_comp['Liq', 'H2O']))
+                        iscale.set_scaling_factor(self.mass_frac_phase_comp[pair], sf)
+                        # transforming constraints
+                        iscale.constraint_scaling_transform(self.eq_mass_frac_phase_comp[pair], sf/10)
+
+        if self.is_property_constructed('dens_mass_phase'):
+            if iscale.get_scaling_factor(self.dens_mass_phase) is None:
+                iscale.set_scaling_factor(self.dens_mass_phase, 1e-3)
```
TimBartholomew, line 602: this is already set by the default_scaling_factors you created on the parameter block above.

watertap/property_models/coagulation_prop_pack.py
```diff
@@ -0,0 +1,640 @@
...
+
+            # transforming constraints
+            sf = iscale.get_scaling_factor(self.dens_mass_phase)
+            iscale.constraint_scaling_transform(self.eq_dens_mass_phase['Liq'], sf)
+            iscale.constraint_scaling_transform(self.eq_dens_mass_phase['Sol'], sf)
```
TimBartholomew, line 607: The constraint transformation should not be in the conditional checking if the scaling factor exists. Otherwise if the scaling factor does exist, then the constraint won't be transformed. Usually, I have a section setting all the scaling factors and then a section for transforming the constraints.

watertap/property_models/tests/test_coag_prop_pack.py
```diff
@@ -0,0 +1,271 @@
...
+from idaes.core.util.scaling import badly_scaled_var_generator
+from idaes.core.util import get_solver
+
+__author__ = "Austin Ladshaw"
+
+solver = get_solver()
+
+# -----------------------------------------------------------------------------
+# Start test class
+class TestCoagulationPropPack():
```
TimBartholomew, line 45: Did you try using the WaterTAP property test harness? It has a set of standard tests that are helpful. For example, it would have caught that the density constraint was not transformed for scaling.

TimBartholomew (commented):
A couple of comments. I believe you will need to eventually update the enthalpy flow (whether you do that in this PR, or in one where you use this property package in a unit model, is up to you). The other comment is did you consciously choose to not run this through the WaterTAP property package test harness?

watertap/property_models/coagulation_prop_pack.py
andrewlee94: That is not quite true - it is specifically that `get_enthalpy_flow_terms` is provided an index when it is called, but it just returns a expression (unindexed). Thus, if you only have one phase you are free to ignore the index and just return `self.enth_flow`.

watertap/property_models/tests/test_coag_prop_pack.py
aladshaw3: I don't know how to use it and I don't really have time to mess with it at the moment. This needs to get in ASAP because it is holding up the unit model that has to be in by March (and I really don't have the time right now).

watertap/property_models/coagulation_prop_pack.py
aladshaw3: Will be fixed.

watertap/property_models/coagulation_prop_pack.py
aladshaw3: Applying fixes

watertap/property_models/coagulation_prop_pack.py
aladshaw3: done

TimBartholomew (approved):

srikanthallu (approved):

PR #328, opened 2022-02-11
Scaled bound_relax_factor

### LINKED ISSUE #325: ipopt-watertap should (nearly) honor the scaled bounds on scaled variables
This is especially needed for variables like `A_comp` in the RO unit model: https://github.com/watertap-org/watertap/blob/f522d360f250b45d514e72afd671d4dc24fa70d0/watertap/unit_models/_reverse_osmosis_base.py#L263-L270

See the related discussion in coin-or/Ipopt#555 for details.

bknueven:
## Summary/Motivation:
@adam-a-a and I discovered that sometimes Ipopt will not honor bounds on small variables, even when appropriately scaled. The issue can be summarized with the following problem:
```python
import pyomo.environ as pyo                                                                                                                                                            
import watertap #for ipopt-watertap

m = pyo.ConcreteModel()
m.x = pyo.Var(bounds=(0.5e-12, 1.5e-12), initialize=1e-12)
m.scaling_factor = pyo.Suffix(direction=pyo.Suffix.EXPORT)
m.scaling_factor[m.x] = 1e+12

m.o = pyo.Objective(expr=1e+12*m.x)
ipopt = pyo.SolverFactory("ipopt-watertap")
ipopt.solve(m)

print(f"optimal solution: {m.x.value}")
```

On main, this prints `optimal solution: -9.999499999997493e-09`, well outside the relative lower bound of 0.5e-12. With this PR, it prints `optimal solution: 4.999999925092976e-13`, which nearly satisfies the bound.

- Modify `ipopt-watertap` so that *it* does Ipopt's bound relaxation step, taking into account the variable scaling.
- Tweaking a few regression test values for the modified solver
- Setting `bound_relax_factor` manually for a few chemistry test initialization steps
- Update property test harness for values near but not equal 0.
- Reformulating a few constraints in the crystallization property package for better numerical stability
- Move a couple full_treatment_train unit tests into `tests` subdirectory

### Legal Acknowledgement


adam-a-a (approved):
LGTM- this is an important modification for ipopt that can prevent erroneous solutions for variables that are actually out of bounds.

andrewlee94 (approved):
Looks OK to me. My one concern is that the `bound_relax_factor` appears to be case specific, meaning more effort for the user to correctly set things up. However, that might be unavoidable when dealing with these small numbers.

bknueven:
@andrewlee94 one thing to note is that where set, we set that option only for the unit model initialization, and that it's only needed on the sometimes numerically troublesome chemistry tests. Because these tend to use small numbers, they probably had variables which were over-relaxed before.

It's possible that changing the scaling on the affected instances would have also resolved the difference -- I did not attempt to do so.

PR #331, opened 2022-02-14
Migrating WT3 Clarifier and Sedimentation Units

adam-a-a:
## Summary/Motivation:
This PR migrates the zero-order clarifier and sedimentation units from WT3 into WaterTAP using the new framework. 

- add clarifier_zo and tests
- add sedimentation_zo and tests

### Legal Acknowledgement


watertap/unit_models/zero_order/clarifier_zo.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+    Zero-Order model for a Clarifier unit operation.
+    """
+
+    CONFIG = ZeroOrderBaseData.CONFIG()
+
+    def build(self):
+        super().build()
+
+
```
andrewlee94: To many blank lines.

adam-a-a:
@lbianchi-lbl @ksbeattie  - The test for py3.9/win64 failed but I'm not sure why. Any thoughts?

watertap/unit_models/zero_order/tests/test_clarifier_zo.py
```diff
@@ -0,0 +1,299 @@
...
+                value(model.fs.unit.properties_treated[0].conc_mass_comp["tss"]))
+        assert (pytest.approx(90.7645, rel=1e-5) ==
+                value(model.fs.unit.properties_treated[0].conc_mass_comp["foo"]))
+        assert (pytest.approx(2.9825e-3, rel=1e-5) ==
+                value(model.fs.unit.properties_byproduct[0].flow_vol))
+        assert (pytest.approx(996.648, rel=1e-5) ==
+                value(model.fs.unit.properties_byproduct[0].conc_mass_comp["tss"]))
+        assert (pytest.approx(0, rel=1e-5) ==
+                value(model.fs.unit.properties_byproduct[0].conc_mass_comp["foo"]))
+        assert (pytest.approx(0, rel=1e-5) ==
```
andrewlee94: It appears that the electricity consumption/intensity is set to zero here. Is this something we expect to be universal, or is it just that we don't have sufficient data? If the former, then I have two questions/comments:

1. Should we even include electricity consumption in this model? It might be worth it for consistency, but I am not sure.
2. If we are going to include electricity consumption, should we have a new "zero_consumption` method that just fixes the consumption to zero to avoid having the additional constraints?

andrewlee94 (commented):
I have a couple of comments based on the current state of this model. The most significant one is on whether we expect electricity consumption/demand to always be zero for this type of unit operation.

watertap/unit_models/zero_order/tests/test_clarifier_zo.py
adam-a-a: I think that there might be energy consumption related to pumping, but the yaml file doesn't include any estimates for that. I think we could just leave this as zero for now and either (a) revise to a non-zero value later as info becomes available or (b) leave room for the user to provide their own value for electricity intensity.

lbianchi-lbl:
> @lbianchi-lbl @ksbeattie - The test for py3.9/win64 failed but I'm not sure why. Any thoughts?

@adam-a-a I've never seen this particular error before, but given the absence of logged output that's specific to us, my guess is that it might be an internal error in GitHub Actions:
image

 When an unexpected error occurs, my usual recommendation is to try if re-running the CI job(s) using the "Re-run all jobs" button (highlighted in red in the screenshot above) is enough to address the issue.

andrewlee94 (dismissed):

watertap/unit_models/zero_order/tests/test_clarifier_zo.py
```diff
@@ -0,0 +1,299 @@
...
+        assert (pytest.approx(1.001752e-2, rel=1e-5) ==
+                value(model.fs.unit.properties_treated[0].flow_vol))
+        assert (pytest.approx(2.7474, rel=1e-5) ==
+                value(model.fs.unit.properties_treated[0].conc_mass_comp["tss"]))
+        assert (pytest.approx(2.9825e-3, rel=1e-5) ==
+                value(model.fs.unit.properties_byproduct[0].flow_vol))
+        assert (pytest.approx(996.648, rel=1e-5) ==
+                value(model.fs.unit.properties_byproduct[0].conc_mass_comp["tss"]))
+        assert (pytest.approx(0, rel=1e-5) ==
+                value(model.fs.unit.electricity[0]))
```
bknueven: Something to keep in mind -- there is no "relative comparison with 0", so this test for floating point equality. Same applies below.

I would recommend using the `abs` keyword instead to your preferred tolerance.

bknueven (dismissed):
I only see a minor issue. Otherwise looks reasonable.

watertap/data/techno_economic/sedimentation.yaml
```diff
@@ -1,4 +1,13 @@
...
+  energy_electric_flow_vol_inlet:
+    value: 0.0
+    units: kWh/m^3
+  cap_basis_par:
+    value: 13572
+    units: $/ft^2
+  cap_exp_par:
+    value: 0.3182
+    units: dimensionless
   recovery_frac_mass_H2O:
```
andrewlee94, line 11: I think we should consider what to do in cases like this where recovery is set to 0.9999. This is either an indication of high but unknown recovery, or a case where recovery is actually 1.

For the latter case (if any actually exist), we should consider not having recovery as a parameter, or at least set to to 1. For the former, we should as t least ask ourselves if 0.9999 is the correct value to use.,

watertap/unit_models/zero_order/sedimentation_zo.py
```diff
@@ -0,0 +1,64 @@
...
+
+        def rule_basin_surface_area(b, t):
+            return (b.basin_surface_area[t] ==
+                    pyunits.convert(
+                        b.properties_in[t].flow_vol
+                        / b.settling_velocity[t],
+                        to_units=pyunits.ft**2))
+        self.basin_surface_area_constraint = Constraint(self.flowsheet().time, rule=rule_basin_surface_area)
+
+
```
andrewlee94: Extra lines at end of file - there should be only 1.

watertap/unit_models/zero_order/tests/test_sedimentation_zo.py
```diff
@@ -0,0 +1,322 @@
...
+                value(model.fs.unit.properties_in[0].conc_mass_comp["tss"]))
+        assert (pytest.approx(1.00265e-2, rel=1e-5) ==
+                value(model.fs.unit.properties_treated[0].flow_vol))
+        assert (pytest.approx(2.7449, rel=1e-5) ==
+                value(model.fs.unit.properties_treated[0].conc_mass_comp["tss"]))
+        assert (pytest.approx(2.9735e-3, rel=1e-5) ==
+                value(model.fs.unit.properties_byproduct[0].flow_vol))
+        assert (pytest.approx(999.66, rel=1e-5) ==
+                value(model.fs.unit.properties_byproduct[0].conc_mass_comp["tss"]))
+        assert (pytest.approx(0, rel=1e-5) ==
```
andrewlee94: As @bknueven noted, make this `abs=1e-5`.

andrewlee94 (commented):
Only a couple of minor comments.

watertap/data/techno_economic/sedimentation.yaml
adam-a-a: 0.9999 is probably not the right value to use, but I think we can refine/revisit these as data become available or are discovered.

andrewlee94 (dismissed):

bknueven (dismissed):

adam-a-a:
@andrewlee94 @bknueven I had to update test reports since a couple 0 values changed to something on the order of 2e-7 as a result of the bound relax mod. Just needs a quick reapproval (assuming all tests pass now).

bknueven (approved):

andrewlee94 (approved):

PR #332, opened 2022-02-15
Migrating WT3 Coagulation/Flocculation Unit

adam-a-a:
## Summary/Motivation:
This PR brings in the coagulation/flocculation unit from WT3, which is used in the HRSD case study and others.

- add coag/flocc unit and tests
- augment coag/flocc yaml file with relevant parameters
- add viscosity to zero order prop model
- add expr_dict to zero order base for reporting expressions

### Legal Acknowledgement


watertap/data/techno_economic/coag_and_floc.yaml
```diff
@@ -7,8 +7,47 @@ default:
...
+  num_floc_processes:
+    value: 2
+    units: dimensionless
+  num_coag_processes:
+    value: 1
+    units: dimensionless
+  num_floc_injection_processes:
+    value: 1
+    units: dimensionless
+  G_rapid_mix:
```
andrewlee94: I would suggest more descriptive names here.

watertap/unit_models/zero_order/coag_and_floc_zo.py
```diff
@@ -0,0 +1,226 @@
...
+        @self.Expression(self.flowsheet().time,
+                         doc='Total power consumption')
+        def total_power(b, t):
+            return b.power_floc[t] + b.power_rapid_mix[t]
+
+        self._perf_expr_dict["Total Power Consumption (kW)"] = self.total_power
+
+        @self.Constraint(self.flowsheet().time,
+                         doc='Electricity intensity with respect to inlet flowrate of unit')
+        def rule_energy_electric_flow_vol_inlet(b, t):
```
andrewlee94: Do we actually need this here, seeing as we are calculating consumption separately?

watertap/unit_models/zero_order/coag_and_floc_zo.py
```diff
@@ -0,0 +1,226 @@
...
+            elif j == 'polymer':
+                chemical_dosage = blk.polymer_dose[t]
+            return (blk.chemical_flow_mass[t, j] ==
+                    pyunits.convert(chemical_dosage * blk.properties[t].flow_vol,
+                                    to_units=pyunits.kg/pyunits.s))
+        self.chemical_flow_constraint = Constraint(self.flowsheet().time,
+                                                   ["alum", "polymer"],
+                                                   rule=rule_chem_flow)
+
+        def rule_anionic_polymer_dose(blk, t):
```
andrewlee94, line 178: I wonder if it would be best if these were Vars and Constraints instead of Expressions, just to keep things more consistent.

andrewlee94 (commented):
A couple of comments for discussion.

watertap/data/techno_economic/coag_and_floc.yaml
adam-a-a: I also consider "velocity_gradient_rapid_mix" but "G" is well known in the discipline when it comes to mixing energy calculations, so I opted for that. Would you prefer I change it to the former?

watertap/unit_models/zero_order/coag_and_floc_zo.py
adam-a-a: I asked myself the same question, but the WT3 method for computing energy consumption returns the "electricity intensity". So I just kept consistent with the original implementation in case it was needed for the case studies for any reason.

watertap/unit_models/zero_order/coag_and_floc_zo.py
adam-a-a: I can do that.

watertap/data/techno_economic/coag_and_floc.yaml
andrewlee94: The IDAES standard has always been to be verbose for clarity. It also helps make things more accessible to those new to the domain,

watertap/unit_models/zero_order/coag_and_floc_zo.py
andrewlee94: I see a few things here:

1. Electricity consumption is probably the more important metric and the one that will be used at the flowsheet level - intensity is inherently specific to the unit and thus not so useful at the flowsheet level (you can calculate a separate flowsheet intensity if needed, but it will not be the sum across the units).
2. In many cases, WT3 ends up doing something like `intensity = some_expression/Q` followed by `electricity = intensity*Q` - i.e. they should have just calculated the electricity demand straight up and forgotten about intensity (or at least calculated it from demand instead of the other way around).
3. Base on the above, I don't see much of a need for calculating intensity in these cases - personally I would only include it in the cases where it is a parameter used to calculate electricity demand.

adam-a-a: Fair enough- I'll remove it.

bknueven (dismissed):
Seems reasonable.

Some of the new files have some extra new lines at the end of them -- not a huge deal, but maybe they should be removed.

watertap/unit_models/zero_order/tests/test_coag_and_floc_zo.py
```diff
@@ -224,33 +218,32 @@ def test_report(self, model):
...
-                         Number of Rapid Mix Processes :     1.0000 :  True : (None, None)
-                                Number of Rapid Mixers :     1.0000 :  True : (None, None)
-                                 Polymer Dosage (mg/L) :    0.10000 :  True : (0, None)
-                                   Polymer Flow (kg/s) : 1.3000e-06 : False : (0, None)
-                          Rapid Mix Basin Volume (m^3) :   0.071500 : False : (None, None)
-                                  Rapid Mix Power (kW) :   0.057915 : False : (None, None)
-                          Rapid Mix Retention Time (s) :     5.5000 :  True : (None, None)
-                     Rapid Mix Velocity Gradient (1/s) :     900.00 :  True : (None, None)
+    Key                                : Value      : Fixed : Bounds
+                    Alum Dosage (mg/L) :     10.000 :  True : (0, None)
```
andrewlee94: As a general question, do we really want all of this information in the report? I generally feel that a report should be a summary f the key details rather than a comprehensive listing.

watertap/core/zero_order_base.py
```diff
@@ -90,6 +90,7 @@ def build(self):
         # Attributed for storing contents of reporting output
         self._stream_table_dict = {}
         self._perf_var_dict = {}
+        self._perf_expr_dict = {}
```
andrewlee94: I suggest we pull these changes out for now - lets avoid Expressions unless we really need them.

andrewlee94 (changes requested):
Only a couple of minor things, most significantly removing the support for expression from the base class.

watertap/unit_models/zero_order/tests/test_coag_and_floc_zo.py
adam-a-a: I assumed that if someone wanted to go as far as to report results directly from this unit, this would all be relevant information. I'd prefer to leave it all in, but I'll scan it again and see what I can remove.

watertap/core/zero_order_base.py
adam-a-a: OK- I can do that, but not sure why we want to avoid Expressions.

andrewlee94 (approved):
Looks good.

bknueven (approved):

PR #335, opened 2022-02-16
SISO class & UV-AOP ZO Model 

adam-a-a:
## Summary/Motivation:

- Create a new single-in/single-out zero-order class where composition changes at the inlet and outlet (e.g., UV reactor).

- Migrate UV and UV-AOP models from WT3.

- add new siso_zo class and tests
- adding uv_zo, uv_aop_zo and tests
-modify uv and uv_aop yaml files

### Legal Acknowledgement


watertap/core/zero_order_siso.py
```diff
@@ -0,0 +1,216 @@
...
+def build_siso(self):
+    """
+    Helper method for constructing material balances for zero-order type models
+    with SISO behavior.
+
+    Two StateBlocks are added with two corresponding Ports:
+        * properties_in --> inlet
+        * properties_treated ---> treated
+
+#todo: revisit this after formulation complete
```
bknueven: todone?

watertap/unit_models/zero_order/uv_aop_zo.py
```diff
@@ -0,0 +1,130 @@
...
+    """
+
+    CONFIG = ZeroOrderBaseData.CONFIG()
+
+    CONFIG.declare("has_aop", ConfigValue(
+            default=False,
+            domain=In([True, False]),
+            description="Oxidant construction flag",
+            doc="""Indicates whether oxidant dose and flow terms should be constructed or not.
+            **default** - False."""))
```
bknueven: Looking at the code below, are we sure we shouldn't be just using inheritance for this? E.g., the class `UVZO` would be all the code with `has_aop == False`, and the class `UVAOPZO` would inherit from it and build the additional Vars/Constraints in its build method.

It would also seem you wouldn't need the over-written `load_parameters_from_database` in either class, simplifying the code.

bknueven (commented):
I have a question about the `UVAOPZO` unit.  I think it might be better as two classes, with one deriving from the other.

watertap/core/zero_order_siso.py
```diff
@@ -0,0 +1,216 @@
...
+        initialize=0.01,
+        units=pyunits.dimensionless,
+        doc='Solute removal fraction on a mass basis')
+
+    # Add performance constraints
+
+    # Flow balance
+    @self.Constraint(self.flowsheet().time, doc='Overall flow balance')
+    def water_balance(b, t):
+        return (b.properties_in[t].flow_mass_comp["H2O"] ==
```
andrewlee94: I know I said otherwise at the beginning, but seeing as we are already throwing conservation of mass out the window with this for solutes, should we extent the same approach to water? I.e. include a recovery and just assume it is 1 unless we are told otherwise?

The YAML files could still leave out recovery in most cases, and only add it in for those where is is not 1.

I could see this as maybe being useful for evaporation ponds, where water leaves the system, but we do not capture the evaporated water.

watertap/unit_models/zero_order/uv_aop_zo.py
```diff
@@ -0,0 +1,130 @@
...
+                f"{self.name} - zero-order uv/aop operations "
+                "require the process_subtype configuration argument to be set")
+
+        build_siso(self)
+        constant_intensity(self)
+
+        self.uv_reduced_equivalent_dose = Var(self.flowsheet().time,
+                                              units=pyunits.mJ/pyunits.cm**2,
+                                              doc="Reduced equivalent dosage")
+        self.uvt_in = Var(self.flowsheet().time,
```
andrewlee94: I would recommend naming this `uv_transmitance_in` for clarity.

watertap/unit_models/zero_order/uv_aop_zo.py
```diff
@@ -0,0 +1,130 @@
...
+            raise NotImplementedError(
+                f"{self.name} derived zero order unit model has not "
+                f"implemented the _tech_type attribute. This is required "
+                f"to identify the database file to load parameters from.")
+
+        # Get parameter dict from database
+        pdict = self.config.database.get_unit_operation_parameters(
+            self._tech_type, subtype=self.config.process_subtype)
+
+        if not self.config.has_aop:
```
andrewlee94: I am not sure you need to overload this method. Why not just conditionally add this oxidation dose vars to `._perf_var_dict` for the AOP case?

andrewlee94 (commented):
A few comments, some building off Ben's.

watertap/unit_models/zero_order/uv_aop_zo.py
adam-a-a: I conditionally added oxidation dose in the build. The reason for this overload was only to raise an exception when the user keeps `has_aop=False` but chooses a subtype that _does_ include aop. In that case, the chemical flow and dose vars would not be created, and chemical flow wouldn't be calculated.

watertap/unit_models/zero_order/uv_aop_zo.py
adam-a-a: Will do.

watertap/unit_models/zero_order/uv_aop_zo.py
andrewlee94: Wouldn't the base method have raised an exception anyway?

watertap/unit_models/zero_order/uv_aop_zo.py
adam-a-a: Now that I think about it, this could work nicely because `UVAOPZO` would inherit the same tech type name of "uv_aop" and point to the same yaml file. I guess the main reason I did all this was to merge those two yaml files (uv_irradiation and uv_aop) into a single uv_aop, which would include both uv alone and uv/aop.

andrewlee94: You could do that, although I would be cautious.

The other thing to note is that this might be a case for supporting multiple subtypes in a single unit (but that has its own issues).

watertap/unit_models/zero_order/uv_aop_zo.py
adam-a-a: No, it doesn't. It does raise it when `has_aop` is True and the user picks a subtype without aop. This is because when attempting to fix `oxidant_dose`, no such parameter would be found in the subtype.

When `has_aop` is False and the user picks a subtype with aop, the build doesn't create `oxidant_dose` and therefore never attempts to fix an oxidant_dose parameter, and `load_parameters_from_database` doesn't expect to find `oxidant_dose`. Thus, no exception is raised.

andrewlee94: OK, that I think is the concrete argument for having two separate models and two separate YAML files.

The data for systems with AOP is only valid in the case where AOP is present (i.e. the removal fractions are assuming both the effects of UV and oxidant), thus I don't think it should be combined with those cases where it is not. The AOP model can still inherit from the UV model, but I think they should be separate classes with separate YAML files (and thus technology type). Otherwise, the line between the two gets too blurred.

Note that in other applications with subtypes, the model form is the same across all subtypes, and only the parameter values change, which is not the case here.

watertap/unit_models/zero_order/uv_aop_zo.py
adam-a-a: With that said, although using inheritance would result in cleaner code, is that something we think I should spend time on to refine when the code (seems to) works fine as is? In my mind, it might be better to make only the subtle adjustments suggested and move on to the numerous other models that we have to merge. What are your thoughts @andrewlee94 and @bknueven ?

adam-a-a: Given @andrewlee94 's response elsewhere in this PR, I'll proceed with separating the two and using inheritance. It shouldn't take long.

watertap/core/zero_order_siso.py
adam-a-a: In that case, how about including recovery in the YAML files and removing the try/except that I placed around setting recovery/removal?

andrewlee94: On that I am not sure - lets leave it as is for now.

watertap/unit_models/zero_order/uv_aop_zo.py
```diff
@@ -0,0 +1,61 @@
...
+# Technology Laboratory (subject to receipt of any required approvals from
+# the U.S. Dept. of Energy). All rights reserved.
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and license
+# information, respectively. These files are also available online at the URL
+# "https://github.com/watertap-org/watertap/"
+#
+###############################################################################
+"""
+This module contains a zero-order representation of a UV-A0P unit.
```
bknueven:
```suggestion
This module contains a zero-order representation of a UV-AOP unit.
```

watertap/data/techno_economic/uv_aop.yaml
```diff
@@ -30,5 +68,25 @@ default:
...
+    units: kWh/m^3
+
+
+
+
+
+
+
+
+
```
bknueven: It's a nit --
```suggestion
```

watertap/core/zero_order_base.py
```diff
@@ -178,7 +178,14 @@ def set_recovery_and_removal(self, data, use_default_removal=False):
...
         """
-        self.set_param_from_data(self.recovery_frac_mass_H2O, data)
+        try:
+            self.set_param_from_data(self.recovery_frac_mass_H2O, data)
+        except KeyError:
+            if self.recovery_frac_mass_H2O[:].fixed:
+                pass
+            else:
+                raise KeyError('fs.unit - database provided does not contain '
+                               'an entry for recovery_frac_mass_H2O for technology.')
```
bknueven:
```suggestion
                raise
```
No need to repeat the message here, I think

bknueven (dismissed):
I think this looks pretty good -- I just found a couple of optional things for cleanup.

bknueven (approved):

watertap/core/tests/test_zero_order_siso.py
```diff
@@ -0,0 +1,179 @@
...
+
+    @pytest.mark.component
+    def test_solve(self, model):
+        results = solver.solve(model)
+
+        # Check for optimal solution
+        assert check_optimal_termination(results)
+
+    @pytest.mark.component
+    def test_solution(self, model):
```
luohezhiming, line 139: I got a failure when running this test_solution in my laptop: 
Review_err1

bknueven: Did `test_solve` also fail?

luohezhiming: No, it works.

andrewlee94 (approved):
Looks good to me.

watertap/core/tests/test_zero_order_siso.py
andrewlee94: @luohezhiming This looks like it is something on your end - the answer definitely should be 1000 if it solved correctly. You probably need to look at what is happening during the solver call step (and maybe copy all the important code into a separate script and run it outside of `pytest` so you have full control over it.

luohezhiming (approved):

PR #340, opened 2022-02-18
Adding methods for zero order reactive separators

andrewlee94:
## Summary/Motivation:
The PR adds a new helper method for creating zero-order mass balances for systems involving both reactions and separations.

- New `zero_order_sido_reactive module
- Updates documentation

### Legal Acknowledgement

### DIFF
--- a/watertap/core/zero_order_sido_reactive.py
+++ b/watertap/core/zero_order_sido_reactive.py
@@ -1,0 +1,418 @@
+###############################################################################
+# WaterTAP Copyright (c) 2021, The Regents of the University of California,
+# through Lawrence Berkeley National Laboratory, Oak Ridge National
+# Laboratory, National Renewable Energy Laboratory, and National Energy
+# Technology Laboratory (subject to receipt of any required approvals from
+# the U.S. Dept. of Energy). All rights reserved.
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and license
+# information, respectively. These files are also available online at the URL
+# "https://github.com/watertap-org/watertap/"
+#
+###############################################################################
+"""
+This module contains methods for constructing the material balances for
+zero-order single inlet-double outlet (SIDO) unit models with chemical
+reactions.
+"""
+
+import idaes.logger as idaeslog
+from idaes.core.util import get_solver
+import idaes.core.util.scaling as iscale
+
+from pyomo.environ import NonNegativeReals, Param, Set, Var, units as pyunits
+
+# Some more inforation about this module
+__author__ = "Andrew Lee"
+
+# Set up logger
+_log = idaeslog.getLogger(__name__)
+
+
+def build_sido_reactive(self):
+    """
+    Helper method for constructing material balances for zero-order type models
+    with one inlet and two outlets including chemcial reactions.
+
+    Three StateBlocks are added with corresponding Ports:
+        * properties_inlet
+        * properties_treated
+        * properties_byproduct
+
+    Additional variables are:
+        * recovery_vol (indexed by time)
+        * removal_frac_mass_solute (indexed by time and solute)
+        * extent_of_reaction (indexed by time and reactions)
+
+    Four additional constraints are added to represent the material balances
+        * water_recovery_equation (indexed by time)
+        * flow_balance (indexed by time)
+        * solute_removal_equation (indexed by time and solute)
+        * solute_treated_equation (indexed by time and solute)
+
+    This method also sets private attributes on the unit model with references
+    to the appropriate initialization and scaling methods to use and to return
+    the inlet volumetric flow rate.
+    """
+    self._has_recovery_removal = True
+    self._initialize = initialize_sidor
+    self._scaling = calculate_scaling_factors_sidor
+
+    # Create state blocks for inlet and outlets
+    tmp_dict = dict(**self.config.property_package_args)
+    tmp_dict["has_phase_equilibrium"] = False
+    tmp_dict["defined_state"] = True
+
+    self.properties_in = self.config.property_package.build_state_block(
+        self.flowsheet().time,
+        doc="Material properties at inlet",
+        default=tmp_dict)
+
+    tmp_dict_2 = dict(**tmp_dict)
+    tmp_dict_2["defined_state"] = False
+
+    self.properties_treated = \
+        self.config.property_package.build_state_block(
+            self.flowsheet().time,
+            doc="Material properties of treated water",
+            default=tmp_dict_2)
+    self.properties_byproduct = \
+        self.config.property_package.build_state_block(
+            self.flowsheet().time,
+            doc="Material properties of byproduct stream",
+            default=tmp_dict_2)
+
+    # Create Ports
+    self.add_port("inlet", self.properties_in, doc="Inlet port")
+    self.add_port("treated",
+                  self.properties_treated,
+                  doc="Treated water outlet port")
+    self.add_port("byproduct",
+                  self.properties_byproduct,
+                  doc="Byproduct outlet port")
+
+    # Load reaction identifiers from database
+    dbparams = self.config.database.get_unit_operation_parameters(
+        self._tech_type,
+        subtype=self.config.process_subtype)
+    try:
+        rxn_ids = list(dbparams["reactions"].keys())
+    except KeyError:
+        raise KeyError(
+            f"{self.name} - database provided does not contain a list of "
+            f"reactions for this technology.")
+    # Create indexing set for reactions
+    self.reaction_set = Set(initialize=rxn_ids)
+
+    # Add performance variables
+    self.recovery_frac_mass_H2O = Var(
+        self.flowsheet().time,
+        initialize=0.8,
+        domain=NonNegativeReals,
+        units=pyunits.dimensionless,
+        bounds=(1E-8, 1.0000001),
+        doc='Mass recovery fraction of water in the treated stream')
+    self.removal_frac_mass_solute = Var(
+        self.flowsheet().time,
+        self.config.property_package.solute_set,
+        domain=NonNegativeReals,
+        initialize=0.01,
+        units=pyunits.dimensionless,
+        doc='Solute removal fraction on a mass basis')
+
+    self.reaction_conversion = Var(
+        self.flowsheet().time,
+        self.reaction_set,
+        initialize=0.5,
+        units=pyunits.dimensionless,
+        doc='Reaction conversion based on key reactant')
+
+    # Load reaction conversions
+    for (t, r), v in self.reaction_conversion.items():
+        try:
+            conv = dbparams["reactions"][r]["conversion"]
+        except KeyError:
+            raise KeyError(
+                f"{self.name} - database provided does not "
+                f"contain an entry for conversion for reaction {r}.")
+        v.fix(conv)
+
+    # Intermediate variables and parameters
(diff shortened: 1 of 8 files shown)


watertap/core/zero_order_sido_reactive.py
```diff
@@ -0,0 +1,416 @@
...
+        units=pyunits.dimensionless,
+        bounds=(1E-8, 1.0000001),
+        doc='Mass recovery fraction of water in the treated stream')
+    self.removal_frac_mass_solute = Var(
+        self.flowsheet().time,
+        self.config.property_package.solute_set,
+        domain=NonNegativeReals,
+        initialize=0.01,
+        units=pyunits.dimensionless,
+        doc='Solute removal fraction on a mass basis')
```
TimBartholomew, line 121: I think it would be best to have `removal_frac_mass_comp` rather than solute. This would include water. Then we can just have a constraint that enforces `removal_frac_mass_comp[t, 'H2O'] = 1 - recovery_frac_mass_H2O[t]` and do all of components mass balances with `removal_frac_mass_comp`.

This would be more general and I think better aligns with AMO in that several of the components wouldn't really be considered solutes.

I also don't see the harm in doing this for the WaterTAP3 type models. What value do we get out of the components being solutes?

watertap/core/zero_order_sido_reactive.py
```diff
@@ -0,0 +1,416 @@
...
+                f"{self.name} - database provided does not "
+                f"contain an entry for key_reactant for reaction {r}.")
+        if key_reactant not in b.properties_in[0].component_list:
+            raise ValueError(
+                f"{self.name} - key_reactant {key_reactant} for reaction {r} "
+                f"is not in the component list used by the assigned property "
+                f"package.")
+        return (b.extent_of_reaction[t, r] ==
+                b.reaction_conversion[t, r] *
+                b.properties_in[t].flow_mass_comp[key_reactant])
```
TimBartholomew, line 167: Isn't this not technically the extent_of_reaction? This is the amount of the key reagent reacted. I like how the model is formulated, but I think we should rename the variable to not cause confusion.

watertap/core/zero_order_sido_reactive.py
```diff
@@ -0,0 +1,416 @@
...
+
+    self.generation_ratio = Param(
+        self.reaction_set,
+        self.config.property_package.component_list,
+        initialize=0,
+        mutable=True,
+        units=pyunits.dimensionless,
+        doc="Mass ratio for generation of species w.r.t. key species")
```
TimBartholomew, line 175: Is there a reason you chose `generation_ratio` rather than `conversion_ratio`? The temporary variable you use to calculate this is `cratio` and the comments refer to it as the conversion ratio.

TimBartholomew: Why did you decide to go with a param here? Isn't the IDAES convention to just use fix vars?

watertap/core/zero_order_sido_reactive.py
```diff
@@ -0,0 +1,416 @@
...
+                b.properties_byproduct[t].flow_mass_comp["H2O"])
+
+    # Solute removal
+    @self.Constraint(self.flowsheet().time,
+                     self.config.property_package.solute_set,
+                     doc='Solute removal equations')
+    def solute_removal_equation(b, t, j):
+        return (b.removal_frac_mass_solute[t, j] *
+                (b.properties_in[t].flow_mass_comp[j] +
+                 sum(b.generation_ratio[r, j]*b.extent_of_reaction[t, r]
```
TimBartholomew: I think it would be helpful to the users if we actually made a generation term (i.e. indexed variable and constraint) rather than just using the product of the generation_ratio and extent of reaction. I could see users frequently wanting to just check how much of a component was generated.

watertap/core/zero_order_sido_reactive.py
```diff
@@ -0,0 +1,416 @@
...
+                            f"{self.name} - database provided does not "
+                            f"contain an entry for molecular_weight w.r.t. "
+                            f"species {j} in reaction {r}.")
+                    try:
+                        MW_k = cratio = stoich[key_reactant]["molecular_weight"]
+                    except KeyError:
+                        raise KeyError(
+                            f"{self.name} - database provided does not "
+                            f"contain an entry for molecular_weight w.r.t. "
+                            f"species {key_reactant} in reaction {r}.")
```
TimBartholomew: remember to remove these repeated `cratio =`

watertap/core/zero_order_sido_reactive.py
TimBartholomew: In another comment, I suggest creating a variable for the generation term, we could just use that with the key reagent index instead of this variable.

TimBartholomew (commented):
A few comments, could be faster to just have a call about them.

watertap/core/zero_order_sido_reactive.py
andrewlee94: 1. Extent of reaction is always relative to something, and normally it is expressed relative to the key component so I would say that yes this is the extent of reaction.
2. Rather than a Var for generation I would suggest an Expression if you think it is important. There is no need to introduce additional Vars and Constraints into the problem for this, as that is just making the problem bigger than it needs to be.

watertap/core/zero_order_sido_reactive.py
andrewlee94: 1. I think the two terms are essentially synonymous, and that is just what I picked.
2. No, the IDAES convention is to use Vars for anything that might conceivably be solved for, however this is something that definitely should not be left free. Hence the use of a Param.

watertap/core/zero_order_sido_reactive.py
andrewlee94: I strongly suggest not including water in here for two reasons.

1. It adds the possibility of over-specifying the problem as a user could specify both the removal and recovery of water (or worse yet, specify both of those and leave another removal unfixed and wonder why their problem with 0 degrees of freedom does not solve). Believe me, someone will try to do this. On top of this, think of the additional logic that would need to be implemented to work out which of these to fix when loading parameters from a datafile. What do we do if a user provides values for both recovery and removal?
4. It adds an additional Var and Constraint to the problem that are not really required. We want to keep these models as simple as possible, so I don't see much value in adding an additional Var and Constraint for this.

On the comment about naming, that is primarily semantics. At the moment, "solute" just means anything that is not water, and yes, not everything in there is strictly a solute (it also includes suspended solids, other solvents and potentially entrained gases). However, going with the full component list is not the solution to this. At most, we should use a more inclusive name (but I have no idea what that is). Further, if we do that then we need to go change everything else in the zero-order models too.

TimBartholomew: For comment 1, this is the case in all models. Users could fix variables that they shouldn't, and the model isn't solvable but still have zero degrees of freedom. The most similar model to these is the IDAES separator model, there users have the option to fix split fractions for either outlet stream. If they fix incompatible split fraction for one component in both outlet streams and omit a split fraction for another component, the model has zero degrees of freedom and is not solvable.

For comment 2, I understand the reasoning for not wanting another variable and constraint, especially for introducing a bilinear term into the mass balance for water. An alternative would be to fix the removal fraction of water given the recovery (rather than a constraint), i.e. we just make model only consider removal fractions and drop the recovery (in this case everyone knows they need to supply the removal fraction for water, which is 1 - recovery).

It just seems like an odd choice to make all the components solutes when they aren't and that the variables names explicitly include solute in them. It also makes the water balance different than all the other components when it is not really necessary.

For example, one AMO project has a reactor that creates a product gas with negligible water content. Then that product gas is sent to a membrane unit which should be modeled as zero order. If I use our zero order models, it's weird the removal fraction has solute in the name. It's also weird that there is a separate water balance and a requirement to provide its recovery, it would make more sense if the removal fraction of water was just automatically populated with the default value since it doesn't matter.

In summary, I feel like differentiating water from all other components is adding an additional layer of complexity, when really it can be treated the same. Then making all other components a solute is an unnecessary restriction that just is odd in several cases.

TimBartholomew (approved):

watertap/core/zero_order_sido_reactive.py
adam-a-a: I was a little confused by this as well. Following the definition of extent of reaction, represented by delta_n/stoich #, and working out how you formulated things, I think the extent of reaction should be  
`extent_of_reaction == reaction_conversion * flow_mass_comp[key_reactant] / MW[key_reactant]` which should work out to be equivalent to moles/s generated divided by stoichiometric number.

andrewlee94: The problem is that we will not always have a MW for the components (we explicitly allow for empirical mass conversion ratios), hence we can't do this on a molar basis. Thus, this is a mass-based extent of reaction instead.

watertap/core/zero_order_sido_reactive.py
```diff
@@ -0,0 +1,424 @@
...
+                b.properties_byproduct[t].flow_mass_comp["H2O"])
+
+    # Solute removal
+    @self.Constraint(self.flowsheet().time,
+                     self.config.property_package.solute_set,
+                     doc='Solute removal equations')
+    def solute_removal_equation(b, t, j):
+        return (b.removal_frac_mass_solute[t, j] *
+                (b.properties_in[t].flow_mass_comp[j] +
+                 sum(b.generation_rxn_comp[t, r, j]
```
adam-a-a, line 264: While this formulation works, I am thinking about cases where this formulation may not be applicable (or easily applicable) such as in biological processes where Monod kinetics would be used to model microbial growth. Thus, I am wondering if there should be a base class that includes the `generation_rxn_comp` term, and then this formulation would inherit the specifics that arrive at defining that generation term. Subsequently, as different representations for the generation term might be useful or desired, we could inherit said base class and add the necessary specifics to define that generation term. I hope that makes sense.

adam-a-a: I just backtracked and recalled that this is not a class, so I guess that suggestion isn't directly applicable. Still something to think about -- we _may_ need alternative representations for generation/decay.

andrewlee94: That is in part why we didn't insist on stoichiometric reactions, and left things fairly empirical.

watertap/core/zero_order_sido_reactive.py
adam-a-a: Just my two cents- while I agree with @TimBartholomew that we don't _need_ to differentiate water from all other components, I think that we don't have time to spare to step back and revise everything that is already in place for WaterTAP3-related models because it would set us back. Instead, I think our time would be better spent by pushing forward and get all models implemented first. So I have to agree with @andrewlee94 on this one. With regard to not all "solutes" necessarily being "solutes", I suppose a more appropriate name would be something like "constituent", but again, since time is limited and altering the name across the board wouldn't change code functionality, I think we can live with what is in place now. Once all of our models are in place and we shift our focus towards refinement, we could go back and refine as needed.

watertap/core/zero_order_sido_reactive.py
andrewlee94: Also, kinetics don't belong in zero-order models.

adam-a-a (approved):
I think this is good to go. I had a couple of comments to consider but nothing that should hold up the PR.

PR #342, opened 2022-02-23
Some suggested edits for #301

bknueven:
## Summary/Motivation:
Keeping the termination condition from the solver adds minimal information and adds complexity. Note that we also need to support the case when the call to `solve` completely fails: under certain circumstances Ipopt may return a non-zero exit code, which will raise an Exception in Pyomo. Even that aside, the termination condition contains very little information for debugging *why* a solve failed.

- Keep `solve_successful` bool in place of termination condition
- Lots of minor edits for internal consistency and consistency with Pyomo

### Legal Acknowledgement


k1nshuk (approved):
Looks good to me.

PR #344, opened 2022-02-23
WT3 Unit Migration: Ion Exchange

adam-a-a:
## Summary/Motivation:
Migrating WT3 anion exchange zero-order model.

- add unit and tests
- modify yaml and merge anion & cation exchange yaml files into ion exchange

### Legal Acknowledgement


watertap/unit_models/pressure_exchanger.py
```diff
@@ -1,3 +1,5 @@
+
```
andrewlee94: Extra line spaces

watertap/unit_models/zero_order/__init__.py
```diff
@@ -20,3 +20,5 @@
 from .coag_and_floc_zo import CoagulationFlocculationZO
 from .uv_zo import UVZO
 from .uv_aop_zo import UVAOPZO
+from .anion_exchange_zo import AnionExchangeZO
```
andrewlee94: Should this just be `ion_exchange`, with different subtypes for anion and cation exchange (and different resins in future)?

watertap/unit_models/zero_order/anion_exchange_zo.py
```diff
@@ -0,0 +1,44 @@
...
+    def build(self):
+        super().build()
+
+        self._tech_type = "anion_exchange"
+
+        build_siso(self)
+        self._Q = Reference(self.properties_in[:].flow_vol)
+        pump_electricity(self, self._Q)
+        self.eta_pump = 0.8  # mutable parameter; default value found in WT3 for anion exchange
+        self.lift_height = 69.91052  # mutable parameter; default value of 2 bar converted to feet head
```
andrewlee94: You should do this using `self.lift_height.set_value(2*pyo.units.bar)`

watertap/unit_models/zero_order/tests/test_anion_exchange_zo.py
```diff
@@ -0,0 +1,247 @@
...
+
+        m.fs.unit.inlet.flow_mass_comp[0, "H2O"].fix(10000)
+        m.fs.unit.inlet.flow_mass_comp[0, "tds"].fix(1)
+
+        return m
+
+    @pytest.mark.unit
+    def test_build(self, model):
+        assert model.fs.unit.config.database == model.db
+
```
andrewlee94: You should also check for the material balances in these cases.

andrewlee94 (changes requested):
All minor things , but there are few things you should change or add.

watertap/unit_models/zero_order/__init__.py
adam-a-a: I considered that but decided to take the route of separating them for now. I could look into cation_exchange and make sure it lines up with the anion_exhange unit in WT3.

watertap/unit_models/zero_order/anion_exchange_zo.py
adam-a-a: And doing this would automatically convert bar to feet?

adam-a-a: I just realized that wouldn't, so I'll set_value to that converted value (in feet).

andrewlee94: Ugh, no, it wouldn't convert to feet unfortunately but it would still be good practice to do `self.lift_height.set_value(69.91052*pyo.units.feet)`

watertap/unit_models/zero_order/tests/test_anion_exchange_zo.py
```diff
@@ -0,0 +1,247 @@
...
+            "database": m.db})
+
+        m.fs.unit.inlet.flow_mass_comp[0, "H2O"].fix(10000)
+        m.fs.unit.inlet.flow_mass_comp[0, "tds"].fix(1)
+
+        return m
+
+    @pytest.mark.unit
+    def test_build(self, model):
+        assert model.fs.unit.config.database == model.db
```
bknueven: We should probably be using `is` for this check. I see it as equals elsewhere, but I think we mean that `model.fs.unit.config.data` is the object `model.db`, and not equivalent to `model.db`.

Since nobody has defined `__eq__` on `Database` this is the same, for now.

bknueven (commented):
I have an optional comment. I think this looks good.

bknueven (approved):

andrewlee94 (approved):
Looks good.

PR #345, opened 2022-02-23
Migrating WT3 dual media filtration model

luohezhiming:
## Summary/Motivation:

add dual media filtration model

- add unit model for dual media filtration
- add corresponding test for dual media filtration unit

### Legal Acknowledgement


watertap/unit_models/zero_order/tests/test_dual_media_filtration_zo.py
```diff
@@ -0,0 +1,176 @@
...
+
+        assert isinstance(model.fs.unit.electricity, Var)
+        assert isinstance(model.fs.unit.energy_electric_flow_vol_inlet, Var)
+        assert isinstance(model.fs.unit.electricity_consumption, Constraint)
+
+    @pytest.mark.component
+    def test_load_parameters(self, model):
+        data = model.db.get_unit_operation_parameters("dual_media_filtration")
+
+        model.fs.unit.load_parameters_from_database()
```
adam-a-a, line 67: @luohezhiming - you might want to add another test when using `load_parameters_from_database(use_default_removal=True)` to ensure that a default removal rate is fixed for solutes that are not listed in the yaml file.

adam-a-a (commented):
Everything looks good. I just think you might want to add the extra test I suggested to check when `use_default_removal` is enabled.

watertap/unit_models/zero_order/tests/test_dual_media_filtration_zo.py
luohezhiming: Should I add `load_parameters_from_database(use_default_removal=True)` and check all parameters again? Or just set `use_default_removal=True` to this line 67?

adam-a-a: I think you could do something similar to the testing done in `test_nanofiltration_zo.py` and other unit model test files where you add a new set of tests for a solute_list that includes a component that doesn't exist in the yaml file.

luohezhiming: OK. I figure out what I missed, I should add a test class with `use_defalt_removal=True` . Will do.

luohezhiming: I've added the test for the case with extra solute.

andrewlee94 (commented):
Well done on your first model @luohezhiming.

II had a quick look at the YAML file for this model in addition to the code, and I had one observation that I thought I should bring up (which is an issue with the original data, not something you did). The YAML file sets the default recovery to 0.99 - I just wanted to ask if this is the correct value for this type of unit. @kurbansitterley can you comment on this value?

kurbansitterley:
Water recovery for media filtration would depend on the backwash frequency and that would depend on water quality. So hard to say what "correct" is. That being said, I am sure this value is used due to lack of data.

andrewlee94:
@luohezhiming You also have your first Git conflict to fix - this one will be easy though. If you want some help understanding what this means and how to fix it, just ask (I am busy today but should have time tomorrow).

luohezhiming:
@andrewlee94 I will try to handle the Git merge conflicts following the online tutorial. If I need any help, I will let you know! Thank you!

adam-a-a:
I think this can be merged once the conflict is resolved. However, just want to note that given our previous discussion, we may end up wanting to change this to a SIDO model instead of SISO. Since much of this is on-the-go, I think we can always revisit that and make the subtle modification once we have more confidence in it.

adam-a-a:
> I think this can be merged once the conflict is resolved. However, just want to note that given our previous discussion, we may end up wanting to change this to a SIDO model instead of SISO. Since much of this is on-the-go, I think we can always revisit that and make the subtle modification once we have more confidence in it.

Disregard that! It's actually a SIDO already.

adam-a-a (dismissed):
Great job!

adam-a-a (dismissed):

adam-a-a (approved):

bknueven (approved):

PR #346, opened 2022-02-23
WT3 Fixed Bed Unit

adam-a-a:
## Summary/Motivation:
Adding fixed_bed_zo unit with gravity_basin and pressure_vessel subtypes.

- add fixed_bed_zo unit with tests
- merge fixed_bed_gravity_basin and fixed_bed_pressure_vessel yaml files into one fixed_bed yaml

### Legal Acknowledgement


andrewlee94 (dismissed):
Looks good - I do wonder about the two sub-types as they both have the same values, but that is probably lack of data.

andrewlee94:
@adam-a-a It looks like you have a couple of conflicts to fix.

adam-a-a:
> Looks good - I do wonder about the two sub-types as they both have the same values, but that is probably lack of data.

Yes, I added those because no removal data were included. So I added removal rates for BOD with fairly arbitrary values. We can go back and refine when we know more.

bknueven (dismissed):

bknueven (approved):

andrewlee94 (approved):

PR #354, opened 2022-02-25
Migrating WT3 buffer tank unit

luohezhiming:
## Summary/Motivation:
Add buffer tank model

- add unit model for buffer tank
- add corresponding test for buffer tank model

### Legal Acknowledgement


watertap/data/techno_economic/buffer_tank.yaml
```diff
@@ -28,3 +28,8 @@ default:
   default_removal_frac_mass_solute:
     value: 0
     units: dimensionless
+  removal_frac_mass_solute:
```
adam-a-a: this model is just a tank without any removal, so you should remove this

watertap/unit_models/zero_order/buffer_tank_zo.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+
+    CONFIG = ZeroOrderBaseData.CONFIG()
+
+    def build(self):
+        super().build()
+
+        self._tech_type = "buffer_tank"
+
+        build_sido(self)
```
adam-a-a: I think this should be a pass-through model type, not SIDO.

adam-a-a (changes requested):
This should be a pass-through unit instead of a SIDO. It doesn't remove or recover anything. Later, when we add costing, I suppose more parameters would be included to estimate costs of the tank, but that might be about it.

watertap/unit_models/zero_order/buffer_tank_zo.py
luohezhiming: OK. Will modify this.

watertap/data/techno_economic/buffer_tank.yaml
luohezhiming: Removed.

bknueven (dismissed):

adam-a-a (dismissed):

bknueven:
@luohezhiming this has some conflicts. If you like, I can try to resolve them.

luohezhiming:
@bknueven I think I resolved those merge conflicts right now, let's see if all checks will pass.

bknueven (dismissed):

adam-a-a (dismissed):

bknueven:
Closing because `git` is confused by our similar-looking models. Replaced with #377.

PR #356, opened 2022-02-25
WT3 treated storage

kurbansitterley:
## Fixes/Addresses:

- Fix CT value units in ``ozonation`` and ``ozone_aop`` modules (from mg/(L *min) to (mg *min)/L) and make necessary changes in .yaml and test files. I confirmed these changes with the reference.
- Rename testing class for OzoneAOP from to include "AOP" in the name

## Summary/Motivation:

- Combine WT3 units ``treated_storage`` and ``holding_tank`` into``storage_tank``. Costing data is the same. The only difference between the two is holding_tank has a surge capacity variable that is included when calculating the storage tank volume. If that variable = 0, the storage volume is the same. 
- Units for CT value were wrong in WT3 unit. 

- create ``storage_tank`` model and associated .yaml and test files
- correct CT units in ``ozonation`` and ``ozone_aop`` and associated .yaml and test files

### Legal Acknowledgement


bknueven (approved):
Would ideally be two separate PRs, but that shouldn't hold this up.

adam-a-a (approved):
LGTM

kurbansitterley:
> Would ideally be two separate PRs, but that shouldn't hold this up.

You mean regarding the fixing of units in ozone? I will keep that in mind for next time.

PR #357, opened 2022-02-25
General 0-Order VFA Recovery Unit

adam-a-a:
## Summary/Motivation:
Adding a zero-order volatile fatty acids (VFA) recovery unit for the AMO effort.

- add VFA recovery unit and test
- add yaml
-  add new component to list for ffCOD, representative of VFAs

### Legal Acknowledgement


watertap/data/techno_economic/component_list.yaml
```diff
@@ -64,6 +64,8 @@ nitrite:
   constituent_longform: Nitrite
 nitrogen:
   constituent_longform: Nitrogen
+nonbiodegradable_cod:
+  constituent_longform: Nonbiodegradable Chemical Oxygen Demand (ffCOD)
```
bknueven, line 68: Should this (and the `nonvolatile_toc` below) have names without acronyms? E.g.
```suggestion
nonbiodegradable_chemical_oxygen_demand:
  constituent_longform: Nonbiodegradable Chemical Oxygen Demand (ffCOD)
```

bknueven (approved):
I have a naming question. But this is okay as-is.

watertap/data/techno_economic/component_list.yaml
andrewlee94: Generally we have used shorter names as the key, and expanded on them in the `constituent_longform`.

For my part, what does the `ff` in `ffCOD` mean (i.e. just making sure it isn't a typo or issue with importing the csv files)?

EDIT: I see a more detailed longform in a later file - we should probably make these the same.

andrewlee94 (approved):
Looks good to me.

watertap/data/techno_economic/component_list.yaml
adam-a-a: ffCOD stands for flocculated and filtered chemical oxygen demand. In the case of this particular unit, ffCOD is representative of volatile fatty acids (VFAs), hence the more elaborate longform presented in the unit's yaml file. However, the general longform that isn't unit specific should not describe this component as "representative of volatile fatty acids", which is why I wrote it this way here.

andrewlee94: As an observation then, we want to be really careful with things like this. What happens if we have two unit operations in a process which use different definitions for the same component (i.e. one unit uses the more elaborate definition, and one uses a more general definition)?

PR #360, opened 2022-02-27
WT3 bioreactor

TimBartholomew:
## Summary/Motivation:
Adds zero order bioreactor, which has selenium and boron removal

- Added unit and tests

### Legal Acknowledgement

### DIFF
--- a/watertap/unit_models/zero_order/__init__.py
+++ b/watertap/unit_models/zero_order/__init__.py
@@ -13,6 +13,7 @@
 
 from .aeration_basin_zo import AerationBasinZO
 from .air_flotation_zo import AirFlotationZO
+from .bioreactor_zo import BioreactorZO
 from .chemical_addition_zo import ChemicalAdditionZO
 from .feed_zo import FeedZO
 from .nanofiltration_zo import NanofiltrationZO
--- a/watertap/unit_models/zero_order/bioreactor_zo.py
+++ b/watertap/unit_models/zero_order/bioreactor_zo.py
@@ -1,0 +1,37 @@
+###############################################################################
+# WaterTAP Copyright (c) 2021, The Regents of the University of California,
+# through Lawrence Berkeley National Laboratory, Oak Ridge National
+# Laboratory, National Renewable Energy Laboratory, and National Energy
+# Technology Laboratory (subject to receipt of any required approvals from
+# the U.S. Dept. of Energy). All rights reserved.
+#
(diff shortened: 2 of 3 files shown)


TimBartholomew:
While creating a basic unit is really fast, creating the test takes some time, I would think that a test harness would be pretty valuable.

Also note that I combined the default removal fraction test with the original, rather than having two separate tests. I'm not sure what the value is of not including a component that needs the default value originally.

adam-a-a:
> Also note that I combined the default removal fraction test with the original, rather than having two separate tests. I'm not sure what the value is of not including a component that needs the default value originally.

I think you're right about this. Originally, we based our tests on @andrewlee94 's first unit (nanofiltration) where he was testing functionality between inclusion/exclusion of default removal in early stages. However, we could probably save effort by only adding the test _with_ default removal. 

With that said, we might want to add the other test for checking subtypes although this unit doesn't have any yet. When/if we do add subtypes, that test would cover new additions. You can choose to add that in now or save it for when subtypes are added later.

adam-a-a:
Also, we might think about modeling this as a sido_r model for the AMO effort, but siso works for the WT3 implementation.

bknueven:
> While creating a basic unit is really fast, creating the test takes some time, I would think that a test harness would be pretty valuable.


This seems likely, but I would recommend any such code just reflect the current structure of these models. E.g., maybe there's a test function you call for the `build_siso` function or the `constant_intensity` function. Similarly for the basic database functionality.

So in my mind this looks quite different from the test harnesses we've used in the past and reflects the structure of the basic models. Non-basic models would have additional test code to  reflect the additional variables and constraints.

adam-a-a:
> > While creating a basic unit is really fast, creating the test takes some time, I would think that a test harness would be pretty valuable.
> 
> This seems likely, but I would recommend any such code just reflect the current structure of these models. E.g., maybe there's a test function you call for the `build_siso` function or the `constant_intensity` function. Similarly for the basic database functionality.
> 
> So in my mind this looks quite different from the test harnesses we've used in the past and reflects the structure of the basic models. Non-basic models would have additional test code to reflect the additional variables and constraints.

To add to that, I think the tests are easy enough to copy/paste/revise to fast-track model migration. A test harness approach would be a nice-to-have, but I think our time would be better spent on directly migrating models and pasting/revising tests.

adam-a-a (approved):
LGTM - we might modify this later or make a new bioreactor unit for AMO needs, but this is sufficient for WT3 migration.

bknueven (approved):

PR #368, opened 2022-03-01
Fix multi-sweep analysis within full_treatment_train

### LINKED ISSUE #361: Parameter sweep error for full treatment train
When running [multi_sweep.py](https://github.com/watertap-org/watertap/blob/main/watertap/examples/flowsheets/full_treatment_train/analysis/multi_sweep.py) for the full treatment train analysis, I receive the same error for every case:
```
  File "watertap\examples\flowsheets\full_treatment_train\analysis\multi_sweep.py", line 244, in <module>
    global_results, sweep_params = run_analysis(case_num, nx, RO_type)
  File "watertap\examples\flowsheets\full_treatment_train\analysis\multi_sweep.py", line 208, in run_analysis
    global_results = parameter_sweep(m, sweep_params, outputs, output_filename,
  File "c:\users\timvb\idaes\watertap\watertap\tools\parameter_sweep.py", line 810, in parameter_sweep
    global_save_data = _save_results(sweep_params, outputs, local_values, global_values, local_results, global_results, global_output_dict,
  File "c:\users\timvb\idaes\watertap\watertap\tools\parameter_sweep.py", line 665, in _save_results
    global_results_clean = _interp_nan_values(global_values, global_results)
  File "c:\users\timvb\idaes\watertap\watertap\tools\parameter_sweep.py", line 327, in _interp_nan_values
    yi = griddata(x0, y0, global_values, method='linear', rescale=True).reshape(-1)
  File "C:\Users\timvb\anaconda3\envs\watertap-dev\lib\site-packages\scipy\interpolate\ndgriddata.py", line 253, in griddata
    ip = interp1d(points, values, kind=method, axis=0, bounds_error=False,
  File "C:\Users\timvb\anaconda3\envs\watertap-dev\lib\site-packages\scipy\interpolate\interpolate.py", line 474, in __init__
    self._y = self._reshape_yi(self.y)
  File "C:\Users\timvb\anaconda3\envs\watertap-dev\lib\site-packages\scipy\interpolate\polyint.py", line 110, in _reshape_yi
    return yi.reshape((yi.shape[0], -1))
ValueError: cannot reshape array of size 0 into shape (0,newaxis)
```

@k1nshuk @eyoung55 can you check if this is due to the changes in the parameter sweep tool? I'm running this on windows, python 3.8, and without mpi (so I get the warning: `Could not import mpi4py from current environment (defaulting to serial)`).

k1nshuk:
## Fixes/Addresses:

This PR brings the flowsheets within `flowsheets/full_treatment_train/analysis` up to date with the recent changes in the parameter sweep tool.

## Summary/Motivation:

https://github.com/watertap-org/watertap/pull/301#issue-1108408629 brought changes to `parameter_sweep` where the optimize function for parameter sweep is now expected to return the Pyomo results object for proper functioning.

- Change the `optimize`, `simulate`, and `solve_block` functions to return the Pyomo results object.
- Add a keyword argument that (defaults to True) for optional checking optimal termination condition.

### Legal Acknowledgement


bknueven (changes requested):
Please add a test so the code affected in #361, [mutli_sweep.py](https://github.com/watertap-org/watertap/blob/main/watertap/examples/flowsheets/full_treatment_train/analysis/multi_sweep.py), is covered. We don't want to be fixing something again after-the-fact.

The test doesn't have to be fancy, executing and making sure the code doesn't error out would be an improvement over the status quo.

watertap/examples/flowsheets/full_treatment_train/analysis/tests/test_multi_sweep.py
```diff
@@ -0,0 +1,44 @@
...
+                                                               flowsheet_NF_no_bypass,
+                                                               flowsheet_single_stage,
+                                                               flowsheet_two_stage,
+                                                               flowsheet_NF_two_stage,
+                                                               flowsheet_softening,
+                                                               flowsheet_softening_two_stage)
+from watertap.examples.flowsheets.full_treatment_train.analysis.multi_sweep import *
+from watertap.tools.parameter_sweep import _init_mpi, LinearSample, parameter_sweep
+
+@pytest.mark.component
```
bknueven: We should be using[`pytest.mark.parametrize`](https://docs.pytest.org/en/6.2.x/parametrize.html) here instead of the `for` loop below.

watertap/examples/flowsheets/full_treatment_train/analysis/tests/test_multi_sweep.py
```diff
@@ -0,0 +1,50 @@
...
+    try:
+        # raise ValueError()
+        global_results, sweep_params = run_analysis(case_num, nx, RO_type, interp_nan_outputs=False)
+
+    except:
+        fail_counter += 1
+        failed_cases.append((case_num, RO_type))
+
+    if fail_counter > 0:
+        pytest.fail(f"multi_sweep.py failed for cases {failed_cases} ")
```
bknueven: I think `parametrize` takes care of everything -- if an exception is raised in `run_analysis` the test will fail.
```suggestion
    nx = 1
    global_results, sweep_params = run_analysis(case_num, nx, RO_type, interp_nan_outputs=False)
```

watertap/examples/flowsheets/full_treatment_train/analysis/tests/test_multi_sweep.py
```diff
@@ -0,0 +1,50 @@
...
+from watertap.examples.flowsheets.full_treatment_train.analysis.multi_sweep import *
+from watertap.tools.parameter_sweep import _init_mpi, LinearSample, parameter_sweep
+
+pytest_parameterize_list = []
+for case_num in [1, 2, 3, 4, 8, 9]:
+    for ro_type in ['0D', '1D']:
+        test_case = (case_num, ro_type)
+        pytest_parameterize_list.append(test_case)
+
+@pytest.mark.parametrize('case_num, RO_type', pytest_parameterize_list)
```
bknueven, line 23: This seems to best fit the theme of an integration test.
```suggestion
@pytest.mark.parametrize('case_num, RO_type', pytest_parameterize_list)
@pytest.mark.integration
```

watertap/examples/flowsheets/full_treatment_train/analysis/tests/test_multi_sweep.py
```diff
@@ -0,0 +1,50 @@
...
+from pyomo.environ import value
+from watertap.examples.flowsheets.full_treatment_train.analysis import (flowsheet_NF,
+                                                               flowsheet_NF_no_bypass,
+                                                               flowsheet_single_stage,
+                                                               flowsheet_two_stage,
+                                                               flowsheet_NF_two_stage,
+                                                               flowsheet_softening,
+                                                               flowsheet_softening_two_stage)
+from watertap.examples.flowsheets.full_treatment_train.analysis.multi_sweep import *
+from watertap.tools.parameter_sweep import _init_mpi, LinearSample, parameter_sweep
```
bknueven: (Not currently using this)
```suggestion
```

watertap/examples/flowsheets/full_treatment_train/analysis/tests/test_multi_sweep.py
```diff
@@ -0,0 +1,50 @@
...
+import pytest
+import os
+from pyomo.environ import value
+from watertap.examples.flowsheets.full_treatment_train.analysis import (flowsheet_NF,
+                                                               flowsheet_NF_no_bypass,
+                                                               flowsheet_single_stage,
+                                                               flowsheet_two_stage,
+                                                               flowsheet_NF_two_stage,
+                                                               flowsheet_softening,
+                                                               flowsheet_softening_two_stage)
```
bknueven: It seems like these imports aren't needed.
```suggestion
```

bknueven (commented):
Just a few more suggested cleanups in the new test. I like how much the code coverage has already increased.

watertap/examples/flowsheets/full_treatment_train/analysis/tests/test_multi_sweep.py
eyoung55: agreed, fixed!

watertap/examples/flowsheets/full_treatment_train/analysis/tests/test_multi_sweep.py
eyoung55: yep, added the parameterize but kept the old way of testing, fixed now!

watertap/examples/flowsheets/full_treatment_train/analysis/tests/test_multi_sweep.py
eyoung55: good cleanup, fixed.

eyoung55 (dismissed):
Looks good to me, thanks for knocking this out @k1nshuk!

watertap/examples/flowsheets/full_treatment_train/analysis/tests/test_multi_sweep.py
```diff
@@ -0,0 +1,26 @@
...
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and license
+# information, respectively. These files are also available online at the URL
+# "https://github.com/watertap-org/watertap/"
+#
+###############################################################################
+import pytest
+from watertap.examples.flowsheets.full_treatment_train.analysis.multi_sweep import *
+
+pytest_parameterize_list = []
+for case_num in [1, 2, 3, 4, 8, 9]:
```
bknueven: Last question: what's going on with cases 5, 6, and 7?

k1nshuk: Cases 4, 5, 6, and 7 seem to be using the same flow sheet: NF Two stage, and the multi_sweep tests add a good amount of time to the pytests. So in the interest of keeping the time elapsed for pytests in check, I chose to exclude them. I can include them if we don't mind the additional test time.

bknueven: It looks like we should include one of case 5, 6 or 7, based on that criteria: case 4 uses `flowsheet_NF`, and case 5, 6 and 7 use `flowsheet_NF_two_stage`.

So, maybe we should include case 6, since it only has a single parameter to sweep? And leave a comment in this file explaining why we're skipping 5 and 7?

watertap/examples/flowsheets/full_treatment_train/analysis/multi_sweep.py
```diff
@@ -94,7 +94,7 @@ def run_analysis(case_num, nx, RO_type, interp_nan_outputs=True):
 
     elif case_num == 4:
         # ================================================================
-        # NF Two Stage
+        # NF
```
bknueven, line 97: You can't always trust the comments :-)

bknueven (approved):

eyoung55 (approved):
Looking even better now, thanks for the help, @bknueven.

PR #375, opened 2022-03-03
Migrating WT3 landfill model

luohezhiming:
## Summary/Motivation:
add the landfill model

- add the landfill unit model
- add corresponding tests

### Legal Acknowledgement

### DIFF
--- a/watertap/unit_models/zero_order/landfill_zo.py
+++ b/watertap/unit_models/zero_order/landfill_zo.py
@@ -1,0 +1,60 @@
+###############################################################################
+# WaterTAP Copyright (c) 2021, The Regents of the University of California,
+# through Lawrence Berkeley National Laboratory, Oak Ridge National
+# Laboratory, National Renewable Energy Laboratory, and National Energy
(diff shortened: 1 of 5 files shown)


bknueven (approved):

watertap/unit_models/zero_order/landfill_zo.py
```diff
@@ -0,0 +1,60 @@
...
+        self._tech_type = "landfill"
+
+        build_pt(self)
+        constant_intensity(self)
+
+        self.capacity_basis = Var(self.flowsheet().time,
+                                units=pyunits.kg/pyunits.hr,
+                                doc="capacity basis for capital cost")
+
+        self.total_mass = Var(self.flowsheet().time,
```
adam-a-a, line 47: I'm not sure that we'd want to have total mass as a separate variable since we can just sum the component mass flow rates. @andrewlee94 thoughts? This would be used in costing I think.

adam-a-a (approved):
Since `total_mass` is used in costing and we aren't adding Expressions in unit models (sticking to variables), we can merge this.

PR #380, opened 2022-03-03
Migrate WT3 co2_addition unit

kurbansitterley:
## Summary/Motivation:
Migrate WT3 co2_addition unit

- create co2_addition model, test files, and .yaml

### Legal Acknowledgement


watertap/data/techno_economic/co2_addition.yaml
```diff
@@ -0,0 +1,20 @@
+default:
+  energy_electric_flow_vol_inlet:
+    value: 0.0
+    units: kWh/m^3
+  specific_capital_cost_flow_vol_inlet:
+    base_fixed_cap_cost:
```
adam-a-a: I actually labeled `base_fixed_cap_cost` as `cap_basis_par` in other merged PRs, with the thought that it might be easier to use that same name (just a different cost_function) when @andrewlee94 brings costing in. However, I'm not sure how useful that'd be.

watertap/data/techno_economic/co2_addition.yaml
```diff
@@ -0,0 +1,20 @@
...
+      value: 0.7
+      units: dimensionless
+    vol units:
+      value: Mgallons/day
+    capital cost units:
+      value: M$
+    year:
+      value: 2019
+    cost_function_form:
+      function f(flow_vol_in): base_fixed_cap_cost*flow_vol_in^cap_exp
```
adam-a-a: If we do change `base_fixed_cap_cost`, we'd have to change it here too

watertap/data/techno_economic/co2_addition.yaml
kurbansitterley: That is fine with me. I wasn't sure so was just using the variables in WT3.

watertap/data/techno_economic/co2_addition.yaml
kurbansitterley: Yes, and maybe a bigger question for @andrewlee94 - should I be updating .yaml files with the costing parameters?

kurbansitterley:
```suggestion
      function f(flow_vol_in): cap_basis_par*flow_vol_in^cap_exp_par
```

watertap/data/techno_economic/co2_addition.yaml
kurbansitterley:
```suggestion
    cap_basis_par:
```

watertap/data/techno_economic/tests/test_unit_parameter_files.py
```diff
@@ -52,6 +52,7 @@ def test_unit_parameter_files(tech):
                     "municipal_drinking",
                     "pump",
                     "storage_tank",
+                    "co2_addition"
```
bknueven:
```suggestion
                    "co2_addition",
```

bknueven (dismissed):

bknueven (approved):

adam-a-a (approved):

watertap/data/techno_economic/co2_addition.yaml
andrewlee94: I used a different name myself in the costing - for now I suggest we not touch these until we can review the changes in the costing code and settle on a convention.

watertap/data/techno_economic/co2_addition.yaml
andrewlee94: Not yet - a lot of this information isn't actually needed, and we need to settle on the naming conventions which would best be done when looking at the costing code.

PR #386, opened 2022-03-04
Costing package for Zero Order models

andrewlee94:
## Summary/Motivation:
This PR adds a general purpose costing package for zero-order models.

-
-

### Legal Acknowledgement


watertap/core/zero_order_costing.py
```diff
@@ -0,0 +1,221 @@
...
+        """
+        # Set the base year for all costs
+        self.base_currency = pyo.units.USD_2018
+        # Set a base period for all operating costs
+        self.base_period = pyo.units.year
+
+        # Define expected flows
+        self.defined_flows = {
+            "electricity":
+                0.0595*pyo.units.USD_2019/pyo.units.kW/pyo.units.hour}
```
luohezhiming: Unsure why we use `USD_2019` here, the base year is set as `USD_2018`.

watertap/core/tests/test_zero_order_costing.py
```diff
@@ -0,0 +1,128 @@
...
+            "database": model.db})
+
+        model.fs.unit.inlet.flow_mass_comp[0, "H2O"].fix(10000)
+        model.fs.unit.inlet.flow_mass_comp[0, "sulfur"].fix(1)
+        model.fs.unit.inlet.flow_mass_comp[0, "toc"].fix(2)
+        model.fs.unit.inlet.flow_mass_comp[0, "tss"].fix(3)
+        model.fs.unit.load_parameters_from_database()
+        assert degrees_of_freedom(model.fs.unit) == 0
+
+        model.fs.costing = ZeroOrderCosting()
```
adam-a-a: Why do we need to make an instance of ZeroOrderCosting and then replace with UnitModelCostingBlock()?

andrewlee94: We don't. Note the first one is attached to `m.fs`, and the second one to `m.fs.unit`.

adam-a-a: Oops- you didn't replace anything. This is on flowsheet level and the next is on the unit level.

watertap/core/tests/test_zero_order_costing.py
```diff
@@ -0,0 +1,144 @@
...
+
+        model.fs.unit.costing = UnitModelCostingBlock(default={
+            "flowsheet_costing_block": model.fs.costing,
+            "costing_method": ZeroOrderCostingData.exponential_flow_form})
+
+        assert isinstance(model.fs.costing.nanofiltration, Block)
+        assert isinstance(model.fs.costing.nanofiltration.capital_a_parameter,
+                          Var)
+        assert isinstance(model.fs.costing.nanofiltration.capital_a_parameter,
+                          Var)
```
TimBartholomew: duplicate line

watertap/core/zero_order_costing.py
```diff
@@ -0,0 +1,333 @@
...
```
TimBartholomew: Is there no kWh unit?

TimBartholomew: I just checked pint and they support a "watt_hour" or "watthour", so I think "kwatt_hour" or "kwatthour" could work.

watertap/core/zero_order_costing.py
```diff
@@ -0,0 +1,333 @@
...
+                blk.unit_model.electricity[t0], "electricity")
+
+    # -------------------------------------------------------------------------
+    # Map costing methods to unit model classes
+    unit_mapping = {ZeroOrderBase: exponential_flow_form}
+
+
+def _add_tech_parameter_block(blk, parameter_dict):
+    # Parameters for this technology haven't been added yet
+    pblock = pyo.Block()
```
TimBartholomew: Is the idea that this would support different forms? Or is this exponential form the standard for WT3?

watertap/core/zero_order_costing.py
```diff
@@ -0,0 +1,333 @@
...
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and license
+# information, respectively. These files are also available online at the URL
+# "https://github.com/watertap-org/watertap/"
+#
+###############################################################################
+"""
+General costing package for zero-order processes.
+"""
+import pyomo.environ as pyo
```
TimBartholomew, line 19: I know this isn't a big deal, but in WaterTAP and most IDAES models (I think) we generally import the objects from pyomo environ rather than use pyo.Var() for example.

TimBartholomew (commented):
Looks pretty good to me. I'm not sure what the name should be. At least when I imagined us making multiple costing packages, I thought one clear way to name them is by their reference. For example, a textbook or a paper. Following that thought, this could be called watertap3 costing package and the reference points to the published documentation.

andrewlee94:
@TimBartholomew I (and others) was envisioning this would be like the rest of the WT3 models- there is a single core model which loads parameters from a provided YAML file (I am working on that feature now). Thus naming is not an issue.

watertap/core/zero_order_costing.py
andrewlee94: This method is only for exponential forms (which is the majority of the WT3 models). There are a few however which currently use custom forms (but those also tend to be the ones regressed from data).

watertap/core/zero_order_costing.py
andrewlee94: This is the standard form Pyomo themselves use for their documentation. We might want to standardize on that form for public facing code to help users.

watertap/core/zero_order_costing.py
```diff
@@ -0,0 +1,510 @@
...
+            doc="Insurance and taxes as % FCI")
+
+        self.wacc = pyo.Var(initialize=0.05,
+                            units=pyo.units.dimensionless,
+                            doc='Weighted Average Cost of Capital (WACC)')
+        self.capital_recovery_factor = pyo.Expression(
+            expr=((self.wacc *
+                   (1 + self.wacc)**(self.plant_lifetime/self.base_period)) /
+                  (((1 + self.wacc)**(self.plant_lifetime/self.base_period)) -
+                   1) / self.base_period))
```
kurbansitterley, line 115: ``base_period`` isn't in the WT3 formula for ``capital_recovery_factor`` - what is it for?

watertap/core/zero_order_costing.py
```diff
@@ -0,0 +1,510 @@
...
+        # Fix all Vars from database
+        for v in global_params:
+            try:
+                value = cs_def["global_parameters"][v]["value"]
+                units = cs_def["global_parameters"][v]["units"]
+                getattr(self, v).fix(value*getattr(pyo.units, units))
+            except KeyError:
+                raise KeyError(
+                    f"Invalid case study definition file - no entry found "
+                    f"for {v}, or entyr lacks value and units.")
```
kurbansitterley:
```suggestion
                    f"for {v}, or entry lacks value and units.")
```

watertap/core/zero_order_costing.py
andrewlee94: This is required because plant lifetime has units, and you cannot have units in an exponent. Notably, the exponent is not so much the plant lifetime, as the number of accounting cycles over the plant life (in this case the accounting cycle (i.e. base period) is yearly, so the effect is to just cancel units).

watertap/core/zero_order_costing.py
```diff
@@ -0,0 +1,510 @@
...
+        This method also registers the chemical flow and electricity demand as
+        costed flows.
+        """
+        chem_name = blk.unit_model.config.process_subtype
+
+        t0 = blk.flowsheet().time.first()
+        chem_flow_mass = (blk.unit_model.chemical_dosage[t0] *
+                          blk.unit_model.properties[t0].flow_vol /
+                          blk.unit_model.ratio_in_solution)
+        sizing_term = chem_flow_mass / (pyo.units.lb/pyo.units.day)
```
kurbansitterley: Sort of hard for me to follow but does ``sizing_term`` go to gal/day at some point? Most (if not all) of the costing relationships for chemical additions are based off solution flow in gal/day.

watertap/data/techno_economic/default_case_study.yaml
```diff
@@ -0,0 +1,89 @@
...
+    purity: 1
+global_parameters:
+  plant_lifetime:
+    value: 30
+    units: year
+  utilization_factor:
+    value: 1
+    units: dimensionless
+  land_cost_percent_FCI:
+    value: 0.07
```
kurbansitterley:
```suggestion
    value: 0.0015
```

watertap/data/techno_economic/default_case_study.yaml
```diff
@@ -0,0 +1,89 @@
...
+    units: dimensionless
+  working_capital_percent_FCI:
+    value: 0.008
```
kurbansitterley:
```suggestion
    value: 0.05
```

watertap/data/techno_economic/default_case_study.yaml
```diff
@@ -0,0 +1,89 @@
...
+    units: dimensionless
+  salaries_percent_FCI:
+    value: 0.07
```
kurbansitterley:
```suggestion
    value: 0.001
```

watertap/data/techno_economic/default_case_study.yaml
kurbansitterley: I see where you get these values- as the initialized values for the Var() in WT3... but they are fixed to something somewhere else. I know it is confusing so apologies

watertap/data/techno_economic/default_case_study.yaml
```diff
@@ -0,0 +1,89 @@
...
+    units: 1/year
+  benefit_percent_of_salary:
+    value: 0.07
```
kurbansitterley:
```suggestion
    value: 0.9
```

watertap/data/techno_economic/default_case_study.yaml
```diff
@@ -0,0 +1,89 @@
...
+    units: dimensionless
+  maintenance_costs_percent_FCI:
+    value: 0.07
```
kurbansitterley:
```suggestion
    value: 0.008
```

watertap/data/techno_economic/default_case_study.yaml
```diff
@@ -0,0 +1,89 @@
...
+    units: 1/year
+  laboratory_fees_percent_FCI:
+    value: 0.07
```
kurbansitterley:
```suggestion
    value: 0.003
```

watertap/data/techno_economic/default_case_study.yaml
```diff
@@ -0,0 +1,89 @@
...
+    units: 1/year
+  insurance_and_taxes_percent_FCI:
+    value: 0.07
```
kurbansitterley:
```suggestion
    value: 0.002
```

watertap/core/zero_order_costing.py
andrewlee94: Not sure why I did this either - I think I got confused by the units for costing the flow versus the capital cost.

luohezhiming (approved):

watertap/core/zero_order_costing.py
andrewlee94: This is because the cost for electricity was provided in 2019 dollars. It automatically gets converted tot the base currency when it is used however.

PR #392, opened 2022-03-08
WT3 Santa Barbara case 

TimBartholomew:
## Fixes/Addresses:
None

## Summary/Motivation:
- WaterTAP needs to demonstrate it can simulate WT3 cases

-
-

### Legal Acknowledgement


TimBartholomew:
@kurbansitterley this PR currently doesn't include much that is specific to Santa Barbara, for example defaults are all used for pumps, chlorination, static mixer, media filtration, cartridge filtration, and RO.

watertap/examples/case_studies/santa_barbara/santa_barbara.py
```diff
@@ -0,0 +1,443 @@
...
+    # ---posttreatment---
+    # storage tank 2
+    psttrt.storage_tank_2.load_parameters_from_database(use_default_removal=True)
+    psttrt.storage_tank_2.storage_time.fix(1)
+
+    # uv aop
+    m.db.get_unit_operation_parameters("uv_aop")
+    psttrt.uv_aop.load_parameters_from_database(use_default_removal=True)
+    psttrt.uv_aop.uv_reduced_equivalent_dose.fix(350)  # TODO: check this was the right thing to fix
+    psttrt.uv_aop.uv_transmittance_in.fix(0.95)  # TODO: check this was the right thing to fix
```
adam-a-a: These look like the right things to fix. And by loading parameters, I suppose you also fixed `oxidant_dose` to its default value of 5 mg/L which matches what I see in `treatment_train_setup.csv`

adam-a-a (commented):
Everything looks good. Do we think there should be flowsheet level expressions for specific energy consumption of the whole plant and overall recovery rate? I'm not sure what system-level metrics were reported for case studies @kurbansitterley , but these two seem like straightforward ones that we should have (aside from LCOW, which can come in when costing goes through).

andrewlee94:
@adam-a-a The zero-order costing package already has methods to report both LCOE and overall electricity intensity (intensity is not a cost per se, but the costing framework already includes the functionality to aggregate the electricity demands).

adam-a-a:
> @adam-a-a The zero-order costing package already has methods to report both LCOE and overall electricity intensity (intensity is not a cost per se, but the costing framework already includes the functionality to aggregate the electricity demands).

@andrewlee94 So basically, the costing package will give us our specific energy consumption (or at least an intermediate to translate to whatever representation we want to provide for energy consumption, e.g., total energy consumption divided by total recovered water). Is that right?

andrewlee94:
@adam-a-a More accurately, the costing package will give you the total consumption of electricity (and any other chemicals as well).

However, it (by itself) does not know what the ultimate feed and/or product flow rates are, so this needs to be provided in order to calculate specific quantities. As I said, I have already added a `add_electricity_intensity` method, however this has one required argument which is the flow rate of water to use in the calculation.

watertap/examples/case_studies/santa_barbara/santa_barbara.py
```diff
@@ -0,0 +1,443 @@
...
+
+    # uv aop
```
kurbansitterley:
```suggestion
    # co2 addition
```

kurbansitterley:
> Everything looks good. Do we think there should be flowsheet level expressions for specific energy consumption of the whole plant and overall recovery rate? I'm not sure what system-level metrics were reported for case studies @kurbansitterley , but these two seem like straightforward ones that we should have (aside from LCOW, which can come in when costing goes through).

System level metrics reported in WT3:

- LCOW (we also reported LCOW for each costing component. For example "TCI LCOW" is the LCOW but only calculated with TCI and leaving out operating costs, "Chemical LCOW" leaves out TCI and all other operating expenses besides chemical costs, etc.... some teams requested these metrics so we just put them all in, I don't know how conventional they are in TEA analysis)
- Capital costs
- Operating (total, fixed, electricity, chemical, other, also reported all the fixed operating components like salaries, tax/insurance, etc.) 
- Electricity intensity
- Water recovery
- Flow rates (inlet, treated, waste)

Operating costs are reported for the plant lifetime and for an annual basis. We also reported most of the same metrics for each unit process as well as the constituent flows for each unit process (which seemed a bit overkill to me, but it did help in troubleshooting some issues along the way).

I ended up putting pretty much every metric calculated in the model in the results file. Also reported all the TEA inputs.

kurbansitterley (dismissed):
This all looks good to me. Appears all the values that were fixed correspond to the values we used in WT3.

adam-a-a (approved):

luohezhiming (approved):

PR #393, opened 2022-03-08
WaterTAP Costing

bknueven:
## Summary/Motivation:
Switch existing WaterTAP flowsheets to use the new IDAES costing framework in IDAES/idaes-pse#741.

- [x] Switch tracking IDAES to Andrew's branch #397  
- [x] Create new `WaterTAPCosting` Package, which replaces existing scattered financials modules
- [ ] Tests for new `WaterTAPCosting` Package
- [ ] Documentation for new `WaterTAPCosting` Package

### Legal Acknowledgement


watertap/costing/watertap_costing_package.py
```diff
@@ -0,0 +1,357 @@
...
+        ReverseOsmosis0D,
+        ReverseOsmosis1D,
+        NanoFiltration0D,
+        NanoFiltrationZO,
+        PressureExchanger,
+        Pump,
+        )
+
+class ROType(str, Enum):
+    standard = "standard"
```
andrewlee94, line 39: FYI, pylint doesn't seem to like this for some reason. However, it you first declare a `StrEnum` class which defines `__str__` and inherit that it seems to work (or at least hide the issue from pylint). For convenience, IDAES defined a `StrEnum` class in `idaes-pse.core.util.misc`.

watertap/costing/watertap_costing_package.py
```diff
@@ -0,0 +1,357 @@
...
+    pressure_exchangers = "pressure_exchanger"
+    
+    def __str__(self):
+        return self.value
+
+@declare_process_block_class("WaterTAPCosting")
+class WaterTAPCostingData(FlowsheetCostingBlockData):
+
+    # Register currency and conversion rates based on CE Index
+    register_idaes_currency_units()
```
andrewlee94: Another FYI, but this can be moved inside the `build_global_params` method (just as long as it is done before you try to use the currency units).

watertap/costing/watertap_costing_package.py
```diff
@@ -0,0 +1,357 @@
...
+
+
+    def build_global_params(self):
+
+        # Set the base year for all costs
+        self.base_currency = pyo.units.USD_CE500
```
andrewlee94: I would suggest we just pick a base year for this rather than CE500. In the end it doesn't mean much as we can always convert values, but a year means a lot more than some arbitrary point.

watertap/costing/watertap_costing_package.py
```diff
@@ -0,0 +1,357 @@
...
+        # Set a base period for all operating costs
+        self.base_period = pyo.units.year
+
+        # Define standard material flows and costs
+        self.defined_flows["electricity"] = 0.07 * pyo.units.USD_CE500 / pyo.units.kWh
```
andrewlee94: Defining a year is probably important here, as the value was chosen for some reference point. Note that the year here dos not need to match the base year.

watertap/costing/watertap_costing_package.py
```diff
@@ -0,0 +1,357 @@
...
+                initialize=0.1,
+                doc='Capital annualization factor [fraction of investment cost/year]',
+                units=pyo.units.year**-1)
+        self.factor_membrane_replacement = pyo.Var(
+                initialize=0.2,
+                doc='Membrane replacement factor [fraction of membrane replaced/year]',
+                units=pyo.units.year**-1)
+        self.reverse_osmosis_membrane_cost = pyo.Var(
+                initialize=30,
+                doc='Membrane cost [$/m2]',
```
andrewlee94: Minor comment but I generally avoid putting units in docs strings now that we have explicit units on the Vars - it just adds the potential for the tow to get out of synch.

watertap/costing/watertap_costing_package.py
```diff
@@ -0,0 +1,357 @@
...
+        Add specific energy consumption (kWh/m**3) to costing block.
+        Args:
+            flow_rate - flow rate of water (volumetric) to be used in
+                        calculating specific energy consumption
+        """
+
+        self.specific_energy_consumption = pyo.Expression(expr = 
+                self.aggregate_flow_electricity / pyo.units.convert(flow_rate, to_units=pyo.units.m**3/pyo.units.hr))
+
+    def cost_nanofiltration(blk):
```
andrewlee94, line 251: This method appears to be missing the `@staticmethod` decorator.

watertap/costing/watertap_costing_package.py
```diff
@@ -0,0 +1,357 @@
...
+        """
+        _make_captial_cost_var(blk)
+        cost_by_flow_volume(blk, blk.costing_package.pressure_exchanger_cost,
+                pyo.units.convert(blk.unit_model.low_pressure_side.properties_in[0].flow_vol, (pyo.units.meter**3/pyo.units.hours)))
+
+    ## TODO; Mixer and Separator
+
+    # Define default mapping of costing methods to unit models
+WaterTAPCostingData.unit_mapping = {
+        Pump: WaterTAPCostingData.cost_pump,  # TODO: this may be carrying over a limitation of IDAES
```
andrewlee94: Is this comment referring to the issue with scaling, and if so can it be removed? If not , what issue is it referring to?

andrewlee94 (commented):
A few minor comments for thoguh.

watertap/costing/watertap_costing_package.py
bknueven: @TimBartholomew @adam-a-a do you think there's a "best year" we can use for the ProteusLib financials package? Ideally we'd pick a single year and use it so our results are replicate, at least for now.

bknueven:
This could use more documentation and testing, though because it's used in several places the new WaterTAPCosting package already has good coverage.

We might merge this more-or-less as-is and check off the last two boxes in a subsequent PR.

andrewlee94 (dismissed):
Looks good to me - someone should confirm the changes to the results in a few places are not significant however.

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
```diff
@@ -94,22 +95,21 @@ def build():
...
+    m.fs.P2.costing = UnitModelCostingBlock(default={
+            "flowsheet_costing_block":m.fs.costing})
+            
+    m.fs.RO.costing = UnitModelCostingBlock(default={
+            "flowsheet_costing_block":m.fs.costing})
+    m.fs.PXR.costing = UnitModelCostingBlock(default={
+            "flowsheet_costing_block":m.fs.costing})
+    m.fs.costing.cost_process()
+    m.fs.costing.add_LCOW(m.fs.product.properties[0].flow_vol)
+    m.fs.costing.add_specific_energy_consumption(m.fs.product.properties[0].flow_vol)
```
adam-a-a, line 112: There could be cases where we want more than one specific energy consumption (e.g., wrt to product flow and/or wrt feed flow). We could always create an expression ourselves outside of this to calculate an alternative SEC I suppose, but just wanted to point that out. Right now, it looks like calling `add_specific_energy_consumption` more than once would overwrite the previous expression.

andrewlee94: @adam-a-a At that point however, it is probably easier for the user to create/add the expressions manually. Otherwise, we need to start doing string manipulations, which get messy.

bknueven:
> Looks good to me - someone should confirm the changes to the results in a few places are not significant however.

Summarizing based on my understanding:
-  RO_with_energy_recovery: The small differences in LCOW are due to electricity being costed over an astronomical year (as pint does converting seconds to year) as opposed to the calendar year we were using before.
- test_flowsheet_NF: I am uncertain about this one, but the difference is small. Both `main` and `costing` solve to the same objective value.
- LSRRO: Where LCOW goes down its because of a bug in how were we costing electricity which is fixed in this PR, and @adam-a-a had previously discovered. We were over-estimating the electricity cost by a factor of 4-5. The switches from 1.0 to some other value is due to proper initialization of the flowsheet costing block.

watertap/examples/flowsheets/full_treatment_train/model_components/mixer.py
```diff
@@ -11,29 +11,5 @@
...
-        super().build()
-
-    def calculate_scaling_factors(self):
-        super().calculate_scaling_factors()
-
-    def get_costing(self, module=None, **kwargs):
-        self.costing = Block()
-        module.Mixer_costing(self.costing, **kwargs)
-
+from idaes.generic_models.unit_models import Mixer
```
adam-a-a, line 15: At some point, I guess we could delete the Mixer and Separator files here under model components and just import them where needed directly in associated flowsheets. We originally did this to handle costing of these units with our financials file.

watertap/examples/flowsheets/lsrro/tests/test_lsrro.py
```diff
@@ -256,16 +260,16 @@ class TestLSRRO_1Stage(_TestLSRRO):
...
+    number_of_constraints = 193
 
     display_system = \
 """----system metrics----
 Feed: 1.00 kg/s, 70000 ppm
 Product: 0.180 kg/s, 1592 ppm
 Volumetric water recovery: 18.0%
 Energy Consumption: 3.8 kWh/m3
-Levelized cost of water: 2.48 $/m3
+Levelized cost of water: 1.73 $/m3
```
adam-a-a, line 272: This looks like a pretty significant change in LCOW. Was this because you factored in the recovered energy from the ERD that may have been omitted previously in this result (i.e., subtracted electricity cost due to energy recovery)?

watertap/examples/flowsheets/full_treatment_train/model_components/mixer.py
bknueven: Of course -- this is merely expedience.

watertap/examples/flowsheets/lsrro/tests/test_lsrro.py
```diff
@@ -498,10 +502,10 @@ def optimization_data(self, model):
...
 
         data[fs.product.flow_mass_phase_comp[0,'Liq','H2O']]   = 0.732036
-        data[fs.product.flow_mass_phase_comp[0,'Liq','NaCl']]  = 0.445058e-3
-        data[fs.disposal.flow_mass_phase_comp[0,'Liq','H2O']]  = 0.197964
-        data[fs.disposal.flow_mass_phase_comp[0,'Liq','NaCl']] = 0.695545e-1
-        data[fs.costing.LCOW]   = 1.51258
+        data[fs.product.flow_mass_phase_comp[0,'Liq','NaCl']]  = 0.448848e-3
+        data[fs.disposal.flow_mass_phase_comp[0,'Liq','H2O']]  = 0.197967
+        data[fs.disposal.flow_mass_phase_comp[0,'Liq','NaCl']] = 0.695512e-1
+        data[fs.costing.LCOW]   = 1.46933
```
adam-a-a, line 508: Just want to make sure the changes in LCOW are only attributed to reduction in electricity cost (added operating cost for pressure exchanger which will be a negative value)

watertap/unit_models/coag_floc_model.py
```diff
@@ -446,7 +446,7 @@ def fix_tss_turbidity_relation_defaults(self):
         self.intercept.fix()
 
     # initialize method
-    def initialize(
+    def initialize_build(
```
adam-a-a, line 449: Why are we changing to initialize_build for unit models?

watertap/examples/flowsheets/full_treatment_train/model_components/mixer.py
andrewlee94: As a note, IDAES is discussing the idea of having "model libraries" where we import a selection of models of interest to a specific application, so that users can go to a single central location to import the models of interest.

watertap/examples/flowsheets/lsrro/tests/test_lsrro.py
bknueven: The specific energy consumption, without the ERD, is something like 13-14 kWh/m^3 for this process. I didn't carefully cross-check this; the reduction in LCOW seem about right to me.

The correct way to do this would be to fix the original bug first so we could have a new baseline. Discovering it in process here makes doing proper cross-validation more difficult.

watertap/costing/watertap_costing_package.py
```diff
@@ -0,0 +1,501 @@
...
+                units=pyo.units.dimensionless)
+
+        # fix the parameters
+        for var in self.component_objects(pyo.Var):
+            var.fix()
+
+        # Define standard material flows and costs
+        self.defined_flows["electricity"] = self.electricity_base_cost
+        self.defined_flows["NaOCl"] = 74.44e-3 * (pyo.units.kg/pyo.units.mol) * (self.naocl_cost/self.naocl_purity)
+        self.defined_flows["CaOH2"] = 74.093e-3 * (pyo.units.kg/pyo.units.mol) * (self.caoh2_cost/self.caoh2_purity)
```
adam-a-a, line 172: Should these values (74.44e-3 and 74.093e-3) be Params? These are the molecular weights of each chemical.

watertap/costing/watertap_costing_package.py
```diff
@@ -0,0 +1,501 @@
...
+            raise BurntToast(f"Unrecognized pump_type: {pump_type}")
+
+    @staticmethod
+    def cost_high_pressure_pump(blk):
+        """
+        High pressure pump costing method
+
+        TODO: describe equations
+        """
+        _make_captial_cost_var(blk)
```
```suggestion
        _make_capital_cost_var(blk)

```

watertap/unit_models/coag_floc_model.py
andrewlee94: This is the new way to do things in IDAES - it is needed to handle initialization of costing blocks (and potentially other model add-ons).

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
adam-a-a: That works. I just wanted everyone to be aware of that for SEC (and LCOW); there can be more than one variation of interest in a given analysis.

watertap/costing/watertap_costing_package.py
```diff
@@ -0,0 +1,501 @@
...
+        self.factor_maintenance_labor_chemical = pyo.Var(
+                initialize=0.03,
+                doc='Maintenance-labor-chemical factor [fraction of investment cost/year]',
+                units=pyo.units.year**-1)
+        self.factor_capital_annualization = pyo.Var(
+                initialize=0.1,
+                doc='Capital annualization factor [fraction of investment cost/year]',
+                units=pyo.units.year**-1)
+        self.factor_membrane_replacement = pyo.Var(
+                initialize=0.2,
```
adam-a-a, line 88: I ran the LSRRO test on my end without the new costing. In one example, the old erroneous LCOW was $2.48/m3 while the new one accounting for energy savings ends up being $1.43/m3. This result is even lower than the new value resulting from the new costing approach. So technically, your test values in LSRRO for LCOW are a little higher than expected, and this can be explained in part by the membrane replacement rate here.  I think we used 0.15 (not 0.2) for LSRRO originally, so this would end up increasing LCOW.

bknueven: No, it is `0.2` in the existing financials module, unless this is updated somewhere else: https://github.com/watertap-org/watertap/blob/b89075c774e9c1eca0a0a64c1b9bf313a27c1868/watertap/examples/flowsheets/lsrro/financials.py#L36-L38

adam-a-a: OK - I must've set it back to the intended 0.15 then locally. I reran with 0.2, and the LCOW is now 1.48, so the new result is still about 16.8% higher. Should I make a separate PR to update the baseline (account for the bug and add the separate membrane costs for ro and lsrro in financials)? There could be another local change I made to financials that is throwing things off, so we can settle this by making that separate PR. What do you think @bknueven ?

bknueven: @adam-a-a that sounds reasonable to me.

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
bknueven: We could provide an optional argument which would allow the user to determine the name of the SEC Expression and LCOW Var/Constraint. That would give flexibility while keeping the simple use case simple (e.g., no weird indexing by flow name or something similar).

adam-a-a (dismissed):
Although there might be little quirks (with LSRRO from what I could see) that we'll want to revisit to ensure that any peculiarities were _not _ attributed to the costing implementation, I think we should go ahead and merge this so that costing of zero-order models can proceed.

watertap/costing/watertap_costing_package.py
bknueven: Maybe? My main thought on hard-coding them is that unlike `electricity_base_cost`, I can't imagine needing a reason for the user to set these.

andrewlee94: Out of interest, why do you even need the mole to mass conversion? Costing supports both mass and mole based flows, so it would make more sense to have the cost and flow use the same units and not have to worry about conversion.

bknueven: I doubt it is strictly necessary -- I'm just replicating what was in the financials module.

andrewlee94 (dismissed):

setup.py
```diff
@@ -43,7 +43,8 @@
     # update with a tag from the nawi-hub/idaes-pse
     # when a version of IDAES newer than the latest stable release from PyPI
     # will become needed for the watertap development
-    "idaes-pse[prerelease] @ https://github.com/watertap-org/idaes-pse/archive/1.12.1.watertap.2022.02.04.zip",
+    #"idaes-pse[prerelease] @ https://github.com/watertap-org/idaes-pse/archive/1.12.1.watertap.2022.02.04.zip",
+    "idaes-pse[prerelease] @ git+https://github.com/andrewlee94/idaes-pse.git@costing_2"
```
```suggestion
    "idaes-pse[prerelease] @ https://github.com/watertap-org/idaes-pse/archive/1.14.0dev0.watertap.2022.03.11.zip"
```

I've just created the tag on watertap-org/idaes-pse: https://github.com/watertap-org/idaes-pse/tree/1.14.0dev0.watertap.2022.03.11

TimBartholomew (approved):

andrewlee94 (approved):

Issue #156, opened 2021-07-07
Clean up pre-processing for records loaded into EDB


dangunter:
Pre-processing of input to be placed in the DB is being split between the database class and the data model classes, which is somewhat messy and confusing. Should either all be in one place or perhaps even pulled out into its own module.

aladshaw3:
@dangunter Can you provide some specific guidance on this issue? I might be able to help clean things up.

aladshaw3:
Code in 2 different modules essentially doing same thing. They should be put into 1 class to remove redundancy.  

Comments somewhere in code about this.

dangunter:
See lines 435-436 in  [watertap.edb.db_api](https://github.com/watertap-org/watertap/blob/main/watertap/edb/db_api.py) module
```
   # XXX: This preprocessing overlaps with data_model.DataWrapper subclasses.
    # XXX: It should all be moved to one place
```

aladshaw3:
> See lines 435-436 in [watertap.edb.db_api](https://github.com/watertap-org/watertap/blob/main/watertap/edb/db_api.py) module
> 
> ```
>    # XXX: This preprocessing overlaps with data_model.DataWrapper subclasses.
>     # XXX: It should all be moved to one place
> ```

@dangunter I am still confused as to why the pre-processing needs to be cleaned up. As far as I can tell, there is only 1 '_preprocess' function under the 'DataWrapper' classes (line 816 [here](https://github.com/watertap-org/watertap/blob/main/watertap/edb/data_model.py)) and no other sub-class of 'DataWrapper' has a preprocess method defined. 

Also, the preprocess method that is defined for the 'Component' object does not perform the same task as the preprocess method in the 'db_api' function call. So I am not sure what the overlap is you are referring to in your comment. 

Is this perhaps something that was already resolved and the comment was just left in place?

PR #395, opened 2022-03-09
Fixed minor error in EDB and cleaned up some comments/notes about pre-processing

aladshaw3:
## Fixes/Addresses:

#156 

## Summary/Motivation:

Following the desire to clean up pre-processing, dug into the EDB and found a minor issue that would only appear if the database were expanded to include some additional components that may mistakenly place a species as a solvent. For the purposes of WaterTAP, the solvent will only ever be H2O. 

- Fixed minor bug/issue in data_model under preprocess of component 
- Added some comments for future

### Legal Acknowledgement


adam-a-a:
@aladshaw3 There could be at least one case in the future where water isn't the only solvent that WaterTAP _ever_ considers: ionic liquid solvent extraction, where there is water and amine in the process to ultimately remove solutes. That said, for the foreseeable future, water is the only solvent of interest.

aladshaw3:
> @aladshaw3 There could be at least one case in the future where water isn't the only solvent that WaterTAP _ever_ considers: ionic liquid solvent extraction, where there is water and amine in the process to ultimately remove solutes. That said, for the foreseeable future, water is the only solvent of interest.

Ok, I will keep that in mind. This is just part of the automated preprocessing that the EDB does, so in case something was not specified as a solvent, it automatically assigns that role to water. The previous line would have added any species that contained 'H' and 'O' as elements as the solvent, which would have selected components that probably don't make sense.

TimBartholomew (approved):
LGTM

srikanthallu (approved):

adam-a-a (approved):

PR #397, opened 2022-03-09
switching to Andrew's costing branch

bknueven:
## Summary/Motivation:
To keep moving forward with implementing the new IDAES costing framework on WaterTAP, we need to start merging PRs based on https://github.com/IDAES/idaes-pse/pull/741, with the supposition that not much will change at this point.

- Point WaterTAP's idaes-pse[prerelease] to @andrewlee94's costing fork 

### Legal Acknowledgement

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -207,6 +207,11 @@ jobs:
         echo '::group::Output of "conda install" commands'
         conda install --quiet --yes pip=21.1 wheel setuptools pandoc
         echo '::endgroup::'
+        # TODO: COSTING_UPDATE: 6.3.0 isn't bundled on PyPI with some files
+        #       need for PyNumero. For now, we'll install 6.3.0 from GitHub
+        echo '::group::Install Pyomo from GitHub'
+        pip install pyomo https://github.com/pyomo/pyomo/archive/6.3.0.zip
+        echo '::endgroup::'
         echo '::group::Output of "pip install" commands'
         pip install --progress-bar off -r requirements-dev.txt
         echo '::endgroup::'
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -207,13 +207,13 @@ jobs:
         echo '::group::Output of "conda install" commands'
         conda install --quiet --yes pip=21.1 wheel setuptools pandoc
         echo '::endgroup::'
+        echo '::group::Output of "pip install" commands'
+        pip install --progress-bar off -r requirements-dev.txt
+        echo '::endgroup::'
         # TODO: COSTING_UPDATE: 6.3.0 isn't bundled on PyPI with some files
         #       need for PyNumero. For now, we'll install 6.3.1.dev0 from GitHub
         echo '::group::Install Pyomo from GitHub'
         pip install pyomo https://github.com/pyomo/pyomo/archive/043b2a5.zip
-        echo '::endgroup::'
-        echo '::group::Output of "pip install" commands'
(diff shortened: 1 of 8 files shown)


adam-a-a:
It looks like, at least in one of the failures, that when a particular IDAES unit is initialized in the full treatment train example (e.g., an IDAES mixer), the initialization routine goes to use initialization for `unit_costing` in IDAES and can't find `purchase_cost` anymore. Although we had our own costing for the mixer in the financials file for the full treatment train example, the initialization still touches costing on IDAES.

lbianchi-lbl (commented):
I think this is reasonable in general.

My main doubt is whether we should pin to a particular commit hash (a static, immutable reference) rather than the branch name (a movable reference, that will change if commits are pushed to that branch). The main pros/cons of using a commit hash:

- :+1: Until the commit hash is updated in the WaterTAP `setup.py`, the same version of `idaes-pse` will be used; i.e., the outcome of the CI check in WaterTAP is entirely dependent on the content of the WaterTAP repository at any given time
- :-1: The commit hash must be updated whenever a new "version" (i.e. a particular commit on the external branch) is desired

In general, I'd favor the commit hash solution because it's more explicit; however, there's a tradeoff depending on specific details on how we're planning to proceed. Off the top of my head:

- How many PRs/branches here in WaterTAP are expected to strictly depend on this?
  - In other words, how many PRs would have to be updated each time the commit hash needs to be updated in our `setup.py`?

andrewlee94:
@lbianchi-lbl Right now, the bigger issue is that the branch is still waiting for review and thus might see some changes that are important. Ultimately, we would want to pin to a tag or hash on idaes-pse instead, but we are currently stuck waiting for the IDAES PR to be approved and merged.

andrewlee94:
@adam-a-a That is (one of the reasons) why the IDAES costing needed to be rebuilt, as it made far to many presumptions about model forms. The easiest solution here is to either a) switch to the new costing form, or b) add a dummy `initialize` method to the WaterTap costing blocks.

lbianchi-lbl:
> @lbianchi-lbl Right now, the bigger issue is that the branch is still waiting for review and thus might see some changes that are important. Ultimately, we would want to pin to a tag or hash on idaes-pse instead, but we are currently stuck waiting for the IDAES PR to be approved and merged.

@andrewlee94 this makes sense. To clarify, when referring to the "hash vs branch name" decision in my previous comment, I meant it in the context of the (currently unmerged) head branch of IDAES/idaes-pse#741; in other words, whether it's better to refer to individual commit hashes on the `andrewlee94:costing_2` branch, or the `andrewlee94:costing_2` branch (i.e. with an implicit reference to its current HEAD).

bknueven:
I'm actively working on switching our existing financials to the new costing framework in #393. I think we should just `xfail` these tests for now, and I'll remove those as part of #393.

For now, I'm in favor of pointing to the current `HEAD` of the `costing_2` branch. It's obviously more difficult to manage, but I would rather know sooner rather than later if something is broken.

bknueven:
The test failures in macOS are related to Pyomo/pyomo#2317. This was fixed in Pyomo/pyomo#2319, but that has not found its way to a release as-of-yet.

bknueven:
> @adam-a-a That is (one of the reasons) why the IDAES costing needed to be rebuilt, as it made far to many presumptions about model forms. The easiest solution here is to either a) switch to the new costing form, or b) add a dummy `initialize` method to the WaterTap costing blocks.

As #393 now fully implements (a), along with the changes in this PR, I'm going to close this.

PR #403, opened 2022-03-11
Add WT3 crystallizer as a subtype of brine concentrator

luohezhiming:
## Summary/Motivation:

Add WT3 crystallizer unit model as a subtype model of brine concentrator unit model

- add crystallizer as a subtype in the brine_concentrator yaml file
- add electric coefficients as varaibles
- fixed a typo in the original brince_concentrator_zo.py

### Legal Acknowledgement


luohezhiming:
@adam-a-a It seems only the check for nmacOS setup failed. Do you have any idea what happened here?

bknueven:
> @adam-a-a It seems only the check for nmacOS setup failed. Do you have any idea what happened here?

I think #393 will fix this, as this failing test will be skipped. The conda-ipopt was updated earlier today, which I expect is what's causing this sensitive test to fail.

luohezhiming:
> I think https://github.com/watertap-org/watertap/pull/393 will fix this, as this failing test will be skipped. The conda-ipopt was updated earlier today, which I expect is what's causing this sensitive test to fail.

OK. Will rerun it when #393 get merged.

watertap/data/techno_economic/brine_concentrator.yaml
```diff
@@ -144,3 +160,166 @@ default:
...
+  elec_coeff_3:
+    value: 9.47
+    units: kWh/m^3
+    reference: Survey of High-Recovery and Zero Liquid Discharge Technologies for Water Utilities (2008).
+  elec_coeff_4:
+    value: 8.63E-4
+    units: kWh/m^6*hr
+    reference: Survey of High-Recovery and Zero Liquid Discharge Technologies for Water Utilities (2008).
+  recovery_frac_mass_H2O:
+    value: 0.95
```
kurbansitterley, line 182: @luohezhiming did you change this value? Nothing wrong with it if there is a reference, I just know water_recovery is 0.9 in WT3.

luohezhiming: I copy this water_recovery from the original crystallizer.yaml file. The water_recovery for a brine_concentrator is 0.9, but is given 0.95 in the original crystallizer.yaml file.

adam-a-a (approved):
LGTM!

bknueven (commented):
Seems reasonable -- should we be explicitly testing the new subtype?

watertap/data/techno_economic/brine_concentrator.yaml
kurbansitterley: Ah perfect, I missed that this was for the subtype

kurbansitterley:
I am not a pro on what warrants a separate test, but my gut says yes since they recovery is different, they would have different flows

luohezhiming:
@bknueven @kurbansitterley The original test file include a subtype testing `class Testbrine_concentratorZOsubtype`, I think it is OK because other units we migrated that has subtypes are also tested by this way.

andrewlee94 (approved):
My only comment is that I don't think we want the electricity intensity coefficients in the report, but that can be cleaned up later.

PR #404, opened 2022-03-11
First WT3 costing examples

andrewlee94:
## Fixes/Addresses:

Adds example costing and tests for aeration basis and air flotation (plus NF and chemical addition) units.

## Things to Note:
- These examples should work for any basic unit (I hope).
- Make sure to check the year used for the costing, and update if necessary.
- Note also the units for the A parameter - the old YAML format generally uses M$
- Cost factor line is for TPEC or TIC (if required)

### Legal Acknowledgement


luohezhiming (approved):
Looks good to me.

adam-a-a (approved):
I might have questions as we go during implementation but looks good to me!

PR #407, opened 2022-03-11
Add costing model for WT3 unit models

luohezhiming:
## Summary/Motivation:

Add costing models for some WT3 unit models

- add the costing model for buffer_tank, catridfe_filtration, dissolved_air_flotation, dual_media_filtration, feed_water_tank, media filtration, microscreen filtration, and ultra_filtration models
- change their yaml file to be consistent with convention

### Legal Acknowledgement


adam-a-a (dismissed):
LGTM

watertap/data/techno_economic/ultra_filtration.yaml
```diff
@@ -2,25 +2,18 @@ default:
...
-    flow_basis_par:
+  capital_cost:
+    basis: flow_vol
+    cost_factor: None
+    reference_state:
       value: 4731.764235
       units: m^3/hr
-    cap_basis_par:
+    capital_a_parameter:
       value: 75.0
```
bknueven, line 10: Does this need to be multiplied by 1e6?

luohezhiming: Yeah, it should be. Thank you for pointing out this error.

luohezhiming:
```suggestion
      value: 75.0e6
```

watertap/data/techno_economic/media_filtration.yaml
```diff
@@ -2,25 +2,18 @@ default:
   energy_electric_flow_vol_inlet:
     value: 0.00015 #todo: revisit this value later on
     units: kWh/m^3
-  specific_capital_cost_flow_vol_inlet:
-    flow_basis_par:
+  capital_cost:
```
andrewlee94, line 5: This model has a different form in the old WT3 files, so it will need custom code (unless something has changed).

That said, we should check with @kurbansitterley where these values came from seeing as they were already here.

andrewlee94 (changes requested):
Only one comment from me - I checked all these models against the old WT3 files and one of them has a non-standard form which we should clarify with the WT3 team.

Otherwise, looks good.

watertap/data/techno_economic/media_filtration.yaml
luohezhiming: OK. The previous yaml file does not make sense, I will build the custom costing model for this media filtration unit.

kurbansitterley: I am not sure what differences you are seeing. The media_filtration.py hasn't been changed in a long time. However, there is also a media_filtration basic_unit that has capital costing in the C = A * Q/Qbasis ** b form. That form could also be used

luohezhiming: @kurbansitterley I am really confused here, what is the media filtration unit in the flowsheet, I thought it would be the WT3 media_filtration.py. Is this file deprecated? I cannot find two media_filtration.yaml

andrewlee94: @kurbansitterley The issue is that we don't know which of the two version we should be using  -the basic unit version or the one in WT3's `media_filtration.py`.

kurbansitterley: They are both used throughout the case studies with very similar results. I don't know why that was done. I don't think there is any use in porting both of them. I linked to the `media_filtration.py` WT3 model in our tracking spreadsheet and marked it as not a basic unit because in general I think it is better to have cost based on some physical/operational parameter (in this case e.g. loading rate and filter surface area) rather than just flow (as is the case with the basic_units). 

Using the `media_filtration.py` version rather than the basic_unit version will result in slightly different TEA results, but as we've discussed, we are likely going to have different TEA results in the watertap version vs. those from WT3 for many reasons so I don't think that is a reason to bring one over the other.

Apologies for the confusion around this.

luohezhiming: OK. At this time point, I think directly treating the media_filtration unit as a standard costing model is reasonable. I will restore the original version.

watertap/unit_models/zero_order/tests/test_buffer_tank_zo.py
```diff
@@ -125,4 +127,46 @@ def test_report(self, model):
...
+
+    assert isinstance(m.fs.unit1.costing.capital_cost, Var)
+    assert isinstance(m.fs.unit1.costing.capital_cost_constraint,
+                      Constraint)
+
+    assert_units_consistent(m.fs)
+    assert degrees_of_freedom(m.fs.unit1) == 0
+
+    assert m.fs.unit1.electricity[0] in \
+        m.fs.costing._registered_flows["electricity"]
```
kurbansitterley, line 172: @andrewlee94 do we want to be testing the solution for the costing tests as we do in the `test_solution()` method for the model build?

andrewlee94: We should be, but for now we haven't been doing that, in part due to lack of verification results on hand. I.e. we really need to have some external result to compare against (e.g. the case studies), as otherwise we are jut assuming the value we got the first time was right,

kurbansitterley (approved):
LGTM

andrewlee94 (approved):

PR #408, opened 2022-03-14
Add 0-D crystallizer model

OOAmusat:
## Summary/Motivation:

First draft of 0-D crystallizer model

- Add crystallizer model
-

### Legal Acknowledgement


watertap/unit_models/crystallizer.py
```diff
@@ -0,0 +1,696 @@
...
+        units_meta = self.config.property_package.get_metadata().get_derived_units
+
+        # Add unit variables
+
+        self.hex_max_temp_increase = Param(
+            initialize=4,
+            units=pyunits.K,
+            doc="Maximum temperature difference between inlet and outlet of a crystallizer heat exchanger.\
+            Lewis et al. suggests 1-2 degC but use 5degC in example; Tavare example used 4 degC.\
+            Default is 4 degC")
```
TimBartholomew: What does "hex" mean in this variable name? I'd write out "temp" as "temperature" to be more consistent with IDAES.

watertap/unit_models/crystallizer.py
```diff
@@ -0,0 +1,696 @@
...
+            units=pyunits.m/pyunits.s,
+            doc="Crystal growth rate, m/s"
+            )
+
+
+        self.k_param = Param(
+            initialize=0.04,
+            units=pyunits.m/pyunits.s,
+            doc="Sounders-Brown constant, set at 0.04 m/s based on Dutta et al. \
+            Lewis et al suggests 0.024 m/s, while Tavare suggests about 0.06 m/s ")
```
TimBartholomew: I'm not sure, but there are probably better descriptive names for the variables.

I realized you made them all Params, that's fine, but typically we make most unit model parameters except constants be variables. If they are variables, we can do parameter estimation problems with them. Additionally, users will have to explicitly fix them, which is what we usually want.

The biggest deciding factor on whether this should be a variable or parameter is do you envision users needing to be explicitly aware of their values and that they would change them? For example, in RO, the A and B parameters for the membrane properties are variables because users should select their values, where the solvent density is a parameter because we don't expect users to interact with it.

watertap/unit_models/crystallizer.py
```diff
@@ -0,0 +1,696 @@
...
+
+        self.work_mechanical = Var(
+            self.flowsheet().config.time,
+            initialize=1e5,
+            bounds=(-5e6, 5e6),
+            units=pyunits.kJ/pyunits.s,
+            doc="Crystallizer thermal energy requirement"
+            )
+
+        self.D_crystallizer = Var(
```
TimBartholomew: I'd write this out as `diameter` and drop the crystallizer part. Typically, we write out the full name of dimensions, like `length` and `width` for the RO model.

watertap/unit_models/crystallizer.py
```diff
@@ -0,0 +1,696 @@
...
+            initialize=3,
+            bounds=(0, 25),
+            units=pyunits.m,
+            doc="Minimum diameter of crystallizer"
```
TimBartholomew: This is just the diameter of the crystallizer, right? Not the minimum.

watertap/unit_models/crystallizer.py
```diff
@@ -0,0 +1,696 @@
...
+            units=pyunits.m,
+            doc="Slurry height in crystallizer"
+            )
+
+        self.H_crystallizer = Var(
+            initialize=3,
+            bounds=(0, 25),
+            units=pyunits.m,
+            doc="Crystallizer height"
+            )
```
TimBartholomew: Same as above, I'd write this out as `height`. I now see why you added the crystallizer qualifier. You can keep it for the height and diameter if you like.

watertap/unit_models/crystallizer.py
```diff
@@ -0,0 +1,696 @@
...
+            units=pyunits.hr,
+            doc="Residence time in crystallizer"
+            )
+
+        self.V_suspension = Var(
+            initialize=0,
+            bounds=(0, None),
+            units=pyunits.m**3,
+            doc="Crystallizer minimum active volume, i.e. volume of liquid-solid suspension"
+            )
```
TimBartholomew: I'd also write this out as `volume`, is this also not the crystallizer volume? Do you have a variable for the crystallizer volume?

watertap/unit_models/crystallizer.py
```diff
@@ -0,0 +1,696 @@
...
+                    == b.properties_out[0].flow_mass_phase_comp['Liq', j] + b.properties_solids[0].flow_mass_phase_comp['Sol', j])
+            else:
+                return Constraint.Skip
+
+        ## (ii) Solvent mass balances
+        @self.Constraint(doc="Mass balance for solvent")
+        def eq_mass_balance_solvent(b):
+                return (b.properties_in[0].flow_mass_phase_comp['Liq', 'H2O'] 
+                    == b.properties_out[0].flow_mass_phase_comp['Liq', 'H2O']
+                    + b.properties_vapor[0].flow_mass_phase_comp['Vap', 'H2O'])
```
TimBartholomew: I think you can just do the species mass balance and not differentiate between the solute and solvents, and include all of the relevant variables:
A constraint indexed by component could:
```
return sum(b.properties_in[0].flow_mass_phase_comp[p, j] for all p) ==
           sum(b.properties_out[0].flow_mass_phase_comp[p, j] for all p)
           sum(b.properties_vapor[0].flow_mass_phase_comp[p, j] for all p)
           sum(b.properties_solids[0].flow_mass_phase_comp[p, j] for all p)
```
This won't work directly because it doesn't use phase_component_set for the p summation, and I'm not 100% sure how to do that at the moment, but there is probably a way to pull all the relevant phases and sum across them.

This type of formulation would force you to fix the unneeded variables, but I saw that you do that anyway below.

watertap/unit_models/crystallizer.py
adam-a-a: We _could_ set these as variables and then fix them here in the unit model (similar to what we do in property models). That way, if we wanted to use parameter estimation later on, it wouldn't be an issue.

watertap/unit_models/crystallizer.py
```diff
@@ -0,0 +1,696 @@
...
+                + b.properties_in[0].flow_mass_phase_comp['Liq', 'H2O']
+                - b.properties_vapor[0].flow_mass_phase_comp['Vap', 'H2O']
+                )
+            # return (b.relative_supersaturation[j] * b.properties_out[0].solubility_mass_frac_phase_comp['Liq', j] == 
+            # (mass_frac_after_evap - b.properties_out[0].solubility_mass_frac_phase_comp['Liq', j])
+            # )
+            return (b.relative_supersaturation[j] == 
+                (mass_frac_after_evap - b.properties_out[0].solubility_mass_frac_phase_comp['Liq', j])
+                / b.properties_out[0].solubility_mass_frac_phase_comp['Liq', j]
+                )
```
TimBartholomew: I think there is an issue here. `j` is fed into the function, but this is changed in the sum with `for j in solute_set`.

watertap/unit_models/crystallizer.py
```diff
@@ -0,0 +1,696 @@
...
+                return Constraint.Skip
+
+        # (iii) Fix vapor: solid and vapour flows must be zero. Vapour is also pure solvent.
+        @self.Constraint(self.properties_vapor[0].phase_component_set,
+            doc="Empty components in vapour stream")
+        def eq_vapor_composition(b, p, j):
+            if p != 'Vap':
+                return (b.properties_vapor[0].flow_mass_phase_comp[p, j] == 1e-8 * pyunits.kg/pyunits.s)
+            else:
+                return Constraint.Skip
```
TimBartholomew: Instead of setting the variables equal to a constant via a constraint, it would be better to just fix the variables.

watertap/unit_models/crystallizer.py
```diff
@@ -0,0 +1,696 @@
...
+
+
+        # 5. Add an energy balance for the system
+        ## (iii) Enthalpy balance: based on Lewis et al. Enthalpy is exothermic, and the hC in property package is -ve
+        @self.Constraint(doc="Enthalpy balance over crystallization system")
+        def eq_enthalpy_balance(b):
+            return (b.properties_in[0].enth_flow 
+                - b.properties_out[0].enth_flow - b.properties_vapor[0].enth_flow - b.properties_solids[0].enth_flow
+                + self.work_mechanical[0]
+                - (b.properties_solids[0].flow_vol_phase['Sol'] * b.properties_solids[0].dens_mass_solute['Sol'] * b.properties_solids[0].dh_crystallization['NaCl'])
```
TimBartholomew: Can't you grab the mass flowrate directly `b.properties_solids[0].flow_mass_phase_comp['Sol', j]? You could also sum across the solutes, whenever we do multiple solutes.

watertap/unit_models/crystallizer.py
```diff
@@ -0,0 +1,696 @@
...
+        def eq_T_con1(b):
+            return self.temperature_operating - b.properties_solids[0].temperature == 0
+
+        @self.Constraint()    
+        def eq_T_con2(b):
+            return self.temperature_operating - b.properties_vapor[0].temperature == 0
+
+        @self.Constraint()
+        def eq_T_con3(b):
+            return self.temperature_operating - b.properties_out[0].temperature == 0
```
TimBartholomew: It's not a big deal, but typically IDAES and WaterTAP style is to just write the equality with a LHS and RHS, rather than subtracting one over and having it equal to 0.

watertap/unit_models/crystallizer.py
```diff
@@ -0,0 +1,696 @@
...
+
+
+        # 7. Heat exchanger minimum circulation flow rate calculations - see Lewis et al. or Tavare et al.
+        @self.Constraint(doc="Constraint on mimimum circulation rate through crystallizer heat exchanger")
+        def eq_minimum_hex_circulation_rate_constraint(b):
+            rho_cp_dt = self.hex_max_temp_increase * \
```
TimBartholomew: I'd write out `rho` as `dens`, although I'm not 100% understanding this variable. It seems to be the average of the product of density and cp. I think `dens_cp_avg` could work. It's also minor because this is just a temporary expression, but might as well be consistent.

watertap/unit_models/crystallizer.py
```diff
@@ -0,0 +1,696 @@
...
+            units=pyunits.kg/pyunits.m**3,
+            doc="Magma density, i.e. mass of crystals per unit volume of suspension"
+            )
+
+        self.slurry_density = Var(
+            initialize=1000,
+            bounds=(1, 5000),
+            units=pyunits.kg/pyunits.m**3,
+            doc="Suspension density, i.e. density of solid-liquid mixture before separation"
+            )
```
TimBartholomew: `dens` is the IDAES convention for density, I'd go with `dens_mass_magma` and `dens_mass_slurry`

watertap/unit_models/crystallizer.py
```diff
@@ -0,0 +1,696 @@
...
+                b.product_volumetric_solids_fraction * b.properties_solids[0].dens_mass_solute['Sol'] 
+                + (1 - b.product_volumetric_solids_fraction) * b.properties_out[0].dens_mass_phase['Liq']
+                )
+
+
+        # 9. Residence time calculation
+        @self.Constraint(doc="Residence time")
+        def eq_residence_time(b):
+            return (
+                b.t_res == b.L_d / (3.67 * pyunits.convert(b.G_rate, to_units=pyunits.m/pyunits.hr))
```
TimBartholomew: What is the 3.67? We should probably make that a parameter or variable for the unit model.

watertap/unit_models/crystallizer.py
```diff
@@ -0,0 +1,696 @@
...
+                b.V_suspension == (b.properties_solids[0].flow_vol + b.properties_out[0].flow_vol) * pyunits.convert(b.t_res, to_units=pyunits.s)
+                )
+
+
+        # 11. Minimum diameter of evaporation zone
+        def eq_max_allowable_velocity(b):
+            return b.k_param * \
+            (b.properties_out[0].dens_mass_phase['Liq']/ b.properties_vapor[0].dens_mass_solvent['Vap']) ** 0.5
+        
+        self.v_max = Expression(rule=eq_max_allowable_velocity, doc='maximum allowable vapour linear velocity in m/s')
```
TimBartholomew: You can use the same decorator syntax you've been using for constraints for an expression.

TimBartholomew (commented):
Looks pretty good. Mostly comments on naming and consistency, and a few on model formulation (i.e. params vs variables, grouping of constraints)

watertap/unit_models/crystallizer.py
OOAmusat: This is not the crystallizer volume. It is the volume of the solid/liquid mixture in the crystallizer. I have no variable for the crystallizer volume at the moment.

watertap/unit_models/crystallizer.py
OOAmusat: hex - heat exchanger.

temp replaced by temperature

watertap/unit_models/crystallizer.py
OOAmusat: Yes. Corrected.

watertap/unit_models/crystallizer.py
OOAmusat: Done

watertap/unit_models/crystallizer.py
OOAmusat: Done.

watertap/unit_models/crystallizer.py
OOAmusat: Yes, it is a weighted average of the sum of the product of density and cp for the solid/liquid mixture. Name changed.

watertap/unit_models/crystallizer.py
OOAmusat: Changed.

watertap/unit_models/crystallizer.py
OOAmusat: It is not a variable. It is a constant in the equation relating residence time, crystal growth rate and crystal length:

crystal length L = 3.67 * Growth rate (m/s)  * residence time (s)

watertap/unit_models/crystallizer.py
OOAmusat: Done.

watertap/unit_models/crystallizer.py
OOAmusat: Good point. Changed just in case.

watertap/unit_models/crystallizer.py
OOAmusat: Done.

watertap/unit_models/crystallizer.py
OOAmusat: @TimBartholomew  So just to be clear, you'd prefer this was written as:

```
        for p, j in self.properties_solids[0].phase_component_set:
            if p != 'Sol':
                self.properties_solids[0].flow_mass_phase_comp[p, j].fix(1e-8)

        for p, j in self.properties_out[0].phase_component_set:
            if p != 'Liq':
                self.properties_out[0].flow_mass_phase_comp[p, j].fix(1e-8)

        for p, j in self.properties_vapor[0].phase_component_set:         
            if p != 'Vap':
                self.properties_vapor[0].flow_mass_phase_comp[p, j].fix(1e-8)
```

OOAmusat: Done.

watertap/property_models/cryst_prop_pack.py
```diff
@@ -1141,13 +1141,15 @@ def calculate_scaling_factors(self):
...
+                        default=1, warning=True))
 
         if self.is_property_constructed('cp_solvent'):
-            for p in ['Vap', 'Liq']:
-                if iscale.get_scaling_factor(self.cp_solvent) is None:
-                    iscale.set_scaling_factor(self.cp_solvent[p], iscale.get_scaling_factor(self.cp_phase['Liq']))
+            for p in ['Liq', 'Vap']:
+                if iscale.get_scaling_factor(self.cp_solvent[p]) is None:
+                    iscale.set_scaling_factor(self.cp_solvent[p], iscale.get_scaling_factor(self.cp_phase['Liq'], 
+                        default=1, warning=True))
```
adam-a-a: Revisit why we need to do this in a subsequent PR (i.e., resolve the issue that requires us to do this)

watertap/unit_models/crystallizer.py
OOAmusat: Converted to variables and naming improved.

OOAmusat: @adam-a-a  Done

watertap/unit_models/crystallizer.py
OOAmusat: 3.67 moved ito a parameter called ``shape_factor_constant``

watertap/unit_models/crystallizer.py
OOAmusat: Variable renamed as ``approach_temperature_heat_exchanger`` for clarity.

watertap/unit_models/crystallizer.py
OOAmusat: Fixed.

@TimBartholomew I have reformulated this as:

```
      @self.Constraint(self.config.property_package.component_list, doc="Mass balance for components")
      def eq_mass_balance_constraints(b, j):
          return (
              sum(b.properties_in[0].flow_mass_phase_comp[p, j] for p in self.config.property_package.phase_list if (p,j) in b.properties_in[0].phase_component_set) == \
              sum(b.properties_out[0].flow_mass_phase_comp[p, j] for p in self.config.property_package.phase_list if (p,j) in b.properties_out[0].phase_component_set)
              + sum(b.properties_vapor[0].flow_mass_phase_comp[p, j] for p in self.config.property_package.phase_list if (p,j) in b.properties_vapor[0].phase_component_set)
              + sum(b.properties_solids[0].flow_mass_phase_comp[p, j] for p in self.config.property_package.phase_list if (p,j) in b.properties_solids[0].phase_component_set) 
         )
```

watertap/property_models/cryst_prop_pack.py
OOAmusat: Addressed. Supplied the scaling factor directly rather than calling get_scaling_factor

OOAmusat:
@TimBartholomew @adam-a-a The tests are done and the scaling errors have been addressed, so this is ready for review. I already made all the changes you suggested in your comments.

watertap/property_models/cryst_prop_pack.py
```diff
@@ -1114,6 +1115,15 @@ def calculate_scaling_factors(self):
             if iscale.get_scaling_factor(self.solubility_mass_frac_phase_comp) is None:
                 iscale.set_scaling_factor(self.solubility_mass_frac_phase_comp, 1e0)
 
+        # if self.is_property_constructed('dens_mass_solvent')
+        # for p in ['Liq', 'Vap']:
+        #     if p == 'Liq':
+        #         if iscale.get_scaling_factor(self.dens_mass_solvent[p]) is None:
+        #             iscale.set_scaling_factor(self.dens_mass_solvent[p], 1e-3)
+        #     elif p == 'Vap':
```
adam-a-a: Remove commented code

OOAmusat: Done

TimBartholomew (approved):
LGTM

adam-a-a (approved):
LGTM

PR #412, opened 2022-03-14
Migrating WT3 energy recovery model

luohezhiming:
## Summary/Motivation:

Add energy recovery model

- add energy recovery model originally developed by @MarkusHolly
- remove the bounds for the electricity variable in the zero_order_electricity.py
- add costing model
- add corresponding tests

### Legal Acknowledgement


watertap/data/techno_economic/energy_recovery.yaml
```diff
@@ -2,29 +2,15 @@ default:
...
+    reference_state:
       value: 3.943
       units: m^3/hr
-    cap_basis_par:
-      value: 0.1
-      units: M$
-    cap_exp_par:
+    capital_a_parameter:
+      value: 1e5
+      units: USD_2020
```
andrewlee94, line 13: Check the year for this - below it says 2020.

watertap/unit_models/zero_order/energy_recovery_zo.py
```diff
@@ -0,0 +1,40 @@
...
+    CONFIG = ZeroOrderBaseData.CONFIG()
+
+    def build(self):
+        super().build()
+
+        self._tech_type = "energy_recovery"
+
+        build_pt(self)
+        constant_intensity(self)
+        self.electricity.setlb(None)
```
andrewlee94: This line is not needed. The bound on electricity is now set in the `constant_intensity` method,

andrewlee94 (changes requested):
Two minor comments for correction.

watertap/data/techno_economic/energy_recovery.yaml
luohezhiming: I set the unit as USD_2020 as indicated in the original yaml file. What is the problem here?

andrewlee94: I have no idea a this point. For some reason when I first read this I thought the year here was different.

watertap/unit_models/zero_order/energy_recovery_zo.py
luohezhiming: The bounds on `electricity` in the `constant_intensity` are set as` (0,None)`. However, the `electricity` in this unit is negative, I have to reset the lower bound to be `None`, otherwise the unit cannot be solved.

andrewlee94: This is going to create some other challenges then, as the costing framework assumes all flows are positive with a lower bound of 0 (and will raise a warning if this is not the case, which we don't want the users seeing).

andrewlee94: The first thing you need to do is also set an upper bound of 0 as well, which might be sufficient to resolve this - either way the flow needs to be bounded to 0 at one end.

luohezhiming: I've set the upper bound to be 0 for this case, I think it is reasonable at this point, the costing test passed.

andrewlee94 (dismissed):

adam-a-a (approved):
LGTM

kurbansitterley (approved):
LGTM

PR #413, opened 2022-03-15
API for the WaterTAP UI

dangunter:
## Fixes/Addresses:

Issue #488 

## Summary/Motivation:

This PR provides an API framework for achieving this that will then be used for specific treatment train and unit models.

- One module with the API
- Utility model for misc. things
- Unit tests

### Legal Acknowledgement


dangunter:
@TimBartholomew and @adam-a-a  this is mostly to reduce the size of the eventual PR , so a review with only nitpicks or even a rubber stamp would be fine. The API internals will no doubt get changed as I work through a real example. To see the external interface, look under `Technical Reference` -> `WaterTAP UI` in the generated docs.

dangunter:
OK, this is passing except the RTD build, which is irrelevant since they are only warnings, and the codecov test which is irrelevant since it's codecov.

dangunter:
@TimBartholomew et al., I'm just going to merge this

aladshaw3:
@dangunter This was not passing the doc tests. How and why was this merged without all checks or approvals?

PR #414, opened 2022-03-15
WT3 UV/UVAOP & Ozone/OzoneAOP Costing

adam-a-a:
## Summary/Motivation:
This PR brings in costing methods for uv/uv-aop and ozonation/ozone-aop. UV-AOP is utilized in the Santa Barbara case study.

Some differences were identified for calculating oxidant dose in ozone/aop and implemented; these changes should be reviewed since the calculated oxidant dose is significantly higher than the previously assumed value originally provided in the yaml file. Additionally, new fitted parameters were implemented for UV capex using ALAMO.

Thanks to @andrewlee94 for putting together the capital cost relationship for ozonation and assisting with setting this up.

- add defined flow for hydrogen peroxide in default_case_study.yaml for costing chemical flow
- add costing parameters and methods for uv (new fitted parameters obtained using ALAMO) and aop capex
- add costing parameters and method for ozonation capex from @andrewlee94 's set of fitted parameters
- add costing tests for uv, uv-aop, ozonation, ozone-aop

### Legal Acknowledgement


watertap/data/techno_economic/ozone_aop.yaml
```diff
@@ -44,12 +67,32 @@ default:
     value: 5      
     units: kWh/lb
     reference: B. Mundy et al. (2018) doi.org/10.1080/01919512.2018.1467187 - Energy consumption 
-      for ozone generator, ozone destruct, nitrogen feed system, cooling water pumps, and other misc. energy uses. 
-  oxidant_dose:
-    value: 5
-    units: mg/L
+      for ozone generator, ozone destruct, nitrogen feed system, cooling water pumps, and other misc. energy uses.
 
 hydrogen_peroxide:
```
andrewlee94, line 72: Do we need to second sub-type here? And if so, does this mean  we should use "oxidant_to_ozone_ratio" instead of explicitly stating H2O2 in the parameter name below?

watertap/data/techno_economic/uv_aop.yaml
```diff
@@ -41,6 +62,26 @@ default: # default is for hydrogen peroxide
     units: kWh/m^3
 
 hydrogen_peroxide:
```
andrewlee94, line 64: Again, do we need the sub-type as H2O2 is the default assumption?

watertap/unit_models/zero_order/ozone_aop_zo.py
```diff
@@ -29,31 +29,58 @@ class OzoneAOPZOData(OzoneZOData):
...
 
-    CONFIG = OzoneZOData.CONFIG()
-
     def build(self):
         super().build()
 
         self._tech_type = "ozone_aop"
 
-
         self.oxidant_dose = Var(self.flowsheet().time,
```
andrewlee94, line 37: In other models, I think we have kept the dosage in mg/L and just done unit conversions as necessary.

watertap/unit_models/zero_order/ozone_aop_zo.py
```diff
@@ -29,31 +29,58 @@ class OzoneAOPZOData(OzoneZOData):
...
+                                                 units=pyunits.dimensionless,
+                                                 doc="Ratio of hydrogen peroxide to ozone")
+
+        self._fixed_perf_vars.append(self.hydrogen_peroxide_ozone_ratio)
+
+        @self.Constraint(self.flowsheet().time,
+                         doc="Ozone/TOC ratio constraint")
+        def ozone_toc_ratio_constraint(b, t):
+            return (b.ozone_toc_ratio[t] == 1
+                    + b.concentration_time[t] / b.contact_time[t]     #TODO: think this should be dividing by contact time, not multiplying
```
andrewlee94: This should have been detected by any unit consistency checks in the tests. Double check to make sure those checks are present,

watertap/unit_models/zero_order/ozone_aop_zo.py
```diff
@@ -29,31 +29,58 @@ class OzoneAOPZOData(OzoneZOData):
...
+                    / pyunits.convert(b.properties_in[t].conc_mass_comp['toc'],
```
andrewlee94: I would also put the entire expression inside this `convert` statement:

```
pyunits.convert(conc_time/contact_time/conc_mass_comp, to_units=dimensionless)
```

This will also help make sure the units in this expression are right,

andrewlee94 (commented):
A few comments to think about and to try to make sure the units are consistent,

watertap/data/techno_economic/uv_aop.yaml
adam-a-a: Although they are basically duplicated at this point because I didn't have a better placeholder for the default, I wanted the user to be able to explicitly specify hydrogen peroxide for the aop process; future alternatives could be ozone and chlorine.

watertap/data/techno_economic/ozone_aop.yaml
adam-a-a: Maybe for ozonation/aop, we only need hydrogen peroxide although I'm unaware of potential alternatives.

watertap/unit_models/zero_order/ozone_aop_zo.py
adam-a-a: Well, the origin of what I thought was an error came from WT3 directly. I know that Kurby had revised this here (e.g., concentration_time, i.e., CT, was originally implemented as concentration divided by time but was corrected to be concentration*time). I'm pretty sure we have an assert_units_consistent in the tests.

watertap/data/techno_economic/ozone_aop.yaml
andrewlee94: If we feel the need to have H2O2 as a sub-type however, that implies we see a need for different oxidants. I think we either need to have only the default sub-type (i.e. assume it is always H2O2), or make the name of the ratio more general as it implies we are allowing for other oxidants.

adam-a-a: OK, I am going to going with renaming the ratio then.

watertap/unit_models/zero_order/ozone_aop_zo.py
```diff
@@ -48,28 +48,28 @@ def build(self):
                                    units=pyunits.dimensionless,
                                    doc="Ratio of ozone to total organic carbon")
 
-        self.hydrogen_peroxide_ozone_ratio = Var(self.flowsheet().time,
+        self.oxidant_ozone_ratio = Var(self.flowsheet().time,
                                                  units=pyunits.dimensionless,
                                                  doc="Ratio of hydrogen peroxide to ozone")
```
andrewlee94: Minor thing, but this comment needs to be updated too.

watertap/unit_models/zero_order/ozone_aop_zo.py
```diff
@@ -48,28 +48,28 @@ def build(self):
...
+        self._fixed_perf_vars.append(self.oxidant_ozone_ratio)
 
         @self.Constraint(self.flowsheet().time,
                          doc="Ozone/TOC ratio constraint")
         def ozone_toc_ratio_constraint(b, t):
             return (b.ozone_toc_ratio[t] == 1
-                    + b.concentration_time[t] / b.contact_time[t]     #TODO: think this should be dividing by contact time, not multiplying
-                    / pyunits.convert(b.properties_in[t].conc_mass_comp['toc'],
-                                      to_units=pyunits.mg/pyunits.liter))
+                    + pyunits.convert(b.concentration_time[t] / b.contact_time[t]     #TODO: think this should be dividing by contact time, not multiplying
```
andrewlee94: I think we can drop this comment,

watertap/data/techno_economic/uv.yaml
```diff
@@ -1,4 +1,18 @@
...
+      units: USD_2014
+    uv_capital_b_parameter:
+      value: 12268.649303773354830582
+      units: USD_2014
+    uv_capital_c_parameter:
+      value: 0.19242621822957126230235e-3
+      units: USD_2014
+    uv_capital_d_parameter:
+      value: -12779.106944244096766283
+      units: USD_2014
```
luohezhiming, line 15: Where do you get these regression parameter values? Is there a file?

andrewlee94 (dismissed):

luohezhiming (dismissed):

watertap/data/techno_economic/uv.yaml
adam-a-a: Oh, I fit these myself to avoid using the uv_regress function from WT3.
"Cost parameters retrieved by fitting TWDB data in ALAMO (MADp modeler, max terms=4)"

kurbansitterley (approved):
LGTM, hopefully the capex never goes negative!

andrewlee94 (approved):

PR #417, opened 2022-03-15
Wt3 well field

kurbansitterley:
## Summary/Motivation:

Migrate WT3 well field unit process to watertap, create tests, yaml, and add costing. 

- add well_field as passthtough unit
- add/create tests, .yaml, and costing
- default subtype includes no pipe cost, but subtype ``emwd`` (for the case study EMWD) does
- well_field costing includes the capex for the well field itself and the piping cost to the treatment train (default pipe distance is 0 miles) 
- add ``cost_well_field()`` function in ``zero_order_costing.py`` to add pipe cost since capex form is ``C = A * Q ** b + pipe_cost`` 

### Legal Acknowledgement


watertap/data/techno_economic/well_field.yaml
```diff
@@ -3,8 +3,62 @@ default:
     value: 0.9999
     units: dimensionless
     reference: Voutchkov, N. (2018). Desalination Project Cost Estimating and Management.
-      Taylor & Francis, a CRC title, part of the Taylor & Francis imprint, a member
-      of the Taylor & Francis Group, the academic division of T&F Informa, plc.
   default_removal_frac_mass_solute:
     value: 0
     units: dimensionless
+  capital_cost:
+    basis: flow_vol
```
andrewlee94: Minor note, but `basis` is not required in most cases and I have been leaving it out.

watertap/unit_models/zero_order/well_field_zo.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+
+    CONFIG = ZeroOrderBaseData.CONFIG()
+
+    def build(self):
+        super().build()
+
+        self._tech_type = "well_field"
+        build_pt(self)
+        self._Q = Reference(self.properties[:].flow_vol)
```
andrewlee94, line 40: Do we really need this `Reference`?

watertap/core/zero_order_costing.py
```diff
@@ -440,6 +440,71 @@ def cost_chlorination(blk):
...
+            initialize=1,
+            units=blk.config.flowsheet_costing_block.base_currency,
+            bounds=(0, None),
+            doc="Capital cost of unit operation")
+        blk.pipe_cost = pyo.Var(
+            initialize=100,
+            units=pyo.units.USD_2018,
+            bounds=(0, None),
+            doc="Piping cost for well field")
+        blk.pipe_cost_basis = pyo.Var(
```
andrewlee94: This is not needed - the costing parameter Blocks already contains a `Var` for this so you should just use that directly.

watertap/core/zero_order_costing.py
```diff
@@ -440,6 +440,71 @@ def cost_chlorination(blk):
...
+            initialize=100,
+            units=pyo.units.USD_2018/(pyo.units.miles*pyo.units.inch),
+            bounds=(0, None),
+            doc="Piping cost for well field")
+        blk.pipe_cost_basis.fix(pipe_cost_basis)
+        blk.pipe_cost_constr = pyo.Constraint(
```
andrewlee94: Why not just build this directly into the capital cost expression below?

watertap/data/techno_economic/well_field.yaml
```diff
@@ -3,8 +3,62 @@ default:
...
+    cost_factor: None
+    reference_state:
+      value: 1
+      units: m^3/hr
+    capital_a_parameter:
+      value: 4731.6
```
andrewlee94, line 15: I think this is correct, but just to confirm this factor is in $ and not M$?

andrewlee94 (changes requested):
You seem to be duplicating a number of components that aren't really required, rather than just using the existing components directly.

Also, is there any need to create a separate Var for the pipe cost, over just rolling it into the capital cost?

watertap/unit_models/zero_order/well_field_zo.py
kurbansitterley: I am not sure if we need the `Reference` specifically but a flow is required input for `pump_electricity()`. Doing `pump_electricity(self, self.properties[:].flow_vol)` throws an error. I got this implementation from @adam-a-a 's `municipal_drinking` unit.

andrewlee94: What error do you get?

watertap/data/techno_economic/well_field.yaml
kurbansitterley: Confirmed.

watertap/unit_models/zero_order/well_field_zo.py
andrewlee94: Oh, I see the problem. Yes, you will need the `Reference` then.

watertap/core/zero_order_costing.py
kurbansitterley: My thought was if there was interest in knowing the piping cost separate from the well field cost. But I will move it.

watertap/core/zero_order_costing.py
kurbansitterley: I saw that but I also saw it was in the `cost_chlorination()` method so I included it. I will remove it.

watertap/data/techno_economic/well_field.yaml
kurbansitterley: Is it only needed if the basis is mass?

andrewlee94: No, it is only needed if you are using the general power law costing method. That method explicitly looks for the basis as it needs to know whether to use `flow_vol` or `flow_mass`.

watertap/core/zero_order_costing.py
kurbansitterley: @andrewlee94 Actually I am not seeing how `blk.capital_cost = Var()` can be excluded from `cost_well_field()`. Just like in `cost_chlorination()` the capex formula is a different form than that in `_general_power_law_form()` so that method is never called and the capital_cost variable is never created. Unless I am missing something? I get an error when it is excluded.

andrewlee94: This is referring to `blk.pipe_cost_basis`, not `blk.captial_cost`. The capital cost Var needs to exist, but `pipe_cost_basis` is just duplicating a `Var` that exists elsewhere (plus I am not sure that your line a little further on fixing this Var is correct either).

kurbansitterley: Ah ok ok I see. You are correct. Didn't realize `_get_tech_parameters()` returned them as `Var()`s

andrewlee94 (dismissed):

bknueven (dismissed):

luohezhiming (dismissed):

adam-a-a (approved):

luohezhiming (dismissed):

adam-a-a (dismissed):

luohezhiming (approved):

adam-a-a (approved):

PR #415, opened 2022-03-15
Cleaning up ZO brine concentrator and adding costing

andrewlee94:
## Part of WT3 merger

Adds costing for zero-order brine concentrator models, as well as some general clean up.

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #418, opened 2022-03-15
Migrating WT3 mbr model

luohezhiming:
## Summary/Motivation:

Add membrane bioreactor model

- add mbr unit model originally developed by @MarkusHolly
- add costing model
- add corresponding tests

### Legal Acknowledgement


watertap/unit_models/zero_order/tests/test_mbr_zo.py
```diff
@@ -0,0 +1,504 @@
...
+
+        for (t, j), v in model.fs.unit.removal_frac_mass_solute.items():
+            assert v.fixed
+            assert v.value == data["removal_frac_mass_solute"][j]["value"]
+
+        assert model.fs.unit.energy_electric_flow_vol_inlet.fixed
+        assert model.fs.unit.energy_electric_flow_vol_inlet.value == data[
+            "energy_electric_flow_vol_inlet"]["value"]
+
+def test_costing():
```
andrewlee94: We should parameterize this test to check each process subtype. Take a look at the brine concentrator tests in  https://github.com/watertap-org/watertap/pull/415 for an example.

andrewlee94 (changes requested):
One minor comment on testing all process subtypes, plus one generally question for the wider group: should we expand `mbr` to be more descriptive?

andrewlee94:
R.e. the failing tests - it looks like the general tests for the YAML files were set to expect electricity intensity to be between 0 and 1, which is probably a copy-and-paste error (this limitation makes sense for removal and recovery fractions, but not electricity intensity). I think the checks on the value of electricity intensity in the general tests should be removed.

watertap/unit_models/zero_order/tests/test_mbr_zo.py
luohezhiming: I've parameterized this test with respect to subtype as suggested.

luohezhiming:
> R.e. the failing tests - it looks like the general tests for the YAML files were set to expect electricity intensity to be between 0 and 1, which is probably a copy-and-paste error (this limitation makes sense for removal and recovery fractions, but not electricity intensity). I think the checks on the value of electricity intensity in the general tests should be removed.

I removed the upper bound (1) for the electricity intensity for now because it makes no sense. I am not so sure if I need to also delete the lower bound, but the test will pass right now.

andrewlee94 (approved):

kurbansitterley (approved):
LGTM

luohezhiming:
> One minor comment on testing all process subtypes, plus one generally question for the wider group: should we expand `mbr` to be more descriptive?

@kurbansitterley We explain in the document that `mbr` is a membrane bioreactor, should we consider  change the name of `mbr` to be more descriptive as `membrane_reactor`?

andrewlee94:
@luohezhiming Don't change it now. The tests are almost done and I am getting ready to merge this.

luohezhiming:
> @luohezhiming Don't change it now. The tests are almost done and I am getting ready to merge this.

OK

PR #419, opened 2022-03-15
Migrating WT3 municipal wwtp model

luohezhiming:
## Summary/Motivation:
 Add WT3 municipal wwtp model

- add municipal wwtp model originally developed by @MarkusHolly
- add costing model
- add corresponding tests

### Legal Acknowledgement

### DIFF
--- a/watertap/unit_models/zero_order/tests/test_municipal_wwtp_zo.py
+++ b/watertap/unit_models/zero_order/tests/test_municipal_wwtp_zo.py
@@ -55,7 +55,7 @@ class TestMunicipalWWTPZO:
 
     @pytest.mark.unit
     def test_build(self, model):
-        assert model.fs.unit.config.database == model.db
+        assert model.fs.unit.config.database is model.db
 
         assert isinstance(model.fs.unit.electricity, Var)
         assert isinstance(model.fs.unit.energy_electric_flow_vol_inlet, Var)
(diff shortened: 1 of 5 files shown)


watertap/unit_models/zero_order/tests/test_municipal_wwtp_zo.py
```diff
@@ -0,0 +1,179 @@
...
+            "database": m.db})
+
+        m.fs.unit.inlet.flow_mass_comp[0, "H2O"].fix(1000)
+        m.fs.unit.inlet.flow_mass_comp[0, "foo"].fix(1)
+
+        return m
+
+    @pytest.mark.unit
+    def test_build(self, model):
+        assert model.fs.unit.config.database == model.db
```
kurbansitterley:
```suggestion
        assert model.fs.unit.config.database is model.db
```

kurbansitterley (dismissed):
just changed the `==` to `is` in the test, LGTM

luohezhiming:
> just changed the `==` to `is` in the test, LGTM

@kurbansitterley I don't think we need to change all `==` to `is`, if the object for comparison is just some value, it is fine to use `==`. If the object is some other instance like `model.db`, then we'd better use `is`.

kurbansitterley:
> > just changed the `==` to `is` in the test, LGTM
> 
> @kurbansitterley I don't think we need to change all `==` to `is`, if the object for comparison is just some value, it is fine to use `==`. If the object is some other instance like `model.db`, then we'd better use `is`.

I agree, no need to change all `==`, just applying some guidance I got for this particular test. Sorry, wasn't suggesting that all `==` need to be changed to `is` only commenting on the suggested change I made.

luohezhiming:
> @pytest.mark.unit
>     def test_build(self, model):
>         assert model.fs.unit.config.database == model.db

OK. The change is already commited and resolved, I missed this, sorry about that.

bknueven (dismissed):

andrewlee94 (approved):

adam-a-a (approved):

bknueven (approved):

PR #424, opened 2022-03-15
WT3 pump-based use units

kurbansitterley:
## Summary/Motivation:

Determine if the WT3 pump-based use units can be `pump` subtypes. If not, then add them as separate unit processes.

These use units from the tracking sheet have identical TEA parameters to the `default` pump subtype. For this reason, I did not add them as pump subtypes:
- agriculture
- discharge
- industrial
- irrigation

TEA factors for these units warranted adding them as separate units:
- intrusion_mitigation
- injection_well_disposal

- add model files, tests, .yaml, and costing for intrusion_mitigation and injection_well_disposal as passthrough models with constant electricity intensity

### Legal Acknowledgement


watertap/data/techno_economic/injection_well_disposal.yaml
```diff
@@ -2,29 +2,21 @@ default:
...
+    cost_factor: None
+    reference_state:
       value: 473.176473
       units: m^3/hr
-    cap_basis_par:
+    capital_a_parameter:
       value: 22.5
-      units: M$
-    cap_exp_par:
+      units: MUSD_2011
```
luohezhiming: Is there any concern that we use MUSD as the currency unit here?

watertap/data/techno_economic/injection_well_disposal.yaml
```diff
@@ -2,29 +2,21 @@ default:
...
-    vol units:
-      value: m^3/hr
-    capital cost units:
-      value: M$
-    year:
-      value: 2011
-    cost_function_form:
-      function f(flow_vol_in): cap_basis_par*(flow_vol_in/flow_basis_par)^cap_exp
-      units: $
   recovery_frac_mass_H2O:
```
luohezhiming: Why does this pass-through unit has water recovery and removal? Just a little curious here because I delete these two attributes when I create a pass-through model and I should add the unit to the pass-through list in the `test_unit_parameter_files.py`. And I don't see any test with respect to the recovery and removal.

luohezhiming (commented):
LGTM. Just want to confirm if a pass-through need recovery and removal in the yaml file.

watertap/data/techno_economic/injection_well_disposal.yaml
andrewlee94: Whilst it doesn't matter in the end, it would probably be a good idea to be consistent with what we use here.

kurbansitterley: Purely convenience! Conversion to proper units is handled elsewhere I believe.

watertap/data/techno_economic/injection_well_disposal.yaml
kurbansitterley: Great point, I neglected to remove them when updating the .yaml. I will remove them and ensure tests are passing.

andrewlee94 (approved):

adam-a-a (approved):

luohezhiming (approved):

luohezhiming (approved):

PR #425, opened 2022-03-15
Migrating WT3 wind-aided intensified evaporation model

luohezhiming:
## Summary/Motivation:

Add wind-aided intensified evaporation model

- add waiv unit model originally developed by @MarkusHolly
- add costing model
- add corresponding tests

### Legal Acknowledgement


watertap/data/techno_economic/waiv.yaml
```diff
@@ -2,29 +2,27 @@ default:
...
-    year:
-      value: 2020
-    cost_function_form:
-      function f(flow_mass_in): cap_basis_par*(flow_mass_in/flow_basis_par)^cap_exp
-      units: $
   recovery_frac_mass_H2O:
-    value: 0.9999
+    value: 0.0001
     units: dimensionless
     reference: .nan
```
kurbansitterley:
```suggestion

```

watertap/data/techno_economic/component_list.yaml
```diff
@@ -118,4 +118,6 @@ viruses_enteric:
 zinc:
   constituent_longform: Zinc
 carbon_dioxide:
-  constituent_longform: Carbon Dioxide
\ No newline at end of file
+  constituent_longform: Carbon Dioxide
+foo:
```
kurbansitterley: Why is `foo` added to this list? I don't think we should have non-real constituents in our component list.

kurbansitterley (changes requested):
I don't think we should have `foo` in our constituent list or in removal fractions for any unit process. There are plenty of real constituents to test with.

watertap/data/techno_economic/component_list.yaml
luohezhiming: Yeah, this model is previously developed by Markus, I thought there is a reason he used `foo` but it turns out not. I've revised the code accordingly.

kurbansitterley (approved):
LGTM, thanks

watertap/data/techno_economic/waiv.yaml
```diff
@@ -2,29 +2,26 @@ default:
...
-    capital cost units:
-      value: M$
-    year:
-      value: 2020
-    cost_function_form:
-      function f(flow_mass_in): cap_basis_par*(flow_mass_in/flow_basis_par)^cap_exp
-      units: $
   recovery_frac_mass_H2O:
-    value: 0.9999
+    value: 0.0001
```
andrewlee94, line 18: Just to confirm is this change correct?

andrewlee94 (commented):
I had one question about a change in a parameter value - could someone check and confirm this.

watertap/data/techno_economic/waiv.yaml
luohezhiming: @kurbansitterley @adam-a-a Could you please check this?

kurbansitterley: I would say this is correct and good catch by @luohezhiming. In WT3 the WAIV was always a terminal unit so we didn't care about the recovery because the flow went nowhere. But this process is intended to have most of the inflow be evaporated so would in reality have very little water recovery

andrewlee94 (approved):

PR #435, opened 2022-03-18
Migrating electrodialysis reversal model

luohezhiming:
## Summary/Motivation:
 
Add electrodialysis reversal model (Irwin project)

- add electrodialysis reversal model
- add corresponding tests

### Legal Acknowledgement

### DIFF
--- a/watertap/unit_models/zero_order/electrodialysis_reversal_zo.py
+++ b/watertap/unit_models/zero_order/electrodialysis_reversal_zo.py
@@ -1,0 +1,84 @@
+###############################################################################
+# WaterTAP Copyright (c) 2021, The Regents of the University of California,
+# through Lawrence Berkeley National Laboratory, Oak Ridge National
+# Laboratory, National Renewable Energy Laboratory, and National Energy
+# Technology Laboratory (subject to receipt of any required approvals from
+# the U.S. Dept. of Energy). All rights reserved.
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and license
+# information, respectively. These files are also available online at the URL
+# "https://github.com/watertap-org/watertap/"
+#
+###############################################################################
+"""
+This module contains a zero-order representation of an electrodialysis reversal unit.
+"""
+
+from pyomo.environ import Constraint, units as pyunits, Var, Param
+from idaes.core import declare_process_block_class
+
+from watertap.core import build_sido, constant_intensity, ZeroOrderBaseData
+
+# Some more information about this module
+__author__ = "Chenyu Wang"
+
+
+@declare_process_block_class("ElectrodialysisReversalZO")
+class ElectrodialysisReversalZOData(ZeroOrderBaseData):
+    """
+    Zero-Order model for an electrodialysis reversal unit operation.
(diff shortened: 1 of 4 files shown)


kurbansitterley (approved):

watertap/unit_models/zero_order/electrodialysis_reversal_zo.py
```diff
@@ -0,0 +1,84 @@
...
+                'TDS must be included in the solute list for determining'
+                ' electricity intensity and power consumption of the electrodialysis '
+                'reversal unit.')
+
+        self.elec_coeff_1 = Param(initialize=0.2534,
+                                  units=pyunits.kWh/pyunits.m**3,
+                                  doc="Constant 1 in electricity intensity equation")
+        self.elec_coeff_2 = Param(initialize=5.149E-4,
+                                  units=pyunits.L/pyunits.mg*pyunits.kWh/pyunits.m**3,
+                                  doc="Constant 2 in electricity intensity equation")
```
adam-a-a, line 52: following how Andrew redid the brine concentrator unit and added crystallizer as a subset with costing, we might want to follow the same changes here by (1) add these as variables here and (2) adding them to the yaml file.

adam-a-a (changes requested):
Everything looks good. My only suggested changed is to take those electricity coefficient parameters and put them in the yaml. You can look at @andrewlee94 's PR where he changed some things around with the brine concentrator and crystallizer.

adam-a-a:
FYI: this is the PR I'm referring to: https://github.com/watertap-org/watertap/pull/415

luohezhiming:
> Everything looks good. My only suggested changed is to take those electricity coefficient parameters and put them in the yaml. You can look at @andrewlee94 's PR where he changed some things around with the brine concentrator and crystallizer.

The reason we took those coefficients as variables for the brine concentrator case is that it has a subtype and those coefficients will be different for different subtypes. For this unit, I think it is fine to use parameters.

adam-a-a:
> > Everything looks good. My only suggested changed is to take those electricity coefficient parameters and put them in the yaml. You can look at @andrewlee94 's PR where he changed some things around with the brine concentrator and crystallizer.
> 
> The reason we took those coefficients as variables for the brine concentrator case is that it has a subtype and those coefficients will be different for different subtypes. For this unit, I think it is fine to use parameters.

The only thing is that electrodialysis will most likely have subtypes. I can think of at least one needed for AMO work, and we might also want a conventional ED subtype (without reversal). With that said, I suppose we can just merge this, and then when/if subtypes arise, we'd make that change.

adam-a-a (approved):

adam-a-a:
@luohezhiming I'll approve this now, and if we need subtypes, we'll change it later.

andrewlee94:
I would tend to agree with @adam-a-a's initial comment here. Just because we don't have different parameters yet doesn't mean we won't in future. We should avoid hard-coding parameters wherever possible (but I see I am too late now).

luohezhiming:
@adam-a-a @andrewlee94 I will add those two parameters as variables right now.

PR #439, opened 2022-03-21
Add costing model for coagulation/flocculation unit

luohezhiming:
## Summary/Motivation:

Add costing model for coagulation/flocculation

- add costing model
- modify `total_power` to `electricity` and register electricity
- add corresponding tests

### Legal Acknowledgement

### DIFF
--- a/watertap/core/zero_order_costing.py
+++ b/watertap/core/zero_order_costing.py
@@ -588,10 +588,7 @@ class ZeroOrderCostingData(FlowsheetCostingBlockData):
                 cost_floc_inj,
                 to_units=blk.config.flowsheet_costing_block.base_currency))
 
-        if factor == "TPEC":
-            expr *= blk.config.flowsheet_costing_block.TPEC
-        elif factor == "TIC":
-            expr *= blk.config.flowsheet_costing_block.TIC
+        expr *= blk.config.flowsheet_costing_block.TPEC
 
         blk.capital_cost_constraint = pyo.Constraint(
             expr=blk.capital_cost == expr)
--- a/watertap/core/zero_order_costing.py
+++ b/watertap/core/zero_order_costing.py
@@ -588,7 +588,10 @@ class ZeroOrderCostingData(FlowsheetCostingBlockData):
                 cost_floc_inj,
                 to_units=blk.config.flowsheet_costing_block.base_currency))
 
-        expr *= blk.config.flowsheet_costing_block.TPEC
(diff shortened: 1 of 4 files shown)


watertap/core/zero_order_costing.py
```diff
@@ -518,6 +519,84 @@ def cost_chlorination(blk):
...
+                cost_floc,
+                to_units=blk.config.flowsheet_costing_block.base_currency) +
+            pyo.units.convert(
+                cost_coag_inj,
+                to_units=blk.config.flowsheet_costing_block.base_currency) +
+            pyo.units.convert(
+                cost_floc_inj,
+                to_units=blk.config.flowsheet_costing_block.base_currency))
+
+        expr *= blk.config.flowsheet_costing_block.TPEC
```
andrewlee94: Whilst the original WT3 models hard-coded this decision, I have generally been setting this in the YAML file so as to allow more flexibility in future.

watertap/unit_models/zero_order/tests/test_coag_and_floc_zo.py
```diff
@@ -17,17 +17,20 @@
 
 from io import StringIO
 from pyomo.environ import (
+    Block, check_optimal_termination,
```
andrewlee94: Is `check_optimal_termination` being used in this test? I did a quick search and couldn't see it being used, which would be on oversight, as we should be testing that we can solve the model.

andrewlee94 (changes requested):
Two fairly minor things that should be fixed, but otherwise looks good.

watertap/unit_models/zero_order/tests/test_coag_and_floc_zo.py
luohezhiming: No, this is not used in the test. I cleaned up this.

watertap/core/zero_order_costing.py
luohezhiming: OK. I revised following the convention and added an option for choosing from `TPEC` or `TIC`.

watertap/unit_models/zero_order/tests/test_coag_and_floc_zo.py
adam-a-a: `assert_optimal_termination` was (and still is) included to account for this.

andrewlee94 (approved):

adam-a-a (approved):

PR #440, opened 2022-03-21
Migrating WT3 cooling tower model

luohezhiming:
## Summary/Motivation:

Add cooling tower model

- add cooling tower model
- add corresponding tests

### Legal Acknowledgement

### DIFF
--- a/watertap/data/techno_economic/cooling_tower.yaml
+++ b/watertap/data/techno_economic/cooling_tower.yaml
@@ -1,6 +1,6 @@
 default: # make_up_mass_balance method in the WT3 cooling_tower.py is used
   energy_electric_flow_vol_inlet:
-    value: 0.000000001
+    value: 0
     units: kWh/m^3
   capital_cost:
     basis: flow_vol
--- a/watertap/data/techno_economic/cooling_tower.yaml
+++ b/watertap/data/techno_economic/cooling_tower.yaml
@@ -1,8 +1,30 @@
-default:
+default: # make_up_mass_balance method in the WT3 cooling_tower.py is used
+  energy_electric_flow_vol_inlet:
+    value: 0.000000001
+    units: kWh/m^3
+  capital_cost:
+    basis: flow_vol
+    cost_factor: None
+    reference_state:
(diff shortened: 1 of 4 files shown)


watertap/unit_models/zero_order/cooling_tower_zo.py
```diff
@@ -0,0 +1,60 @@
+###############################################################################
+# WaterTAP Copyright (c) 2021, The Regents of the University of California,
```
tarnold17, line 2: Should the copyright date in the header still be 2021 even though it is now 2022?

luohezhiming: I didn't pay attention to this before, but all our migrated models from WT3 use the same copyright. @adam-a-a

andrewlee94: We should not change the copyright header ourselves. Firstly, there is a lot of paperwork that goes into getting it approved, and secondly there is an automated tool for updating the headers when required.

watertap/data/techno_economic/cooling_tower.yaml
```diff
@@ -1,8 +1,30 @@
-default:
+default: # make_up_mass_balance method in the WT3 cooling_tower.py is used
+  energy_electric_flow_vol_inlet:
+    value: 0.000000001
```
andrewlee94: I just want to confirm this value before proceeding - it is very small. @kurbansitterley Are the units correct, or is this value meant to be effectively zero (but given a small positive value instead)?

andrewlee94 (dismissed):
Looks good, but I had a question about the value of electricity intensity.

watertap/data/techno_economic/cooling_tower.yaml
kurbansitterley: Should be effectively zero.

kurbansitterley (dismissed):

andrewlee94 (approved):

adam-a-a (approved):

PR #441, opened 2022-03-22
WT3: Adding costing for GAC units

andrewlee94:
- Rework GAC and fixed bed models to include electricity intensity calculations regressed from EPA data
- Add correlations for chemical demands in GAC and fixed beds
- Add capital cost methods for GAC and fixed bed units

### Legal Acknowledgement

### DIFF
--- a/watertap/core/zero_order_costing.py
+++ b/watertap/core/zero_order_costing.py
@@ -30,6 +30,7 @@ from watertap.unit_models.zero_order import (
     ChemicalAdditionZO,
     ChlorinationZO,
     CoagulationFlocculationZO,
+    FixedBedZO,
     GACZO,
     LandfillZO,
     IonExchangeZO,
@@ -601,6 +602,18 @@ class ZeroOrderCostingData(FlowsheetCostingBlockData):
         blk.config.flowsheet_costing_block.cost_flow(
             blk.unit_model.electricity[t0], "electricity")
 
+    def cost_fixed_bed(blk):
+        """
+        General method for costing fixed bed units. This primarily calls the
+        cost_power_law_flow method.
+
+        This method also registers demand for a number of additional material
+        flows..
+        """
+        t0 = blk.flowsheet().time.first()
+
+        ZeroOrderCostingData.cost_power_law_flow(blk)
+
     def cost_gac(blk):
         """
         General method for costing granular activated carbon processes. Capital
@@ -1156,6 +1169,7 @@ class ZeroOrderCostingData(FlowsheetCostingBlockData):
                     ChemicalAdditionZO: cost_chemical_addition,
                     ChlorinationZO: cost_chlorination,
                     CoagulationFlocculationZO: cost_coag_and_floc,
+                    FixedBedZO: cost_fixed_bed,
                     GACZO: cost_gac,
                     LandfillZO: cost_landfill,
                     IonExchangeZO: cost_ion_exchange,
--- a/watertap/core/zero_order_costing.py
+++ b/watertap/core/zero_order_costing.py
@@ -30,6 +30,7 @@ from watertap.unit_models.zero_order import (
     ChemicalAdditionZO,
     ChlorinationZO,
     CoagulationFlocculationZO,
+    GACZO,
     LandfillZO,
     IonExchangeZO,
     OzoneZO,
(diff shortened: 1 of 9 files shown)


watertap/data/techno_economic/gac.yaml
```diff
@@ -1,4 +1,7 @@
 default: # TODO: confirm any differentiation in removal across subtypes
+  volumetric_electricity_intensity:
+    value: 0.106310
+    units: kW/m^3
```
adam-a-a, line 4: This is technically power intensity if we have units of kW/m3 (whereas we've been using electricity intensity for energy consumption per volumetric flowrate at inlet in units of kWh/m3).

adam-a-a: Although I see why you called it this-- dividing by volume rather than volume/time.

adam-a-a (dismissed):
Just one minor comment on `volumetric_electricity_intensity`, but looks good to me.

watertap/data/techno_economic/gac.yaml
andrewlee94: I am not sure that is quite right either. For this unit, we have `intensity = a*ebct` where this is `a`. I.e. this is a parameter for calculating intensity based on the empty bed contact time.

This reason for the lack of time units here is because it gets canceled by the units on `ebct`.

luohezhiming (dismissed):

watertap/data/techno_economic/gac.yaml
adam-a-a: Actually, doesn't multiplying by `ebct` turn intensity into kWh/m3  (i.e., multiplying by time) rather than canceling out time units?

andrewlee94: No, intensity has units of kWh/m^3 (at least for every other model we have done).

adam-a-a: Yes, that's what I am saying too-- we'll end up with kWh/m^3, which is correct. 

To clarify my point, I think a better name for `volumetric_electricity_intensity` would be `power_electric_flow_vol_inlet` (analogous to our naming for `energy_electric_flow_vol_inlet`, where energy is kWh, and in `power_electric_flow_vol_inlet`, power is kW).

Again, very minor, but I think this would be more consistent with our naming convention at this point.

watertap/unit_models/zero_order/gac_zo.py
```diff
@@ -37,7 +37,74 @@ def build(self):
...
+        self.energy_electric_flow_vol_inlet = Var(
+            units=pyunits.kWh/pyunits.m**3,
+            doc="Electricity intensity with respect to inlet flowrate of unit")
+
+        @self.Constraint(
+            doc='Electricity intensity based on empty bed contact time.')
+        def electricity_intensity_constraint(b):
+            return (b.energy_electric_flow_vol_inlet ==
+                    b.volumetric_electricity_intensity *
+                    b.empty_bed_contact_time)
```
kurbansitterley, line 71: Where does getting energy from electricity_intensity*EBCT come from? Is that in WT3? I don't think I have seen this relationship before.

andrewlee94: That is from the EPA cost curves - the data shows that intensity is a function of `ebct`, however with only two data points the best we can do is a linear relationship.

kurbansitterley: There should be three data points for each curve. But the EBCT column in that dataset is only supposed to be used for selecting data. We fit the electricity intensity data to E = A * Q ** B.

andrewlee94: We did basically the same thing here - we just let the A term be a function of `ebct` as well. This way we support values that are not 30 and 60 minutes as well.

watertap/core/zero_order_costing.py
```diff
@@ -600,6 +602,96 @@ def cost_coag_and_floc(blk):
         blk.config.flowsheet_costing_block.cost_flow(
             blk.unit_model.electricity[t0], "electricity")
 
+    def cost_fixed_bed(blk):
+        """
+        General method for costing fixed bed units. This primarily calls the
+        cost_power_law_flow method.
+
+        This method also registers demand for a number of additional material
+        flows..
```
luohezhiming:
```suggestion
        flows.
```

watertap/data/techno_economic/gac.yaml
andrewlee94: I renamed it to `electricity_intensity_parameter`

luohezhiming (dismissed):

luohezhiming (dismissed):

luohezhiming (approved):

adam-a-a (approved):

PR #443, opened 2022-03-22
Adding power usage / performance equations to Coagulation-Flocculation 

aladshaw3:
## Fixes/Addresses:

Adds the necessary power usage expressions and constraints to the coagulation-flocculation model

## Summary/Motivation:

Initial coagulation-flocculation model did not include power usage functions. This PR adds those functions to the unit model and tests them. 

- Modified the Coagulation model to include power functions
- Modified the unit tests to account for changes 
- Add dynamic viscosity calculation to the property package

### Legal Acknowledgement


watertap/unit_models/coag_floc_model.py
```diff
@@ -12,6 +12,7 @@
 ###############################################################################
 
 # Import Pyomo libraries
+from math import pi
```
andrewlee94: It ends up being the same thing, but we generally use `from idaes.core.util.constants import pi` for this (mostly just to advertize the IDAES Constants module).

watertap/unit_models/tests/test_coag_floc_model.py
```diff
@@ -104,19 +104,50 @@ def test_build_model(self, coag_obj_w_chems):
...
+
     @pytest.mark.unit
     def test_stats(self, coag_obj_w_chems):
         model = coag_obj_w_chems
 
         # Check to make sure we have the correct DOF and
         #   check to make sure the units are correct
         assert_units_consistent(model)
-        assert degrees_of_freedom(model) == 11
+        assert degrees_of_freedom(model) == 11 + 11
```
andrewlee94: Why not just do 22? At most, I would add a comment explaining what the difernt DoF are, but just test against the total number,

I see this in a few other places below too.

andrewlee94 (dismissed):
I only had two minor comments ,but nothing that should stop this being merged.

watertap/unit_models/coag_floc_model.py
aladshaw3: Made the suggested change

watertap/unit_models/tests/test_coag_floc_model.py
aladshaw3: Updated all asserts based on suggestion.

watertap/unit_models/coag_floc_model.py
```diff
@@ -369,6 +372,109 @@ def build(self):
             doc="Moles of the produced salts from 1 mole of chemical additives",
         )
 
+        # Below set of Vars are for the power usage of the unit
```
lbibl, line 375: I feel `domain` repeats `bounds` in defining `Var` in this section so recommend deleting `domain = ` where bounds are defined.  Correct me if I am wrong.

watertap/unit_models/coag_floc_model.py
```diff
@@ -524,6 +630,173 @@ def eq_mass_transfer_term(self, t, p, j):
...
+        )
+
+        @self.Constraint(
+            self.flowsheet().config.time,
+            doc="Constraint for the power usage of the rapid mixing basins",
+        )
+        def eq_rapid_mixing_power(self, t):
+            vel_grad = pyunits.convert(
+                self.rapid_mixing_vel_grad[t], to_units=units_meta("time") ** -1
+            )
```
lbibl, line 681: Is this unit converting  necessary?  I think the two are both `s ** -1` ?

watertap/unit_models/coag_floc_model.py
```diff
@@ -714,6 +1043,97 @@ def calculate_scaling_factors(self):
...
+        # set scaling for total_power
+        if iscale.get_scaling_factor(self.total_power[t]) is None:
+            sf1 = iscale.get_scaling_factor(self.flocculation_power)
+            sf2 = iscale.get_scaling_factor(self.rapid_mixing_power)
+            sf = (sf1 + sf2) / 2
+            iscale.set_scaling_factor(self.total_power, sf)
+
+            for ind, c in self.eq_total_power.items():
+                iscale.constraint_scaling_transform(c, sf)
+
```
lbibl, line 1136: It seems to be a sophisticated method of assigning scaling factors.  Without checking eq details, I would double if there is a chance of sf turning out 0, which may cancel a var or constraint?

watertap/unit_models/coag_floc_model.py
lbibl: I have a general question on the use of `pyunits.convert`.  It looks to me that this command forces the unit change of a pyomo var or param.  If so, I would wonder why this is needed.  If an equation is well formulated, the unit must be balanced on two sides.  If a unit conversion is needed, does that suggest any issue with the equation?

aladshaw3: Yes, this is always needed. This is because the `units_meta` comes from the property package, which the user may define (if using a generic package). The variable `rapid_mixing_vel_grad` is defined by the developer in the unit model and is defined to have specific units (in this case `s**-1`). In order to ENSURE that the units are consistent between the user defined property package and the developer defined unit model variables, we should always convert our defined units into the units defined in `units_meta`. 

For instance, with this conversion, if the user defines the units of time for the flowsheet/properties to be in `min`, then the conversion with apply the appropriate factor to `rapid_mixing_vel_grad` to the `min` time basis before constructing the constraint. 

In this particular case, both are already in `s**-1`, but we cannot assume that will always be the case.

watertap/unit_models/coag_floc_model.py
aladshaw3: The `domain` is slightly different than `bounds`, but in many cases they have some overlap. The `domain` refers more to the 'type' of numbers that apply for the variable (e.g., complex, real, integers, etc), whereas the `bounds` actually apply inequality constraints to the variables to bound them between the given values. 

In certain cases, such as `domain = NonNegativeReals` with `bounds = (0, None)`, they would effectively have the same impact since the full set of 'NonNegativeReals' is within the specified bounds. 

In other cases, however, it is necessary to specify both (e.g., `domain = PositiveIntegers` with `bounds = (1, 10)` would be needed for integer optimization that is bounded. If only the bounds were provided, then pyomo would assume that non-integers are allowed since the default `domain` is just 'Reals'). 

It is never harmful to apply both a domain and bounds for variables (unless they are contradictory). They each mean something different, even if they sometimes overlap.

andrewlee94: That is debatable, but there is not really any problem with having both a `domain` and `bounds` in most cases (there are a few edge cases, but they normally don't show up). One reason for having both is so that the `domain` forces an absolute bound that cannot be violated (even if the user tries), whilst the `bounds` can be used to provide adjustable (tighter) limits as needed.

watertap/unit_models/coag_floc_model.py
andrewlee94: @lbibl `pyunits.convert` does not change the defined units of the `Var` or `Param`. Instead, it determines the necessary unit conversion factor and adds it to the expression under construction. E.g.:

```
m = ConcreteModel()

m.v = Var(units=pyunits.kg)

print(pyunits.convert(m.v, to_units=pyunits.pound)

>>> (2.20462*pounds/kg)*m.v
```

watertap/unit_models/coag_floc_model.py
aladshaw3: Since scaling factors are always positive numbers (and these are all built off of other scaling factors, which are in turn built on scaling factors the user provides) it should only every give a `0` for a scaling factor if the user gives a `0` for a scaling factor, which the user should never do.

adam-a-a (approved):
LGTM

lbibl (approved):

PR #444, opened 2022-03-22
Wt3 evap pond

kurbansitterley:
## Summary/Motivation:

Add WT3 evaporation pond model

- add evaporation pond model, yaml, and necessary tests

### Legal Acknowledgement


watertap/core/zero_order_costing.py
```diff
@@ -913,6 +914,50 @@ def cost_uv_aop(blk):
...
+            C*blk.unit_model.land_cost[t0]+
+            D*blk.unit_model.land_clearing_cost[t0]+
+            E*blk.unit_model.dike_height[t0]), 
+            to_units=blk.config.flowsheet_costing_block.base_currency)
+
+        blk.capital_cost_constraint = pyo.Constraint(
+            expr=blk.capital_cost == expr)
+
+        # Register flows
+        blk.config.flowsheet_costing_block.cost_flow(
```
andrewlee94: Does an evaporation pond even need an electricity demand?

watertap/data/techno_economic/evaporation_pond.yaml
```diff
@@ -144,3 +203,5 @@ default:
       value: 0.98
       units: dimensionless
       constituent_longform: zinc
+
+
```
andrewlee94: Extra line break at end of file - should be just 1.

watertap/unit_models/zero_order/evaporation_pond_zo.py
```diff
@@ -0,0 +1,161 @@
...
+    def build(self):
+        super().build()
+
+        build_sido(self)
+        constant_intensity(self)
+        self._Q_out = Reference(self.properties_byproduct[:].flow_vol)
+
+        self._tech_type = "evaporation_pond"
+
+        self.air_temp = Var(self.flowsheet().time,
```
andrewlee94: We generally use `temperature` rather than `temp`.

watertap/unit_models/zero_order/evaporation_pond_zo.py
```diff
@@ -0,0 +1,161 @@
...
+
+        self.solar_radiation = Var(self.flowsheet().time,
+                    units=pyunits.mJ/pyunits.m**2,
+                    doc="Daily solar radiation incident")
+
+        self.liner_thickness = Var(self.flowsheet().time,
+                    units=pyunits.mil,
+                    doc="Thickness of evaporation pond liner")
+
+        self.land_cost = Var(self.flowsheet().time,
```
andrewlee94: Parameters specifically related to cost should generally go in the costing Block rather than the main unit model.

watertap/unit_models/zero_order/evaporation_pond_zo.py
```diff
@@ -0,0 +1,161 @@
...
+    """
+
+    CONFIG = ZeroOrderBaseData.CONFIG()
+
+    def build(self):
+        super().build()
+
+        build_sido(self)
+        constant_intensity(self)
+        self._Q_out = Reference(self.properties_byproduct[:].flow_vol)
```
andrewlee94: Do you actually need the `Reference` to Q? From what I saw later, you should be able to use the Var directly.

watertap/unit_models/zero_order/tests/test_evaporation_pond_zo.py
```diff
@@ -0,0 +1,329 @@
...
+        assert isinstance(model.fs.unit.adj_area, Var)
+        assert isinstance(model.fs.unit.evaporation_rate_pure, Var)
+        assert isinstance(model.fs.unit.evaporation_rate_salt, Var)
+        assert isinstance(model.fs.unit.electricity_consumption, Constraint)
+        assert isinstance(model.fs.unit.evap_rate_pure_constraint, Constraint)
+        assert isinstance(model.fs.unit.evap_rate_salt_constraint, Constraint)
+        assert isinstance(model.fs.unit.area_constraint, Constraint)
+        assert isinstance(model.fs.unit.area_adj_constraint, Constraint)
+
+
```
andrewlee94: To many line breaks. PEP8 standards have only 1 line between methods within classes.

andrewlee94 (changes requested):
A couple of minor things I think need to be fixed, plus a few formatting issues.

watertap/core/zero_order_costing.py
kurbansitterley: It doesn't have one but I wasn't sure if we wanted to have something registered for every unit process just as a convention regardless. But I will remove it

watertap/data/techno_economic/evaporation_pond.yaml
kurbansitterley:
```suggestion
```

watertap/unit_models/zero_order/evaporation_pond_zo.py
kurbansitterley: You are correct. No need for it.

watertap/unit_models/zero_order/evaporation_pond_zo.py
kurbansitterley: Ok I will move them. `dike_height` is used as part of the `area` calculation, so it needs a value on this block.

andrewlee94 (approved):
Assuming tests pass, I think this is good to go.

adam-a-a (approved):

PR #445, opened 2022-03-23
Migrating WT3 deep well injection model

luohezhiming:
## Summary/Motivation:

Add deep well injection model

- add deep well injection model
- add costing
- add tests

### Legal Acknowledgement

### DIFF
--- a/watertap/core/zero_order_costing.py
+++ b/watertap/core/zero_order_costing.py
@@ -30,6 +30,7 @@ from watertap.unit_models.zero_order import (
     ChemicalAdditionZO,
     ChlorinationZO,
     CoagulationFlocculationZO,
+    DeepWellInjectionZO,
     LandfillZO,
     IonExchangeZO,
     OzoneZO,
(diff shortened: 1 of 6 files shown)


watertap/core/zero_order_costing.py
```diff
@@ -600,6 +601,59 @@ def cost_coag_and_floc(blk):
...
+                subtype=blk.unit_model.config.process_subtype)
+
+        # Get costing parameter sub-block for this technology
+        A, B, C = _get_tech_parameters(
+            blk,
+            parameter_dict,
+            blk.unit_model.config.process_subtype,
+            ["well_pump_cost",
+             "pipe_cost_basis",
+             "capital_b_parameter"])
```
andrewlee94: I am not sure this is the best name for this parameter in this case (notably there is no `a` parameter). Maybe `flow_exponent` instead?

watertap/data/techno_economic/deep_well_injection.yaml
```diff
@@ -0,0 +1,33 @@
+default:
+  recovery_frac_mass_H2O:
+    value: 0.9999
+    units: dimensionless
+  default_removal_frac_mass_solute:
+    value: 0
+    units: dimensionless
+  capital_cost:
+    cost_factor: None
+    reference_state:
```
andrewlee94: I don't think this is used, so it can be removed.

andrewlee94 (changes requested):
Two minor comments to fix.

watertap/data/techno_economic/deep_well_injection.yaml
luohezhiming: I used the inlet `flow_vol` for costing (sizing term). I added `basis: flow_vol` here to clarify.

andrewlee94: I was referring to `reference_state`, which is not used at all in the code so shouldn't be in the yaml file either (there is a separate flow basis reference later which covers this, but is also used elsewhere in the model).

watertap/core/zero_order_costing.py
luohezhiming: Previously I just follow the convention that we typically used b_parameter to represent the exponent term. I've modified to `flow_exponent`

watertap/data/techno_economic/deep_well_injection.yaml
andrewlee94: Also, the `flow_basis` isn't used either, so it shouldn't be here as well. We don't need to make all the YAML files look the same just for the sake of it, and we would be better off including only what is actually necessary (i.e. used by the code).

andrewlee94 (approved):

watertap/data/techno_economic/deep_well_injection.yaml
tarnold17: @luohezhiming @andrewlee94 I see that `reference_state` has been removed from this YAML file, but `flow_basis` is still included. Is it necessary to remove the latter before merging this PR?

andrewlee94: @tarnold17 In this case `flow_basis` is still required for the actual unit model (rather than just the capital cost calculations as in other cases). Good question to ask however.

adam-a-a (approved):

PR #446, opened 2022-03-23
Streamline results storage within parameter_sweep.py

k1nshuk:
## Fixes/Addresses:

This PR consolidates how the results for CSV and the H5 files were being collected for subsequent writing. 

## Summary/Motivation:

Previously the data structures for writing the CSV and H5 files were being updated separately within the parameter sweep loop. The old way of creating the arrays for the CSV file was left in place to ensure nothing broke while implementing H5 outputs. This is no longer needed as we test it appropriately and leaving it adds confusion and increases chances of bugs in the future

- The arrays needed for storing the CSV files are now created from the results dictionary
- Changes to the function arguments of `parameter_sweep`
- Code rearrangement to use more private functions
- Updated documentation to reflect the above changes

### Legal Acknowledgement


watertap/tools/parameter_sweep.py
```diff
@@ -582,110 +643,88 @@ def _do_param_sweep(model, sweep_params, outputs, local_values, optimize_functio
...
-        _write_outputs(global_output_dict, dirname, h5_results_file, txt_options="keys")
+        _write_outputs(global_results_dict, dirname, fname_no_ext, txt_options="keys")
 
     return global_save_data
 
 
 # ================================================================
 
-def parameter_sweep(model, sweep_params, outputs=None, csv_results_file=None, h5_results_file=None,
+def parameter_sweep(model, sweep_params, outputs=None, results_file_name=None, write_csv=False, write_h5=False,
```
bknueven, line 734: As a general comment -- I do not particularly like this interface change. Before I just needed to specify a single parameter for a results file, and now I need to specify two.

Also, if we're going with this `results_file_name` become a positional argument (nonsensical *not* to specify it) and `write_h5` should be switched to `True` by default. This will at least avoid some potential pitfalls.

watertap/tools/parameter_sweep.py
```diff
@@ -376,19 +378,23 @@ def _create_component_output_skeleton(component, num_samples):
...
     if hasattr(component, 'get_units'):
         unit_obj = component.get_units()
         if unit_obj is not None:
             comp_dict["units"] = component.get_units().name
         else:
             comp_dict["units"] = "None"
 
+    # Add information to this output that WILL NOT be written as part
+    # of the file saving step.
+    comp_dict['_pyo_obj'] = component
```
bknueven, line 391: I'm a bit worried that we'll be sending Pyomo objects over MPI now. That should generally be avoided -- they have to be pickled and reconstructed on the other end. This could become a major bottleneck on rank 0 for large scale sweeps.

bknueven (commented):
I have a couple of high-level comments.

watertap/tools/parameter_sweep.py
k1nshuk: I added a condition where if a file extension is provided but its corresponding write bool is false, it gets overridden. So the user does not need to specify two parameters to get output files now.
https://github.com/k1nshuk/watertap/blob/d3ec6653b9fa1ab0b4bab6005cd44d58ae4380b4/watertap/tools/parameter_sweep.py#L692-L701

watertap/tools/parameter_sweep.py
k1nshuk: We don't send promo objects using MPI. The only items that are collected on rank 0 are numpy arrays.
https://github.com/k1nshuk/watertap/blob/d3ec6653b9fa1ab0b4bab6005cd44d58ae4380b4/watertap/tools/parameter_sweep.py#L446-L448

Those pyomo objects remain in the local output dictionary and are popped off before any h5 files are written.

watertap/tools/parameter_sweep.py
```diff
@@ -137,9 +137,23 @@ def _init_mpi(mpi_comm=None):
...
+# ================================================================
+
+def _process_results_filename(results_file_name):
+    # Get the directory path
+    dirname = os.path.dirname(results_file_name)
+    # Get the file name without the extension
+    known_extensions = ['.h5', '.csv']
+    for ext in known_extensions:
+        fname_no_ext, status, extension = _strip_extension(results_file_name, ext)
+        if status == 1:
```
bknueven: A return handle like this would be more pythonic -- no need for the `status` variable.

```suggestion
        fname_no_ext, extension = _strip_extension(results_file_name, ext)
        if extension is not None:
```

watertap/tools/parameter_sweep.py
```diff
@@ -137,9 +137,23 @@ def _init_mpi(mpi_comm=None):
 
 def _strip_extension(file_name, extension):
     if file_name.lower().endswith(extension):
-        return file_name[:-len(extension)]
+        return file_name[:-len(extension)], 1, extension
```
```suggestion
        return file_name[:-len(extension)], extension
```

watertap/tools/parameter_sweep.py
```diff
@@ -137,9 +137,23 @@ def _init_mpi(mpi_comm=None):
...
     else:
-        return file_name
+        return file_name, 0, None
```
```suggestion
        return file_name, None
```

watertap/tools/parameter_sweep.py
```diff
@@ -582,110 +643,95 @@ def _do_param_sweep(model, sweep_params, outputs, local_values, optimize_functio
...
+            # run_successful remains false. We try to reinitialize and solve again
+            if reinitialize_function is not None:
+                try:
+                    reinitialize_function(model, **reinitialize_kwargs)
+                    with capture_output():
+                        results = optimize_function(model, **optimize_kwargs)
+                    pyo.assert_optimal_termination(results)
+
+                except:
+                    pass # Run Successful is still False
```
```suggestion
                    pass # run_successful is still False
```

watertap/tools/parameter_sweep.py
```diff
@@ -582,110 +643,95 @@ def _do_param_sweep(model, sweep_params, outputs, local_values, optimize_functio
...
+                warnings.warn("An H5 results filename was provided. Outputs will be created", UserWarning)
+                write_h5 = True
+            elif extension == '.csv' and write_csv is False:
+                warnings.warn("A CSV results filename was provided. Outputs will be created", UserWarning)
+                write_csv = True
+
+            if (write_h5 or write_csv) and dirname != '':
                 os.makedirs(dirname, exist_ok=True)
+            # elif not write_h5 and not write_csv:
+            #     warnings.warn("A results filename was provided but neither options to write H5 or csv was selected. No file will be written.", UserWarning)
```
bknueven: This code should either be removed or un-commented -- probably un-commented, given the warnings above.
```suggestion
            elif not write_h5 and not write_csv:
                warnings.warn("A results filename was provided but neither options to write H5 or csv was selected. No file will be written.", UserWarning)
```

bknueven (commented):
I have a few more minor nits, which I left as code suggestions. Otherwise I think this is good to go, and helps refactor the code to minimize special logic when MPI is used.

eyoung55 (dismissed):
Looks good to me after Ben's clean ups are addressed.

bknueven (dismissed):

watertap/tools/tests/test_parameter_sweep.py
```diff
@@ -71,7 +72,7 @@ def test_init_mpi(self):
     def test_strip_extension(self):
         input_list = ['/my_dir/my_file.h5', 'my_file.csv', '/my_dir/myfile']
         extension_list = ['.h5', '.csv', '.h5']
-        true_list = ['/my_dir/my_file', 'my_file', '/my_dir/myfile']
+        true_list = [('/my_dir/my_file',1,'.h5'), ('my_file',1,'.csv'), ('/my_dir/myfile',0, None)]
```
bknueven:
```suggestion
        true_list = [('/my_dir/my_file', '.h5'), ('my_file', '.csv'), ('/my_dir/myfile', None)]
```

bknueven (approved):

eyoung55 (approved):
Looks great, thanks Kinshuk and Ben.

PR #452, opened 2022-03-24
Gas sparged membrane ZO

adam-a-a:
## Summary/Motivation:
Last unit in one of the AMO projects for wastewater resource recovery

- yaml, unit, and tests for gas_sparged membrane

### Legal Acknowledgement


watertap/data/techno_economic/gas_sparged_membrane.yaml
```diff
@@ -0,0 +1,18 @@
...
+    units: dimensionless
+  removal_frac_mass_solute:
+    cod:
+      value: 1    # Todo: revisit this value later
+      units: dimensionless
+      constituent_longform: Chemical Oxygen Demand
+  gas_mass_influent_ratio:   # 5 L H2/ L/day influent --> convert to mass ratio (multiply by rho_h2/rho_solution)
+    value: 0.00041175
+    units: dimensionless
+#  energy_electric_flow_vol_inlet:
```
andrewlee94: Commented code.

I assume this one doesn't have capital costing yet?

watertap/unit_models/zero_order/gas_sparged_membrane_zo.py
```diff
@@ -0,0 +1,308 @@
...
+
+        self._perf_var_dict["Water Recovery"] = self.recovery_frac_mass_H2O
+        self._perf_var_dict["Solute Removal"] = self.removal_frac_mass_solute
+
+        self._get_Q = _get_Q_gas_extraction
+
+        self._Q = Reference(self.properties_in[:].flow_vol)
+        pump_electricity(self, self._Q)
+
+def initialize_gas_extraction(blk, state_args=None, outlvl=idaeslog.NOTSET,
```
andrewlee94: Does this differ from the one in the SIDO method, and same for the scaling method below? If not, you could just import the method from there instead to save duplicated code (and automatically get any updates made to those).

andrewlee94 (dismissed):
A couple of minor comments but I don't think anything needs to hold this up.

Just to confirm however, the extracted gas is not considered to be a solute in the influent stream and the `gas_mass_influent_ratio` is assumed to be a fixed parameter, correct? Or to put it another way, we are not expecting conservation of mass for the gas.

watertap/data/techno_economic/gas_sparged_membrane.yaml
adam-a-a: No, no costs yet.

watertap/unit_models/zero_order/gas_sparged_membrane_zo.py
adam-a-a: There's one addition to the scaling method, but I can do that for the initialization method.

adam-a-a:
> A couple of minor comments but I don't think anything needs to hold this up.
> 
> Just to confirm however, the extracted gas is not considered to be a solute in the influent stream and the `gas_mass_influent_ratio` is assumed to be a fixed parameter, correct? Or to put it another way, we are not expecting conservation of mass for the gas.

That's correct- not at this stage at least with the level of information available.

watertap/unit_models/zero_order/tests/test_gas_sparged_membrane.py
```diff
@@ -202,6 +204,10 @@ def test_conservation(self, model):
 
     @pytest.mark.component
     def test_report(self, model):
+        if not is_solver_from_idaes_ext:
```
andrewlee94: Is this really needed for the test? Notably, the solve test above is not xfailed, so this should run in the standard test environment. I am less concerned about test failures showing up if a user runs in a non-standard environment - those would be legitimate failures indicating that they need to fix something.

andrewlee94 (dismissed):

watertap/unit_models/zero_order/tests/test_gas_sparged_membrane.py
adam-a-a: I just put this in because I saw test_report fail for the Mac check (instead of one of the values being displayed as 1e-15, it was displayed as 1e-18, or something like that). Since this is a harmless difference, I put this in to pass the check.

watertap/unit_models/zero_order/gas_sparged_membrane_zo.py
```diff
@@ -0,0 +1,223 @@
...
+
+@declare_process_block_class("GasSpargedMembraneZO")
+class GasSpargedMembraneZOData(ZeroOrderBaseData):
+    """
+    Zero-Order model for a gas-sparged membrane.
+    This unit is similar to a SIDO, but there is technically a third outlet for gas extraction.
+    Three StateBlocks are added with corresponding Ports:
+        * properties_inlet
+        * properties_treated
+        * properties_byproduct
```
luohezhiming, line 39: If gas extraction is a third outlet, why don't set it as another stream with a port?

watertap/unit_models/zero_order/tests/test_gas_sparged_membrane.py
```diff
@@ -0,0 +1,238 @@
...
+            else:
+                assert (abs(value(model.fs.unit.inlet.flow_mass_comp[t, j] -
+                                  model.fs.unit.treated.flow_mass_comp[t, j] -
+                                  model.fs.unit.byproduct.flow_mass_comp[t, j] -
+                                  model.fs.unit.flow_mass_gas_extraction[t]))
+                        <= 1e-6)
+
+    @pytest.mark.component
+    def test_report(self, model):
+        if not is_solver_from_idaes_ext:
```
lbianchi-lbl: This is definitely not critical, but I just wanted to mention that you can use the `requires_idaes_solver` pytest mark for this: simply decorate the test function (you can apply multiple pytest marks), and the test will be skipped/xfailed if the solver is not one of the IDAES solvers (i.e. on macOS):

```suggestion
    @pytest.mark.requires_idaes_solver
    @pytest.mark.component
    def test_report(self, model):
        # the rest of the test, without the if clause
```

andrewlee94 (approved):

luohezhiming (approved):

watertap/unit_models/zero_order/gas_sparged_membrane_zo.py
adam-a-a: I had a comment somewhere pointing out that I could've tried that, but this was the quickest route to get the initial unit in. The other thing was that H2 is not tracked at this point, and adding a third port would've required state variables, making the first pass at this trickier to implement.

Issue #387, opened 2022-03-05
Potential Bug in the Ipopt-Watertap Solver factory 


aladshaw3:
I came across this strange error (on windows) when trying to test out the development of a new unit model. I had done this same test on Linux (about a week ago) and it ran everything fine, but now on Windows the code gets caught in an infinite loop of errors whenever I attempt to initialize or solve the model. 

I was able to narrow it down to the solver factor for 'ipopt-watertap'. If I just use the standard solver factory from IDAES (i.e., change 'ipopt-watertap' to 'ipopt', the code runs fine. 

Here is how to recreate the issue:

(1) My basic unit model (this model is incomplete, but enough is there to run a solve on): https://github.com/aladshaw3/watertap/blob/solver-issue/watertap/unit_models/electrodialysis_0D.py 

(2) File to run the unit model: https://github.com/aladshaw3/watertap/blob/solver-issue/watertap/unit_models/run_elec_example.py 

In that above file (2) there is a boolean I use to change the solver from 'ipopt-watertap' to 'ipopt'. If set to false, it uses 'ipopt-watertap'. Setting that to true uses 'ipopt'. 

If you run that file as is (with the boolean set to false), then you get the following errors.
[ipopt-watertap-error.txt](https://github.com/watertap-org/watertap/files/8189430/ipopt-watertap-error.txt)

However, if you set that boolean to true (i.e., just use the default solver of IDAES, 'ipopt'), then the model initializes and solves fine.

bknueven:
This example runs fine on my system (Linux).

It looks to be an issue with complied libraries in an upstream package. What is the output of:
- `pip list`
- `python -c "from pyomo.contrib.pynumero.asl import AmplInterface; print(AmplInterface.available())"` ?

aladshaw3:
The example also runs fine on my linux machine. For whatever reason, this issue seems isolated to windows?

Here are the outputs of pip-list for windows and linux.

[pip-list-windows.txt](https://github.com/watertap-org/watertap/files/8199431/pip-list-windows.txt)
[pip-list-linux.txt](https://github.com/watertap-org/watertap/files/8199432/pip-list-linux.txt)

For the second command you listed, both windows and linux actually return 'False' for me.

aladshaw3:
Also, I have already re-installed watertap on both machines and have created a clean conda environment, but still have the same issues with Windows.

bknueven:
The second command should print `True`. Have you run `idaes get-extensions` recently on either machine?

[...]

PR #390, opened 2022-03-07
[DO NOT MERGE] Testing some issues through GitHub Actions

aladshaw3:
## Fixes/Addresses:

Related to #387

## Changes proposed in this PR:
-
-

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #453, opened 2022-03-24
Add documentation for running models in Python scripts

### LINKED ISSUE #431: Document workarounds for errors when running module-level code using `get_solver()` on Windows
- This issue was first noticed in #387 and tested using #390 
- In #387 we figured out the cause for this error
- We should document somewhere, most likely in a FAQ/How-To (Fix) section in the Sphinx docs

lbianchi-lbl:
## Summary/Motivation:

- Document recommended structure for Python scripts to avoid errors on Windows such as e.g. #387 

- Add how-to guide on running models in Python scripts

### Legal Acknowledgement


lbianchi-lbl:
Currently, for the example code [here](https://watertap--453.org.readthedocs.build/en/453/how_to_guides/how_to_run_models_in_a_py_script.html#example-python-file-with-recommended-structure), I mostly used the [snippet](https://github.com/aladshaw3/watertap/blob/dd81ed04d00c108664892cae807aa0b7fd28e5fb/basic_test.py) that @aladshaw3 was using in #390 to investigate the error, but if you can think of a simple way to make the example code in more representative (e.g. actually initialize/solve the model?), I'd be happy to add it.

docs/how_to_guides/how_to_run_models_in_a_py_script.rst
```diff
@@ -0,0 +1,113 @@
...
+
+
+   m = pyo.ConcreteModel()
+   m.x = pyo.Var()
+   m.c = pyo.Constraint(expr=(0, m.x, 1))
+   m.o = pyo.Objective(expr=m.x)
+
+   nlp = PyomoNLP(m)
+
+
```
aladshaw3, line 166:
```suggestion


Example: Python file with recommended structure to run RO model
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

.. code-block::

   # Import concrete model from Pyomo
   from pyomo.environ import ConcreteModel
   # Import flowsheet block from IDAES core
   from idaes.core import FlowsheetBlock
   # Import NaCl property model
   import watertap.property_models.NaCl_prop_pack as props
   # Import utility tool for calculating scaling factors
   from idaes.core.util.scaling import calculate_scaling_factors
   # Import RO model
   from watertap.unit_models.reverse_osmosis_0D import ReverseOsmosis0
   # import the solver 
   from idaes.core.util import get_solver
   
   # Get default watertap solver 
   solver = get_solver()

   # Put all the model constructors, initialization, and solver in a separate function
   def main():

       # Create a concrete model, flowsheet, and NaCl property parameter block.
       m = ConcreteModel()
       m.fs = FlowsheetBlock(default={"dynamic": False})
       m.fs.properties = props.NaClParameterBlock()

       # Add an RO unit to the flowsheet.
       m.fs.unit = ReverseOsmosis0D(default={"property_package": m.fs.properties})
       
       # Specify system variables.
       m.fs.unit.inlet.flow_mass_phase_comp[0, 'Liq', 'NaCl'].fix(0.035)  # mass flow rate of NaCl
       m.fs.unit.inlet.flow_mass_phase_comp[0, 'Liq', 'H2O'].fix(0.965)  # mass flow rate of water
       m.fs.unit.inlet.pressure[0].fix(50e5)  # feed pressure
       m.fs.unit.inlet.temperature[0].fix(298.15)  # feed temperature
       m.fs.unit.area.fix(50)  # membrane area
       m.fs.unit.A_comp.fix(4.2e-12)  # membrane water permeability
       m.fs.unit.B_comp.fix(3.5e-8)  # membrane salt permeability
       m.fs.unit.permeate.pressure[0].fix(101325)  # permeate pressure
       
       # Set scaling factors for component mass flowrates.
       m.fs.properties.set_default_scaling('flow_mass_phase_comp', 1, index=('Liq', 'H2O'))
       m.fs.properties.set_default_scaling('flow_mass_phase_comp', 1e2, index=('Liq', 'NaCl'))
       
       # Calculate scaling factors.
       calculate_scaling_factors(m)
       
       # Initialize the model passing default solver options 
       m.fs.unit.initialize(optarg=solver.options)
       
       # Solve the model (using the tee=True option to display solver info)
       results = solver.solve(m, tee=True)

   # Call that function in the "__main__" for the script 
   if __name__ == "__main__":
       main()
```

aladshaw3 (changes requested):
I think we can keep the simple examples you have and just add another simple example of the 'proper' way to write the script. Here, I show the proper format for the RO unit.

docs/how_to_guides/how_to_run_models_in_a_py_script.rst
lbianchi-lbl: Ooh, I really like this. I'll add this as-is as the "good example", and the same code without the `main()` function for the "bad example".

lbianchi-lbl:
@aladshaw3 I've changed the examples to use the code you suggested.

The only change I made is moving the call to `get_solver()` to within the `main()` function (since that's what the rest of the guide prescribes); I'm not sure it's strictly required (i.e. whether it would cause the same sort of errors on Windows as #387), but it's more consistent. Let me know if you can think of any drawbacks for this.

**EDIT** I'm sorry, I've read your review comment more carefully, and realized you were suggesting to have the more realistic example *in addition* to the simple/trivial examples, whereas I've changed to using it *instead* of the simple examples. Let me know if you think the current (i.e. c431b1b) is still OK, otherwise I'll add the simpler examples back.

aladshaw3 (approved):
Current version (i.e. https://github.com/watertap-org/watertap/commit/c431b1ba129a20eb39e03deefd8937bb8d86cd06) is fine to me. So long as it shows a good and bad example for this, I think it is fine. 

You may need to 'kick' github actions to re-run the checks. It seems like it is stuck waiting for something.

ksbeattie (approved):
LGTM

PR #457, opened 2022-03-25
Municipal treatment case

TimBartholomew:
## Summary/Motivation:

- WaterTAP should include a few demonstration of using the zero order models, this PR includes a demonstration for municipal wastewater treatment

- Creates a flowsheet for municipal wastewater treatment
- Updates bounds on parameters in fixed_bed, which were returning a warning because the default value was outside the bounds
- Adds initial values and bounds to variables is ozone_zo

### Legal Acknowledgement


TimBartholomew:
There are a few things to note:
- This PR doesn't include costing
- I use the raw water pump and unfix the specific energy consumption, so that I can fix the power for the three pumps. While this works, the costing data for the pumps will likely not line up with WT3. We may want to consider implementing the WT3 "water_pumping_station"
- The ozone_zo model for some reason is unstable. It requires me to initialize all of the models before a solve, which is usually not necessary for systems entirely composed zero order models. If I skipped the ozone model, I could solve without initialization. Interestingly, I couldn't just initialize the ozone model, I had to initialize all the units leading up to it.
- The initialization method for zero order unit models fails because the stateblock initialization doesn't have any constraints (since most models don't request any properties)

TimBartholomew:
@andrewlee94 @kurbansitterley @adam-a-a I think this is ready to be merged, only thing left is costing and that can be a separate PR

watertap/examples/case_studies/municipal_treatment/municipal_treatment.py
```diff
@@ -0,0 +1,301 @@
...
+    m.fs.chlorination.load_parameters_from_database(use_default_removal=True)
+
+    # storage
+    m.db.get_unit_operation_parameters("storage_tank")
+    m.fs.storage.load_parameters_from_database(use_default_removal=True)
+    m.fs.storage.storage_time.fix(6)
+
+    # recharge pump
+    m.fs.recharge_pump.load_parameters_from_database()
+    m.fs.recharge_pump.electricity.fix(250)
```
adam-a-a: If you want to match the case study, I believe these numbers were originally in HP and should be converted to kW (for all three pumping stations).

watertap/examples/case_studies/municipal_treatment/municipal_treatment.py
```diff
@@ -0,0 +1,301 @@
...
+    conc_mass_toc = 0.004 * pyunits.kg / pyunits.m ** 3
+
+    m.fs.feed.flow_vol[0].fix(flow_vol)
+    m.fs.feed.conc_mass_comp[0, "tds"].fix(conc_mass_tds)
+    m.fs.feed.conc_mass_comp[0, "tss"].fix(conc_mass_tss)
+    m.fs.feed.conc_mass_comp[0, "toc"].fix(conc_mass_toc)
+    solve(m.fs.feed)
+
+    # intake pump
+    m.db.get_unit_operation_parameters("water_pumping_station")
```
andrewlee94: Why are you calling `get_unit_operation_parameters` here? This is done automatically as part of `load_parameters_from_database` so all of these lines are unnecessary.

watertap/examples/case_studies/municipal_treatment/tests/test_municipal_treatment.py
```diff
@@ -0,0 +1,128 @@
...
+from watertap.examples.case_studies.municipal_treatment.municipal_treatment import main
+from idaes.core.util import get_solver
+
+# -----------------------------------------------------------------------------
+@pytest.mark.component
+def test_municipal_treatment():
+    m = main()
+
+    report_io = StringIO()
+    m.fs.feed.report(ostream=report_io)
```
andrewlee94: I don't think you should be testing the string outputs here - these are rather fragile and we only do them in the unit models tests to make sure the method is functional. At the flowsheet level, you would be much better off just testing the expected value of the key variables with `pytest.approx`.

watertap/examples/case_studies/municipal_treatment/tests/test_municipal_treatment.py
```diff
@@ -0,0 +1,128 @@
...
+    Mass Concentration H2O    999.36
+    Mass Concentration tds   0.63000
+    Mass Concentration toc 0.0040000
+    Mass Concentration tss 0.0065250
+    Volumetric Flowrate      0.92240
+====================================================================================
+"""
+    assert output == report_io.getvalue()
+
+    report_io = StringIO()
```
andrewlee94: Not really important following my previous comment, but you do not need to create a new `StringIO` for each step - you can just use the previous one and the new output should be appended. The only change you would need is to make the final check an `in` rather than `==`.

watertap/examples/case_studies/municipal_treatment/tests/test_municipal_treatment.py
```diff
@@ -0,0 +1,128 @@
...
+    Volumetric Flowrate       0.88491    0.88491
+    Mass Concentration H2O     999.93     999.93
+    Mass Concentration tds   0.065669   0.065669
+    Mass Concentration tss 1.8719e-06 1.8719e-06
+    Mass Concentration toc  0.0010710  0.0010710
+====================================================================================
+"""
+    assert output == report_io.getvalue()
+#
+#     assert value(m.LCOW) == pytest.approx(VALUE, rel=1e-3) # TODO: add test for LCOW
```
andrewlee94: You really need tests for value rather than string output here. Also, there is no test to ensure the model actually solved here - this would probably require the `main` method to return the `results` object in addition to the model (or append the results to the model).

watertap/unit_models/zero_order/fixed_bed_zo.py
```diff
@@ -115,7 +115,7 @@ def ferric_chloride_demand_equation(b, t):
             doc="Pre-exponential factor for activated carbon demand")
         self.activated_carbon_parameter_b = Var(
             units=pyunits.dimensionless,
-            bounds=(0, None),
+            bounds=(None, None),
```
andrewlee94, line 118: Why are you changing this bound? Are there really any cases where this could be negative?

andrewlee94 (changes requested):
A few general comments, but the main ones are the need for better testing and why you changed the bounds on some fixed Vars.

watertap/examples/case_studies/municipal_treatment/municipal_treatment.py
TimBartholomew: Ok, I saw this before every `load_parameters_from_database()` in the tests, but I see that it was only to get the data object to do other things. The commercial seawater desalination case will need to be updated as well.

watertap/unit_models/zero_order/fixed_bed_zo.py
TimBartholomew: Yes all of these parameters are negative in the yamls.

watertap/examples/case_studies/municipal_treatment/tests/test_municipal_treatment.py
TimBartholomew: the solve in main asserts the optimal termination

andrewlee94: Whilst that probably works, it is not explicitly part of the tests - i.e. it relies on the assertion in the main file being there, and there is no way for the test to know for sure. I still think the test should explicitly test for that, even if the main file does it as well.

watertap/examples/case_studies/municipal_treatment/tests/test_municipal_treatment.py
TimBartholomew: I got rid of the string comparisons

watertap/examples/case_studies/municipal_treatment/tests/test_municipal_treatment.py
TimBartholomew: I got rid of the report comparisons.

watertap/unit_models/zero_order/fixed_bed_zo.py
```diff
@@ -211,7 +211,7 @@ def anthracite_demand_equation(b, t):
             doc="Pre-exponential factor for cationic polymer demand")
         self.cationic_polymer_parameter_b = Var(
             units=pyunits.dimensionless,
-            bounds=(0, None),
+            bounds=(None, None),
```
adam-a-a, line 214: Just a _minor_ note- I guess you could just remove these bounds all together.

watertap/examples/case_studies/municipal_treatment/municipal_treatment.py
```diff
@@ -63,17 +63,17 @@ def main():
...
+    results = solve(m)
     display_results(m)
 
     # TODO: add costing, currently costing functions only pass
     add_costing(m)
     initialize_costing(m)
     assert_degrees_of_freedom(m, 0)
 
-    solve(m)
+    # results = solve(m)
```
andrewlee94, line 74: I see a couple of issues here - first, you are not doing a final solve of the model with costing, and secondly you have left this as a commented line.

To me, this really implies that you should write two separate tests - one before costing and one after, each with their own checks for optimal solution.

TimBartholomew: Costing is currently not added, each of the functions just `pass`. They are called so that codecov will pass and I expect costing to be implemented soon. We can break apart the costing tests later, when it is implemented.

andrewlee94 (approved):

adam-a-a (approved):
LGTM

PR #459, opened 2022-03-25
WT3 surface discharge ZO model

tarnold17:
## Summary/Motivation:
Migrate surface discharge ZO model from WT3

- Add surface discharge ZO model and unit tests
- Update YAML

### Legal Acknowledgement


watertap/core/zero_order_costing.py
```diff
@@ -939,6 +940,49 @@ def cost_storage_tank(blk):
...
+            initialize=1,
+            units=blk.config.flowsheet_costing_block.base_currency,
+            bounds=(0, None),
+            doc="Capital cost of unit operation")
+
+        Q = pyo.units.convert(
+            blk.unit_model.properties[t0].flow_vol,
+            to_units=pyo.units.m ** 3 / pyo.units.hour)
+        expr = pyo.units.convert(
+            A * (Q / ref_state) ** B
```
andrewlee94: In this case, what you really want ot do is `pyo.units.convert(flow_vol/ref_state, to_units=pyo.units.dimensionless)`.

This allows for future cases where the units of `ref_state` are not known.

watertap/data/techno_economic/surface_discharge.yaml
```diff
@@ -1,9 +1,49 @@
...
+    units: miles
+emwd:
+  capital_cost:
+    basis: flow_vol
+    cost_factor: None
+    reference_state:
+      value: 10417
+      units: m^3/hr
+    capital_a_parameter:
+      value: 35e6 # TODO does this need to be scaled to account for $ vs MM$ ?
```
andrewlee94: Most $ values in the WT3 code were for millions of dollars ,so we generally have to multiply by 1e6 in our YAML files (alternatively, you could set the cost units as `MUSD_2020`, but we have generally just used 1e6).

watertap/unit_models/zero_order/surface_discharge_zo.py
```diff
@@ -0,0 +1,56 @@
...
+                                 doc='Piping distance')
+
+        self.pipe_diameter = Var(self.flowsheet().config.time,
+                                 units=pyunits.inches,
+                                 doc='Pipe diameter')
+
+        self._fixed_perf_vars.append(self.pipe_distance)
+        self._fixed_perf_vars.append(self.pipe_diameter)
+
+        self._perf_var_dict["Pipe Distance (miles)"] = self.pipe_distance
```
andrewlee94: We generally do not put units in here, as we will (eventually) report units with the values in the report output. and thus these hard-coded units might not be correct.

watertap/unit_models/zero_order/tests/test_surface_discharge.py
```diff
@@ -0,0 +1,330 @@
...
+
+class TestSurfaceDischargeZO:
+    @pytest.fixture(scope="class")
+    def model(self):
+        m = ConcreteModel()
+        m.db = Database()
+
+        m.fs = FlowsheetBlock(default={"dynamic": False})
+        m.fs.params = WaterParameterBlock(
+                default={"solute_list": ["toc", "nitrate", "sulfate", "bar", "crux"]})
```
andrewlee94: What is `crux` in this case, and why is it needed?

andrewlee94 (changes requested):
A couple of minor comments, but otherwise looks good. Well done for your first model.

watertap/core/zero_order_costing.py
tarnold17: I'm not sure that I precisely follow your suggestion here - are you saying that I need to wrap `(Q / ref_state)` in the call to `pyo.units.convert()` that you suggested? I.e., I should have two calls to `pyo.units.convert()`, one nested inside the other?

watertap/data/techno_economic/surface_discharge.yaml
tarnold17: Is there an action required here? The A parameter that I found in WT3 was 35 [(see here)](https://github.com/NREL/WaterTAP3/blob/master/watertap3/watertap3/wt_units/surface_discharge.py) so I have multiplied by 1e6.

watertap/unit_models/zero_order/surface_discharge_zo.py
tarnold17: So these strings should simply read `"Pipe Distance"` and `"Pipe Diameter"`? I can change that. This is something that I just copied and pasted from the well field ZO model.

watertap/unit_models/zero_order/tests/test_surface_discharge.py
tarnold17: @kurbansitterley told me that it is just meant to be a dummy component. This is another instance of copy/paste from the well field ZO model. I can remove all references to "crux" if you'd like.

watertap/core/zero_order_costing.py
andrewlee94: Yes, you can nest convert statements as needed. So:

```
expr = pyo.units.convert(
            A * pyo.units.convert(Q / ref_state, to_units=pyo.units.dimensionless)**B,
            to_units=base_currency)
```

watertap/data/techno_economic/surface_discharge.yaml
andrewlee94: No, I don't think so. I just wanted to clarify,

watertap/unit_models/zero_order/surface_discharge_zo.py
andrewlee94: I think that is better. It is on the list of thing to do in IDAES to add units automatically to these reports.

watertap/unit_models/zero_order/tests/test_surface_discharge.py
andrewlee94: You already have `"bar"` for a dummy component in this case however.

tarnold17: True. I will remove "crux" on my next update.

luohezhiming (dismissed):
Looks good to me.

watertap/core/zero_order_costing.py
tarnold17: I addressed this on my most recent update, so I am marking resolved here.

watertap/unit_models/zero_order/surface_discharge_zo.py
tarnold17: I addressed this on my most recent update, so I am marking resolved here.

watertap/unit_models/zero_order/tests/test_surface_discharge.py
tarnold17: I addressed this on my most recent update, so I am marking resolved here.

watertap/core/zero_order_costing.py
```diff
@@ -972,7 +972,7 @@ def cost_surface_discharge(blk):
             blk.unit_model.properties[t0].flow_vol,
             to_units=pyo.units.m ** 3 / pyo.units.hour)
         expr = pyo.units.convert(
-            A * (Q / ref_state) ** B
+            A * pyo.units.convert(Q / ref_state, to_units=pyo.units.dimensionless) ** B
```
andrewlee94: By doing this, you no longer need the unit conversion for `Q` above - you can just use `blk.unit_model.properties[t0].flow_vol` here.

andrewlee94 (commented):
One more small comment to fix.

watertap/core/zero_order_costing.py
```diff
@@ -939,6 +940,46 @@ def cost_storage_tank(blk):
...
+        # Add cost variable and constraint
+        blk.capital_cost = pyo.Var(
+            initialize=1,
+            units=blk.config.flowsheet_costing_block.base_currency,
+            bounds=(0, None),
+            doc="Capital cost of unit operation")
+
+        expr = pyo.units.convert(
+            A * pyo.units.convert(blk.unit_model.properties[t0].flow_vol / ref_state, to_units=pyo.units.dimensionless) ** B
+            + (pipe_cost_basis * blk.unit_model.pipe_distance[t0] * blk.unit_model.pipe_diameter[t0]),
```
andrewlee94: You should probably separate the unit conversion to be around each part of the sum here. I.e:

```
expr = (
    pyo.units.convert(
        A * pyo.units.convert(blk.unit_model.properties[t0].flow_vol / ref_state, to_units=pyo.units.dimensionless) ** B,
        to_units=blk.config.flowsheet_costing_block.base_currency) +
    pyo.units.convert(
        pipe_cost_basis * blk.unit_model.pipe_distance[t0] * blk.unit_model.pipe_diameter[t0]),
        to_units=blk.config.flowsheet_costing_block.base_currency))
```

The reason for this is that Pyomo will object if the two parts of the sum do not have same units before conversion. This way, we force each part of the sum to have the same units before doing the addition.

watertap/data/techno_economic/surface_discharge.yaml
andrewlee94: You should probably remove this comment no that this is resolved.

andrewlee94 (commented):
Two more small things to fix.

andrewlee94 (approved):

luohezhiming (approved):

PR #460, opened 2022-03-26
Add Water Pumping Station

adam-a-a:
## Summary/Motivation:
Add water pumping station to support PR#457 and provide ability that @TimBartholomew asked for.

- add unit, yaml, and tests
- add config option "fix_pump_power" which is set to True by default.

### Legal Acknowledgement

### DIFF
--- a/watertap/data/techno_economic/tests/test_unit_parameter_files.py
+++ b/watertap/data/techno_economic/tests/test_unit_parameter_files.py
@@ -63,6 +63,7 @@ def test_unit_parameter_files(tech):
                     "smp",
                     "sw_onshore_intake",
                     "tramp_oil_tank",
+                    "water_pumping_station"
                    ]
 
     siso_full_recovery = ["uv_aop", "uv", "fixed_bed", "decarbonator", "chlorination"]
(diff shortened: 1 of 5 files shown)


TimBartholomew (dismissed):
LGTM, I'm adding an issue to our master list for after March that we may want to consider reorganizing all of the zero-order pump models

watertap/data/techno_economic/tests/test_unit_parameter_files.py
```diff
@@ -63,6 +63,7 @@ def test_unit_parameter_files(tech):
                     "smp",
                     "sw_onshore_intake",
                     "tramp_oil_tank",
+                    "water_pumping_station"
```
bknueven:
```suggestion
                    "water_pumping_station",
```

watertap/data/techno_economic/tests/test_unit_parameter_files.py
```diff
@@ -114,7 +115,8 @@ def test_unit_parameter_files(tech):
                                          "agglom_stacking",
                                          "storage_tank",
                                          "evaporation_pond",
-                                         "deep_well_injection"]
+                                         "deep_well_injection",
+                                         "water_pumping_station"]
```
bknueven:
```suggestion
                                         "water_pumping_station",
                                         ]
```

bknueven (dismissed):
This looks reasonable.

I would strongly encourage the use of trailing commas on list-of-strings to avoid potential bugs. Note that:
```
["foo"
 "bar"]
```
is *not* a syntax error -- it results in `["foobar"]`, as opposed to `["foo", "bar"]`, which is likely what was intended. However, this mistake can be easy to make by forgetting to add a comma to the previously last element.

bknueven (approved):

luohezhiming (approved):

PR #463, opened 2022-03-28
MABR and DMBR ZO

luohezhiming:
## Summary/Motivation:

Add mabr and dmbr models for 1598_Skerlos case study

- add mabr and dmbr model
- add corresponding tests

### Legal Acknowledgement

### DIFF
--- a/watertap/core/zero_order_costing.py
+++ b/watertap/core/zero_order_costing.py
@@ -31,9 +31,11 @@ from watertap.unit_models.zero_order import (
     ChlorinationZO,
     CoagulationFlocculationZO,
     DeepWellInjectionZO,
+    DMBRZO,
     FixedBedZO,
     GACZO,
     LandfillZO,
+    MABRZO,
     IonExchangeZO,
     IronManganeseRemovalZO,
     OzoneZO,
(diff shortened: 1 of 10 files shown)


watertap/core/zero_order_costing.py
```diff
@@ -661,6 +663,58 @@ def cost_deep_well_injection(blk):
...
+        blk.capital_cost = pyo.Var(
+            initialize=1,
+            units=blk.config.flowsheet_costing_block.base_currency,
+            bounds=(0, None),
+            doc="Capital cost of unit operation")
+
+        Q_treated = pyo.units.convert(blk.unit_model.properties_treated[t0].flow_vol,
+                              to_units=pyo.units.L / pyo.units.hour)
+
+        ES_reactor = (1/pyo.units.convert(A, to_units=pyo.units.L / pyo.units.m**2 / pyo.units.hour))*Q_treated
```
andrewlee94: A minor comment, but I think you could avoid a unit conversion by replacing lines 696 to 699:

```
ES_reactor = pyo.units.convert(blk.unit_model.properties_treated[t0].flow_vol/A, to_units=pyo.units.m**2)
```

watertap/core/zero_order_costing.py
```diff
@@ -661,6 +663,58 @@ def cost_deep_well_injection(blk):
...
+
+        expr = pyo.units.convert(ES_reactor*B, to_units=blk.config.flowsheet_costing_block.base_currency)
```
andrewlee94: You could save another unit conversion here by combining this with the one above.

```
expr = pyo.units.convert(
    blk.unit_model.properties_treated[t0].flow_vol/A*B,
    to_units=blk.config.flowsheet_costing_block.base_currency)
```

watertap/core/zero_order_costing.py
```diff
@@ -752,6 +806,72 @@ def cost_gac(blk):
...
+
+        SV_reactor = Q_treated*ammonium_reacted
+
+        ES_reactor = (1/pyo.units.convert(A, to_units=pyo.units.kg / pyo.units.m**2 / pyo.units.hour))*SV_reactor
+
+        DCC_reactor = pyo.units.convert(ES_reactor*B, to_units=blk.config.flowsheet_costing_block.base_currency)
+
+        ES_blower = C*pyo.units.convert(blk.unit_model.blower_size, to_units=pyo.units.m ** 2)
+
+        DCC_blower = pyo.units.convert(ES_blower*D, to_units=blk.config.flowsheet_costing_block.base_currency)
```
andrewlee94: As above, you can save some unit conversion operations by combining many of these. In the end, I think you should only need two unit conversions.

watertap/examples/case_studies/wastewater_resource_recovery/case_1598.yaml
```diff
@@ -0,0 +1,48 @@
...
+    units: dimensionless
+  TPEC:
+    # Total Purchased Equipment Cost
+    value: 3.4
+    units: dimensionless
+  TIC:
+    # Total Installed Cost
+    value: 1.65
+    units: dimensionless
+# currency_definitions:
```
andrewlee94: You should remove this comment as it is no longer needed. I am guessing this is also in the default file as well, so it should be removed there too.

andrewlee94 (changes requested):
A couple of minor comments, most about reducing overhead by combining expressions to reduce the number of unit conversion calls.

luohezhiming:
> A couple of minor comments, most about reducing overhead by combining expressions to reduce the number of unit conversion calls.

Thank you for all your suggestions, I've refined the code accordingly to avoid multiple unit conversions.

andrewlee94 (dismissed):

adam-a-a (dismissed):
LGTM. Just a quick note: there are a few comments that I originally entered in the yaml files that were kept in here (some of them still apply, while others may not). We can revisit/cleanup at a later point though.

luohezhiming:
> LGTM. Just a quick note: there are a few comments that I originally entered in the yaml files that were kept in here (some of them still apply, while others may not). We can revisit/cleanup at a later point though.

@andrewlee94 @adam-a-a I just cleaned up those deprecated comments in the yaml file.

andrewlee94 (approved):

adam-a-a (approved):
LGTM

PR #465, opened 2022-03-29
Initial Nanofiltration DSPM-DE (0D) with modified ion property model

adam-a-a:
## Summary/Motivation:
The Donnan Steric Pore Model with Dielectric Exclusion (DSPM-DE) enables us to model how multiple ions in a given feed composition will each be rejected by a nanofiltration membrane. The model accounts for diffusion, convection, and electromigration across the membrane. 

This PR presents the initial implementation DSPM-DE 0D model and modifications to the associated ion property model that can be provided any set of ions.  

All _key_ model equations and variables have been implemented, and the model initializes and solves. Note, there are also several #TODOs scattered throughout files. However, this PR is meant to serve only as the initial version of the model, and the plan is to thoroughly vet the model through a validation and refinement phase (outlined in issue#464). 

Developers can follow these slides as a reference for variables and constraints to aid in review (https://app.box.com/s/9juyc1n4qd1r98lutpaw2qyhyob5pzs1).

- addition of DSPM-DE 0D model with tests
- modification of ion DSPM-DE property model with tests

### Legal Acknowledgement


watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -53,6 +57,11 @@ class ActivityCoefficientModel(Enum):
     ideal = auto()                    # Ideal
```
andrewlee94, line 57: Out of interest, why use `auto()` here rather than assigning a nominal value?

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -208,7 +283,7 @@ class _DSPMDEStateBlock(StateBlock):
     whole, rather than individual elements of indexed Property Blocks.
     """
 
-    def initialize(self, state_args=None, state_vars_fixed=False,
+    def initialize(blk, state_args=None, state_vars_fixed=False,
```
andrewlee94: Minor comment, but `self` is probably the more Pythonic name to use here. Using `blk` will likely cause some issues if/when we start running `pylint` as part of the tests.

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -576,18 +682,63 @@ def _act_coeff_phase_comp(self):
...
             elif b.params.config.activity_coefficient_model == ActivityCoefficientModel.davies:
-                raise NotImplementedError(f"Davies model has not been implemented yet.")
+                I = b.ionic_strength
+                return (log(b.act_coeff_phase_comp['Liq', j]) == -b.debye_huckel_constant
+                        * b.charge_comp[j] ** 2
+                        * (I**0.5
+                        / (1 * pyunits.mole ** 0.5 / pyunits.kg ** 0.5
+                           + I ** 0.5)
+                           - b.params.debye_huckel_b * I))
+                # raise NotImplementedError(f"Davies model has not been implemented yet.")
```
andrewlee94: Is this comment still needed?

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -761,6 +938,31 @@ def calculate_scaling_factors(self):
...
                             sf_dens * iscale.get_scaling_factor(self.mass_frac_phase_comp['Liq', j],default=1,warning=True))
 
+        if self.is_property_constructed('conc_mol_phase_comp'):
+            for j in self.params.component_list:
+                if iscale.get_scaling_factor(self.conc_mol_phase_comp['Liq', j]) is None:
+                    sf = (iscale.get_scaling_factor(self.conc_mass_phase_comp['Liq', j])
+                          / iscale.get_scaling_factor(self.mw_comp[j]))
+                    iscale.set_scaling_factor(self.conc_mol_phase_comp['Liq', j], sf)
+
+
```
andrewlee94: Extra line breaks.

watertap/unit_models/tests/test_nanofiltration_DSPMDE_0D.py
```diff
@@ -0,0 +1,310 @@
...
+        assert len(unscaled_var_list) == 0
+
+        # not all constraints have scaling factor so skipping the check for unscaled constraints
+
+    @pytest.mark.requires_idaes_solver
+    @pytest.mark.component
+    def test_initialize(self, NF_frame):
+        m= NF_frame
+        initialization_tester(m)
+        # m.fs.unit.initialize(automate_rescale=True)  # this is the default for NFDSPM initialize but writing out anyway
```
andrewlee94: Commented line,

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -0,0 +1,1394 @@
...
+        self.scaling_factor = Suffix(direction=Suffix.EXPORT)
+
+        # For stateblock-specific scaling in calculate_scaling_factors
+        self._sb_scaled_properties = ComponentSet()
+
+        units_meta = self.config.property_package.get_metadata().get_derived_units
+
+        self.io_list = io_list = Set(initialize=[0, 1])  # inlet/outlet set
+
+        self._process_config()
```
andrewlee94: A minor comment, but I would do this at the earliest opportunity possible (in this case immediately after `build()`?).

It is a minor thing, but you might as well fail as early as possible to avoid wasting cycles.

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -0,0 +1,1394 @@
...
+
+        if hasattr(self.config.property_package,'ion_set'):
+            solute_set = self.config.property_package.ion_set
+        elif hasattr(self.config.property_package,'solute_set'):
+            solute_set = self.config.property_package.solute_set
```
andrewlee94, line 197: You probably need a final case here for what to do if neither of the above is true (probably raise an Exception).

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -0,0 +1,1394 @@
...
+            solver=solver,
+            state_args=state_args['interface_in'],)
+        blk.pore_exit.initialize(
+            outlvl=outlvl,
+            optarg=optarg,
+            solver=solver,
+            state_args=state_args['permeate'],)
+        init_log.info_high("Initialization Step 2 Complete.")
+
+        # # Double-check for poorly scaled variables after state block initialization
```
andrewlee94: Double comment marks through a lot of this.

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -0,0 +1,1394 @@
...
+            res = opt.solve(blk, tee=slc.tee)
+            if not check_optimal_termination(res):
+                init_log.warn("Trouble solving NanofiltrationDSPMDE0D unit model with deactivated constraint.")
+        # # ---------------------------------------------------------------------
+        # # Solve unit
+        blk.eq_solute_solvent_flux.activate()
+        with idaeslog.solver_log(solve_log, idaeslog.DEBUG) as slc:
+            res = opt.solve(blk, tee=slc.tee)
+            if not check_optimal_termination(res):
+                init_log.warn("Trouble solving NanofiltrationDSPMDE0D unit model. Trying one more time.")
```
andrewlee94, line 1043: You should raise an `InitializationError` here instead. Without an exception, it is easy for this to pass unnoticed, whereas with an exception the user at least needs to acknowledge it with a `try/except` (which could just `pass` on an exception indicating they are happy to continue with the failing step).

IDAES is try to make it standard for all models to raise an `InitializationError` in these cases.

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -0,0 +1,1394 @@
...
+                    * (1 / b.feed_side.properties_in[t].dielectric_constant - 1 / b.dielectric_constant_pore[t]))
+
+        @self.Expression(self.flowsheet().config.time,
+                         solute_set,
+                         doc="Born solvation contribution to partitioning")
+        def partition_factor_born_solvation_comp(b, t, j):
+            return (-b.gibbs_solvation_comp[t, j]
+                    / (Constants.boltzmann_constant * b.feed_side.properties_in[t].temperature))
+
+    def initialize(
```
andrewlee94: In order to facilitate initialization of costing methods, IDAES has changed how initialization is called. There is now a common, centralized `initialize` method in the base classes which calls to a model specific `initialize_build` method (i.e. this method). This way, the central method can handle determining what to do with "add-ons" like costing, and also handles the final solve of the full model (including raising an `InitializationError` if the final step fails).

Rather than overload `initialize`, you should instead just rename this to `initialize_build` and everything should still work.

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -0,0 +1,1394 @@
...
+                "Permeate Outlet": self.permeate,
+            },
+            time_point=time_point,
+        )
+
+    def get_costing(self, module=None, **kwargs):
+        self.costing = Block()
+        module.Nanofiltration_costing(self.costing, **kwargs)
+
+    def _get_state_args(self, source, mixed_permeate_properties, initialize_guess, state_args):
```
andrewlee94, line 1196: Could you comment on the purpose of this method (it probably should be included in the doc string).

andrewlee94 (commented):
I can't speak to the details of the model itself, but I had a number of comments on the code.

watertap/property_models/ion_DSPMDE_prop_pack.py
adam-a-a: I suppose it's more a matter of convenience, since this will also assign unique integer values without having to manually assign them.

watertap/unit_models/nanofiltration_DSPMDE_0D.py
adam-a-a: I added something to describe this method. I adopted this from what @bknueven did for RO; it returns `state_args` for each of the stateblocks in the unit model which subsequently get passed to those stateblocks during initialization.

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -160,13 +187,58 @@ def build(self):
...
+            doc='Mass density parameter B3')
+        self.dens_mass_param_B4 = Var(
+            within=Reals, initialize=-3.060e-5, units=dens_units * t_inv_units ** 3,
+            doc='Mass density parameter B4')
+        self.dens_mass_param_B5 = Var(
+            within=Reals, initialize=-1.613e-5, units=dens_units * t_inv_units ** 2,
+            doc='Mass density parameter B5')
+
+        for v in self.component_objects(Var):
+            v.fix()
```
bknueven, line 234: Note that this will *also* fix any `Var`s created by the call to `super().build()` above, or, alternatively, all the `Var`s currently on the block.

Probably doesn't matter in this case, but we might consider being more explicit here.

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -440,7 +517,7 @@ def _mass_frac_phase_comp(self):
             self.params.phase_list,
             self.params.component_list,
             initialize=lambda b,p,j : 0.4037 if j == "H2O" else 0.0033, #todo: revisit
-            bounds=(1e-6, 1.001),
+            bounds=(1e-8, 1.001),
```
bknueven, line 521: This is more of comment on how we bound variables like this in general: why over-relax the upper bound but under-relax the lower bound?

If we're trying to avoid 0, I think we should settle on a standard epsilon for the codebase. Sometimes these bounds become binding unexpectedly.

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -640,21 +790,68 @@ def define_state_vars(self):
...
+        else:
+            for j in self.params.solute_set:
+                if self.flow_mol_phase_comp['Liq', j].is_fixed():
+                    raise AssertionError(
+                        f"{self.flow_mol_phase_comp['Liq', j]} was fixed. Either set defined_state=True or unfix "
+                        f"flow_mol_phase_comp for each solute to check that electroneutrality is satisfied.")
+
+        if get_property is not None:
+            for i in [get_property]:
+                getattr(self, i)
```
bknueven:
```suggestion
            getattr(self, get_property)
```

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -640,21 +790,68 @@ def define_state_vars(self):
...
+        self.conc_mol_phase_comp
```
bknueven, line 822: We might just comment that we're accessing these properties here to ensure their creation. In a "typical" program this and the `getattr` above would be a no-op.

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -640,21 +790,68 @@ def define_state_vars(self):
                 "temperature": self.temperature,
                 "pressure": self.pressure}
 
-    def assert_electroneutrality(self, tol=None, tee=False):
+    def assert_electroneutrality(self, tol=None, tee=True, defined_state=True, adjust_by_ion=None,
+                                 get_property=None, solve=True):
         if tol is None:
-            tol = 1e-6
+            tol = 0
```
bknueven: Is 0 the right default? I worry we'll just be compounding solver error. Maybe `1e-08` or something similar (scaled by the constraint) would be more appropriate?

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -640,21 +790,68 @@ def define_state_vars(self):
...
+                # self.flow_mol_phase_comp['Liq', adjust_by_ion].fix()
+                ion_adjusted = self.flow_mol_phase_comp['Liq', adjust_by_ion].value
+                if defined_state:
+                    self.flow_mol_phase_comp['Liq', adjust_by_ion].fix(ion_adjusted)
+                    msg = f"{adjust_by_ion} was adjusted and flow_mol_phase_comp['Liq',{adjust_by_ion}] was fixed " \
+                          f"to {ion_adjusted}."
+                else:
+                    msg = f"{adjust_by_ion} was adjusted and the value computed for flow_mol_phase_comp['Liq',{adjust_by_ion}]" \
+                          f" is {ion_adjusted}."
+                del self.charge_balance
```
bknueven: It's hard to verify that this constraint gets deleted in every case, and if the logic of this method changes there's the potential for bugs, even if none exist presently.

Maybe we should just check for its existence at the end of the method and delete it if it's still there?

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -0,0 +1,1424 @@
...
+                initialize=0.6,
+                # bounds=(1e2, 5e3),  #TODO:unsure of value ranges at the moment
+                domain=NonNegativeReals,
+                units=units_meta('length'),
+                doc="Characteristic length of spacer")
+        ###############################################################################################################
+        # Expressions
+        ###############################################################################################################
+        # Make expressions that don't depend on any variables
+        self._make_expressions()
```
bknueven, line 428: This is a minor thing, but I'm a bit confused by the existence of this method alongside the Expressions below.

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -0,0 +1,1424 @@
...
+                         doc="Volumetric water flux at inlet and outlet")
+        def flux_vol_water(b, t, x):
+            prop = b.feed_side.properties_in[t]
+            return b.flux_mol_phase_comp[t, x, 'Liq', 'H2O'] * prop.mw_comp['H2O'] / prop.dens_mass_solvent
+
+        # Average Volumetric Water Flux ------------------------------------#
+        @self.Expression(self.flowsheet().config.time,
+                         doc="Average volumetric water flux")
+        def flux_vol_water_avg(b, t):
+            return sum(b.flux_vol_water[t, x] for x in io_list) * 0.5
```
bknueven: Again, a minor thing -- dividing by `len(io_list)` would be a bit more obvious.

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -0,0 +1,1424 @@
...
+
+        # 3. Feed-solution/membrane electroneutrality, DOF=1 *2 for inlet/outlet: DOF= 2
+        @self.Constraint(self.flowsheet().config.time,
+                         io_list,
+                         phase_list,
+                         doc="Electroneutrality at feed-side membrane interface")
+        def eq_electroneutrality_interface(b, t, x, p):
+            return (sum(b.feed_side.properties_interface[t, x].conc_mol_phase_comp[p, j] *
+                    b.feed_side.properties_interface[t, x].charge_comp[j] for j in solute_set)
+                    == b.tol_electroneutrality)
```
bknueven: Since `tol_electroneutrality` is used in an equality constraint like this, I'm not sure "tolerance" is the right word. To me, this would be a proper tolerance if `tol_electroneutrality` were a `Var` with bounds `(-epsilon, epsilon)`. As it-is, it's more like an accumulated error term, and since it's used in a bunch of constraints, the accumulated error is probably different in all of them. Therefore I think having this as a parameter with a value *other than* 0 is a modeling error.

Another way to relax these constraints would be to apply a constraint scaling transform to the electroneutrality constraints, thereby relaxing them a bit (e.g., by scaling them by `1e-4`, for example).

watertap/unit_models/nanofiltration_DSPMDE_0D.py
bknueven: Maybe we should ultimately put this method in a base class?

bknueven (commented):
I have a few comments, most of them minor and can be safely ignored.

My largest concern is the `tol_electroneutrality` parameter in the new nanofiltration model. As-it-is, setting it to anything other than 0 seems like a modeling error.

watertap/property_models/ion_DSPMDE_prop_pack.py
andrewlee94: We probably should codify this somewhere. The lower bound is indeed to avoid potential singularities due to a value of zero so we probably should have a standard epsilon somewhere. The relaxation of the upper bound is because we have  found (by trial and error) that it greatly improves tractability - there probably should be a standard epsilon for this too.

watertap/unit_models/nanofiltration_DSPMDE_0D.py
adam-a-a: This is something I was planning to take out in the next PR.

watertap/unit_models/nanofiltration_DSPMDE_0D.py
adam-a-a: I'm going to clean this up in the next PR. Essentially I was toggling between Expressions and Vars with Constraints while troubleshooting the model.

watertap/property_models/ion_DSPMDE_prop_pack.py
adam-a-a: This is the way we usually do it (e.g., in NaCl and seawater properties) although there is a comment that goes with this: 
`# traditional parameters are the only Vars currently on the block and should be fixed`

watertap/property_models/ion_DSPMDE_prop_pack.py
adam-a-a: I agree.

watertap/property_models/ion_DSPMDE_prop_pack.py
adam-a-a: Just to verify, I don't remember if this is the case or not, but would passing a list directly to `getattr()` work?

bknueven: No, but my suggested code is equivalent. One line 819 we make a single-element list, `[get_property]`, just to iterate over it.

watertap/unit_models/nanofiltration_DSPMDE_0D.py
bknueven: Maybe update the default value to be `0`?

bknueven (dismissed):
I can't really comment on the model itself, but my concerns about the code have been adequately addressed in the discussion.

watertap/property_models/ion_DSPMDE_prop_pack.py
adam-a-a: I'm changing this so that it can take a string or list of strings.

watertap/property_models/ion_DSPMDE_prop_pack.py
adam-a-a: The constraint only gets created if `adjust_by_ion` is set to some solute (as a string). I added a couple tests to verify this, but I think any further refinement on this should be reserved for a subsequent PR.

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -640,21 +791,81 @@ def define_state_vars(self):
...
+            if adjust_by_ion is not None:
+                del self.charge_balance
+                ion_adjusted = self.flow_mol_phase_comp['Liq', adjust_by_ion].value
+                if defined_state:
+                    self.flow_mol_phase_comp['Liq', adjust_by_ion].fix(ion_adjusted)
+                    # touch on-demand property desired
+                    if get_property is not None:
+                        if isinstance(get_property, str):
+                            getattr(self, get_property)
+                        elif isinstance(get_property, list):
```
bknueven:
```suggestion
                        elif isinstance(get_property, (list, tuple)):
```
We might make this slightly more general. But it's also fine as-is.


I guess we could also use a different ask-for-forgiveness pattern:
```python
try:
    getattr(self, get_property)
except AttributeError: #assume a list or other iterator of properties
    for i in get_property:
        getattr(self,i)
except:
    raise TypeError("get_property must be a string or an iterable of strings.")
```

bknueven (dismissed):
I had one more remark on the changed code. But I think this is ready to merge.

bknueven (approved):

TimBartholomew (approved):

PR #467, opened 2022-03-30
Adding costing and minor cleanup of municipal wastewater treatment flowsheet

bknueven:
## Summary/Motivation:
Adds costing to municipal treatment case, follow-on from #457.

- Add costing
- Minor cleanup
- Update `capital_a_parameter` in sedimentation.yaml

## Notes:
~~I have no idea if the reported costing values are reasonable. There's probably more to be done.~~ I think the costing values are now reasonable.

### Legal Acknowledgement

### DIFF
--- a/watertap/data/techno_economic/sedimentation.yaml
+++ b/watertap/data/techno_economic/sedimentation.yaml
@@ -5,7 +5,7 @@ default:
   capital_cost:
     cost_factor: TPEC
     capital_a_parameter:
-      value: 13572e6
+      value: 13572
       units: USD_2007
     capital_b_parameter:
       value: 0.3182
@@ -34,7 +34,7 @@ phosphorus_capture:    # AMO subtype
   capital_cost:
     cost_factor: TPEC
     capital_a_parameter:  # kept from default
-      value: 13572e6
+      value: 13572
       units: USD_2007
     capital_b_parameter:  # kept from default
       value: 0.3182
--- a/watertap/examples/flowsheets/case_studies/municipal_treatment/municipal_treatment.py
+++ b/watertap/examples/flowsheets/case_studies/municipal_treatment/municipal_treatment.py
@@ -15,24 +15,15 @@ from pyomo.environ import (ConcreteModel,
                            TransformationFactory,
                            units as pyunits,
                            assert_optimal_termination,
-                           Block)
-from pyomo.network import Arc
-from pyomo.util.check_units import assert_units_consistent
+                           )
+from pyomo.network import Arc, SequentialDecomposition
 
 from idaes.core import FlowsheetBlock
 from idaes.core.util import get_solver
-from idaes.core.util.initialization import (propagate_state,
-                                            fix_state_vars,
-                                            revert_state_vars)
-from idaes.core.util.exceptions import ConfigurationError
-from idaes.generic_models.unit_models.translator import Translator
-from idaes.generic_models.unit_models import Mixer, Separator, Product
-from idaes.generic_models.unit_models.mixer import MomentumMixingType
+from idaes.generic_models.unit_models import Product
 import idaes.core.util.scaling as iscale
-import idaes.logger as idaeslog
 from idaes.generic_models.costing import UnitModelCostingBlock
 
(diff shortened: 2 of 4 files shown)


watertap/examples/case_studies/municipal_treatment/municipal_treatment.py
```diff
@@ -208,47 +201,10 @@ def set_operating_conditions(m):
...
-            inlet_blk = getattr(unit, unit_inlet_properties_dic[u_str])
-            propagate_state(arc)
-            if unit_inlet_properties_dic[u_str] == 'properties_in':
-                flags = fix_state_vars(inlet_blk)
-                solve(unit)
-                revert_state_vars(inlet_blk, flags)
+    seq = SequentialDecomposition()
+    seq.options.tear_set = []
+    seq.options.iterLim = 1
+    seq.run(m, lambda u: u.initialize())
```
adam-a-a: Just curious- why did we need to include this? I'd think we wouldn't need to go this far for a flowsheet with zero-order models.

bknueven: I'm not sure. Tim had a much more complicated initialization routine which I replaced with this. I didn't try removing it altogether.

And this really isn't complicated ... we should probably write a utility that just does this. It just initializes everything in order (which is fine because this flowsheet is a DAG).

watertap/examples/case_studies/municipal_treatment/municipal_treatment.py
```diff
@@ -264,28 +220,74 @@ def display_results(m):
...
+    for u in m.fs.costing._registered_unit_costing:
+        print(u.name,
+              " :   ",
+              value(pyunits.convert(u.capital_cost,
+                                    to_units=pyunits.USD_2018)))
+    for u in m.fs.costing._registered_unit_costing:
+        print(u.name,
+              " :   ",
+              value(pyunits.convert(u.capital_cost,
+                                    to_units=pyunits.USD_2018)))
```
adam-a-a: duplicate code?

watertap/examples/case_studies/municipal_treatment/tests/test_municipal_treatment.py
```diff
@@ -39,4 +39,5 @@ def test_municipal_treatment():
     assert value(m.fs.recharge_pump.properties[0].flow_mass_comp['toc']) == pytest.approx(9.477e-4, rel=1e-3)
     assert value(m.fs.recharge_pump.properties[0].flow_mass_comp['tss']) == pytest.approx(1.656e-6, rel=1e-3)
 
-    # assert value(m.LCOW) == pytest.approx(VALUE, rel=1e-3)  # TODO: add test for LCOW
+    assert value(m.fs.costing.LCOW) == pytest.approx(5.1533e-7, rel=1e-3) # in M$/m**3
```
adam-a-a: Does that mean LCOW is displayed in M$/m3 in display_costing? Shouldn't hold anything up, but this would be strange to see for the user initially since LCOW is always reported in $/m3 (and usually never exceeds a few dollars per m3 in extreme cases).

adam-a-a (dismissed):
Just a couple of _minor_ comments, but looks good to me.

watertap/examples/case_studies/municipal_treatment/municipal_treatment.py
bknueven: Yeah -- thanks. I didn't notice because I copied it from the seawater_RO_desalination flowsheet, which uses both costing packages.

watertap/examples/case_studies/municipal_treatment/tests/test_municipal_treatment.py
bknueven: No, I explicitly convert it in `display_costing`. The `Expression` itself is in M$/m^3 -- maybe that should be changed in the zero order costing package.

TimBartholomew (dismissed):
LGTM

TimBartholomew (approved):
LGTM

andrewlee94 (approved):
Looks good to me, but I don't know the details of the process.

PR #477, opened 2022-04-01
Low pressure pump model for AMO

luohezhiming:
## Summary/Motivation:

Develop low pressure pump model for Skerlos case study

- add low pressure model
- add tests

### Legal Acknowledgement


watertap/unit_models/zero_order/low_pressure_pump_zo.py
```diff
@@ -0,0 +1,71 @@
...
+    CONFIG = ZeroOrderBaseData.CONFIG()
+
+    def build(self):
+        super().build()
+
+        self._tech_type = "low_pressure_pump"
+
+        build_pt(self)
+
+        self.lift_height = Var(units=pyunits.feet,
```
adam-a-a: For this pump unit, I think we wanted lift_height to be in meters, not feet.

watertap/unit_models/zero_order/low_pressure_pump_zo.py
```diff
@@ -0,0 +1,71 @@
...
+        self._fixed_perf_vars.append(self.eta_pump)
+        self._fixed_perf_vars.append(self.eta_motor)
+
+        @self.Constraint(self.flowsheet().time,
+                         doc='Constraint for electricity consumption based on '
+                             'pump flowrate.')
+        def electricity_consumption(b, t):
+            q_in = pyunits.convert(b.properties[t].flow_vol,
+                                   to_units=pyunits.gallon / pyunits.minute)
+            A = 3960 * pyunits.gallon * pyunits.foot / pyunits.minute / pyunits.horsepower
```
adam-a-a: You can remove this because we don't want to convert from horsepower this time.

watertap/unit_models/zero_order/low_pressure_pump_zo.py
```diff
@@ -0,0 +1,71 @@
...
+            return b.electricity[t] == pyunits.convert(
+                q_in * self.lift_height / (A * self.eta_pump * self.eta_motor),
```
adam-a-a: here, to calculate pump energy consumption using SI units, you would multiply `lift_height` in meters by density of solution and gravitational acceleration. Density is taken from the property model, and you can use `acceleration_gravity` after importing `from idaes.core.util.constants import Constants` to handle gravitational acceleration.

adam-a-a (changes requested):
For this unit, we'd want pump head (lift_height) in meters, and the corresponding relationship for `electricity` should be modified so that conversion from horsepower is not necessary.

watertap/unit_models/zero_order/low_pressure_pump_zo.py
luohezhiming: I think we need the A for a correct unit conversion. Otherwise there is no unit of energy for the `electricity` (watt). I am not so sure what this 3960 coefficient is, but we still need another coefficient for unit consistency.

adam-a-a: Yes, you would need this conversion factor if we were keeping flow in units of gpm and lift_height in feet, which I am saying we should not. Instead, we'd multiply density * gravitational_acceleration * lift_height, which should work out to be in Pascals, and multiply that by flowrate in m^3/s. That should bring you to Watts, which can then be converted to kW.

adam-a-a:
One more suggestion- I think we should name the unit something else, since the unit calculations don't necessarily limit this to a low-pressure pump. Maybe we can call the unit `PumpElectricityZO`, similar to the name we used for the function that does these calculations (although in different units).

watertap/unit_models/zero_order/low_pressure_pump_zo.py
andrewlee94: In the original case, `A` was a lumped term combining gravity and density in the units WT3 wanted (hence the strange units and the 3960). In this case, we should do what @adam-a-a suggested and do the full calculation instead, which will implicitly capture the correct units.

andrewlee94:
@adam-a-a Actually, can we just use the existing IDAES `Pump` model? It already does all of this for us, and I don't think there should be any issues using the simple zero-order properties in the pump model.

watertap/unit_models/zero_order/low_pressure_pump_zo.py
```diff
@@ -0,0 +1,72 @@
...
+        self._fixed_perf_vars.append(self.lift_height)
+        self._fixed_perf_vars.append(self.eta_pump)
+        self._fixed_perf_vars.append(self.eta_motor)
+
+        @self.Constraint(self.flowsheet().time,
+                         doc='Constraint for electricity consumption based on '
+                             'pump flowrate.')
+        def electricity_consumption(b, t):
+            q_in = pyunits.convert(b.properties[t].flow_vol,
+                                   to_units=pyunits.m ** 3 / pyunits.s)
```
adam-a-a: This shouldn't be necessary, unless we are anticipating a new zero order property model with different units, which I doubt.

watertap/unit_models/zero_order/low_pressure_pump_zo.py
```diff
@@ -0,0 +1,72 @@
...
+            return b.electricity[t] == pyunits.convert(
+                pyunits.convert(b.lift_height, to_units=pyunits.m) *
```
adam-a-a: I don't think we _need_ to convert lift_height to meters here.

luohezhiming:
> One more suggestion- I think we should name the unit something else, since the unit calculations don't necessarily limit this to a low-pressure pump. Maybe we can call the unit `PumpElectricityZO`, similar to the name we used for the function that does these calculations (although in different units).

Renamed to `PumpElectricityZO`.

watertap/unit_models/zero_order/low_pressure_pump_zo.py
luohezhiming: I refined the code, with only one necessary unit conversion.

watertap/unit_models/zero_order/low_pressure_pump_zo.py
luohezhiming: I refined the code, with only one necessary unit conversion.

watertap/unit_models/zero_order/low_pressure_pump_zo.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+    Zero-Order model for a low pressure pump unit operation.
+    """
+
+    CONFIG = ZeroOrderBaseData.CONFIG()
+
+    def build(self):
+        super().build()
+
+        self._tech_type = "low_pressure_pump"
```
adam-a-a: Let's change tech_type to "pump_electricity" too

luohezhiming: Done

andrewlee94 (approved):

adam-a-a (approved):

PR #479, opened 2022-04-04
Photothermal membrane and CANDO+P ZO models

tarnold17:
## Summary/Motivation:

Add ZO photothermal membrane and CANDO+P models needed for 1595 case study.

- Add models and tests

### Legal Acknowledgement


tarnold17 (commented):
Please note that I changed the name of

`watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/case_1598.yaml`

to

`watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/wastewater_default_case_study.yaml`

I did this at @adam-a-a 's suggestion, the idea being that this YAML file that @luohezhiming originally created in PR 463 for the MABR and DMBR models can be reused for other case studies as well, and so should be thusly named.

watertap/data/techno_economic/CANDOP.yaml
```diff
@@ -0,0 +1,13 @@
+default:
+  energy_electric_flow_vol_inlet:
+    value: 0
+    units: kWh/m^3
```
adam-a-a: I think you can remove this since I don't think it is being put to use, unless we anticipate that it will be later and assigned a value. 

EDIT: Just realized this is for CANDO+P, so you probably haven't finished this yet.

watertap/unit_models/zero_order/tests/test_photothermal_membrane_zo.py
```diff
@@ -0,0 +1,181 @@
...
+        data = model.db.get_unit_operation_parameters("photothermal_membrane")
+
+        model.fs.unit.load_parameters_from_database(use_default_removal=True)
+        assert model.fs.unit.recovery_frac_mass_H2O[0].fixed
+        assert model.fs.unit.recovery_frac_mass_H2O[0].value == \
+               data["recovery_frac_mass_H2O"]["value"]
+
+        for (t, j), v in model.fs.unit.removal_frac_mass_solute.items():
+            assert v.fixed
+            if j in data["removal_frac_mass_solute"]:
```
adam-a-a: Should this be `if j not in data["removal_frac_mass_solute"]: `? Although the test will pass because the removal fractions are all 0 for the components you added to the yaml, if one were to update the yaml for any reason and include non-zero values, the test would fail. Moreover, this isn't checking what we're intending to check here, which is that if j is not a listed solute with an associated removal fraction, check that it gets assigned the default value of 0.

watertap/data/techno_economic/CANDOP.yaml
tarnold17: Yep, I just put it in so that the unit tests would pass for now.

watertap/unit_models/zero_order/tests/test_photothermal_membrane_zo.py
tarnold17: You're right, I got it backwards here. Good catch, I just pushed your suggested fix.

watertap/unit_models/zero_order/CANDOP_zo.py
```diff
@@ -0,0 +1,83 @@
...
+        self._tech_type = "CANDO_P"
+
+        build_sido_reactive(self)
+
+        # Create electricity variable
+        self.electricity = Var(self.flowsheet().time,
+                               units=pyunits.kW,
+                               bounds=(0, None),
+                               doc="Electricity consumption of unit")
+        self._perf_var_dict["Electricity Demand"] = self.electricity
```
luohezhiming, line 45: Probably also add the unit to the dictionary:
`self._perf_var_dict["Electricity Demand (kW)"] = self.electricity`

watertap/unit_models/zero_order/CANDOP_zo.py
```diff
@@ -0,0 +1,83 @@
...
+            return b.electricity[t] == (pyunits.convert(b.extent_of_reaction[t, "n_reaction"] * b.electricity_intensity_N, to_units=pyunits.kW))
+
+        # Create oxygen demand variables and constraint. The amount of oxygen
+        # consumed is assumed to be a linear function of the amount of
+        # nitrogen reacted.
+        self.O2_demand = Var(self.flowsheet().time,
+                             units=pyunits.kg/pyunits.s,
+                             bounds=(0, None),
+                             doc="Oxygen demand")
+        self._perf_var_dict["Oxygen Demand"] = self.O2_demand
```
luohezhiming, line 72: Also, provide the unit of oxygen demand will be helpful:
`self._perf_var_dict["Oxygen Demand (kg/s)"] = self.O2_demand`

watertap/unit_models/zero_order/CANDOP_zo.py
```diff
@@ -0,0 +1,83 @@
...
+        self.oxygen_nitrogen_ratio = Var(units=pyunits.dimensionless,
+                                         bounds=(0, None),
+                                         doc="Oxygen consumed / nitrogen reacted ratio")
+        self._fixed_perf_vars.append(self.oxygen_nitrogen_ratio)
+        self._perf_var_dict["Oxygen consumed / nitrogen reacted ratio (mass basis)"] = self.oxygen_nitrogen_ratio
```
luohezhiming, line 77: Probably don't need to report this ratio if this is just an intermediate variable.

luohezhiming (commented):
I think we should provide the unit for the defined variables in the report, other staff looks good.

andrewlee94:
@luohezhiming We have avoided putting units in the report as we generally have no way of knowing what they are (these models are an exception). There is an issue in IDAES to get units automatically added to the report instead.

luohezhiming:
> @luohezhiming We have avoided putting units in the report as we generally have no way of knowing what they are (these models are an exception). There is an issue in IDAES to get units automatically added to the report instead.



> @luohezhiming We have avoided putting units in the report as we generally have no way of knowing what they are (these models are an exception). There is an issue in IDAES to get units automatically added to the report instead.

luohezhiming:
> @luohezhiming We have avoided putting units in the report as we generally have no way of knowing what they are (these models are an exception). There is an issue in IDAES to get units automatically added to the report instead.



> @luohezhiming We have avoided putting units in the report as we generally have no way of knowing what they are (these models are an exception). There is an issue in IDAES to get units automatically added to the report instead.

luohezhiming:
@andrewlee94 OK, thank you for clarification. I am so sorry that I accidently close the PR.

watertap/data/techno_economic/CANDO_P.yaml
```diff
@@ -0,0 +1,51 @@
...
+    value: 0
+    units: dimensionless
+  removal_frac_mass_solute:
+    phosphates:
+      value: 0
+      units: dimensionless
+    nitrogen:
+      value: 0
+      units: dimensionless
+    bcp:
```
adam-a-a: What is bcp?

adam-a-a: I see... bioconcentrated phosphorus. I would rename `bcp` to that name, which is more in line with how we have named other components. 

I'm also not clear on the difference between "phosphates" and "bioconcentrated phosphorus" when considering the reaction that's taking place.

watertap/unit_models/zero_order/CANDOP_zo.py
adam-a-a: It also doesn't hurt to show it. I guess it comes down to preference. For me, I'd prefer to include it.

watertap/data/techno_economic/CANDO_P.yaml
tarnold17: I can rename BCP.

> I'm also not clear on the difference between "phosphates" and "bioconcentrated phosphorus" when considering the reaction that's taking place.

The spreadsheet I have lists phosphorous (liquid) and BCP (solid) as different components. Phosphorous is converted into BCP via reaction. I have actually used the component "phosphates" instead of "phosphorous" in the model because you and @andrewlee94 have cautioned me against adding additional components to the code base if it might not be necessary.

watertap/unit_models/zero_order/CANDOP_zo.py
tarnold17: I included `oxygen_nitrogen_ratio` in the report because I thought it might be useful for the end user to see. I included `electricity_intensity_N` in the report as well for the same reason, at @andrewlee94 's suggestion. My inclination is to keep it as is, I think it's a helpful piece of information to display to the user.

watertap/data/techno_economic/CANDO_P.yaml
adam-a-a: Ah, thanks! I didn't realize it was meant to represent phosphorus in solid phase. That makes sense.

watertap/unit_models/zero_order/photothermal_membrane_zo.py
```diff
@@ -0,0 +1,56 @@
...
+        self._fixed_perf_vars.append(self.water_flux)
+
+        # Create membrane area variable
+        self.membrane_area = Var(units=pyunits.m**2,
+                                 bounds=(0, None),
+                                 doc="Membrane area")
+        self._perf_var_dict["Membrane Area"] = self.membrane_area
+
+        @self.Constraint(self.flowsheet().time,
+                         doc='Constraint for oxygen consumption.')
```
adam-a-a:
```suggestion
                         doc='Constraint for water flux.')
```

adam-a-a (approved):
LGTM. Good job!

luohezhiming (approved):

PR #480, opened 2022-04-05
Zero-order NF subtype for dye rejection

adam-a-a:
## Summary/Motivation:
This NF subtype incorporates the water flux equation to compute membrane area, as well as rejection rates of solutes.

- add NF subtype with costing
- add tests
- add variable to pump_electricity for applied pressure, matching the var name in the NF subtype

### Legal Acknowledgement

### DIFF
--- a/watertap/core/zero_order_costing.py
+++ b/watertap/core/zero_order_costing.py
@@ -25,6 +25,7 @@ from idaes.generic_models.costing.costing_base import (
     FlowsheetCostingBlockData, register_idaes_currency_units)
 
 from watertap.core.zero_order_base import ZeroOrderBase
+from watertap.costing.watertap_costing_package import cost_membrane
 from watertap.unit_models.zero_order import (
     BrineConcentratorZO,
     ChemicalAdditionZO,
@@ -38,6 +39,7 @@ from watertap.unit_models.zero_order import (
     MABRZO,
     IonExchangeZO,
     IronManganeseRemovalZO,
+    NanofiltrationZO,
     OzoneZO,
     OzoneAOPZO,
     PumpElectricityZO,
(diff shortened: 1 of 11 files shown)


watertap/core/zero_order_costing.py
```diff
@@ -1510,6 +1512,69 @@ def cost_well_field(blk):
...
+
+        t0 = blk.flowsheet().time.first()
+
+        # Get parameter dict from database
+        parameter_dict = \
+            blk.unit_model.config.database.get_unit_operation_parameters(
+                blk.unit_model._tech_type,
+                subtype=blk.unit_model.config.process_subtype)
+
+        if (blk.unit_model.config.process_subtype == "default"
```
andrewlee94: I think I will suggest we make this more explicit, and require the YAML file to define which method to use in the costing parameters, rather than rely on the "subtype". I think this will be more amendable to possible future expansion.

I.e. add a new entry in the cost parameters part of the YAML, then look that up here to determine which method to use.

Secondly, I suggest we break this into a "gateway" method that does the check and then calls out to sub-methods to do the actual implementation (i.e. it calls either `cost_power_law_flow` or a new sub-method (e.g.  `cost_membrane_area`).

andrewlee94 (changes requested):
I think there are a couple of small changes we can make to make this more amenable to future expansion if required, but otherwise looks good.

watertap/unit_models/zero_order/nanofiltration_zo.py
```diff
@@ -37,4 +37,45 @@ def build(self):
...
+            constant_intensity(self)
+        else:
+            self.rejection_comp = Var(self.flowsheet().time,
+                                      self.config.property_package.config.solute_list,
+                                      units=pyunits.dimensionless,
+                                      doc="Component rejection")
+
+            self.water_permeability_coefficient = Var(self.flowsheet().time,
+                                                      units=pyunits.L/pyunits.m**2/pyunits.hour/pyunits.bar,
+                                                      doc="Membrane water permeability coefficient, A")
```
luohezhiming, line 51: I am unsure why there is a `A` in the doc of` water_permeability_coefficient`? Does it have a professional meaning?

watertap/core/zero_order_costing.py
```diff
@@ -25,6 +25,7 @@
     FlowsheetCostingBlockData, register_idaes_currency_units)
 
 from watertap.core.zero_order_base import ZeroOrderBase
+from watertap.costing.watertap_costing_package import cost_membrane
```
bknueven:
```suggestion
```
You ended up not using this

watertap/unit_models/zero_order/tests/test_nanofiltration_zo.py
```diff
@@ -30,6 +30,7 @@
 from watertap.core.wt_database import Database
 from watertap.core.zero_order_properties import WaterParameterBlock
 from watertap.core.zero_order_costing import ZeroOrderCosting
+from watertap.costing.watertap_costing_package import cost_membrane
```
bknueven:
```suggestion
```
Same here.

watertap/core/zero_order_costing.py
```diff
@@ -1510,6 +1512,69 @@ def cost_well_field(blk):
...
+        else:
+            # Get costing parameter sub-block for this technology
+            mem_cost, rep_rate = _get_tech_parameters(
+                blk,
+                parameter_dict,
+                blk.unit_model.config.process_subtype,
+                ["membrane_cost", "membrane_replacement_rate"])
+
+            # Determine if a costing factor is required
+            factor = parameter_dict["capital_cost"]["cost_factor"]
```
bknueven: This is a minor thing -- I would prefer we declare this variable just before its use.

watertap/unit_models/zero_order/nanofiltration_zo.py
```diff
@@ -16,11 +16,11 @@
 """
 
 from idaes.core import declare_process_block_class
-
-from watertap.core import build_sido, constant_intensity, ZeroOrderBaseData
+from pyomo.environ  import Var, Constraint, units as pyunits
+from watertap.core import build_sido, constant_intensity, ZeroOrderBaseData, pump_electricity
```
bknueven:
```suggestion
from watertap.core import build_sido, constant_intensity, ZeroOrderBaseData
```
I don't see `pump_electricity` used here.

bknueven (commented):
I agree with @andrewlee94's larger comment. I found a few minor things to pick on which are not critical.

watertap/core/zero_order_costing.py
luohezhiming: Can you clarify where do you use this `cost_membrane`?

luohezhiming (commented):
I added a few comments, and it seems that the nanofilctration unit does not account for any electricity. If the major objective of addressing electricity is just for operation cost, then it is fine (operational cost depends on some other variables for this unit). But I am not so sure if the electricity is also considered as a necessary factor of unit models.

watertap/unit_models/zero_order/nanofiltration_zo.py
adam-a-a: Yes, the water and salt permeability coefficients are commonly referred to as A and B values, respectively.

watertap/core/zero_order_costing.py
adam-a-a: Originally, I was just going to use this method from `watertap_costing_package` but then wrote the function myself and forgot to delete this import.

luohezhiming (dismissed):

watertap/core/zero_order_costing.py
```diff
@@ -1758,6 +1766,24 @@ def _get_tech_parameters(blk, parameter_dict, subtype, param_list):
...
+    This is meant for units with different cost methods between subtypes.
+    """
+    # Get parameter dict from database
+    parameter_dict = \
+        blk.unit_model.config.database.get_unit_operation_parameters(
+            blk.unit_model._tech_type,
+            subtype=blk.unit_model.config.process_subtype)
+
+    if "cost_method" not in parameter_dict['capital_cost']:
+        raise KeyError("Add a cost_method as a sub-key under capital_cost "
```
andrewlee94: I think this error message could be more descriptive:

```
f"Costing for {blk.unit_model._tech_type} requires a cost_method argument, however "
f"this was not defined for process sub-type {blk.unit_model.config.process_subtype}."
```

watertap/core/zero_order_costing.py
```diff
@@ -1518,62 +1517,71 @@ def cost_nanofiltration(blk):
...
+        if factor == "TPEC":
+            capex_expr *= blk.config.flowsheet_costing_block.TPEC
+        elif factor == "TIC":
+            capex_expr *= blk.config.flowsheet_costing_block.TIC
+
+        blk.capital_cost_constraint = \
+            pyo.Constraint(expr=blk.capital_cost == capex_expr)
+
+        blk.fixed_operating_cost_constraint = \
+            pyo.Constraint(expr=blk.fixed_operating_cost ==
```
andrewlee94: A couple of comments regarding consistency with other zero-order unit operations:

1. Is this really a fixed cost, or a variable cost (i.e. strictly replacement is a function of usage).
2. In other similar models, we have generally just registered material replacement as a flow (even if it is a batch replacement process).

andrewlee94 (commented):
A couple more comments.

watertap/core/zero_order_costing.py
adam-a-a: I suppose that it's technically a variable cost. I just implemented it the same way that it was implemented in `watertap_costing`. 

I noticed that we registered as flows for everything else,  but thought this might be a different category than the rest (all other flows are either per unit kg or per kWh).

adam-a-a: I changed this to a variable cost instead. If you feel strongly about including as a defined flow instead, let me know.

bknueven (approved):

andrewlee94 (approved):

PR #481, opened 2022-04-06
Adding Units for CAS and secondary treatment in WWTPs 

adam-a-a:
## Summary/Motivation:
This PR brings in units for conventional activated sludge as well as a WWTP with secondary treatment. These may be utilized for AMO flowsheets.

- add conventional_activated_sludge from WT3 implementation, with modifications
- collect data for secondary treatment and add corresponding unit
- add tests
- add components

### Legal Acknowledgement


watertap/unit_models/zero_order/tests/test_conventional_activated_sludge_zo.py
```diff
@@ -0,0 +1,256 @@
...
+        data = db.get_unit_operation_parameters("conventional_activated_sludge", subtype=subtype)
+
+        model.fs.unit.load_parameters_from_database()
+
+        for (t, j), v in model.fs.unit.removal_frac_mass_solute.items():
+            assert v.fixed
+            assert v.value == data["removal_frac_mass_solute"][j]["value"]
+
+
+def test_costing():
```
luohezhiming: Maybe also test the subtype for `test_costing`.

watertap/unit_models/zero_order/tests/test_secondary_treatment_wwtp_zo.py
```diff
@@ -0,0 +1,256 @@
...
+====================================================================================
+"""
+
+        assert output in stream.getvalue()
+
+db = Database()
+params = db._get_technology("secondary_treatment_wwtp")
+
+
+class Test_SecondaryTreatmentWWTPZOsubtype:
```
luohezhiming, line 188: Why there is a subtype test here? The yaml file does not have a subtype.

luohezhiming (commented):
Overall, LGTM. I have a few comments on the two test files, please check.

watertap/unit_models/zero_order/tests/test_secondary_treatment_wwtp_zo.py
adam-a-a: Oh, it will in the future, so the test is setup for that.

watertap/unit_models/zero_order/tests/test_conventional_activated_sludge_zo.py
adam-a-a: OK, done!

bknueven (dismissed):

watertap/unit_models/zero_order/tests/test_conventional_activated_sludge_zo.py
luohezhiming: I see you added the subtype test for costing the `test_secondary_treatment_wwtp_zo.py`, but here I denote the `test_conventional_activated_sludge_zo.py` since the model has two subtypes.

watertap/data/techno_economic/component_list.yaml
```diff
@@ -82,8 +84,12 @@ perfluorootanoate_pfoa:
   constituent_longform: Perfluorootanoate (PFOA)
 phosphates:
   constituent_longform: Phosphates
+phosphorus:
```
andrewlee94, line 89: As a general question, is there sufficient difference in the definition of "phosphorous" from "phosphates" in this case?

@tarnold17 and I were discussing this in another PR, and decided to just go with phosphates there.

andrewlee94 (commented):
Rather than move the old YAML files to `depr`, at this point I think we should just start deleting the ones we don't use. As part of this, we should go through the list of YAML files and make sure there aren't any others hanging around that are unused as well and decide whether they should be deleted or have a unit model made for them.

watertap/data/techno_economic/component_list.yaml
adam-a-a: I also had the same discussion with @tarnold17 and had encouraged him to only account for phosphates. However, I created a new yaml for secondary wastewater treatment and estimated average removal efficiencies based on a table in Metcalf & Eddy. Long story short-- I decided to stay true to the referenced table, which provides an entry for "total phosphorus".

adam-a-a:
> Rather than move the old YAML files to `depr`, at this point I think we should just start deleting the ones we don't use. As part of this, we should go through the list of YAML files and make sure there aren't any others hanging around that are unused as well and decide whether they should be deleted or have a unit model made for them.

Ah, OK. I usually deleted them until I noticed the `depr` folder and thought that I'd switch gears. I'll get rid of the two that I sent there but will leave the other stuff in `depr` for now.

watertap/unit_models/zero_order/tests/test_conventional_activated_sludge_zo.py
adam-a-a: Good catch. I added it in.

luohezhiming (dismissed):

luohezhiming (approved):

bknueven (approved):

PR #485, opened 2022-04-14
fix parameter sweep issue in write_to_csv

adam-a-a:
## Summary/Motivation:
Running multisweep for the full treatment train example no longer works, and the issue was an obstacle during efforts to bring in our AMO use case. A subtle mod is made in this PR to resolve the issue for the time being.

- edit _write_to_csv in parameter_sweep to avoid conflict when directory name is in filename path provided

### Legal Acknowledgement

### DIFF
--- a/watertap/tools/parameter_sweep.py
+++ b/watertap/tools/parameter_sweep.py
@@ -553,7 +553,6 @@ def _write_to_csv(
     global_results_arr,
     rank,
     write_csv,
-    dirname,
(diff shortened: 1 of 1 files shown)


bknueven:
I'm pretty sure this also affects the h5 writer, and it would be nice to have a test which fails under the current bug.

bknueven:
#487 replaces this PR.

PR #493, opened 2022-04-14
Metab flowsheet 2

TimBartholomew:
## Fixes/Addresses:

-
-

### Legal Acknowledgement


adam-a-a:
- [x] Tables 1,2,3 : breakdown with original vals, factors applied to original vals, Breakdown with new values
- [x] Consider sensitivity on H2 and CH4 prices

watertap/unit_models/mvc/components/evaporator.py
```diff
@@ -366,7 +366,6 @@ def initialize(
     ):
         """
         General wrapper for pressure changer initialization routines
-
```
andrewlee94: Why are these lines being deleted in this PR?

watertap/unit_models/zero_order/metab_zo.py
```diff
@@ -0,0 +1,130 @@
...
+
+    CONFIG = ZeroOrderBaseData.CONFIG()
+
+    def build(self):
+        super().build()
+
+        self._tech_type = "metab"
+        build_sido_reactive(self)
+
+        self._gas_comp = self.config.process_subtype
```
andrewlee94: Why duplicate this? You should be able to just do `self.config.process_subtype` anywhere you would have used `self._gas_comp`.

Secondly, I am not sure that the underscore is appropriate here either as that indicates a private attribute and generally private attributes should only be accessed by the object itself. However, I have already seen one instance of this being used in the costing module.

watertap/core/zero_order_costing.py
```diff
@@ -965,6 +966,177 @@ def cost_mabr(blk):
...
+        # Register operating cost flows
+        blk.config.flowsheet_costing_block.cost_flow(
+            blk.unit_model.electricity[t0], "electricity"
+        )
+        blk.config.flowsheet_costing_block.cost_flow(blk.unit_model.heat[t0], "heat")
+        blk.config.flowsheet_costing_block.cost_flow(
+            blk.unit_model.properties_byproduct[0].flow_mass_comp[
+                blk.unit_model._gas_comp
+            ],
+            blk.unit_model._gas_comp + "_product",
```
andrewlee94: First, see my comment in the METAB unit model. Secondly, this could give rise to unexpected behaviours if `blk.unit_model._gas_comp` were not a string for some reason (and we have no guards in place to prevent that).

Finally, I am a little hesitant with the current setup as this means our list of costed components is now dynamic. I supposed this just required the end-user to declare any new product component they wish to add via a METAB reactor, but that probably needs to be clearly documented (and runs against what we have generally done in the WT3 models, where everything was concretely defined).

watertap/core/zero_order_costing.py
```diff
@@ -2022,9 +2195,9 @@ def _get_tech_parameters(blk, parameter_dict, subtype, param_list):
                 pblock.add_component(p, vobj)
             except KeyError:
                 raise KeyError(
-                    "Error when trying to retrieve costing parameter"
-                    " from {p} database. Please check the YAML "
-                    "file for this technology for errors."
+                    "Error when trying to retrieve costing parameter "
+                    "for {p}. Please check the YAML "
+                    "file for this technology for errors.".format(p=p)
```
andrewlee94: The way this was written, it was intended to be an `f-string` which would be cleaner and maybe slightly faster.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/metab/metab.py
```diff
@@ -0,0 +1,622 @@
...
+    initialize_system(m)
+
+    results = solve(m)
+    display_results(m)
+
+    add_costing(m)
+    initialize_costing(m)
+    assert_degrees_of_freedom(m, 0)
+
+    m.fs.metab_hydrogen.hydraulic_retention_time.fix(6)  # default - 12 hours, 0.5x
```
andrewlee94, line 60: Why are these being fixed as part of the `main` method, whereas the rest was done by calling out to sub-methods? It seems odd to me that these are not part of the `set_operating_conditions` method for one.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/metab/metab.py
```diff
@@ -0,0 +1,622 @@
...
+    m.fs.costing.metab.bead_cost["methane"].fix(14.40)  # default 1440, 0.01x
+    m.fs.costing.metab.reactor_cost["methane"].fix(78.9)  # default 789, 0.1x
+    m.fs.costing.metab.vacuum_cost["methane"].fix(136.0)  # default 1360, 0.1x
+    m.fs.costing.metab.mixer_cost["methane"].fix(27.40)  # default 2740, 0.1x
+    m.fs.costing.metab.membrane_cost["methane"].fix(498)  # default 498
+
+    results = solve(m)
+    display_costing(m)
+
+    print(
```
andrewlee94, line 98: Similarly, why is this not part of the `display_results` method (or a similar sub-method)?

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/metab/multi_sweep.py
```diff
@@ -0,0 +1,139 @@
...
+        # bead cost
+        m.fs.costing.metab.bead_cost.display()
+        sweep_params["bead_cost"] = LinearSample(
+            m.fs.costing.metab.bead_cost["hydrogen"], 5, 50, nx
+        )
+        output_filename = "sensitivity_" + str(case_num) + ".csv"
+
+    elif case_num == 2:
+        pass
+        # # ================================================================
```
andrewlee94: Commented code.

andrewlee94 (commented):
A few general comments on the code.

On a side note, we should probably go and clean up some of the YAML files for the zero-order models and the general tests of these. A lot of these tests were there to help us find things that didn't follow the standard format, and many of those could be removed or streamlined now that everything has been standardized.

Issue #402, opened 2022-03-11
WaterTAP Costing Package Enhancements


bknueven:
We should do a bit of clean-up from #393 
- [x] Remove stub files in watertap/examples/flowsheets/full_treatment_train/model_componets/ https://github.com/watertap-org/watertap/pull/393#discussion_r823917728 #495 
- [x] Support more than one LCOW / specific energy consumption https://github.com/watertap-org/watertap/pull/393#discussion_r823893782 (#497)
- [x] Clean-up NaOCl / CaOH2 flow costing https://github.com/watertap-org/watertap/pull/393#discussion_r823932410 (#496)
- [x] Cost electricity flow with units as in Zero-Order package (#504) https://github.com/watertap-org/watertap/blob/53ac11da291f2c48a7fbc5d14b493f5d1fbcb960/watertap/core/zero_order_costing.py#L296-L298
- [ ] Add better documentation for WaterTAPCosting package
- [x] Add explicit tests for WaterTAPCosting package (Question: should tests live with the costing package or the unit model? https://github.com/watertap-org/watertap/pull/515#discussion_r863022357)
- [x] Incorporate any upcoming changes to LSRRO flowsheet if needed (Issue #461, #518)
- [ ] Organize parameters for specific unit models:
    - Put parameters for a unit model on their own block
    - Lazily create said block the first time that unit model is costed.

bknueven:
I'm going to close this -- the open issues are captured in #676

PR #497, opened 2022-04-20
Enabling different variants of LCOW, etc.

bknueven:
## Fixes/Addresses:
Part of #402.

## Summary/Motivation:
Sometimes we might want to track different variants of specific energy consumption and/or LCOW; see https://github.com/watertap-org/watertap/pull/393#discussion_r823893782

- Separating out annual_water_production
- Allowing user-specified names for annual_water_production, LCOW, and specific_energy_consumption so different flow basis can be analyzed
- Adding calls to `add_annual_water_production` where appropriate

## Questions:
- Should we add any more safety checking around these methods? Calling them more than once with the same name is probably a mistake.

### Legal Acknowledgement


watertap/costing/watertap_costing_package.py
```diff
@@ -252,50 +256,85 @@ def initialize_build(self):
             self.total_operating_cost, self.total_operating_cost_constraint
         )
 
-        if hasattr(self, "LCOW"):
-            calculate_variable_from_constraint(self.LCOW, self.LCOW_constraint)
+        for var, con in self._registered_LCOWs.values():
+            calculate_variable_from_constraint(var, con)
 
-    def add_LCOW(self, flow_rate):
+    def add_LCOW(self, flow_rate, name="LCOW"):
```
adam-a-a, line 258: Can you add a test demonstrating, say, LCOW with respect to feed flow?

TimBartholomew (approved):
LGTM

adam-a-a (approved):
LGTM - I guess we don't _need_ tests for the mods, but they wouldn't hurt.

bknueven:
> LGTM - I guess we don't _need_ tests for the mods, but they wouldn't hurt.

Note that the way this is structured, it's more-or-less tested already. I also have a separate checkbox on #402 to work on testing for the costing package, but I was focusing on feature enhancement before moving on to testing and documentation.

PR #498, opened 2022-04-21
Add EnergyRecoveryDevice

### LINKED ISSUE #461: Revise implementation of ERD in LSRRO flowsheet
The "work_fluid" should be multiplied by efficiency for the ERD, but the LSRRO flowsheet currently uses the isothermal pump unit to represent a turbine-type ERD and divides by efficiency instead, which is correct for a pump but incorrect for an ERD.

bknueven:
## Summary/Motivation:
As pointed out in #461, when we use a WaterTAP Pump as an energy recovery device, the efficiency is "on the wrong side". This leads to very optimistic values for energy consumption.

In some places in WaterTAP this was already dealt with in an ad-hoc fashion. This PR would formalize this with an `EnergyRecoveryDevice` unit model.

- Add `EnergyRecoveryDevice` unit model & tests
- Update affected example flowsheets
- Starting tests for `WaterTAPCosting`.

### Legal Acknowledgement


andrewlee94 (approved):
Looks good to me.

adam-a-a (approved):

PR #504, opened 2022-04-26
Costing flows in the WaterTAP costing package methods

bknueven:
## Fixes/Addresses:
Part of #402.

## Summary/Motivation:
The zero-order costing package also costs electricity and chemical flows in its methods. Generally speaking its a modeling error *not* to do this, so this PR would adopt the same convention for the WaterTAP costing package.

I am not certain this is what we should be doing. But given a lack of conventions for new-style IDAES costing packages it seems best to standardize within WaterTAP.

- Cost flows in WaterTAP costing package methods
- Remove explicit flow costing from affected flowsheets.

### Legal Acknowledgement


adam-a-a (approved):
LGTM

TimBartholomew (approved):

Issue #503, opened 2022-04-26
Need to update our Fork and Tags for IDAES


aladshaw3:
IDAES will be changing a lot of its paths for import statements for specific unit models and properties that we will be dependent on (see the comment and test failures in #502). 

After updating the fork and tags for installing the latest IDAES, we will likely see a lot of depreciation warnings for those import statements. We will also need to then go through our code base and update those import paths. 

- [ ] Update fork of IDAES and tag for installation 
- [ ] Update all import paths in WaterTAP to IDAES models/properties

aladshaw3:
I am not sure what the priority on this will be, but we should discuss this at the next meeting so we understand how much of a change this is and what this may 'break' in WaterTAP.

bknueven:
Based on the issues in #500, we should ideally update to at least this commit: https://github.com/IDAES/idaes-pse/commit/2cb6df02923ca775a0235ec0782a5f8133375f6b (or just 2.0.0a1), and also update our Pyomo requirement to `pyomo>=6.3` to get the latest version of PyNumero.

adam-a-a:
> Based on the issues in #500, we should ideally update to at least this commit: [IDAES/idaes-pse@2cb6df0](https://github.com/IDAES/idaes-pse/commit/2cb6df02923ca775a0235ec0782a5f8133375f6b) (or just 2.0.0a1), and also update our Pyomo requirement to `pyomo>=6.3` to get the latest version of PyNumero.

Yes, and I would consider this high priority since this is something that can hold up simple PRs.

ksbeattie:
I will update the fork and put a tag on the 2cb6df0 hash, and open a PR to update our setup.py to use that.

aladshaw3:
Closed by #512

PR #500, opened 2022-04-22
How-To Guide for Using a Property Model

MarcusHolly:
## Fixes/Addresses:

Issue #275 

## Summary/Motivation:

This PR adds a How-To guide for using a property model.

## Changes proposed in this PR:
- Adds an .rst file to the list of How-To guides
- Updates index file

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

Issue #464, opened 2022-03-29
Nanofiltration DSPM-DE Enhancement


adam-a-a:
The initial version of the nanofiltration DSPM-DE was pushed in PR#465, but the model needs to undergo validation and refinement.
To Do List:
**DSPM-DE Model:**
_Model Enhancement_
- [ ] Add pressure loss relationship for spiral wound membrane module
- [ ] Add modified mass transfer coefficient due to suction effect
- [ ] Revise documentation as needed: add degrees of freedom and identify variables typically fixed, etc.
- [ ] Expand to 1D and possibly 2D to account for gradient across membrane pore and membrane length
- [ ] Reformulate potentially problematic constraints

_Stability and Testing_
- [x] Refine initialization as needed to enhance stability
- [x] Refine scaling as needed to enhance stability- thanks to @bknueven 
- [x] Augment and refine tests
- [ ] Continue refinement of initialization for more robust model
- [x] Continue refinement of scaling for more robust model - thanks to @bknueven @avdudchenko for contributions 
- [ ] Rigorously validate the model and refine implementation as needed

_Parameter Estimation_
- [ ] Enable parameter estimation of pore radius, effective thickness, pore dielectric constant, and charge density

**MCAS Property Model**
- [ ] add calc option for Stokes radius from Stokes Einstein
- [ ] get relationship for viscosity as a func of temp and concentration for undefined solute set
- [ ] get relationship for density as a func of temp and concentration for undefined solute set (using seawater correlation currently)
- [ ] consider adding the Pitzer activity coefficient model (nice-to-have)
- [ ] consider adding OLI API functionality (nice-to-have)
- [ ] Refine documentation as needed
- [x] Remove H2O from molecular weight data input requirement. MCAS should contain MW of H2O by default.
- [x] consider changing Solute to Anion or Cation (#501)
- [x] **ADD documentation** @lbibl - the issue is being tracked here. #858 
- [x] Rename the `pressure_osm` property to `pressure_osm_phase` in order to be more consistent with IDAES generic property package (#501)
- [x] Refine initialization, scaling, and bounds on variables/constraints (#508)

aladshaw3:
In #508 I have suggested (and merged in) skipping of the nanofiltration tests from initialization onwards. Results of my analysis on the initialization failures suggest that the changes in the property model are not the issue with the failing tests, but the unit model itself (which may need to be reformulated in some manner). 

[...]

PR #508, opened 2022-04-27
Ion pp mod

lbibl:
## Fixes/Addresses:

- Added new vars including `electrical_mobility_comp` and `electrical_conductivity_phase` and related methods.
- Modified the initialization method to set values for quantities calculated from state vars. 
- Modified scaling methods as needed. 
- Modified default values for some vars as needed. 
- Modified the test file to reflect changes in the property package. 

### Legal Acknowledgement


aladshaw3:
@lbibl @adam-a-a Looks like the changes are causing initialization failures in the nanofiltration model. More testing and modifications will be needed (most likely with the scaling of the effected models).

adam-a-a:
> @lbibl @adam-a-a Looks like the changes are causing initialization failures in the nanofiltration model. More testing and modifications will be needed (most likely with the scaling of the effected models).

@aladshaw3 - Yeah, which was what I was worried about. The NF certainly needs refinement, but that's something that I need to work on for the quarter. Maybe we'll get by with a subtle adjustment so that tests pass for this PR. Otherwise, we might want to consider creating a separate dev branch for this (or some other solution) so that the refinement of the NF model doesn't impede your workflow with ED.

bknueven:
> > @lbibl @adam-a-a Looks like the changes are causing initialization failures in the nanofiltration model. More testing and modifications will be needed (most likely with the scaling of the effected models).
> 
> @aladshaw3 - Yeah, which was what I was worried about. The NF certainly needs refinement, but that's something that I need to work on for the quarter. Maybe we'll get by with a subtle adjustment so that tests pass for this PR. Otherwise, we might want to consider creating a separate dev branch for this (or some other solution) so that the refinement of the NF model doesn't impede your workflow with ED.

Given there are relatively minor changes here, shouldn't we question if the scaling changes made in this PR are correct? I think it would be better to make the minimal changes required for this feature *first* and then maybe in a separate PR make unrelated changes to the way this property package is scaled, being sure to test it against every place its used.

adam-a-a:
> > > @lbibl @adam-a-a Looks like the changes are causing initialization failures in the nanofiltration model. More testing and modifications will be needed (most likely with the scaling of the effected models).
> > 
> > 
> > @aladshaw3 - Yeah, which was what I was worried about. The NF certainly needs refinement, but that's something that I need to work on for the quarter. Maybe we'll get by with a subtle adjustment so that tests pass for this PR. Otherwise, we might want to consider creating a separate dev branch for this (or some other solution) so that the refinement of the NF model doesn't impede your workflow with ED.
> 
> Given there are relatively minor changes here, shouldn't we question if the scaling changes made in this PR are correct? I think it would be better to make the minimal changes required for this feature _first_ and then maybe in a separate PR make unrelated changes to the way this property package is scaled, being sure to test it against every place its used.

That makes sense. However, I was able to get the NF test working just now locally. I just had to reduce the default scaling factors that I had set for two ions and adjust some regression tests (for a variable that has already been problematic in NF model). I just can't push the changes because @lbibl needed to enable permission to do so.

aladshaw3:
> > > > @lbibl @adam-a-a Looks like the changes are causing initialization failures in the nanofiltration model. More testing and modifications will be needed (most likely with the scaling of the effected models).
> > > 
> > > 
> > > @aladshaw3 - Yeah, which was what I was worried about. The NF certainly needs refinement, but that's something that I need to work on for the quarter. Maybe we'll get by with a subtle adjustment so that tests pass for this PR. Otherwise, we might want to consider creating a separate dev branch for this (or some other solution) so that the refinement of the NF model doesn't impede your workflow with ED.
> > 
> > 
> > Given there are relatively minor changes here, shouldn't we question if the scaling changes made in this PR are correct? I think it would be better to make the minimal changes required for this feature _first_ and then maybe in a separate PR make unrelated changes to the way this property package is scaled, being sure to test it against every place its used.
> 
> That makes sense. However, I was able to get the NF test working just now locally. I just had to reduce the default scaling factors that I had set for two ions and adjust some regression tests (for a variable that has already been problematic in NF model). I just can't push the changes because @lbibl needed to enable permission to do so.

Are the changes significant? Can you just use the GitHub interface to commit those changes directly in this PR?

bknueven:
Screen Shot 2022-04-27 at 5 24 30 PM
@lbibl you need to check this box so @adam-a-a can commit his changes to this PR. In general, there's no real reason to leave this unchecked.

aladshaw3:
@lbibl You will also need to run black on both these files to pass the format check. You can do this in terminal as:

`black {file_name}`

lbibl:
Hi all, thanks for your messages. Before I opened the PR,  `pytest test_nanofiltration_DSPMDE_0D.py` was able to pass on my local but did show 4 "xfailed " .  So I wonder if my arm64 env still missed something and will look into that.   Meanwhile, if @adam-a-a can make reasonable/easy changes on the nf unit model as a fix, that'll be great too.  I have checked that box to allow changes from maintainers and will have it checked for the future.  Also just blacked the two files to pass the formatting checks and the commit was pushed.  Sorry for the negligence in my first PR.  @aladshaw3 @bknueven @adam-a-a

aladshaw3:
@adam-a-a I think there is something fundamentally wrong with the scaling in the nanofiltration model (since we are getting different model results / failing tests on different python versions and OSs). I can have a look at it more today, but we may need to just disable the nanofiltration model and revisit it later.

aladshaw3:
@adam-a-a I have checked the log during initialization for the nanofiltration model. The results indicate that the unit model itself (nanofiltration_DSPMDE) is causing the issue. The properties are properly scaled and solve well during initialization, but the unit model never solves. (see attached).

[nanofiltration_DSPMDE_initialization_log.txt](https://github.com/watertap-org/watertap/files/8583036/nanofiltration_DSPMDE_initialization_log.txt)

This tells me that any issue with failing tests is caused by the nanofiltration model and not the changes made in the property package. I think we may want to consider disabling the nanofiltration model tests and revisiting at another time so that this PR can get done soon.

aladshaw3:
For some reason, there is a documentation failure now. It is from the newly merged docs (#500). That PR passed all its checks, so I am not sure why it is failing now. 

[FailingDocLog-python3_8.txt](https://github.com/watertap-org/watertap/files/8583330/FailingDocLog-python3_8.txt)

adam-a-a:
> @adam-a-a I have checked the log during initialization for the nanofiltration model. The results indicate that the unit model itself (nanofiltration_DSPMDE) is causing the issue. The properties are properly scaled and solve well during initialization, but the unit model never solves. (see attached).
> 
> [nanofiltration_DSPMDE_initialization_log.txt](https://github.com/watertap-org/watertap/files/8583036/nanofiltration_DSPMDE_initialization_log.txt)
> 
> This tells me that any issue with failing tests is caused by the nanofiltration model and not the changes made in the property package. I think we may want to consider disabling the nanofiltration model tests and revisiting at another time so that this PR can get done soon.

@aladshaw3 Yes, I'm aware of the issue and realized that some tests pass on Windows while Linux tests fail. The NF model requires a bunch of refinement (outlined in issue #464 ). 

At first, I was alarmed by your suggestion, until I realized that you only intended to skip specific tests rather than the whole set of tests for NF DSPM. So-- I think I'm OK with that.

adam-a-a:
> For some reason, there is a documentation failure now. It is from the newly merged docs (#500). That PR passed all its checks, so I am not sure why it is failing now.
> 
> [FailingDocLog-python3_8.txt](https://github.com/watertap-org/watertap/files/8583330/FailingDocLog-python3_8.txt)

I noticed this issue elsewhere too, but I think I managed to make that go away by rerunning the failed test. In any case, we probably want to resolve this sooner rather than later (i.e., updating IDAES tag as you mentioned in issue #503 ).

aladshaw3:
> > @adam-a-a I have checked the log during initialization for the nanofiltration model. The results indicate that the unit model itself (nanofiltration_DSPMDE) is causing the issue. The properties are properly scaled and solve well during initialization, but the unit model never solves. (see attached).
> > [nanofiltration_DSPMDE_initialization_log.txt](https://github.com/watertap-org/watertap/files/8583036/nanofiltration_DSPMDE_initialization_log.txt)
> > This tells me that any issue with failing tests is caused by the nanofiltration model and not the changes made in the property package. I think we may want to consider disabling the nanofiltration model tests and revisiting at another time so that this PR can get done soon.
> 
> @aladshaw3 Yes, I'm aware of the issue and realized that some tests pass on Windows while Linux tests fail. The NF model requires a bunch of refinement (outlined in issue #464 ).
> 
> At first, I was alarmed by your suggestion, until I realized that you only intended to skip specific tests rather than the whole set of tests for NF DSPM. So-- I think I'm OK with that.

As another point of comparison, I ran the nanofiltration model before the changes made in this PR and here is that log.
[nanofiltration_DSPMDE_initialization_log_pre_changes.txt](https://github.com/watertap-org/watertap/files/8583566/nanofiltration_DSPMDE_initialization_log_pre_changes.txt)

Comparing to the new log (with the changes proposed in this PR), you can see that the initialization of the properties section has improved because of the changes made by @lbibl.

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -640,7 +701,7 @@ def build(self):
         self.flow_mol_phase_comp = Var(
             self.params.phase_list,
             self.params.component_list,
-            initialize=100,  # todo: revisit
+            initialize=0.1,  # todo: revisit
             bounds=(1e-8, None),
```
aladshaw3, line 743:
```suggestion
            bounds=(1e-16, None),
```

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -669,9 +730,7 @@ def _mass_frac_phase_comp(self):
         self.mass_frac_phase_comp = Var(
             self.params.phase_list,
             self.params.component_list,
-            initialize=lambda b, p, j: 0.4037
-            if j == "H2O"
-            else 0.0033,  # todo: revisit
+            initialize=0.5,
             bounds=(1e-8, 1.001),
```
aladshaw3, line 772:
```suggestion
            bounds=(1e-16, 1.001),
```

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -771,7 +830,7 @@ def _conc_mol_phase_comp(self):
         self.conc_mol_phase_comp = Var(
             self.params.phase_list,
             self.params.component_list,
-            initialize=10,
+            initialize=500,
             bounds=(1e-8, None),
```
aladshaw3, line 872:
```suggestion
            bounds=(1e-16, None),
```

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -811,7 +870,7 @@ def _flow_mass_phase_comp(self):
         self.flow_mass_phase_comp = Var(
             self.params.phase_list,
             self.params.component_list,
-            initialize=100,
+            initialize=0.5,
             bounds=(1e-8, None),
```
aladshaw3, line 912:
```suggestion
            bounds=(1e-16, None),
```

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -831,7 +890,7 @@ def _mole_frac_phase_comp(self):
         self.mole_frac_phase_comp = Var(
             self.params.phase_list,
             self.params.component_list,
-            initialize=0.1,
+            initialize=0.5,
             bounds=(1e-8, 1.001),
```
aladshaw3, line 932:
```suggestion
            bounds=(1e-16, 1.001),
```

aladshaw3 (commented):
Correct me if I am wrong, but I thought we had discussed changing the 'bounds' on many of these variables because they didn't go low enough. Especially, for things like mole fractions and concentrations, we will need to allow for smaller concentration values if we want to consider very dilute systems. 

Same to be said of the variables on lines (784 - `dens_solvent`, 854 - `conc_mass`, 912 - `molality`, and 960 - `act_coeff`). I don't think the bounds on these variables are correct or accurate. For example, the activity coefficient can go greater than 1 and mass concentration should be allowed to go lower than 1e-8.

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -460,6 +477,50 @@ def initialize(
...
         flags = fix_state_vars(self, state_args)
+
+        # initialize vars caculated from state vars
+        for k in self.keys():
+            for j in self[k].params.component_list:
+                if self[k].is_property_constructed("mass_frac_phase_comp"):
+                    self[k].conc_mass_phase_comp["Liq", j].set_value(
+                        self[k].dens_mass_phase["Liq"]
+                        * self[k].mass_frac_phase_comp["Liq", j]
+                    )
```
aladshaw3: This section doesn't seem right to me. You check to see if `mass_frac` is constructed, but then you are initializing `conc_mass` instead. Is something missing here? Shouldn't this line check to see if `conc_mass` is constructed then initializing it? 

Shouldn't there also be a step before where you initialize 'mass_frac'? I would think that the initialization of the mole fractions should be the first step (since it only depends on the state var `flow_mol`). Then you can initialize the `mass_frac` based on the `mole_frac` (since they are closely related).

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -252,6 +256,15 @@ def build(self):
...
         )
+        # Ion electrical mobility
+        self.electrical_mobility_comp = Param(
+            self.ion_set | self.solute_set,
+            mutable=True,
+            default=5.19e-8,  # default as Na+
+            initialize=self.config.electrical_mobility_data,
+            units=pyunits.meter**2 * pyunits.volt**-1 * pyunits.second**-1,
+            doc="Ion electrical mobility",
+        )
```
aladshaw3, line 267: This parameter is indexed by `ion_set` and `solute_set` (meaning that it is currently indexed for all dissolved species (both charged and uncharged). However, when later used to calculate conductivity you only iterate over and use these values from the `ion_set`.

Should this parameter only be for the `ion_set` or should the calculation of conductivity iterate over both the `ion_set` and `solute_set`? (Or does it really matter?)

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -999,6 +1063,27 @@ def rule_pressure_osm_phase(b):
...
+
+        def rule_electrical_conductivity_phase(b):
+            return b.electrical_conductivity_phase["Liq"] == sum(
+                Constants.faraday_constant
+                * abs(b.charge_comp[j])
+                * b.electrical_mobility_comp[j]
+                * b.conc_mol_phase_comp["Liq", j]
+                for j in self.params.ion_set
+            )  # maybe revisit for other emprical calculation or non-ideal situation
```
aladshaw3: The new parameter for electrical mobility includes both `ion_set` and `solute_set`, but we only iterate through `ion_set` here. (See comment above about the new parameter).

watertap/property_models/ion_DSPMDE_prop_pack.py
lbibl: This comment on "bounds" has led to an active discussion among the chemistry team on how we set up near-zero bounds throughout all property packages in WaterTAP.  @bknueven opened a separate PR  #511 to work on improving bounds.  I am taking this note as a resolution for all bound-related comments in this PR.

watertap/property_models/ion_DSPMDE_prop_pack.py
aladshaw3: Per conversion with Xiangyu, this section is correct. The existence of mass_frac already implies that conc_mass is constructed.

watertap/property_models/ion_DSPMDE_prop_pack.py
aladshaw3: Per conversation with Xiangyu, this indexing is necessary because of how the solutes (both charged and uncharged) are specified in the configuration args.

aladshaw3 (dismissed):
This looks good to go. We are skipping some nanofiltration_DSPMDE tests for now. That model needs to be refined in a separate PR.

adam-a-a (dismissed):
LGTM

lbibl:
I am correcting a few minor issues per comments above.  Please wait for at least another commit from me before merging.

watertap/property_models/ion_DSPMDE_prop_pack.py
lbibl: I changed the index here to be `ion_set | solute_set`.  Per discussion with Austin, we may modify the `solute_list` configuration in the future to make charge and electrical mobility indexed by `ion_set` only.  For now the two were indexed also by neutral species whose values will be set zero. No issues by this way for the current unit models.

watertap/property_models/ion_DSPMDE_prop_pack.py
lbibl: Thanks for @aladshaw3 comment.  Actually I did find an error in a second look and fixed it in the latest commit.   Also, I went ahead to add initialization methods for a few other vars.

lbibl:
In the latest commit (25aa85b) I fixed minor issues and added initialization method for a few other vars.  Test for ion property package passed.  Maybe @adam-a-a  or @aladshaw3 may try if the added initialization for osmosis pressure improves the nanofiltation model.  This would be good to go if no other input at this stage.

aladshaw3 (approved):
LGTM

adam-a-a (approved):
LGTM

PR #502, opened 2022-04-25
Adding ASM1 model

andrewlee94:
## Summary/Motivation:
This PR adds an implementation of the Activated Sludge Model No.1 in the form of a reaction property package along with an associated (very simple) thermophysical property package. These property packages can be combined with the core IDAES reactor models to model activated sludge processes.

## Changes proposed in this PR:
- Adds property and reaction packages for ASM1
- Includes a verification test based on a report for a 5 stage biological wasterwater treatment plant.

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #509, opened 2022-04-28
Preliminary Documentation for ASM1 Property Model

MarcusHolly:
## Summary/Motivation:

This PR adds documentation for the activated sludge model no.1 (ASM1) property model. This documentation is modelled loosely after the coagulation property model. That being said, I have a few questions:

1.) Should this documentation retain the density and viscosity correlations from the coagulation example?

2.) Should the observed conversion rates from the reference link be included in the Relationships table?

3.) Should the stoichiometric and kinetic parameters from the reference link be included in the Parameters table?

- Adds the property model documentation for ASM1

### Legal Acknowledgement


andrewlee94:
@MarcusHolly

1. For now, the ASM1 properties do not have viscosity at all, and density is assumed to be constant, so all of that should be dropped.
2. I would not include any values for the conversions - this documentation should focus on the actual property package alone (i.e. only the rate expressions - conversion comes at the unit level). Discussion of conversion should be left to later documentation of a case study/example.
3. Listing the default parameter values is probably a good idea, referencing Alex et al. as the source.

Finally, I would use the original report on the model instead of the application one by Alex et al.:

Henze, M., Grady, C.P.L., Gujer, W., Marais, G.v.R., Matsuo, T.,
"Activated Sludge Model No. 1", 1987, IAWPRC Task Group on Mathematical Modeling
for Design and Operation of Biological Wastewater Treatment

docs/technical_reference/property_models/activated_sludge_model.rst
```diff
@@ -0,0 +1,144 @@
+ASM1 Property Package
+=====================
+
+This package implements property relationships for the treatment of activated sludge as provided in `Alex, J. et al. (2008) <https://www.iea.lth.se/publications/Reports/LTH-IEA-7229.pdf>`_.
+
+This Activated Sludege Model no.1 (ASM1) property package:
+   * supports only 'H2O', 'TDS', 'TSS', and 'Sludge' as Components
```
adam-a-a: This looks like it was copy/pasted from the coag/floc rst file. You should scan through and remove/revise similar parts that aren't related to the ASM1 model.

MarcusHolly:
> @MarcusHolly
> 
> 1. For now, the ASM1 properties do not have viscosity at all, and density is assumed to be constant, so all of that should be dropped.
> 2. I would not include any values for the conversions - this documentation should focus on the actual property package alone (i.e. only the rate expressions - conversion comes at the unit level). Discussion of conversion should be left to later documentation of a case study/example.
> 3. Listing the default parameter values is probably a good idea, referencing Alex et al. as the source.
> 
> Finally, I would use the original report on the model instead of the application one by Alex et al.:
> 
> Henze, M., Grady, C.P.L., Gujer, W., Marais, G.v.R., Matsuo, T., "Activated Sludge Model No. 1", 1987, IAWPRC Task Group on Mathematical Modeling for Design and Operation of Biological Wastewater Treatment

Sounds like a plan. I'll work on this first thing tomorrow morning then.

docs/technical_reference/property_models/activated_sludge_model.rst
```diff
@@ -0,0 +1,144 @@
...
+  "Slowly Biodegradable Substrate", ":math:`X_S`", "biodegradable", "[p]", ":math:`\text{kg/m}^{3}`"
+  "Active Heterotrophic Biomass", ":math:`X_{B,H}`", "heterotrophic", "[p]", ":math:`\text{kg/m}^{3}`"
+  "Active Autotrophic Biomass", ":math:`X_{B,A}`", "autotrophic", "[p]", ":math:`\text{kg/m}^{3}`"
+  "Particulate Products Arising from Biomass Decay", ":math:`X_P`", "decay_particulate", "[p]", ":math:`\text{kg/m}^{3}`"
+  "Oxygen", ":math:`S_O`", "oxygen", "[p]", ":math:`\text{kg/m}^{3}`"
+  "Nitrate and Nitrite Nitrogen", ":math:`S_{NO}`", "nitrates", "[p]", ":math:`\text{kg/m}^{3}`"
+  "NH4 :math:`^{+}` + NH :math:`_{3}` Nitrogen", ":math:`S_{NH}`", "ammonium", "[p]", ":math:`\text{kg/m}^{3}`"
+  "Soluble Biodegradable Organic Nitrogen", ":math:`S_{ND}`", "nitrogen_soluble", "[p]", ":math:`\text{kg/m}^{3}`"
+  "Particulate Biodegradable Organic Nitrogen", ":math:`X_{ND}`", "nitrogen_particulate", "[p]", ":math:`\text{kg/m}^{3}`"
+  "Alkalinity", ":math:`S_{ALK}`", "alkalinity", "[p]", ":math:`\text{kg/m}^{3}`"
```
adam-a-a: These will need to be updated in accordance with the latest changes in PR #502 . Most of the names were altered for these. Be sure to look at both the thermophysical property model and the reaction prop model for ASM1.

docs/technical_reference/property_models/activated_sludge_model.rst
```diff
@@ -0,0 +1,144 @@
...
+ "Reference density (at 273 K)", ":math:`\rho_{ref}`", "ref_dens_liq", "None", ":math:`\text{kg/}\text{m}^3`"
+ "Relative density contribution from salt mass fraction", ":math:`\rho_s`", "dens_slope", "None", ":math:`\text{kg/}\text{m}^3`"
+ "First density temperature coefficient", ":math:`A`", "dens_param_A", "None", ":math:`\text{K}^{-2}`"
+ "Second density temperature coefficient", ":math:`B`", "dens_param_B", "None", ":math:`\text{K}^{-1}`"
+ "Third density temperature coefficient", ":math:`C`", "dens_param_C", "None", ":math:`\text{dimensionless}`"
+ "First pressure coefficient", ":math:`\alpha`", "ref_pressure_correction", "None", ":math:`\text{dimensionless}`"
+ "Second pressure coefficient", ":math:`\beta`", "ref_pressure_slope", "None", ":math:`\text{Pa}^{-1}`"
+ "Reference viscosity (at 273 K)", ":math:`\mu_{ref}`", "mu_A", "None", ":math:`\text{kg/}\text{m/}\text{s}`"
+ "First viscosity temperature coefficient", ":math:`\mu_B`", "mu_B", "None", ":math:`\text{K}`"
+ "Second viscosity temperature coefficient", ":math:`\mu_C`", "mu_C", "None", ":math:`\text{K}`"
```
adam-a-a: You can remove anything here that is not in the ASM1 property model. Be sure to add anything from the ASM1 prop model that is missing here.

docs/technical_reference/property_models/activated_sludge_model.rst
```diff
@@ -0,0 +1,144 @@
...
+  "Phases", ":math:`p`", "['Liq']"
+
+State variables
+---------------
+.. csv-table::
+   :header: "Description", "Symbol", "Variable", "Index", "Units"
+
+   "Component mass flowrate", ":math:`M_j`", "flow_mass_phase_comp", "[p, j]", ":math:`\text{kg/s}`"
+   "Temperature", ":math:`T`", "temperature", "None", ":math:`\text{K}`"
+   "Pressure", ":math:`P`", "pressure", "None", ":math:`\text{Pa}`"
```
adam-a-a: You'll want to review the state variables as defined in the ASM1 property model in PR #502 and update these.

docs/technical_reference/property_models/activated_sludge_model.rst
```diff
@@ -0,0 +1,144 @@
...
+   * is formulated on a mass basis
+   * does NOT support formulations on a molar basis
+   * includes mass density correction for fraction of suspended/dissolved solids
+
+Sets
+----
+.. csv-table::
+  :header: "Description", "Symbol", "Indices"
+
+  "Components", ":math:`j`", "['H2O', 'TDS', 'TSS', 'Sludge']"
```
adam-a-a: This component set is for the coag/floc package, so you can revise this to match what's in the ASM1 property model in PR #502

docs/technical_reference/property_models/activated_sludge_model.rst
```diff
@@ -0,0 +1,144 @@
...
+=====================
+
+This package implements property relationships for the treatment of activated sludge as provided in `Alex, J. et al. (2008) <https://www.iea.lth.se/publications/Reports/LTH-IEA-7229.pdf>`_.
+
+This Activated Sludege Model no.1 (ASM1) property package:
+   * supports only 'H2O', 'TDS', 'TSS', and 'Sludge' as Components
+   * supports only liquid phase
+   * is formulated on a mass basis
+   * does NOT support formulations on a molar basis
+   * includes mass density correction for fraction of suspended/dissolved solids
```
adam-a-a: Same applies here- you can take any assumptions listed in the ASM1 property and reaction packages from PR #502 , if any were mentioned in doc strings. If not, you can delete these in the interim, and we can add relevant assumptions later.

adam-a-a (changes requested):
I know this is still a work in progress, but I have a few comments for revising.

MarcusHolly:
I couldn't find some of the names of the kinetic and stoichiometric parameters, so I just listed the description as the variable name with the default parameter value next to it in parentheses.

I was also unsure if the components should only be H20 and sludge as well as what the variable should be for "Alkalinity in molar concentration" (in state variables).

andrewlee94:
@MarcusHolly You should look at the actual model code for reference:

https://github.com/watertap-org/watertap/blob/main/watertap/property_models/activated_sludge/asm1_properties.py.
https://github.com/watertap-org/watertap/blob/main/watertap/property_models/activated_sludge/asm1_reactions.py

There are 14 different components present, which you are in the first file. The second file defines all the parameters, which use the same names as those in the origin paper.

docs/technical_reference/property_models/activated_sludge_model.rst
```diff
@@ -0,0 +1,144 @@
...
+  "Slowly biodegradable substrate X_S", ":math:`X_S`", "X_S", "[p]", ":math:`\text{kg/m}^{3}`"
+  "Active heterotrophic biomass X_B,H", ":math:`X_{B,H}`", "X_BH", "[p]", ":math:`\text{kg/m}^{3}`"
+  "Active autotrophic biomass X_B,A", ":math:`X_{B,A}`", "X_BA", "[p]", ":math:`\text{kg/m}^{3}`"
+  "Particulate products arising from biomass decay, X_P", ":math:`X_P`", "X_P", "[p]", ":math:`\text{kg/m}^{3}`"
+  "Oxygen, S_O", ":math:`S_O`", "S_O", "[p]", ":math:`\text{kg/m}^{3}`"
+  "Nitrate and nitrite nitrogen, S_NO", ":math:`S_{NO}`", "S_NO", "[p]", ":math:`\text{kg/m}^{3}`"
+  "NH4 :math:`^{+}` + NH :math:`_{3}` Nitrogen, S_NH", ":math:`S_{NH}`", "S_NH", "[p]", ":math:`\text{kg/m}^{3}`"
+  "Soluble biodegradable organic nitrogen, S_ND", ":math:`S_{ND}`", "S_ND", "[p]", ":math:`\text{kg/m}^{3}`"
+  "Particulate biodegradable organic nitrogen, X_ND", ":math:`X_{ND}`", "X_ND", "[p]", ":math:`\text{kg/m}^{3}`"
+  "Alkalinity, S_ALK", ":math:`S_{ALK}`", "S_ALK", "[p]", ":math:`\text{kg/m}^{3}`"
```
yalinli2: kg of what per m3? (alkalinity is not a component, could be CaCO3, C, CO3 and the results will be different)

yalinli2 (commented):
I only had one minor comment on the alkalinity (and it goes back to the multiplier of 14 we mentioned the other day), I'm just not clear what exactly the unit of alkalinity is

docs/technical_reference/property_models/activated_sludge_model.rst
MarcusHolly: Based on the original report (hyperlink at the top of the file) alkalinity is assumed to be HCO3- only

andrewlee94: This should be kmol HCO3-/m^3 - alkalinity is measured on a molar basis.

andrewlee94:
@yalinli2 After digging through the ASM2d model, I finally worked out where the 14 comes from. It stems from a charge balance and the fact that most things are done on a mass basis except for alkalinity.

For example, consider a reaction that consumes NO3-; in order to maintain the charge balance, alkalinity (as HCO3-) needs to be produced on a 1:1 molar basis. However, all the calculations are done on a mass basis, so we have:

`NO3- / 14 == HCO3- /61`

I have updated the model to be explicitly about this. However, the original paper left HCO3- on a molar basis thus the 61 g/mol did not appear (in the case of this implementation, the 61 cancels out deeper in the code).

yalinli2:
> @yalinli2 After digging through the ASM2d model, I finally worked out where the 14 comes from. It stems from a charge balance and the fact that most things are done on a mass basis except for alkalinity.
> 
> For example, consider a reaction that consumes NO3-; in order to maintain the charge balance, alkalinity (as HCO3-) needs to be produced on a 1:1 molar basis. However, all the calculations are done on a mass basis, so we have:
> 
> `NO3- / 14 == HCO3- /61`
> 
> I have updated the model to be explicitly about this. However, the original paper left HCO3- on a molar basis thus the 61 g/mol did not appear (in the case of this implementation, the 61 cancels out deeper in the code).

OK, so essentially that 14 was because it was for NO3 and NO3 was expressed as kg N/m3? If so, then that makes sense, and yea noting it is on a molar basis is great, thanks!

yalinli2 (commented):
Looks good, thanks!

yalinli2 (dismissed):

docs/technical_reference/property_models/activated_sludge_model.rst
```diff
@@ -0,0 +1,144 @@
...
+   * supports only 'H2O', 'S_I', 'S_S', 'X_I', 'X_S', 'X_BH', 'X_BA', 'X_P', 'S_O', 'S_NO', 'S_NH', 'S_ND', 'X_ND', and 'S_ALK' as components
+   * supports only liquid phase
+   * is formulated on a mass basis
+
+Sets
+----
+.. csv-table::
+  :header: "Description", "Symbol", "Indices"
+
+  "Components", ":math:`j`", "['H2O', 'S_I', 'S_S', 'X_I', 'X_S', 'X_BH', 'X_BA', 'X_P', 'S_O', 'S_NO', 'S_NH', 'S_ND', 'X_ND', 'S_ALK']"
```
adam-a-a, line 16: I think after "Sets", it would be good to add another section for "Components" where we have a csv-table that gives the names for each component in connection with symbol names. For example:
"Active heterotrophic biomass",  ":math:`\text{X}_{b, h}`"

adam-a-a: Similar to what you have for Parameters, except we'd only need the Description and Symbol columns

docs/technical_reference/property_models/activated_sludge_model.rst
```diff
@@ -0,0 +1,144 @@
...
+State variables
+---------------
+.. csv-table::
+   :header: "Description", "Symbol", "Variable", "Index", "Units"
+
+   "Total volumetric flowrate", ":math:`Q`", "flow_vol", "None", ":math:`\text{m}^3\text{/s}`"
+   "Temperature", ":math:`T`", "temperature", "None", ":math:`\text{K}`"
+   "Pressure", ":math:`P`", "pressure", "None", ":math:`\text{Pa}`"
+   "Component mass concentrations", ":math:`C_j`", "conc_mass_comp", "[p]", ":math:`\text{kg/}\text{m}^3`"
+   "Alkalinity in molar concentration", ":math:`A`", "alkalinity", "[p]", ":math:`\text{kmol HCO}_{3}^{-}\text{/m}^{3}`"
```
adam-a-a, line 47: @andrewlee94 - might want to double-check this. This was changed to kmol HCO3/m3 based on Yalin's feedback, but I can't recall if that is the case here, considering that alkalinity is later converted based on a charge balance.

docs/technical_reference/property_models/activated_sludge_model.rst
```diff
@@ -0,0 +1,144 @@
...
+
+Parameters
```
adam-a-a:
```suggestion
Stoichiometric Parameters
```
Although we generally have a section for just "Parameters", renaming this might be more in line with what's in the model as well as typical literature on ASM1.

docs/technical_reference/property_models/activated_sludge_model.rst
adam-a-a: Nevermind- I see now that you already accounted for these in the "Properties" section!

docs/technical_reference/property_models/activated_sludge_model.rst
andrewlee94: Alkalinity is in kmol/m^3 - I put the conversion in the actual model to be explicit about what is happening.

docs/technical_reference/property_models/activated_sludge_model.rst
```diff
@@ -0,0 +1,144 @@
...
+  "Slowly biodegradable substrate X_S", ":math:`X_S`", "X_S", "[p]", ":math:`\text{kg/m}^{3}`"
+  "Active heterotrophic biomass X_B,H", ":math:`X_{B,H}`", "X_BH", "[p]", ":math:`\text{kg/m}^{3}`"
+  "Active autotrophic biomass X_B,A", ":math:`X_{B,A}`", "X_BA", "[p]", ":math:`\text{kg/m}^{3}`"
+  "Particulate products arising from biomass decay, X_P", ":math:`X_P`", "X_P", "[p]", ":math:`\text{kg/m}^{3}`"
+  "Oxygen, S_O", ":math:`S_O`", "S_O", "[p]", ":math:`\text{kg/m}^{3}`"
+  "Nitrate and nitrite nitrogen, S_NO", ":math:`S_{NO}`", "S_NO", "[p]", ":math:`\text{kg/m}^{3}`"
+  "NH4 :math:`^{+}` + NH :math:`_{3}` Nitrogen, S_NH", ":math:`S_{NH}`", "S_NH", "[p]", ":math:`\text{kg/m}^{3}`"
+  "Soluble biodegradable organic nitrogen, S_ND", ":math:`S_{ND}`", "S_ND", "[p]", ":math:`\text{kg/m}^{3}`"
+  "Particulate biodegradable organic nitrogen, X_ND", ":math:`X_{ND}`", "X_ND", "[p]", ":math:`\text{kg/m}^{3}`"
+  "Alkalinity, S_ALK", ":math:`S_{ALK}`", "S_ALK", "[p]", ":math:`\text{kmol HCO}_{3}^{-}\text{/m}^{3}`"
```
adam-a-a: Based on how the property model is formulated, these are listed as components, while calling `conc_mass_comp['S_I"]`  would give us the mass concentration of soluble inert organic matter (for example). 

Maybe we'd want to move these to a separate "Components" section then, right after "Sets" and eliminate the Index and Units columns.

docs/technical_reference/property_models/activated_sludge_model.rst
```diff
@@ -0,0 +1,144 @@
...
+  m.fs.properties = props.CoagulationParameterBlock()               # Needs to be replaced with ASM
+
+  # set scaling for component mass flowrate
+  m.fs.properties.set_default_scaling('flow_mass_phase_comp', 1, index=('Liq','H2O'))
+  m.fs.properties.set_default_scaling('flow_mass_phase_comp', 1e2, index=('Liq','TDS'))
+  m.fs.properties.set_default_scaling('flow_mass_phase_comp', 1e2, index=('Liq','TSS'))
+  m.fs.properties.set_default_scaling('flow_mass_phase_comp', 1e3, index=('Liq','Sludge'))
+
+  # calculate scaling factors
+  calculate_scaling_factors(m.fs)
```
adam-a-a: Maybe we can exclude this bit on scaling and maybe just include any default scaling factors.

adam-a-a (changes requested):
A few more suggested changes. We probably also want to make note of how the model is structured. That is, the model is broken down into a property model and reaction model. Then we could describe what can be located in each. However, perhaps that could be reserved for a subsequent PR.

docs/technical_reference/property_models/activated_sludge_model.rst
andrewlee94: For now, the AMS1 implementation has no scaling factors set, so I would leave this section out entirely, or just say that is has yet to be implemented.

docs/technical_reference/property_models/activated_sludge_model.rst
```diff
@@ -0,0 +1,116 @@
+ASM1 Property Package
+=====================
+
+This package implements property relationships for the treatment of wastewater using an activated sludge biological reactor as provided in `Henze, M. et al. (1987) <https://belinra.inrae.fr/doc_num.php?explnum_id=4467>`_.
+
+This Activated Sludge Model no.1 (ASM1) property package:
+   * supports only 'H2O', 'S_I', 'S_S', 'X_I', 'X_S', 'X_BH', 'X_BA', 'X_P', 'S_O', 'S_NO', 'S_NH', 'S_ND', 'X_ND', and 'S_ALK' as components
```
adam-a-a:
```suggestion
   * supports 'H2O', 'S_I', 'S_S', 'X_I', 'X_S', 'X_BH', 'X_BA', 'X_P', 'S_O', 'S_NO', 'S_NH', 'S_ND', 'X_ND', and 'S_ALK' as components
```

andrewlee94 (approved):

adam-a-a (approved):
LGTM

PR #515, opened 2022-04-29
Add capital cost function for crystallizer model

OOAmusat:
## Summary/Motivation:

This PR adds two captal cost estimation functions for the crystallizer model.

- Add mass- and volume-based capital cost estimates for crystallizer

### Legal Acknowledgement


TimBartholomew:
@bknueven can you look into these test failures for LSRRO? It's weird that they are failing for the number of variables and Mayo didn't touch the LSRRO files.

TimBartholomew:
@OOAmusat 
Ben's response was:
It’s these lines: https://github.com/watertap-org/watertap/pull/515/files#diff-b8409e0826e208143b2dc914e86c9f79da1e396679bd80ab49edb430ead318e1R201-R236

These new parameters appear on every costing block. Therefore we should expect an increase in the number of Vars for any flowsheet using the watertap costing package.

watertap/costing/watertap_costing_package.py
```diff
@@ -586,6 +629,67 @@ def cost_caoh2_mixer(blk):
...
+                                blk.unit_model.solids.flow_mass_phase_comp[0, "Sol", j]
+                                for j in blk.unit_model.config.property_package.solute_set
+                            )
+                            / blk.costing_package.fc_crystallizer_ref_capacity
+                        )
+                        ** blk.costing_package.fc_crystallizer_ref_exponent
+                    ),
+                    to_units=blk.costing_package.base_currency,
+                )
+            )
```
bknueven: I've generally written different costing types as their own method (see `cost_pump` for example) for maximal flexibility. I.e. you can specify the method directly *or* use the `CrystallizerCostType` attribute.

watertap/unit_models/tests/test_crystallizer.py
```diff
@@ -286,6 +290,19 @@ def test_calculate_scaling(self, Crystallizer_frame):
...
+        m.fs.unit.costing = UnitModelCostingBlock(
+            default={
+                "flowsheet_costing_block": m.fs.costing,
+                "costing_method_arguments": {
+                    "cost_type": CrystallizerCostType.mass_basis
+                },
+            },
+        )
+        m.fs.costing.cost_process()
+
```
bknueven, line 305: Are these tests in the right place?

On one hand, our general philosophy has been to put tests in the same directory as the code, i.e., in the `watertap/costing/tests/` directory. On the other hand, breaking them out into the unit model tests makes a huge amount of sense as well -- we already have the code to set up the unit model, etc.

bknueven:
> @OOAmusat Ben's response was: It’s these lines: https://github.com/watertap-org/watertap/pull/515/files#diff-b8409e0826e208143b2dc914e86c9f79da1e396679bd80ab49edb430ead318e1R201-R236
> 
> These new parameters appear on every costing block. Therefore we should expect an increase in the number of Vars for any flowsheet using the watertap costing package.

That said, none of the other flowsheets *check* the number of variables. I would be okay with eliminating the check altogether. And the LSRRO tests will need to be re-done in the near future anyways.

watertap/costing/watertap_costing_package.py
OOAmusat: Done.

watertap/unit_models/tests/test_crystallizer.py
OOAmusat: I could do this. However, are we sure that is how we want to handle the costing tests? That just seems to add a lot of unnecessary overhead to me.

bknueven: I'm happy with the tests as they are because the costing tests are *not* currently centralized in `watertap/costing/tests/`.

Looking at the code I'm not sure centralizing *is* the right idea -- perhaps it should be with the unit test as these are.

watertap/costing/watertap_costing_package.py
```diff
@@ -586,6 +629,84 @@ def cost_caoh2_mixer(blk):
...
+            or cost_type == CrystallizerCostType.mass_basis
+        ):
+            WaterTAPCostingData.cost_crystallizer_by_crystal_mass(blk)
+        elif cost_type == CrystallizerCostType.volume_basis:
+            WaterTAPCostingData.cost_crystallizer_by_volume(blk)
+        else:
+            raise ConfigurationError(
+                f"{blk.unit_model.name} received invalid argument for cost_type:"
+                f" {cost_type}. Argument must be a member of the CrystallizerCostType Enum."
+            )
```
bknueven, line 651: I do test that this error is raised for the other methods here: https://github.com/watertap-org/watertap/blob/main/watertap/costing/tests/test_watertap_costing_package.py

watertap/unit_models/crystallizer.py
```diff
@@ -664,6 +664,10 @@ def initialize(
         blk.properties_in.release_state(flags, outlvl=outlvl)
         init_log.info("Initialization Complete: {}".format(idaeslog.condition(res)))
 
+    def get_costing(self, module=None, **kwargs):
+        self.costing = Block()
+        module.Crystallization_costing(self.costing)
```
bknueven: I'm not sure this code is strictly necessary -- we do not currently use the "old" IDAES costing framework anywhere in WaterTAP, and we should probably consider removing this method from our other unit models.

bknueven (commented):
A few suggestions for increasing the coverage.

watertap/unit_models/crystallizer.py
OOAmusat: Done

watertap/costing/watertap_costing_package.py
OOAmusat: Test added.

bknueven (approved):
I cannot comment on the two costing models, but their implementation looks good.

adam-a-a (approved):
LGTM

PR #517, opened 2022-05-02
Modify MABR model for AMO project

luohezhiming:
## Summary/Motivation:

Modify MABR model for project 1598 of AMO

- Modify blower size based on the membrane size of MABR
- Modify blower energy usage based on air flow rate

### Legal Acknowledgement


luohezhiming:
@bknueven @adam-a-a The checks for macOS setup failed, it seems that one solution value is inconsistent, but all other checks passed. Do you have idea on this?

bknueven:
> @bknueven @adam-a-a The checks for macOS setup failed, it seems that one solution value is inconsistent, but all other checks passed. Do you have idea on this?

```python
>       assert pytest.approx(1.414246e-13, rel=1e-2) == value(
            model.fs.unit.properties_byproduct[0].flow_vol
        )
E       assert 1.414246e-13 ± 1.0e-12 == 2.9013322984899295e-12
E         comparison failed
E         Obtained: 2.9013322984899295e-12
E         Expected: 1.414246e-13 ± 1.0e-12

watertap/unit_models/zero_order/tests/test_mabr_zo.py:148: AssertionError
```

If this value is supposed to be indistinguishable from `0`, I would suggest we compare against `0` with an absolute tolerance. If not, then just use the `requires_idaes_solver` pytest mark.

luohezhiming:
> > @bknueven @adam-a-a The checks for macOS setup failed, it seems that one solution value is inconsistent, but all other checks passed. Do you have idea on this?
> 
> ```python
> >       assert pytest.approx(1.414246e-13, rel=1e-2) == value(
>             model.fs.unit.properties_byproduct[0].flow_vol
>         )
> E       assert 1.414246e-13 ± 1.0e-12 == 2.9013322984899295e-12
> E         comparison failed
> E         Obtained: 2.9013322984899295e-12
> E         Expected: 1.414246e-13 ± 1.0e-12
> 
> watertap/unit_models/zero_order/tests/test_mabr_zo.py:148: AssertionError
> ```
> 
> If this value is supposed to be indistinguishable from `0`, I would suggest we compare against `0` with an absolute tolerance. If not, then just use the `requires_idaes_solver` pytest mark.

Thanks, it should be zero.

watertap/unit_models/zero_order/mabr_zo.py
```diff
@@ -37,14 +37,89 @@ def build(self):
...
         )
 
-        self._fixed_perf_vars.append(self.blower_size)
+        @self.Constraint(
+            self.flowsheet().time,
+            doc="Constraint for blower size",
+        )
+        def blower_size_consumption(b, t):
+            return b.blower_size == pyunits.convert(
+                b.properties_treated[t].flow_mass_comp["ammonium_as_nitrogen"]
```
adam-a-a, line 61: Since we are hard-coding this component within the constraint, you'll need to add a check above to ensure that the "ammonium_as_nitrogen" component is provided in the solute list. If not, raise an exception.

luohezhiming: @adam-a-a Good point, but since the `ammonium_as_nitrogen` is the `key_reactant`, so there is already a check in `build_sido_reactive` to ensure that it is must be included.

PR #519, opened 2022-05-05
Enable recursive parameter sweep to attempt to provide a guaranteed number of solutions

k1nshuk:
## Fixes/Addresses:

This PR adds a new function called `recursive_parameter_sweep` which runs the parameter sweep multiple times in an attempt to provide a guaranteed number of valid solutions.

## Summary/Motivation:

Depending on the flowsheet being used, it is possible to get NaNs in the results. This function looks for such "failed" solves, resamples the parameters, and attempts to solve in order to replace these failed solves

- Addition of a new file `recursive_parameter_sweep.py` with the necessary functions to enable this recursion
- All other functions and files are preserved as is.

### Legal Acknowledgement


watertap/tools/recursive_parameter_sweep.py
```diff
@@ -0,0 +1,293 @@
...
+
+    # Set up optimize_kwargs
+    if optimize_kwargs is None:
+        optimize_kwargs = dict()
+    # Set up reinitialize_kwargs
+    if reinitialize_kwargs is None:
+        reinitialize_kwargs = dict()
+
+    n_samples_remaining = copy.deepcopy(req_num_samples)
+    num_total_samples = copy.deepcopy(req_num_samples)
```
bknueven: These look like they should be `int`, no need to deepcopy (they're not mutable).

watertap/tools/recursive_parameter_sweep.py
```diff
@@ -0,0 +1,293 @@
...
+        n_samples_remaining -= global_success_count
+
+        # The total number of samples to generate at the next iteration is a multiple of the total remaining samples
+        safety_factor = 2
+        if success_prob > 0:
+            num_total_samples = int(
+                np.ceil(safety_factor * n_samples_remaining / success_prob)
+            )
+        else:
+            num_total_samples = int(np.ceil(safety_factor * n_samples_remaining))
```
bknueven: Maybe we should just limit how much we "blow up" the sample size at a given iteration, e.g.,
```suggestion
        # The total number of samples to generate at the next iteration is a multiple of the total remaining samples
        scale_factor = 2.0 / max(success_prob, 0.10)
        num_total_samples = int(
            np.ceil(scale_factor * n_samples_remaining)
        )
```

In this way we limit the amount we scale up by 20x in a single iteration...we might pick something more conservative by adjusting either the numerator or denominator.

watertap/tools/recursive_parameter_sweep.py
```diff
@@ -0,0 +1,293 @@
...
+    _create_global_output,
+    _default_optimize,
+    _divide_combinations,
+    _do_param_sweep,
+    _init_mpi,
+    _process_sweep_params,
+    _process_results_filename,
+    _save_results,
+    _write_outputs,
+)
```
bknueven: Please eliminate the unused imports. Normally it's not a big deal, but I think half of these are not used.

watertap/tools/recursive_parameter_sweep.py
```diff
@@ -0,0 +1,293 @@
...
+
+np.set_printoptions(linewidth=200)
```
bknueven: I know we do this in the other file ... what's the reasoning? Generally we importing a module shouldn't have weird side effects like this.

If it's needed, perhaps we can write a little context manager to modify this value for when its needed and reset it automatically.

watertap/tools/recursive_parameter_sweep.py
```diff
@@ -0,0 +1,293 @@
...
+
+
+def _filter_recursive_solves(model, sweep_params, outputs, recursive_local_dict, comm):
+
+    # Figure out how many filtered solves did this rank actually do
+    filter_counter = 0
+    for case, content in recursive_local_dict.items():
+        filter_counter += sum(
+            content["solve_successful"]
+        )  # content["solve_successful"].count(filter_keyword)
```
bknueven: Please remove commented code.

watertap/tools/recursive_parameter_sweep.py
```diff
@@ -0,0 +1,293 @@
...
+
+    # Populate local_successful_outputs
+    offset = 0
+    for case_number, content in recursive_local_dict.items():
+        # Filter all of the sucessful solves
+        optimal_indices = list(
+            itertools.compress(
+                range(len(content["solve_successful"])), content["solve_successful"]
+            )
+        )  # [idx for idx, status in enumerate(content["solve_status"]) if status == "optimal"]
```
bknueven: Commented code.

In this case the commented code is also easier to read -- I would suggest using it in place of `itertools.compress` ... which I had to look up.

bknueven:
```suggestion
        optimal_indices = [idx for idx, success in enumerate(content["solve_successful"]) if success]
```
More readable, IMHO.

watertap/tools/tests/test_recursive_parameter_sweep.py
```diff
@@ -0,0 +1,377 @@
...
+    build,
+    set_operating_conditions,
+    initialize_system,
+    solve,
+    optimize,
+)
+
+# import logging
+# logging.getLogger('pyomo.core').setLevel(logging.ERROR)
+
```
bknueven: commented code

bknueven (commented):
This looks pretty good -- I have a few relatively minor comments.

Big picture: this feature needs documentation, but given we're planning to refactor the parameter sweep tool I think it can wait.

watertap/tools/recursive_parameter_sweep.py
k1nshuk: Setting a wider line width for numpy print options simply enables easier debugging of code using print statements as it doesn't wrap numpy arrays. The 200 character width is definitely arbitrary in the sense that it was sufficient in making sure the output was easily readable for debugging and understanding the code. Since we don't really print numpy arrays during analysis runs, it doesn't affect the outputs.

I am open to having a mechanism to change the line width value. That said I just disagree with the default value of 75 characters in this day and age of HD and 4K screens.

bknueven: If the answer is a matter of debugging print statements, I would suggest we eliminate this line in this file and `parameter_sweep.py`.

It's easy enough to change when we're working on the code an removing it once done, or write a one-off function for printing which uses the context manager instead: https://numpy.org/doc/stable/reference/generated/numpy.printoptions.html#numpy.printoptions. We should not modify the options in a widely-used third-party package globally without a very compelling reason.

k1nshuk: Removed it in the most recent commit.

bknueven (approved):

eyoung55 (approved):

PR #511, opened 2022-04-28
Experimental: Changing some bounds to 0

bknueven:
## Changes proposed in this PR:
-
-

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #520, opened 2022-05-05
Update idaes-pse requirement to 2.0.0.dev0.watertap.2022.05.18 tag

lbianchi-lbl:
## Summary/Motivation:

- Assess the impact of the latest IDAES changes before the IDAES May release in preparation for the WaterTAP June release

### Legal Acknowledgement

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -143,6 +143,7 @@ jobs:
       if: matrix.cov_report
       uses: codecov/codecov-action@v2
     - name: Test documentation code
+      if: 'false'  # TODO: re-enable this ASAP
       run: |
         make -C docs doctest -d
     # TODO: this should be moved to a dedicated job/workflow
--- a/setup.py
+++ b/setup.py
@@ -43,8 +43,7 @@ SPECIAL_DEPENDENCIES_FOR_PRERELEASE = [
     # update with a tag from the nawi-hub/idaes-pse
     # when a version of IDAES newer than the latest stable release from PyPI
     # will become needed for the watertap development
-    # "idaes-pse[prerelease] @ https://github.com/watertap-org/idaes-pse/archive/1.12.1.watertap.2022.02.04.zip",
-    "idaes-pse[prerelease] @ https://github.com/watertap-org/idaes-pse/archive/1.14.0dev0.watertap.2022.04.16.zip"
+    "idaes-pse[prerelease] @ https://github.com/watertap-org/idaes-pse/archive/2.0.0.dev0.watertap.2022.05.05.zip"
 ]
 
 # Arguments marked as "Required" below must be included for upload to PyPI.
--- a/setup.py
+++ b/setup.py
@@ -43,7 +43,7 @@ SPECIAL_DEPENDENCIES_FOR_PRERELEASE = [
     # update with a tag from the nawi-hub/idaes-pse
     # when a version of IDAES newer than the latest stable release from PyPI
     # will become needed for the watertap development
-    "idaes-pse[prerelease] @ https://github.com/watertap-org/idaes-pse/archive/2.0.0.dev0.watertap.2022.05.05.zip"
+    "idaes-pse[prerelease] @ https://github.com/watertap-org/idaes-pse/archive/2.0.0.dev0.watertap.2022.05.18.zip"
 ]
 
 # Arguments marked as "Required" below must be included for upload to PyPI.
--- a/watertap/core/tests/test_zero_order_base.py
+++ b/watertap/core/tests/test_zero_order_base.py
@@ -17,7 +17,7 @@ import pytest
 
 from idaes.core import declare_process_block_class, FlowsheetBlock
 from idaes.core.util.exceptions import ConfigurationError
-from pyomo.environ import ConcreteModel, Var
+from pyomo.environ import ConcreteModel, Var, units as pyunits
 
 from watertap.core import ZeroOrderBaseData
 from watertap.core import WaterParameterBlock
@@ -433,7 +433,14 @@ class TestZOBase:
                 "Mass Concentration C": pytest.approx(250.000, rel=1e-4),
                 "Mass Concentration H2O": pytest.approx(250.000, rel=1e-4),
                 "Volumetric Flowrate": pytest.approx(0.004, rel=1e-4),
-            }
+            },
+            "Units": {
+                "Mass Concentration A": getattr(pyunits.pint_registry, "kg/m**3"),
+                "Mass Concentration B": getattr(pyunits.pint_registry, "kg/m**3"),
+                "Mass Concentration C": getattr(pyunits.pint_registry, "kg/m**3"),
+                "Mass Concentration H2O": getattr(pyunits.pint_registry, "kg/m**3"),
+                "Volumetric Flowrate": getattr(pyunits.pint_registry, "m**3/s"),
+            },
         }
 
         assert stable.to_dict() == expected
--- a/watertap/edb/tests/test_data_model.py
+++ b/watertap/edb/tests/test_data_model.py
@@ -21,7 +21,7 @@ from typing import Dict
 from pyomo.environ import units as pyunits
 
 # Used to test IDAES config -> DataWrapper
-from idaes.generic_models.properties.core.pure import Perrys
+from idaes.models.properties.model_properties.pure import Perrys
 from idaes.generic_models.properties.core.pure.NIST import NIST
 from idaes.generic_models.properties.core.phase_equil.forms import fugacity
 from idaes.generic_models.properties.core.reactions.equilibrium_forms import (
--- a/watertap/edb/tests/test_data_model.py
+++ b/watertap/edb/tests/test_data_model.py
@@ -21,7 +21,7 @@ from typing import Dict
 from pyomo.environ import units as pyunits
 
 # Used to test IDAES config -> DataWrapper
-from idaes.models.properties.model_properties.pure import Perrys
+from idaes.models.properties.modular_properties.pure import Perrys
 from idaes.generic_models.properties.core.pure.NIST import NIST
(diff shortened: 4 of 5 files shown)


ksbeattie (dismissed):
LGTM

bknueven:
Lots of issues here -- but maybe we could collectively agree that checking the output of `report()` is bad and we should never do it again? And also eliminate the tests that *do* check the output. This was a major source of failing tests on #511 which had to be addressed.

IDAES can test that the output is reasonable, I think WaterTAP should just call it to make sure it doesn't error-out (and *not* check the output).

andrewlee94:
@bknueven I am torn on checking the output of `report` - for one, if you don't check the output then what are you really checking? Also, it is relatively unusual for the `report` method itself to change (this is the first time since it was implemented). On the other hand, the values being reported are subject to noise at times and adding the appropriate code to test the approximate values would be excessive.

bknueven:
@andrewlee94 we would still be testing if the values returned from `_get_stream_contents` and `_get_performance_contents` are interpretable by the `report` method itself.

Alternatively, we could test `_get_performance_contents` and `_get_stream_contents` directly -- which would have the benefit of being easier to do relative check with.

lbianchi-lbl:
> Lots of issues here -- but maybe we could collectively agree that checking the output of `report()` is bad and we should never do it again? And also eliminate the tests that _do_ check the output. This was a major source of failing tests on #511 which had to be addressed.
> 
> IDAES can test that the output is reasonable, I think WaterTAP should just call it to make sure it doesn't error-out (and _not_ check the output).

I'm not familiar enough with the structure of the code to have a clear idea of the tradeoffs of checking something else, but I definitely agree that we should try to find an alternative to using the human-formatted output of `report()`.

In general, as long as the data we want to look at has a sufficiently simple structure (e.g. a JSON-able `dict` without too much nesting), pytest should "just work" (including, AFAIK, supporting `pytest.approx()` without us needing to wrap every single value).

bknueven:
> In general, as long as the data we want to look at has a sufficiently simple structure (e.g. a JSON-able `dict` without too much nesting), pytest should "just work" (including, AFAIK, supporting `pytest.approx()` without us needing to wrap every single value).

Generally speaking, the correctness of the values should be checked in the other tests, which is what makes this situation so weird.

The contents of `_get_performance_contents` is a dictionary whose leaf values are Pyomo objects ... which is probably an issue for `pytest`, but it probably wouldn't be difficult to write a checker for.

aladshaw3:
The 2 chemistry example errors seem to be tied into some change in how phase equilibrium problems are initialized in IDAES. 

` idaes.core.util.exceptions.InitializationError: fs.unit.control_volume.properties_out Unexpected degrees of freedom during initialization at phase equilibrium step: 2`

@andrewlee94 Has something in IDAES changed recently with how the Generic Properties handles phase equilibrium? 

It is worth noting that the full `solve` step did not fail, just the initialization step. The other failure was due to what appear to be some minor py.approx diffs.

aladshaw3:
@andrewlee94  Here is exactly where the error is coming from in IDAES from `generic_property.py` lines 1490-1537

Somehow, there are now 2 additional degrees of freedom at this point. 

        n_cons = 0
        dof = 0
        skip = False
        Tfix = {}  # In enth based state defs, need to also fix T until later
        for k in blk.keys():
            if blk[k].params.config.phase_equilibrium_state is not None and (
                not blk[k].config.defined_state or blk[k].always_flash
            ):
                if not blk[k].temperature.fixed:
                    blk[k].temperature.fix()
                    Tfix[k] = True
                for c in blk[k].component_objects(Constraint):
                    # Activate common constraints
                    if c.local_name in (
                        "total_flow_balance",
                        "component_flow_balances",
                        "sum_mole_frac",
                        "equilibrium_constraint",
                        "phase_fraction_constraint",
                    ):
                        c.activate()
                    if c.local_name == "log_mole_frac_phase_comp_eqn":
                        c.activate()
                        for p, j in blk[k].params._phase_component_set:
                            calculate_variable_from_constraint(
                                blk[k].log_mole_frac_phase_comp[p, j],
                                blk[k].log_mole_frac_phase_comp_eqn[p, j],
                            )
                for pp in blk[k].params._pe_pairs:
                    # Activate formulation specific constraints
                    blk[k].params.config.phase_equilibrium_state[
                        pp
                    ].phase_equil_initialization(blk[k], pp)

            n_cons += number_activated_constraints(blk[k])
            dof += degrees_of_freedom(blk[k])
            if degrees_of_freedom(blk[k]) < 0:
                # Skip solve if DoF < 0 - this is probably due to a
                # phase-component flow state with flash
                skip = True

        if n_cons > 0 and not skip:
            if dof > 0:
                raise InitializationError(
                    f"{blk.name} Unexpected degrees of freedom during "
                    f"initialization at phase equilibrium step: {dof}."
                )


                idaes.core.util.exceptions.InitializationError: fs.unit.control_volume.properties_out Unexpected degrees of freedom during initialization at phase equilibrium step: 2.

andrewlee94:
@aladshaw3 I think the main change is that we got more strict in checking these types of things - i.e. this issue existed before but we didn't raise an Exception for it.

As for what to do about it, I am not sure. Experience would seem to indicate that in this case we can get away with having degrees of freedom there, but on the other hand that is never a good idea (hence the exception).

aladshaw3:
> @aladshaw3 I think the main change is that we got more strict in checking these types of things - i.e. this issue existed before but we didn't raise an Exception for it.
> 
> As for what to do about it, I am not sure. Experience would seem to indicate that in this case we can get away with having degrees of freedom there, but on the other hand that is never a good idea (hence the exception).

So, my question here is then, why are there 2 dof during initialization, but before I call initialization the dof = 0? This doesn't appear to be an issue that is resolvable on the watertap side.

andrewlee94:
@aladshaw3 Yes, the issue is something deeper. I would need to look deeper into the problem however to work out why. The first step would be to work out what the unexpected degrees of freedom are.

aladshaw3:
@lbianchi-lbl  

Another error in the `test_data_model` file from the EDB stuff is caused by the depreciation/renaming of methods in IDAES. 

If you go into file `edb/tests/test_data_model.py`, all you need to do for the time being is replace line 24 with the following:

`from idaes.models.properties.modular_properties.pure import Perrys`

aladshaw3:
@lbianchi-lbl  

Unfortunately, that change in `test_data_model.py` has caused an issue somewhere else. It might be best to just undo that last change and then skip the one test in `test_data_model.py` that was failing and then fix it later. 

The issue here is driven entirely by the renaming and moving of methods in IDAES. Fixing that might be best reserved for its own PR (or we would update everything in this PR). 

==================================== ERRORS ====================================
____________ ERROR collecting watertap/edb/tests/test_data_model.py ____________
ImportError while importing test module '/home/runner/work/watertap/watertap/watertap/edb/tests/test_data_model.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/share/miniconda/envs/watertap-dev/lib/python3.7/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
watertap/edb/tests/test_data_model.py:24: in <module>
    from idaes.models.properties.model_properties.pure import Perrys
E   ModuleNotFoundError: No module named 'idaes.models.properties.model_properties'

andrewlee94:
@aladshaw3 @lbianchi-lbl The issue in the error above is that the import was misnamed - it is not `model_properties` but `modular_properties`.

aladshaw3:
If we need this PR in quickly, we should just skip the last 3 tests on `test_recarbonation_process.py` (located in `examples/chemistry/tests`) and open an issue to figure out what is going wrong. I don't think that will be a quick fix.

aladshaw3:
This PR needs to come after #538 and will also require a new [PR ](https://github.com/IDAES/idaes-pse/pull/867)from IDAES, which means we will need to update our own fork of IDAES for WaterTAP after that IDAES PR is merged.

lbianchi-lbl:
> This PR needs to come after #538 and will also require a new [PR ](https://github.com/IDAES/idaes-pse/pull/867)from IDAES, which means we will need to update our own fork of IDAES for WaterTAP after that IDAES PR is merged.

IDAES/idaes-pse#867 has been merged, so I've created a [new tag](https://github.com/watertap-org/idaes-pse/releases/tag/2.0.0.dev0.watertap.2022.05.18) for `watertap-org/idaes-pse` and updated the requirement in `setup.py` in this PR accordingly.

aladshaw3:
Looking over the most recent error logs shows there are now only 4 errors. 

3 of those errors should be fixed by PR #538 

The last error is in one of the zero order models `test_zero_order_base.py` during a dict comparison. The dict being compared has an additional argument compared to what is expected. 

See error log below.
[zero_order_dict_error.txt](https://github.com/watertap-org/watertap/files/8729094/zero_order_dict_error.txt)
 
Looks like the update has added something that comes out of the `_get_stream_table_contents()` function to include a sub-dict containing units. Quickest fix is probably to just compare the `test port` sub-dicts from each instead of the full dict. Or to add in the `Units` sub-dict to the `expected` dict.

bknueven:
> The last error is in one of the zero order models `test_zero_order_base.py` during a dict comparison. The dict being compared has an additional argument compared to what is expected.
> 
> See error log below. [zero_order_dict_error.txt](https://github.com/watertap-org/watertap/files/8729094/zero_order_dict_error.txt)
> 
> Looks like the update has added something that comes out of the `_get_stream_table_contents()` function to include a sub-dict containing units. Quickest fix is probably to just compare the `test port` sub-dicts from each instead of the full dict. Or to add in the `Units` sub-dict to the `expected` dict.

~~I think we should fix this properly as part of this PR, meaning we add the `Units` to the sub-dict.~~ I fixed this in 945a5e2.

aladshaw3 (dismissed):

bknueven (approved):

aladshaw3 (approved):

PR #516, opened 2022-05-02
[WIP] Initial model framework for 1D electrodialysis

aladshaw3:
## Fixes/Addresses:

Partially addresses #489 

## Summary/Motivation:

This is being developed in parallel with a 0D electrodialysis model. The 1D model is being built and tested against both the ion property package in WaterTAP and the Generic Property Package in IDAES. 

## Changes proposed in this PR:
- Modified variable bounds on `pressure_osm_phase` in ion prop pack
- Added framework model for 1D electrodialysis
- Added unit tests for 1D electrodialysis 

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #523, opened 2022-05-06
Electrodialysis 0 d

lbibl:
## Summary/Motivation:
- This branch adds the first version of Electrodialysis_0D unit model, including the main model file and test file. 

### Legal Acknowledgement


watertap/unit_models/electrodialysis_0d.py
```diff
@@ -0,0 +1,938 @@
...
+        # Create essential sets.
+        self.membrane_set = Set(initialize=["cem", "aem"])
+
+        # Create unit model parameters and vars
+        self.water_density = Param(
+            initialize=1000,
+            mutable=False,
+            units=pyunits.kg * pyunits.m**-3,
+            doc="density of water",
+        )
```
aladshaw3, line 204: Remind me again, why is this here? This should be a property calculated in the property package. What is this used for and why can't we use the property package?

watertap/unit_models/electrodialysis_0d.py
```diff
@@ -0,0 +1,938 @@
...
+        self.water_MW = Param(
+            initialize=18.015e-3,
+            mutable=False,
+            units=pyunits.kg * pyunits.mole**-1,
+            doc="molecular weight of water",
+        )
```
aladshaw3: This is a piece of information that should be coming from the property package. Why do we need this here?

watertap/unit_models/electrodialysis_0d.py
```diff
@@ -0,0 +1,938 @@
...
+        self.cell_pair_num = Var(
+            initialize=1,
+            domain=NonNegativeIntegers,
+            bounds=(1, Inf),
```
aladshaw3:
```suggestion
            bounds=(1, 1000),
```
We should actually bound this variable to something reasonable.

watertap/unit_models/electrodialysis_0d.py
```diff
@@ -0,0 +1,938 @@
...
+            bounds=(1e-6, 1e-1),
+            units=pyunits.meter,
+            doc="Membrane thickness",
+        )
+        self.solute_diffusivity_membrane = Var(
+            self.membrane_set,
+            self.config.property_package.ion_set
+            | self.config.property_package.solute_set,
+            initialize=1e-10,
+            # bounds=(1e-11, 1e-6),
```
aladshaw3:
```suggestion
            bounds=(1e-11, 1e-6),
```
Why is this being commented out? Is the lower bound low enough for this variable?

watertap/unit_models/electrodialysis_0d.py
```diff
@@ -0,0 +1,938 @@
...
+        self.ion_trans_number_membrane = Var(
+            self.membrane_set,
+            self.config.property_package.ion_set,
+            bounds=(0, 1),
+            units=pyunits.dimensionless,
+            doc="Ion transference number in the membrane phase",
+        )
+        self.water_trans_number_membrane = Var(
+            self.membrane_set,
+            initialize=5,
```
aladshaw3, line 259:
```suggestion
            initialize=1,
```
Don't initialize a value outside of its bounds.

watertap/unit_models/electrodialysis_0d.py
```diff
@@ -0,0 +1,938 @@
...
+            bounds=(0, 1),
+            units=pyunits.dimensionless,
+            doc="Transference number of water in membranes",
+        )
+        self.water_permeability_membrane = Var(
+            self.membrane_set,
+            initialize=5,
```
aladshaw3:
```suggestion
            initialize=1e-14,
```
We should not be giving an unreasonable default value for any variable.

watertap/unit_models/electrodialysis_0d.py
```diff
@@ -0,0 +1,938 @@
...
+            self.flowsheet().config.time,
+            initialize=0.9,
+            bounds=(0, 1),
+            units=pyunits.dimensionless,
+            doc="The overall current efficiency for deionizaiton",
+        )
+        self.power_electrical = Var(
+            self.flowsheet().config.time,
+            initialize=1,
+            bounds=(0, Inf),
```
aladshaw3:
```suggestion
            bounds=(0, 1e5),
```
We should try and provide a reasonable upper limit (whatever that limit may be)

watertap/unit_models/electrodialysis_0d.py
```diff
@@ -0,0 +1,938 @@
...
+            domain=NonNegativeReals,
+            units=pyunits.watt,
+            doc="Electrical power consumption of a stack",
+        )
+        self.specific_power_electrical = Var(
+            self.flowsheet().config.time,
+            initialize=1e6,
+            bounds=(0, Inf),
```
aladshaw3:
```suggestion
            bounds=(0, 1e5),
```
Provide reasonable upper limits to variables.

watertap/unit_models/electrodialysis_0d.py
```diff
@@ -0,0 +1,938 @@
...
+    **default** - False. The filtration unit does not have defined volume, thus
+    this must be False.""",
+        ),
+    )
+
+    CONFIG.declare(
+        "operation_mode",
+        ConfigValue(
+            default="Constant Current",
+            domain=In(["Constant Current", "Constant Voltage"]),
```
aladshaw3:
```suggestion
            default="const_current",
            domain=In(["const_current", "const_voltage"]),
```
Avoid spaces in names of options, variables, and objects. It may also be better if these were done as an Enum, rather than strings.

watertap/unit_models/electrodialysis_0d.py
```diff
@@ -0,0 +1,938 @@
...
+            )
+
+        @self.Constraint(
+            self.flowsheet().config.time,
+            self.config.property_package.phase_list,
+            self.config.property_package.component_list,
+            doc="Equation for electrical migration flux_in",
+        )
+        def eq_elec_migration_flux_in(self, t, p, j):
+            if j == "H2O":
```
aladshaw3, line 448: Because the unit model specifically requires a component named 'H2O', we may want to check the property package component list during the build of the object and raise a configuration error if 'H2O' is not present. This way, the user will get a clear message from the unit model saying why things have failed and telling them they need 'H2O' as a component in their property package.

watertap/unit_models/electrodialysis_0d.py
```diff
@@ -0,0 +1,938 @@
...
+                        "Liq", j
+                    ]
+                    * self.config.property_package.charge_comp[j]
+                    for j in self.config.property_package.cation_set
+                )
+                * Constants.faraday_constant
+            )
+
+    # initialize method
+    def initialize(
```
aladshaw3:
```suggestion
    def initialize_build(
```
I believe newer version of IDAES will require the `initialize` function for custom unit models to be called `initialize_build`. It is basically the same function, but has some different args. Have a look at #516 for an example.

watertap/unit_models/electrodialysis_0d.py
```diff
@@ -0,0 +1,938 @@
...
+            sf = (
+                iscale.get_scaling_factor(self.solute_diffusivity_membrane)
+                / iscale.get_scaling_factor(self.membrane_thickness)
+                * iscale.get_scaling_factor(
+                    self.concentrate_channel.properties_in[ind[0]].conc_mol_phase_comp[
+                        ind[1], ind[2]
+                    ]
+                )
+            )
+            # print('########', ind, sf)
```
aladshaw3:
```suggestion
```
Don't need this line anymore.

aladshaw3 (changes requested):
Looking good so far. Some additional changes suggested below.

watertap/unit_models/tests/test_electrodialysis_0d.py
```diff
@@ -0,0 +1,672 @@
...
+        m.fs.unit.config.operation_mode = "Constant Voltage"
+        assert len(m.fs.unit.config) == 7
+        assert not m.fs.unit.config.dynamic
+        assert not m.fs.unit.config.has_holdup
+        assert m.fs.unit.config.operation_mode == "Constant Voltage"
+        assert m.fs.unit.config.material_balance_type == MaterialBalanceType.useDefault
+        assert (
+            m.fs.unit.config.momentum_balance_type == MomentumBalanceType.pressureTotal
+        )
+        assert m.fs.unit.config.property_package is m.fs.properties
```
aladshaw3: These asserts should be in the `test_build_model` function below.

watertap/unit_models/tests/test_electrodialysis_0d.py
```diff
@@ -0,0 +1,672 @@
...
+        m.fs.unit.config.operation_mode = "Constant Current"
+        assert len(m.fs.unit.config) == 7
+        assert not m.fs.unit.config.dynamic
+        assert not m.fs.unit.config.has_holdup
+        assert m.fs.unit.config.operation_mode == "Constant Current"
+        assert m.fs.unit.config.material_balance_type == MaterialBalanceType.useDefault
+        assert (
+            m.fs.unit.config.momentum_balance_type == MomentumBalanceType.pressureTotal
+        )
+        assert m.fs.unit.config.property_package is m.fs.properties
```
aladshaw3: These assertions should be in the `test_build_model` function below

watertap/unit_models/tests/test_electrodialysis_0d.py
```diff
@@ -0,0 +1,672 @@
...
```
aladshaw3: These assertions should be in the `test_build_model` function below

watertap/unit_models/tests/test_electrodialysis_0d.py
```diff
@@ -0,0 +1,672 @@
...
+# through Lawrence Berkeley National Laboratory, Oak Ridge National
+# Laboratory, National Renewable Energy Laboratory, and National Energy
+# Technology Laboratory (subject to receipt of any required approvals from
+# the U.S. Dept. of Energy). All rights reserved.
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and license
+# information, respectively. These files are also available online at the URL
+# "https://github.com/nawi-hub/proteuslib/"
+#
+###############################################################################
```
aladshaw3: This is the old copyright statement from ProteusLib. This needs to be replaced with the copyright from WaterTAP (the one you used in the unit model). If this was copy-pasted from another file, then it needs to be changed in that other file as well. 

If you could, also replace the copyright statement in the file you got this from in this PR.

watertap/unit_models/electrodialysis_0d.py
lbibl: This was left in the unit model because it was used to build a unit_model var that does not distinguish the two control_volumes (concentrate and diluate). Taking it from property pack  will have to bind it to `concentrate_channel` or `diluate_channel`. Although that'll be technically the same number, it may make the equation confusing.  So I would prefer to keep it here.

watertap/unit_models/electrodialysis_0d.py
lbibl: I removed this one in the new commit and take the value from property pack.  No issue for this one because mw_comp is a universal param in the ion pp.

watertap/unit_models/electrodialysis_0d.py
lbibl: I kept the strings but removed the space.   Maybe revisit the option of Enum if this config becomes more useful (for now it is more or less a "label" only).

watertap/unit_models/tests/test_electrodialysis_0d.py
lbibl: Yes, I also corrected the file "test_coag_..."  where I took it from .

watertap/unit_models/electrodialysis_0d.py
lbibl: I believe "H2O" is built as solvent always in the ion property pack. But I did add a `assert` for testing "H2O" in `component_list` in case other property packs are used.

watertap/unit_models/electrodialysis_0d.py
lbibl: I put it back.  It should be a temporary disabling during my debug.  It does not cause a problem till now.

lbibl:
New commits are pushed to address Austin's comments (Thank you for catching some small but important issues!)

watertap/unit_models/electrodialysis_0d.py
aladshaw3: Should we consider adding a temperature dependence to this term as part of a later model refinement?

watertap/unit_models/electrodialysis_0d.py
```diff
@@ -0,0 +1,934 @@
...
+        self.ion_trans_number_membrane = Var(
+            self.membrane_set,
+            self.config.property_package.ion_set,
+            bounds=(0, 1),
+            units=pyunits.dimensionless,
+            doc="Ion transference number in the membrane phase",
+        )
+        self.water_trans_number_membrane = Var(
+            self.membrane_set,
+            initialize=5,
```
aladshaw3, line 259: Is this correct? I thought these we numbers between 0 and 1?

watertap/unit_models/electrodialysis_0d.py
aladshaw3: What I meant was to check for 'H2O' in the build function of this unit model and raise an exception if not present (not to just check in the tests). This is because ideally, we may want to use other property packages with this unit model (if available). 

One thing to understand is that since property packages and unit models are separate entities, we may not always be able to assume what the structure of the property package is. Something to keep in mind for future models and refinement.

aladshaw3 (dismissed):
I had a few other minor comments/responses. Other than that, this looks good to me.

watertap/unit_models/electrodialysis_0d.py
lbibl: I do not think it is technically necessary because water density variation with temperature from 0 to 60 Celsius degree is <2%, which would likely be well below the error of temperature measurements in reality.  The error caused by this would be even much less significant for the ED Perfromance modeling.   I would recommend leaving this out for now but we may revisit this if the model refinement shows this becomes a critical factor to change.

srikanthallu (commented):
Can you provide references from where you got the values used in the testing of the unit_model ?

lbibl:
> Can you provide references from where you got the values used in the testing of the unit_model ?

I put a brief note on the testing scenario reference  in the testing code file.  Formal references will be given in the technical documentation.

aladshaw3 (approved):

srikanthallu (approved):

PR #528, opened 2022-05-11
Removing testing the output of the `report` method

bknueven:
## Summary/Motivation:
Testing the output of report has been a pain point for our tests. Needed for #520.

- Remove testing the output of `report`
- Replace with testing values directly, where not already tested

### Legal Acknowledgement

### DIFF
--- a/watertap/core/tests/test_zero_order_electricity.py
+++ b/watertap/core/tests/test_zero_order_electricity.py
@@ -14,7 +14,6 @@
 Tests for general zero-order property package
 """
 import pytest
-from io import StringIO
 
 from idaes.core import declare_process_block_class, FlowsheetBlock
 from idaes.core.util.model_statistics import degrees_of_freedom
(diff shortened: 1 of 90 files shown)


kurbansitterley (dismissed):
Looks straightforward to me.

andrewlee94:
FYI, IDAES is changing how it does these tests as well: https://github.com/IDAES/idaes-pse/pull/848

watertap/core/tests/test_zero_order_electricity.py
```diff
@@ -116,31 +115,7 @@ def test_solution(self, model):
...
-------------------------------------------------------------------------------------
-    Stream Table
-    Empty DataFrame
-    Columns: []
-    Index: []
-====================================================================================
-"""
-
-        assert output == stream.getvalue()
+        model.fs.unit.report()
```
aladshaw3, line 119: So, I guess my issue with this is that we are no longer checking the values of the solutions on output. So this test is effectively testing nothing now. 

Are there assertions for the values that are getting reported? I saw some were added, but do we have them everywhere?

bknueven: Most of the time, no. For example, this test already has a `test_solution` method for the "Electricity Demand".

I'm adding explicit solution tests in a few more places that I missed where we don't check it otherwise.

bknueven:
> FYI, IDAES is changing how it does these tests as well: [IDAES/idaes-pse#848](https://github.com/IDAES/idaes-pse/pull/848)

For units that implement `_get_performance_contents` or `_get_stream_table_contents` and were explicitly checking the output of `report` before, I've implemented similar tests as part of this PR.

aladshaw3 (approved):
LGTM

TimBartholomew (approved):
LGTM

PR #530, opened 2022-05-11
Modify MABR model for AMO project

luohezhiming:
## Fixes/Addresses:

This PR is an updated version of #517 , the old one is closed.

## Summary/Motivation:
Modify MABR model for project 1598 of AMO

-Modify blower size based on the membrane size of MABR
-Modify blower energy usage based on air flow rate

### Legal Acknowledgement


luohezhiming:
@adam-a-a I've added the tests for key reactant in this new PR. I am sorry that I messed up the two branches AMO_1598 and amo_1598 (#517). When I realized this, I've already pushed my fork. I do not find an easy way to handle this, so I decided to post this PR and close #517.

watertap/unit_models/zero_order/tests/test_mabr_zo.py
```diff
@@ -145,16 +145,10 @@ def test_solution(self, model):
...
-        assert pytest.approx(4e-12, rel=1e-2) == value(
+        assert pytest.approx(0, abs=1e-9) == value(
             model.fs.unit.properties_byproduct[0].flow_vol
         )
-        assert pytest.approx(200, rel=1e-5) == value(
-            model.fs.unit.properties_byproduct[0].conc_mass_comp["bod"]
-        )
-        assert pytest.approx(200, rel=1e-5) == value(
-            model.fs.unit.properties_byproduct[0].conc_mass_comp["nitrate"]
-        )
```
tarnold17, line 156: Why have the two assert statements on lines 151 - 156 been removed?

luohezhiming: The byproduct stream is zero for this unit, so no need for these composition tests.

tarnold17: Are you saying that for `MABRZO`, the byproduct stream always has a flowrate of zero?

luohezhiming: Yeah, the byproduct stream flowrate is zero based on their provided data.

tarnold17: In that case, does it make more sense to build this model as a SISO instead of a SIDO?

watertap/unit_models/zero_order/mabr_zo.py
```diff
@@ -37,14 +37,89 @@ def build(self):
...
+        def electricity_consumption(b, t):
+            return b.electricity[t] == (
+                b.energy_electric_flow_vol_inlet
+                * pyunits.convert(
+                    b.air_flow_vol[t], to_units=pyunits.m**3 / pyunits.hour
+                )
+            )
 
-        self._perf_var_dict["Blower Size (m^2)"] = self.blower_size
+        self._fixed_perf_vars.append(self.energy_electric_flow_vol_inlet)
```
adam-a-a, line 122: I don't see this in the yaml file, so is this actually being fixed to any value?

watertap/unit_models/zero_order/tests/test_mabr_zo.py
adam-a-a: I think it's better to keep as a SIDO in the case where alternative data is fed into the unit, allowing for some byproduct flowrate.

watertap/unit_models/zero_order/tests/test_mabr_zo.py
```diff
@@ -187,21 +182,22 @@ def test_report(self, model):
...
-                       Electricity Demand :   0.57380 : False : (0, None)
-                    Electricity Intensity : 0.0069300 :  True : (None, None)
-    Reaction Extent [ammonium_to_nitrate] :    1.4000 : False : (None, None)
-    Solute Removal [ammonium_as_nitrogen] :    0.0000 :  True : (0, None)
-                     Solute Removal [bod] :    0.0000 :  True : (0, None)
-                 Solute Removal [nitrate] :    0.0000 :  True : (0, None)
-                     Solute Removal [tss] :    0.0000 :  True : (0, None)
-                           Water Recovery :    1.0000 :  True : (1e-08, 1.0000001)
+    Key                                   : Value      : Fixed : Bounds
+                              Blower Size : 3.9877e+07 : False : (0, None)
```
adam-a-a: I'm not sure about this blower size in m**2. Mathematically, it works out in the relationships that we have, but what area is this really? I guess it depends on the `air_flow_rate` parameter. Does it represent volume of air flowrate required per effective area of the tank or reactor receiving the air? The way I see it, it can only be that or the effective cross-sectional area of the blower outlet, which seems unlikely due to the large size.

adam-a-a: Since we know at the least that this just serves as a sizing parameter for getting the blower flowrate, maybe we should leave it out of the report unless we rename `blower size` to something else. Perhaps reactor area?

adam-a-a (dismissed):
LGTM- just a small suggestion to rename blower size, since that area is more likely representative of the reactor area, and representing blower size as an area doesn't make much sense (unless it represents cross-sectional area of the outlet, which I think it doesn't).

watertap/unit_models/zero_order/tests/test_mabr_zo.py
luohezhiming: The blower size is the sizing variable for calculating blower cost. You are correct that it represents the volume of air flowrate required per effective area of the reacotr receiving the air. I calculated this value by the nitrogen removal flowrate as suggested, but I don't know if this large value make any sense. Probably I rename it as `reactor area` and doesn't report it.

luohezhiming:
> LGTM- just a small suggestion to rename blower size, since that area is more likely representative of the reactor area, and representing blower size as an area doesn't make much sense (unless it represents cross-sectional area of the outlet, which I think it doesn't).

This is pretty good point, the area is just the effectivee area as a sizing variable. Actually I still doubt this value but I rename it as `reactor area`. And I also deleted the report as Kurby's PR.

watertap/unit_models/zero_order/mabr_zo.py
```diff
@@ -37,14 +37,87 @@ def build(self):
...
-        self.blower_size = Var(
+        self.nitrogen_removal_rate = Var(
+            units=pyunits.g / pyunits.m**2 / pyunits.day,
+            bounds=(0, None),
+            doc="Nitrogen removal rate per day",
+        )
+
+        self._fixed_perf_vars.append(self.nitrogen_removal_rate)
+
+        self.reactor_area = Var(
```
adam-a-a, line 49: One last comment- let's include it in the variables that will be reported so that when we get to the use case analysis, we can easily see this value and confirm whether it makes sense with the group. If anything, we can always refine as needed, but better to have this variable showing up in a report rather than hiding behind the scenes.

watertap/unit_models/zero_order/tests/test_mabr_zo.py
```diff
@@ -173,44 +166,27 @@ def test_conservation(self, model):
...
-    Mass Concentration H2O                    434.78   434.78     200.00
-    Mass Concentration bod                    217.39   217.39     200.00
-    Mass Concentration tss                    217.39   217.39     200.00
-    Mass Concentration ammonium_as_nitrogen   86.957   26.087     200.00
-    Mass Concentration nitrate                43.478   104.35     200.00
-====================================================================================
-"""
+        model.fs.unit.report()
 
-        assert output in stream.getvalue()
```
tarnold17, line 213: Why have you removed all of the lines in the `test_report` method? I thought the only one we were concerned with was the line reporting "Blower Size"? You have removed the only `assert` statement, so the test no longer actually tests anything, it just prints the report.

adam-a-a: @tarnold17 I think because we started leaning towards removing the string report tests (see @bknueven 's PR #528 )

watertap/unit_models/zero_order/mabr_zo.py
luohezhiming: Agree, added the variable to the report.

watertap/unit_models/zero_order/tests/test_mabr_zo.py
tarnold17: Ah I was unaware, thanks for explaining, Adam.

bknueven: It's better to test the values you think need to be tested directly.

adam-a-a (approved):
LGTM

tarnold17 (approved):

Issue #492, opened 2022-04-14
Boron Removal Model Development


aladshaw3:
To Do List:
-----------
- [x] Equilibrium reactor for boron speciation [**June/July**] (#531)
- [x] Zero order boron removal unit [**June**] (#531)
- [x] Boron Removal Documentation (#536)
- [ ] Demonstration for boron removal in flowsheet [**June**]
- [ ] Boron specific RO unit ?? [**If needed**]

aladshaw3:
Should be done as a unit model that contains the IDAES unit model for equilibrium reactor including H+, OH-, H2O, Boron speciation, and added NaOH (or other base)

aladshaw3:
May be good to validate against an OLI simulation

PR #531, opened 2022-05-11
Boron removal model

aladshaw3:
## Fixes/Addresses:

#492 

## Summary/Motivation:

Creates a simplified equilibrium model for boron removal. Works with or without protons and hydroxide as explicit species in the property package. Calculates pH change and boron/borate distribution based on addition of a caustic additive defined by the user. Works with the ion_prop_pack in watertap and technically will work with the generic property package in IDAES. 

- Adds boron removal and associated tests
- Modifies the bounds on the ion_prop_pack (REQUIRED to facilitate the pH dependence) 
- Updates ion_prop_pack tests 

### Legal Acknowledgement


watertap/unit_models/boron_removal.py
```diff
@@ -0,0 +1,1123 @@
...
+        # build always starts by calling super().build()
+        # This triggers a lot of boilerplate in the background for you
+        super().build()
+
+        # this creates blank scaling factors, which are populated later
+        self.scaling_factor = Suffix(direction=Suffix.EXPORT)
+
+        # Next, get the base units of measurement from the property definition
+        units_meta = self.config.property_package.get_metadata().get_derived_units
+
```
lbibl, line 229: Can this units_meta be removed since all units are/can be adopted from `pyunits` directly?

andrewlee94: This functionality is intended to allow for general models where the units of measurement are not know a priori - i.e. in many cases the units of measurement are determined by the property package, and thus the units definition in the unit model will need to change depending on what the property package is using.

watertap/unit_models/boron_removal.py
```diff
@@ -0,0 +1,1123 @@
...
+                )
+
+        # Constraints for mass transfer terms
+        @self.Constraint(
+            self.flowsheet().config.time,
+            doc="Electroneutrality condition",
+        )
+        def eq_electroneutrality(self, t):
+            Alk = 0
+            for j in self.ion_charge:
```
lbibl, line 590: Isn't `self.ion_charge` the charge values (e.g., +1 or -1) instead of the ion names?  but it must be right as the code passes properly.

watertap/unit_models/boron_removal.py
```diff
@@ -0,0 +1,1123 @@
...
+                    to_units=units_meta("amount") * units_meta("length") ** -3,
+                )
+                input_rate = self.control_volume.properties_in[t].flow_mol_phase_comp[
+                    p, self.boron_name_id
+                ]
+                exit_rate = (
+                    self.control_volume.properties_out[t].flow_vol_phase[p] * boron_out
+                )
+
+                loss_rate = input_rate - exit_rate
```
lbibl: Just a question-  This constraint seems to be only for a unit conversion essentially ?

watertap/unit_models/boron_removal.py
aladshaw3: ion_charge is a pyomo Param indexed by name, so the values of j will be the names. 

If it were just a list, then you would be correct.

watertap/unit_models/boron_removal.py
aladshaw3: No, this constraint is basing the loss rate off of the solution to the boron concentration inside the unit at the given pH and applies that loss rate such that we achieve that end result.

lbibl (dismissed):
I made a few minor comments. The overall content makes sense looks good to me.  A general comment: this seems to calculate Boron speciation upon a pH increase and I assume this unit is usually adapted to a RO to make the RO reaction of Boron more efficient; if this is correct, should we name it something else rather than "Boron removal"?  (Correct me if I am wrong. )    I approve in the scope of this pH increase treatment scheme.

watertap/unit_models/boron_removal.py
aladshaw3: It does include in it a unit conversion, but that is just to ensure that the units always work out regardless of how the property package was setup.

watertap/unit_models/boron_removal.py
```diff
@@ -0,0 +1,1123 @@
...
+                )
+                input_rate = self.control_volume.properties_in[t].flow_mol_phase_comp[
+                    p, self.hydroxide_name_id
+                ]
+                exit_rate = (
+                    self.control_volume.properties_out[t].flow_vol_phase[p] * h_out
+                )
+
+                loss_rate = input_rate - exit_rate
+                return self.control_volume.mass_transfer_term[t, p, j] == -loss_rate
```
adam-a-a: I skimmed this quickly so may be wrong, but can these be condensed? It looks like you might only need the conditional to decide which component molarity will be converted.

e.g.,: 
```
if j == self.proton_name_id:
    molarity = self.mol_H[t]
elif j == self.hydroxide_name_id:
   molarity = self.mol_OH[t]
.....
```

adam-a-a: Wouldn't change functionality but would reduce number of lines

watertap/unit_models/boron_removal.py
```diff
@@ -0,0 +1,1123 @@
...
+        )
+        self.mol_Borate = Var(
+            self.flowsheet().config.time,
+            initialize=1e-2,
+            bounds=(0, None),
+            domain=NonNegativeReals,
+            units=pyunits.mol / pyunits.m**3,
+            doc="Resulting molarity of Borate",
+        )
+
```
adam-a-a: So, aren't these the molarities at the outlet? Is there a reason why these couldn't be represented by `self.properties_out[t].conc_mol_phase_comp`?

watertap/unit_models/boron_removal.py
```diff
@@ -0,0 +1,1123 @@
...
+                        self.control_volume.properties_in[t].flow_mass_phase_comp[
+                            ind
+                        ] = value(
+                            self.control_volume.properties_in[t0].flow_mol_phase_comp[
+                                ind
+                            ]
+                            * self.control_volume.properties_in[t0].mw_comp[ind[1]]
+                        )
+
```
adam-a-a: shouldn't this already be handled by the prop model?

adam-a-a: Oh, I guess you are updating these values to match the instance relevant to the unit model.

watertap/unit_models/boron_removal.py
aladshaw3: Yes, I think that may work for most of them.

watertap/unit_models/boron_removal.py
aladshaw3: Depending on which property package is used, these may be initialized differently (or at different times). Thus, I have this function to reinitialize them to the most relevant values for the model. It helps **significantly** with scaling. I highly recommend doing this reinitialization in any unit model for setting initial property values relevant to the unit model's problem.

watertap/unit_models/boron_removal.py
aladshaw3: My reasoning was initially for testing purposes to make sure that the internal speciation problem was solved correctly without explicitly needing to couple with the properties. After I was satisfied with the results, I made constraints on mass transfer terms to essentially enforce this equality. 

Doing it in the way I am doing it also makes the unit model a little less dependent on the scaling of the property package, as I can scale the internal representations separately from the scaling employed by the package. I also get better control of the units on the variables, which can be helpful if this is used with multiple property packages.

adam-a-a: I'm all about approaches that make things more helpful although this might deviate from our conventions. I don't feel strongly about changing what you have necessarily, but we probably want @andrewlee94 and @TimBartholomew 's opinions on this too.

watertap/unit_models/boron_removal.py
adam-a-a: I think I like it. =)

watertap/unit_models/boron_removal.py
aladshaw3: Will update names to standard naming conventions to make this more clear.

aladshaw3:
@adam-a-a @TimBartholomew @lbibl 

Updated the unit model per suggestions during our meeting. Also changed one of the unit tests to solve for the dosage of caustic additive needed to achieve a specific outlet rate of boron.

watertap/unit_models/boron_removal.py
```diff
@@ -0,0 +1,1141 @@
...
+                ]
+                exit_rate = (
+                    self.control_volume.properties_out[t].flow_vol_phase[p] * h_out
+                )
+
+                loss_rate = input_rate - exit_rate
+                return self.control_volume.mass_transfer_term[t, p, j] == -loss_rate
+            elif j == self.cation_name_id:
+                c_out = pyunits.convert(
+                    self.caustic_cation_charge * self.caustic_dose[t] / self.caustic_mw,
```
aladshaw3:
```suggestion
                    self.caustic_dose[t] / self.caustic_mw,

```

aladshaw3:
**NOTE**: This PR will be paused pending some revisions and updates per conversations with the Chem Team. 

- [x] Update model to enforce user provide a caustic cation species in the state variables (so we can properly track changes in salinity from inlet to outlet)
- [x] Update chemical dose to a total mass of additive (rather than concentration)
- [x] Update unit to have a volume variable set by the user (used to determine retention time)

@lbibl Let me know if I am missing anything here we discussed.

lbibl (dismissed):
Looking good to me.

watertap/unit_models/boron_removal.py
```diff
@@ -0,0 +1,1202 @@
...
+                )
+                input_rate = self.control_volume.properties_in[t].flow_mol_phase_comp[
+                    p, self.hydroxide_name_id
+                ]
+                exit_rate = (
+                    self.control_volume.properties_out[t].flow_vol_phase[p] * h_out
+                )
+
+                loss_rate = input_rate - exit_rate
+                return self.control_volume.mass_transfer_term[t, p, j] == -loss_rate
```
bknueven: I'm not a huge fan of the repeated code in this constraint formulation. It could be eliminated by using a simple dictionary mapping:
```python
{ self.hydroxide_name_id : self.conc_molOH,
...
}
```

watertap/unit_models/boron_removal.py
```diff
@@ -0,0 +1,1202 @@
...
+            iscale.set_scaling_factor(
+                blk.conc_mol_OH[t], 100 / blk.conc_mol_OH[t].value
+            )
+            iscale.set_scaling_factor(blk.conc_mol_H[t], 100 / blk.conc_mol_H[t].value)
+            iscale.set_scaling_factor(
+                blk.conc_mol_Boron[t], 100 / blk.conc_mol_Boron[t].value
+            )
+            iscale.set_scaling_factor(
+                blk.conc_mol_Borate[t], 100 / blk.conc_mol_Borate[t].value
+            )
```
bknueven: It seems a bit dangerous to rescale these purely based on value without an upper limit.

watertap/unit_models/boron_removal.py
```diff
@@ -0,0 +1,1202 @@
...
+                            ]
+                            * self.control_volume.properties_in[t0].mw_comp[ind[1]]
+                            / sum(
+                                self.control_volume.properties_in[
+                                    t0
+                                ].flow_mol_phase_comp[ind[0], j]
+                                * self.control_volume.properties_in[t0].mw_comp[j]
+                                for j in self.config.property_package.component_list
+                            )
+                        )
```
bknueven: It might make sense to move this to a helper function so reduce the indentation two levels, making the code easier to read and reason about (e.g., am I still in this `if` block?)

watertap/unit_models/boron_removal.py
```diff
@@ -0,0 +1,1202 @@
...
+    def outlet_pH(self, time=0):
+        return -log10(value(self.conc_mol_H[time]) / 1000)
+
+    def outlet_pOH(self, time=0):
+        return -log10(value(self.conc_mol_OH[time]) / 1000)
+
+    def propogate_initial_state(self):
+        units_meta = self.config.property_package.get_metadata().get_derived_units
+        i = 0
+        for t in self.control_volume.properties_in:
```
bknueven: Since it seems like we need `i`, it would be better to use `enumerate` here, e.g.:
```suggestion
        for i, t in enumerate(self.control_volume.properties_in):
```

That said, since it seems like you just use this to grab the first time index, you might instead just put this before the loop:
`t0 = self.flowsheet().time.first()`

bknueven (commented):
I can't really speak to the model, but I have a few comments on the code and a comment on the auto-scaling the initialization is doing.

watertap/unit_models/boron_removal.py
aladshaw3: It should not really be much of an issue for these variables. Most of these 'values' will not even approach 1. I guess I could add an upper limit to the variables to be safe.

watertap/unit_models/boron_removal.py
aladshaw3: Sure, I can do that.

watertap/unit_models/boron_removal.py
aladshaw3: I will see about condensing this down.

watertap/unit_models/boron_removal.py
bknueven: Sorry, I mean an upper limit on `100 / var.value`, i.e., the applied scaling factor.

bknueven (approved):

lbibl (approved):

PR #529, opened 2022-05-11
Basic .rst file example for WT3 documentation

adam-a-a:
## Summary/Motivation:
Basic .rst file template for zero order units that only use helper functions.

## Changes proposed in this PR:
- add draft template for .rst file
- add index .rst

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #532, opened 2022-05-11
Automatically Generate Zero Order Unit Documentation

adam-a-a:
## Summary/Motivation:
We need documentation for zero-order unit models. I wrote a script to automatically generate .rst files for all units. The intention is that we can modify .rst files for customized units on the go using the template generated here as a starting point (and possibly remove or relocate a couple, e.g., feed_zo). Any additional changes (e.g., like revising titles to be all CAPS --> Candop to CANDOP) can be made in subsequent PRs as we build out customized docs for select units and do general cleanup.

**EDIT**: See the corresponding template used for all unit model .rst files in draft PR #529 (with slight mods made in this PR).

- automatically generate .rst files for all zero order units
- automatically generate index.rst
- add automation script and corresponding excel used for dataframe in the case that anyone wants to modify and rerun the automation script.

### Legal Acknowledgement


docs/technical_reference/unit_models/zero_order_unit_models/automate_rst_file_wt3.py
```diff
@@ -0,0 +1,72 @@
+import pandas as pd
```
bknueven, line 16: Should we add the header to this file?

bknueven (commented):
Screen Shot 2022-05-12 at 7 18 55 AM

Something not quite right is happening with automodule and the "Class Documentation".

Maybe it would be better if the unit classification file were a csv?

docs/technical_reference/unit_models/zero_order_unit_models/automate_rst_file_wt3.py
adam-a-a: I suppose so.

adam-a-a:
> Screen Shot 2022-05-12 at 7 18 55 AM
> 
> Something not quite right is happening with automodule and the "Class Documentation".
> 
> Maybe it would be better if the unit classification file were a csv?

Ah- thanks- I just needed to add a space before `automodule` (I think). Let's see how it turns out now.

bknueven (dismissed):

docs/technical_reference/unit_models/zero_order_unit_models/CANDOP_zo.rst
```diff
@@ -0,0 +1,29 @@
+Candop  (ZO)
```
tarnold17, line 1: This would be better formatted as "CANDO+P". Several other units whose names are acronyms are similarly affected - e.g., the page for DMBR shows "Dmbr", UV shows as "Uv", etc.

bknueven (approved):

luohezhiming (approved):

docs/technical_reference/unit_models/zero_order_unit_models/CANDOP_zo.rst
bknueven: We will modify individual unit model documentation in subsequent PRs.

PR #540, opened 2022-05-19
Parameter Sweep Tool: resetting variable values before reinitialization

bknueven:
## Summary/Motivation:
In the parameter sweep tool, sometimes reinitialization fails because some variables (e.g., those calculated or in properties) have wondered too far off their initial values. This PR would address that by resetting all non-fixed variables every time initialization is called to the value they had at the beginning of the parameter sweep.

- Before reinitializing a flowsheet, either after a failed solve or at the beginning of each parameter value, reset all non-fixed variable values to the value the variable had when the parameter sweep was started.

### Legal Acknowledgement


k1nshuk (approved):

avdudchenko (approved):
This looks good and have tested in a different version of parameter sweep tool. Thanks for adding it in.

PR #526, opened 2022-05-09
Configure Sphinx to generate apidoc files automatically

lbianchi-lbl:
## Summary/Motivation:

- It turns out that there is a way to register arbitrary functions in `conf.py` so that they run as part of `sphinx-build`
- This PR uses this functionality to automatically run `sphinx-apidoc`, which in turns eliminates the need to manually create and commit the files generated by `sphinx-apidoc`

## Changes proposed in this PR:
- Add functionality in `docs/conf.py` to generate apidoc files automatically
- Remove apidoc files that were checked into the Git repository
- Remove extra scripts and documentation used to manually generate apidocs file

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #542, opened 2022-05-19
De-noindex-ation

### LINKED ISSUE #539: Turn off the "noindex" in the apidocs?
In order to find documentation in the apidocs, i.e. when it is not included directly in a webpage, the "noindex" option cannot be present when they are generated.

This has two related side effects:
* Autodoc directives in an RST page will need to include ":noindex:" in their options to avoid warnings
* When you link to a class, module, function, etc. the docs location will be in the apidoc-generated tree (not the documentation page)

The second one is a bit of a kicker, and I think points to a conflict in philosophy that we should resolve.

lbianchi-lbl:
## Summary/Motivation:

- Since #526, the `:noindex:` flag is automatically added to the documentation generated by `sphinx-apidoc`
- Therefore, manually adding `:noindex:` to human-generated documents is not only no longer necessary, but also not desirable, since it would cause in the target module not being indexed at all

- Remove all instances of `:noindex:` flag in human-generated documents
- Remove references to the use of `:noindex:` from developer docs since it no longer applies

### Legal Acknowledgement

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -6,7 +6,7 @@ coverage:
     patch:
       default:
         target: auto
-        threshold: 0%
+        threshold: 10%
     project:
       default:
         target: auto
--- a/docs/getting_started/install.rst
+++ b/docs/getting_started/install.rst
@@ -365,14 +365,7 @@ documentation for your module as generated from the source code (and docstrings)
 
     .. automodule:: watertap.<package_name>.<module_name>
         :members:
-        :noindex:
 
 The meaning of the options is the following:
 
-* ``:noindex:`` (**required**): prevents the creation of an index entry.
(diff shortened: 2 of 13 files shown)


lbianchi-lbl:
There are some `:noindex:`-related errors in the RTD build, so this might actually require a bit more thought.

I think there could be cases where the same module is "targeted" by `.. automodule::` in `N` separate places within the human-generated docs, so we might still have to add `:noindex:` to `N - 1` of those.

lbianchi-lbl:
Closing this since we've decided to go in a different direction (see [this comment](https://github.com/watertap-org/watertap/issues/539#issuecomment-1145333691) for more info).

PR #550, opened 2022-05-31
GAC Model

hunterbarber:
## Fixes/Resolves:
Initial build of GAC model for review

## Summary/Motivation:
Development of GAC model with accompanying testing 
sim_ files used for debugging in flowsheet model structure
test_ files used for GAC model testing

n/a

### Legal Acknowledgement


bknueven (commented):
Why are we adding a new property package and not just making the changes directly to the `ion_DSPMDE_prop_pack`?

adam-a-a:
> Why are we adding a new property package and not just making the changes directly to the `ion_DSPMDE_prop_pack`?




> Why are we adding a new property package and not just making the changes directly to the `ion_DSPMDE_prop_pack`?

@hunterbarber I think you just updated the lower bounds on a few variables, correct? If so, perhaps you can make those updates directly in the original `ion_DSPMDE_prop_pack`, run your unit again using that prop pack, and scrap the duplicated file that has the adjustments.

adam-a-a:
In fact for all the variables with small lower bounds, updating them to 0 should still work. At least it does for me when tinkering the NF DSPM model which the `ion_DSPMDE_prop_pack` was designed for originally.

bknueven:
> In fact for all the variables with small lower bounds, updating them to 0 should still work. At least it does for me when tinkering the NF DSPM model which the `ion_DSPMDE_prop_pack` was designed for originally.

Here is the diff:
```
743c743
<             bounds=(1e-12, None),
---
>             bounds=(0, None),
772c772
<             bounds=(1e-12, 1.001),
---
>             bounds=(0, 1.001),
844c844
<             bounds=(1e-12, None),
---
>             bounds=(0, None),
872c872
<             bounds=(1e-12, None),
---
>             bounds=(0, None),
892c892
<             bounds=(1e-12, 2e3),
---
>             bounds=(0, 2e3),
912c912
<             bounds=(1e-12, None),
---
>             bounds=(0, None),
932c932
<             bounds=(1e-12, 1.001),
---
>             bounds=(0, 1.001),
950c950
<             bounds=(1e-12, 10),
---
>             bounds=(0, 10),
1174c1174
<             tol = 1e-12
---
>             tol = 1e-8
```

hunterbarber:
It is correct that the property package was only minimally changed by editing the lower bounds. I was originally working out of an edited version of the original `ion_DSPMDE_prop_pack` but wasn't sure how it would be read when creating the PR as I hadn't been committing this file, so I made the 'ion_DSPMDE_prop_pack_GACadjusted' just to differentiate the edited and original files. Possible ignorance on my part on what exactly would be merged and would have been able to be reviewed when creating the PR, but I can make these adjustments directly in the `ion_DSPMDE_prop_pack` in the future as Adam has commented.

hunterbarber:
Also, the reason I had not updated the lower bounds to 0 (instead of 1e-12) when testing was that I had uncertainty about the zeros handling from a solver perspective. It is good to know that these can still be solved as I move forward with other examples with trace concentrations.

watertap/unit_models/tests/test_gac.py
```diff
@@ -0,0 +1,143 @@
...
+    stream_table_dataframe_to_string,
+)
+
+from watertap.property_models.ion_DSPMDE_prop_pack_GACadjusted import (
+    DSPMDEParameterBlock,
+)
+from watertap.unit_models.gac import GAC
+
+
+def main():
```
adam-a-a: @hunterbarber Please have a look at other example test files and try following one of those examples to make your test file. This is useful but we'd consider this to be more of simulation file.

adam-a-a: @hunterbarber Please have a look at other example test files and try following one of those examples to make your test file. This is useful but we'd consider this to be more of simulation file.

hunterbarber: Okay, I understood that this wasn't the convention but did need a location to house this early testing prior to the more detailed testing.

watertap/unit_models/gac.py
```diff
@@ -0,0 +1,687 @@
...
+        self.freund_k = Var(
+            initialize=10,
+            bounds=(0, 1000),
+            domain=NonNegativeReals,
+            # TODO: Correct variable units
+            units=pyunits.dimensionless,  # ((units_meta("length") ** 3) * units_meta("mass") ** -1) ** 0.8316,
+            doc="Freundlich k parameter",
+        )
+
+        self.eps_bed = Var(
```
adam-a-a: Consider renaming `bed_void_fraction` or `void_fraction_bed`

watertap/unit_models/gac.py
```diff
@@ -0,0 +1,687 @@
...
+
+        self.thru = Var(
+            initialize=10,
+            bounds=(0, None),
+            domain=NonNegativeReals,
+            units=pyunits.dimensionless,
+            doc="Mass throughput",
+        )
+
+        self.tau = Var(
```
adam-a-a: Consider renaming something like `contact_time`

watertap/unit_models/tests/test_gac.py
lbianchi-lbl: > Okay, I understood that this wasn't the convention but did need a location to house this early testing prior to the more detailed testing.

@hunterbarber as a quick check, if you rename this function from `main()` to something starring with `test_` e.g. `test_gac_main()`, it will be picked up by pytest and counted in the coverage report.

adam-a-a:
@hunterbarber I don't think this was the case before, but now the MAC check is failing. Not sure how the model/test refinements could've affected this yet (since I think the MAC check was passing before), but this is a common issue we see that we usually resolve by adding @pytest.mark.requires_idaes_solver above all tests that depend on a solve. 

See the example highlighted here: https://github.com/adam-a-a/watertap/blame/ccff25f273956195c8007b9cf01e4aa72fedddd7/watertap/unit_models/zero_order/tests/test_gas_sparged_membrane.py#L204

watertap/unit_models/tests/test_gac.py
```diff
@@ -0,0 +1,360 @@
...
+        assert results.solver.status == SolverStatus.ok
+
+    @pytest.mark.component
+    def test_solution_robust(self, gac_frame_robust):
+        m = gac_frame_robust
+
+        # low tolerance due to inconsistent reporting in reference
+        assert pytest.approx(1.14, rel=1e-1) == value(m.fs.unit.mass_throughput)
+        assert pytest.approx(158 * 3600 * 24, rel=1e-1) == value(m.fs.unit.elap_time)
+        assert pytest.approx(10.68, rel=1e-1) == value(m.fs.unit.bed_area)
```
adam-a-a: It might be better to use tolerance of 1e-5 and fill in the values you expect to get rather than using the exact reference values. A note about which reference you are referring to might also be useful in the future. Is it the Crittenden reference mentioned earlier?

adam-a-a (changes requested):
Looks good, but the Mac check is failing now and needs to be resolved. Also, a suggestion about the loose tolerances.

watertap/unit_models/tests/test_gac.py
hunterbarber: Yes, it is the Crittenden, 2012 reference. It uses calculated values reported in days (e.g., 158 days of operation time) which when converted to seconds for the base SI of the model output introduces a lot of rounding error. I can alter these to the expected value from previous calculations that are the same magnitude but off by rounding to the reference calculations.

adam-a-a (approved):
LGTM - any further refinement can be resolved in subsequent PRs when we try extending the model or using it in a flowsheet.

TimBartholomew (approved):
LGTM

PR #554, opened 2022-06-01
UV Model

luohezhiming:
## Summary/Motivation:
Develop first version of UV model for reviewing

- Develop UV model
- Develop NDMA package
- Add corresponding tests

### Legal Acknowledgement


adam-a-a (approved):
Looks good to me. We can make modifications in subsequent PRs as needed (e.g., determining energy demand with multiple components and repercussions on removal of components).

Also, I started to wonder whether the UV dose should be linked to the EE/O, rather than fixing it. Again, this is something we can settle in a subsequent PR but also something to think about.

watertap/unit_models/ultraviolet.py
```diff
@@ -223,13 +224,24 @@ def build(self):
             doc="Inactivation rate coeff.",
         )
 
+        self.rate_constant = Var(
```
adam-a-a, line 224: Just for my understanding- what is the use for the rate_constant in the model? I see it relates to UV intensity and inactivation rate but unsure of its significance beyond that.

luohezhiming: The `rate_constant` can be an alternative input for users instead of `inactivation_rate`. It is easier for users to get the `rate_constant` from the dose response curve with respect to exposure time. I see different formulations of the dose response model, someone provides the `rate_constant` and `exposure_time`, and someone provides `inactivation_rate` and `uv_dose`, so it depends. The `rate_constant` can also be obtained by the kinetic reactions (though haven't considered yet). And, another point is that there is an alternative EEO relationship with respect to the `rate_constant`, I didn't include it since it is just derived from substituting the `rate_constant` into the dose response model and require some unit conversions.

adam-a-a: Thanks for the nice explanation!

watertap/property_models/NDMA_prop_pack.py
```diff
@@ -0,0 +1,781 @@
...
+        # Initialize at current state values (not user provided)
+        self.initialize(solver=solver, optarg=optarg, outlvl=outlvl)
+
+        # Set solver and options
+        opt = get_solver(solver, optarg)
+
+        # Fix variables and check degrees of freedom
+        flags = (
+            {}
+        )  # dictionary noting which variables were fixed and their previous state
```
bknueven:
```suggestion
         # dictionary noting which variables were fixed and their previous state
        flags = {}
```
Putting comments at the end of the line is not very black friendly

watertap/property_models/NDMA_prop_pack.py
```diff
@@ -0,0 +1,781 @@
...
+            self.params.phase_list,
+            initialize=997,
+            bounds=(5e2, 2e3),
+            units=pyunits.kg * pyunits.m**-3,
+            doc="Mass density",
+        )
+
+        def rule_dens_mass_phase(
+            b,
+        ):  # constant density, NDMA concentration is always minimal
```
bknueven:
```suggestion
        def rule_dens_mass_phase(b):
            # constant density, NDMA concentration is always minimal
```
Same comment here

watertap/unit_models/ultraviolet.py
```diff
@@ -0,0 +1,613 @@
...
+
+    def build(self):
+        # Call UnitModel.build to setup dynamics
+        super().build()
+
+        self.scaling_factor = Suffix(direction=Suffix.EXPORT)
+
+        if len(self.config.property_package.phase_list) > 1 or "Liq" not in [
+            p for p in self.config.property_package.phase_list
+        ]:
```
bknueven:
```suggestion
        if list(self.config.property_package.phase_list) != ["Liq"]:
```

bknueven (approved):
I didn't read the model for understanding, but the code looks solid. Only have a few minor optional nits regarding how black changed a few things.

watertap/property_models/tests/test_NDMA_prop_pack.py
```diff
@@ -0,0 +1,120 @@
...
+# Technology Laboratory (subject to receipt of any required approvals from
+# the U.S. Dept. of Energy). All rights reserved.
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and license
+# information, respectively. These files are also available online at the URL
+# "https://github.com/watertap-org/watertap/"
+#
+###############################################################################
+
+import pytest
```
hunterbarber, line 14:
```suggestion
import pytester
```

PR #558, opened 2022-06-02
Refining DSPMDE 

adam-a-a:
## Fixes/Resolves:
#464 
## Summary/Motivation:

Refining NF DSPM-DE model to obtain valid rejection rates for multiple ions. Degenerate constraints needed to be removed and replaced to achieve a square problem and have a well formulated model. 

Moreover, the main objective of this model is to predict ion rejection rates, and the previous implementation was predicting erroneously low rejection rates for all ions. The error originated from an error in the literature, where the reported equation for partitioning due to dielectric exclusion was provided as `-gibbs_energy/boltzmann_constant/temperature` instead of `exp(-gibbs_energy/boltzmann_constant/temperature)`, which led to significant overestimation of ion concentrations in the permeate and thus erroneously low rejection rates for all ions. The fix is implemented here.

- improve initialization/scaling so that model solves for 2 - 5 ions, with or without concentration polarization effects
- correct erroneous constraint from literature
- remove electroneutrality constraints for feed and replace with equal volume flowrate constraints where needed to resolve stateblocks that do not play a role in the overall mass balance
- modify tests

### Legal Acknowledgement


watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -1083,7 +1083,7 @@ def _pressure_osm_phase(self):
         self.pressure_osm_phase = Var(
             self.params.phase_list,
             initialize=1e6,
-            bounds=(5e2, 5e7),
+            bounds=(0, None),
```
arao53, line 1086: Does this variable need to have an explicitly defined domain (e.g. `NonNegativeReals`)?

adam-a-a: @arao53 I think that bounds of (0, None) is equivalent to `NonNegativeReals`

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -350,12 +390,10 @@ def build(self):
             io_list,
             phase_list,
             solvent_solute_set,
-            initialize=(
-                lambda b, t, x, p, j: 2.78e-2 if j in solvent_set else 4e-6
-            ),  # TODO: divided mass solvent by .018 and solute by .25
-            bounds=lambda b, t, x, p, j: (5e-5, 1.5)
-            if j in solvent_set
-            else (4e-12, 1e-2),  # TODO: keep checking these
+            initialize=(lambda b, t, x, p, j: 2.78e-2 if j in solvent_set else 4e-6),
```
aladshaw3, line 394: Is this based on anything physical or chemical? Would we be better off in just providing initialization of these variables in the `initialize_build` method based on other state vars or system conditions?

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -398,32 +436,12 @@ def build(self):
             self.flowsheet().config.time,
             io_list,
             ["pore_entrance", "pore_exit", "permeate"],
-            initialize=0.1,  # TODO:revisit
+            initialize=-1e-3,  # TODO:revisit
             domain=Reals,
             bounds=(-1.001, 1.001),
```
aladshaw3: Is the electric potential actually bounded? What the is basis for choosing to bound the potential in this way?

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -445,51 +463,39 @@ def build(self):
             self.flowsheet().config.time,
             self.config.property_package.phase_list,
             initialize=0.5,
-            bounds=(1e-3, 1 - 1e-6),
+            bounds=(0, 1 - 1e-6),
```
aladshaw3: These bounds seem odd. Is there some significance to the upper bound value?

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -1386,23 +1515,23 @@ def initialize_build(
...
-            self._automate_rescale_variables()
+        # # Rescale any badly scaled vars
+        # if automate_rescale:
+        #     badly_scaled_vars = list(iscale.badly_scaled_var_generator(self))
+        #     if len(badly_scaled_vars) > 0:
+        #         init_log.warn(
+        #             f"After solve: {len(badly_scaled_vars)} poorly scaled "
+        #             f"variable(s) will be rescaled so that each scaled variable value = 1"
+        #         )
+        #     self._automate_rescale_variables()
```
aladshaw3, line 1559: You re-scaled prior to initialization, but not after. Should this be active?

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -1807,10 +1947,10 @@ def calculate_scaling_factors(self):
...
             if iscale.get_scaling_factor(v) is None:
                 iscale.set_scaling_factor(v, 1e4)
-
-        for (t, x), v in self.electric_potential_grad_feed_interface.items():
-            if iscale.get_scaling_factor(v) is None:
-                iscale.set_scaling_factor(v, 1e-2)
+        if hasattr(self, "electric_potential_grad_feed_interface"):
+            for v in self.electric_potential_grad_feed_interface.values():
+                if iscale.get_scaling_factor(v) is None:
+                    iscale.set_scaling_factor(v, 1e-2)
```
aladshaw3: Is this going to be sufficient for the scaling of this variable? Do we know anything about the range of values for the gradients?

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -1859,29 +2009,26 @@ def calculate_scaling_factors(self):
                 iscale.set_scaling_factor(
                     self.feed_side.mass_transfer_term[0, "Liq", j], 1
                 )
```
aladshaw3, line 2036: Since there are constraints to calculate the mass transfer, would it be better to approximate a scaling factor based on some formulaic representation of the expected values for mass transfer terms? I imagine that mass transfer rates could be very different for different ions

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -1902,24 +2049,96 @@ def calculate_scaling_factors(self):
...
+        for con in self.eq_equal_flow_vol_permeate.values():
+            iscale.constraint_scaling_transform(con, 1)
+        for con in self.eq_equal_flow_vol_pore_exit_perm.values():
+            iscale.constraint_scaling_transform(con, 1)
+        for con in self.feed_side.eq_equal_flow_vol_feed_interface.values():
+            iscale.constraint_scaling_transform(con, 1)
+        for con in self.feed_side.eq_feed_interface_isothermal.values():
+            iscale.constraint_scaling_transform(con, 1)
+        for con in self.feed_side.eq_feed_isothermal.values():
+            iscale.constraint_scaling_transform(con, 1)
```
aladshaw3, line 2169: The scaling needs work. The factors we put into the scaling of these variables and constraints should be based on scaling factors for the state variables and other unit variables that go into the formula for the constraint.

For instance, if you have a constraint like:
`var = a * b / c`
Then the scaling factor (for `var` and the constraint) should probably take the form of:
`var_sf = a_sf * b_sf / c_sf`
Where each *`_sf` represents the scaling factor taken from *.

watertap/unit_models/tests/test_nanofiltration_DSPMDE_0D.py
```diff
@@ -337,21 +357,18 @@ def test_conservation(self, NF_frame):
...
-            "H2O": 0.41437,
+            "Na_+": 0.00029831,
+            "Cl_-": 0.00031137,
+            "Ca_2+": 1.9289e-06,
+            "SO4_2-": 1.9731e-06,
+            "Mg_2+": 6.5754e-06,
+            "H2O": 0.032671,
         }
         for j, val in mole_flux_dict.items():
             assert pytest.approx(val, rel=5e-2) == value(
```
aladshaw3: This tolerance needs to be tighter

watertap/unit_models/tests/test_nanofiltration_DSPMDE_0D.py
```diff
@@ -69,21 +70,23 @@
...
     m = ConcreteModel()
     m.fs = FlowsheetBlock(default={"dynamic": False})
-    # NOTE: Not giving actual charges in order to test construction of
-    #   both the ion_set and solute_set
     m.fs.properties = DSPMDEParameterBlock(
-        default={"solute_list": ["Ca_2+", "SO4_2-", "Na_+", "Cl_-", "Mg_2+"]}
+        default={
+            "solute_list": ["Ca_2+", "SO4_2-", "Na_+", "Cl_-", "Mg_2+"],
+            "charge": {"Ca_2+": 2, "SO4_2-": -2, "Na_+": 1, "Cl_-": -1, "Mg_2+": 2},
+        }
```
aladshaw3, line 78: So right now you are only set up for this 1 specific test case, but given the number of issues this model has had, I think we need to add in some additional tests.

(1) Unit tests with and without Concentration Polarization
(2) Unit tests with the different Mass Transfer Coefficient options
(3) Unit tests with different sets of ions (this set + an NaCl set)
(4) Unit tests with different inlet conditions (i.e., stress testing)
(5) Unit tests with fixed outlets instead of inlets (i.e., an inverse solve to show model solution is truly unique)

aladshaw3 (changes requested):
Here is my review on a first pass. 

The main things that need updating are (i) better scaling choices/values and (ii) more unit tests. 

We may not be able to get around to making all the tests I requested. I think the most important additional test is the 5th test (the inverse solve) to verify the solution obtained is unique. The next most important is the stress testing.

watertap/unit_models/nanofiltration_DSPMDE_0D.py
adam-a-a: This is from way back when I was having issues and had little idea what was causing the trouble. I can remove the bounds on this.

watertap/unit_models/nanofiltration_DSPMDE_0D.py
adam-a-a: The original thought on this came from the RO model, where we were cautious about setting recovery to 1 and instead set to a value close to 1. I think this is another relic that can go away

watertap/unit_models/nanofiltration_DSPMDE_0D.py
adam-a-a: I might take the automate_rescale out all together since it is barely being used right now and was able to stabilize the model without it (for the subset of cases I tried).

watertap/unit_models/nanofiltration_DSPMDE_0D.py
adam-a-a: I don't know very much about the range of values, but from what I have observed, I think they should be close to 1, so I can probably just set to a scaling factor of 1.

watertap/unit_models/nanofiltration_DSPMDE_0D.py
adam-a-a: Yes, this is included in some of the improvements I'd like to make. The reason I left it this way was because when I tried scaling the preferred way that you're suggesting, the solve was failing miserably. However, that was back when things were not well formulated and the model still had the erroneous constraint. I was considering updating all of this in a subsequent PR, but I can take a shot at it now too.

watertap/unit_models/tests/test_nanofiltration_DSPMDE_0D.py
adam-a-a: I totally agree. I'll check off what I can, and maybe we can revisit to cast judgement on what should be in this PR and what can be swept to a subsequent PR.

- [x] Unit tests with and without Concentration Polarization
- [x] Unit tests with the different Mass Transfer Coefficient options
- [x] Unit tests with different sets of ions (this set + an NaCl set)
- [x] Unit tests with different inlet conditions (i.e., stress testing)
- [x] Unit tests with fixed outlets instead of inlets (i.e., an inverse solve to show model solution is truly unique)

aladshaw3 (approved):
Looks like this is ready to go to me. The additional unit tests have boosted my confidence that the model is ready. May still need to tune the scaling, but we can do that in another PR if needed.

tarnold17 (approved):

PR #559, opened 2022-06-03
IX single component 0D model

kurbansitterley:
## Summary/Motivation:

Get an initial draft if IX model for review

- initial review for IX model
- need test files for model and prop pack 

### Legal Acknowledgement


watertap/property_models/tests/test_IX_prop_pack.py
```diff
@@ -0,0 +1,373 @@
...
+
+# # Imports from idaes generic models
+# from idaes.generic_models.properties.core.pure.ConstantProperties import Constant
+# from idaes.generic_models.properties.core.state_definitions import FpcTP
+# from idaes.generic_models.properties.core.eos.ideal import Ideal
+
+# # Import the idaes objects for Generic Properties
+# from idaes.generic_models.properties.core.generic.generic_property import (
+#     GenericParameterBlock,
+# )
```
adam-a-a:
```suggestion

```

watertap/property_models/IX_prop_pack.py
```diff
@@ -0,0 +1,1110 @@
...
+    def define_metadata(cls, obj):
+        """Define properties supported and units."""
+        obj.add_properties(
+            {
+                "flow_mol_phase_comp": {"method": None},
+                "temperature": {"method": None},
+                "pressure": {"method": None},
+                "flow_mass_phase_comp": {"method": "_flow_mass_phase_comp"},
+                "diffus_phase_comp": {"method": "_diffus_phase_comp"},
+                # "flow_mol_phase_comp": {"method": "_flow_mol_phase_comp"},
```
adam-a-a:
```suggestion
```

watertap/property_models/IX_prop_pack.py
```diff
@@ -0,0 +1,1110 @@
...
+    # reference_cation and reference_anion for use in future prop pack development
+    CONFIG.declare(
+        "reference_cation",
+        ConfigValue(default="H_+", domain=str, description="Reference Cation"),
+    )
+
+    CONFIG.declare(
+        "reference_anion",
+        ConfigValue(default="OH_-", domain=str, description="Reference Anion"),
+    )
```
adam-a-a: Aside from these two options and the exclusion of others (e.g., stokes_radius), what differentiates this from the ion_prop_model? I won't hold this up, but we might want to consider some consolidation if possible in a subsequent PR.

watertap/property_models/IX_prop_pack.py
```diff
@@ -0,0 +1,1110 @@
...
+
+        def rule_flow_equiv_phase_comp(b, j):
+            if j == "H2O":
+                return (
+                    b.flow_equiv_phase_comp["Liq", j] == b.flow_mol_phase_comp["Liq", j]
+                )
+            return b.flow_equiv_phase_comp["Liq", j] == b.flow_mol_phase_comp[
+                "Liq", j
+            ] / abs(b.params.charge_comp[j])
+
```
lbibl: What is the "equivalent"?  If it is equivalent charge, please check if it should be `eqv == ... * abs(... charge_comp[j]` . The same applies to the "equivalent concentration" var.

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -0,0 +1,1352 @@
...
+    CONFIG.declare(
+        "target_ion",
+        ConfigValue(default="Ca_2+", domain=str, description="Target Ion"),
+    )
+
+    CONFIG.declare(
+        "regenerant",
+        ConfigValue(default="HCl", domain=str, description="Regenerant"),
+    )
+
```
lbibl, line 142: I understand "configuration" as a small number of options to be chosen by the user (in UI will be a checkbox or a similar thing).  So the domain should not be unlimited like `str`.   If the goal is to limit the regeneration agent in a few chemicals, you may use `In()` or `Enum` .  If it can really be any chemicals (maybe not...), it might just be good to set it as a var.

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -0,0 +1,1352 @@
...
+    **Valid values:** {
+    see property package for documentation.}""",
+        ),
+    )
+
+    CONFIG.declare(
+        "target_ion",
+        ConfigValue(default="Ca_2+", domain=str, description="Target Ion"),
+    )
+
```
lbibl, line 137: For the same reason in commented in 127,  I feel this should not be a configuration.

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -0,0 +1,1352 @@
...
+        target_ion = self.config.target_ion
+
+        # this creates blank scaling factors, which are populated later
+        self.scaling_factor = Suffix(direction=Suffix.EXPORT)
+
+        self.gravity = Param(
+            initialize=9.81, units=pyunits.m / pyunits.s**2, doc="Gravity [m/s2]"
+        )
+
+        self.pi = Param(initialize=3.14159, units=pyunits.dimensionless, doc="Pi")
```
lbibl: You can probably just use `pi` to call the constant.

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -0,0 +1,1352 @@
...
+
+        self.diff_ion_resin = Param(
+            initialize=1e-13,
+            units=pyunits.m**2 / pyunits.s,
+            doc="Diffusivity of ion resin [m2/s]",  # Perry's
+        )
```
lbibl, line 176: Consider if it makes more sense to set this as a `Var`

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -0,0 +1,1352 @@
...
+            units=(pyunits.psi * pyunits.hr) / pyunits.m**2,
+            doc="Pressure drop eq B",
+        )
+
+        self.p_drop_C = Param(
+            initialize=8.28e-4,
+            units=(pyunits.psi * pyunits.hr**2) / pyunits.m**3,
+            doc="Pressure drop eq C",
+        )
+
```
lbibl, line 287: The same comment at 162 applies to all `Param` above.  If any of them will be later changed or optimized, it should be easier to set them as `Var`.

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -0,0 +1,1352 @@
...
+            units=pyunits.m,
+            doc="Column height [m]",
+        )
+
+        self.col_vol = Var(
+            initialize=10,
+            units=pyunits.m**3,
+            doc="Column volume of one unit [m3]",
+        )
+
```
lbibl, line 406: Some of the  vars carry a relatively small bound , which would be nice for the solver.  But I wonder if there will be any practical situation where a given bound is too narrow.  So maybe consider enlarging some bounds without hurting the solver's performance.

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -0,0 +1,1352 @@
...
+            doc="Dimensionless time [-]",
+        )
+
+        self.lh = Var(
+            initialize=0,
+            bounds=(-20, 20),
+            units=pyunits.dimensionless,
+            doc="Left hand side of kinetic equation",
+        )
+
```
lbibl, line 481: The meaning of this var is hard to get.

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -0,0 +1,1352 @@
...
+        self.mass_out = Var(
+            ion_set,
+            initialize=1e3,
+            bounds=(0, None),
+            units=pyunits.mol,
+            doc="Effluent mass of ion",
+        )
+
```
lbibl, line 505: Can Vars from 438 to 461 be from the property pack ?

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -0,0 +1,1352 @@
...
+        self.dimensionless_time.fix()
+        self.lh.fix()
+        self.regen_dose.fix()
+        self.regen_ww.fix()
+        self.regen_sg.fix()
+        self.t_bw.fix()
+        self.bw_rate.fix()
+        self.rinse_bv.fix()
+        self.pump_efficiency.fix()
+
```
lbibl: If these are to be fixed in this file, should they just be hard coded as numbers?  Otherwise they can be fixed in the test file (by the user).  Seems odd to fix vars in this file.

lbibl (commented):
This review majorly focuses on the coding structure.  General comments: (1) it can probably just use the current DSPMDE ion property pack by putting some vars in the unit file; (2) consider using the control volume block, and (3) consider adding scaling calculation methods for those potentially carrying  significant variations.

watertap/property_models/IX_prop_pack.py
```diff
@@ -0,0 +1,1110 @@
...
+        self.component_list = Set()
+
+        # components
+        self.H2O = Solvent()
+
+        # blank sets
+        self.cation_set = Set()
+        self.anion_set = Set()
+        self.solute_set = Set()
+        self.ion_set = Set()
```
aladshaw3: To reiterate @lbibl comments, a lot of this already exists as part of the `ion_DSPMDE_prop_pack`. To avoid significant code duplication, we should either just use the existing package, or inherit from that package and add a couple new things (instead of rewriting it).

watertap/property_models/IX_prop_pack.py
```diff
@@ -0,0 +1,1110 @@
...
+                "mass_frac_phase_comp": {"method": "_mass_frac_phase_comp"},
+                "dens_mass_phase": {"method": "_dens_mass_phase"},
+                "flow_vol_phase": {"method": "_flow_vol_phase"},
+                "flow_vol": {"method": "_flow_vol"},
+                "conc_mass_phase_comp": {"method": "_conc_mass_phase_comp"},
+                "conc_mol_phase_comp": {"method": "_conc_mol_phase_comp"},
+                "conc_equiv_phase_comp": {"method": "_conc_equiv_phase_comp"},
+                # "mole_frac_phase_comp": {"method": "_mole_frac_phase_comp"},
+                "visc_d_phase": {"method": "_visc_d_phase"},
+                "visc_k_phase": {"method": "_visc_k_phase"},
```
aladshaw3: It appears the only new things from the existing package are `conc_equiv_phase_comp` and `visc_k_phase`. Might be easier just to add these to the existing package.

watertap/property_models/IX_prop_pack.py
```diff
@@ -0,0 +1,1110 @@
...
+        flags = fix_state_vars(self, state_args)
+
+        for k in self.keys():
+            if self[k].is_property_constructed("visc_k_phase"):
+                self[k].visc_k_phase["Liq"].set_value(
+                    self[k].visc_d_phase["Liq"] / self[k].dens_mass_phase["Liq"]
+                )
+
+            if self[k].is_property_constructed("dens_mass_phase"):
+                self[k].dens_mass_phase["Liq"].set_value(1000)
```
aladshaw3: It may not be a problem here, but in general you do not want to use a property value in the initialization stage before you set its value in the initialization stage. 

Here, you are using `dens_mass_phase` to setup an initial value to `visc_k_phase`, but in the next lines (330-331) you are setting a new value for `dens_mass_phase`. These order of operations need to be reversed.

Also, because your function to set the value for `visc_k_phase` uses `dens_mass_phase` the `dens_mass_phase` will always be constructed if you have a `visc_k_phase`, so it looks a little redundant to then check to see if `dens_mass_phase` has been constructed. Again, in this case it probably doesn't matter, but this is just something to keep in mind when you are doing the initializations.

watertap/property_models/IX_prop_pack.py
```diff
@@ -0,0 +1,1110 @@
...
+
+        self.scaling_factor = Suffix(direction=Suffix.EXPORT)
+
+        # Add state variables
+
+        self.flow_mol_phase_comp = Var(
+            self.params.phase_list,
+            self.params.component_list,
+            initialize=0.1,
+            bounds=(1e-8, None),
```
aladshaw3:
```suggestion
            bounds=(0, None),
```

We are moving away from using these small numbers as the bounds and using 0 instead.

watertap/property_models/IX_prop_pack.py
```diff
@@ -0,0 +1,1110 @@
...
+
+    # -----------------------------------------------------------------------------
+    # Property Methods
+
+    def _flow_mass_phase_comp(self):
+        self.flow_mass_phase_comp = Var(
+            self.params.phase_list,
+            self.params.component_list,
+            initialize=0.5,
+            bounds=(1e-9, None),
```
aladshaw3:
```suggestion
            bounds=(0, None),
```

watertap/property_models/IX_prop_pack.py
```diff
@@ -0,0 +1,1110 @@
...
+        self.eq_flow_mass_phase_comp = Constraint(
+            self.params.component_list, rule=rule_flow_mass_phase_comp
+        )
+
+    def _flow_equiv_phase_comp(self):
+        self.flow_equiv_phase_comp = Var(
+            self.params.phase_list,
+            self.params.component_list,
+            initialize=100,
+            bounds=(1e-8, None),
```
aladshaw3:
```suggestion
            bounds=(0, None),
```

watertap/property_models/IX_prop_pack.py
```diff
@@ -0,0 +1,1110 @@
...
+        )
+
+        def rule_flow_equiv_phase_comp(b, j):
+            if j == "H2O":
+                return (
+                    b.flow_equiv_phase_comp["Liq", j] == b.flow_mol_phase_comp["Liq", j]
+                )
+            return b.flow_equiv_phase_comp["Liq", j] == b.flow_mol_phase_comp[
+                "Liq", j
+            ] / abs(b.params.charge_comp[j])
```
aladshaw3:
```suggestion
            ] * abs(b.params.charge_comp[j])
```

watertap/property_models/IX_prop_pack.py
```diff
@@ -0,0 +1,1110 @@
...
+        self.eq_flow_equiv_phase_comp = Constraint(
+            self.params.ion_set, rule=rule_flow_equiv_phase_comp
+        )
+
+    def _conc_mass_phase_comp(self):
+        self.conc_mass_phase_comp = Var(
+            self.params.phase_list,
+            self.params.component_list,
+            initialize=10,
+            bounds=(1e-89, None),
```
aladshaw3:
```suggestion
            bounds=(0, None),
```

watertap/property_models/IX_prop_pack.py
```diff
@@ -0,0 +1,1110 @@
...
+        self.eq_conc_mass_phase_comp = Constraint(
+            self.params.component_list, rule=rule_conc_mass_phase_comp
+        )
+
+    def _conc_mol_phase_comp(self):
+        self.conc_mol_phase_comp = Var(
+            self.params.phase_list,
+            self.params.component_list,
+            initialize=500,
+            bounds=(1e-8, None),
```
aladshaw3:
```suggestion
            bounds=(0, None),
```

watertap/property_models/IX_prop_pack.py
```diff
@@ -0,0 +1,1110 @@
...
+        self.eq_conc_mol_phase_comp = Constraint(
+            self.params.component_list, rule=rule_conc_mol_phase_comp
+        )
+
+    def _conc_equiv_phase_comp(self):
+        self.conc_equiv_phase_comp = Var(
+            self.params.phase_list,
+            self.params.component_list,
+            initialize=500,
+            bounds=(1e-8, None),
```
aladshaw3:
```suggestion
            bounds=(0, None),
```

watertap/property_models/IX_prop_pack.py
```diff
@@ -0,0 +1,1110 @@
...
+            units=pyunits.mol / pyunits.m**3,
+            doc="Equivalent concentration",
```
aladshaw3: You may want to better explain what is meant by "Equivalent Concentration" here.

watertap/property_models/IX_prop_pack.py
```diff
@@ -0,0 +1,1110 @@
...
+        self.eq_conc_equiv_phase_comp = Constraint(
+            self.params.component_list, rule=rule_conc_equiv_phase_comp
+        )
+
+    def _mass_frac_phase_comp(self):
+        self.mass_frac_phase_comp = Var(
+            self.params.phase_list,
+            self.params.component_list,
+            initialize=0.5,
+            bounds=(1e-9, 1.001),
```
aladshaw3:
```suggestion
            bounds=(0, 1.001),
```

watertap/property_models/IX_prop_pack.py
```diff
@@ -0,0 +1,1110 @@
...
+        def rule_dens_mass_phase(b):
+            return b.dens_mass_phase["Liq"] == 1000 * pyunits.kg * pyunits.m**-3
+
+        self.eq_dens_mass_phase = Constraint(rule=rule_dens_mass_phase)
+
+    def _flow_vol_phase(self):
+        self.flow_vol_phase = Var(
+            self.params.phase_list,
+            initialize=1,
+            bounds=(1e-8, None),
```
aladshaw3:
```suggestion
            bounds=(0, None),
```

watertap/property_models/IX_prop_pack.py
```diff
@@ -0,0 +1,1110 @@
...
+        # components
+        self.H2O = Solvent()
+
+        # blank sets
+        self.cation_set = Set()
+        self.anion_set = Set()
+        self.solute_set = Set()
+        self.ion_set = Set()
+        self.reference_cation = self.config.reference_cation
+        self.reference_anion = self.config.reference_anion
```
aladshaw3: It is not clear to me what these "reference_*" ions are for.

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -0,0 +1,1352 @@
...
+    **default** - None.
+    **Valid values:** {
+    see property package for documentation.}""",
+        ),
+    )
+
+    CONFIG.declare(
+        "target_ion",
+        ConfigValue(default="Ca_2+", domain=str, description="Target Ion"),
+    )
```
aladshaw3, line 136: Are you using this to enforce a matching name between the property package and the unit model? If so, I don't think it makes sense to put in a default value. This would need to be something that is always user supplied and if the user does not supply anything, then it should report an error. 

Also, I am not certain that you would want to make this an `Enum`. That may limit the number of ions that a user could target. If there really only are a few possible targets, then maybe. But in general, you should be able to target any ion that exists in the property package.

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -0,0 +1,1352 @@
...
+            doc="Pe particle eq A parameter",
+        )
+
+        self.Pe_p_exp = Param(
+            initialize=0.48, units=pyunits.dimensionless, doc="Pe particle eq exp"
+        )
+
+        self.Sh_A = Param(
+            initialize=1.09, units=pyunits.dimensionless, doc="Sh eq A parameter"
+        )
```
aladshaw3: A lot of these parameter names are unclear and the `doc` strings added don't clear things up. If these come from one of the references, it may be good to put in a note about that.

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -0,0 +1,1352 @@
...
+            bounds=(0.6, 12),
+            units=pyunits.m,
+            doc="Column height [m]",
+        )
+
+        self.col_vol = Var(
+            initialize=10,
+            units=pyunits.m**3,
+            doc="Column volume of one unit [m3]",
+        )
```
aladshaw3: Is there a difference between `bed_vol` and `col_vol`?

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -0,0 +1,1352 @@
...
+            units=pyunits.hr**-1,
+            doc="Service flow rate [BV/hr]",
+        )
+
+        self.pressure_drop = Var(
+            initialize=14,
+            units=pyunits.psi,
+            bounds=(0, 25),
+            doc="Pressure drop across column",  # max pressure drop is 25 psi, MWH
+        )
```
aladshaw3: If you use a 0D control volume block from IDAES, you probably don't need separate variables for hold_up and pressure_drop.

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -0,0 +1,1352 @@
...
+        self.bed_porosity.fix()
+        self.dimensionless_time.fix()
+        self.lh.fix()
+        self.regen_dose.fix()
+        self.regen_ww.fix()
+        self.regen_sg.fix()
+        self.t_bw.fix()
+        self.bw_rate.fix()
+        self.rinse_bv.fix()
+        self.pump_efficiency.fix()
```
aladshaw3: Should not be fixing variables in your object. That is up to the user to control.

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -0,0 +1,1352 @@
...
+            default=tmp_dict,
+        )
+
+        # Add outlet and waste block
+        tmp_dict["defined_state"] = False  # outlet and waste block is not an inlet
+        self.properties_out = self.config.property_package.state_block_class(
+            self.flowsheet().config.time,
+            doc="Material properties of outlet",
+            default=tmp_dict,
+        )
```
aladshaw3: If you use a 0D control volume block, you won't need to do these things.

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -0,0 +1,1352 @@
...
+            ion_set,
+            doc="Langmuir isotherm",
+        )
+        def eq_langmuir(b, j):
+            if j == target_ion:
+                return b.R_eq[j] == (
+                    b.c_norm[j] * (1 - b.resin_eq_capacity / b.resin_max_capacity)
+                ) / (b.resin_eq_capacity / b.resin_max_capacity * (1 - b.c_norm[j]))
+            else:
+                return Constraint.Skip
```
aladshaw3: Will this unit model be limited to only removing a single target ion? Can this be further generalized for a set of target ions?

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -0,0 +1,1352 @@
...
+        def eq_resin_eq_capacity(b):
+            return b.resin_eq_capacity <= b.resin_max_capacity
+
+        @self.Constraint(doc="Bed velocity ub")
+        def eq_vel_bed_ub(b):
+            return b.vel_bed <= b.vel_bed_ratio * b.vel_min
+
+        @self.Constraint(doc="Bed velocity lb")
+        def eq_vel_bed_lb(b):
+            return self.vel_bed >= self.vel_min
```
aladshaw3: Why do you have inequality constraints? Is this necessary?

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -0,0 +1,1352 @@
...
+
+        @self.Constraint(doc="Left hand side of constant pattern sol'n")
+        def eq_lh(b):
+            return b.lh == b.num_transfer_units * (b.dimensionless_time - 1)
+
+        @self.Constraint(ion_set, doc="Right hand side of fixed pattern soln")
+        def eq_fixed_pattern_soln(b, j):
+            return (
+                b.lh
+                == (log(b.c_norm[j]) - b.R_eq[j] * log(1 - b.c_norm[j]))
```
aladshaw3: Here you are using the `pyomo.environ log` function. This is going to have some pretty poor convergence behavior if your `b.c_norm` vars are very small. You may want to consider using a so-called "safe log" function. 

I believe IDAES has a "safe-log" function. @andrewlee94 may have some input here.

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -0,0 +1,1352 @@
...
+            res = opt.solve(blk, tee=slc.tee)
+        init_log.info("Initialization Step 3 {}.".format(idaeslog.condition(res)))
+        # ---------------------------------------------------------------------
+        # Release Inlet state
+        blk.properties_in.release_state(flags, outlvl=outlvl)
+        init_log.info("Initialization Complete: {}".format(idaeslog.condition(res)))
+
+    def calculate_scaling_factors(self):
+        super().calculate_scaling_factors()
+
```
aladshaw3, line 1242: You may definitely need to revisit the scaling (and initialization) for the model. This seems like a relatively complex model with many variables and I feel like this method for scaling (and initialization) is too simplistic.

aladshaw3 (changes requested):
Sorry it took me a while to get to this. I generally agree with Xiangyu's comments, but I added some other thoughts as well. My major concerns are (i) amounts of code duplication in the property packages, (ii) simplistic scaling of the unit model, and (iii) some oddities in the constraints (inequalities?...  and use of the log function that may have very poor convergence).

kurbansitterley:
@lbibl and @aladshaw3 thank you for the reviews. I am finally getting around to addressing the issues both of you raise and many will be addressed in my next push.

- Reconfiguring the ion DPSMDE prop pack to work with IX. Only a few changes were needed.
- I tried to implement a ``ControlVolume0D`` block but I was stuck on integrating the waste stream since the control volume only has an inlet and outlet. Could be a simple fix with some guidance, though.

watertap/property_models/IX_prop_pack.py
kurbansitterley: This will be fixed, thank you.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: I geared this model so that IX really could be any chemical. Presumably, the target ion (or component) would be part of the larger ``solute_list`` - so is it possible to make a ``Var`` a ``str``? Or how else could I have the user designate a target ion/compound?

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: Same response as above.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: The bounds are more about the costing data I have, less about the solver.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: Yes, thank you. I will clear up my doc strings in next push.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: I suppose I could add them to the DPMSDE prop pack...

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: I saw this in a different unit model but realized later the vars that were fixed were for regressions (and thus would likely not be changed by the user). I will eliminate these.

watertap/property_models/IX_prop_pack.py
kurbansitterley: Yes good call, I will be doing that.

watertap/property_models/IX_prop_pack.py
kurbansitterley: and ``flow_equiv_phase_comp`` -- I am wondering if it is a good idea to add a new pyunit ``equiv``? Right now both use ``mols`` because there is no equivalent pyunit as part of the standard set.

watertap/property_models/IX_prop_pack.py
kurbansitterley: Yes yes, good points. This was my first property pack and I have learned quite a bit since then. I will be using the DPSMDE prop pack moving forward.

watertap/property_models/IX_prop_pack.py
kurbansitterley: Good to know, I will edit the prop pack accordingly.

watertap/property_models/IX_prop_pack.py
kurbansitterley: What do you think about adding a ``equiv`` pyunit? May be a better question for @andrewlee94

andrewlee94: What exactly would `equiv` mean as a unit, and what would its conversion factors be? My initial impression is that the current units are correct, and that @aladshaw3 was referring more to the doc string.

watertap/property_models/IX_prop_pack.py
kurbansitterley: Resin manufacturers present selectivity coefficients in terms of a reference ion for the resin (its like saying, the resin has X times affinity for ion A over the reference ion B). In theory, you can get the selectivity coefficient for any combination of target and presaturant ions using this data. These aren't used in this implementation, but I included them in anticipation of future developments and having a large resin database. I will remove them.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: Yes I would like it to be possible for any component to be the ``target_ion`` (or maybe ``target_component`` would be better). The idea is that the resin has high selectivity for the target and all other components would pass through.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: Yes, I will make better doc strings.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: IX columns include the resin, a distributor, underdrain support, and must include additional height to accomodate bed expansion during backwash. The ``bed_vol`` will probably occupy the majority of the ``col_vol`` but not all of it.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: I tried to use the 0D control volume block but I will need some guidance.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: I would like it to be generalizable for a set of target ions. Those relationships are established for batched processes but I do not believe they can be applied to a fixed bed process. This is something I am actively researching.

watertap/unit_models/ion_exchange_0D.py
andrewlee94: The better approach is to use the IDAES constants module (which also has a number of other useful physical constants with units).

```
from idaes.core.util.constants import Constants

pi = Constants.pi
```

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: In an earlier version I had a ``Constraint`` for ``vel_min`` and I found these inequalities helped it solve. I will remove these and see how it works then.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: Yes agreed. We have had some conversations about both of these in our model developer meeting for this and both will be changed. The scaling will essentially be up to the user.

watertap/property_models/IX_prop_pack.py
aladshaw3: Yes, my comment is aimed more at the doc string. I believe `equiv` is referring to equivalent charge concentrations (i.e., moles of charge per volume), but correct me if I am wrong on this. If this is the case, then using `mol` as the units should be fine.

watertap/property_models/IX_prop_pack.py
aladshaw3: I think is makes sense to use `mol`. If my assumption is correct, then `equiv` is just denoting that this is a charge equivalent concentration (i.e., moles of charge per volume, or something similar). In which case, it is still technically moles of a substance, but based on moles of charge rather than moles of ions.

watertap/property_models/IX_prop_pack.py
kurbansitterley: Yes equivalent charge so ``equiv = mol * charge``. But if using ``mol`` is ok then fine with me.

watertap/unit_models/ion_exchange_0D.py
aladshaw3: So a fixed-bed process can be modeled as a batch process whose control volume moves at the speed of the fluid, so I would assume that almost anything that would apply to the batch would also apply to the fixed bed.

watertap/unit_models/ion_exchange_0D.py
lbibl: The target ions can be designated through the property package. This comment was on whether the regenerate agent should be limited to certain specific chemicals (e.g., HCl or another acid).

watertap/unit_models/ion_exchange_0D.py
lbibl: I guess this config can be removed since the target ion will be passed from the property package's components.

watertap/unit_models/ion_exchange_0D.py
lbibl: I meant to have these calculated by the property pack.  With a second look, these exist probably because the control volume block is not used.

watertap/unit_models/ion_exchange_0D.py
lbibl: My suggestion was considering making the model capable of solving problems in larger var domains (if they are practical), which will translate to the problem's constraints when it is solved.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: How can they be desginated through the property package if not through CONFIG though?

aladshaw3:
> @lbibl and @aladshaw3 thank you for the reviews. I am finally getting around to addressing the issues both of you raise and many will be addressed in my next push.
> 
> * Reconfiguring the ion DPSMDE prop pack to work with IX. Only a few changes were needed.
> * I tried to implement a `ControlVolume0D` block but I was stuck on integrating the waste stream since the control volume only has an inlet and outlet. Could be a simple fix with some guidance, though.

The `reverse_osmosis_0D` model is a good place to start as an example of how to use the `ControlVolume0D` with 2 outlets. It basically comes down to creating an additional `state_block` after the 0D control volume is created, then coupling the flow variables from that new state block with mass transfer terms from the 0D control volume.

watertap/property_models/IX_prop_pack.py
adam-a-a: @kurbansitterley note that if you are transitioning to using the ion property model and eliminating this extra prop model, you can merge main into your branch since a recent PR that was merged included setting all lower bounds (that were previously set to very small numbers) to 0.

adam-a-a (commented):
@kurbansitterley I _just_ realized that I never finalized my review, so all my comments have just been pending. Thus, most, if not all, might be outdated or already acknowledged by you--but I'm going to post anyway.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: Right, but I haven't seen anything in literature suggesting you can apply the batch equilibrium relationships to fixed-bed transitions; I could check again. Either way, I would prefer to get something merged for a single component before trying to expand it to multiple components.

aladshaw3: If there is a plan to expand to multi-component at some point, I would recommend going ahead and having the logic built into the unit model for dealing with a list of target ions, then adding an error check in the build or at specific constraints to report an error if the list is >1. It will be a lot more work to add in multiple species later if you don't plan for it now. 

Specifically, I would make the target ions config option to accept a list of strings (where each string is presumably the name of a target ion). Then, towards the front of the build check to see if the list is larger than 1 and report an error. It may also be worth checking to make sure that the name in the list actually exists as a component/solution/ion name from the property package. 

In the constraint (like this one), instead of checking if `j == target`, you could do `if j in target_list`. Then, that function will get applied whenever one of the j components is found in the list of targets. 

That is a more generalized way to build it that will also allow for easily extending later.

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -0,0 +1,1346 @@
...
+            ion_set,
+            doc="Langmuir isotherm",
+        )
+        def eq_langmuir(b, j):
+            if j == target_ion:
+                return b.separation_factor[j] == (
+                    b.c_norm[j] * (1 - b.resin_eq_capacity / b.resin_max_capacity)
+                ) / (b.resin_eq_capacity / b.resin_max_capacity * (1 - b.c_norm[j]))
+            else:
+                return Constraint.Skip
```
aladshaw3: This is a very odd form of Langmuir function. More commonly you see the Langmuir function take the following form:

`eq_cap = max_cap * (K * C) / (1 + K*C)`

Are we sure this is correct? Is this maybe not named correctly?

aladshaw3: Another thing to note here is that this constraint will probably cause convergence issues because you are dividing by a number that can become zero (i.e., the `1 - c_norm` part can become zero). 

You might want to consider reformatting this constraint to avoid the possibility of a divide by zero.

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -0,0 +1,1346 @@
...
+
+        @self.Constraint(doc="Resin equilibrium concentration bounds")
+        def eq_resin_eq_capacity(b):
+            return b.resin_eq_capacity <= b.resin_max_capacity
```
aladshaw3: I am willing to wager that you needed this inequality to prevent poor convergence that would have come from the manner in which the above 2 constraints are formulated. 

Theoretically, your `eq_capacity` should be able to become equal to your `max_capacity`, but if you let that happen then `separation_factor` in the above constraint goes to zero, which then causes your `selectivity` variable in the constraint above that to need to go towards infinity. 

I think a lot of this could be avoided if you reformulate the first 2 constraints.

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -0,0 +1,1346 @@
...
+        def eq_bed_area(b):
+            return b.bed_area == Constants.pi * (b.bed_diam / 2) ** 2
+
+        @self.Constraint(doc="Bed volume calculation")
+        def eq_bed_vol(b):
+            return self.bed_vol == self.bed_area * self.bed_depth
+
+        @self.Constraint(doc="Bed depth to bed diameter ratio")
+        def eq_bed_depth_to_diam_ratio(b):
+            return b.bed_depth / b.bed_diam >= b.bed_depth_to_diam_ratio
```
aladshaw3: What is this, why do we need it, and why is it an inequality? How would the depth-to-diam ratio ever be less than the ratio of the depth to the diam?

aladshaw3:
@kurbansitterley 

Changes you have made in the ion prop pack are causing numerous scaling failures in several other models that use this package. 

You will need to dive deeper into what exactly each failure is indicating, but here is a sample of a failure in the Boron Removal unit:

```

________________ TestBoronRemoval_IonPropPack_Min.test_scaling _________________
self = <watertap.unit_models.tests.test_boron_removal.TestBoronRemoval_IonPropPack_Min object at 0x7fee95b93290>
min_boron_removal_model = <pyomo.core.base.PyomoModel.ConcreteModel object at 0x7fee961455f0>
@pytest.mark.unit
deftest_scaling(self, min_boron_removal_model):
        m = min_boron_removal_model
# Set some scaling factors and look for 'bad' scaling
        scaling_setup(m)
# check that all variables have scaling factors
        unscaled_var_list = list(iscale.unscaled_variables_generator(m))
assertlen(unscaled_var_list) == 0
# check if any variables are badly scaled
        badly_scaled_var_values = {
            var.name: val
for (var, val) in iscale.badly_scaled_var_generator(
                m, large=1e3, small=1e-3
            )
        }
>       assertnot badly_scaled_var_values
E       assert not {"fs.unit.control_volume.properties_in[0.0].flow_equiv_phase_comp[Liq,'B[OH]4_-']": 100000.0, "fs.unit.control_volume.properties_out[0.0].flow_equiv_phase_comp[Liq,'B[OH]4_-']": 100000.0}
watertap/unit_models/tests/test_boron_removal.py:264: AssertionError
```


What this is showing me is that there is poor scaling for the new `flow_equiv_phase_comp` variable/property that was added. 

Now, what makes this error even more concerning is the fact that this property should not have even been built (unless somehow the `badly_scaled_var_generator` function somehow touches properties that were not invoked). 

This may mean that somewhere in the changes to the property package, these new properties are being touched and constructed, even for models that don't use or need them. 


Solution here is one of two things: (i) Find what changes you made that forced these properties to be invoked for models that don't need them and make a correction to that change so they don't get built or (ii) provide better scaling factors for them so that it doesn't matter that they get built.

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -1308,6 +1391,12 @@ def calculate_scaling_factors(self):
                 )
                 iscale.set_scaling_factor(self.flow_mol_phase_comp["Liq", j], sf)
 
+            if iscale.get_scaling_factor(self.flow_equiv_phase_comp["Liq", j]) is None:
+                sf = iscale.get_scaling_factor(
+                    self.flow_mol_phase_comp["Liq", j], default=1, warning=True
+                )
+                iscale.set_scaling_factor(self.flow_equiv_phase_comp["Liq", j], sf)
+
```
aladshaw3: This (and others like this) may be the culprit to the issue of the bad scaling in other tests of models. You are not checking to see if this property is constructed here, so as soon as you touch this variable, the properties get constructed (even for the models that don't use them).

adam-a-a: @kurbansitterley @aladshaw3 I just hopped on the branch and made a couple fixes to the ion prop model changes. I don't know it'll fix all failed tests, but it should repair at least a couple.

aladshaw3: Looks like that fixed most of the issues. There are now only 3 failing tests, one of which probably just needs to have the assert changed due to the new set of constraints. The other 2 are bad scaling vars that show up in the `Electrodialysis_0D` and `GAC` models.

aladshaw3:
**Remaining test failures:**

```

______ TestElectrodialysis_withNeutralSPecies.test_initialization_scaling ______
self = <watertap.unit_models.tests.test_electrodialysis_0D.TestElectrodialysis_withNeutralSPecies object at 0x7f97a95de700>
electrodialysis_cell3 = <pyomo.core.base.PyomoModel.ConcreteModel object at 0x7f97a95d3130>
@pytest.mark.component
deftest_initialization_scaling(self, electrodialysis_cell3):
        m = electrodialysis_cell3
# set default scaling for state vars
        m.fs.properties.set_default_scaling(
"flow_mol_phase_comp", 1e2, index=("Liq", "H2O")
        )
        m.fs.properties.set_default_scaling(
"flow_mol_phase_comp", 1e4, index=("Liq", "Na_+")
        )
        m.fs.properties.set_default_scaling(
"flow_mol_phase_comp", 1e4, index=("Liq", "Cl_-")
        )
        m.fs.properties.set_default_scaling(
"flow_mol_phase_comp", 1e5, index=("Liq", "N")
        )
        iscale.calculate_scaling_factors(m.fs)
        initialization_tester(m)
        badly_scaled_var_values = {
            var.name: val for (var, val) in iscale.badly_scaled_var_generator(m)
        }
>       assertnot badly_scaled_var_values
E       AssertionError: assert not {'fs.unit.nonelec_flux_in[0.0,Liq,Cl_-]': 3.5946834879070216e-10, 'fs.unit.nonelec_flux_in[0.0,Liq,Na_+]': 1.7207300709562899e-09}
watertap/unit_models/tests/test_electrodialysis_0D.py:697: AssertionError
----------------------------- Captured stdout call -----------------------------
Warning: 6 02:56:58 [WARNING] idaes.core.util.scaling: Missing scaling factor for fs.unit.solute_diffusivity_membrane
Warning: 6 02:56:58 [WARNING] idaes.core.util.scaling: Missing scaling factor for fs.unit.membrane_thickness
Warning: 6 02:56:58 [WARNING] idaes.core.util.scaling: Missing scaling factor for fs.unit.water_permeability_membrane
Warning: 6 02:56:58 [WARNING] idaes.core.util.scaling: Missing scaling factor for fs.unit.cell_length
Warning: 6 02:56:58 [WARNING] idaes.core.util.scaling: Missing scaling factor for fs.unit.cell_width
Warning: 6 02:56:58 [WARNING] idaes.core.util.scaling: Missing scaling factor for fs.unit.spacer_thickness
Warning: 6 02:56:58 [WARNING] idaes.core.util.scaling: Missing scaling factor for fs.unit.membrane_areal_resistance
Warning: 6 02:56:58 [WARNING] idaes.core.util.scaling: Missing scaling factor for fs.unit.electrodes_resistance
Warning: 6 02:56:58 [WARNING] idaes.core.util.scaling: Missing scaling factor for fs.unit.current
Warning: 6 02:56:58 [WARNING] idaes.core.util.scaling: Missing scaling factor for fs.unit.voltage
2022-07-16 02:56:59 [INFO] idaes.init.fs.unit.diluate_channel: Initialization Complete
2022-07-16 02:56:59 [INFO] idaes.init.fs.unit.concentrate_channel: Initialization Complete
2022-07-16 02:56:59 [INFO] idaes.init.fs.unit: Initialization Complete: optimal - Optimal Solution Found
2022-07-16 02:56:59 [INFO] idaes.init.fs.unit: Initialization Complete: optimal - Optimal Solution Found
------------------------------ Captured log call -------------------------------
WARNING  idaes.core.util.scaling:scaling.py:256 Missing scaling factor for fs.unit.solute_diffusivity_membrane
WARNING  idaes.core.util.scaling:scaling.py:256 Missing scaling factor for fs.unit.membrane_thickness
WARNING  idaes.core.util.scaling:scaling.py:256 Missing scaling factor for fs.unit.water_permeability_membrane
WARNING  idaes.core.util.scaling:scaling.py:256 Missing scaling factor for fs.unit.cell_length
WARNING  idaes.core.util.scaling:scaling.py:256 Missing scaling factor for fs.unit.cell_width
WARNING  idaes.core.util.scaling:scaling.py:256 Missing scaling factor for fs.unit.spacer_thickness
WARNING  idaes.core.util.scaling:scaling.py:256 Missing scaling factor for fs.unit.membrane_areal_resistance
WARNING  idaes.core.util.scaling:scaling.py:256 Missing scaling factor for fs.unit.electrodes_resistance
WARNING  idaes.core.util.scaling:scaling.py:256 Missing scaling factor for fs.unit.current
WARNING  idaes.core.util.scaling:scaling.py:256 Missing scaling factor for fs.unit.voltage
__________________ TestGACSimplified.test_var_scaling_robust ___________________
self = <watertap.unit_models.tests.test_gac.TestGACSimplified object at 0x7f97a959d940>
gac_frame_robust = <pyomo.core.base.PyomoModel.ConcreteModel object at 0x7f97c3b239f0>
@pytest.mark.component
deftest_var_scaling_robust(self, gac_frame_robust):
        m = gac_frame_robust
>       assert badly_scaled_var_lst == []
E       assert [(<pyomo.core...554e-07), ...] == []
E         Left contains 7 more items, first extra item: (<pyomo.core.base.var.ScalarVar object at 0x7f97c31a6740>, 3.785230475501351e-06)
E         Full diff:
E           [
E         -  ,
E         +  (<pyomo.core.base.var.ScalarVar object at 0x7f97c31a6740>,
E         +   3.785230475501351e-06),
E         +  (<pyomo.core.base.var.ScalarVar object at 0x7f97c31a6c80>,
E         +   3.028184380400466e-07),
E         +  (<pyomo.core.base.var.ScalarVar object at 0x7f97c31a65f0>,
E         +   9.827944589010264e-07),
E         +  (<pyomo.core.base.var.ScalarVar object at 0x7f97c31a6040>,
E         +   4.3242956191644927e-07),
E         +  (<pyomo.core.base.var.ScalarVar object at 0x7f97c31a60b0>,
E         +   2.113013522929153e-05),
E         +  (<pyomo.core.base.var.ScalarVar object at 0x7f97c31a6cf0>,
E         +   3.251727557656554e-07),
E         +  (<pyomo.core.base.var.ScalarVar object at 0x7f97c32447b0>,
E         +   4.516288282408317e-09),
E           ]
watertap/unit_models/tests/test_gac.py:344: AssertionError
__________________________ TestIonExchange.test_build __________________________
self = <watertap.unit_models.tests.test_ion_exchange_0D.TestIonExchange object at 0x7f97a95a7d00>
IX_frame = <pyomo.core.base.PyomoModel.ConcreteModel object at 0x7f97c30eb130>
@pytest.mark.unit
deftest_build(self, IX_frame):
        m = IX_frame
        ix = m.fs.unit
.
.
.
for sb_str in stateblock_lst:
            sb = getattr(ix, sb_str)
assertisinstance(sb, DSPMDEStateBlock)
# test statistics
assert number_variables(m) == 126
>       assert number_total_constraints(m) == 86
E       assert 89 == 86
E        +  where 89 = number_total_constraints(<pyomo.core.base.PyomoModel.ConcreteModel object at 0x7f97c30eb130>)
watertap/unit_models/tests/test_ion_exchange_0D.py:314: AssertionError
```

lbibl:
I made a major modification- to index the two equiv vars  (`flow_equiv...` and `conc_equiv...`) by `ion_set | solute_set`.  I think this makes more sense because we should never need a charge equivalent for H2O and assigning a charge equivalent of H2O non-zero looks incorrect to me.  This led me to change two numbers (number of vars essentially) in the test_ix but did not affect any model outcomes.  I highly recommend to rename the `equiv` as `charge_equiv`, or `coulomb_equiv` or an abbreviation of either, anything to show what the "equiv" refers to.   I fixed the scaling of ED_0d and the early-shown error was insignificant.  Test_gac passed with x-fail on my local and I am waiting to see how this commit performs by gh action.

lbibl:
I took a look at the test_gac file.  I think the error can be fixed by only testing `badly_scaled_var` after the model is initialized  because running the initialization routine will trigger vars from the prop pack to be set at proper initial values that will affect the meaningful scaling.  In the actual running, it should goes 1) set default or known scaling factors 2) initialize, and 3) solve.  I think testing scaling quality after 2) initialize would be most meaningful.

aladshaw3:
Looking good. Only remaining test failure is in the scaling of GAC model. 

Maybe @adam-a-a or @hunterbarber can have a look at it. I think @lbibl comment above is a good suggestion. Move the scaling check after initialization.

hunterbarber:
The only reason I could see that would relate to the GAC model and testing is the changes made in the `ion_DSPMDE_prop_pack`. Some robustness to variable scaling will likely be added to finalize the GAC model in the near future, but I need to look at this in more detail on Monday to better diagnose what's causing the test failure. Then I can suggest whether adjustments need made to the property package or to delay until the GAC model is edited.

hunterbarber:
> The only reason I could see that would relate to the GAC model and testing is the changes made in the `ion_DSPMDE_prop_pack`. Some robustness to variable scaling will likely be added to finalize the GAC model in the near future, but I need to look at this in more detail on Monday to better diagnose what's causing the test failure. Then I can suggest whether adjustments need made to the property package or to delay until the GAC model is edited.

My guess is that the changes to `flow_mol_phase_comp` and how they relate to `conc_mass_phase_comp` are generating the problem, still looking into solutions.

adam-a-a:
> My guess is that the changes to `flow_mol_phase_comp` and how they relate to `conc_mass_phase_comp` are generating the problem, still looking into solutions

@hunterbarber which changes are you referring to specifically? Can you point out the lines? After a quick scan of the ion prop model changes in this PR, I didn't notice any changes directly made to `flow_mol_phase_comp`

hunterbarber:
> > My guess is that the changes to `flow_mol_phase_comp` and how they relate to `conc_mass_phase_comp` are generating the problem, still looking into solutions
> 
> @hunterbarber which changes are you referring to specifically? Can you point out the lines? After a quick scan of the ion prop model changes in this PR, I didn't notice any changes directly made to `flow_mol_phase_comp`

@adam-a-a Sorry you're right, it was pulling scaling factors from flow_mol to write to flow_equiv

aladshaw3:
More context: Here are the vars that are badly scaled.

```
N_Bi : Biot number
    Size=1, Index=None
    Key  : Lower : Value                 : Upper : Fixed : Stale : Domain
    None :     0 : 3.785230475501351e-06 :  None : False : False : NonNegativeReals
3.785230475501351e-06

min_N_St : Minimum Stanton number to achieve a constant pattern solution
    Size=1, Index=None
    Key  : Lower : Value                 : Upper : Fixed : Stale : Domain
    None :     0 : 3.028184380400466e-06 :  None : False : False : NonNegativeReals
3.028184380400466e-07

min_ebct : Minimum empty bed contact time to achieve a constant pattern solution
    Size=1, Index=None, Units=s
    Key  : Lower : Value                 : Upper : Fixed : Stale : Domain
    None :     0 : 9.827944589010263e-05 :  None : False : False : NonNegativeReals
9.827944589010264e-07

min_res_time : Minimum fluid residence time in the adsorber bed to achieve a constant pattern solution
    Size=1, Index=None, Units=s
    Key  : Lower : Value                 : Upper : Fixed : Stale : Domain
    None :     0 : 4.324295619164492e-05 :  None : False : False : NonNegativeReals
4.3242956191644927e-07

min_elap_time : Minimum elapsed time between GAC replacement in adsorber bed in operationto achieve a constant pattern solution
    Size=1, Index=None, Units=s
    Key  : Lower : Value             : Upper : Fixed : Stale : Domain
    None :     0 : 2.113013522929153 :  None : False : False : NonNegativeReals
2.113013522929153e-05

ebct_mtz_replace : Empty bed contact time of the mass transfer zoneat the time of replacement
    Size=1, Index=None, Units=s
    Key  : Lower : Value                 : Upper : Fixed : Stale : Domain
    None :     0 : 3.251727557656554e-05 :  None : False : False : NonNegativeReals
3.251727557656554e-07

length_mtz_replace : Length of the mass transfer zone at the time of replacement
    Size=1, Index=None, Units=m
    Key  : Lower : Value                  : Upper : Fixed : Stale : Domain
    None :     0 : 4.5162882824083166e-08 :  None : False : False : NonNegativeReals
4.516288282408317e-09
```

aladshaw3:
@adam-a-a @hunterbarber 

After looking at the GAC model in more depth, I think the entirety of the GAC model's scaling needs to be revisited. You are currently only applying scaling to variables (not the constraints) and all the scaling is very simplified (just defined constant values). 

Based on this, my feeling is that this failing test should be skipped for the time being to not hold up this PR and then an issue needs to be opened for revisiting the scaling of the GAC model (and put that at high priority). It doesn't make sense to me to rework all the GAC scaling in this PR.

hunterbarber:
@aladshaw3 @adam-a-a 

I agree that the variable scaling needs improved, but it is still concerning that a change made here differs from the current main repo and test file results. Whether it is the solver or scaling I think may be important to identify.

aladshaw3:
> @aladshaw3 @adam-a-a
> 
> I agree that the variable scaling needs improved, but it is still concerning that a change made here differs from the current main repo and test file results. Whether it is the solver or scaling I think may be important to identify.

So, I don't think this is a solver issue because the GAC model still solves and still produces the same results (at least the ones that are being compared in the last test). If you feel that the model is producing false results, then we need to check more in depth on the full set of GAC results, especially for the vars that have the bad scaling to see if they have changed significantly.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: Yes it is correct.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: The ratio of the bed depth to diameter needs to greater than some value (I have seen >2.5-5) so the hydrodynamics of the column are such that assumptions of the model hold and it results in a valid design. If `bed_depth_to_diam_ratio` is a confusing, maybe should be renamed to `min_bed_depth_to_diam_ratio`.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: I added this constraint because I was getting valid solves where `resin_eq_capacity` was greater than `resin_max_capacity`, which shouldn't happen.

kurbansitterley:
> I made a major modification- to index the two equiv vars (`flow_equiv...` and `conc_equiv...`) by `ion_set | solute_set`. I think this makes more sense because we should never need a charge equivalent for H2O and assigning a charge equivalent of H2O non-zero looks incorrect to me. This led me to change two numbers (number of vars essentially) in the test_ix but did not affect any model outcomes. I highly recommend to rename the `equiv` as `charge_equiv`, or `coulomb_equiv` or an abbreviation of either, anything to show what the "equiv" refers to. I fixed the scaling of ED_0d and the early-shown error was insignificant. Test_gac passed with x-fail on my local and I am waiting to see how this commit performs by gh action.

@aladshaw3 @adam-a-a @lbibl Should I just eliminate `equiv` from the property pack? It is more of a convenience really and since `charge` is part of the property package already I could just replace any use of `equiv` with `mol * charge` in the unit model.

hunterbarber:
@kurbansitterley @aladshaw3 

It also seems like `visc_k_phase` is being constructed in the state blocks of the GAC test model even though it is not being touched/invoked. Should this be made mutable?

kurbansitterley:
> @kurbansitterley @aladshaw3
> 
> It also seems like `visc_k_phase` is being constructed in the state blocks of the GAC test model even though it is not being touched/invoked. Should this be made mutable?

I had a line for `set_default_scaling` for `visc_k_phase` that I removed in this last commit. Maybe that was causing it to be constructed?

lbibl:
I located the issue being several vars were built even when the unit model did not call because of the missing `is_property_constructed()` in the scaling function.   This caused the earlier minor issue for my ED_0D model.  I have this fixed.  But that should not have caused any detrimental issues -  it just led to extra var being built so the scaling turned out different. The scaling is technically ok when the problem is solved as desired.  @hunterbarber may try if this modification can pass your current GAC file (mine system is MacOS so the file always passed with x-fail for some reason... ) .  With this fix, the rest of the property pack should be the same as the current main for your file.   But I agree that scaling in GAC can be advanced.

lbibl:
> > @kurbansitterley @aladshaw3
> > It also seems like `visc_k_phase` is being constructed in the state blocks of the GAC test model even though it is not being touched/invoked. Should this be made mutable?
> 
> I had a line for `set_default_scaling` for `visc_k_phase` that I removed in this last commit. Maybe that was causing it to be constructed?

Oops... we operated at the same time.  My commit had the issue resolved so you can put the `set_default_scaling` back.

aladshaw3:
> > > @kurbansitterley @aladshaw3
> > > It also seems like `visc_k_phase` is being constructed in the state blocks of the GAC test model even though it is not being touched/invoked. Should this be made mutable?
> > 
> > 
> > I had a line for `set_default_scaling` for `visc_k_phase` that I removed in this last commit. Maybe that was causing it to be constructed?
> 
> Oops... we operated at the same time. My commit had the issue resolved so you can put the `set_default_scaling` back.

Before anymore changes, let's see if all the tests pass now. I am also running these tests locally with all gac tests active. If this passes, then we may not need any additional changes (for now).

watertap/unit_models/tests/test_gac.py
```diff
@@ -338,9 +340,14 @@ def test_initialize_robust(self, gac_frame_robust):
...
     @pytest.mark.requires_idaes_solver
     @pytest.mark.component
+    @pytest.mark.skip("Scaling for GAC model needs to be revisited")
     def test_var_scaling_robust(self, gac_frame_robust):
         m = gac_frame_robust
         badly_scaled_var_lst = list(badly_scaled_var_generator(m))
+        for i in badly_scaled_var_lst:
+            i[0].pprint()
+            print(i[1])
+            print()
```
aladshaw3:
```suggestion
    def test_var_scaling_robust(self, gac_frame_robust):
        m = gac_frame_robust
        badly_scaled_var_lst = list(badly_scaled_var_generator(m))
```

Reactivate this test before merging

lbibl:
Thanks to all.  My conclusion is [06e16be](https://github.com/watertap-org/watertap/pull/559/commits/06e16bee843febd5a268e5437b99cd427f093075)  made by me should resolved the problem for the best.  (you wanna keep the setting default scaling for `visa_k_phase` since it might be useful in other scenarios).   I think the authorized user can revoke the last two commits or make a small change.  I can also push my branch again to conclude.  Let m know if needed.

watertap/property_models/tests/test_ion_DSPMDE_prop_pack.py
```diff
@@ -389,6 +390,7 @@ def test_default_scaling(model3):
         ("dens_mass_phase", "Liq"): 1e-3,
         ("visc_d_phase", "Liq"): 1e3,
         ("diffus_phase_comp", "Liq"): 1e10,
+        ("visc_k_phase", "Liq"): 1e6,
```
aladshaw3:
```suggestion
```
Need to remove this line if you are not adding this as a default scaled var anymore.

lbibl: I recommend keeping this and the other removed line because it is a legitimate scaling for `visa_k_phase`.  See my last comment.

kurbansitterley:
> > > > @kurbansitterley @aladshaw3
> > > > It also seems like `visc_k_phase` is being constructed in the state blocks of the GAC test model even though it is not being touched/invoked. Should this be made mutable?
> > > 
> > > 
> > > I had a line for `set_default_scaling` for `visc_k_phase` that I removed in this last commit. Maybe that was causing it to be constructed?
> > 
> > 
> > Oops... we operated at the same time. My commit had the issue resolved so you can put the `set_default_scaling` back.
> 
> Before anymore changes, let's see if all the tests pass now. I am also running these tests locally with all gac tests active. If this passes, then we may not need any additional changes (for now).

Oops sorry, saw this too late and re-added.

lbibl:
> 

@aladshaw3  Please make sure you get my comment.   I don't think we should delete the default scaling for `visc_k_phase` because it should be scaled when constructed.

aladshaw3:
@kurbansitterley @lbibl 

Please run tests locally before committing changes. 

The latest changes from @kurbansitterley cause 3 test failures on my machine. Run the tests locally first to make sure that things are good on your end.

watertap/unit_models/tests/test_gac.py
aladshaw3:
```suggestion
    def test_var_scaling_robust(self, gac_frame_robust):
        m = gac_frame_robust
        badly_scaled_var_lst = list(badly_scaled_var_generator(m))
```

**After the current checks are done. Try committing this suggestion and re-running the tests. It passes now on my local machine. 

This should be the last changed needed in this PR.**

lbibl: Thank you @aladshaw3 .  Yes, it looks the same to me.  The current commit should maintain the original ion property pack intact but only with new vars added.   @hunterbarber  's test_gac should pass with its current version.  we can reactivate that skipped test if we want.  I recommend @hunterbarber to consider testing scaling after modeling being initialized if he is going to improve the scaling method.

hunterbarber: Yes, a GAC model overhaul will be done in a subsequent PR including improvements to scaling. The GAC model should pass without changes to the `test_gac` for this PR, which I also confirmed locally through previous commits after correcting the new variable introductions.

watertap/unit_models/ion_exchange_0D.py
aladshaw3: That leads me to believe that there is a missing constraint here for the "site balance":

Constraint:  `max_cap = available_sites + eq_cap`

You may want to add a variable for `available_sites`, then add in the above constraint. This would be a better way of preventing `eq_cap` from being larger than `max_cap`

watertap/unit_models/ion_exchange_0D.py
lbibl: With a second read,  is your target ion CONFIG to determine which ion to be *selectively* exchanged?   OK to do it this way if this is your point.  but I would wonder, if the solution contains multiple ions, whether you can determine what the target ion is just by choosing a config.  My idea is that the selectivity can be modeled by affinity or material selectivity data for different ions.   In that case, only property block data is needed.  Another suggestion was to make the config domain more specific, something like `In( )` to assign a candidate list.  `Str` would be too loose.  Same idea for the regenerate agent CONFIG.    Depending on your goal for this version, these comments can be treated at your discretion.

aladshaw3 (approved):
LGTM.

I left some additional comments in issue #599 for things to consider refining further in future PRs.

hunterbarber:
- [ ] resin type as enumeration, global declaration of params, and general costing cleanup for this first PR
- [ ] operating costs

watertap/costing/watertap_costing_package.py
```diff
@@ -857,6 +858,50 @@ def cost_crystallizer_by_volume(blk):
             )
         )
 
+    @staticmethod
+    def cost_ion_exchange(blk):
```
adam-a-a, line 1098:
```suggestion
    def cost_ion_exchange(blk, resin_cost_type):

```

watertap/costing/watertap_costing_package.py
```diff
@@ -857,6 +858,50 @@ def cost_crystallizer_by_volume(blk):
...
+            "strong_base_styrenic_gel_2": 6116,
+            "strong_base_styrenic_macro_1": 7298,
+            "strong_base_styrenic_macro_2": 7810,
+            "strong_base_polyacrylic": 8658,
+            "strong_base_nitrate": 6116,
+        }  # cost of resin per m3, adapted to $/m3 from EPA models
+        make_capital_cost_var(blk)
+        num_cols = 2
+        col_cost_params = [9120, 0.49]
+        resin_cost_params = resin_dict["strong_acid_macro"]
```
```suggestion
        resin_cost_params = resin_dict[resin_cost_type]

```

adam-a-a: @kurbansitterley like @TimBartholomew mentioned during our call, you could probably follow the example of the costing the RO membrane, where we have costs for whether a high-pressure module is being used as opposed to "conventional" RO module.

lbibl:
> > I made a major modification- to index the two equiv vars (`flow_equiv...` and `conc_equiv...`) by `ion_set | solute_set`. I think this makes more sense because we should never need a charge equivalent for H2O and assigning a charge equivalent of H2O non-zero looks incorrect to me. This led me to change two numbers (number of vars essentially) in the test_ix but did not affect any model outcomes. I highly recommend to rename the `equiv` as `charge_equiv`, or `coulomb_equiv` or an abbreviation of either, anything to show what the "equiv" refers to. I fixed the scaling of ED_0d and the early-shown error was insignificant. Test_gac passed with x-fail on my local and I am waiting to see how this commit performs by gh action.
> 
> @aladshaw3 @adam-a-a @lbibl Should I just eliminate `equiv` from the property pack? It is more of a convenience really and since `charge` is part of the property package already I could just replace any use of `equiv` with `mol * charge` in the unit model.

I don't have a particular preference about this.  Using a charge equivalent avoids coding the *z_i* in equations dealing with charges.  Since it is already added, I would probably just keep it there and future users can always choose to work with either way.

watertap/costing/watertap_costing_package.py
hunterbarber: It seems like `IonExchangeType()` has been implemented in place of this suggestion. Though, it is defined in the unit model where it is unused and passed to the costing package model. Should this instead be implemented as `MixerType()` given it has no effect on the file it is constructed in?

watertap/unit_models/tests/test_electrodialysis_0D.py
```diff
@@ -687,7 +687,7 @@ def test_initialization_scaling(self, electrodialysis_cell3):
             "flow_mol_phase_comp", 1e4, index=("Liq", "Cl_-")
         )
         m.fs.properties.set_default_scaling(
-            "flow_mol_phase_comp", 1e5, index=("Liq", "N")
+            "flow_mol_phase_comp", 1e6, index=("Liq", "N")
```
hunterbarber, line 696: Unclear about this change, both 1e5 and 1e6 pass locally for me and seemingly unrelated to other changes.

watertap/unit_models/ion_exchange_0D.py
hunterbarber: Costing data bounds may be indicative of validated results, but even then data may be extrapolated if the user so chooses. In this case, bounds would need expanded and potentially accompanying tests validated large- and small-scale results.

hunterbarber (commented):
I agree with many of the comments provided by others on this first model that should be carried into subsequent PRs. Couldn't determine the root cause of the non-macOS failures but they do also fail locally for me. I can rereview after tests are passing.

watertap/unit_models/tests/test_electrodialysis_0D.py
kurbansitterley: I included `IonExchangeType()` in the unit model itself in anticipation of using it in future refinements of model. I'd prefer to keep it in IX for that reason. 

I didn't make the second change, maybe it is a result of me merging from main? Doesn't matter if it is reverted or not.

lbibl: This is a change made by me (I wrote the corresponding file) and it is a trivial change due to a previous failed check but things have been updated since then.  Mark this resolved if no new error is related to this.

lbibl:
Per the discussion in the electrochem meeting today, I am creating a PR (#668) to take over the changes on the DSPMDE property parts  made in this PR, so the changes here can be timely merged to facilitate following works of others.  It's currently in draft for testing. Once I address that to a ready version, I will invite folks to review.  Thanks.

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -1308,11 +1392,29 @@ def calculate_scaling_factors(self):
...
                 iscale.set_scaling_factor(self.mw_comp[j], value(v) ** -1)
 
+        for j, v in self.charge_comp.items():
+            if iscale.get_scaling_factor(v) is None:
+                if value(v) == 0:
+                    sf = 1
+                else:
+                    sf = abs(value(v)) ** -1
+                iscale.set_scaling_factor(self.charge_comp[j], sf)
+
```
lbibl, line 1412: Is there an apparent reason for this scaling being needed? Did you see errors without this scaling?  In principle there is no need to scale constant parameters, even more so for this charge parameter which always stays a single digit.  I recommend removing this change if no special reason.

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -1418,6 +1524,22 @@ def calculate_scaling_factors(self):
...
+        if self.is_property_constructed("conc_equiv_phase_comp"):
+            for j in self.params.ion_set | self.params.solute_set:
+                if (
+                    iscale.get_scaling_factor(self.conc_equiv_phase_comp["Liq", j])
+                    is None
+                ):
+                    if self.is_property_constructed("charge_comp"):
+                        sf = iscale.get_scaling_factor(
+                            self.conc_mol_phase_comp["Liq", j]
+                        ) * iscale.get_scaling_factor(self.charge_comp[j])
```
lbibl: So I think if `conc_equiv` is constructed, "charge" must be constructed as part of the method.  Along with my last comment on the unnecessary scaling of "charge",  I suggest this if statement be removed.  I don't think adding a coefficient of 1/charge (usually 1, 2 or 3)  significantly affect sf carried by the conc or trigger any differences for the solver.

watertap/property_models/ion_DSPMDE_prop_pack.py
kurbansitterley: I didn't see any errors without the scaling. My only thought was since other params like `mw_comp` are scaled that charge should be also. Particularly because the IX model makes heavy use of `conc_equiv`, I would want those variables as well scaled as possible. But your point is taken and will remove in next push.

kurbansitterley: However I will also add that this PR is failing tests right now because of differences in results between the Win and Mac solvers. Could that be a result of poor scaling that removing this change could help?

andrewlee94: @kurbansitterley If the solver found an optimal solution on both systems but gave different solutions, then the cause is (almost) always poor scaling (more generally, any difference is solver behaviour is likely to be due to scaling).

lbibl: The current version of this PR fails for both Win and Linux but passes MacOS.  Aside from your Ix file, two other failing tests are test_posttreatment and test_lssro_multi_sweep.py.   The draft PR #668  takes over the property package in this PR without adding the scaling on change_comp and is not showing failures on these two files so that suggests removing them is better.   But, that version (in PR #668) still has some issue as I described there.   I recommend that at this PR the IX file be focused only and  investigations on DSPMDE instead be conducted by PR #668.  By this way we may target things a little clearly.   If we agreed,  @kurbansitterley can revoke that charge scaling (where I commented) locally.  That means to have a DSPMDE as in PR #668 .  You can then test your IX file, scaling or other things, locally.  If you find critical thing about DSPMDE, you may mention that under #668.

kurbansitterley: Ok thank you for the feedback and info. I will remove this scaling for `charge_comp`.

watertap/property_models/ion_DSPMDE_prop_pack.py
adam-a-a: One note here, @lbibl - `charge_comp` isn't necessarily constructed on demand (if that's what you meant). From what I recall doing, charge_comp is populated by the user through a config option, those data are passed into the corresponding parameter, and later I add an object reference to that param (or fixed var) so that it seems that charge_comp is built on demand, but it is really just adding a reference to something that already existed on the model.

lbibl: Thanks @adam-a-a for your clarification.   Get you.  I think my comment should be minor since charge is really a small number.  Since taking out the `charge_comp` out of scaling does help passing two other files across os and did not affect this file, I left them out in the DSPMDE PR #668 , which is ready to be reviewed now.

kurbansitterley:
This PR was becoming a disaster for me to manage with updates to main so I created a new one: PR #684

PR #564, opened 2022-06-10
Updating property names to better follow standards

### LINKED ISSUE #513: Standardize property names
Some property packages have different names for the same property (e.g., `pressure_osm` vs `pressure_osm_phase`). This may be confusing to the average user. We should go through all property packages and use standardized names. 

May require some discussion on what names should be used for what property.

aladshaw3:
## Fixes/Resolves:

#513 

## Summary/Motivation:

Want all property packages to use the same naming conventions for common properties (such as densities, viscosities, presssures, etc). This PR will make sure that all current properties follow the standard naming conventions mentioned in #513 

- Update coagulation-flocculation models and properties
- Updated crystallization prop pack
- Updated ion prop pack
- Updated NaCl prop pack
- Updated seawater prop names
- Updated water prop names
- Updated NDMA prop names

### Legal Acknowledgement


watertap/property_models/NaCl_prop_pack.py
```diff
@@ -631,24 +631,26 @@ def rule_mole_frac_phase_comp(b, j):
             self.params.component_list, rule=rule_mole_frac_phase_comp
         )
 
-    def _molality_comp(self):
-        self.molality_comp = Var(
+    def _molality_phase_comp(self):
+        self.molality_phase_comp = Var(
             ["NaCl"],
```
bknueven, line 637: Here, and in lots of other places, we also need to update the indexing set (e.g., in this example it should be `[["Liq"],["NaCl"]]`.

Otherwise this will be the source of even further confusion.

adam-a-a: I agree. If we change the name to include the `phase` qualifier, then we need to include said index for the variable.

watertap/property_models/NaCl_prop_pack.py
```diff
@@ -668,17 +670,17 @@ def rule_visc_d_phase(b):  # dynamic viscosity, eq 5 in Bartholomew
 
         self.eq_visc_d_phase = Constraint(rule=rule_visc_d_phase)
 
-    def _diffus_phase(self):
-        self.diffus_phase = Var(
+    def _diffus_phase_comp(self):
+        self.diffus_phase_comp = Var(
             self.params.phase_list,
```
adam-a-a, line 676: SHould also be indexed by component, even though this is the NaCl package

lbibl (commented):
I agree with and appreciate all the changes.  One general comment is that while we are adding `_phase_` to var names where applicable, some of them were not actually indexed by `[p]` in the unit or test files, when it's assumed `'H2O'` being the only case. I think it'd be nice to have all vars that have `_phase` in their names indexed by `[p]` when being used, if that won't overly consume computationally.  This will potentially create a flexibility for say treating a water containing some other solvent.   But either will be fine and having  an agreement might be helpful.

watertap/property_models/NaCl_prop_pack.py
aladshaw3: Done

watertap/property_models/NaCl_prop_pack.py
aladshaw3: Done

watertap/unit_models/_reverse_osmosis_base.py
```diff
@@ -749,10 +752,13 @@ def eq_N_Sh(b, t, x):
...
             def eq_N_Sc(b, t, x):
                 bulk = b.feed_side.properties[t, x]
+                # # TODO: This needs to be revisted. Diffusion is now by component, but
+                #   not H2O and this var should also be by component, but the implementation
+                #   is not immediately clear.
                 return (
                     b.N_Sc[t, x]
                     * bulk.dens_mass_phase["Liq"]
-                    * bulk.diffus_phase["Liq"]
+                    * bulk.diffus_phase_comp["Liq", bulk.params.component_list.last()]
```
adam-a-a, line 761: Yes, this should be revised so that the constraint `eq_N_Sc` is also indexed by solute_set or ion_set as component `j`. I _think_ the NF DSPM-DE model should have this implemented correctly.

adam-a-a: However, since we've only ever used the RO model for either the NaCl or seawater prop packs, we never had to worry about more than one solute, and I think we even have a warning or possibly an exception indicating the model is only compatible with one solute. Thus, this change will suffice for now.

adam-a-a (commented):
LGTM - a comment concerning the Schmidt number in RO base, but I think that is beyond scope of this PR. The only reason I didn't approve yet is because I didn't see the nanofiltration DSPM-DE model (maybe I missed it) and suspect it might need to be resolved too once you merge main.

adam-a-a (approved):
Actually, I see you merged main recently, and after revisiting the DSPM-DE model, it appears everything there is already covered. Good to go! Just need to run Black for the check to pass.

lbibl (approved):

watertap/property_models/NDMA_prop_pack.py
```diff
@@ -542,24 +542,27 @@ def rule_mole_frac_phase_comp(b, j):
...
-                self.molality_comp[j]
+                self.molality_phase_comp["Liq", j]
                 == b.mass_frac_phase_comp["Liq", j]
                 / (1 - b.mass_frac_phase_comp["Liq", j])
                 / b.params.mw_comp[j]
             )
 
-        self.eq_molality_comp = Constraint(["NDMA"], rule=rule_molality_comp)
+        self.eq_molality_phase_comp = Constraint(
+            ["NDMA"], rule=rule_molality_phase_comp
```
bknueven, line 564: Ideally this constraint would also be indexed by phase

watertap/property_models/NaCl_prop_pack.py
```diff
@@ -631,24 +631,27 @@ def rule_mole_frac_phase_comp(b, j):
...
-                self.molality_comp[j]
+                self.molality_phase_comp["Liq", j]
                 == b.mass_frac_phase_comp["Liq", j]
                 / (1 - b.mass_frac_phase_comp["Liq", j])
                 / b.params.mw_comp[j]
             )
 
-        self.eq_molality_comp = Constraint(["NaCl"], rule=rule_molality_comp)
+        self.eq_molality_phase_comp = Constraint(
+            ["NaCl"], rule=rule_molality_phase_comp
```
bknueven, line 653: Similar here (missing phase index in constraint)

watertap/property_models/NaCl_prop_pack.py
```diff
@@ -688,7 +692,7 @@ def rule_diffus_phase(b):  # diffusivity, eq 6 in Bartholomew
                 + b.params.diffus_param["0"]
             )
 
-        self.eq_diffus_phase = Constraint(rule=rule_diffus_phase)
+        self.eq_diffus_phase_comp = Constraint(["NaCl"], rule=rule_diffus_phase_comp)
```
bknueven, line 695: Same here.

The rule above would also ideally not have `"Liq"` and `"NaCl"` hard-coded.

watertap/property_models/NaCl_prop_pack.py
```diff
@@ -708,27 +712,33 @@ def rule_osm_coeff(b):
...
             rhow = (
                 1000 * pyunits.kg / pyunits.m**3
             )  # TODO: could make this variable based on temperature
             return (
-                b.pressure_osm
+                b.pressure_osm_phase[p]
                 == i
                 * b.osm_coeff
-                * b.molality_comp["NaCl"]
+                * b.molality_phase_comp["Liq", "NaCl"]
```
bknueven, line 733:
```suggestion
                * sum(b.molality_phase_comp[p,:])
```

watertap/property_models/NDMA_prop_pack.py
```diff
@@ -542,24 +542,27 @@ def rule_mole_frac_phase_comp(b, j):
...
         )
 
-        def rule_molality_comp(b, j):
+        def rule_molality_phase_comp(b, j):
             return (
-                self.molality_comp[j]
+                self.molality_phase_comp["Liq", j]
                 == b.mass_frac_phase_comp["Liq", j]
                 / (1 - b.mass_frac_phase_comp["Liq", j])
                 / b.params.mw_comp[j]
```
bknueven, line 560: Should this parameter also be indexed by phase?

watertap/property_models/cryst_prop_pack.py
```diff
@@ -1404,30 +1409,30 @@ def rule_cp_solute(b, p):
...
-                b.cp_phase["Liq"]
-                == b.mass_frac_phase_comp["Liq", "NaCl"] * b.cp_solute["Liq"]
-                + b.mass_frac_phase_comp["Liq", "H2O"] * b.cp_solvent["Liq"]
+                b.cp_mass_phase["Liq"]
+                == b.mass_frac_phase_comp["Liq", "NaCl"] * b.cp_mass_solute["Liq"]
+                + b.mass_frac_phase_comp["Liq", "H2O"] * b.cp_mass_solvent["Liq"]
             )
 
-        self.eq_cp_phase = Constraint(rule=rule_cp_phase)
+        self.eq_cp_mass_phase = Constraint(rule=rule_cp_mass_phase)
```
bknueven, line 1435: Should be indexed by phase?

watertap/property_models/cryst_prop_pack.py
```diff
@@ -1713,7 +1724,9 @@ def rule_dh_crystallization(b):
                     f"Temperature-dependent heat of crystallization model has not been implemented yet."
                 )
 
-        self.eq_dh_crystallization = Constraint(rule=rule_dh_crystallization)
+        self.eq_dh_crystallization_mass_comp = Constraint(
```
bknueven, line 1727: Indexed by component?

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -943,25 +942,26 @@ def rule_mole_frac_phase_comp(b, j):
...
+                b.molality_phase_comp["Liq", j]
                 == b.flow_mol_phase_comp["Liq", j]
                 / b.flow_mol_phase_comp["Liq", "H2O"]
                 / b.params.mw_comp["H2O"]
             )
 
-        self.eq_molality_comp = Constraint(
-            self.params.ion_set | self.params.solute_set, rule=rule_molality_comp
+        self.eq_molality_phase_comp = Constraint(
+            self.params.ion_set | self.params.solute_set, rule=rule_molality_phase_comp
```
bknueven, line 964: Indexed by phase?

bknueven (commented):
It's a minor thing for sure, but now a lot of our constraints have the "wrong" index set, i.e., it does not match that of the property it is calculating.

watertap/property_models/NDMA_prop_pack.py
aladshaw3: No, the `mw_comp` is phase independent.

watertap/property_models/NaCl_prop_pack.py
aladshaw3: > It's a minor thing for sure, but now a lot of our constraints have the "wrong" index set, i.e., it does not match that of the property it is calculating.

I agree with your comments. It probably needs to change, but I wonder if that should be for a subsequent PR. The most important part is the names of the properties because that is how a user/developer will interact with the properties in there own modules/models.

aladshaw3:
> It's a minor thing for sure, but now a lot of our constraints have the "wrong" index set, i.e., it does not match that of the property it is calculating.

I agree with your comments. It probably needs to change, but I wonder if that should be for a subsequent PR. The most important part is the names of the properties because that is how a user/developer will interact with the properties in there own modules/models.

lbibl:
> > It's a minor thing for sure, but now a lot of our constraints have the "wrong" index set, i.e., it does not match that of the property it is calculating.
> 
> I agree with your comments. It probably needs to change, but I wonder if that should be for a subsequent PR. The most important part is the names of the properties because that is how a user/developer will interact with the properties in there own modules/models.

An idea of mine:  after we merge in this PR to first have all names updated and make an agreement on how to index everything for water solutions, we ask all relevant authors, who have written files relying on the changed properties, to review their own files to make update on indices, in one collaborated PR or separate PRs.

watertap/property_models/NaCl_prop_pack.py
adam-a-a: @bknueven I think it is OK since this is the NaCl property model.

bknueven:
> > It's a minor thing for sure, but now a lot of our constraints have the "wrong" index set, i.e., it does not match that of the property it is calculating.
> 
> I agree with your comments. It probably needs to change, but I wonder if that should be for a subsequent PR. The most important part is the names of the properties because that is how a user/developer will interact with the properties in there own modules/models.

I'm fine with this as-is, FWIW. If I wasn't I would have used the "request changes" feature.

watertap/property_models/NaCl_prop_pack.py
bknueven: To be clear, I didn't mean here, but in the rule for this constraint.

Again it's being nit-picky. Pyomo passes in the indices to the rule, so it's weird not to use them.

PR #566, opened 2022-06-13
Add crystallizer operating cost

OOAmusat:
## Summary/Motivation:

Add crystallizer operating cost

- Add crystallizer operating cost as sum of heating and pumping cost.

### Legal Acknowledgement


watertap/costing/watertap_costing_package.py
```diff
@@ -233,6 +235,37 @@ def build_global_params(self):
             units=pyo.units.dimensionless,
         )
 
+        # Crystallizer operating cost information from literature
+        self.crystallizer_steam_unit_cost = pyo.Var(
```
adam-a-a: Maybe this can just be steam_unit_cost, similar to how we just have a global electricity_cost

watertap/costing/watertap_costing_package.py
```diff
@@ -233,6 +235,37 @@ def build_global_params(self):
...
         )
 
+        # Crystallizer operating cost information from literature
+        self.crystallizer_steam_unit_cost = pyo.Var(
+            initialize=0.004,
+            units=pyo.units.USD_2018 / (pyo.units.meter**3),
+            doc="Steam cost, Panagopoulos (2019)",
+        )
+
+        self.crystallizer_electricity_unit_cost = pyo.Var(
```
adam-a-a: if we already have electricity_cost, perhaps we can consolidate this.

watertap/costing/watertap_costing_package.py
```diff
@@ -734,6 +805,75 @@ def cost_crystallizer_by_volume(blk):
             )
         )
 
+    @staticmethod
+    def compute_steam_properties(blk):
```
bknueven: Should this be private, *or* a helper function in this module?

I mostly think of the static methods on the costing package as being *costing* methods. But this is mostly about how we organize the code.

bknueven (approved):
I have a minor comment about code organization within the watertap costing module.

watertap/costing/watertap_costing_package.py
OOAmusat: It probably should be a private method (without the static method tag) - I don't expect it to be accessed externally. I'll change the change now. It probably could be a general helper function for steam costing though. 

Thanks.

watertap/unit_models/tests/test_crystallizer.py
```diff
@@ -101,6 +101,52 @@ def Crystallizer_frame(self):
...
+
+        # Define operating conditions
+        m.fs.unit.temperature_operating.fix(crystallizer_temperature)
+        m.fs.unit.crystallization_yield["NaCl"].fix(crystallizer_yield)
+
+        # Fix growth rate, crystal length and Sounders brown constant to default values
+        m.fs.unit.crystal_growth_rate.fix()
+        m.fs.unit.souders_brown_constant.fix()
+        m.fs.unit.crystal_median_length.fix()
+
```
adam-a-a, line 150: I would check that `assert_units_consistent` holds true, once after building the model, and another time after adding costing.

adam-a-a (changes requested):
LGTM - just minor suggested changes, mostly to add an `assert_units_consistent` before and after costing in the tests.

watertap/costing/watertap_costing_package.py
adam-a-a: That is, I don't think we should have 2 separate electricity costs. One global cost should do, and we'd apply that to this unit and others in a flowsheet.

OOAmusat: I already removed this and used the electricity cost already there.

watertap/costing/watertap_costing_package.py
OOAmusat: Done.

watertap/unit_models/tests/test_crystallizer.py
OOAmusat: Done

adam-a-a (approved):

PR #567, opened 2022-06-13
Dye desalination with Reverse Osmosis

arao53:
## Summary/Motivation:
Improves the completeness of the dye desalination case study by adding a zero order waste water treatment plant model for pretreatment and a 0-dimensional reverse osmosis desalination unit downstream of dye separation via nanofiltration. 

- dye_desalination_global_costing.yaml: adds a defined flow for RO permeate recovered to the textile plant. 
- dye_desalination_withRO.py: flowsheet analysis with zero order unit model for WWTP and 0d unit model for RO. 
- test: tests for dye_desalination_withRO.py

### Legal Acknowledgement


tarnold17 (approved):

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_global_costing.yaml
```diff
@@ -14,6 +14,9 @@ defined_flows:
   dye_mass:
     value: 0.1
     units: USD_2020/kg
+  recovered_water:
```
adam-a-a, line 17: Although I would opt for not accounting for the sale of recovered water as the default since it would be recycled to the plant (i.e., default value of 0), I think it is nice to have this as an option that we could adjust with parameter_sweep.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
```diff
@@ -0,0 +1,569 @@
...
+
+def build():
+    # flowsheet set up
+    m = ConcreteModel()
+    m.db = Database()
+
+    m.fs = FlowsheetBlock(default={"dynamic": False})
+
+    # define property packages
+    m.fs.prop_nf = prop_ZO.WaterParameterBlock(default={"solute_list": ["dye", "tds"]})
```
adam-a-a, line 91: Technically, since we are adding the secondary WWTP step, we could add `bod `and `tss` to the feed, and these would be removed by the WWTP. However, at this level, the only significance the addition of the WWTP step has is accounting for the extra treatment costs, so I suppose we could omit the addition of `bod `and `tss` in the feed. It might complicate things with regard to the translator block.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
```diff
@@ -0,0 +1,569 @@
...
+            "property_package": m.fs.prop_ro,
+            "has_pressure_change": True,
+            "pressure_change_type": PressureChangeType.calculated,
+            "mass_transfer_coefficient": MassTransferCoefficient.calculated,
+            "concentration_polarization_type": ConcentrationPolarizationType.calculated,
+        }
+    )
+
+    desal.RO.width.setub(2000)
+    desal.RO.area.setub(20000)
```
adam-a-a, line 146: do we know if these lines are necessary?

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
```diff
@@ -0,0 +1,569 @@
...
+    # desalination
+    desal.P2.efficiency_pump.fix(0.80)
+    operating_pressure = 70e5 * pyunits.Pa
+    desal.P2.control_volume.properties_out[0].pressure.fix(operating_pressure)
+    desal.RO.A_comp.fix(4.2e-12)  # membrane water permeability
+    desal.RO.B_comp.fix(3.5e-8)  # membrane salt permeability
+    desal.RO.channel_height.fix(1e-3)  # channel height in membrane stage [m]
+    desal.RO.spacer_porosity.fix(0.97)  # spacer porosity in membrane stage [-]
+    desal.RO.permeate.pressure[0].fix(pressure)  # atmospheric pressure [Pa]
+    # desal.RO.area.fix(50)
```
```suggestion
```

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
```diff
@@ -0,0 +1,568 @@
...
+    )
+    desal.RO.feed_side.properties_in[0].pressure = value(
+        desal.P2.control_volume.properties_out[0].pressure
+    )
+    solve(desal)
+    desal.RO.initialize()
+    return
+
+
+def solve(blk, solver=None, tee=False, check_termination=True):
```
adam-a-a, line 334: Technically, since we have the RO unit here, we could have a `setup_optimization` function which unfixes certain variables on the RO unit and then solve and find the cost-optimal decision variables (e.g., membrane area)

adam-a-a: (just food for thought rather than a suggested change)

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
```diff
@@ -0,0 +1,568 @@
...
+    dye_sep.P1.costing = UnitModelCostingBlock(
+        default={"flowsheet_costing_block": m.fs.zo_costing}
+    )
+
+    # RO Train
+    # RO equipment is costed using more detailed costing package
+    desal.P2.costing = UnitModelCostingBlock(
+        default={
+            "flowsheet_costing_block": m.fs.ro_costing,
+            "costing_method_arguments": {"cost_electricity_flow": False},
```
adam-a-a: Just curious- why set `cost_electricity_flow` to False instead of True?

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
```diff
@@ -0,0 +1,568 @@
...
+        ),
+        doc="Savings from water recovered back to the plant",
+    )
+
+    # Combine results from costing packages and calculate overall metrics
+    @m.Expression()
+    def total_capital_cost(b):
+        return (
+            pyunits.convert(
+                m.fs.zo_costing.total_capital_cost, to_units=pyunits.USD_2018
```
adam-a-a: Why convert to 2018 instead of keeping at 2020? In that case, I suppose you'd convert `ro_costing` to 2020. I may be wrong, but I thought we were working with USD_2020 for the AMO cases.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
```diff
@@ -0,0 +1,568 @@
...
+            )
+            + m.fs.ro_costing.total_operating_cost
+        )
+
+    @m.Expression()
+    def total_externalities(b):
+        return pyunits.convert(
+            m.fs.water_recovery_revenue
+            + m.fs.dye_recovery_revenue
+            - m.fs.brine_disposal_cost,
```
adam-a-a: I guess adding the WWTP opened a small can of worms, but maybe we should account for sludge disposal too.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
```diff
@@ -0,0 +1,568 @@
...
+            to_units=pyunits.USD_2018 / pyunits.year,
+        )
+
+    @m.Expression()
+    def LCOT(b):
```
adam-a-a: Can you add doc strings for these expressions?

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
arao53: Will the sludge disposal cost be the same/similar to the waste disposal cost that is applied to the brine term? If so, maybe we could sum the wasted flowrates and then multiply by the cost.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
arao53: I believe some of the other examples (metab) were in 2018 so I went with that, but I can easily change to 2020.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
```suggestion
    def LCOT(b, doc="Levelized cost of treatment with respect to volumetric feed flow"):
```

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
```diff
@@ -0,0 +1,568 @@
...
+    return
+
+
+def display_costing(m):
+    # print("\n Unit costing metrics:")
+    # prtrt_capex = m.fs.zo_costing.
+    #
+    #     m.fs.zo_costing.total_variable_operating_cost,
+    #             to_units=pyunits.USD_2018 / pyunits.year,
+    #         )
```
```suggestion

```

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_sweep.py
```diff
@@ -91,11 +91,16 @@ def run_analysis(case_num, nx, interpolate_nan_outputs=True):
...
+        os.path.dirname(os.path.abspath(__file__)),
+        output_filename,
+    )
+
     global_results = parameter_sweep(
         m,
         sweep_params,
         outputs,
-        csv_results_file_name=output_filename,
+        csv_results_file_name=output_path,
```
adam-a-a, line 103: Can this sweep file be applied to the flowsheet w/WWTP and RO? Also, are the high-level metric the same between the 2 flowsheets if we wanted run comparisons between them?

adam-a-a (changes requested):
Great job on this despite my string of comments! There are just a couple of small changes I'm suggesting, while the majority of the comments are just for me to get a better understanding and/or throw ideas out for _potential_ modifications/additions in a subsequent PR.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
adam-a-a: I think that sounds good. I think it's a fair assumption to just use one waste disposal cost. We can always segregate these later if we have reason to do so.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_sweep.py
arao53: I don't think the sweep file can do both, as of now. My plan was to make another sweep file for RO but I can integrate them instead. This specific addition was slipped in from a previous comment and ensures that the sweep files are saved in the same directory as the flowsheet. I'll need to adjust the costing for dye in the previous analysis slightly, but we should be able to compare the levelized cost on a treated feed water basis.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
adam-a-a: Ah ok- that's what I was failing to recall. In that case, it might make sense to use USD_2018 in the flowsheet's global param YAML file

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
```diff
@@ -0,0 +1,562 @@
...
+            - m.fs.brine_disposal_cost,
+            to_units=pyunits.USD_2018 / pyunits.year,
+        )
+
+    @m.Expression()
+    def LCOT(b, doc="Levelized cost of treatment with respect to volumetric feed flow"):
+        return (
+            b.total_capital_cost * b.fs.zo_costing.capital_recovery_factor
+            + b.total_operating_cost
+            - b.total_externalities
```
adam-a-a, line 492: I think this makes sense now (regarding the value of recovered water), if we are considering the LCOT w/respect to the manufacturing plant (i.e., this account for freshwater savings.)

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_global_costing.yaml
arao53: Reset to 0.0000001 since setting the flow cost to 0 caused unit errors.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_sweep.py
adam-a-a: OK- we can possibly get around to having one sweep file that could be applicable to both flowsheets (e.g., by making sure cost metrics are symmetrical between flowsheets) in a subsequent PR.

adam-a-a (approved):
LGTM! @arao53 could you add another issue noting the subtle mods discussed here for future PRs so we don't forget when the time comes?

Issue #489, opened 2022-04-14
Electrodialysis Model Development List - June


aladshaw3:
To Do List for Electrodialysis Unit Model:
--------------------------------------------
- [x] Complete first iteration of 0D electrodialysis model [**April**] (#523)
- [x] Make necessary changes in ion properties for electrodialysis [**April**] 
     - (issue #464)
     - (PR #501)
     - (PR #508)
- [x] Create documentation for 0D electrodialysis [**June**] (#537)
- [x] Complete first iteration of 1D electrodialysis model [**June**] (#571 )
- [x] Create documentation for 1D electrodialysis [**June**] (#579)
- [x] **Costing** for analysis in flowsheets (#582)

lbibl:
I updated with some proposed 0d refinements .

PR #571, opened 2022-06-15
Electrodialysis_1D unit model 

lbibl:
## Fixes/Resolves:

#489  

This PR presents a complete version of electrodialysis_1D unit model.   

-A full electrodialysis_1D model is presented using the same theoretical framework of the current electrodialysis_0D model. 
-A test file for electrodialysis_1D model is created. 
-Minor changes and amendment were made to the electrodialysis_0D unit model, its testing file and documentation. 
-A minor change to the ion_DSPMDE_prop_pack file was made. This should not cause any issue to other models using this property package. 

### Legal Acknowledgement


watertap/unit_models/electrodialysis_1D.py
```diff
@@ -0,0 +1,1308 @@
...
+            balance_type=self.config.material_balance_type, has_mass_transfer=True
+        )
+        # # TODO: Temporarily disabling energy balances
+        if hasattr(self.config, "energy_balance_type"):
+            self.diluate.add_energy_balances(
+                balance_type=self.config.energy_balance_type,
+                has_enthalpy_transfer=False,
+            )
+        self.diluate.add_momentum_balances(
+            balance_type=self.config.momentum_balance_type, has_pressure_change=False
```
```suggestion
            balance_type=self.config.momentum_balance_type, has_pressure_change=self.config.has_pressure_change
```

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -0,0 +1,1308 @@
...
+            initialize=1,
+            bounds=(0, 12100),
+            domain=NonNegativeReals,
+            units=pyunits.watt,
+            doc="Electrical power consumption of a stack",
+        )
+        self.diluate.Dpower_electrical_Dx = DerivativeVar(
+            self.diluate.power_electrical_x,
+            wrt=self.diluate.length_domain,
+            units=pyunits.watt,
```
adam-a-a, line 322: Should the units be per length?

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -0,0 +1,1308 @@
...
+            raise ConfigurationError(
+                "Property Package MUST constain 'H2O' as a component"
+            )
+
+        # Create unit model parameters and vars
+        self.water_density = Param(
+            initialize=1000,
+            mutable=False,
+            units=pyunits.kg * pyunits.m**-3,
+            doc="density of water",
```
adam-a-a: This is already a param or var in the ion prop package, so this isn't needed here. While the RO model does include, this is something that should eventually be removed from the RO model.

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -0,0 +1,1308 @@
...
+        self.cell_width = Var(
+            initialize=0.1,
+            bounds=(1e-3, 1e2),
+            units=pyunits.meter,
+            doc="The width of the electrodialysis cell, denoted as b in the model description",
+        )
+        self.spacer_thickness = Var(
+            initialize=0.0001,
+            units=pyunits.meter,
+            doc="The distance between the concecutive aem and cem",
```
```suggestion
            doc="The distance between the consecutive aem and cem",
```

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -0,0 +1,1308 @@
...
+            units=pyunits.amp,
+            doc="Current across a cell-pair or stack",
+        )
+        self.current_x = Var(
+            self.flowsheet().time,
+            self.diluate.length_domain,
+            initialize=1,
+            bounds=(0, 1000),
+            units=pyunits.amp,
+            doc="Current across a cell-pair or stack",
```
adam-a-a: Same doc string as `current_applied`. Should this be something like "Current across cell-pair or stack, as a function of length"?

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -0,0 +1,1308 @@
...
+        )
+        self.current_density_x = Var(
+            self.flowsheet().time,
+            self.diluate.length_domain,
+            initialize=1,
+            bounds=(0, 1000),
+            units=pyunits.amp * pyunits.meter**-2,
+            doc="Current across a cell-pair or stack",
```
adam-a-a: Update doc string

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -0,0 +1,1308 @@
...
+            units=pyunits.volt,
+            doc="Voltage across a stack, declared under the 'Constant Voltage' mode only",
+        )
+        self.voltage_x = Var(
+            self.flowsheet().time,
+            self.diluate.length_domain,
+            initialize=100,
+            bounds=(0, 1000),
+            units=pyunits.volt,
+            doc="Voltage across a stack, declared under the 'Constant Voltage' mode only",
```
adam-a-a, line 496: Update doc string

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -0,0 +1,1308 @@
...
+                + self.electrodes_resistance
+            )
+
+        @self.Constraint(
+            self.flowsheet().time,
+            self.diluate.length_domain,
+            doc="calcualte current density from the electrical input",
+        )
+        def eq_get_current_density(self, t, x):
+            if self.config.operation_mode == "Constant_Current":
```
adam-a-a, line 593: This works although we usually add an enumerated class to account for the options.

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -0,0 +1,1308 @@
...
+
+        # Mass Transfer for the Concentrate
+        @self.Constraint(
+            self.flowsheet().config.time,
+            self.concentrate.length_domain,
+            self.config.property_package.phase_list,
+            self.config.property_package.component_list,
+            doc="Mass transfer term concentrate channel",
+        )
+        def eq_mass_transfer_term_concentrate(self, t, x, p, j):
```
adam-a-a, line 686: Can we cut down these lines of code by just setting the mass transfer term of the concentrate equal to the negative mass transfer term of diluate?

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -0,0 +1,1308 @@
...
+        blk,
+        state_args=None,
+        outlvl=idaeslog.NOTSET,
+        solver=None,
+        optarg=None,
+        fail_on_warning=False,
+        ignore_dof=False,
+    ):
+        """
+        General wrapper for pressure changer initialization routines
```
```suggestion
        General wrapper for electrodialysis initialization routines
```

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -0,0 +1,1308 @@
...
+                if (
+                    "flow_mol_phase_comp"
+                    not in blk[k].diluate.properties[set].define_state_vars()
+                ):
+                    raise ConfigurationError(
+                        "Electrodialysis1D unit model requires "
+                        "either a 'flow_mol_phase_comp' or 'flow_mass_phase_comp' "
+                        "state variable basis to apply the 'propogate_initial_state' method"
+                    )
+                if "teperature" in blk[k].diluate.properties[set].define_state_vars():
```
```suggestion
                if "temperature" in blk[k].diluate.properties[set].define_state_vars():
```

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -0,0 +1,1308 @@
...
+                    "flow_mol_phase_comp"
+                    not in blk[k].concentrate.properties[set].define_state_vars()
+                ):
+                    raise ConfigurationError(
+                        "Electrodialysis1D unit model requires "
+                        "either a 'flow_mol_phase_comp' or 'flow_mass_phase_comp' "
+                        "state variable basis to apply the 'propogate_initial_state' method"
+                    )
+                if (
+                    "teperature"
```
```suggestion
                    "temperature"
```

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -0,0 +1,1308 @@
...
+        check_solve(
+            res,
+            logger=init_log,
+            fail_flag=fail_on_warning,
+            checkpoint="Initialization Step 3",
+        )
+        # ---------------------------------------------------------------------
+        # Release state
+        blk.diluate.release_state(flags_diluate, outlvl)
+        init_log.info("Initialization Complete: {}".format(idaeslog.condition(res)))
```
```suggestion
```

watertap/unit_models/tests/test_electrodialysis_0D.py
```diff
@@ -224,22 +228,22 @@ def test_solution(self, electrodialysis_cell1):
...
-        ) == pytest.approx(2.51e-2, rel=5e-2)
+        ) == pytest.approx(2.47e-2, rel=5e-2)
         assert value(
             m.fs.unit.outlet_concentrate.flow_mol_phase_comp[0, "Liq", "Na_+"]
-        ) == pytest.approx(1.417e-4, rel=5e-3)
+        ) == pytest.approx(1.19e-4, rel=5e-2)
         assert value(
             m.fs.unit.outlet_concentrate.flow_mol_phase_comp[0, "Liq", "Cl_-"]
-        ) == pytest.approx(1.417e-4, rel=5e-3)
+        ) == pytest.approx(1.19e-4, rel=5e-2)
```
aladshaw3: Your relative tolerances seem too relaxed to me. At a minimum, the tolerances should be 1e-4. These tests are designed to make sure that we get the same expected output from the model regardless of updates made to the solver or OS or other minor changes in how the model is solved. If we cannot get these tolerances, then we may need to revisit some scaling.

watertap/unit_models/tests/test_electrodialysis_0D.py
```diff
@@ -455,10 +463,10 @@ def test_performance_contents(self, electrodialysis_cell2):
...
             perform_dict["vars"]["Electrical power consumption(Watt)"]
-        ) == pytest.approx(3.5, rel=5e-1)
+        ) == pytest.approx(5.4, rel=5e-1)
         assert value(
             perform_dict["vars"]["Specific electrical power consumption (kWh/m**3)"]
-        ) == pytest.approx(2.33, rel=5e-2)
+        ) == pytest.approx(0.36, rel=5e-2)
         assert value(
             perform_dict["vars"]["Current efficiency for deionzation"]
         ) == pytest.approx(0.71, rel=5e-2)
```
aladshaw3: Relative tolerances are too relaxed.

watertap/unit_models/tests/test_electrodialysis_0D.py
```diff
@@ -675,10 +687,10 @@ def test_performance_contents(self, electrodialysis_cell3):
...
             perform_dict["vars"]["Electrical power consumption(Watt)"]
-        ) == pytest.approx(3.5, rel=5e-1)
+        ) == pytest.approx(5.3, rel=5e-1)
         assert value(
             perform_dict["vars"]["Specific electrical power consumption (kWh/m**3)"]
-        ) == pytest.approx(2.31, rel=5e-2)
+        ) == pytest.approx(0.35, rel=5e-2)
         assert value(
             perform_dict["vars"]["Current efficiency for deionzation"]
         ) == pytest.approx(0.71, rel=5e-2)
```
aladshaw3: Relative tolerances are too relaxed.

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -0,0 +1,1308 @@
...
+                        "either a 'flow_mol_phase_comp' or 'flow_mass_phase_comp' "
+                        "state variable basis to apply the 'propogate_initial_state' method"
+                    )
+                if "teperature" in blk[k].diluate.properties[set].define_state_vars():
+                    blk[k].diluate.properties[set].temperature = value(
+                        blk[k].diluate.properties[(0.0, 0.0)].temperature
+                    )
+                if "pressure" in blk[k].diluate.properties[set].define_state_vars():
+                    blk[k].diluate.properties[set].pressure = value(
+                        blk[k].diluate.properties[(0.0, 0.0)].pressure
```
aladshaw3, line 797: Will the first time point and first x point always be (0, 0)? If there is ever an instance where this isn't the case, then we maybe want to extract the first time and x point from their respective sets.

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -0,0 +1,1308 @@
...
+                        blk[k].concentrate.properties[set].flow_mass_phase_comp[
+                            ind
+                        ] = value(
+                            blk[k]
+                            .concentrate.properties[(0.0, 0.0)]
+                            .flow_mass_phase_comp[ind]
+                        )
+
+        # ---------------------------------------------------------------------
+        # Initialize diluate block
```
aladshaw3, line 869: So, it appears you are not initializing any of the other properties (such as concentrations, mole fractions, etc). Is this going to be good enough for initialization of this unit? How difficult is the model to solve? Should we add in logic here to initialize the other properties that get used?

watertap/unit_models/tests/test_electrodialysis_1D.py
```diff
@@ -0,0 +1,691 @@
...
+            "flow_mol_phase_comp", 1e4, index=("Liq", "Na_+")
+        )
+        m.fs.properties.set_default_scaling(
+            "flow_mol_phase_comp", 1e4, index=("Liq", "Cl_-")
+        )
+        # set scaling factors for some vars
+        iscale.set_scaling_factor(m.fs.unit.cell_width, 10)
+        iscale.set_scaling_factor(m.fs.unit.cell_length, 10)
+        iscale.calculate_scaling_factors(m.fs)
+        initialization_tester(m)
```
aladshaw3, line 206: How many iterations does the initialization take? Do you see any regularization during initialization?

watertap/unit_models/tests/test_electrodialysis_1D.py
```diff
@@ -0,0 +1,691 @@
...
+        }
+        assert not badly_scaled_var_values
+        # check to make sure DOF does not change
+        assert degrees_of_freedom(m) == 0
+
+    @pytest.mark.component
+    def test_solve(self, electrodialysis_1d_cell1):
+        m = electrodialysis_1d_cell1
+        # run solver and check for optimal solution
+        results = solver.solve(m)
```
aladshaw3, line 218: How many iterations does the solve take? Is convergence smooth?

watertap/unit_models/tests/test_electrodialysis_1D.py
```diff
@@ -0,0 +1,691 @@
...
+    def test_performance_contents(self, electrodialysis_1d_cell1):
+        m = electrodialysis_1d_cell1
+        perform_dict = m.fs.unit._get_performance_contents()
+        assert "vars" in perform_dict
+        assert value(
+            perform_dict["vars"]["Total electrical power consumption(Watt)"]
+        ) == pytest.approx(3.0, rel=5e-1)
+        assert value(
+            perform_dict["vars"]["Specific electrical power consumption (kW*h/m**3)"]
+        ) == pytest.approx(0.20, rel=5e-2)
```
aladshaw3: Tolerances are too relaxed.

watertap/unit_models/tests/test_electrodialysis_1D.py
```diff
@@ -0,0 +1,691 @@
...
+    def test_performance_contents(self, electrodialysis_1d_cell2):
+        m = electrodialysis_1d_cell2
+        perform_dict = m.fs.unit._get_performance_contents()
+        assert "vars" in perform_dict
+        assert value(
+            perform_dict["vars"]["Total electrical power consumption(Watt)"]
+        ) == pytest.approx(5.8, rel=5e-1)
+        assert value(
+            perform_dict["vars"]["Specific electrical power consumption (kW*h/m**3)"]
+        ) == pytest.approx(0.39, rel=5e-2)
```
aladshaw3: Tolerances are too relaxed.

watertap/unit_models/tests/test_electrodialysis_1D.py
```diff
@@ -0,0 +1,691 @@
...
+        ) == pytest.approx(2.50e-2, rel=5e-2)
+        assert value(
+            m.fs.unit.outlet_concentrate.flow_mol_phase_comp[0, "Liq", "Na_+"]
+        ) == pytest.approx(1.30e-4, rel=5e-2)
+        assert value(
+            m.fs.unit.outlet_concentrate.flow_mol_phase_comp[0, "Liq", "Cl_-"]
+        ) == pytest.approx(1.30e-4, rel=5e-2)
+        assert value(
+            m.fs.unit.outlet_concentrate.flow_mol_phase_comp[0, "Liq", "N"]
+        ) == pytest.approx(7.50e-06, rel=5e-2)
```
aladshaw3: Tolerances are too relaxed.

aladshaw3 (changes requested):
Looking good. I have a few questions and concerns about the convergence behavior since we do not appear to be initializing non-stateful properties (concentrations, mole fractions, etc). 

Also, you need to tighten the tolerances for the unit tests. 

You will also need to resolve conflicts caused by the updating of the IDAES import statements from another PR.

watertap/unit_models/tests/test_electrodialysis_0D.py
lbibl: Re: all scaling tolerance.  I chose all tolerance in accordance with the significant figures of that var.  and the significant figures are determined by the actual physical meaning (extra significant figures are physically meaningless).  I think scaling majorly matters for iterating times of the solver and avoiding local minima by the solver.   The tolerance is what levels of error we actually can tolerate by the physical meaning of the problem.  The solver's tolerance is by itself pretty tight (1e-8 I think) is not what we wrote here.    But the tolerance we care about here is determined by the problem itself.

watertap/unit_models/electrodialysis_1D.py
lbibl: It appears to be pretty good.  I expected and checked that setting up the state var initial values also set up all build-on-demand vars by the property packages.   Initialization only takes 7 iterations and 0.03 s.  After that solving the model takes 5 iterations and 0.023 s.  As long as the electrical input is is not exceeding the limit, model is solved smoothly and stably.

watertap/unit_models/tests/test_electrodialysis_1D.py
lbibl: See the previous comment.

watertap/unit_models/electrodialysis_1D.py
lbibl: For time, we have used t=0 everywhere to represent the steady state so I think it's ok to use 0 here (essentially there is no time index in our models so far) .  For length, the same is true here because the 1d control volume takes a constant continues set as (0, 1) when it was build in the current version.  So writing it as `...first()` in lieu of 0 will not make a difference if we do not give the flexibility to the control volume block's length set.    But ok to change.

watertap/unit_models/tests/test_electrodialysis_0D.py
aladshaw3: I think you don't understand in this case what this relative tolerance is in the `pytest.approx` function. This does not represent the significant figures of the result (although they may be somewhat related). This function is checking to make sure that the value you indicate is within `abs( value - calculated) / value`.

So, when you put in a relative tolerance of 5e-1, you are allowing the solution to vary by up to 50% of the expected result and still pass the test. **This is unacceptable**. Generally, most unit tests in WaterTAP use at least 1e-4 as a relative tolerance. Maybe a 1e-3 relative tolerance is acceptable, but no more than that.

adam-a-a (changes requested):
Just from a first pass -- some minor suggestions so far. Nice job!

watertap/unit_models/tests/test_electrodialysis_0D.py
lbibl: Tol were all set to 5e-3, which allow 0.5% error.  Further tightening will lead to adding extra significant figures.

watertap/unit_models/electrodialysis_1D.py
lbibl: No, the unit isn't per length (otherwise the code won't pass).   Basically, because the `length_domain` is a normalized length, it is dimensionless.

watertap/unit_models/electrodialysis_1D.py
lbibl: Suggestion accepted.

watertap/unit_models/electrodialysis_1D.py
lbibl: Accepted.

watertap/unit_models/electrodialysis_1D.py
lbibl: Good pointed and changed accordingly!

watertap/unit_models/electrodialysis_1D.py
lbibl: Why should this be deleted? Other models seem to have this kept in the release state routine. Also I have this for both control volumes. Can someone clarify?

aladshaw3 (approved):
LGTM. Just need to resolve the merge conflicts. If you need help dealing with this, just let me know.

watertap/unit_models/electrodialysis_1D.py
aladshaw3: This line is redundant. The initialization is complete at line 905, so we don't need to repeat this message here.

lbibl:
The conflict message seems gone after merge wt org main through my fork via command lines.  I did this when I opened the PR and manually addressed the conflicts caused by renaming files.  This might have been overridden by commit [fd9b842](https://github.com/watertap-org/watertap/pull/571/commits/fd9b842983560f3ac3d7f64d32cd3e3c6b65c749) made by @aladshaw3 on the github.   I recommend no more merge commits on github into this PR.  The current version should be ready to go with one more approval.

adam-a-a (approved):
Excellent job, @lbibl!

PR #572, opened 2022-06-15
Add partial support for parallel unit costing

bknueven:
## Summary/Motivation:
WaterTAP3 allowed multiple parallel units to be costed in a simple fashion, without creating multiple parallel unit models. This PR would begin to replicate that functionality in the zero-order costing package.

- Add `number_of_parallel_units` keyword argument to all zero order costing methods utilizing the `_general_power_law_form`.
- Implement parallel unit costing in `_general_power_law_form`.

### Legal Acknowledgement

### DIFF
--- a/watertap/core/zero_order_costing.py
+++ b/watertap/core/zero_order_costing.py
@@ -358,7 +358,7 @@ class ZeroOrderCostingData(FlowsheetCostingBlockData):
 
     # -------------------------------------------------------------------------
     # Unit operation costing methods
-    def cost_power_law_flow(blk):
+    def cost_power_law_flow(blk, number_of_parallel_units=1):
         """
         General method for costing equipment based on power law form. This is
         the most common costing form for zero-order models.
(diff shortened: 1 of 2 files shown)


adam-a-a (approved):
LGTM!

watertap/core/zero_order_costing.py
```diff
@@ -2569,10 +2633,17 @@ def _general_power_law_form(blk, A, B, sizing_term, factor=None):
         )
 
         expr = pyo.units.convert(
-            A * pyo.units.convert(sizing_term, to_units=pyo.units.dimensionless) ** B,
+            A
```
kurbansitterley, line 2636: This is different than the way we did it in WT3.
In WT3 for models with multiple units, the cost is essentially:

``C = (A * X ** b) * num_units``

Where `X` is calculated with the full flow entering the unit. I am not sure if the idea in WT3 was that there would be one unit offline (i.e. not in parallel), so the flow would be flowing through one unit at a time...

But appears this is the proper calculation for having parallel units. Because deviates from how it was implemented in WT3, the validation will still be off, but that is OK.

kurbansitterley (approved):
Slightly different implementation than WT3, but I think it is better and makes more sense. LGTM

PR #573, opened 2022-06-16
Remove IDAES v2 deprecation messages and doctest failures

### LINKED ISSUE #543: Address doctest failures due to IDAES v2 deprecations
- Changes recently introduced in IDAES cause some of the doctests to fail
- In order to merge #520, we're temporarily disabling the doctest check in the GHA CI workflow

## TODO

- [ ] Re-enable doctest check ASAP to avoid false negatives in the documentation code snippets
- [ ] Get rid of the deprecation messages by updating the import paths throughout the WaterTAP codebase

##### Use `.. doctest::` instead of `.. testcode::` for doctests that do not require testing the output

- Most if not all of the doctests are specified using the `.. testcode::` directive
- As I understand, this is supposed to be used in combination with a `.. testoutput::` block containing the expected output to be compared with the output of the code block (see the [testcode section in the sphinx-doctest documentation](https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html#directive-testcode) for more info)
- If no `.. testoutput::` is present, the code block is assumed to produce no output, which might or might not be what we want
- If we don't care about the output, we should consider converting all "lonely" `.. testcode::` directives to use `.. doctest::` instead

##### `stdout` vs `stderr` for IDAES deprecation messages

- I haven't realized this until now, but apparently the IDAES deprecation messages are emitted on `stdout`
- I'm not sure if there was a motivation behind this choice, but it could be argued that the IDAES deprecation messages should be emitted to `stderr` instead
- In any case, this is something that should be addressed at the IDAES level rather than by WaterTAP

lbianchi-lbl:
## Summary/Motivation:

- Get rid of the deprecation messages introduced by change in import paths in IDAES

- In the Python codebase:
  - Search and replace import paths affected by the IDAES changes with the current, non-deprecated version
- In the Sphinx docs:
  - Update the import paths in the code snippets
  - Convert code blocks using `.. testcode::` to `.. doctest::` if they didn't have an accompanying `.. testsetup::` and/or `.. testoutput::` (about half of them)
- In the GitHub Actions workflows:
  - Re-enable Sphinx doctests step (`make doctests`)
    - It had been disabled in response to the (false-positive) failures caused by the deprecation messages

### Legal Acknowledgement


bknueven (approved):

aladshaw3 (approved):
LGTM

PR #576, opened 2022-06-17
AMO 1595 initial flowsheet

tarnold17:
## Fixes/Resolves:

Add flowsheet and sweep for AMO 1595 Chen

-
-

### Legal Acknowledgement


watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1595_photothermal_membrane_candoP/amo_1595.py
```diff
@@ -0,0 +1,330 @@
...
+
+    electricity_intensity = value(
+        pyunits.convert(
+            m.fs.costing.electricity_intensity, to_units=pyunits.kWh / pyunits.m**3
+        )
+    )
+    print(f"Electricity Intensity: {electricity_intensity:.3f} kWh/m^3")
+
+    LCOW = value(
+        pyunits.convert(m.fs.costing.LCOW, to_units=pyunits.USD_2020 / pyunits.m**3)
```
arao53, line 347: What is the flow basis for this LCOW metric?

watertap/unit_models/zero_order/tests/test_CANDOP_zo.py
```diff
@@ -176,12 +173,15 @@ def test_solution(self, model):
...
             < 1e-6
         )
-        assert pytest.approx(500, rel=1e-5) == value(
-            model.fs.unit.properties_byproduct[0].conc_mass_comp[
-                "bioconcentrated_phosphorous"
-            ]
+        assert (
+            value(
+                model.fs.unit.properties_byproduct[0].conc_mass_comp[
+                    "bioconcentrated_phosphorous"
```
arao53, line 179: Is there a reason why pytest.approx is not used in this check?

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1595_photothermal_membrane_candoP/amo_1595.py
tarnold17: Treated water leaving the photothermal membrane unit. This is considered a byproduct, for this process.

watertap/unit_models/zero_order/tests/test_CANDOP_zo.py
tarnold17: I am just trying to make sure that the BCP in the byproduct stream is ~0. I used a < constraint just to make sure that the BCP in that stream is very small, and `pytest.approx` is only meant to be used for equality constraints..

arao53 (approved):

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1595_photothermal_membrane_candoP/amo_1595.py
```diff
@@ -0,0 +1,330 @@
...
+        default={
+            "property_package": m.fs.prop,
+            "database": m.db,
+        },
+    )
+
+    # product streams
+    m.fs.photothermal_water = Product(default={"property_package": m.fs.prop})
+    m.fs.candop_treated = Product(default={"property_package": m.fs.prop})
+    m.fs.candop_byproduct = Product(default={"property_package": m.fs.prop})
```
luohezhiming, line 123: What are the actual products of `candop_treated` and `candop_byproduct`? I am not sure but if there are no requirements or limitations on the `candop_byproduct` term, you probably don't need to add this.

luohezhiming (commented):
Looks good to me. Only have a comment on the byproduct of the CANDOP because I don't see you use it in the costing or other performance metrics.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1595_photothermal_membrane_candoP/amo_1595.py
tarnold17: The CANDO+P byproduct stream is just N2O. The treated stream is everything else - H2O, bioconcentrated phosphorous (BCP), nitrogen, and phosphates.

tarnold17: You can see in this flow diagram that the N2O and BCP streams are supposed to be separate:

image

tarnold17: @adam-a-a suggested that I could consider the price of N2O as one of the costing parameters, and factor the dollars recouped by selling that product into the levelized cost calculations. But I opted not to do that, because as you can see in the flow diagram, the N2O is just passed to the combustion block to generate electricity.

luohezhiming: Yeah, that makes sense. The N2O can pass to the combustion for energy recovery.

luohezhiming (approved):

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1595_photothermal_membrane_candoP/amo_1595.py
adam-a-a: @tarnold17 - I think I was suggesting to account for the N2O by considering some equivalent amount of electricity generated, i.e., using electricity production as a proxy, or assigning some price to N2O that translates to that equivalent.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1595_photothermal_membrane_candoP/amo_1595.py
```diff
@@ -0,0 +1,330 @@
...
+            - m.fs.costing.value_bcp_recovery
+        )
+        / (
+            pyunits.convert(
+                m.fs.feed.properties[0].flow_vol,
+                to_units=pyunits.m**3 / m.fs.costing.base_period,
+            )
+            * m.fs.costing.utilization_factor
+        ),
+        doc="Levelized Cost of Treatment (accounts for salable BCP)",
```
adam-a-a:
```suggestion
        doc="Levelized Cost of Treatment with respect to influent flow rate (accounts for salable BCP)",
```

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1595_photothermal_membrane_candoP/amo_1595.py
```diff
@@ -0,0 +1,330 @@
...
+            * m.fs.costing.utilization_factor
+        ),
+        doc="Levelized Cost of Water when accounting for salable BCP",
+    )
+
+    m.fs.costing.LCOT = Expression(
+        expr=(
+            m.fs.costing.total_capital_cost * m.fs.costing.capital_recovery_factor
+            + m.fs.costing.total_operating_cost
+            - m.fs.costing.value_bcp_recovery
```
adam-a-a, line 291: @tarnold17 maybe we should consider some water selling price as well to account for recovered water

adam-a-a: And we'd just include that in the LCOT metric and not the LCOW_bcp metric. This would set things up for a sensitivity analysis where we could assume no sale of water vs. some sale of water across a range of values.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1595_photothermal_membrane_candoP/amo_1595.py
```diff
@@ -0,0 +1,330 @@
...
+        )
+    )
+    print(f"Total Operating Costs: {total_operating_cost:.3f} M$/year")
+
+    electricity_intensity = value(
+        pyunits.convert(
+            m.fs.costing.electricity_intensity, to_units=pyunits.kWh / pyunits.m**3
+        )
+    )
+    print(f"Electricity Intensity: {electricity_intensity:.3f} kWh/m^3")
```
adam-a-a:
```suggestion
    print(f"Electricity Intensity with respect to influent flow rate: {electricity_intensity:.3f} kWh/m^3")
```

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1595_photothermal_membrane_candoP/amo_1595.py
```diff
@@ -0,0 +1,330 @@
...
+        )
+    )
+    print(
+        f"Levelized Cost of Water (accounting for BCP sale): {LCOW_with_bcp:.3f} $/m^3"
+    )
+
+    LCOT = value(
+        pyunits.convert(m.fs.costing.LCOT, to_units=pyunits.USD_2020 / pyunits.m**3)
+    )
+    print(f"Levelized Cost of Treatment: {LCOW:.3f} $/m^3")
```
adam-a-a:
```suggestion
    print(f"Levelized Cost of Treatment with respect to influent flow rate: {LCOW:.3f} $/m^3")
```

adam-a-a (changes requested):
Nice job on this! Just a couple of suggestions for you to review.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1595_photothermal_membrane_candoP/amo_1595.py
tarnold17: @adam-a-a That seems reasonable. Do you want to suggest a reasonable default price for this water?

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1595_photothermal_membrane_candoP/amo_1595.py
tarnold17: @adam-a-a I'm not sure it makes sense to consider an equivalent amount of electricity for the N2O, since N2O isn't directly flammable. From [wiki](https://en.wikipedia.org/wiki/Nitrous_oxide#:~:text=Internal%20combustion%20engine,-Main%20article%3A%20Nitrous&text=In%20vehicle%20racing%2C%20nitrous%20oxide,to%20produce%20more%20engine%20power.), the way it works in engines is that the N2O "allows the engine to burn more fuel by providing more oxygen during combustion. The increase in oxygen allows for an increase in the injection of fuel, allowing the engine to produce more engine power. The gas is not flammable at a low pressure/temperature, but it delivers more oxygen than atmospheric air by breaking down at elevated temperatures, about 570 degrees F (~300C)." I assume the principle is the same here. My inclination is to follow your other suggestion, assigning a proxy price to the N2O.

adam-a-a: @tarnold17 Right- the idea behind this is you'd have N2O being combined with a methane stream for combustion, which of course isn't a part of this flowsheet, hence the suggestion for some sort of proxy. That said, I think what you have here should meet the mark, and we can always go back and refine as needed.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1595_photothermal_membrane_candoP/amo_1595.py
adam-a-a: @tarnold17 something below 50 cents/m3 would be fine, but let me see what you have.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1595_photothermal_membrane_candoP/amo_1595_case_study.yaml
```diff
@@ -7,6 +7,12 @@ defined_flows:
   bcp:
     value: 0.5
     units: USD_2020/kg
+  water:
+    value: 1.0
```
adam-a-a, line 11: This is fine- we'll adjust this with sensitivity analyses as needed.

adam-a-a (approved):
LGTM - we can refine if needed later once we put together an analysis for this (which I think we should).

PR #579, opened 2022-06-21
Electrodialysis_1D documentation 

lbibl:
## Summary/Motivation:
This PR presents the documentation for the recently merged Electrodialysis_1D unit model. 

-Minor changes on the Electrodialysis_0D documentation are carried along.

### Legal Acknowledgement


docs/technical_reference/unit_models/electrodialysis_1D.rst
```diff
@@ -0,0 +1,237 @@
...
+
+   ion_dict = {
+            "solute_list": ["Na_+", "Cl_-"],
+            "mw_data": {"H2O": 18e-3, "Na_+": 23e-3, "Cl_-": 35.5e-3},
+            "electrical_mobility_data": {"Na_+": 5.19e-8, "Cl_-": 7.92e-8},
+            "charge": {"Na_+": 1, "Cl_-": -1},
+        }
+
+This model, by default, uses H\ :sub:`2`\ O  as the solvent of the feed solution.
+
```
aladshaw3, line 149:
```suggestion

Information regarding the property package this unit model relies on can be found here: 

:mod:`watertap.property_models.ion_DSPMDE_prop_pack`
```

aladshaw3 (changes requested):
Looks good so far. 

I suggest adding something like I mention below (a link to some form of documentation for the property package). This should probably also be added to the `electrodialysis_0D.rst` doc page as well. 

There currently is no 'technical reference' for the ion property package, but we should at least link to the API docs so a user will have some idea of how that property package is set up.

docs/technical_reference/unit_models/electrodialysis_1D.rst
lbibl: Accepted.

aladshaw3 (approved):
LGTM

docs/technical_reference/unit_models/electrodialysis_1D.rst
```diff
@@ -0,0 +1,241 @@
...
+All equations are coded as "constraints" (Pyomo). Isothermal and isobaric conditions apply.
+
+Nomenclature
+------------
+.. csv-table:: **Table 5.** Nomenclature
+   :header: "Symbol", "Description", "Unit"
+   :widths: 10, 20, 10
+
+   "**Variables and Parameters**"
```
adam-a-a: It would be good to have Parameters and Variables separate so that we could easily distinguish between the two.

adam-a-a (changes requested):
To be honest, I haven't looked at the 0D doc yet, but I have to say, this documentation is phenomenal! Extremely well done! I only did some touching up on very minor typos and committed them directly to save you time (please review to make sure you're happy with those). 

The only suggested change that I'd like to see is the separation of the Vars and Parameters. Then, I think it's good to go.

docs/technical_reference/unit_models/electrodialysis_1D.rst
```diff
@@ -0,0 +1,241 @@
...
+of water. By its name, "1D" suggests the variation of mass transfer and solution properties over length (l in Figure 1)
+is mathematically described.  The model relies on the following key assumptions:
+
+* The concentrate and diluate channels have identical geometry.
+* For each channel, component fluxes in the bulk solution are uniform in b and s directions in Figure 1
+  but dependent on the l directions (the 1-dimensional assumption).
+* Steady state: all variables are independent of time.
+* Co-current flow operation. 
+* Electrical current is operated below the limiting current. 
+* Ideality assumptions: activity, osmotic, and van't Hoff coefficients are set at one.
```
adam-a-a, line 41: Is the ideality assumption necessarily applicable? The ion prop model has the Davies model, so at the least, activity can deviate from ideality assumptions. Is there something in the ED unit model that would need to be incorporated, beyond what the ion prop model provides (in this case, just referring to the activity)? If not, and the ED model can technically utilize activities calculated by the Davies model (or might be that you'd need to apply activity coefficients--forgot how it was implemented in prop model), then you can boast that the ED model _can _ account for nonideality in terms of activities. My presumption is that you only tested the model with activity coefficients = 1 but the model would work if provided non-ideal activity coefficients. Is that the case?

lbibl: I didn't investigate the activity calculation in the prop model-  is activity defined separately from conc or absorbed into it?    I think for solutions ideality generally suggests particles are point-like and short-range molecular forces are neglected.  I cannot fully specify all non-ideal elements in this scenario at this point, but I am sure it's more than replacing conc by activity.  Things like the charge correlation force gonna be present when salinity is high.  I would keep this statement since it is given in most of the modeling papers.

lbibl:
> To be honest, I haven't looked at the 0D doc yet, but I have to say, this documentation is phenomenal! Extremely well done! I only did some touching up on very minor typos and committed them directly to save you time (please review to make sure you're happy with those).
> 
> The only suggested change that I'd like to see is the separation of the Vars and Parameters. Then, I think it's good to go.

Thank you Adam! I tried to make it readable :)  I addressed all your comments (good ones!).  For ideality I would keep the writing as is as the model hasn't touched any non-ideality stuff and we can come back to it in later refinements.   Austin put a hold on checking the files and I think it'll be good to go after that.

docs/technical_reference/unit_models/electrodialysis_1D.rst
```diff
@@ -0,0 +1,244 @@
...
+   ":math:`L`", "Water permeability (osmosis)", ":math:`ms^{-1}Pa^{-1}`"
+   ":math:`p_{osm}`", "Osmotic pressure", ":math:`Pa`"
+   ":math:`r_{tot}`", "Total areal resistance", ":math:`\Omega m^2`"
+   ":math:`r`", "Membrane areal resistance", ":math:`\Omega m^2`"
+   ":math:`r_{el}`", "Electrode areal resistance", ":math:`\Omega m^2`"
+   ":math:`s`", "Spacer thickness", ":math:`m`"
+   ":math:`\kappa`", "Solution conductivity", ":math:`S m^{-1}\ or\  \Omega^{-1} m^{-1}`"
+   ":math:`\eta`", "Current efficiency for desalination", "dimensionless"
+   ":math:`P`", "Power consumption", ":math:`W`"
+   ":math:`P_Q`", "Specific power consumption", ":math:`kW\ h\  m^{-3}`"
```
adam-a-a, line 225:
```suggestion
   ":math:`P_Q`", "Specific energy consumption", ":math:`kW\ h\  m^{-3}`"
```

adam-a-a (approved):

PR #580, opened 2022-06-21
Microbial battery ZO model and flowsheet

tarnold17:
## Fixes/Resolves:

Unit model and flowsheet for AMO project 1679 (Bostick)

A few corrections to the 1595 flowsheet are also included

-
-

### Legal Acknowledgement


watertap/examples/flowsheets/case_studies/groundwater_treatment/groundwater_treatment.py
```diff
@@ -0,0 +1,346 @@
...
+        doc="Mass of spent filtration media generated per year (accounting for utilization factor)",
+    )
+
+    m.fs.costing.cost_spent_media_disposal = Expression(
+        expr=pyunits.convert(
+            m.fs.costing.spent_media_recovery_mass
+            * m.fs.costing.filtration_media_disposal_cost,
+            to_units=m.fs.costing.base_currency / m.fs.costing.base_period,
+        ),
+        doc="Disposal cost of spent filtration media per year (accounting for utilization factor)",
```
adam-a-a, line 217: I am thinking we should consider a variable for recovering some fraction of the "spent_media," assuming regeneration of media and reuse. Right now, it seems that the model doesn't consider any regeneration. We may have talked about how we were going to handle this, but I think I need a refresher. For the time being, it seems like the current implementations accounts for the worst-case scenario (no regeneration+recovery of media) and doesn't leave room for the potential case where media could be recovered and reused.

watertap/examples/flowsheets/case_studies/groundwater_treatment/tests/test_groundwater_treatment.py
```diff
@@ -0,0 +1,162 @@
...
+            m.fs.filtered_water.flow_mass_comp[0, "nitrate"]
+        )
+
+        assert pytest.approx(1.157407e-10, rel=1e-3) == value(
+            m.fs.filtered_water.flow_mass_comp[0, "phosphates"]
+        )
+
+        assert value(m.fs.filtered_water.flow_mass_comp[0, "iron"]) < 1e-12
+
+        assert value(m.fs.filtered_water.flow_mass_comp[0, "filtration_media"]) < 1e-12
```
adam-a-a, line 118: Why not use relative tol here like you did above?

watertap/examples/flowsheets/case_studies/groundwater_treatment/groundwater_treatment.py
tarnold17: The purpose of regenerating some of the spent media would be to recycle and reuse it. Adding a recycle stream would be a significant change to the flowsheet, so I think if we want to model that, it might make sense to create a separate flowsheet with the recycle stream, and leave this one as it is. However, I think we might want to wait on that until we have a chance to follow up with Bostick. I recall in our previous call with him that he did mention regeneration and reuse of spent media as something that they wanted to consider, but he didn't give us a lot of details about what that might look like. So I think we'd want to ask him a few questions about the appropriate way to model media regeneration before proceeding in that direction.

watertap/examples/flowsheets/case_studies/groundwater_treatment/tests/test_groundwater_treatment.py
tarnold17: Because the amount of filtration media in the filtered water stream should essentially be 0. Notice in the yaml file that the removal fraction of `filtration_media` is 1.

watertap/examples/flowsheets/case_studies/groundwater_treatment/groundwater_treatment.py
adam-a-a: I think there are a couple of approaches that we could try.
1) Trickier approach, but not _too_ difficult: Add a mixer at the beginning of the flowsheet that takes the feed and recycle stream as inputs, and the output is fed to the treatment unit. At the outlet of the treatment unit, we could have a splitter that sends some specified fraction of the iron media back to the inlet, while some fraction gets wasted along with any other constituents that would end up in the waste sludge. 
2) Simpler approach: add a variable for replacement rate that represents the frequency at which the seed must be resupplied. The same could be applicable to the actual media, if ever there were a need to provide initial media and replace it every so often. Overall, I think the model/flowsheet should be set up in such a way that we could specify how much of the initial seed is needed with costing of that AND a variable related to the frequency at which seed or media would need to be replaced/resupplied. I still need to review your PR again to recall correctly, but I _think_ you have the first part covered and just need the replacement factor.

watertap/examples/flowsheets/case_studies/groundwater_treatment/tests/test_groundwater_treatment.py
adam-a-a: I figured that much-- just nitpicking here.

watertap/examples/flowsheets/case_studies/groundwater_treatment/groundwater_treatment.py
```diff
@@ -0,0 +1,346 @@
...
+    )
+
+    m.fs.costing.fresh_media_mass = Expression(
```
adam-a-a, line 220: I think here, or in `cost_fresh_media`, we can assign some separate utilization rate that accounts for the annual frequency at which fresh media is supplied (and paid for).

tarnold17: I might be misunderstanding your point here, but I think what you are describing is essentially how I already have it modeled.

adam-a-a: The way it appears to me, it seems that fresh media is constantly provided throughout the year. I'm describing the addition of something like (e.g.,) 0.05 as the fraction of the operational time within a year that new media is actually provided... unless we are assuming that the quantity of annual fresh media specified is meant to be equivalent to what I'm describing.

On second thought, I think what you have here is sufficient for the initial implementation, and we can refine/augment as needed in subsequent PRs.

adam-a-a (approved):
LGTM for an initial implementation. We can refine and build on this as needed.

watertap/examples/flowsheets/case_studies/groundwater_treatment/groundwater_treatment.py
tarnold17: The model makes it appear like fresh media is added continuously to the inlet stream, but I do not think that is how it would work in reality. Rather, fresh media would probably need to be dosed at discrete, periodic intervals. The amount of fresh media in the inlet stream is meant to provide the "average" amount of fresh media that enters the reactor over time. You can think of this as the model "annualizing" the addition of fresh media provided to the reactor.

We don't have the capability to model the kind of transient behavior that I think would occur in reality for this system. So the "steady-state" values that this model outputs aren't representative of the system at any single individual point in time. Rather, they represent the "average" steady-state over a year of typical operation. I do agree that we might want to refine this model later.

watertap/examples/flowsheets/case_studies/groundwater_treatment/groundwater_treatment.py
andrewlee94: Given that we are still talking about zero-order models at this point, I would suggest that this would be best handled in an abstract manner. I.e. the new media stream would only include the new, fresh media required (minus an recycling), and that there should be a separate operating cost for media regeneration included in the unit operation (and incorporate any capital costs for regeneration into the unit capital cost) Thus, the flowsheet would not see an of the media regeneration at all.

EDIT: Looking further at the actual model implementation, the filtration media appears to be modeled as a component in the water stream (rather than as some static component in the unit model). Is this really the way we want to model this? I.e. do we actually expect the filter media to leave the unit operation in significant amounts and that we will need to consider it in downstream units?

watertap/data/techno_economic/component_list.yaml
```diff
@@ -42,6 +42,8 @@ endothall:
   constituent_longform: Endothall
 endrin:
   constituent_longform: Endrin
+filtration_media:
```
andrewlee94, line 45: As noted in the discussion, I find it odd that we are including the filtration media as a component to be tracked in the water streams.

watertap/examples/flowsheets/case_studies/groundwater_treatment/groundwater_treatment.py
adam-a-a: Without some recycle/regeneration of the media, it would technically leave the system and would be accounted for in sludge disposal costs, but that is not what was originally intended by the group-- they intended to account for regen/recycle. I'm trying to recall the original intent by @tarnold17 here, and I think it had to do partially with the fact that iron is the key reactant and is oxidized to form the filtration media. So, filtration media takes part in the reaction in this example (yet, iron only needs to be seeded into the system for media generation to commence).

I did suggest that we could do without including the media as a component and just consider related capex and opex as an alternative. At this point though, we only intend to put together some initial cost results and reiterate with the team on our implementation and results, so maybe we can merge this and refine in subsequent PRs as needed.

watertap/core/zero_order_costing.py
```diff
@@ -59,6 +59,7 @@
     WellFieldZO,
     PhotothermalMembraneZO,
     CANDOPZO,
+    MicrobialBatteryZO,
```
bknueven: Please consider putting this in the correct spot alphabetically in this list. Or, maybe in this particular case we should consider a `*` import.

watertap/core/zero_order_costing.py
```diff
@@ -2697,6 +2745,7 @@ def _general_power_law_form(
         WellFieldZO: cost_well_field,
         PhotothermalMembraneZO: cost_photothermal_membrane,
         CANDOPZO: cost_CANDOP,
+        MicrobialBatteryZO: cost_microbial_battery,
```
bknueven, line 3112: Similar comment here.

watertap/examples/flowsheets/case_studies/groundwater_treatment/groundwater_treatment_sweep.py
```diff
@@ -0,0 +1,73 @@
...
+        interpolate_nan_outputs=interpolate_nan_outputs,
+    )
+
+    print(global_results)
+
+    return global_results, sweep_params, m
+
+
+if __name__ == "__main__":
+    results, sweep_params, m = run_analysis(*sys.argv[1:])
```
bknueven, line 73: Similar to this comment: https://github.com/watertap-org/watertap/pull/606/files#r919491937

watertap/examples/flowsheets/case_studies/groundwater_treatment/tests/test_groundwater_treatment_sweep.py
```diff
@@ -0,0 +1,35 @@
...
+for case_num in [1, 2, 3]:
+    sweep_list.append(case_num)
+
+
+@pytest.mark.parametrize("case_num", sweep_list)
+@pytest.mark.integration
+def test_sweep(case_num, tmp_path):
+    cwd = os.getcwd()
+    os.chdir(tmp_path)
+    nx = 11
```
bknueven:
```suggestion
    nx = 2
```
Don't see any reason to run more, given we don't check the outputs.

bknueven (commented):
I have a few minor suggestions, similar to those on #606.

We really should clean up the zero-order costing package..

luohezhiming (approved):
Overall, it looks good to me, the `media` component can be changed to its detailed alternative as @adam-a-a suggested, and that can be addressed later.

watertap/examples/flowsheets/case_studies/groundwater_treatment/groundwater_treatment.py
tarnold17: > I think it had to do partially with the fact that iron is the key reactant and is oxidized to form the filtration media

This is precisely why I included filtration media as one of the components - it is essentially a product of one of the reactions that occurs in the unit. I do agree that there could have been other ways to model this; perhaps there is another approach that would have been better. We can certainly refine if needed.

watertap/data/techno_economic/component_list.yaml
tarnold17: @andrewlee94 see my response in the above thread.

PR #555, opened 2022-06-01
Adding costing to ED model

aladshaw3:
## Fixes/Resolves:

Partially addresses model enhancements in #489 

## Summary/Motivation:

Adds costing methods and variables for electrodialysis to the WaterTAP costing block

## Changes proposed in this PR:
- Adding costing for electrodialysis to WaterTAP costing block
- Adding unit tests for costing in the electrodialysis unit model 
- Minor fixes to electrodialysis model 

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #582, opened 2022-06-22
Adding costing to ED model v2

aladshaw3:
## Fixes/Resolves:

Partially addresses model enhancements in #489

Replaces PR #555

## Summary/Motivation:

Adds costing methods and variables for electrodialysis to the WaterTAP costing block

- Adding costing for electrodialysis to WaterTAP costing block
- Adding unit tests for costing in the electrodialysis unit model
- Minor fixes to electrodialysis model

### Legal Acknowledgement

### DIFF
--- a/watertap/costing/watertap_costing_package.py
+++ b/watertap/costing/watertap_costing_package.py
@@ -918,7 +918,7 @@ def cost_electrodialysis_stack(
 def cost_by_flow_volume(blk, flow_cost, flow_to_cost):
     """
     Generic function for costing by flow volume.
-    
+
     Args:
         flow_cost - The cost of the pump in [currency]/([volume]/[time])
         flow_to_cost - The flow costed in [volume]/[time]
--- a/watertap/costing/watertap_costing_package.py
+++ b/watertap/costing/watertap_costing_package.py
@@ -407,6 +407,7 @@ class WaterTAPCostingData(FlowsheetCostingBlockData):
     def cost_nanofiltration(blk):
         """
         Nanofiltration costing method
+
         TODO: describe equations
         """
         make_capital_cost_var(blk)
@@ -422,7 +423,9 @@ class WaterTAPCostingData(FlowsheetCostingBlockData):
     def cost_reverse_osmosis(blk, ro_type=ROType.standard):
         """
         Reverse osmosis costing method
+
         TODO: describe equations
+
         Args:
             ro_type: ROType Enum indicating reverse osmosis type,
                 default = ROType.standard
@@ -449,10 +452,13 @@ class WaterTAPCostingData(FlowsheetCostingBlockData):
     def cost_pump(blk, pump_type=PumpType.high_pressure, cost_electricity_flow=True):
         """
         Pump costing method
+
         TODO: describe equations
+
         Args:
             pump_type: PumpType Enum indicating pump type,
                 default = PumpType.high_pressure
+
             cost_electricity_flow: bool, if True, the Pump's work_mechanical will be
                 converted to kW and costed as an electricity, default = True
         """
@@ -474,10 +480,13 @@ class WaterTAPCostingData(FlowsheetCostingBlockData):
     ):
         """
         Energy recovery device costing method
+
         TODO: describe equations
+
         Args:
             energy_recovery_device_type: EnergyRecoveryDeviceType Enum indicating ERD type,
                 default = EnergyRecoveryDeviceType.pressure_exchanger.
+
             cost_electricity_flow: bool, if True, the ERD's work_mechanical will
                 be converted to kW and costed as an electricity default = True
         """
@@ -493,7 +502,9 @@ class WaterTAPCostingData(FlowsheetCostingBlockData):
     def cost_high_pressure_pump(blk, cost_electricity_flow=True):
         """
         High pressure pump costing method
+
         `TODO: describe equations`
+
         Args:
             cost_electricity_flow - bool, if True, the Pump's work_mechanical will
                                     be converted to kW and costed as an electricity
@@ -518,7 +529,9 @@ class WaterTAPCostingData(FlowsheetCostingBlockData):
     def cost_low_pressure_pump(blk, cost_electricity_flow=True):
         """
         Low pressure pump costing method
+
         TODO: describe equations
+
         Args:
             cost_electricity_flow - bool, if True, the Pump's work_mechanical will
                                     be converted to kW and costed as an electricity
@@ -545,7 +558,9 @@ class WaterTAPCostingData(FlowsheetCostingBlockData):
     def cost_pressure_exchanger_erd(blk, cost_electricity_flow=True):
         """
         ERD pressure exchanger costing method
+
         TODO: describe equations
+
         Args:
             cost_electricity_flow - bool, if True, the ERD's work_mechanical will
                                     be converted to kW and costed as an electricity
@@ -572,6 +587,7 @@ class WaterTAPCostingData(FlowsheetCostingBlockData):
     def cost_pressure_exchanger(blk):
         """
         Pressure exchanger costing method
+
         TODO: describe equations
         """
         cost_by_flow_volume(
@@ -587,10 +603,13 @@ class WaterTAPCostingData(FlowsheetCostingBlockData):
     def cost_mixer(blk, mixer_type=MixerType.default, **kwargs):
         """
         Mixer costing method
+
         TODO: describe equations
+
         Args:
             mixer_type: MixerType Enum indicating mixer type,
                 default = MixerType.default
+
             `**kwargs`: Additional keywords for the MixerType, e.g., NaOCl
                 and CaOH2 mixers expect the `dosing_rate` keyword
                 argument.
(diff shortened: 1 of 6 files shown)


aladshaw3:
NOTE: This PR replaces #555

watertap/costing/watertap_costing_package.py
```diff
@@ -793,6 +855,66 @@ def cost_membrane(blk, membrane_cost, factor_membrane_replacement):
...
+    blk.membrane_replacement_factor = pyo.Expression(expr=membrane_replacement_factor)
+    blk.spacer_cost = pyo.Expression(expr=spacer_cost)
+    blk.electrode_cost = pyo.Expression(expr=electrode_cost)
+    blk.electrode_replacement_factor = pyo.Expression(expr=electrode_replacement_factor)
+
+    blk.capital_cost_constraint = pyo.Constraint(
+        expr=blk.capital_cost
+        == (blk.membrane_cost + blk.spacer_cost)
+        * (
+            blk.unit_model.cell_pair_num
```
lbibl, line 908: we do not need a `2 * ` here because the multiplier has been included in the unit price, correct?  Fine in this way, but I think it may cause confusion for users. Especially for spacers, they actually need to give a doubled unit price as the unit price when using this model.  For membrane, they'll need to add up unit prices of cem and aem.

watertap/unit_models/electrodialysis_0D.py
```diff
@@ -205,7 +205,7 @@ def build(self):
 
         self.cell_pair_num = Var(
             initialize=1,
-            domain=NonNegativeIntegers,
+            domain=NonNegativeReals,
```
lbibl, line 208: Do we wanna change this now or leave it when optimization is actually needed?  We briefly mentioned that there might be a way to walk around this var when doing optimization by optimizing over the total membrane size.  So maybe we can keep this constrain without hurting any goals.  But this change should not impact functions of the unit model so far.

watertap/unit_models/tests/test_electrodialysis_0D.py
```diff
@@ -260,6 +262,44 @@ def test_performance_contents(self, electrodialysis_cell1):
...
+        assert pytest.approx(388.6800, rel=1e-3) == value(
+            m.fs.costing.total_capital_cost
+        )
+        assert pytest.approx(45.86804, rel=1e-3) == value(
+            m.fs.costing.total_operating_cost
+        )
+        assert pytest.approx(777.3600, rel=1e-3) == value(
+            m.fs.costing.total_investment_cost
+        )
+
```
lbibl, line 303: Are there any references these numbers can be compared against to make sense?

lbibl (commented):
No major issues.  Just a few minor comments that can be considered and a question.

watertap/costing/watertap_costing_package.py
aladshaw3: Yeah, I see what you are saying. I think it probably makes sense as well to divide it up by an actual per unit cost. I remember at a prior meeting it was decided not to break it up, but we can change it.

watertap/unit_models/tests/test_electrodialysis_0D.py
aladshaw3: We would need to look in the literature. Presumably, if the voltage and power usage is correct, then this should be accurate.

watertap/unit_models/electrodialysis_0D.py
aladshaw3: I feel like we might as well change this now. It won't have any impact on the model or how the model solves, but would allow a user to try to optimize for the cell numbers if they choose to do so.

lbibl (approved):

aladshaw3:
@lbibl 

I found an error in the 1D model with the units of one of the constraints. This popped up when applying the costing to the 1D model and asserting unit consistency. For some reason, this error is not caught by the first assertion of consistent units, but if I try to assert units consistent just before or after a solve, then it gives an error. 

See below

[UnitsError-MassTransfer-H2O.txt](https://github.com/watertap-org/watertap/files/8968911/UnitsError-MassTransfer-H2O.txt)

From that error, it appears there is some very small unitless number getting introduced to the constraint:

2.099519999999999e-13 + (96485.33212*C/mol)**2: dimensionless not compatible with coulomb ** 2 / mole ** 2.

It only seems to show up for the water balance and only at the initial node [t=0,x=0]

aladshaw3:
@lbibl 

The error appears ONLY after scaling factors get calculated. So some how, whatever scaling factor you applied to that constraint is not a unitless number. 

Let me see if I can fix that in this PR.

lbibl:
> @lbibl
> 
> The error appears ONLY after scaling factors get calculated. So some how, whatever scaling factor you applied to that constraint is not a unitless number.
> 
> Let me see if I can fix that in this PR.

I am not exactly sure what you are doing.  But one thing to be aware is that the mass transfer term in 1D control volume block carries a different unit from that in 0D.  So the costing for 1D might need some modification from 0D.  All scaling factors were calculated in the 1D unit test so what you described should have been caught in that test if there were any error

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -1070,7 +1072,7 @@ def calculate_scaling_factors(self):
                     )
                     ** 0.5
                 )
-                sf_eleosm = Constants.faraday_constant
+                sf_eleosm = Constants.faraday_constant * pyunits.mol / pyunits.coulomb
```
lbibl: OK. I see what happened here.  Interesting.  So it was because the `constant` carries a unit in addition to its value.  I need to investigate why this wasn't caught in the unit test.

aladshaw3: As far as I can tell, this was not caught because we checked units before calculating scaling factors. The scaling factor is applied to the constraint after it is calculated, so maybe we need to check units after scaling from now on.

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -696,6 +696,8 @@ def eq_mass_transfer_term_concentrate(self, t, x, p, j):
             self.diluate.length_domain,
             doc="Electrical power consumption of a stack",
         )
+        # # TODO: Check this constraint formulation. Is this correct?
+        #           it states: dP/dx = b*L*u(x)*i(x)
```
lbibl: Yes, I think it is. Be noted that the power for 1D is indexed by x so, when doing costing, you'll need to use `power_electrical_x[0, self.diluate.length_domain.last()]` to represent the total power consumed through the entire length.

aladshaw3: Thanks for checking on this.

What I did for the power was to used the variables from 0D or 1D based on the `_get_performance_contents` function. Then I also made sure that each key in the performance table had the same name. Thus, in the Costing package, I could reference the same performance metric regardless of unit model type.

watertap/unit_models/electrodialysis_1D.py
lbibl: Ok. get you.  But I think that also suggests whether or not scaling carries units doesn't matter as IPOPT will strip the unit (if any). Otherwise the solver will fail.  So this is an error defined by us in that sense.  But I will  make sure to only take in values when calculating scaling factors as a good habit :)

lbibl: I would also try using the `value()` or `.value` method here.  The way you are writing now is converting the unit to `dimensionless`, which might still be a unit (not 100% sure without looking into the source ). I think the best way is to pass only a value.

watertap/unit_models/electrodialysis_1D.py
lbibl: Then it should not be of an issue since the index has been specified in the `_get_performance_contents` function.  Let me know if you ran into any related problems.

watertap/costing/watertap_costing_package.py
```diff
@@ -411,6 +447,9 @@ def cost_reverse_osmosis(blk, ro_type=ROType.standard):
                 f"{blk.unit_model.name} received invalid argument for ro_type:"
                 f" {ro_type}. Argument must be a member of the ROType Enum."
             )
+        make_capital_cost_var(blk)
+        make_fixed_operating_cost_var(blk)
+
```
bknueven:
```suggestion
```

watertap/costing/watertap_costing_package.py
```diff
@@ -383,6 +416,9 @@ def cost_nanofiltration(blk):
 
         TODO: describe equations
         """
+        make_capital_cost_var(blk)
+        make_fixed_operating_cost_var(blk)
+
```
bknueven:
```suggestion
```

watertap/costing/watertap_costing_package.py
```diff
@@ -776,10 +856,7 @@ def cost_membrane(blk, membrane_cost, factor_membrane_replacement):
         membrane_cost - The cost of the membrane in currency per area
         factor_membrane_replacement - Membrane replacement factor
                                       [fraction of membrane replaced/year]
-
     """
-    make_capital_cost_var(blk)
-    make_fixed_operating_cost_var(blk)
```
bknueven, line 854:
```suggestion

    make_capital_cost_var(blk)
    make_fixed_operating_cost_var(blk)
```

bknueven (changes requested):
Please remove the changes to the membrane costing methods. Otherwise looks good.

bknueven (approved):

PR #592, opened 2022-06-23
Add CI workflow for linkchecking

### LINKED ISSUE #429: Move check for HTML links to dedicated CI workflow/job
- It used to be part of the PR checks until #428 
- Since links become broken more often than not because of external rather than internal changes, it would make sense to have this check run on schedule and/or less often than for every PR commit

lbianchi-lbl:
## Summary/Motivation:

- Instead of running for every PR, this workflow will run:
  - On schedule (currently 1/day)
  - On demand (will become available once the PR is merged), with the possibility to set options

### Legal Acknowledgement

### DIFF
--- a/.github/workflows/linkcheck.yml
+++ b/.github/workflows/linkcheck.yml
@@ -8,7 +8,7 @@ on:
         default: https://watertap.readthedocs.io/en/latest/
         required: false
         description: Base url to check
-      exclude_url_regex:
+      ignore_url_regex:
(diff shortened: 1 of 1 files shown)


adam-a-a:
@lbianchi-lbl Was there anything else that you needed to add/modify before taking this out of draft mode?

adam-a-a:
@lbianchi-lbl Do you think we can get this one through, or should it remain on the backburner?

PR #603, opened 2022-07-01
Simple example

TimBartholomew:
## Fixes/Resolves:

NA

## Summary/Motivation:
The simple examples used for onboarding new users/developers should be in the repository, rather than just on Box.

- Add the simple property package file and demo
- Add the simple filter file and demo
- Add test files

### Legal Acknowledgement


bknueven:
I haven't reviewed the code thoroughly, but I do have a few questions:
1. Does a "simple demo" really involve building your own unit model and property package? Why not use one we've already included?
2. Should the unit model and property package be in the main part of the repository?
3. As-is, this doesn't actually *use* anything in watertap. Maybe it should go in the IDAES examples repository?

bknueven:
If this is to be for developers, I think we need to come up with a different name for it.

adam-a-a (approved):
LGTM - we might want to consider converting to documentation (i.e., tutorials) or point directly to this simple_demo from the documentation in a future PR.

bknueven (approved):

PR #606, opened 2022-07-01
AMO Swine wastewater treatment flowsheet

adam-a-a:
## Summary/Motivation:
Develop flowsheet for swine wastewater treatment along with some minimal sweep analysis to get things started.

- revise unit models and associated tests as needed
-  add costing to unit model YAML files and costing package
- implement flowsheet and tests with minimal set of regression tests
- implement small subset of parameter sweeps in multi_sweep to initiate sensitivity analysis

### Legal Acknowledgement


watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/swine_wwt/swine_wwt.py
```diff
@@ -0,0 +1,711 @@
...
+    # Annual food waste tipping fee revenue
+    costing.annual_costs_revenues.annual_food_waste_revenue = Expression(
+        expr=(
+            pyunits.convert(
+                m.fs.costing.food_waste_tipping_fee_cost
+                * costing.annual_production.annual_food_waste,
+                to_units=m.fs.costing.base_currency / m.fs.costing.base_period,
+            )
+        ),
+        doc="Annual sludge disposal cost",
```
adam-a-a:
```suggestion
        doc="Annual food waste tipping fee revenue",
```

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/swine_wwt/swine_wwt.py
```diff
@@ -0,0 +1,711 @@
...
+    m = build()
+
+    set_operating_conditions(m)
+    assert_degrees_of_freedom(m, 0)
+    assert_units_consistent(m)
+
+    initialize_system(m)
+
+    results = solve(m)
+    assert_optimal_termination(results)
```
andrewlee94, line 62: Should we be adding a `try/except` on this (or perhaps doing a `check_optimal_termination` instead of `assert_optimal_termination`)? In the current form, if the result in not optimal the user will see an `AssertionError` with  default message which might not be that useful for explaining what happened.

watertap/unit_models/zero_order/ion_exchange_zo.py
```diff
@@ -41,9 +41,10 @@ def build(self):
         pump_electricity(self, self._Q)
 
         # mutable parameter; default value found in WT3 for anion exchange
-        self.eta_pump.set_value(0.8)
-        # mutable parameter; default value of 2 bar converted to feet head
-        self.lift_height.set_value(69.91052 * pyunits.feet)
+        if self.config.process_subtype != "clinoptilolite":
```
andrewlee94: A minor point, but for future maintenance, it might be better to call out the exceptions instead:

```
if self.config.process_subtype != "clinoptilolite":
    pass
else:
    self.eta_pump.set_value(0.8)
    # mutable parameter; default value of 2 bar converted to feet head
    self.lift_height.set_value(69.91052 * pyunits.feet)
```

andrewlee94 (commented):
Only a couple of minor comments/suggestions,

watertap/unit_models/zero_order/ion_exchange_zo.py
adam-a-a: We could do that, except the logic would be the other way around. Basically, the default WT3 version specified said parameter values, while for this "clinoptilolite" subtype, I want to ignore those specifications. So, for now, I only check whether subtype is or isn't "clinoptilolite". If it isn't, keep the old specifications from WT3; if it is "clinoptilolite", don't touch the default parameter values which are typically used across most other models with pump electricity.

andrewlee94: Sorry, that is what I meant.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/swine_wwt/swine_wwt.py
adam-a-a: We _could_, but then again, we've been doing this for all the AMO/zero-order model flowsheets, so we'd have to change this in all those flowsheets as well.  And currently, we're able to solve without issues, so if this were to fail, it would be because the user changed something in terms of fixing variables under `set_operating_conditions`. I think this serves its purpose as is for now.

Moreover, at the moment, our main objective is to get initial flowsheets implemented, produce techno-economic results, and iterate with teams on those results and model setup to determine next steps. @dangunter is working on a UI, so maybe this is something we'd want to consider at that level (i.e., a useful message for the user if/when result is not optimal).

watertap/unit_models/zero_order/ion_exchange_zo.py
```diff
@@ -41,9 +41,10 @@ def build(self):
...
+            self.eta_pump.set_value(0.8)
+            # mutable parameter; default value of 2 bar converted to feet head
+            self.lift_height.set_value(69.91052 * pyunits.feet)
```
adam-a-a:
```suggestion
        if self.config.process_subtype == "clinoptilolite":
            pass
        else:
            self.eta_pump.set_value(0.8)
            # mutable parameter; default value of 2 bar converted to feet head
            self.lift_height.set_value(69.91052 * pyunits.feet)
```

watertap/core/zero_order_costing.py
```diff
@@ -856,6 +927,113 @@ def cost_coag_and_floc(blk):
...
+        """
+        Method for costing constructed wetlands.
+        """
+        t0 = blk.flowsheet().time.first()
+
+        # Get parameter dict from database
+        parameter_dict = blk.unit_model.config.database.get_unit_operation_parameters(
+            blk.unit_model._tech_type, subtype=blk.unit_model.config.process_subtype
+        )
+        parameter_dict
```
bknueven:
```suggestion 

```

watertap/data/techno_economic/tests/test_unit_parameter_files.py
```diff
@@ -102,6 +103,8 @@ def test_unit_parameter_files(tech):
         "water_pumping_station",
         "well_field",
         "mbr",
+        "vfa_recovery",
+        "cofermentation",
```
bknueven: It would be preferable to maintain the alphabetical order here, so it's easy to know if something is in the list by inspection.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/swine_wwt/multi_sweep.py
```diff
@@ -0,0 +1,100 @@
...
+        optimize_function=opt_function,
+        optimize_kwargs=optimize_kwargs,
+        interpolate_nan_outputs=interpolate_nan_outputs,
+    )
+
+    return global_results, sweep_params, m
+
+
+if __name__ == "__main__":
+    results, sweep_params, m = run_analysis(*sys.argv[1:])
```
bknueven: It would be nice to put a basic help message here, e.g.,
```suggestion
    if len(sys.argv) == 1:
        print("Usage: python multi_sweep.py case_number number_of_samples interpolate_nan_outputs")
        print(f"Results will be written to {os.path.dirname(os.path.abspath(__file__))}")
    else:
        results, sweep_params, m = run_analysis(*sys.argv[1:])
```

bknueven (commented):
I have a few minor suggestions, mostly relating to maintenance.

At a high level, the zero order costing package has too much repeated code, and this PR is not helping matters. But it should be cleaned up subsequently.

adam-a-a:
> I have a few minor suggestions, mostly relating to maintenance.
> 
> At a high level, the zero order costing package has too much repeated code, and this PR is not helping matters. But it should be cleaned up subsequently.

@bknueven I addressed your suggestions. As you mentioned, we could refactor the zero order costing package in a subsequent PR to eliminate repeated code.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/swine_wwt/swine_wwt.py
adam-a-a: @andrewlee94 I think replacing with `check_optimal_termination` in all AMO flowsheets in a separate PR would be good. Any other issues that need to be addressed now before you approve?

andrewlee94 (approved):

bknueven (approved):

PR #608, opened 2022-07-11
Replace removal_frac_mass_solute with removal_frac_mass_comp

### LINKED ISSUE #347: Renaming "solute" in zero-order models and omitting water recovery from input data
Not all components will be solutes, so the naming of `removal_frac_mass_solute` can be changed to `removal_frac_mass_comp`. From this, `recovery_frac_mass_H2O` would no longer be needed in YAML files to load into unit models. Instead it could be calculated as 1 - `removal_frac_mass_comp['H2O']`. 

Changing this would alter our WT3 migration work and could slow down our effort for WT3 and AMO model implementation, so it might be best to consider addressing this issue once models are in place.

See the conversation in PR #340 for more details on the discussion.

adam-a-a:
## Summary/Motivation:
This PR replaces `removal_frac_mass_solute` with `removal_frac_mass_comp` in all files related to zero-order modeling.  

Note: this PR does not alter the water recovery variable and still only includes removal fractions for all components aside from water. The PR simply swaps names.

- replace  `removal_frac_mass_solute` with `removal_frac_mass_comp`
- revise helper methods documentation to reflect these changes

### Legal Acknowledgement


andrewlee94:
My one comment/concern with this is that the name `removal_frac_mass_comp` does not indicate that the indexing set is only the solute species (i.e. does not include water). Whilst we do not need to blindly follow the IDAES naming conventions, the conventions were set up to clearly indicate the indexing set via the suffix, and thus `comp` would indicate that the variable is indexed over all components (including water), whereas `solute` indicates it is indexed only by solutes.

A second (minor) question is whether we need to do a proper deprecation for this or not. I suspect we are not yet at the point where we need to be too struct with this, but we should keep this in mind and get in the habit of asking these questions whenever we make API changes like this.

adam-a-a:
> My one comment/concern with this is that the name `removal_frac_mass_comp` does not indicate that the indexing set is only the solute species (i.e. does not include water). Whilst we do not need to blindly follow the IDAES naming conventions, the conventions were set up to clearly indicate the indexing set via the suffix, and thus `comp` would indicate that the variable is indexed over all components (including water), whereas `solute` indicates it is indexed only by solutes.
> 
> A second (minor) question is whether we need to do a proper deprecation for this or not. I suspect we are not yet at the point where we need to be too struct with this, but we should keep this in mind and get in the habit of asking these questions whenever we make API changes like this.

To your first point, I agree with the sentiment, but I think this is an intermediate step until we make a final decision for water recovery. In my view, we can resolve this simply by being explicit in the documentation that removal_frac_mass_comp is for everything but water. So I see this as the minimal change that we can make for the time being. I'm also fine with keeping `removal_frac_mass_solute` and documenting that it is indexed by solute_set but can include any constituent in water, not just solutes.

To the second point, I haven't done a "proper deprecation" thus far, so would need to check with you to make sure all boxes were checked if we were to do so. However, I don't think we need it for this.

andrewlee94:
@adam-a-a In that case, the issue here is more about the naming of the `solute_set` then, rather that the removal fraction and we should probably fix that instead. That should also be an easy fix (a find and replace just like this one), so all we would need is a preferred name (which is often hard to agree upon).

Some ideas:

* `contaminant_set` - probably better than `solute_set` in that it would include both solutes and suspended solids, but not all species are unwanted (which is implied by "contaminant").
* `aqueous_set` - aqueous tends to apply to dissolved species only (and often carries the connotation of ions).
* `constituent_set` - probably too obscure, and water is also a constituent

adam-a-a:
> @adam-a-a In that case, the issue here is more about the naming of the `solute_set` then, rather that the removal fraction and we should probably fix that instead. That should also be an easy fix (a find and replace just like this one), so all we would need is a preferred name (which is often hard to agree upon).
> 
> Some ideas:
> 
> * `contaminant_set` - probably better than `solute_set` in that it would include both solutes and suspended solids, but not all species are unwanted (which is implied by "contaminant").
> * `aqueous_set` - aqueous tends to apply to dissolved species only (and often carries the connotation of ions).
> * `constituent_set` - probably too obscure, and water is also a constituent

I like `constituent_set` because in WaterTAP, the constituents are generally in reference to constituents _in water_.  However, would it be that straight forward? I would imagine that solute_set from outside the zero-order modeling framework could be mixed in with search results.

andrewlee94:
@adam-a-a You would need to make sure you only touched the zero-order files, but you should be able to limit your search to the a couple of folders (the ZO models are fairly well contained).

However, one thing that this would do is it would limit the zero-order models to only use zero-order property packages (i.e. ones with the `contaminant_set` defined). At the moment, the zero-order units models can theoretically use any property package (as long as water is a component and it has a solute set). That may not be a huge issue though,

adam-a-a:
@andrewlee94 Let's hear @TimBartholomew 's thoughts on this since I think he was the main proponent of changing the naming of `removal_frac_mass_solute`. 

I just put this up to scratch off an issue and don't have much of a preference, except for the idea that we shouldn't overdo these types of changes and should instead focus time/effort on more impactful tasks.

adam-a-a:
> @adam-a-a You would need to make sure you only touched the zero-order files, but you should be able to limit your search to the a couple of folders (the ZO models are fairly well contained).
> 
> However, one thing that this would do is it would limit the zero-order models to only use zero-order property packages (i.e. ones with the `contaminant_set` defined). At the moment, the zero-order units models can theoretically use any property package (as long as water is a component and it has a solute set). That may not be a huge issue though,

@andrewlee94 I'd prefer to keep `solute_set` as the index for the time being to avoid any issues like this. And going back to your original comment about using the `comp` qualifier and how that technically should include water as a component--I should point out that we might already have several properties (e.g., in NaCl prop pack) that use `phase_comp` qualifiers and are indexed by solute set rather than component list. So, we wouldn't be deviating from what we've been doing in WaterTAP so far.

adam-a-a:
> > @adam-a-a You would need to make sure you only touched the zero-order files, but you should be able to limit your search to the a couple of folders (the ZO models are fairly well contained).
> > However, one thing that this would do is it would limit the zero-order models to only use zero-order property packages (i.e. ones with the `contaminant_set` defined). At the moment, the zero-order units models can theoretically use any property package (as long as water is a component and it has a solute set). That may not be a huge issue though,
> 
> @andrewlee94 I'd prefer to keep `solute_set` as the index for the time being to avoid any issues like this. And going back to your original comment about using the `comp` qualifier and how that technically should include water as a component--I should point out that we might already have several properties (e.g., in NaCl prop pack) that use `phase_comp` qualifiers and are indexed by solute set rather than component list. So, we wouldn't be deviating from what we've been doing in WaterTAP so far.

Actually, what I said isn't entirely true. There are only select cases where we use the `comp` qualifier and only index by solute set and/or ion set, but not in the NaCl prop pack.

TimBartholomew (approved):
LGTM

tarnold17 (approved):

PR #617, opened 2022-07-14
GAC Costing Package

hunterbarber:
## Fixes/Resolves:
n/a

## Summary/Motivation:
Initial implementation of capital and operating costs for a pressure GAC unit

- added GAC costing parameters and model into costing package
- updated testing in GAC unit model test file

### Legal Acknowledgement


hunterbarber:
Draft PR until there's some review and suggestions on costing package conventions

watertap/costing/watertap_costing_package.py
```diff
@@ -859,6 +914,109 @@ def cost_crystallizer_by_volume(blk):
...
+            == blk.costing_package.gac_num_contactors
+            * pyo.units.convert(
+                (
+                    blk.costing_package.gac_contactor_cost_coeff_3
+                    * blk.unit_model.bed_volume**3
+                    + blk.costing_package.gac_contactor_cost_coeff_2
+                    * blk.unit_model.bed_volume**2
+                    + blk.costing_package.gac_contactor_cost_coeff_1
+                    * blk.unit_model.bed_volume**1
+                    + blk.costing_package.gac_contactor_cost_coeff_0
```
adam-a-a: Just brainstorming here-- suppose someone wanted to set `gac_num_contactors` to something like 10, with 8 operational and 2 redundant, for example. If we are assuming operational and redundant vessels are sized the same, then we only care about `gac_num_contactors`. However, in the cost correlation, should it then be
```suggestion
                    blk.costing_package.gac_contactor_cost_coeff_3
                    * (blk.unit_model.bed_volume/blk.costing_package.gac_num_contactors)**3
                    + blk.costing_package.gac_contactor_cost_coeff_2
                    * (blk.unit_model.bed_volume/blk.costing_package.gac_num_contactors)**2
                    + blk.costing_package.gac_contactor_cost_coeff_1
                    * (blk.unit_model.bed_volume/blk.costing_package.gac_num_contactors)**1
                    + blk.costing_package.gac_contactor_cost_coeff_0

```

watertap/costing/watertap_costing_package.py
```diff
@@ -859,6 +914,109 @@ def cost_crystallizer_by_volume(blk):
...
+                    * pyo.exp(
+                        blk.unit_model.bed_mass_gac
+                        * blk.costing_package.gac_adsorbent_unit_cost_exp_coeff
+                    ),
+                    to_units=blk.costing_package.base_currency * pyo.units.kg**-1,
+                )
+            )
+        else:
+            blk.adsorbent_unit_cost_constraint = pyo.Constraint(
+                expr=blk.adsorbent_unit_cost
```
adam-a-a: Just a thought- do we want to consider using an Expr_IF here?

adam-a-a: Also, you'll want a test to make sure this works as intended.

watertap/costing/watertap_costing_package.py
adam-a-a: If `unit_model.bed_volume` is representative of the GAC's total bed volume which would then be split up accordingly as defined by number of contactors, I'd think the costing should account for the volume of each contactor, then add them all up (i.e., multiply by number of contactors).

EDIT: Actually, that isn't _entirely_ correct. The unit bed volume should probably represent the total OPERATIONAL bed volume, so then you'd divide by number of operational contactors (within cost equation where `bed_volume` is specified) and then multiply by total number of contactors. I could be wrong, but let's discuss.

luohezhiming (commented):
I recommend you to add corresponding costing tests in `test_gac.py`.

watertap/costing/watertap_costing_package.py
hunterbarber: Yes, with respect with what you said in the edit. For moderate design flows the 1 operating, 1 redundant, is fine to cost and works as intended in the current implementation. However, with large design flows, the 1 operating vessel may scale too large in size than practical and return economy of scale results that says having this 1000+ cubic meter vessel is feasible and cost effective (which is unrealistic), so the user option should be available. I intend on doing what you said with fractionally decreasing the costed volume based on `gac_num_contactors_operational` within the correlation, and accounting for `gac_num_contactors_operational + gac_num_contactors_redundant` as the multiplier. Just need to figure out the best way to do it with creating the least and most intuitive amount of additional variables.

watertap/costing/watertap_costing_package.py
hunterbarber: I agree that there is a better way to write this. This is simply trying to impose a maximum on the variable that if purchasing more than the 18143.68 kg, the unit price isn't improved. Didn't know whether to overwrite a temp variable and write only one equation or do something similar to this, but I would like to find a current WaterTAP model that handles this kind of if statement to recode in a similar way.

hunterbarber: I'm assuming Expr_IF handles conditionals flipping between true and false as a variable moves from the initial to solution value? That would be the concern with how it's currently done.

bknueven: I would consider using `smooth_min`, e.g., `bed_mass_for_cost == smooth_min(18143.68, blk.unit_model.bed_mass_gac)`

watertap/costing/watertap_costing_package.py
```diff
@@ -292,6 +294,91 @@ def build_global_params(self):
...
+        )
+
+        self.gac_num_contactors_redundant = pyo.Var(
+            initialize=1,
+            units=pyo.units.dimensionless,
+            doc="Number of off-line redundant GAC contactors in parallel",
+        )
+
+        self.gac_contactor_cost_coeff_0 = pyo.Var(
+            initialize=10010.859100985,
```
hunterbarber:
```suggestion
            initialize=10010.9,

```

watertap/costing/watertap_costing_package.py
```diff
@@ -292,6 +294,91 @@ def build_global_params(self):
...
+            units=pyo.units.USD_2020,
+            doc="GAC contactor polynomial cost coefficient 0",
+        )
+
+        self.gac_contactor_cost_coeff_1 = pyo.Var(
+            initialize=2204.95245503305,
```
hunterbarber:
```suggestion
            initialize=2204.95,

```

watertap/costing/watertap_costing_package.py
```diff
@@ -292,6 +294,91 @@ def build_global_params(self):
...
+            units=pyo.units.USD_2020 * (pyo.units.m**3) ** -1,
+            doc="GAC contactor polynomial cost coefficient 1",
+        )
+
+        self.gac_contactor_cost_coeff_2 = pyo.Var(
+            initialize=-15.9377885738609,
```
hunterbarber:
```suggestion
            initialize=-15.9378,

```

watertap/costing/watertap_costing_package.py
```diff
@@ -292,6 +294,91 @@ def build_global_params(self):
...
+            units=pyo.units.USD_2020 * (pyo.units.m**3) ** -2,
+            doc="GAC contactor polynomial cost coefficient 2",
+        )
+
+        self.gac_contactor_cost_coeff_3 = pyo.Var(
+            initialize=0.110591952020239,
```
hunterbarber:
```suggestion
            initialize=0.110592,

```

watertap/costing/watertap_costing_package.py
```diff
@@ -292,6 +294,91 @@ def build_global_params(self):
...
+            units=pyo.units.USD_2020 * (pyo.units.m**3) ** -3,
+            doc="GAC contactor polynomial cost coefficient 3",
+        )
+
+        self.bed_mass_gac_max_ref = pyo.Var(
+            initialize=18143.68,
```
hunterbarber:
```suggestion
            initialize=18143.7,

```

watertap/costing/watertap_costing_package.py
```diff
@@ -292,6 +294,91 @@ def build_global_params(self):
...
+            units=pyo.units.kg,
+            doc="Reference maximum value of GAC mass needed for initial charge where "
+            "economy of scale no longer discounts the unit price",
+        )
+
+        self.gac_adsorbent_unit_cost_coeff = pyo.Var(
+            initialize=4.58341889627683,
```
hunterbarber:
```suggestion
            initialize=4.58342,

```

watertap/costing/watertap_costing_package.py
```diff
@@ -292,6 +294,91 @@ def build_global_params(self):
...
+            units=pyo.units.USD_2020 * pyo.units.kg**-1,
+            doc="GAC adsorbent exponential cost pre-exponential coefficient",
+        )
+
+        self.gac_adsorbent_unit_cost_exp_coeff = pyo.Var(
+            initialize=-1.25310852043246e-5,
```
hunterbarber:
```suggestion
            initialize=-1.25311e-5,

```

watertap/costing/watertap_costing_package.py
```diff
@@ -292,6 +294,91 @@ def build_global_params(self):
...
+            units=pyo.units.kg**-1,
+            doc="GAC adsorbent exponential cost parameter coefficient",
+        )
+
+        self.gac_other_cost_coeff = pyo.Var(
+            initialize=16660.6931081066,
```
hunterbarber:
```suggestion
            initialize=16660.7,

```

watertap/costing/watertap_costing_package.py
```diff
@@ -292,6 +294,91 @@ def build_global_params(self):
...
+            units=pyo.units.USD_2020,
+            doc="GAC other cost power law coefficient",
+        )
+
+        self.gac_other_cost_exp = pyo.Var(
+            initialize=0.552206579,
```
hunterbarber:
```suggestion
            initialize=0.552207,

```

watertap/costing/watertap_costing_package.py
```diff
@@ -292,6 +294,91 @@ def build_global_params(self):
...
+        )
+
+        self.gac_regen_frac = pyo.Var(
+            initialize=0.70,
+            units=pyo.units.dimensionless,
+            doc="Fraction of spent GAC adsorbent that can be regenerated for reuse",
+        )
+
+        self.gac_regen_unit_cost = pyo.Var(
+            initialize=4.283522249590950,
```
hunterbarber:
```suggestion
            initialize=4.28352,

```

watertap/costing/watertap_costing_package.py
```diff
@@ -292,6 +294,91 @@ def build_global_params(self):
...
+            units=pyo.units.USD_2020 * pyo.units.kg**-1,
+            doc="Unit cost to regenerate spent GAC adsorbent by an offsite regeneration facility",
+        )
+
+        self.gac_makeup_unit_cost = pyo.Var(
+            initialize=4.582232978299970,
```
hunterbarber:
```suggestion
            initialize=4.58223,

```

hunterbarber (commented):
Standardize regressed variables to 6 significant figures

watertap/costing/watertap_costing_package.py
```diff
@@ -857,6 +944,180 @@ def cost_crystallizer_by_volume(blk):
...
+                to_units=blk.costing_package.base_currency * pyo.units.kg**-1,
+            )
+        )
+        blk.adsorbent_cost_constraint = pyo.Constraint(
+            expr=blk.adsorbent_cost
+            == blk.adsorbent_unit_cost * blk.unit_model.bed_mass_gac
+        )
+        other_process_cost_units = (
+            pyo.units.m**3
+        ) ** -blk.costing_package.gac_other_cost_exp
```
bknueven: I would just inline this below for clarity.

bknueven (approved):
Seems reasonable.

I would like somebody else's opinion on the use of `smooth_min`.

watertap/costing/watertap_costing_package.py
andrewlee94: Either an `expr_if` or a `smooth_min` would work here, and both have advantages and disadvantages.

`smooth_min` is smooth and continuous which is good, but it is not the nicest non-linear constraint.
`expr_if` on the other hand does have an issue with being discontinuous in the 1st derivative, which can cause issues if the solution lies close to that point, but is otherwise probably OK.

watertap/costing/watertap_costing_package.py
hunterbarber: @bknueven Just for clarification, do you mean inline commenting to clarify the "other" (`# other accompanying vessels, pipes, instrumentation, and controls`) or actual code structure changes?

bknueven: No, just replace `other_process_cost_units` below with `(pyo.units.m**3)** - blk.costing_package.gac_other_cost_exp`

hunterbarber: Got it, more so for conciseness and minimizing declaring intermediate variables (I think I got fixated on the word clarify). I've gone ahead and made the edit to inline.

watertap/costing/watertap_costing_package.py
```diff
@@ -857,6 +944,177 @@ def cost_crystallizer_by_volume(blk):
...
+            other process costs calculated power law with total contactor volume
+        Capital costs calculated using required makeup and regenerated GAC adsorbent
+            Energy for backwash and booster pumps considered negligible compared to regeneration costs
+        """
+        make_capital_cost_var(blk)
+        blk.contactor_cost = pyo.Var(
+            initialize=1e5,
+            domain=pyo.NonNegativeReals,
+            units=blk.costing_package.base_currency,
+            doc="Unit contactor capital cost",
```
hunterbarber:
```suggestion
            doc="Unit contactor(s) capital cost",
```

watertap/costing/watertap_costing_package.py
```diff
@@ -857,6 +944,177 @@ def cost_crystallizer_by_volume(blk):
...
+            domain=pyo.NonNegativeReals,
+            units=pyo.units.kg,
+            doc="Reference value of GAC mass needed for initial charge where "
+            "economy of scale no longer discounts the unit price",
+        )
+        blk.adsorbent_unit_cost = pyo.Var(
+            initialize=2,
+            domain=pyo.NonNegativeReals,
+            units=blk.costing_package.base_currency * pyo.units.kg**-1,
+            doc="Unit adsorbent unit cost per mass",
```
hunterbarber:
```suggestion
            doc="GAC adsorbent cost per unit mass",
```

hunterbarber (commented):
Wording changes corresponding to #637

hunterbarber:
Subsequent PR #637 for documentation of costing package

adam-a-a (approved):
LGTM

PR #623, opened 2022-07-15
Simple fix to remove _get_performance_contents from ED costing

### LINKED ISSUE #613: Modify cost_electrodialysis function not to use _get_performance_contents
The `_get_performance_contents` function shouldn't be used for calculating the costing for the electrodialysis unit model, it typically only accounts for obtaining system performance. I suggest to change it to directly use `power_electrical` .

aladshaw3:
## Fixes/Resolves:

#613 

## Summary/Motivation:
Remove reference to the get performance dict from costing file. Simple solution is to have a function that just points to the correct var in both ED models. 

Alternative would be to use `add_object_reference` to the 1D ED model, but then we still need to handle the different indices that each of the `power_electrical` vars have. I think this is a good quick compromise. 

- Adds a get_power_electrical function for 0D and 1D ED
- Uses get_power_electrical function in the costing package 

### Legal Acknowledgement

### DIFF
--- a/watertap/costing/watertap_costing_package.py
+++ b/watertap/costing/watertap_costing_package.py
@@ -749,9 +749,7 @@ class WaterTAPCostingData(FlowsheetCostingBlockData):
         if cost_electricity_flow:
             blk.costing_package.cost_flow(
                 pyo.units.convert(
-                    blk.unit_model._get_performance_contents(t0)["vars"][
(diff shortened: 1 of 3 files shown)


luohezhiming (approved):
Looks good to me!

adam-a-a (approved):
LGTM

PR #624, opened 2022-07-18
Add UV cost

luohezhiming:
## Summary/Motivation:

Add costing method for UV system

- add costing
- add corresponding tests

### Legal Acknowledgement


watertap/costing/watertap_costing_package.py
```diff
@@ -1058,3 +1100,35 @@ def cost_by_flow_volume(blk, flow_cost, flow_to_cost):
     blk.capital_cost_constraint = pyo.Constraint(
         expr=blk.capital_cost == blk.flow_cost * flow_to_cost
     )
+
+
+def cost_uv_aop_stack(blk, flow_cost, electricity_cost, factor_uv_replacement):
```
hunterbarber: Not sure `uv_aop_stack` is the nomenclature that best represents the UV reactor configuration. ED is costed as such because it seems to be completely modular, so the stack of cells makes sense. Here, something like `uv_aop_lamp_bundle` would be a physical equivalent, but the model is not being costed as such and therefore I'm not sure of the reason this is presented as its own function, similar to the ED costing. Maybe you and others can comment for clarification.

watertap/costing/watertap_costing_package.py
```diff
@@ -1058,3 +1100,35 @@ def cost_by_flow_volume(blk, flow_cost, flow_to_cost):
...
+
+    reactor_volume = pyo.units.convert(
+        blk.unit_model.reactor_volume, to_units=pyo.units.m**3
+    )
+
+    electricity_demand = pyo.units.convert(
+        blk.unit_model.electricity_demand[0], to_units=pyo.units.kW
+    )
+
+    blk.capital_cost_constraint = pyo.Constraint(
```
hunterbarber, line 1385: Little confused with how `blk.electricity_cost * electricity_demand` is a capital cost, just thinking that some electricity demand in kW (or GJ/yr or any equivalent) should be a fixed operating cost. Are you costing the uv (lamp?) replacement as a function of the electricity demand? If so, why is the equation not something similar to:
`blk.fixed_operating_cost_constraint = pyo.Constraint(
        expr=blk.fixed_operating_cost
        == blk.factor_uv_replacement * blk.uv_replacement_cost * electricity_demand + blk.electricity_cost * electricity_demand
    )`

watertap/unit_models/tests/test_uv_aop.py
```diff
@@ -234,5 +236,29 @@ def test_solution(self, UV_frame):
...
+                "flowsheet_costing_block": m.fs.costing,
+            },
+        )
+        m.fs.costing.cost_process()
+        results = solver.solve(m)
+
+        assert results.solver.termination_condition == TerminationCondition.optimal
+
+        # Check solutions
+        assert pytest.approx(967.423, rel=1e-5) == value(m.fs.unit.costing.capital_cost)
```
hunterbarber: At 1 kg/s of feed flow, is this reasonable cost and is there any need to also test an industrial scale?

hunterbarber (commented):
General comments on costing implementation provided

watertap/costing/watertap_costing_package.py
luohezhiming: I've changed to `cost_uv_aop_bundle` as suggested, the UV system include uv reactor, lamps and other equipment. I previously thought "stack" has a similar meaning with "bundle", but I agree "bundle" is more appropriate in this case.

watertap/costing/watertap_costing_package.py
luohezhiming: Electricity is registered as another operating cost flow in the `watertap_costing_package`, and it does not belong to the fixed operating cost (should be "annual operating cost" or something else). And you are correct, the `blk.electricity_cost * electricity_demand` represents lamps and other relevant equipment capital cost which is sized by electricity demand, and the fixed operating costs are simply consumable cost of these devices.

watertap/unit_models/tests/test_uv_aop.py
luohezhiming: I've changed the value to account for an industrial scale.

watertap/costing/watertap_costing_package.py
```diff
@@ -1058,3 +1100,35 @@ def cost_by_flow_volume(blk, flow_cost, flow_to_cost):
     blk.capital_cost_constraint = pyo.Constraint(
         expr=blk.capital_cost == blk.flow_cost * flow_to_cost
     )
+
+
+def cost_uv_aop_bundle(blk, flow_cost, electricity_cost, factor_uv_replacement):
```
```suggestion
def cost_uv_aop_bundle(blk, flow_cost, lamp_cost, factor_uv_replacement):

```

watertap/costing/watertap_costing_package.py
```diff
@@ -29,6 +29,7 @@
 from idaes.models.unit_models import Mixer
 
 from watertap.unit_models import (
+    Ultraviolet0D,
```
hunterbarber: If you look across the __init__, readthedocs, and `from watertap.unit_models import` sequences of unit models, UV-AOP appears in a different location each time. There isn't a standardization in the order these appear I think.

watertap/costing/watertap_costing_package.py
```diff
@@ -436,6 +452,31 @@ def add_specific_energy_consumption(
         )
 
     # Define costing methods supported by package
+
+    @staticmethod
+    def cost_uv_aop(blk, cost_electricity_flow=True):
+        """
+        UV-AOP costing method
+
+        TODO: describe equations
```
hunterbarber: Is this actually still TODO?

watertap/costing/watertap_costing_package.py
```diff
@@ -928,6 +969,7 @@ def _compute_steam_properties(blk):
 
 # Define default mapping of costing methods to unit models
 WaterTAPCostingData.unit_mapping = {
+    Ultraviolet0D: WaterTAPCostingData.cost_uv_aop,
```
hunterbarber: Here again, seems like the order models go: copied from IDAES models, then WaterTAP unique models in the order they appear in the __init__ and readthedocs. Obviously minor but will be more relevant as this list becomes longer.

watertap/unit_models/uv_aop.py
```diff
@@ -595,12 +670,26 @@ def calculate_scaling_factors(self):
         if iscale.get_scaling_factor(self.UVA) is None:
             iscale.set_scaling_factor(self.UVA, 100)
 
+        if iscale.get_scaling_factor(self.num_of_reactors) is None:
+            iscale.set_scaling_factor(self.num_of_reactors, 100)
```
hunterbarber: What are practical values for the number of reactors? Would it always be between 1 and ~10? Not going to affect the convergence too much, since the scaled variable would then be between 100 and ~1000 respectively but can you comment on the reasoning for this sf?

hunterbarber (approved):
Left a couple comments but all were extremely minor and may not even need changed therefore I'm marking this as approved. May be good to get some discussion on the couple of comments before merging though.

watertap/costing/watertap_costing_package.py
luohezhiming: Typically, we arrange the sequence alphabetically, but since there are not so many detailed models, so people just add them following the last one. I add it at the first just for avoiding potential merging conflicts because I believe there are other detailed model PRs (e.g. yours and Kurby's). If we all add it at the last, then we will have merging conflicts. But I agree with you that we can follow the sequence in the **init** file at present.

watertap/unit_models/uv_aop.py
luohezhiming: Typically in a UV system, there are 1-5 reactors. So the sf should be 1. I copied from my last sentence and forgot to change the value.

watertap/costing/watertap_costing_package.py
luohezhiming: Deleted

watertap/costing/watertap_costing_package.py
hunterbarber: @TimBartholomew @adam-a-a Is it worth creating an issue for standardizing the order of the unit models when presented as a list in various locations? See `watertap/docs/technical_reference/unit_models/index.rst`, `watertap/watertap/unit_models/__init__.py`, and `watertap/watertap/costing/watertap_costing_package.py -> imports` as examples that all appear differently. This is extremely minor and more ease of use I would think, but may be worth converting to alphabetical as the list gets longer and as @luohezhiming said, to more consistently avoid merge conflicts with new models' PRs. Currently it seemed as though the list had the newest models at the end of the list but that has gotten twisted as the aforementioned issues occur.

bknueven: We should generally do things alphabetically. This, along with the way `black` formats the code, will very often avoid merge conflicts.

watertap/unit_models/uv_aop.py
```diff
@@ -457,12 +504,39 @@ def eq_electricity_demand_phase_comp(b, t, p, j):
...
+            self.flowsheet().config.time,
+            doc="Constraints for total electricity demand of the UV reactor.",
+        )
+        def eq_electricity_demand(b, t):
+            return (
+                b.electricity_demand[t]
+                == b.electricity_demand_minimum[
+                    t, b.config.property_package.solute_set.last()
+                ]
+            )
```
bknueven: This whole set-up doesn't make a huge amount of sense to me from a mathematical modeling perspective. What quantity should `electricity_demand` be equal to?

If it's the `max(b.electricity_demand_phase_comp[t, "Liq", :])` then we need to be doing something different.

watertap/unit_models/uv_aop.py
```diff
@@ -457,12 +504,39 @@ def eq_electricity_demand_phase_comp(b, t, p, j):
             )
 
         # TODO: add minimum electricity demand for multiple solutes
```
bknueven:
```suggestion
```

bknueven (changes requested):
I need some clarification on the model.

watertap/unit_models/uv_aop.py
luohezhiming: The `electricity_demand` is equal to `max(b.electricity_demand_phase_comp[t, "Liq", :])`. All of these set-up is to avoid using `max` function because it is not continuous and may have problems when using IDAES ipopt solver. So I follow the way in this https://github.com/IDAES/idaes-pse/blob/d1a2add96bf660681048b05ea8e90fb4b51294ee/idaes/models/unit_models/mixer.py#L632

bknueven: This implementation is a bit different than the one you point to in IDAES, which looks correct to me.

I would also change the variable name (from min to max) to be less confusing.

I can think of another implementation which might be a bit more efficient, if you have a lot of values. @andrewlee94 I wonder if we should add a helper in IDAES to deal with these recursive min/max formulations, since we have at least two examples.

andrewlee94: We could add a helper function for this, although I am always a bit concerned about making things like this too easy to use for users who don't know the issues associated with them. PRs are always welcome.

On a different topic, I note that all the doc strings refer to a minimum energy demand whilst the function itself is a max. Whilst this does make some sense in the global process level if you stop and think about it for a while, in the more local sense you are really finding the maximum of the potential electricity demands. I think it might make more sense to call this `max_component_electricity_demand` or something similar to better indicate what this represents.

bknueven: Your second comment brings up a good point -- if this is really the "minimum" energy the process could use, but it could use more, then we could use a standard inequality representation and skip using `smooth_max`.

andrewlee94: @bknueven I am not sure that is a good idea - it leaves the energy unbounded to the upper end without any additional constraints. My initial though as I was reading this was that this would give us the minimum theoretical energy required, and the unit model could then add an efficiency onto this to get the actual energy demand - this would require that this be well-defined and not an inequality.

luohezhiming: I agree that we should change the name to `max_component_electricity_demand` because bas an intermedium vairable it is easier for understanding. @bknueven we talked about the inequality representation before, but we think it might have bounding issues when doing optimization problems.

watertap/unit_models/uv_aop.py
```diff
@@ -456,13 +526,75 @@ def eq_electricity_demand_phase_comp(b, t, p, j):
...
+        def eq_max_phase_electricity_demand(b, t, p, j):
+            if p == b.config.property_package.phase_list.first():
+                return (
+                    b.max_phase_electricity_demand[t, p, j]
+                    == b.electricity_demand_phase_comp[t, p, j]
+                )
+            else:
+                return b.max_phase_electricity_demand[t, p, j] == (
+                    smooth_max(
+                        b.electricity_demand_phase_comp[
```
bknueven:
```suggestion
                        b.max_phase_electricity_demand[
```

watertap/unit_models/uv_aop.py
```diff
@@ -456,13 +526,75 @@ def eq_electricity_demand_phase_comp(b, t, p, j):
...
+        def eq_max_electricity_demand(b, t, j):
+            if j == b.config.property_package.solute_set.first():
+                return (
+                    b.max_component_electricity_demand[t, j]
+                    == b.electricity_demand_comp[t, j]
+                )
+            else:
+                return b.max_component_electricity_demand[t, j] == (
+                    smooth_max(
+                        b.electricity_demand_comp[
```
bknueven:
```suggestion
                        b.max_component_electricity_demand[
```

bknueven (dismissed):
My original comment was that the formulation which calculates the `max` is incorrect, which wasn't caught because it is not tested, and even with a test may not have been caught.

Hence my suggestion for a utility, which we could at least test/verify separately, and then use where needed. This would be especially useful if this is a modeling paradigm we need to repeat elsewhere. However, it would be enough to just correct and test it in this unit model.

luohezhiming:
> My original comment was that the formulation which calculates the `max` is incorrect, which wasn't caught because it is not tested, and even with a test may not have been caught.
> 
> Hence my suggestion for a utility, which we could at least test/verify separately, and then use where needed. This would be especially useful if this is a modeling paradigm we need to repeat elsewhere. However, it would be enough to just correct and test it in this unit model.

Thank you for catching this mistake that I did realize before! I've committed your suggested change.

luohezhiming:
@bknueven The Checks for macOS setup always directly cancelled, I see your PR #635 also cancelled this check, do you know what happened?

hunterbarber:
> @bknueven The Checks for macOS setup always directly cancelled, I see your PR #635 also cancelled this check, do you know what happened?

Looks like the following is happening with [macOS tests outlined here with scheduled brownouts](https://github.com/actions/virtual-environments/issues/5583). My PR #617 is now also encountering this issue when updating the branch although it is ready to merge and was previously passing all tests. @lbianchi-lbl Does any action need taken with this?

lbianchi-lbl:
> > @bknueven The Checks for macOS setup always directly cancelled, I see your PR #635 also cancelled this check, do you know what happened?
> 
> Looks like the following is happening with [macOS tests outlined here with scheduled brownouts](https://github.com/actions/virtual-environments/issues/5583). My PR #617 is now also encountering this issue when updating the branch although it is ready to merge and was previously passing all tests. @lbianchi-lbl Does any action need taken with this?

@hunterbarber Thanks a lot for the heads-up, I remember seeing some kind of deprecation warnings but I had no idea that the recent cancellations were due to deliberate brown-outs. I'll open a PR to update the macOS runner version.

adam-a-a:
> > My original comment was that the formulation which calculates the `max` is incorrect, which wasn't caught because it is not tested, and even with a test may not have been caught.
> > Hence my suggestion for a utility, which we could at least test/verify separately, and then use where needed. This would be especially useful if this is a modeling paradigm we need to repeat elsewhere. However, it would be enough to just correct and test it in this unit model.
> 
> Thank you for catching this mistake that I did realize before! I've committed your suggested change.

Actually, correct me if I'm wrong, but I think @bknueven was referring to his suggested changes on one of the arguments in the smooth_max function, which I don't think @luohezhiming incorporated. However, from just a glance and recalliing the original intention, I'm not sure those changes need to be made because after using smooth_max, later on, the last element which contains that actual max is grabbed and assigned to the actual energy consumption. 

Perhaps with @bknueven 's suggestion, we would not need to grab the last element -- not sure without trying myself -- but it looks that way.

adam-a-a:
> > > My original comment was that the formulation which calculates the `max` is incorrect, which wasn't caught because it is not tested, and even with a test may not have been caught.
> > > Hence my suggestion for a utility, which we could at least test/verify separately, and then use where needed. This would be especially useful if this is a modeling paradigm we need to repeat elsewhere. However, it would be enough to just correct and test it in this unit model.
> > 
> > 
> > Thank you for catching this mistake that I did realize before! I've committed your suggested change.
> 
> Actually, correct me if I'm wrong, but I think @bknueven was referring to his suggested changes on one of the arguments in the smooth_max function, which I don't think @luohezhiming incorporated. However, from just a glance and recalliing the original intention, I'm not sure those changes need to be made because after using smooth_max, later on, the last element which contains that actual max is grabbed and assigned to the actual energy consumption.
> 
> Perhaps with @bknueven 's suggestion, we would not need to grab the last element -- not sure without trying myself -- but it looks that way.

Nevermind-- I see now that it wouldn't have worked out, and although it appeared that @bknueven 's suggestions weren't adopted by looking at the conversation, that was outdated.

bknueven:
I won't hold this up anymore, but I am still concerned about this max formulation not being covered by tests.

If we merge this as-is, we should open an issue to track the uncovered code.

watertap/unit_models/tests/test_uv_aop.py
```diff
@@ -207,32 +218,60 @@ def test_solve(self, UV_frame):
...
+        assert pytest.approx(2.3e-3, rel=1e-3) == value(
+            m.fs.unit.photolysis_rate_constant["Liq", "NDMA"]
+        )
+        assert pytest.approx(1156.99, rel=1e-3) == value(
+            pyunits.convert(m.fs.unit.electricity_demand[0], to_units=pyunits.kW)
+        )
+
+    @pytest.mark.requires_idaes_solver
+    @pytest.mark.component
+    def test_costing(self, UV_frame):
```
adam-a-a, line 726: @luohezhiming Based on @bknueven 's previous suggestion, I think it'd be good to have a quick test checking that max energy consumption gets selected as intended, so maybe a separate test with two target components (two EE/Os), and assert that the expected energy consumption is set to the actual energy consumption.

adam-a-a: Not sure if you did include that somewhere and I missed it...if not, do you think you can add that?

luohezhiming: I could try to add another dummy component to the test to validate the test.

watertap/unit_models/tests/test_uv_aop.py
```diff
@@ -207,32 +220,317 @@ def test_solve(self, UV_frame):
...
+
+
+class TestUltraviolet_with_multiple_comps:
+    @pytest.fixture(scope="class")
+    def UV_frame(self):
+        m = ConcreteModel()
+        m.fs = FlowsheetBlock(default={"dynamic": False})
+
+        # m.fs.properties = props.NDMAParameterBlock()
+
```
adam-a-a:
```suggestion
```

watertap/costing/watertap_costing_package.py
```diff
@@ -113,6 +114,11 @@ def build_global_params(self):
             doc="Membrane replacement factor [fraction of membrane replaced/year]",
             units=pyo.units.year**-1,
         )
+        self.factor_uv_replacement = pyo.Var(
+            initialize=0.3993,
+            doc="UV replacement factor [fraction of uv replaced/year]",
```
adam-a-a: Is this representative of UV _lamp_ replacement?

luohezhiming: I sum up the consumable costs of lamps, sleeves, ballasts and sensors, so it accounts for all of them.

watertap/costing/watertap_costing_package.py
```diff
@@ -1317,3 +1357,37 @@ def cost_by_flow_volume(blk, flow_cost, flow_to_cost):
...
+    )
+
+    blk.capital_cost_constraint = pyo.Constraint(
+        expr=blk.capital_cost
+        == blk.reactor_cost * flow_in * blk.unit_model.num_of_reactors
+        + blk.lamp_cost * electricity_demand
+    )
+    blk.fixed_operating_cost_constraint = pyo.Constraint(
+        expr=blk.fixed_operating_cost
+        == blk.factor_uv_replacement * blk.lamp_cost * electricity_demand
```
adam-a-a: Can we rename this `factor_uv_lamp_replacement` so that it is immediately obvious?

watertap/unit_models/tests/test_uv_aop.py
```diff
@@ -160,9 +174,9 @@ def test_build(self, UV_frame):
                 assert hasattr(blk[0], obj_str)
 
         # test statistics
-        assert number_variables(m) == 35
-        assert number_total_constraints(m) == 22
-        assert number_unused_variables(m) == 0  # vars from property package parameters
+        assert number_variables(m) == 41
+        assert number_total_constraints(m) == 27
+        assert number_unused_variables(m) == 1  # vars for watertap costing package
```
adam-a-a: Which unused var is it that gets used for costing? I only ask because the first thought that comes to mind is that maybe that var should only be in the costing package but not sure.

watertap/unit_models/tests/test_uv_aop.py
```diff
@@ -192,7 +206,7 @@ def test_initialize(self, UV_frame):
     def test_var_scaling(self, UV_frame):
         m = UV_frame
         badly_scaled_var_lst = list(badly_scaled_var_generator(m))
-        [print(i[0]) for i in badly_scaled_var_lst]
+        [print(i[0].name, i[1]) for i in badly_scaled_var_lst]
```
adam-a-a: I don't have a problem with this line being here, but we typically delete these print lines since they are meant for debugging purposes and we usually clean this kind of stuff up before merging.

watertap/unit_models/tests/test_uv_aop.py
luohezhiming: The unused variable is `num_of_reactors`, since the costing data I used is for a uv system, so I add this `num_of_reactors` to account for number of reactors in a uv system.

watertap/costing/watertap_costing_package.py
luohezhiming: Yeah, I think it is a good idea to be more clear

watertap/unit_models/tests/test_uv_aop.py
luohezhiming: Deleted

watertap/unit_models/tests/test_uv_aop.py
```diff
@@ -207,32 +221,539 @@ def test_solve(self, UV_frame):
...
+        m.fs.unit.costing = UnitModelCostingBlock(
+            default={
+                "flowsheet_costing_block": m.fs.costing,
+            },
+        )
+        m.fs.costing.cost_process()
+        m.fs.costing.add_LCOW(m.fs.unit.control_volume.properties_out[0].flow_vol)
+        results = solver.solve(m)
+
+        assert results.solver.termination_condition == TerminationCondition.optimal
```
adam-a-a: You could replace this line with `assert_optimal_termination(results)`, which I think checks that `TerminationCondition.optimal` is true _and_ that solver status is OK

watertap/unit_models/tests/test_uv_aop.py
```diff
@@ -207,32 +221,539 @@ def test_solve(self, UV_frame):
...
```
adam-a-a: same as above- basically, replace with assert_optimal_termination or add assertion to check that solver status is OK

watertap/unit_models/tests/test_uv_aop.py
```diff
@@ -207,32 +221,539 @@ def test_solve(self, UV_frame):
...
+
+    @pytest.mark.component
+    def test_initialize(self, UV_frame):
+        initialization_tester(UV_frame)
+
+    @pytest.mark.component
+    def test_var_scaling(self, UV_frame):
+        m = UV_frame
+        badly_scaled_var_lst = list(badly_scaled_var_generator(m))
+        [print(i[0].name, i[1]) for i in badly_scaled_var_lst]
```
adam-a-a: same comment as before regarding print lines

watertap/unit_models/tests/test_uv_aop.py
```diff
@@ -207,32 +221,539 @@ def test_solve(self, UV_frame):
...
+        assert pytest.approx(662.018, rel=1e-3) == value(
+            pyunits.convert(
+                m.fs.unit.electricity_demand_phase_comp[0, "Liq", "DCE"],
+                to_units=pyunits.kW,
+            )
+        )
+        assert pytest.approx(1153.52, rel=1e-3) == value(
+            pyunits.convert(m.fs.unit.electricity_demand[0], to_units=pyunits.kW)
+        )
+
```
adam-a-a, line 723: Great!

watertap/unit_models/tests/test_uv_aop.py
```diff
@@ -207,32 +221,539 @@ def test_solve(self, UV_frame):
...
+        m.fs.unit.costing = UnitModelCostingBlock(
+            default={
+                "flowsheet_costing_block": m.fs.costing,
+            },
+        )
+        m.fs.costing.cost_process()
+        m.fs.costing.add_LCOW(m.fs.unit.control_volume.properties_out[0].flow_vol)
+        results = solver.solve(m)
+
+        assert results.solver.termination_condition == TerminationCondition.optimal
```
adam-a-a: same comment as before regarding solver status OK

watertap/unit_models/uv_aop.py
```diff
@@ -279,6 +278,20 @@ def build(self):
             units=units_meta("time"),
             doc="Exposure time of UV light.",
         )
+        self.reactor_volume = Var(
+            initialize=1,
+            bounds=(1e-18, 10000),
```
adam-a-a: Can you change all these small lower bounds (1e-18) to 0 and ensure solver convergence is still achieved?

watertap/costing/watertap_costing_package.py
adam-a-a:
```suggestion
            doc="UV replacement factor accounting for lamps, sleeves, ballasts and sensors [fraction of uv replaced/year]",

```

watertap/unit_models/tests/test_uv_aop.py
adam-a-a: Does this represent number of reactors in series, parallel, or both? If parallel, would the var end up being used in the model without costing?

watertap/unit_models/uv_aop.py
luohezhiming: Yeah, I've revised all these bounds to `(0,None) `and all tests are passed.

watertap/unit_models/tests/test_uv_aop.py
luohezhiming: It represent reactors in parallel. This var is onlly used for costing the uv reactor.

luohezhiming: I think we could keep this `num_of_reactors` as a unit model variable because we could use it for the uv dose expression (related to UVT, flow rate and number of banks).

watertap/costing/watertap_costing_package.py
```diff
@@ -1317,3 +1357,37 @@ def cost_by_flow_volume(blk, flow_cost, flow_to_cost):
...
+    )
+
+    electricity_demand = pyo.units.convert(
+        blk.unit_model.electricity_demand[0], to_units=pyo.units.kW
+    )
+
+    blk.capital_cost_constraint = pyo.Constraint(
+        expr=blk.capital_cost
+        == blk.reactor_cost * flow_in * blk.unit_model.num_of_reactors
+        + blk.lamp_cost * electricity_demand
```
adam-a-a: I'm a little unsure about this. This seems like the control volume for your UV model only accounts for 1 reactor, so you multiply by number of (parallel) reactors to accommodate a system with higher capacity.... but then the lamp_cost is multiplied by power demand, which I would expect to increase as well with number of reactors. Moreover, couldn't our control volume represent the whole system of reactors without having to ever specify number of reactors?

luohezhiming: This is a pretty good point! Actually different uv reactors have different numbers of lamps and other equipments, but as Tim suggested, I'd better not add so many variablles such as number of lamps, etc. I think for a simple solution, we could delete this `num_of_reactors` variable and use our control volume to represent the whole uv system as you suggested.

watertap/unit_models/uv_aop.py
```diff
@@ -456,13 +526,75 @@ def eq_electricity_demand_phase_comp(b, t, p, j):
...
+                    == b.electricity_demand_phase_comp[t, p, j]
+                )
+            else:
+                return b.max_phase_electricity_demand[t, p, j] == (
+                    smooth_max(
+                        b.max_phase_electricity_demand[
+                            t, b.config.property_package.phase_list.prev(p), j
+                        ],
+                        b.electricity_demand_phase_comp[t, p, j],
+                        b.eps_electricity,
```
adam-a-a, line 541: I suppose we are missing a test to check that the solution works as intended with more than one phase. Nevertheless, I think this was already an additional capability that you didn't need to add, so maybe we can just make an issue to keep track of any UV-related things we want to revisit in future PRs for later improvement. In this case, we can note that a test should be added to check electricity demand for a multi-component and multi-phase example.

watertap/costing/watertap_costing_package.py
adam-a-a: OK, then it seems this is the only further change needed before merging, considering that you already addressed my other comments.

adam-a-a (changes requested):
Excellent job on this. I think we should be ready to merge this once we remove the `num_of_reactors` being used in the capital cost constraint. Please also note things that we might want to consider in future PRs by adding to an issue for UV. We can have one master issue for all things related to UV.

watertap/costing/watertap_costing_package.py
adam-a-a: _Unless_ the number of reactors accounts for redundant reactors that aren't used all at the same time. Then I could understand why only 1 reactor is considered for energy demand, while 3 (in your test examples) are considered for capital cost, which could represent 1 operational reactor and 2 redundant.

watertap/unit_models/uv_aop.py
luohezhiming: I don't know what kind of waterborne pathogens is solid or gas phase, maybe we could add a dummy solute for verifying the test? e.g. solid `DCE`

adam-a-a: We could try something like that. Another option could be to use the IDAES generic property framework to define liquid and solid. Whichever path we take, I think this test can be reserved for a subsequent PR and noted in your master issue for UV, but if you think it'd be quick and easy for you to implement, I won't stop you!

watertap/costing/watertap_costing_package.py
luohezhiming: @adam-a-a in the given example, there are 3 parallel reactors and 1 is redundant, there are 15 average number of lamps working over the simulation. Overall, the energy demand depends on the working lamps but the capital cost includes the redundant reactor, lamps and other equipments.

luohezhiming:
> Excellent job on this. I think we should be ready to merge this once we remove the `num_of_reactors` being used in the capital cost constraint. Please also note things that we might want to consider in future PRs by adding to an issue for UV. We can have one master issue for all things related to UV.

@adam-a-a I've deleted the `num_of_reactors` var and addressed all other comments. I will create a master issue for further refinement.

adam-a-a (approved):
Nice job!

PR #626, opened 2022-07-18
GAC scaling revisions and other general enhancements

### LINKED ISSUE #625: Revisit GAC scaling and other issues
Corresponding PR #626 
It was discovered in #559 that the scaling implemented for the GAC model may be too simplified. 

- [x] Revisit the scaling values used for the model variables
- [x] ~~Add scaling for model constraints~~

Other issues include:

- [x] Tolerance used for testing simplified solution is way too loose
      - Currently using a tolerance of 0.1 for comparison. This needs to be at least 1e-3. 
      - Line 207 in `test_solution_simplified`
  
- [x] Remove the `within` keyword arg for Param `saturation_mtz_upstream`
      - Not proper Param definition in Pyomo
      - Line 251 from `gac.py`

hunterbarber:
## Fixes/Resolves:
Minor formatting and nomenclature corrections, corresponding to #625  

## Summary/Motivation:
Improvements to close the initial GAC model

- Minor modeling convention fixes
- Scaling improvements
- Add `def _get_performance_contents` and `def _get_stream_table_contents`
- Addressing multiple components

### Legal Acknowledgement


hunterbarber:
This is up as a draft to coincide with progress marked as complete on #625

watertap/unit_models/gac.py
```diff
@@ -222,7 +223,7 @@ def eq_isothermal(b, t):
             self.flowsheet().config.time,
             doc="Isothermal assumption for absorbed contaminant",
         )
-        def eq_isothermal_adsorbate(b, t):
+        def eq_isothermal_adsorbed_contam(b, t):
```
adam-a-a, line 250: While we're back here, these constraints should have doc strings.

watertap/unit_models/gac.py
```diff
@@ -248,7 +249,7 @@ def eq_isobaric_adsorbate(b, t):
         self.saturation_mtz_upstream = Param(
             default=0.95,
             initialize=0.95,
-            within={0.95, 0.99},
+            domain={0.95, 0.99},
```
adam-a-a, line 276: I guess you can do this, but I think we'd typically use `bounds=(0.95, 0.99)` in this case.

hunterbarber: This is a convention with defining the mass transfer zone upstream edge as typically 0.95 or 0.99 exclusively, but I just looked in literature and there are rare cases of 0.98, so I think converting to bounds is valid.

watertap/unit_models/gac.py
```diff
@@ -1058,52 +1058,78 @@ def initialize_build(
...
+            },
+            time_point=time_point,
+        )
 
     # ---------------------------------------------------------------------
     def calculate_scaling_factors(self):
         super().calculate_scaling_factors()
 
-        # overwrite default scaling
+        # scale based on molar flow traditionally provided by user for building flowsheets
```
adam-a-a, line 1188: What if a property package defines ions and nonpolar solutes? `solute_set` might not pick everything up when using the ion DSPMDE prop model. I am pretty sure that, hypothetically, if you had multiple components in your feed comprising ions and uncharged solutes, the ions would be listed under `ion_set`, while uncharged under `solute_set`. I'm not familiar enough with your model yet, but if you anticipate ions AND uncharged solutes ever being in this solute list, you should/could include something along the lines of
`solute_set = self.config.property_package.solute_set | self.config.property_package.ion_set`
and then loop across `solute_set`.

watertap/unit_models/gac.py
hunterbarber: @adam-a-a I'm confused here, do they not already have doc strings? Maybe some clarification would help

watertap/unit_models/gac.py
hunterbarber: @adam-a-a Actually looks like since Param() doesn't have a bounds argument, the solution would have to be 
```suggestion
            domain=NonNegativeReals,
```
with the doc string just stating the conventional values

adam-a-a: Ah, I overlooked the fact that this was a Param. In that case, no need to change to `NonNegativeReals` domain.

watertap/unit_models/tests/test_gac.py
```diff
@@ -367,7 +364,96 @@ def test_solution_robust(self, gac_frame_robust):
...
+            ub_scale=1e3,
+            tol=1e2,
+            zero=1e-10,
+        ):
+            test_scale = [lb_scale, ub_scale]
+            var_hist = {}
+            for scale in test_scale:
+                temp_blk = blk.clone()
+                # loop through variables and scale previously fixed inlet flow
+                for vt, vm in temp_blk.component_data_iterindex(
```
adam-a-a: Interesting- I haven't used this function yet. I'm wondering if there was a reason why you used this as opposed to `component_data_objects`.

watertap/unit_models/tests/test_gac.py
```diff
@@ -367,7 +364,96 @@ def test_solution_robust(self, gac_frame_robust):
...
+                        )
+                        temp_blk.fs.properties.set_default_scaling(
+                            vt[0], dsf * scale**-1, index=vt[1]
+                        )
+                        vm.fix(vm * scale)
+                calculate_scaling_factors(temp_blk)
+                temp_blk.fs.unit.initialize(outlvl=idaes.logger.ERROR)
+                results = solver.solve(temp_blk)
+                if (
+                    results.solver.termination_condition
```
adam-a-a: FYI- we should have a `check_optimal_termination` function that handles this (`from pyomo.environ import check_optimal_termination`)

watertap/unit_models/tests/test_gac.py
```diff
@@ -367,7 +364,96 @@ def test_solution_robust(self, gac_frame_robust):
...
+                    is not TerminationCondition.optimal
+                    or results.solver.status is not SolverStatus.ok
+                ):
+                    print("Failed run on", scale, "scale")
```
adam-a-a: Should we use logging here instead?

watertap/unit_models/tests/test_gac.py
```diff
@@ -367,7 +364,96 @@ def test_solution_robust(self, gac_frame_robust):
...
+        m.fs.unit.report()
+
+    @pytest.mark.component
+    def test_variable_sensitivity_robust(self, gac_frame_robust):
+        m = gac_frame_robust
+
+        import idaes.core.util.scaling as iscale
+        from pyomo.common.collections import ComponentSet
+
+        def var_sens_generator(
```
adam-a-a: This looks like something that would need its own testing. I also think it would benefit from more comments to make it easier to follow. We could think about adding a utility under `water.core.util.scaling`, then bring the function in here, but as mentioned, you'd also need a test file for the new scaling function.

watertap/unit_models/tests/test_gac.py
```diff
@@ -478,3 +562,80 @@ def test_costing_max_gac_ref_robust(self, gac_frame_robust):
...
+        m.fs.unit.tort.fix(1)
+        m.fs.unit.spdfr.fix(1)
+        m.fs.unit.sphericity.fix(1.5)
+        m.fs.unit.a0.fix(0.8)
+        m.fs.unit.a1.fix(0)
+        m.fs.unit.b0.fix(0.023)
+        m.fs.unit.b1.fix(0.793673)
+        m.fs.unit.b2.fix(0.039324)
+        m.fs.unit.b3.fix(0.009326)
+        m.fs.unit.b4.fix(0.08275)
```
adam-a-a: Regarding the Stanton and throughput equation parameters, how are these determined? That is, what are they primarily dependent on?

adam-a-a (commented):
It looks like things are on the right track. I have a few comments and questions.

watertap/unit_models/tests/test_gac.py
hunterbarber: `component_data_objects` outputs `vm` in this function. However this variable object is not directly application with                         `set_default_scaling(attrbute, value, index=None)` because the components are indexed individually and at different values. For example, `component_data_iterindex` will output the following `('flow_mol_phase_comp', ('Liq', 'H2O')) fs.unit.process_flow.properties_out[0.0].flow_mol_phase_comp[Liq,H2O]`. If vm or the `component_data_objects` was solely available, you could not identify the correct variable in the default scaling function as `temp_blk.fs.properties.set_default_scaling("flow_mol_phase_comp", dsf * scale**-1, ('Liq', 'H2O'))`. I found this function to be easiest to gain the variable and index in the appropriate string form.

watertap/unit_models/tests/test_gac.py
hunterbarber: There are niches with this single use case that may need redefined if adding to `water.core.util.scaling`, say testing the handling vol/mass/mole flow and other compatibilities. I think if this does get migrated and more work done testing and commenting are sure to be included.

andrewlee94: It does strike me as odd that you are defining this method inside a test. I would have thought that this is either something more generally useful that would warrant a utility function, or something that can be achieved in other ways.

watertap/unit_models/tests/test_gac.py
andrewlee94: @hunterbarber Could you clarify exactly what you are trying to do here? Without comments or documentation, I am struggling to understand what is going on (especially seeing this is part of a test).

watertap/unit_models/tests/test_gac.py
andrewlee94: Seeing as this is a test, you should generally not be logging or printing anything, as this tends to get eaten by `pytest`.

adam-a-a: I suppose I was looking at it from the perspective of this _potentially_ becoming a utility function under `watertap.core.util` somewhere. In that case, I was thinking that logging might be preferred.

watertap/unit_models/tests/test_gac.py
adam-a-a: I think @hunterbarber intended to get eyes on this function before taking the next step of adding it as a utility function. So we should first make sure we know everything that's happing in the function and make a determination on whether it should be added in the first place. Thus, more comments describing what is happening throughout the function would be helpful in proceeding.

watertap/unit_models/tests/test_gac.py
adam-a-a: The reason I ask is because I wonder if these should be kept as Vars in the unit model and then fixed, like we do with certain parameters in property models. Then these default values would be used as the initial vals. The user could then overwrite them as needed by fixing new values to them (or potentially using parameter estimation down the line). Not sure if we'd want to take that route or not as I anticipate that these parameters are highly specific, but it could help reduce the number of vars that need to be fixed by the user that might not have this information at the ready.

watertap/unit_models/tests/test_gac.py
hunterbarber: So where `badly_scaled_var_generator` outputs which variables are outside of a properly scaled tolerance for a solved frame, it allows for the model dev to set scaling factors in the `calculate_scaling_factors` function to be well posed for the data frames used in testing. Though, I found this mainly lends itself to setting constant scaling factors in initial model development with whatever process scale are used in the tests. What this method intended to do was to take a previously defined and solved model frame in testing that has fixed inlet flow and default scaling factors. Then, it would downscale and upscale the process inlet flowrate and default scaling factors to approximate small-scale and large-scale limitations of the process. By varying the process scale and reporting values and scaled values (with initial constant scaling factors) of all model variables, the tool can show variables that are linearly or nonlinearly proportional (or inversely proportional) to the inlet flow. These are then good candidates to have non-constant scaling factors or user defined scaling factors in `calculate_scaling_factors`. Specifically, I had some scaled variables that changed between 100 and 0.01 in different model frames that weren't caught by `badly_scaled_var_generator` but were caught by this tool by changing in orders of magnitude across different process scales.

watertap/unit_models/gac.py
hunterbarber: The only thing I have left is handling multi-component property packs and I can consider this when making the edits. I still have the consideration that for GAC it is easy to define the saturation capacity of the particle and adsorption rate of the single solute. When multiple solutes are present, you can still reasonably determine the saturation capacity of the particle and adsorption rate of the target solute, where the background solutes reduce both values. The decision is to then treat the backgrounds as inert where in = out with reductions to saturation capacity of the particle and adsorption rate, or to also determine the saturation capacity of the particle and adsorption rate of the background solute(s) which gets a little more complicated. These background solutes would then also selectively adsorb into the particle and be contained in the waste stream.

watertap/unit_models/tests/test_gac.py
hunterbarber: Yes, printing is mostly for debugging in this draft status.

watertap/unit_models/tests/test_gac.py
andrewlee94: First, this sounds like it needs to be a separate function with a set of dedicated tests to ensure it does what is expected. This might also be something that should go in IDAES rather than WaterTap.

Secondly however, I would be weary about using this in a test (unless it is a test for the method) - tests are generally best written in concrete terms, where you have control over as much as possible.

watertap/unit_models/gac.py
adam-a-a: That sounds fine, but my main point was that-- if I were to pass in even a single ion via the ion prop model (or IDAES generic prop model), though it wouldn't make sense in terms of electroneutrality, that ion would not be captured by `self.config.property_package.solute_set`. That is because when a component is defined as either Cation or Anion type, which would happen when a non-zero charge is provided for a component, that component would be listed under `self.config.property_package.ion_set`. Neutral components (zero charge) are listed in `self.config.property_package.solute_set`.

watertap/unit_models/tests/test_gac.py
hunterbarber: Both a#s and b#s are functions of 1/n (fixed and highly specific to the contaminants) and Bi (calculated) already present in the model. These parameters are tabulated alongside the constant pattern model development (Hand et al., 1984) which is relatively available. In literature, parameter values are referenced and linearly interpolated as needed to get their value. Concerning their nonlinearity with 1/n and Bi, I am not sure of the goodness of fit to any simple regression formula which would remove the need to fix and introduce new equations.

watertap/unit_models/gac.py
hunterbarber: Yes, I gotcha. If treating all others as inert, it may even be,

`target_component = "TCE"`
`inert_set = self.config.property_package.component_list.remove(target_component)`

Depends on how indexing and scaling is easiest to implement wrt the sets.

watertap/unit_models/gac.py
adam-a-a: False alarm. =)

hunterbarber (commented):
> @adam-a-a @TimBartholomew we mentioned the potential of having a .yaml for [this spreadsheet (gac.rst, line 341)](https://github.com/watertap-org/watertap/pull/626/files#diff-80399b21c78362c74b0c4bbc0382355b2f8c1f0f690baa638ca04079c42e8e36R341) for its place in time with parameters from 2021 data, do you find this still necessary?

Addressed by having .xls in box folder for future reference and citation in documentation.

docs/technical_reference/unit_models/gac.rst
```diff
@@ -1,12 +1,8 @@
 Granular Activated Carbon (GAC)
 ===============================
-.. note::
-
-    The GAC model is an ongoing project and subject to change, as is this documentation.
-
 This is an empirical-performance-based granular activated carbon (GAC) model that works under the following criteria and assumptions:
```
adam-a-a:
```suggestion
This is an empirical, performance-based granular activated carbon (GAC) model that works under the following criteria and assumptions:
```

docs/technical_reference/unit_models/gac.rst
```diff
@@ -1,12 +1,8 @@
...
 ===============================
-.. note::
-
-    The GAC model is an ongoing project and subject to change, as is this documentation.
-
-This is an empirical-performance-based granular activated carbon (GAC) model that works under the following criteria and assumptions:
+This is an empirical, performance-based granular activated carbon (GAC) model that works under the following criteria and assumptions:
    * supports a single liquid phase only
-   * supports a single solute and single solvent (water) only
+   * supports adsorption of a single solute species only, other species are considered inert
```
adam-a-a:
```suggestion
   * supports adsorption of a single solute species only while other species are considered inert
```

watertap/unit_models/gac.py
```diff
@@ -851,7 +867,7 @@ def eq_approx_saturation(b):
                 doc="Molecular diffusion coefficient",
             )
 
-            # TODO: Determine whether the LeBas method can be implemented
+            # TODO: Determine whether the LeBas method can be implemented or embed in prop pack
```
adam-a-a, line 870: Yes, this is something that should be in a property model, but it can be resolved in a future PR. If this isn't in an issue, please add it.

watertap/unit_models/gac.py
```diff
@@ -1058,52 +1075,186 @@ def initialize_build(
...
+        ] = self.gac_mass_replace_rate
+        var_dict["Liquid phase film transfer coefficient"] = self.kf
+        var_dict["Surface diffusion coefficient"] = self.ds
+        var_dict["Solute distribution parameter"] = self.dg
+        if hasattr(self, "diffus_liq"):
+            var_dict["Molecular diffusion coefficient"] = self.diffus_liq
+        if hasattr(self, "molal_volume"):
+            var_dict["Molal volume"] = self.molal_volume
+
+        # loop throuhg desired state block properties indexed by [phase, comp]
```
adam-a-a:
```suggestion
        # loop through desired state block properties indexed by [phase, comp]
```

watertap/unit_models/gac.py
```diff
@@ -1058,52 +1075,186 @@ def initialize_build(
...
+                ):
+                    var_dict[f"{prop_str} of {j} @ process feed outlet"] = getattr(
+                        self.process_flow.properties_out[time_point], prop_name
+                    )["Liq", j]
+                if self.adsorbed_contam[time_point].is_property_constructed(prop_name):
+                    var_dict[f"{prop_str} of {j} @ process feed outlet"] = getattr(
+                        self.adsorbed_contam[time_point], prop_name
+                    )["Liq", j]
+
+        # loop throuhg desired state block properties indexed by [phase]
```
adam-a-a:
```suggestion
        # loop through desired state block properties indexed by [phase]
```

watertap/unit_models/gac.py
```diff
@@ -1058,52 +1075,186 @@ def initialize_build(
...
     def calculate_scaling_factors(self):
         super().calculate_scaling_factors()
 
-        # overwrite default scaling
-        for j in self.config.property_package.solute_set:
-            sf = 100 * iscale.get_scaling_factor(
+        # scale based on molar flow traditionally provided by user for building flowsheets
+        for j in self.config.target_species:
+            sf_solute = iscale.get_scaling_factor(
+                self.process_flow.properties_in[0].flow_mol_phase_comp["Liq", j]
```
adam-a-a: We might want to add a warning and default here when the user does not provide the scaling factor.

watertap/unit_models/tests/test_gac.py
```diff
@@ -60,421 +56,604 @@
...
+            assert len(port.vars) == 3  # number of state variables for property package
             assert isinstance(port, Port)
 
         # test statistics
-        assert number_variables(m) == 77
-        assert number_total_constraints(m) == 45
-        assert number_unused_variables(m) == 10  # dens parameters from properties
+        assert number_variables(ms) == 77
+        assert number_total_constraints(ms) == 44
+        assert number_unused_variables(ms) == 10  # dens parameters from properties
```
adam-a-a, line 156: I noticed this is missing in some of our other unit model tests, but I think you should add an `assert_units_consistent(ms)` here.

watertap/unit_models/tests/test_gac.py
```diff
@@ -60,421 +56,604 @@
...
         assert results.solver.status == SolverStatus.ok
 
-    @pytest.mark.requires_idaes_solver
+    @pytest.mark.component
+    def test_var_scaling_solve_simplified(self, gac_frame_simplified):
+        ms = gac_frame_simplified
+        badly_scaled_var_lst = list(
+            badly_scaled_var_generator(ms, large=1e2, small=1e-2, zero=1e-8)
+        )
+        [print(i[0].name, i[0].value, i[1]) for i in badly_scaled_var_lst]
```
adam-a-a: Can delete these print lines

watertap/unit_models/gac.py
hunterbarber: I think molal_volume may be similarly added to the property model. I will create the issue now.

watertap/unit_models/tests/test_gac.py
```diff
@@ -60,421 +56,604 @@
...
     @pytest.mark.component
-    def test_var_scaling_robust(self, gac_frame_robust):
-        m = gac_frame_robust
-        badly_scaled_var_lst = list(badly_scaled_var_generator(m))
+    def test_var_scaling_init_robust(self, gac_frame_robust):
+        mr = gac_frame_robust
+        badly_scaled_var_lst = list(
+            badly_scaled_var_generator(mr, large=1e2, small=1e-2, zero=1e-8)
+        )
+        [print(i[0].name, i[0].value, i[1]) for i in badly_scaled_var_lst]
```
adam-a-a: delete print line

watertap/unit_models/gac.py
```diff
@@ -172,6 +173,21 @@ class GACData(UnitModelBlockData):
         **SurfaceDiffusionCoefficientType.calculated** - calculates surface diffusion coefficient}""",
         ),
     )
+    CONFIG.declare(
+        "target_species",
+        ConfigValue(
+            default=None,
+            domain=set,
```
adam-a-a, line 180: Not sure what difference it would make here, but why use `domain=set` instead of `domain=dict`?

adam-a-a (changes requested):
Just a couple of subtle comments to address and this should be ready to go.

watertap/unit_models/gac.py
hunterbarber: Here, I just mirrored how the sets are configured in the property package such as solute set, and wanted to be able to use set operators in later code (see inert_species). In future model development I could see the whole solute_set or ion_set getting passed into the target_species which I think would make this a more seamless transition to higher fidelity models.

TimBartholomew (approved):
LGTM

watertap/unit_models/gac.py
```diff
@@ -1058,52 +1075,190 @@ def initialize_build(
 
     # ---------------------------------------------------------------------
 
-    # TODO: def _get_performance_contents(self, time_point=0):
-    # TODO: def _get_stream_table_contents(self, time_point=0):
+    def _get_performance_contents(self, time_point=0):
+        var_dict = {}
+
```
luohezhiming, line 1080: Typically what variables should we consider as `performance_contents`? I see you almost list all vairables in your model, but maybe some variables are specific inputs but not performance indices? Since this is also my TODO list of  UV model, I want to confirm with you.

luohezhiming (approved):
Looks great, I only have a trivial question on performance_contents.

watertap/unit_models/gac.py
hunterbarber: I chose variables that are relevant for interpreting the results of the model. This includes variables that are typically fixed, indicative of the size of the unit, indicative of the performance, or other important variables. Just so happened that this turned out to be a lot, but there are some lesser relevant variables that are not included.

adam-a-a: @luohezhiming You can report any variable, expression, and/or parameter values that you think would be of interest. These would then be displayed when using the `report()` method (e.g., `m.fs.unit.report()`).

adam-a-a (approved):

PR #627, opened 2022-07-18
[WIP] ADM1 Model

adam-a-a:
## Summary/Motivation:
Implement Anaerobic Digestion Model No. 1 (ADM1) for detailed wastewater treatment processes. 

Checklist:

- [x] add Petersen matrix values
- [x] add process rate equations
- [x] add inhibition functions, which are included in process rate equations
- [x] verify Petersen matrix values
- [x] Add quasi-steady-state components and related constraints and verify for a known state
- [ ] If possible without VLE piece, verify rate equations 
- [ ] Add the necessary properties for VLE to the thermo package
- [ ] Build the anaerobic digestor unit model and verify the whole system

- 
-

### Legal Acknowledgement


adam-a-a:
@andrewlee94 @luohezhiming I will be committing to this model bit by bit. Right now, I have some version of the property model in place and getting to work on the reaction model (added stoich parameters and stoichiometric coefficients for a couple of the reactions so far). Feel free to check out what I have so far.

watertap/property_models/anaerobic_digestion/adm1_properties.py
```diff
@@ -0,0 +1,346 @@
...
+        self.S_I = Solute(doc="Soluble inerts")
+
+        self.X_c = Solute(doc="Composites")
+        self.X_ch = Solute(doc="Carbohydrates")
+        self.X_pr = Solute(doc="Proteins")
+        self.X_li = Solute(doc="Lipids")
+        self.X_su = Solute(doc="Sugar degraders")
+        self.X_aa = Solute(doc="Amino acid degraders")
+        self.X_fa = Solute(doc="Long chain fatty acid (LCFA) degraders")
+        self.X_c4 = Solute(doc="Valerate degraders")
```
luohezhiming:
```suggestion
        self.X_c4 = Solute(doc="Valerate and butyrate degraders")
```

watertap/property_models/anaerobic_digestion/adm1_reactions.py
```diff
@@ -0,0 +1,904 @@
...
+        # R14: Decay of X_aa
+        # R15: Decay of X_fa
+        # R16: Decay of X_c4
+        # R17: Decay of X_pro
+        # R18: Decay of X_ac
+        # R19: Decay of X_h2
+
+        self.rate_reaction_idx = pyo.Set(initialize=[f"R{i}" for i in range(1, 20)])
+
+        # Stoichiometric Parameters (Table 6.1 in Batstone et al., 2002)
```
luohezhiming, line 154: Where is this citation `Batstone et al., 2002`? It can be added to the reference.

watertap/property_models/anaerobic_digestion/adm1_reactions.py
```diff
@@ -0,0 +1,904 @@
...
+            domain=pyo.PositiveReals,
+            doc="Acetate from amino acids",
+        )
+        # Kinetic Parameters
+
+        # Reaction Stoichiometry
+        # This is the stoichiometric part of the Peterson matrix in dict form.
+        # Note that reaction stoichiometry is on a mass basis.
+        # See Table 3.1 in Batstone et al., 2002.
+        mw_n = 14 * pyo.units.kg / pyo.units.kmol
```
luohezhiming: Why add this molar weight of nitrogen here?

luohezhiming (commented):
Excellant work, will continue reviewing it.

watertap/property_models/anaerobic_digestion/adm1_reactions.py
adam-a-a: Didn't get to that yet- old reference for asm1 will be replaced

watertap/property_models/anaerobic_digestion/adm1_reactions.py
adam-a-a: This was leftover from ASM1. I didn't delete, just in case it ends up being necessary for the next set of things that I'm adding. If unneeded after that, will remove.

luohezhiming: OK. That makes sense, it's under developing, I'll focus on double checking those parameter values and expressions

watertap/property_models/anaerobic_digestion/adm1_properties.py
```diff
@@ -0,0 +1,346 @@
...
+        self.X_aa = Solute(doc="Amino acid degraders")
+        self.X_fa = Solute(doc="Long chain fatty acid (LCFA) degraders")
+        self.X_c4 = Solute(doc="Valerate and butyrate degraders")
+        self.X_pro = Solute(doc="Propionate degraders")
+        self.X_ac = Solute(doc="Acetate degraders")
+        self.X_h2 = Solute(doc="Hydrogen degraders")
+        self.X_I = Solute(doc="Particulate inerts")
+
+        # TODO: Additional components not in Table but referred to in text
+        self.S_cat = Solute(doc="Total cation equivalents concentration")
```
andrewlee94: I haven't read the papers but I wonder if these should really be treated as components, or if they would be better as separate variables on the state blocks.

The answer would lie in how these are calculated - if they are calculated by unit level mass balances (i.e. mass is conserved between states, subject to reaction terms) then they should be components. If however they are calculated based off other concentration (e.g. to maintain a charge balance, which I vaguely recall) then they should not be components (as that would over-specify the problem).

watertap/property_models/anaerobic_digestion/adm1_reactions.py
```diff
@@ -0,0 +1,1433 @@
...
+            "X_c4": 0.0313,
+            "X_pro": 0.0313,
+            "X_ac": 0.0313,
+            "X_h2": 0.0313,
+            "X_I": 0.03,  # varies
+            # "S_an": 0,
+            # "S_cat": 0,
+        }
+        # TODO: Ci might need to go to the property model instead to be used for conc_mol_phase_comp
+        self.Ci = pyo.Var(
```
andrewlee94, line 139: This looks like it belongs here - the thermo properties are only concerned about actual components, not elemental concentrations (until someone want's to do elemental balances, but then you wouldn't use these reactions).

watertap/property_models/anaerobic_digestion/adm1_reactions.py
```diff
@@ -0,0 +1,1433 @@
...
+        self.pH = pyo.Expression(rule=rule_pH, doc="pH of solution")
+
+    def _I_IN_lim(self):
+        def rule_I_IN_lim(b):
+            # TODO: revisit molar concentration of S_IN as it is not in property model as of yet
+            return 1 / (
+                1 + b.params.K_S_IN / b.state_ref.conc_mol_phase_comp["Liq", "S_IN"]
+            )
+
+        self.I_IN_lim = pyo.Expression(rule=rule_I_IN_lim)
```
andrewlee94: Some doc strings on these might help users (and readers of the code like me) understand what these terms are.

watertap/property_models/anaerobic_digestion/adm1_properties.py
adam-a-a: There are other components (i.e., H+, NH3, NH4, CO2, etc.) that end up being used in a charge balance an elemental balances from what I recall. I've been thinking about how to incorporate those as I knew I wanted to avoid adding them as components. For those cases, should they just be their own variables/expressions (e.g., `conc_mol_H_+`, `conc_mol_NH3`, etc.) on the property model, and should I make them on-demand properties?

As for S_cat and S_an, so far, I think these should be components. Additionally, in the example in the Rossen & Jeppsson reference, they specify inlet concentrations and these two are specified as well (though there is no change between inlet/outlet concentrations in the example, there could be in a different case).

watertap/property_models/anaerobic_digestion/adm1_reactions.py
adam-a-a: So this ties to another issue related to the Petersen matrix. All components are on a kg COD/m3 basis, except for inorganic carbon and inorganic nitrogen, which are in units kmol C/m3 and kmol N/m3, respectively. This `Ci` converts the mass concentration to molarity for inorganic carbon, and I've been debating with myself on the _best_ way to handle the discrepancy in units while staying true to the form taken in the literature.

watertap/property_models/anaerobic_digestion/adm1_reactions.py
adam-a-a: Will do.

watertap/property_models/anaerobic_digestion/adm1_properties.py
andrewlee94: That sound like the anions and cations should be components then. I seem to recall the specific ions are calculated based on other species in the system (e.g. NH4+ is calculated from NH3) and OH- and H+ are used to "fill in the remainder". If that is the case, then those should all be property level variables/expressions. I am not sure whether they should be build-on-demand or not - that really depends on how often they are _not_ required.

watertap/property_models/anaerobic_digestion/adm1_reactions.py
andrewlee94: Look at the ASM1 and ASM2d example - they have similar issues with alkalinity. You will see that there is a hard coded conversion for molecular weight of C and N in those models to handle the conversion.

adam-a-a: Yes, I saw that. In my case, I think I could just omit the conversion that takes place when multiplying the corresponding stoichiometric coefficients by `Ci`, which is kmol C/kg COD.

andrewlee94: I kept the conversion to keep the model (and hence parameter and stoichiometric values) in-line with the published material.

PR #631, opened 2022-07-21
Add ability to specify a separate probe function within parameter sweep

k1nshuk:
## Fixes/Resolves:

This PR adds an option to specify a light weight test function to check whether the pyomo model can be initialized and solved with the current values.

## Summary/Motivation:

Larger and more complicated Watertap models can take a long time to initialize and then solve. It is possible to cheaply check whether the model within the parameter sweep loop can be solved without attempting to initialize and solve the full model.

- Create a separate function called _param_sweep_kernel which contains the kernel being called in the parameter sweep loop.
- Add an if statement that checks whether a test_function was specified or if it returns a True value.

### Legal Acknowledgement

### DIFF
--- a/watertap/tools/parameter_sweep.py
+++ b/watertap/tools/parameter_sweep.py
@@ -770,13 +770,16 @@ def _read_output_h5(filepath):
 
 # ================================================================
 
-def _param_sweep_kernel(model, optimize_function,
+
+def _param_sweep_kernel(
+    model,
+    optimize_function,
     optimize_kwargs,
     reinitialize_before_sweep,
     reinitialize_function,
     reinitialize_kwargs,
-    reinitialize_values
-    ):
+    reinitialize_values,
+):
 
     run_successful = False  # until proven otherwise
 
@@ -821,7 +824,9 @@ def _param_sweep_kernel(model, optimize_function,
 
     return run_successful
 
+
 # ================================================================
+
 
 def _do_param_sweep(
     model,
@@ -866,12 +871,16 @@ def _do_param_sweep(
 
         run_successful = False  # until proven otherwise
 
-        if test_function is None:
-            run_successful = _param_sweep_kernel(model, optimize_function, optimize_kwargs, reinitialize_before_sweep,
-                reinitialize_function, reinitialize_kwargs, reinitialize_values)
-        elif test_function(model):
-            run_successful = _param_sweep_kernel(model, optimize_function, optimize_kwargs, reinitialize_before_sweep,
-                reinitialize_function, reinitialize_kwargs, reinitialize_values)
+        if test_function is None or test_function(model):
+            run_successful = _param_sweep_kernel(
+                model,
+                optimize_function,
+                optimize_kwargs,
+                reinitialize_before_sweep,
+                reinitialize_function,
+                reinitialize_kwargs,
+                reinitialize_values,
+            )
 
         # Update the loop based on the reinitialization
         _update_local_output_dict(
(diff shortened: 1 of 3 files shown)


avdudchenko (approved):
This looks good to me. Thanks for adding it in.

watertap/tools/parameter_sweep.py
```diff
@@ -811,44 +871,16 @@ def _do_param_sweep(
 
         run_successful = False  # until proven otherwise
```
```suggestion
```

bknueven (commented):
Looks pretty good. Just a minor cleanup suggestion.

watertap/tools/parameter_sweep.py
k1nshuk: I don't think that line can be removed because, if the conditional on line https://github.com/k1nshuk/watertap/blob/063fe6ab51324c0b3ce8a87ff6cf065b5935a4a8/watertap/tools/parameter_sweep.py#L874 is not satisfied, `run_successful` does not get defined. I can put the above assignment within an `else:` if that is better from a readability standpoint.

watertap/tools/parameter_sweep.py
```diff
@@ -771,6 +771,63 @@ def _read_output_h5(filepath):
...
+                    v.set_value(val, skip_validation=True)
+            reinitialize_function(model, **reinitialize_kwargs)
+
+    try:
+        # Simulate/optimize with this set of parameter
+        with capture_output():
+            results = optimize_function(model, **optimize_kwargs)
+        pyo.assert_optimal_termination(results)
+
+    except:
```
lbianchi-lbl, line 804: I'd recommend to specify the exception type(s) (the more specific the better).

watertap/tools/parameter_sweep.py
```diff
@@ -771,6 +771,63 @@ def _read_output_h5(filepath):
...
+            for v, val in reinitialize_values.items():
+                if not v.fixed:
+                    v.set_value(val, skip_validation=True)
+            try:
+                reinitialize_function(model, **reinitialize_kwargs)
+                with capture_output():
+                    results = optimize_function(model, **optimize_kwargs)
+                pyo.assert_optimal_termination(results)
+
+            except:
```
lbianchi-lbl, line 816: See above re: exception types.

bknueven (approved):

lbianchi-lbl (dismissed):

watertap/tools/parameter_sweep.py
bknueven: I don't think we should make this change, because we basically *do* want to catch everything here.

Further, this is not new code and is just reorganizing existing code into a separate function. If we do want to make this change, it should be done in a separate PR.

lbianchi-lbl:
```suggestion
    except RuntimeError:  # raised by pyo.assert_optimal_termination()
```

watertap/tools/parameter_sweep.py
lbianchi-lbl:
```suggestion
            except RuntimeError:  # raised by pyo.assert_optimal_termination()
```

watertap/tools/parameter_sweep.py
lbianchi-lbl: @bknueven: :man_shrugging: :+1:

bknueven: This is not things raised by `assert_optimal_termination`, per se, because otherwise we would use `check_optimal_termination`.

Sometimes the parameter sweep tool finds weird ways to make either Ipopt *or* Pyomo, *or* the combination of both fail in ways that are difficult to both find and enumerate (e.g., inside the `optimize_function`). So any change to this blanket `except` needs to be very carefully considered.

lbianchi-lbl: @bknueven I see, thanks for the clarification and the extra context. I agree that the "catch-all" approach is more justified here.

PR #633, opened 2022-07-22
Example flowsheets for activated sludge processes

andrewlee94:
## Summary/Motivation:
This PR adds some initial flowsheets for activated sludge processes.

-
-

### Legal Acknowledgement


watertap/unit_models/cstr_injection.py
```diff
@@ -0,0 +1,338 @@
...
+# Solutions of Sandia, LLC, Carnegie Mellon University, West Virginia University
+# Research Corporation, et al.  All rights reserved.
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and
+# license information.
+#################################################################################
+"""
+Modified CSTR model which includes terms for injection of species/reactants.
+
+This is copied the standard IDAES CSTR with the addition of mass trasnfer terms.
```
tarnold17:
```suggestion
This is copied from the standard IDAES CSTR with the addition of mass transfer terms.
```

watertap/examples/flowsheets/case_studies/activated_sludge/ASM1_flowsheet_seq.py
```diff
@@ -0,0 +1,422 @@
...
+
+    # Fifth reactor (aerobic) - CSTR with injection
+    m.fs.R5 = CSTR_Injection(
+        default={
+            "property_package": m.fs.props,
+            "reaction_package": m.fs.rxn_props,
+        }
+    )
+    m.fs.R5.volume.fix(1333 * pyo.units.m**3)
+    m.fs.R5.injection[...].fix(0)
```
adam-a-a:
```suggestion
    m.fs.R5.injection[...].fix(0)
```
What does the [...] do? Is it sort of like using (:) in MATLAB, where this would grab all indices?

andrewlee94: Yes, it is a slicer, but `:` and `...` have different meanings. `:` means all elements of a single index, `...` means all elements of all indices (i.e. `:, :, etc`).

watertap/property_models/activated_sludge/asm1_properties.py
```diff
@@ -282,39 +283,76 @@ def build(self):
...
+                    * (12 * pyo.units.kg / pyo.units.kmol)
+                )
+            else:
+                return self.flow_vol * self.conc_mass_comp[j]
+
+        self.material_flow_expression = pyo.Expression(
+            self.component_list,
+            rule=material_flow_expression,
+            doc="Material flow terms",
         )
```
adam-a-a, line 304: Why switch to expressions instead of using the former methods? My guess is because it'd be more desirable for you to point to an Expression object instead of calling a function.

andrewlee94: Because we need to provide scaling factors for those expressions so that the control volumes can do scaling.

watertap/property_models/activated_sludge/asm1_properties.py
```diff
@@ -282,39 +283,76 @@ def build(self):
...
+
+    def get_enthalpy_flow_terms(self, p):
+        return self.enthalpy_flow_expression
+
+    def get_material_density_terms(self, p, j):
+        return self.material_density_expression[j]
+
+    def get_energy_density_terms(self, p):
+        return self.energy_density_expression
+
```
adam-a-a, line 355: Ah, I see now. You just moved these down and have them returning the new expressions.

watertap/unit_models/cstr_injection.py
```diff
@@ -0,0 +1,338 @@
...
+            return b.control_volume.rate_reaction_extent[t, r] == (
+                b.volume[t] * b.control_volume.reactions[t].reaction_rate[r]
+            )
+
+        # Set references to balance terms at unit level
+        if (
+            self.config.has_heat_transfer is True
+            and self.config.energy_balance_type != EnergyBalanceType.none
+        ):
+            self.heat_duty = Reference(self.control_volume.heat[:])
```
adam-a-a, line 300: Still not 100% clear on when/why we'd choose to use `Reference` over `add_object_reference`

andrewlee94: In general you should use `Reference` wherever possible.

There are two exceptions where using `add_object_reference` is appropriate:

1. When you want to reference a scalar object - using `Reference` will make it appear to be indexed by `[None]` which is unfortunate and likely confusing (there are reasons behind this).
2. If you explicitly do not want the `Reference` to appear in `component_objects`. `References` get picked up by this, so when iterating over all components the referenced component can show up twice. There are other ways to avoid this however, so this should only be done where there is a very good cause.

watertap/unit_models/cstr_injection.py
adam-a-a: Without having gone through the CSTR class to sort out the details myself, could this have just been inherited from CSTR, followed by setting has_mass_transfer True for the control volume?

andrewlee94: Not without changing the CSTR model, and I decided that was not appropriate in this case. However, I think we should generalize this model a bit more to also cover the anaerobic digestor case (and any liquid-phase reactor with vapour headspace we might have).

adam-a-a (commented):
Looks good so far - just a couple of comments/questions for now.

watertap/examples/flowsheets/case_studies/activated_sludge/ASM1_flowsheet_seq.py
adam-a-a: I see. That was what I meant by my MATLAB example, since using (:) in MATLAB gives you all elements of all indices.

watertap/examples/flowsheets/case_studies/activated_sludge/ASM1_flowsheet.py
```diff
@@ -0,0 +1,364 @@
...
+# by the software owners: The Regents of the University of California, through
+# Lawrence Berkeley National Laboratory,  National Technology & Engineering
+# Solutions of Sandia, LLC, Carnegie Mellon University, West Virginia University
+# Research Corporation, et al.  All rights reserved.
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and
+# license information.
+#################################################################################
+"""
+Example of activated sludge process model using ASM1.
```
yalinli2: Maybe add a bit note here that the system is based on BSM1 (hence the reference) with 2 anoxic chambers and 3 aerobic chambers.

watertap/examples/flowsheets/case_studies/activated_sludge/ASM1_flowsheet.py
```diff
@@ -0,0 +1,364 @@
...
+
+
+Unit opeations are modeled as follows:
+
+    * Anoixic reactors as standard CSTRs (CSTR)
+    * Aerobic reactors as CSTRs with additioanl injection terms (CSTR_Injection)
```
yalinli2: Maybe explicitly say that `S_O` is the injection term

yalinli2 (approved):
Looks good, thanks!

watertap/examples/flowsheets/case_studies/activated_sludge/ASM1_flowsheet.py
andrewlee94: @yalinli2 Whilst `S_O` is the only term we will use in the example, there are injection terms for all species.

yalinli2: oh I see, thanks!

watertap/examples/flowsheets/case_studies/activated_sludge/ASM1_flowsheet.py
```diff
@@ -0,0 +1,381 @@
...
+# license information.
+#################################################################################
+"""
+Example of activated sludge process model based on BSM1 study using ASM1 reactions.
+This process contains 2 anoxic tanks followed by 3 aerobic tanks and a clarifier
+with recycle streams.
+
+Unit operations are modeled as follows:
+
+    * Anoixic reactors as standard CSTRs (CSTR)
```
adam-a-a:
```suggestion
    * Anoxic reactors as standard CSTRs (CSTR)
```

yalinli2 (approved):

watertap/examples/flowsheets/case_studies/activated_sludge/ASM1_flowsheet.py
```diff
@@ -0,0 +1,381 @@
...
+with recycle streams.
+
+Unit operations are modeled as follows:
+
+    * Anoxic reactors as standard CSTRs (CSTR)
+    * Aerobic reactors as CSTRs with additional injection terms (CSTR_Injection)
+    * Clarifier as a Separator with split fractions by components
+
+Note that a pressure changer is required in the recycle stream to ensure the
+pressure inside the recycle loop is bounded. As the inlet Mixer uses a pressure
```
luohezhiming, line 25:
```suggestion
pressure inside the recycle loop is bounded. As the inlet mixer uses a pressure
```

watertap/property_models/activated_sludge/asm1_properties.py
```diff
@@ -264,7 +265,7 @@ def build(self):
         self.temperature = pyo.Var(
             domain=pyo.NonNegativeReals,
             initialize=298.15,
-            bounds=(298.15, 323.15),
+            bounds=(293.15, 323.15),
```
luohezhiming, line 268: I think the lower bound for temperature is still a little high. Even though it is an exothermic reaction, the initial environmental temperature can be low in winter, so probably set to 273.15?

watertap/property_models/activated_sludge/asm1_properties.py
```diff
@@ -282,39 +283,76 @@ def build(self):
...
+        # Material and energy flow and density expressions
+        def material_flow_expression(self, j):
+            if j == "H2O":
+                return self.flow_vol * self.params.dens_mass
+            elif j == "S_ALK":
+                # Convert moles of alkalinity to mass of C assuming all is HCO3-
+                return (
+                    self.flow_vol
+                    * self.alkalinity
+                    * (12 * pyo.units.kg / pyo.units.kmol)
```
luohezhiming, line 295: Is this `12 * pyo.units.kg / pyo.units.kmol` the molar mass of the solute? Then why not set it as a variable?

watertap/examples/flowsheets/case_studies/activated_sludge/ASM1_flowsheet.py
```diff
@@ -0,0 +1,381 @@
...
+minimization constraint and there is no pressure drop in the rectors, if pressure
```
adam-a-a:
```suggestion
minimization constraint and there is no pressure drop in the reactors, if pressure
```

watertap/examples/flowsheets/case_studies/activated_sludge/ASM1_flowsheet.py
andrewlee94: This is a deliberate choice - `Mixer` is the class used for the unit model.

watertap/property_models/activated_sludge/asm1_properties.py
andrewlee94: No, this is the molecular weight of carbon, and is a constant which will not change.

watertap/property_models/activated_sludge/asm1_properties.py
andrewlee94: Activated sludge processes don't work at low temperatures, so I think this is probably OK - either way that is something the user should update on a case-by-case basis.

luohezhiming (approved):
Looks great to me!

watertap/property_models/activated_sludge/asm1_properties.py
luohezhiming: A minor point, I think keep this is OK, but maybe add it as a parameter can be another option.

PR #636, opened 2022-07-24
Add suboxic ASM model

luohezhiming:
## Summary/Motivation:

Add suboxic activated sludge model for AMO project 1581

- add suboxicASM unit model
- add flowsheet
- add sensitivity analysis
- add corresponding tests
- revise typo for biomembrane filtration test

### Legal Acknowledgement


watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/suboxic_activated_sludge_process/suboxic_activated_sludge_process.py
```diff
@@ -0,0 +1,208 @@
...
+    return m, results
+
+
+def build():
+    m = ConcreteModel()
+    m.db = Database()
+
+    m.fs = FlowsheetBlock(default={"dynamic": False})
+    m.fs.prop = prop_ZO.WaterParameterBlock(
+        default={"solute_list": ["bod", "tss", "tkn", "phosphorus"]}
```
tarnold17, line 73: @adam-a-a #608 removed a lot of instances of the word "solute" - is it still appropriate to use it here, as `solute_list`? Or does it need to be changed to something else?

tarnold17 (approved):

watertap/data/techno_economic/suboxic_activated_sludge_process.yaml
```diff
@@ -0,0 +1,89 @@
...
+      units: dimensionless
+      constituent_longform: Biological Oxygen Demand (BOD)
+    tss:
+      value: 0.9851
+      units: dimensionless
+      constituent_longform: Total Suspended Solids (TSS)
+    tkn:
+      value: 0.8846
+      units: dimensionless
+      constituent_longform: Total Kjeldahl Nitrogen (TKN)
```
adam-a-a, line 23: I know that you're following info provided, but when we go to make flowsheets and think about the situation where we mix and match different unit models, if all other models include something like `nitrogen` or `ammonia_as_nitrogen`, having this component as TKN for this unit might not be desirable.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/suboxic_activated_sludge_process/multi_sweep.py
```diff
@@ -0,0 +1,85 @@
...
+        optimize_function=opt_function,
+        optimize_kwargs=optimize_kwargs,
+        interpolate_nan_outputs=interpolate_nan_outputs,
+    )
+
+    return global_results, sweep_params, m
+
+
+if __name__ == "__main__":
+    results, sweep_params, m = run_analysis(*sys.argv[1:])
```
adam-a-a: There is a more user-friendly message that can be displayed here. We haven't done for all flowsheets yet, but you might want to look at another example and add that in (maybe look at the example for swine_wwt?)

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/suboxic_activated_sludge_process/suboxic_activated_sludge_process.py
adam-a-a: @tarnold17 Oh, we are definitely keeping `solute_list` around for the time being. This is something that exists in property models, whether zero-order or not. The main concern was over the variable name for removal fraction, which was something new that we introduced, whereas `solute_list`, to my knowledge, has been around for some time and is likely interlaced throughout a lot of IDAES code. Technically, we could deviate from that convention in our zero_order_properties, but personally, I'd rather focus our efforts on more fruitful tasks, like implementing new models and adding detail to existing ones. Thanks for catching that and probing though!

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/suboxic_activated_sludge_process/tests/test_multi_sweep.py
```diff
@@ -0,0 +1,34 @@
...
+for case_num in [1, 2, 3, 4]:
+    sweep_list.append(case_num)
+
+
+@pytest.mark.parametrize("case_num", sweep_list)
+@pytest.mark.integration
+def test_multi_sweep(case_num, tmp_path):
+    cwd = os.getcwd()
+    os.chdir(tmp_path)
+    nx = 11
```
adam-a-a:
```suggestion
    nx = 2
```

adam-a-a: We don't need to extend to 11 since testing already consumes a lot of time. 2 steps should be sufficient.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/suboxic_activated_sludge_process/tests/test_suboxic_activated_sludge_process.py
```diff
@@ -0,0 +1,60 @@
...
+        m.fs.suboxicASM.properties_treated[0].flow_mass_comp["tss"]
+    ) == pytest.approx(0.001780, rel=1e-3)
+    assert value(
+        m.fs.suboxicASM.properties_treated[0].flow_mass_comp["tkn"]
+    ) == pytest.approx(0.002134, rel=1e-3)
+    assert value(
+        m.fs.suboxicASM.properties_treated[0].flow_mass_comp["phosphorus"]
+    ) == pytest.approx(0.0003556, rel=1e-3)
+
+    assert value(m.fs.costing.LCOW) == pytest.approx(0.05581, rel=1e-3)  # in M$/m**3
```
adam-a-a:
```suggestion
    assert value(m.fs.costing.LCOW) == pytest.approx(0.05581, rel=1e-3)  # in $/m**3
```
Is this really in million $/m3, or is that comment outdated?

adam-a-a: If it somehow is in M$/m3, then something is wrong with costing.

watertap/unit_models/zero_order/tests/test_suboxic_activated_sludge_process_zo.py
```diff
@@ -0,0 +1,223 @@
...
+    assert isinstance(
+        m.fs.costing.suboxic_activated_sludge_process.automated_valve_cost, Var
+    )
+    assert isinstance(
+        m.fs.costing.suboxic_activated_sludge_process.advanced_predicted_control_cost,
+        Var,
+    )
+    assert isinstance(
+        m.fs.costing.suboxic_activated_sludge_process.ammonia_probe_cost, Var
+    )
```
adam-a-a: Is this everything? The list looked longer in the cost method.

watertap/data/techno_economic/suboxic_activated_sludge_process.yaml
```diff
@@ -0,0 +1,89 @@
+default:
+  energy_electric_flow_vol_inlet:
+    value: 0.5
+    units: kWh/m^3
```
adam-a-a, line 4: This is likely the most important aspect for the unit in comparison to the conventional approach. In reality, we would calculate the compressor work required to supply the amount of air/oxygen going to the tank, so I'd think we'd want to approach this similar to how we have applied `pump_electricity` to calculate pump work. 

Now, if this is meant as an early first step due lack of info, then I would say that we should at least compare this value with what is listed for other related zero-order units, like `conventional_activated_sludge` and `aeration_basin` (or `aeration_tank` ?)

adam-a-a (changes requested):
Nice job on this - just some comments/questions that I think need to be addressed before we merge.

watertap/data/techno_economic/suboxic_activated_sludge_process.yaml
luohezhiming: The TKN is defined as the sum of organic nitrogen and ammonia, while the total nitrogen include TKN plus nitrate nitrogen and nitrite nitrogen. So `tkn` is a component of `nitrogen`, and `ammonia_as_nitrogen` is a component of `tkn`.

andrewlee94: @adam-a-a's point is more general than that - there are many ways that we can go about defining nitrogen in a waste stream, and if we start using specific ones in every model then there is no way to link them all together. Thus, we have tended to take a more hands-off approach and just say "nitrogen" or "ammonia" and leave the finer details of how that is measured up to the user (who needs to set the concentrations and removal fractions accordingly).

I.e. if one unit uses `tdk_nitrogen` and another uses `my_nitrogen`, then the user will have to do some sort of manipulation to get the two to work together. If they both use `nitrogen`, then the user only needs to set the parameters according (which they should be doing anyway).

watertap/data/techno_economic/suboxic_activated_sludge_process.yaml
luohezhiming: We don't have information for the `pump_electricity` right now, but their original data sheet does provide an alternative electricity metrics based on daily oxygen demand instead of inlet flow. But I check the `conventional_activated_sludge` and `aeration_basin` models, they also both use the electricity method based on the inlet flow. The `energy_electric_flow_vol_inlet` for `conventional_activated_sludge` is 0.14 kWh/m3 and for `aeration_basin` is 0.41226735 kWh/m3

watertap/data/techno_economic/suboxic_activated_sludge_process.yaml
adam-a-a: Yes, @andrewlee94 hit the nail on the head. While your implementation here might be technically accurate, our objection is about compatibility between units in flowsheets.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/suboxic_activated_sludge_process/multi_sweep.py
luohezhiming: I revised based on `swine_wwt`

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/suboxic_activated_sludge_process/tests/test_suboxic_activated_sludge_process.py
luohezhiming: Yeah, the M$/m3 is outdated, I've revised to $/m3.

watertap/unit_models/zero_order/tests/test_suboxic_activated_sludge_process_zo.py
luohezhiming: Added all other variables in the test.

watertap/data/techno_economic/suboxic_activated_sludge_process.yaml
adam-a-a: Then at a minimum, I would expect that the default value used here should be less than either of those for `conventional_activated_sludge` or `aeration_basin` (assuming the default values for the latter two are relevant in the first place).

watertap/data/techno_economic/suboxic_activated_sludge_process.yaml
luohezhiming: @andrewlee94 @adam-a-a Thank you for clarification, I understand the points right now. This amo team only provides the removal fraction for `tkn` at present. If the user wants to make a comparison, they should provide extra information for calculating `nitrogen` or `ammonia_as_nitrogen` accordingly based on their own data.

watertap/data/techno_economic/suboxic_activated_sludge_process.yaml
luohezhiming: The value is provided by the Tanja's team, they said their major advantage is their advanced control technology, which use less money than the conventional control. I think they might need to consider relating it to the energy consumption.

tarnold17 (approved):

tarnold17 (approved):

watertap/data/techno_economic/suboxic_activated_sludge_process.yaml
adam-a-a: OK, I would say that any issue here can be resolved after iteration with the team. To my understanding, a key advantage should be lower energy consumption in comparison to conventional approach, so it's something that should be revisited later.

adam-a-a (approved):
LGTM

watertap/data/techno_economic/suboxic_activated_sludge_process.yaml
luohezhiming: Yeah, I fully agree your thought. I think the advantages of their control strategy should be somehow adressed by low energy consumption. I will make a note and revist this value.

PR #640, opened 2022-07-26
Zero Order Variable Efficiency Pump (low pressure)

arao53:
## Fixes/Resolves:

Issue #641 item 1

## Summary/Motivation:
Adds a simple formulation for variable efficiency for low pressure pumps based on the pump's best-efficiency-point (BEP) flowrate. 

- Added a yaml and ZO unit model that mimic the PumpElectricityZO model. 
- Uses the costing parameters that were developed for PumpElectricityZO. 
- Added corresponding tests. 

### Legal Acknowledgement


watertap/unit_models/zero_order/pump_electricity_zo.py
```diff
@@ -90,3 +90,70 @@ def applied_pressure_constraint(b, t):
 
         self._perf_var_dict["Electricity (kW)"] = self.electricity
         self._perf_var_dict["Applied Pressure (bar)"] = self.applied_pressure
+
+
+@declare_process_block_class("PumpVariableEfficiencyZO")
+class PumpVariableEfficiencyZOData(PumpElectricityZOData):
```
adam-a-a: Although this can work too, I think it would be better, at least for automated documentation reasons, to make this a separate file.

adam-a-a: That means the yaml should also be separate too.

watertap/unit_models/zero_order/pump_electricity_zo.py
```diff
@@ -90,3 +90,70 @@ def applied_pressure_constraint(b, t):
...
+            bounds=(0, 1),
+            doc="Ratio between the true efficiency and the best efficiency due to variable operation",
+        )
+
+        @self.Constraint(
+            self.flowsheet().time, doc="Constraint for variable pump flowrate ratio"
+        )
+        def flow_ratio_constraint(b, t):
+            return b.flow_ratio[t] == b.properties[t].flow_vol / pyunits.convert(
+                b.properties.flow_bep, to_units=pyunits.m**3 / pyunits.s
```
adam-a-a:
```suggestion
                b.flow_bep, to_units=pyunits.m**3 / pyunits.s
```

watertap/unit_models/zero_order/pump_electricity_zo.py
```diff
@@ -90,3 +90,70 @@ def applied_pressure_constraint(b, t):
...
+        def eta_ratio_constraint(b, t):
+            if b.config.process_subtype is "variable_lowpressure":
+                # if the flow is too far from the design point, the efficiency is set to 40%
+                if b.flow_ratio[t] < 0.6 or b.flow_ratio[t] > 1.4:
+                    return b.eta_ratio[t] == 0.4
+                # otherwise, use the correlation
+                else:
+                    return b.eta_ratio[t] == (
+                        -0.995 * b.flow_ratio[t] ** 2 + 1.977 * b.flow_ratio[t] + 0.025
+                    )
```
adam-a-a: This _might_ require an `expr_if`

adam-a-a (commented):
Nice start! Let's break these into separate files for now. By the way, it looks like this relationship doesn't consider the potential impact of pump pressure (lift height) on efficiency... but that's OK since we are going to consider alternatives. I can help with the conditional for `eta_ratio`

watertap/unit_models/zero_order/pump_variable_zo.py
```diff
@@ -0,0 +1,133 @@
...
+    CONFIG = ZeroOrderBaseData.CONFIG()
+
+    def build(self):
+        super().build()
+
+        self._tech_type = "pump_variable"
+
+        build_pt(self)
+
+        self.lift_height = Var(units=pyunits.m, doc="Lift height for pump")
```
adam-a-a:
```suggestion
        self.lift_height = Var(self.flowsheet().config.time, units=pyunits.m, doc="Lift height for pump")
```

adam-a-a: Although the current formulation for efficiency seems to be independent of pump head, I think we should index this by time since we'll want this to change as well.

watertap/unit_models/zero_order/pump_variable_zo.py
```diff
@@ -0,0 +1,133 @@
...
+
+        self.eta_pump = Var(units=pyunits.dimensionless, doc="Efficiency of pump")
+
+        self.eta_motor = Var(units=pyunits.dimensionless, doc="Efficiency of motor")
```
adam-a-a: I think these should be indexed by time as well.

watertap/unit_models/zero_order/pump_variable_zo.py
```diff
@@ -0,0 +1,133 @@
...
+        @self.Constraint(
+            self.flowsheet().time, doc="Constraint for actual pump energy consumption"
+        )
+        def electricity_consumption(b, t):
+            return b.electricity[t] == pyunits.convert(
+                b.lift_height
+                * b.properties[t].flow_vol
+                * b.properties[t].dens_mass
+                * Constants.acceleration_gravity
+                / (b.eta_pump * b.eta_motor * b.eta_ratio[t]),
```
adam-a-a: We'll want to think about this, but additional details down the line could include part-load motor efficiency and _maybe_ the same for a VFD if we add that in.

adam-a-a: In my previous work, I just assumed the variable efficiency to be represented by the pump- the simplifying assumption being that the pump efficiency represented the overall efficiency. That is, I didn't directly account for motor or VFD efficiency. So accounting for these effects would be an improvement.

watertap/unit_models/zero_order/pump_variable_zo.py
```diff
@@ -0,0 +1,133 @@
...
+        )
+        def eta_ratio_constraint(b, t):
+            # if the flow is too far from the design point, the efficiency is set to 40%
+            if b.flow_ratio[t] < 0.6 or b.flow_ratio[t] > 1.4:
+                return b.eta_ratio[t] == 0.4
+            # otherwise, use the correlation
+            else:
+                return b.eta_ratio[t] == (
+                    -0.995 * b.flow_ratio[t] ** 2 + 1.977 * b.flow_ratio[t] + 0.025
+                )
```
adam-a-a: I think you can start adding tests. I expect that this might not work out, but let's see what happens after testing.

watertap/unit_models/zero_order/tests/test_pump_variable_zo.py
```diff
@@ -0,0 +1,91 @@
...
+        assert isinstance(model.fs.unit.flow_ratio, Var)
+        assert isinstance(model.fs.unit.flow_ratio_expr, Expression)
+
+    @pytest.mark.component
+    def test_load_parameters(self, model):
+        data = model.db.get_unit_operation_parameters("pump_variable")
+
+        model.fs.unit.load_parameters_from_database()
+
+        assert model.fs.unit.flow_bep.fixed
```
arao53, line 89: @adam-a-a this assertion is causing some issues, but the line below (91) recognizes `data["flow_bep"]["value"]`. Is there something else that needs to be added to `model.fs.unit.load_parameters_from_database()` (line 88)?

watertap/unit_models/zero_order/pump_variable_zo.py
arao53: Adding a time index for the lift height throws an error: `TypeError: Argument for expression is an indexed numeric value`. This is similar to what happens when `eta_motor` and `eta_pump` are also index by time.  I think this is because these variables are set within the yaml. 

In the NF examples, the lift height was unfixed at the flowsheet level and an additional constraint for applied pressure was added to connect the pump to other units.

adam-a-a: I think this was just because wherever they were used in their respective constraints, they were not being called with their indices. However, this should be resolved now.

watertap/unit_models/zero_order/pump_variable_zo.py
TimBartholomew: My modeling approach has been to just have a single pump efficiency that is the combined pump and motor efficiency. In some papers I would give both values - but I'd model the efficiency as one variable, but I think the key thing we need to think about is how do we expect users to interact with these parameters. For me, if someone wanted to vary the efficiency of the pump for a sensitivity they'd just want to vary one value rather than both, and I don't think anyone would really want to vary the two efficiencies separately - which leads me to think we should just combine them.

andrewlee94: A lot of people like to vary the parameters separately - not so much for sensitivity studies but for tracking different contributions. E.g. we have one collaboration in IDAES where the compressors will have 3 efficiencies (thermodynamic, mechanical and electric).

adam-a-a: I would also opt for having separate efficiencies. If we were to ever to go to the highest level of detail for flexible pump operation. While assuming the pump efficiency to be representative of the overall efficiency is a valid assumption, in the future, what we'd (I'd) really want is to have  pump, motor, and VFD efficiencies, all with part-load performance, but no need to go there yet.

arao53:
Thank you all for the feedback. This has been converted to a draft PR since our implementation will focused on using the 0D formulation PR#654.

adam-a-a:
Closing this out since the focus will be on PR #654

PR #646, opened 2022-07-27
slight mod in zo props

adam-a-a:
## Fixes/Resolves:

-
-

### Legal Acknowledgement


adam-a-a:
Resolving this issue simply by pointing to `self.config.property_package.solute_set ` with respect to a unit model instead, since config options are intended to be set by users, not internally.

PR #647, opened 2022-07-28
Add support for Python 3.10

### LINKED ISSUE #615: Add support for Python 3.10
I see IDAES now tests against Python 3.10. We should do the same, and if for some reason we have an incompatibility we should address it.

lbianchi-lbl:
## Summary/Motivation:

- It would be nice

- Add 3.10 to Python version specs in `setup.py`
- Add 3.10 to CI workflow job matrix

### Legal Acknowledgement


setup.py
```diff
@@ -97,7 +98,7 @@
         "fastjsonschema",  # schema validation
         "click",  # command-line tools with Click
         # tutorial tests
-        "nbformat",
+        # "nbformat",
```
```suggestion
        "nbformat",
```

setup.py
```diff
@@ -86,7 +87,7 @@
     ],
     keywords="water systems, chemical engineering, process modeling, filtration, desalination, nawi",
     packages=find_namespace_packages(),
-    python_requires=">=3.7, <4",
+    python_requires=">=3.7,<3.11",
```
bknueven:
```suggestion
    python_requires=">=3.7",
```
Is the `<3.11` worth it? Before this PR WaterTAP would happily install on 3.10...

bknueven:
```suggestion
    python_requires=">=3.7,<3.99",
```
An alternative ...

bknueven (approved):
Happy with this as-is -- just a minor question over compatibility with unreleased Python versions.

PR #652, opened 2022-08-01
Remove ref to missing var in tutorial flowsheet 

dangunter:
Related to https://github.com/watertap-org/watertap-ui/pull/31

### Legal Acknowledgement

### DIFF
--- a/tutorials/workshop_flowsheet.py
+++ b/tutorials/workshop_flowsheet.py
@@ -451,12 +451,12 @@ def export_ui_variables(fs):
                 "display_units": "{USD_2018}/(m**3/hr)",
                 "category": Category.cost,
             },
-            "load_factor": {
-                "display_name": "Load factor",
-                "scale_factor": 100,
-                "display_units": "%",
-                "category": Category.cost,
-            },
+            # "load_factor": {
+            #     "display_name": "Load factor",
+            #     "scale_factor": 100,
+            #     "display_units": "%",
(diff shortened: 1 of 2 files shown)


dangunter:
@TimBartholomew or @adam-a-a , this is a band-aid. I think we should pick a better home for the tutorials since finding/importing them from that one-off directory is confusing

tutorials/workshop_flowsheet.py
```diff
@@ -451,12 +451,12 @@ class Category:
                 "display_units": "{USD_2018}/(m**3/hr)",
                 "category": Category.cost,
             },
-            "load_factor": {
-                "display_name": "Load factor",
-                "scale_factor": 100,
-                "display_units": "%",
-                "category": Category.cost,
-            },
+            # "load_factor": {
```
TimBartholomew, line 454: It's name got changed to "utilitization_factor", we can probably just update it to that rather than removing it

TimBartholomew (commented):
I see why you brought up the maintenance concerns for the UI and the UI connection. Is there a way for us to integrate this into the CI testing?

Because if it was then the person who renamed the variable would know they'd have to change it in the UI connection.

dangunter:
This can be fixed by simply dropping this flowsheet entirely and wrapping instead 1-2 of the case study flowsheets (and adding diagrams for same). Then we could drop this tutorials directory entirely.

Issue #641, opened 2022-07-26
Variable efficiency models for pumps


arao53:
Implement efficiency correlations or characteristic curves for pumps that operate outside the best efficiency point. Constant efficiency is modeled in `pump_electricity_ZO`.

- [x] Low pressure (< 10bar) pumps with efficiency-flow correlations for ZO pump model. PR #640 
- [x] 0D pump unit that inherits from the IDAES pump unit model (similar to WaterTAP Isothermal Pump). PR #654 
- [x] Implement flow-head-efficiency correlations in the 0D pump unit model. #719

arao53:
Zero order pump is being replaced by 0D pump model, which builds on top of the WaterTAP Isothermal Pump model. Next steps include implementing pump efficiency correlations with relevance to desalination applications.

arao53:
Items 1 and 2 have been closed and item 3 is now being tracked by issue #719.

PR #654, opened 2022-08-03
Variable efficiency pump 0D

arao53:
## Fixes/Resolves:

Issue #641 item 2

## Summary/Motivation:
Models part-load-performance in pumps with an efficiency-flow correlation that is scaled to the pump's best-efficiency-point (BEP). Performance matches steady state performance at the best-efficiency-point, but this allows for more robust quasi-steady and dynamic analyses. 

- A new pump unit model that inherits from the IDAES base pump class with configuration options for pump efficiency correlations.
- Implemented basic correlations for variable efficiency based on pump flowrate and pump BEP.

### Legal Acknowledgement


watertap/unit_models/pressure_changer.py
```diff
@@ -11,35 +11,301 @@
...
+
+            self.flow_ratio = Var(
+                self.flowsheet().time,
+                initialize=1.0,
+                doc="Ratio of pump flowrate to best efficiency point flowrate",
+                units=pyunits.dimensionless,
+            )
+
+            # unfix the default efficiency from the base class
+            self.efficiency_pump.unfix()
```
adam-a-a: Is it fixed to begin with? I didn't check, but the only way I'd see this happening was if it was a variable that was treated as a parameter.

adam-a-a: Otherwise, if it isn't fixed within the model already, there would be no need for this line. You could quickly test this by asserting it isn't fixed after uncommenting this line and checking the build.

watertap/unit_models/pressure_changer.py
```diff
@@ -11,35 +11,301 @@
...
+
+            # add constraints
+            @self.Constraint(self.flowsheet().time, doc="Pump flow ratio")
+            def flow_ratio_constraint(b, t):
+                return b.flow_ratio[t] * b.bep_flow == (
+                    b.control_volume.properties_out[t].flow_vol.expr()
```
adam-a-a: Do we need the `.expr()`? I would think not.

watertap/unit_models/pressure_changer.py
```diff
@@ -11,35 +11,301 @@
...
+                    0.4,
+                    Expr_if(
+                        b.flow_ratio[t] > 1.4,
+                        0.4,
+                        -0.995 * b.flow_ratio[t] ** 2 + 1.977 * b.flow_ratio[t] + 0.018,
+                    ),
+                )
+
+        elif self.config.variable_efficiency is VariableEfficiency.flow_head:
+            raise ValueError(
```
adam-a-a: I think we may have a `NotImplemented` exception although in the past, others have recommended not using it since users may expect to have such features readily available. In this case though, I don't see a problem with using the NotImplemented exception.

watertap/unit_models/pressure_changer.py
```diff
@@ -11,35 +11,301 @@
...
+            #     initialize=1.0,
+            #     doc="Ratio of pump head to best efficiency point head",
+            #     units=pyunits.dimensionless,
+            # )
+            #
+            # @self.Constraint(self.flowsheet().time, doc="Pump head ratio")
+            # def head_ratio_constraint(b, t):
+            #     if b.control_volume.properties_in[t].dens_mass_phase["Liq"] is None:
+            #         raise Exception(
+            #             "Density is not implemented in property package as 'dens_mass_phase'"
```
adam-a-a: I don't think you should worry about catching this exception. Generally, I think most if not all of our models expect density with this naming convention.

watertap/unit_models/pressure_changer.py
```diff
@@ -11,35 +11,301 @@
...
+
+        for ind, c in self.control_volume.isothermal_balance.items():
+            sf = iscale.get_scaling_factor(
+                self.control_volume.properties_in[0].temperature
+            )
+            iscale.constraint_scaling_transform(c, sf)
+
+        if hasattr(self, "bep_flow"):
+            if iscale.get_scaling_factor(self.bep_flow) is None:
+                iscale.set_scaling_factor(self.bep_flow, 10)
```
adam-a-a: This is likely one the user should provide, so we can have a default scaling factor with a warning. I would opt to set it to the inverse of whatever bep_flow value is (or something close to that).

watertap/unit_models/pressure_changer.py
```diff
@@ -11,35 +11,301 @@
...
+                iscale.set_scaling_factor(self.flow_ratio_constraint[0], 1)
+
+        if hasattr(self, "head_ratio_constraint"):
+            if iscale.get_scaling_factor(self.head_ratio_constraint[0]) is None:
+                iscale.set_scaling_factor(self.head_ratio_constraint[0], 1)
+
+        if hasattr(self, "eta_constraint"):
+            if iscale.get_scaling_factor(self.eta_constraint[0]) is None:
+                iscale.set_scaling_factor(self.eta_constraint[0], 1)
+
```
adam-a-a: These are constraints, so we typically use constraint_scaling_transform. Also, we might be able to get away with omitting constraint scaling for these.

watertap/unit_models/pressure_changer.py
```diff
@@ -11,35 +11,301 @@
...
+
+@declare_process_block_class("PumpVar")
+class PumpVariableData(PumpData):
+    """
+    Variable efficiency Pump Unit Model Class
+    Uses pump affinity laws for industrial scale centrifugal pumps.
+    Atia et al. (2019) https://doi.org/10.1016/j.desal.2019.07.002
+    """
```
adam-a-a: Did you mean to remove this class?

watertap/unit_models/pressure_changer.py
arao53: I was using it earlier to initialize `bep_eta` but then I realized all it was doing was passing the value that initialized `efficiency_pump`. I have removed both instances.

watertap/unit_models/pressure_changer.py
arao53: Removed.

watertap/unit_models/pressure_changer.py
arao53: Turns out it works well without constraint scaling.

watertap/unit_models/pressure_changer.py
arao53: Yup, I meant to remove it. Not sure why those changes weren't pushed before but they should be now.

watertap/unit_models/pressure_changer.py
```diff
@@ -40,6 +69,93 @@ def build(self):
...
+            #     return b.head_ratio[t] * b.bep_head * Constants.acceleration_gravity == (
+            #             (
+            #                     b.control_volume.properties_out[t].pressure
+            #                     / b.control_volume.properties_out[t].dens_mass_phase["Liq"]
+            #             )
+            #             - (
+            #                     b.control_volume.properties_in[t].pressure
+            #                     / b.control_volume.properties_in[t].dens_mass_phase["Liq"]
+            #             )
+            #     )
```
luohezhiming: Why have all these variables and constraint commented? Is this still a in-progress PR?

arao53: The commented variables and constraints have been removed.

adam-a-a:
@lbianchi-lbl will be adding an IDAES tag here which incorporates changes from [PR 925 in IDAES](https://github.com/IDAES/idaes-pse/pull/925). That way, we can check a couple more potential issues that can occur with WaterTAP's pump and/or ERD (compressor) unit model. Thanks in advance, @lbianchi-lbl!

watertap/unit_models/pressure_changer.py
andrewlee94: I second this - I do not see a reason why the `.expr()` should be required (and many good reasons why it should not).

For one, this assumes that `flow_vol` is an `Expression`, which maybe true in the current case, but is hardly general (and even then it should not be needed).

watertap/unit_models/pressure_changer.py
```diff
@@ -11,35 +11,126 @@
...
+                    0.4,
+                    Expr_if(
+                        b.flow_ratio[t] > 1.4,
+                        0.4,
+                        -0.995 * b.flow_ratio[t] ** 2 + 1.977 * b.flow_ratio[t] + 0.018,
+                    ),
+                )
+
+        elif self.config.variable_efficiency is VariableEfficiency.flow_head:
+            raise ValueError(
```
andrewlee94: I would suggest that this should be a `NotImplementedError` rather than a `ValueError` - the value is valid, but the method is not implemented.

watertap/unit_models/pressure_changer.py
```diff
@@ -11,35 +11,126 @@
...
+                "Config option 'VariableEfficiency.flow_head' is not fully implemented yet"
+            )
+            # TODO - Implement pump efficiency expression based on flow and head (bep_head, head_ratio)
```
andrewlee94, line 126: Should there be a closing `else` statement here? Even if it is not required, it can be good practice to put `else: pass` here to make it clear that this is expected behaviour.

watertap/unit_models/tests/test_pressure_changer.py
```diff
@@ -259,3 +262,108 @@ def test_solution(self, Pump_frame):
...
+            "flow_ratio",
+        ]
+        for v in var_list:
+            assert hasattr(m.fs.unit, v)
+            var = getattr(m.fs.unit, v)
+            assert isinstance(var, Var)
+
+        assert isinstance(m.fs.unit.flow_ratio_constraint, Constraint)
+
+    @pytest.mark.component
```
andrewlee94: This looks like it should be a unit test, not a component test.

watertap/unit_models/tests/test_pressure_changer.py
```diff
@@ -259,3 +262,108 @@ def test_solution(self, Pump_frame):
...
+
+        # check that all variables and constraints have scaling factors
+        unscaled_var_list = list(unscaled_variables_generator(m))
+        assert len(unscaled_var_list) == 0
+
+        unscaled_constraint_list = list(unscaled_constraints_generator(m))
+        assert len(unscaled_constraint_list) == 0
+
+    @pytest.mark.component
+    def test_solution(self, Pump_frame):
```
andrewlee94, line 395: Should there be a test for the initialization method before this?

It would also be good to have a test for unit consistency as well.

andrewlee94 (changes requested):

watertap/unit_models/tests/test_pressure_changer.py
```diff
@@ -299,8 +301,9 @@ def Pump_frame(self):
 
         m.fs.unit.bep_eta.fix(efi_pump)
 
-        flow_vol = m.fs.unit.control_volume.properties_in[0].flow_vol.expr()
-        m.fs.unit.bep_flow.fix(flow_vol)
+        m.fs.unit.flow_ratio.fix(1)
+        # flow_vol = m.fs.unit.control_volume.properties_in[0].flow_vol.expr()
```
andrewlee94: These lines should be removed if not needed.

watertap/unit_models/tests/test_pressure_changer.py
```diff
@@ -354,6 +402,7 @@ def test_solution(self, Pump_frame):
         assert pytest.approx(1, rel=1e-3) == value(m.fs.unit.eta_ratio[0])
 
         # Test low bep flow case
+        m.fs.unit.flow_ratio.unfix()
         m.fs.unit.bep_flow.fix(default_flow_vol / 2)
         results = solver.solve(m)
```
andrewlee94, line 404: This seems to be duplicating the solve that is now done in an earlier test. Ideally, this test would just check the values in the model based on the previous solve.

andrewlee94 (commented):
A couple of minor comments.

watertap/unit_models/tests/test_pressure_changer.py
arao53: For the default case (line 402), the initial solve is used. The additional solves test different regions of the step-wise flow-efficiency correlation.

adam-a-a: @andrewlee94 - We used a nested `Expr_if `to account for 3 different regions of a curve representative of pump efficiency ratio as a function of pump flow ratio and wanted to ensure each region worked as expected.

luohezhiming (approved):
LGTM

watertap/unit_models/pressure_changer.py
```diff
@@ -49,6 +143,36 @@ def calculate_scaling_factors(self):
...
+
+        # scale constraints
+
+        if hasattr(self, "flow_ratio_constraint"):
+            if iscale.get_scaling_factor(self.flow_ratio_constraint[0]) is None:
+                iscale.set_scaling_factor(self.flow_ratio_constraint[0], 1)
+
+        if hasattr(self, "eta_constraint"):
+            if iscale.get_scaling_factor(self.eta_constraint[0]) is None:
+                iscale.set_scaling_factor(self.eta_constraint[0], 1)
```
adam-a-a: I'm not sure we only want to scale the first time element. Perhaps we should remove `[0]` unless there is some reasoning behind this.

adam-a-a: Actually, I think this should be indexed and looped through by time

adam-a-a (changes requested):
@arao53 Just a quick adjustment with scaling time-dependent vars and I think this is good to go!

watertap/unit_models/pressure_changer.py
adam-a-a: @arao53 I think this suggested change still stands, but it might not be much of an issue. I just thought that anything related to flowrate will highly depend on the user's decision, so perhaps there should be a warning to the user indicating such.

adam-a-a (approved):
LGTM - I had one comment about scaling, but I think any further modifications can be handled in a subsequent PR as needed.

watertap/unit_models/pressure_changer.py
arao53: @adam-a-a sorry about that! Somehow this got lost along the way. I'll update that shortly.

andrewlee94 (approved):
LGTM

PR #660, opened 2022-08-09
Generator for testing variable sensitivity across large- and small- scale flow perturbations of a specified model

hunterbarber:
## Fixes/Resolves:
n/a

## Summary/Motivation:
This is a draft for a potential generation tool to vary the magnitude of inlet flow for a model to upper and lower bounds representing the limitations of large- and small-scale process designs. The generator is intended to assist in catching variables proportional to flow to guide sf, variables badly scaled at different magnitudes, and potentially hitting variable bounds at different magnitudes. This is posed as a draft for discussion on whether the generator would be useful for generating model statistics across other unit models. If so, this would motivate further development of the tool. Therefore, the current file location and model structure are subject to change or will be removed.

n/a

### Legal Acknowledgement


hunterbarber:
@adam-a-a @andrewlee94 I have migrated this generator (discussed in #626) to a new draft PR just to decouple it from the GAC model development. It may be that this is scraped, but I wanted to at least keep the discussion open for now and allow for the other PR to be escalated from a draft soon.

watertap/unit_models/tests/test_gac.py
```diff
@@ -367,6 +367,19 @@ def test_solution_robust(self, gac_frame_robust):
...
+    def test_variable_sensitivity_robust(self, gac_frame_robust):
+        m = gac_frame_robust
+
+        from watertap.core.util.scaling import variable_sens_generator
+
+        # test
+        sens_var_lst = list(variable_sens_generator(m))
+        for i in sens_var_lst:
+            print(i)
+
```
adam-a-a: We'd opt to omit the print lines for the final test.

watertap/core/util/scaling.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def variable_sens_generator(blk, lb_scale=1e-1, ub_scale=1e3, tol=1e2, zero=1e-10):
+    """
+
+    Keyword Arguments:
+
+    Returns:
+
+    """
```
adam-a-a: Thanks for taking the initiative on this. I think once you document this more clearly via docstrings, add logging (similar to what's done in `watertap/core/util/infeasible` and `watertap/core/util/initialization`), and then add dedicated testing to verify the intended functionality of this tool, it will probably be a helpful addition to our utilities.

watertap/core/util/scaling.py
```diff
@@ -0,0 +1,96 @@
...
+
+    test_scale = [lb_scale, ub_scale]
+    var_hist = {}
+    for scale in test_scale:
+        temp_blk = blk.clone()
```
adam-a-a, line 77: @andrewlee94 This is something I haven't looked into, but would you happen to know the difference between using `deepcopy` vs `clone`?

andrewlee94: They have very different uses:

`clone` is for copying Pyomo blocks (and is a Pyomo method). `deepcopy` is a Python method for creating copies of `dicts`.

hunterbarber:
@adam-a-a @andrewlee94 There's some improved commenting in the most recent commit that I think would be a good time to get some more feedback. Pending your comments, I think it may be time to switch over to working on a test file with it to see it it is universally applicable across different unit models. Right now it is testing the old (and soon to be updated with #626) GAC model because I was familiar with the lower fidelity of the scaling factor methods at that point, and knew what the generator should be outputting and causing the test to fail correctly. With testing other models it will be evident on the checks that will need added to ensure it works as intended.

Let me know any feedback you may have.

watertap/core/util/scaling.py
```diff
@@ -0,0 +1,149 @@
...
+# Technology Laboratory (subject to receipt of any required approvals from
+# the U.S. Dept. of Energy). All rights reserved.
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and license
+# information, respectively. These files are also available online at the URL
+# "https://github.com/watertap-org/watertap/"
+#
+###############################################################################
+"""
+This module contains utility functions for model statistics of WaterTAP models.
```
andrewlee94: Minor comment: This says the utility is for model statistics, but the name implies scaling.

watertap/core/util/scaling.py
```diff
@@ -0,0 +1,149 @@
...
+"""
+
+import pyomo.environ as pyo
+from pyomo.environ import check_optimal_termination
```
andrewlee94: Why import this separately from above? You can just do `pyo.check_optimal_termination`.

andrewlee94 (commented):
I did a quick read though and this looks good - the extra documentation makes it a lot clearer what is going on. I only had two very minor comments.

hunterbarber:
As an idea, may be beneficial to include functions from #691 once merged

watertap/core/util/scaling.py
```diff
@@ -0,0 +1,149 @@
...
+        blk : pyomo ConcreteModel block which holds the flowsheet to be tested
+        lb_scale : the lower bound scale which adjusts the magnitude for the
+            fixed inlet flows of the process (default = 1e-2)
+        ub_scale : the upper bound scale which adjusts the magnitude for the
+            fixed inlet flows of the process (default = 1e2)
+        tol : tolerance for the change in magnitude of scaled variables between
+            the lower and upper bounds (default = 1e3), should always be smaller
+            than the difference in magnitude of lb_scale and up_scale
+        zero: magnitude that is considered to be zero, scaled variables with a value of
+            zero are okay, and not reported (default = 1e-10)
```
bknueven:
```suggestion
        zero: magnitude that is considered to be zero, scaled variables with an absolute value
            of less than this value are okay, and not reported (default = 1e-10)
```

watertap/core/util/scaling.py
```diff
@@ -0,0 +1,149 @@
...
+
+    # TODO: Check DOF is 0
+
+    sv_hist = {}
+
+    for scale in [lb_scale, ub_scale]:
+
+        temp_blk = (
+            blk.clone()
+        )  # clone flowsheet to re-solve at conditions supplied by scale
```
bknueven:
```suggestion
        # clone flowsheet to re-solve at conditions supplied by scale
        temp_blk = blk.clone()
```

This is a really minor nit, but we need to be putting comments *above* the line of code they apply to. Black comes along as does this to it, which makes things harder to read.

watertap/core/util/scaling.py
```diff
@@ -0,0 +1,149 @@
...
+
+        # loop through all vars to wipe previous sf and reset dsf considering new scale
+        for (var_name, var_index), var_obj in temp_blk.component_data_iterindex(
+            ctype=pyo.Var, active=True, descend_into=True
+        ):
+
+            unset_scaling_factor(
+                var_obj
+            )  # remove prior sf which are reestablished on init
```
bknueven: Minor nit again

watertap/core/util/scaling.py
```diff
@@ -0,0 +1,149 @@
...
+                blk.fs.properties.get_default_scaling(var_name, index=var_index)
+                is not None
+                and "flow" in var_name
+                and var_obj.fixed
+            ):
+
+                # overwrite old dsf with one considering the new scale
+                dsf = blk.fs.properties.get_default_scaling(var_name, index=var_index)
+                temp_blk.fs.properties.set_default_scaling(
+                    var_name, dsf * scale**-1, index=var_index
```
bknueven:
```suggestion
                    var_name, dsf / scale, index=var_index
```

watertap/core/util/scaling.py
```diff
@@ -0,0 +1,149 @@
...
+        results = solver.solve(temp_blk)
+
+        # ensure model solves wrt new scale
+        if not pyo.check_optimal_termination(results):
+            yield "Failed run on", scale, "scale"
+            break
+
+        # check variable scaling wrt new scale
+        for b in list(badly_scaled_var_generator(temp_blk, zero=zero)):
+            yield f"badly scaled variable for scaled flow of {scale} ", b[0].name, b[1]
```
bknueven:
```suggestion
        for var, scaled_value in badly_scaled_var_generator(temp_blk, zero=zero):
            yield f"badly scaled variable for scaled flow of {scale} ", var.name, scaled_value
```
No need to create an intermediate list, and tuple unpacking makes the intent clear

watertap/core/util/scaling.py
```diff
@@ -0,0 +1,149 @@
...
+
+        # store results for sv to sv_hist dict for current model copy
+        # TODO: Can this be moved to eliminate the need for the sv_hist dict
+        for v in ComponentSet(
+            temp_blk.component_data_objects(pyo.Var, active=True, descend_into=True)
+        ):
```
bknueven:
```suggestion
        for v in temp_blk.component_data_objects(pyo.Var, active=True, descend_into=True):
```
No need to create a container for these iterators.

watertap/core/util/scaling.py
```diff
@@ -0,0 +1,149 @@
...
+
+            val = pyo.value(v, exception=False)
+            if val is None:
+                continue
```
bknueven: We might check `v.stale` here -- shouldn't all non-stale variables have a value since at this point we're post-successful solve?

watertap/core/util/scaling.py
```diff
@@ -0,0 +1,149 @@
...
+import idaes.logger
+from idaes.core.solvers import get_solver
+from idaes.core.util.scaling import (
+    calculate_scaling_factors,
+    badly_scaled_var_generator,
+    get_scaling_factor,
+    unset_scaling_factor,
+)
+
+solver = get_solver()
```
bknueven: I would move this inside the function -- no need for this solver object to live once the module is imported.

watertap/core/util/scaling.py
```diff
@@ -0,0 +1,149 @@
...
+            than the difference in magnitude of lb_scale and up_scale
+        zero: magnitude that is considered to be zero, scaled variables with a value of
+            zero are okay, and not reported (default = 1e-10)
+
+    Yields:
+        string denoting what step failed, variable name as string,
+            sensitivity of sv between lower and upper solutions
+    """
+
+    # TODO: Check DOF is 0
```
bknueven: I'm not sure this is relevant to this tool. I wouldn't do this.

watertap/core/util/scaling.py
```diff
@@ -0,0 +1,149 @@
...
+            fixed inlet flows of the process (default = 1e2)
+        tol : tolerance for the change in magnitude of scaled variables between
+            the lower and upper bounds (default = 1e3), should always be smaller
+            than the difference in magnitude of lb_scale and up_scale
+        zero: magnitude that is considered to be zero, scaled variables with a value of
+            zero are okay, and not reported (default = 1e-10)
+
+    Yields:
+        string denoting what step failed, variable name as string,
+            sensitivity of sv between lower and upper solutions
```
bknueven, line 57: I think this is okay -- but if we're yielding at not just printing to the screen, ideally his should return a status Enum for the first element of this tuple.

I suspect this would be most useful if it just printed its information to the screen, or maybe optionally a passed-in logging object.

bknueven (commented):
This looks pretty useful, and the core of the code looks correct.

I would like a test case that actually produces results -- according to the codecov report the test case "passes", which doesn't particularly make this a great test.

I have a lot of relatively minor comments -- mostly on style, though I do think it's important not to build intermediate objects just for iteration.

hunterbarber:
Thanks for the comments @bknueven, I will address these in future commits. I also wanted to log this challenge and see if you have any comments on it.

So I'm testing the generator across different unit models in the test file to identify issues with the universality of it. I'm on the electrodialysis_0D model and have encountered an issue with the old way the generator was constructed. I was looping through all variables using [this method](https://github.com/watertap-org/watertap/blob/6afcf52260db7762477beecd42981356adabbb6a/watertap/core/util/scaling.py#L85) to unfix sf. All old sf should be unset so that if a variable has a `if iscale.get_scaling_factor(self.var, warning=True) is None:`, this conditional statement is re-performed for each solve. The challenge arose when the electrolysis model uses [this method to establish sf](https://github.com/watertap-org/watertap/blob/main/watertap/unit_models/electrodialysis_0D.py#L790). The actual object that is captured in the `component_data_objects()` and used in model calculations is `current[t]`, but the sf is established and propagated through `current` which is not indexed and has no data(?) so not captured in this function. I've added [another loop](https://github.com/watertap-org/watertap/blob/6afcf52260db7762477beecd42981356adabbb6a/watertap/core/util/scaling.py#L72) using `component_objects()` but this may just be another temporary fix. I figure the code needs condensed into more efficient loops so I'm open to any suggestions.

**UPDATE** I think this is fixed and consolidated appearing in commits after 755336c

watertap/core/util/scaling.py
hunterbarber: I believe this was added because when looking through the `component_data_objects` there are `flow_mol_phase_comp_ref` and `flow_mol_phase_comp` variables (along with some other state variables) that appear separate but are the same effective data object. Adding the container eliminates the `_ref` variables that are duplicated objects. Let me know if you have a better solution, but I also need to double check that removing the container causes an error.

watertap/core/util/scaling.py
hunterbarber: Can you elaborate what you believe is the best reporting for this? I'm not sure of the most intuitive way the results may be displayed and don't think I exactly understood this.

bknueven: So, for printing to the screen, yielding a string is fine. For anything else, it's cumbersome.

So, we might define an `StrEnum` in this file to help us:
```python
from idaes.core.util.misc import StrEnum

class FailureType(StrEnum):
    run_failed = "Failed run on"
    badly_scaled_flow = "badly scaled flow"
    high_sensitivity = "high sensitivity of scaled variable"
```

Alternatively, maybe instead of yielding we should just print to the screen or (better) to a user-specified logger.

ksbeattie:
@hunterbarber any updates on this?

hunterbarber:
> @hunterbarber any updates on this?

No updates, but I hope to reevaluate it soon. I need to decide on how to keep it tested.

hunterbarber:
Looking to make it more generalizable with user options to make it a more adoptable tool

hunterbarber:
Closing this for now and repoening later, since it's remained in the backlog for too long and I don't see time in the foreseeable future for it

PR #668, opened 2022-08-12
DSPMDE property pack update

lbibl:
## Summary/Motivation:

This PR is to merge some major changes on the DSMPDE property package file from PR#559 to facilitate following refinement works on this file and on other unit models. 

- Added variable needed for the ion exchange unit model.
- Improved initialization and scaling. 

### Legal Acknowledgement


lbibl:
@lbianchi-lbl: 
May I invite you to take a look at this PR in draft, which failed only in Linux but not Win or Mac? The failure was due to a solver outcome error, where ipopt was used to optimize a system with dof of 2.  The inconsistency of Linux (not the experimental Mac) with Win is especially concerning.  I saw this inconsistency of Linux in a previous PR (merged) and resolved that by modifying the file.  But different solver outcome between Linux and Win might be worthy of an investigation.  Thank you for your attention.

adam-a-a:
@lbibl You have an absolute tolerance of 0.001 for specific energy consumption. Can you start by trying a relative tolerance?

adam-a-a:
You might also want to revisit scaling in the ED model.

lbibl:
> @lbibl You have an absolute tolerance of 0.001 for specific energy consumption. Can you start by trying a relative tolerance?

Thanks @adam-a-a .  Is this comment specific to the OS solver difference issue?  So I took a look and I think changing tol to `rel = 0.001`, which corresponds to about `abs = 0.0014` can make it pass here.  But I wonder if this can rule out the issue of solver failing only in Linux.  I think in the previous case the outcome value was much larger.  I can look that one back.  @lbianchi-lbl may hold off for another failure example on the issue I described above.

lbibl:
> @lbianchi-lbl:
> May I invite you to take a look at this PR in draft, which failed only in Linux but not Win or Mac? The failure was due to a solver outcome error, where ipopt was used to optimize a system with dof of 2. The inconsistency of Linux (not the experimental Mac) with Win is especially concerning. I saw this inconsistency of Linux in a previous PR (merged) and resolved that by modifying the file. But different solver outcome between Linux and Win might be worthy of an investigation. Thank you for your attention.

A previous PR failure that better reflected the above described issue is here: https://github.com/watertap-org/watertap/runs/7735049456?check_suite_focus=true 

In this PR, only Linux failed at assert 0.96==1, while Win passed.  The difference was much significant here.   I resolved it by modifying the problem set up a little.  Again, what is concerning here is the different outcomes between Linux and Win.

lbianchi-lbl:
Hi @lbibl, thanks for bringing this to my attention. If the discrepancies between different OSes are localized to the solver (as opposed to e.g. the Python code, or installation issues), then I don't think I'd be able to say much since my experience with the solver infrastructure (and non-trivial optimization problems in general) is very limited. If the evidence seems to be pointing in this direction, it seems reasonable to me to bring this up with the rest of the team to decide what the next steps could look like.

andrewlee94:
@lbibl From the look of it, the issue is due to the solver returning slightly different values on different systems, and not the execution of the code. This means that minor differences in the underlying linear algebra packages on the different OSes are likely the cause, which in turn is generally due to a poorly scaled problem.

lbibl:
Report at this point:  the changes in the DSPMDE property package, which was initiated in PR #559, is causing some inconsistencies to other files. These should not be expected since the changes only adde a few variables (`conc_equiv_phase_comp` , `viscosity_k_phase`, and `flow_equiv_phase_comp`), which should NOT be constructed to other files without being called and should not impact using other existing variables.   Initially, the inconsistencies were seen on nanofiltration, ED0d and GAC models.  A few of us consecutively addressed the errors either by modifying the unit models or improving the property model (e.g., I corrected a scaling neglect to prevent the construct of a few variables without being called) .  However, I am again seeing an inconsistency on the recently added electrodialysis flowsheet.  In addition to re-visiting scaling methods as a routine, I did notice this updated DSPMDE file makes one more variable in the optimization test than the original, caught by the model statistics.  Since there are over 1500 variables totally in that model, I cannot immediately catch what is the extra viable.  But this suggests that the changes made here are impacting the use of other variables, which we do not want.  I am going to investigate this more. Anyone having insights please suggest.   @adam-a-a  @kurbansitterley

lbibl:
@adam-a-a @kurbansitterley , hi both, here I got the DSPMDE ion property pack taken from the ix0d PR before the charge_comp scaling was added.  I believe it is compatible to the ix0d unit model @kurbansitterley is working on.  After some scaling effort to address the os inconsistency, this version is passing all files and I think is good to be merged so we can continue to add more things ( I will add a few other things for the ED refinement).  Please review when you can.  Thank you!

adam-a-a (approved):
LGTM

tarnold17 (approved):

PR #162, opened 2021-07-15
1DRO Unit Model

adam-a-a:
## Summary/Motivation:
Complete the 1DRO unit model with all detailed calculations included. Any feedback on improving functionality where possible would be much appreciated. 

Additionally, some of the newest variables in the 0DRO model (e.g., over pressure ratio) were not added in this PR; those variables can be added to this PR or a subsequent one based on reviewer feedback.

## Changes proposed in this PR:
- add all config options for concentration polarization, mass transfer coefficient, and pressure drop
- add associated testing
-  adjust initialization method and associated testing in 0dro along the way
-  add `has_full_reporting` config option and methods to compute expressions for averaged variables that will show up in a detailed report of results (via `report()`)

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #679, opened 2022-08-18
Membrane channel refactor

bknueven:
## Summary/Motivation:
Separate out various calculations in RO to "Membrane Channel" control volume.

-
-

### Legal Acknowledgement


watertap/core/membrane_channel0d.py
```diff
@@ -0,0 +1,262 @@
...
+@declare_process_block_class("MembraneChannel0DBlock")
+class MembraneChannel0DBlockData(MembraneChannelMixin, ControlVolume0DBlockData):
+
+    # overwrite CV0D `add_geometry`
+    def add_geometry(self, include_length_and_width=True):
+        """
+        Method to create spatial domain and volume Var in ControlVolume.
+
+        Args:
+            include_length_and_width - (optional) add a length and width
```
TimBartholomew: "has_length_and_width" is probably better aligned with naming conventions in IDAES

watertap/core/membrane_channel0d.py
```diff
@@ -0,0 +1,262 @@
...
+        self, information_flow=FlowDirection.forward, has_phase_equilibrium=None
+    ):
+        """
+        This method constructs the state blocks for the
+        control volume.
+
+        Args:
+            information_flow: a FlowDirection Enum indicating whether
+                               information flows from inlet-to-outlet or
+                               outlet-to-inlet
```
TimBartholomew: I'm assuming this is for co-current flow and counter-current flow, however, I'm wary of this phrasing. Flow direction always goes from inlet to outlet, its just for counter-current flow the inlet and outlet are on opposite sides.

I'll see how this is implemented, but I don't think we want a state block or port named with inlet - actually being the outlet for the counter flow.

watertap/core/membrane_channel1d.py
```diff
@@ -0,0 +1,295 @@
...
+        domain=In(DistributedVars),
+        description="Argument for defining form of area variable",
+        doc="""Argument defining whether area variable should be spatially
+variant or not. **default** - DistributedVars.uniform.
+**Valid values:** {
+DistributedVars.uniform - area does not vary across spatial domain,
+DistributedVars.variant - area can vary over the domain and is indexed
+by time and space.}""",
+    ),
+)
```
TimBartholomew, line 57: Is this a new configuration argument? I can see the utility of it, but I don't know if we have an example using it.

watertap/core/membrane_channel1d.py
```diff
@@ -0,0 +1,295 @@
...
+        description="Discretization scheme to use for DAE transformation",
+        doc="""Discretization scheme to use when transforming domain. See
+Pyomo documentation for supported schemes.""",
+    ),
+)
+
+CONFIG_Template.declare(
+    "finite_elements",
+    ConfigValue(
+        default=20,
```
TimBartholomew: This a minor comment - but we probably want to lower this default. Usually theres not much benefit going beyond 10, well honestly even 5. This figure comes from one of my papers for the error we can expect from typical operation of membrane processes compared to 100 nodes - it also includes a monte carlo across multiple parameters to define the distribution: 
image

watertap/core/membrane_channel1d.py
```diff
@@ -0,0 +1,295 @@
...
+    def add_geometry(self, flow_direction=FlowDirection.forward, **kwargs):
+        """
+        Method to create spatial domain and volume Var in ControlVolume.
+
+        Args:
+            flow_direction - argument indicating direction of material flow
+                            relative to length domain. Valid values:
+                                - FlowDirection.forward (default), flow goes
+                                  from 0 to 1.
+                                - FlowDirection.backward, flow goes from 1 to 0
```
TimBartholomew, line 138: Ok I like this better. Could we do a similar set up for the 0D model?

Replace flow_information config argument with this. I guess we would still have the issue with the control volume properties_in and properties_out having the wrong names, but the reference would be fine. I'm not sure whats the best approach.

watertap/core/membrane_channel_base.py
```diff
@@ -0,0 +1,904 @@
...
+        @self.Constraint(
+            self.flowsheet().config.time,
+            self.length_domain,
+            doc="Volumetric flow at interface of inlet",
+        )
+        def eq_equal_flow_vol_interface(b, t, x):
+            return (
+                b.properties_interface[t, x].flow_vol_phase["Liq"]
+                == b.properties[t, x].flow_vol_phase["Liq"]
+            )
```
TimBartholomew: We might want to change this wording. We aren't necessarily doing a volumetric flowrate balance, we are just providing extensive flow information to the interface. Maybe "add_extensive_flow_to_interface" would be best.

watertap/core/membrane_channel_base.py
```diff
@@ -0,0 +1,904 @@
...
+                == self.properties[t, x].conc_mass_phase_comp["Liq", j]
+                * self.cp_modulus[t, x, j]
+            )
+
+        if concentration_polarization_type == ConcentrationPolarizationType.calculated:
+            if mass_transfer_coefficient == MassTransferCoefficient.none:
+                raise ConfigurationError()
+
+            # NOTE: calculated concentration polarization is relegated
+            #       to the unit model as it depends on flux
```
TimBartholomew: Ohh interesting, so the concentration polarization equation will live in different places depending on the configuration arguments. I see why that's necessary for initialization, I'm wondering if we always just want the concentration polarization equation on the unit for it to not move based on config arguments.

watertap/unit_models/reverse_osmosis_base.py
```diff
@@ -0,0 +1,789 @@
...
+            self, "spacer_porosity", self.feed_side, "spacer_porosity"
+        )
+        _add_object_reference_if_exists(self, "N_Re", self.feed_side, "N_Re")
+        _add_object_reference_if_exists(self, "N_Re_avg", self.feed_side, "N_Re_avg")
+        _add_object_reference_if_exists(self, "N_Sc", self.feed_side, "N_Sc")
+        _add_object_reference_if_exists(self, "N_Sh", self.feed_side, "N_Sh")
+        _add_object_reference_if_exists(self, "velocity", self.feed_side, "velocity")
+        _add_object_reference_if_exists(
+            self, "friction_factor_darcy", self.feed_side, "friction_factor_darcy"
+        )
```
TimBartholomew, line 157: This is all for backwards compatibility right? I think that's a good approach for this PR, but in the near future we may want to change our tests to the new format and not support this. At least for WaterTAP we don't have many external users who would be mad about this breaking change.

watertap/unit_models/tests/test_reverse_osmosis_1D.py
```diff
@@ -308,34 +307,22 @@ def test_build(self, RO_frame):
...
-            "N_Re_avg": Expression,
-            "Kf_avg": Expression,
             "flux_mass_phase_comp_avg": Expression,
         }
-        for (obj_str, obj_type) in unit_objs_type_dict.items():
-            obj = getattr(m.fs.unit, obj_str)
-            assert isinstance(obj, obj_type)
+        # for (obj_str, obj_type) in unit_objs_type_dict.items():
+        #    obj = getattr(m.fs.unit, obj_str)
+        #    assert isinstance(obj, obj_type)
```
TimBartholomew: Yeah we can probably just get rid of these types of tests. If an expected object doesn't exist or isn't the right type, eventually we would get a pyomo error in the build.

TimBartholomew (commented):
Looks good to me. The only unexpected consequence for me with making the flux be a unit model variable is that the concentration polarization equation then has to be on the unit model. I think that's fine though.

TimBartholomew:
Oh and the other thing being the trouble with setting up the naming for counterflow operation in the 0D.

I think since the CV0D automatically creates properties_in and properties_out, maybe the best way to handle this is if it is counterflow the reference properties (to make it quack like 1D) could change whether properties_in or properties_out is assigned to 0 or 1. I'm not sure if that is your current implementation.

watertap/core/membrane_channel0d.py
andrewlee94: This look like it is referring to a feature in the core IDAES code, which allows for systems where it is more robust to solve a process in reverse order (e.g. some refrigeration cycles). It is not related (directly) to co- or counter-current flow.

I am not sure whether this feature is required here, and unless there is a need for it I would suggest it might be best to drop it.

watertap/core/membrane_channel1d.py
andrewlee94: This is also a core IDAES feature - this makes me wonder if this should be importing a standard config block from somewhere in IDAES, as it looks like it is re-creating its own.

Also, do not feel that you need to duplicate all config argument from IDAES - there are far more than you will need for any one case, and you should only implement those that you need. The underlying framework will default to the most general case if no input is given.

watertap/core/membrane_channel_base.py
andrewlee94: I would second this - having things move around makes code maintenance and testing much harder. If you can put the polarization equation in the same place across all models, it will make your life much easier in future.

bknueven: So, we could move the flux variable, but *not* the flux calculation, to the membrane channel. That would allow the concentration polarization equation to move to the membrane channel as well.

Since the flux calculation depends on the state on the "oppose side", (e.g., in the current RO units the permeate stateblock, in future OARO on another membrane channel), it necessarily has to stay on the unit model.

This gets at the heart of the "seam" question. I think adding a flux variable to the channel is a good idea, but of course it's up to the modeller to calculate it in a reasonable fashion.

@adam-a-a what do you think?

watertap/unit_models/reverse_osmosis_base.py
bknueven: Yes. I agree we should probably break it, and should be in a separate PR as it requires touching a huge number of files.

watertap/unit_models/tests/test_reverse_osmosis_1D.py
bknueven: I agree, done in 68bd79d.

watertap/core/membrane_channel0d.py
bknueven: Done in [8e9eda1](https://github.com/watertap-org/watertap/pull/679/commits/8e9eda1c8aad94d68ac8bb2e616968fdd7b2c530).

watertap/core/membrane_channel1d.py
bknueven: Done in [75001f7](https://github.com/watertap-org/watertap/pull/679/commits/75001f7d0e210eb40b208911be74cfb8b55990c8)

watertap/core/membrane_channel_base.py
bknueven: Done in [4300d89](https://github.com/watertap-org/watertap/pull/679/commits/4300d89a88728121b1e43b98bbcfedf862b6b384)

watertap/core/membrane_channel0d.py
bknueven: Right, the only reason it is here is for the call to `super().add_state_blocks` -- I gave the overloaded methods the same signature as that in the control volume. However, if we think for membrane channels we'll always use the default, I'm happy to eliminate the option.

watertap/core/membrane_channel1d.py
bknueven: I think it's a hold-over from 1DRO (which this is copied from) which wasn't used. ~~Will eliminate.~~ Removed in [115392b](https://github.com/watertap-org/watertap/pull/679/commits/115392b850bddde937b06d259997cdfe2519fdc4).

adam-a-a: I just want to note that this was added to RO1D since the area can technically vary across the spatial domain. Although `DistributedVars.variant` wasn't being used, I included this for a potential future application, if we ever want to consider variable area across the length of the membrane channel.

watertap/core/membrane_channel_base.py
```diff
@@ -0,0 +1,904 @@
...
+
+            return
+
+        if concentration_polarization_type not in (
+            ConcentrationPolarizationType.fixed,
+            ConcentrationPolarizationType.calculated,
+        ):
+            raise ConfigurationError(
+                f"Unrecognized concentration_polarization_type {concentration_polarization_type}"
+            )
```
adam-a-a, line 394: Should we get rid of this if we are adding the `eq_concentration_polarization` constraint when CP type is none?

watertap/core/membrane_channel_base.py
```diff
@@ -0,0 +1,904 @@
...
+
+        self.cp_modulus = Var(
```
adam-a-a, line 360: should there be an elif above this, basically checking that CP type is not none? If not none, then create cp_modulus

watertap/core/membrane_channel_base.py
adam-a-a: Well, I am thinking we'll want to have this on the unit model and that it would be better to have `eq_concentration_polarization` in the same place across config types. 

Moreover, `eq_concentration_polarization`, at a minimum, will differ based on the flow direction and whether the channel is facing the active side of the membrane or the porous support layer of the membrane, which I would think is decided at the unit model level. While it'd be nice to have the CP constraint for each channel control volume, I think it'd be ok to let this one slide and have it setup at the unit model level.

watertap/core/membrane_channel_base.py
bknueven: I agree that these lines are not strictly necessary.

It's just some defensive programming against somebody adding a fourth CP type and forgetting to update this method.

Or worse, passing in `"foo"` as the CP type.

adam-a-a: Nevermind- you're returning in the bit before this so this only happens when someone specifies some alternative option besides the 3 that we've defined.

watertap/core/membrane_channel_base.py
bknueven: One thing to keep in mind is the membrane channel could have different build options for these different configurations (e.g., which side of the membrane am I on, which direction is the flow going).

adam-a-a: OK, so running with that, how about handling a custom relationship for `eq_concentration_polarization`?

EDIT: Would it make a difference to the developer whether this would need to be applied in the membrane channel or unit model?

bknueven: Who is doing the customizing?

If we put the flux variables in membrane channel, you would add another option there, I think.

adam-a-a: I was just imagining if someone wanted to use a custom relationship to account for CP, e.g., some empirical relationship related to some particular module geometry. In any case, I suppose we could have an arg for custom_CP_con or something like that for the add_concentration_polarization method.

adam-a-a: So maybe we should see how things work out if we put the flux vars on the channel, even though they would technically be shared between adjacent channels.

bknueven: > EDIT: Would it make a difference to the developer whether this would need to be applied in the membrane channel or unit model?


Ah, for adding a new one, no.

But, for utilizing an existing one in a new unit model, yes. Instead of copying+pasting, you call the appropriate method (with the appropriate options) on the membrane channel.

This is the crux of what @andrewlee94 is saying above.

watertap/core/membrane_channel1d.py
bknueven: Done in 68d8ae1, which also makes the membrane channel, and by extension the RO unit model, more robust to the transformation scheme (forward basically wasn't supported before).

watertap/unit_models/reverse_osmosis_0D.py
```diff
@@ -24,55 +24,43 @@
...
-        # for quacking like 1D model -> 0. is "in", 1. is "out"
-        self.length_domain = Set(
-            ordered=True, initialize=(0.0, 1.0)
-        )  # inlet/outlet set
-        add_object_reference(self, "difference_elements", self.length_domain)
-        self.first_element = self.length_domain.first()
-
-        # Build control volume for feed side
-        self.feed_side = ControlVolume0DBlock(
+    def _add_feed_side_membrane_channel_and_geometery(self):
```
adam-a-a:
```suggestion
    def _add_feed_side_membrane_channel_and_geometry(self):
```

watertap/unit_models/reverse_osmosis_1D.py
```diff
@@ -134,28 +79,15 @@ def _process_config(self):
...
-        Args:
-            None
-
-        Returns:
-            None
-        """
-        # Call UnitModel.build to setup dynamics
-        super().build()
-
+    def _add_feed_side_membrane_channel_and_geometery(self):
```
adam-a-a:
```suggestion
    def _add_feed_side_membrane_channel_and_geometry(self):
```

watertap/unit_models/reverse_osmosis_1D.py
```diff
@@ -134,28 +79,15 @@ def _process_config(self):
...
         self._process_config()
 
-        # Build 1D Control volume for feed side
-        self.feed_side = feed_side = ControlVolume1DBlock(
+        # Build 1D Membrane Channel
+        self.feed_side = MembraneChannel1DBlock(
             default={
                 "dynamic": self.config.dynamic,
                 "has_holdup": self.config.has_holdup,
-                "area_definition": self.config.area_definition,
```
adam-a-a: I don't think there was any harm in leaving the area definition since a variable area could come into play in the future.

watertap/core/membrane_channel_base.py
```diff
@@ -0,0 +1,928 @@
...
+class MembraneChannelMixin:
+    def _add_pressure_change(self, pressure_change_type=PressureChangeType.calculated):
+        raise NotImplementedError()
+
+    def _skip_element(self, x):
+        raise NotImplementedError()
+
+    def _add_var_reference(self, pyomo_var, reference_name, param_name):
+        if pyomo_var is not None:
+            # Validate length_Var and add a reference
```
adam-a-a: Should we delete this comment or did you want to keep as an example?

watertap/examples/flowsheets/lsrro/lsrro.py
```diff
@@ -762,7 +763,7 @@ def do_backward_initialization_pass(m, optarg):
     for stage in reversed(m.fs.NonFinalStages):
         m.fs.Mixers[stage].initialize(optarg=optarg)
         propagate_state(m.fs.mixer_to_stage[stage])
-        m.fs.ROUnits[stage].initialize(optarg=optarg)
+        m.fs.ROUnits[stage].initialize(optarg=optarg, raise_on_failure=False)
```
adam-a-a, line 766: Why did we need to add these `raise_on_failure=False` args? Did something affect LSRRO initialization? Or was this for debugging purposes?

watertap/unit_models/reverse_osmosis_base.py
```diff
@@ -0,0 +1,822 @@
...
+            raise ConfigurationError(
+                "Membrane models only support one liquid phase ['Liq'],"
+                "the property package has specified the following phases {}".format(
+                    [p for p in self.config.property_package.phase_list]
+                )
+            )
+
+        validate_membrane_config_args(self)
+
+        self._add_feed_side_membrane_channel_and_geometery()
```
adam-a-a:
```suggestion
        self._add_feed_side_membrane_channel_and_geometry()
```

watertap/core/membrane_channel0d.py
TimBartholomew: Oh its not related to co- or counter-current flow? I think it can be eliminated then, although I don't know anything about the benefits of solving a process in reverse order (e.g. Andrew's comment about refrigeration cycles).

watertap/core/membrane_channel0d.py
```diff
@@ -0,0 +1,298 @@
...
+        super().calculate_scaling_factors()
+
+        if hasattr(self, "area"):
+            if iscale.get_scaling_factor(self.area) is None:
+                iscale.set_scaling_factor(self.area, 100)
+
+        if hasattr(self, "dP_dx"):
+            for v in self.dP_dx.values():
+                if iscale.get_scaling_factor(v) is None:
+                    iscale.set_scaling_factor(v, 1e-4)
```
TimBartholomew, line 277: Minor comment and it shouldn't hold up the PR, but maybe we want to include a warning that the area scaling factor is being set by default. dp_dx probably doesn't need this warning because its an intensive variable and this is probably a good scaling factor for all practical uses of RO.

watertap/core/membrane_channel1d.py
TimBartholomew: Wait, I may have misinterpreted this. In my original review I thought this flow_direction was actually the direction of flow and could be used by us to set co-current or counter-current for OARO (i.e. the permeate channel would have backward flow). But Andrew's comment suggest that this is just the direction of the solve rather than the actual flow. I'm not sure, but we can figure this all out with implementing the OARO model.

watertap/core/membrane_channel0d.py
andrewlee94: This feature has never actually been used in IDAES - it was added based on a suggestion/request from a team member for future applications.

watertap/core/membrane_channel1d.py
```diff
@@ -0,0 +1,277 @@
...
+            for v in self.pressure_change_total.values():
+                if iscale.get_scaling_factor(v) is None:
+                    iscale.set_scaling_factor(v, 1e-4)
+
+        if hasattr(self, "dP_dx"):
+            for v in self.pressure_dx.values():
+                iscale.set_scaling_factor(v, 1e-5)
+        else:
+            for v in self.pressure_dx.values():
+                iscale.set_scaling_factor(v, 1e5)
```
TimBartholomew: At first I thought the difference here was because of the naming conventions of the 1D model. But I'm confused, should for the 0D and 1D model this is the pressure loss per unit length? Shouldn't they have the same scaling factor?

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
```diff
@@ -274,6 +274,8 @@ def set_operating_conditions(
         m.fs.P1.control_volume.properties_out[0].pressure
     )
 
+    m.fs.RO.area.fix(50)  # guess area for RO initialization
```
TimBartholomew, line 277: Were we previously initializing with 1 degree of freedom? Oops :)

watertap/unit_models/reverse_osmosis_base.py
```diff
@@ -0,0 +1,822 @@
...
+            def eq_concentration_polarization(b, t, x, j):
+                jw = b.flux_mass_phase_comp[t, x, "Liq", "H2O"] / self.dens_solvent
+                js = b.flux_mass_phase_comp[t, x, "Liq", j]
+                return b.feed_side.properties_interface[t, x].conc_mass_phase_comp[
+                    "Liq", j
+                ] == b.feed_side.properties[t, x].conc_mass_phase_comp["Liq", j] * exp(
+                    jw / self.feed_side.K[t, x, j]
+                ) - js / jw * (
+                    exp(jw / self.feed_side.K[t, x, j]) - 1
+                )
```
TimBartholomew, line 441: Ok so this concentration polarization constraint is added on the unit model and the constraint for the other concentration polarization types (i.e. fixed, none) are on the membrane channel? I don't know if there is a better solution.

What's the downside with moving the constraints for the other concentration polarization types to the unit model? I understand that the channel initialization won't have concentration polarization constraint when it could have for fixed and none options, but for the calculated option the channel initialization we would have to assume the interface is at the same state as the bulk - which is also fine for the fixed and none options.

TimBartholomew (commented):
A few comments, I think we just need a quick discussion on the location of the concentration polarization constraints but otherwise I think this is good.

watertap/core/membrane_channel1d.py
andrewlee94: @TimBartholomew Note that there are two similarly named argument - `flow_direction` and `information_flow`. `flow_direction` is use for the physical flow (basically does the `d/dx` term need a -1 in front of it).

watertap/unit_models/reverse_osmosis_base.py
bknueven: @adam-a-a discussed this and thought this was the best, though not great, solution.

Putting the flux on the channel doesn't make much sense as it is the mass transfer term. This constraint is also different on different sides of the channel. However, all the other concentration polarization equations are the same (as far as I understand) on both sides of the channel, so conceptually this separation makes some sense. In any case, we now create a `cp_modulus` Var in every case except `none`, so in some ways the channel doesn't really care beyond "create cp_modulus or not" -- and this constraint serves to calculate it.

We could put them all the unit model; however, note that any equations we put on the unit model need to be re-implemented in each unit.

watertap/unit_models/tests/test_reverse_osmosis_0D.py
```diff
@@ -261,68 +264,8 @@ def test_build(self, RO_frame):
...
-        cv_objs_type_dict = {
-            "eq_concentration_polarization": Constraint,
-            "eq_equal_temp_interface": Constraint,
-            "eq_equal_pressure_interface": Constraint,
-            "eq_equal_flow_vol_interface": Constraint,
-        }
-        for (obj_str, obj_type) in cv_objs_type_dict.items():
-            obj = getattr(m.fs.unit.feed_side, obj_str)
-            assert isinstance(obj, obj_type)
+        assert isinstance(m.fs.unit.feed_side, MembraneChannel0DBlock)
```
adam-a-a, line 267: Should the test also be checking for `eq_concentration_polarization` since that is built at unit model level in ro base, only when CP type is calculated? Also, I am guessing assertions on the cp_modulus constraint would be for the future membrane channel tests

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
bknueven: This snuck is as part of #690, so it hasn't been there long ...

watertap/core/membrane_channel1d.py
bknueven: We should probably think through this -- it's just copied from the analogous code in the current 1DRO model: 

https://github.com/watertap-org/watertap/blob/2d2a8fdca20acc0a343433afa511396fd788a91f/watertap/unit_models/reverse_osmosis_1D.py#L561-L566

watertap/unit_models/reverse_osmosis_base.py
andrewlee94: Does this suggest we should have a common utility function/method for creating these constraints at the unit level then (e.g. on a common base class for the models)?

adam-a-a (commented):
I'm glad the tests work out thus far and I think we'll be able to start OARO soon. I will say though that the code is a bit trickier to navigate now, just in terms of knowing where things live and the order in which they connect. Hopefully, it will just be a matter of getting used to the changes.

watertap/unit_models/reverse_osmosis_base.py
adam-a-a: @TimBartholomew I think we now only have this `eq_concentration_polarization` at the unit level when CP type is calculated. When none or fixed, we now only have a cp_modulus constraint in the membrane channel CV which uses the cp_modulus to get interface concentration. When CP type is none, concentrations at bulk and interface are set equal using `eq_cp_modulus`.

adam-a-a: I think what we have now is fine actually. In the membrane channel CV, we just use a cp_modulus constraint, which makes sense from a channel perspective (i.e., each channel will have its own cp_modulus w/respect to length across the channel, and this is just the ratio of interface to bulk concentration). 

However, the constraint for calculating the membrane interface concentration can take on different forms, depending on whether we are using a first principles relationship, an empirical relationship, or depending on _which_ channel and flow configuration we are considering. Thus, it makes sense for any detailed calculation for CP to be at the unit model; meanwhile, we do no such calculation at the channel and merely compute interface concentration by an assumed ratio. 

Overall, we no longer have the same constraint living in either the unit model or channel model when switching between Enum types for concentration polarization. `eq_concentration_polarization` only lives on the unit model and `eq_cp_modulus` only lives on the channel CV.

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
adam-a-a: Yeah, there probably should be tests for the RO/ERD flowsheet to prevent that in the future.

watertap/unit_models/reverse_osmosis_base.py
bknueven: Always catching our spelling mistakes :-)

watertap/examples/flowsheets/lsrro/lsrro.py
bknueven: This was essentially the default behavior of the RO model before, which this PR changes to be more IDAES compliant. Note the `fail_on_warning` flag:
https://github.com/watertap-org/watertap/blob/2d2a8fdca20acc0a343433afa511396fd788a91f/watertap/unit_models/reverse_osmosis_0D.py#L353-L362
https://github.com/watertap-org/watertap/blob/2d2a8fdca20acc0a343433afa511396fd788a91f/watertap/unit_models/reverse_osmosis_1D.py#L407-L416

In any case, for LSRRO sometimes these units fail to initialize on the first or second pass, especially at high salinities where the first-stage probably *is* infeasible. So nothing has really changed besides the default RO initialization flag, which I think is more correct in this PR.

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
bknueven: The RO initialization will now fail if the degrees of freedom aren't 0, which it wasn't doing before. (Which is how I found this.)

watertap/core/membrane_channel0d.py
bknueven: I've remove it as part of 495fbfa9

watertap/core/membrane_channel1d.py
adam-a-a: So I traced this back to PR #162, back in July of last year. I have no memory of why I did this, but if I had to guess, it was likely out of desperation when toggling between `has_pressure_drop` True or False and running into some scaling issue at the time. It was my 4th month so probably an unnecessary inclusion.

watertap/unit_models/reverse_osmosis_1D.py
bknueven: I've reverted the removal of this in 495fbfa.

watertap/core/membrane_channel1d.py
bknueven: @adam-a-a would like to maintain this, so I reverted the change that removed it. I don't have strong feelings either way.

watertap/core/membrane_channel1d.py
bknueven: You are correct -- `dP_dx` will only be created if there's a pressure change.

In the event that you do not, I'm guessing these `pressure_dx` variables will be small, hence the scaling factor. I'm inclined to leave it as-is with an additional comment. Thoughts?

watertap/core/membrane_channel_base.py
bknueven: I updated the comment, which got find-replace mangled.

watertap/unit_models/tests/test_reverse_osmosis_0D.py
bknueven: I think @TimBartholomew and I had a conversation about removing tests for specific components broadly, and relying more on performance values.

I'm happy to add these back in for those components that still live on the RO model, though I often find they're a bit more trouble than they're worth -- especially those checking specific constraint names, which don't particularly affect the API of the model.

watertap/core/membrane_channel0d.py
bknueven: Interestingly we don't do this currently -- note that this is the cross-sectional area in this context (to match the name with 1DCV).

TimBartholomew (approved):

bknueven:
> I will say though that the code is a bit trickier to navigate now, just in terms of knowing where things live and the order in which they connect. Hopefully, it will just be a matter of getting used to the changes.

FWIW I agree -- the benefit will come in (1) more rapid development of unit models and (2) consolidating our other existing membrane models to extend and use the membrane channel CV.

watertap/core/membrane_channel1d.py
adam-a-a: An additional comment would be helpful if we leave it in. However, I am wondering if we should even have to do this or not. That is, is `feed_side.pressure_dx` unintentionally being created somewhere by us (e.g., faulty if statement logic), or is this just how things are setup in ControlVolume1D? If the latter, I think it's fine to keep; if the former, maybe we should provide a fix.

bknueven:
Good point — I will look into it

watertap/core/membrane_channel1d.py
bknueven: Removed in https://github.com/watertap-org/watertap/pull/679/commits/a86e13148826e885990157b680b3ac6a22b9eaa2.

Based on the way the 1DCV is set-up, I think we still need these variables to enforce the isobaric assumption when there's no pressure change. Or rather, we could get rid of them but we would need to enforce that constraint ourselves -- does not seem worth the trouble.

adam-a-a (approved):
I think we can merge this and if anything needs to be resolved, we can address issues as OARO implementation is pursued. Furthermore, we should eventually include tests for membrane_channel classes.

PR #682, opened 2022-08-20
Require do_xxx methods for API

dangunter:
## Fixes/Resolves:

See https://github.com/watertap-org/watertap-ui/issues/37

## Summary/Motivation:
Needed for https://github.com/watertap-org/watertap-ui/pull/36

- fail immediately if `do_build`, `do_export`, or `do_solve` are not provided when building the flowsheet interface

### Legal Acknowledgement

### DIFF
--- a/examples/ui/api_example.py
+++ b/examples/ui/api_example.py
@@ -1,16 +1,20 @@
 from watertap.ui.fsapi import FlowsheetInterface
+from idaes.core.solvers import get_solver
 
 
 def export_to_ui():
     return FlowsheetInterface(
-        do_export=export_variables, do_build=build_flowsheet, do_solve=solve_flowsheet
+        name="foo",
+        do_export=export_variables,
+        do_build=build_flowsheet,
+        do_solve=solve_flowsheet,
     )
 
 
 def export_variables(flowsheet=None, exports=None):
     for n in 1, 2:
         t = getattr(flowsheet, f"Tank{n}")
-        exports.add(obj=t.inlet.flow_vol[0], name="Tank {n} inlet flowrate")
+        exports.add(obj=t.inlet.flow_vol[0], name=f"Tank {n} inlet flowrate")
     # ..etc..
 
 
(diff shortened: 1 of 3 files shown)


MichaelPesce (approved):

dangunter:
So, @lbianchi-lbl  I'm confused by the codecov/patch requirements. How is it 94.89%? Yes, I missed testing a few lines, but the project diff is -0.02% so really this shouldn't be blocking a merge. And I would expect a target of 80% perhaps?

lbianchi-lbl:
> So, @lbianchi-lbl I'm confused by the codecov/patch requirements. How is it 94.89%? Yes, I missed testing a few lines, but the project diff is -0.02% so really this shouldn't be blocking a merge. And I would expect a target of 80% perhaps?

:man_shrugging: I'm not sure where the 40% (current at the time of writing) comes from, I don't know exactly step-by-step how to reproduce the Codecov calculations (especially for patch coverage). The current settings for the `patch` coverage is the default (i.e. the current project) with 10% tolerance.

In general I really don't have a preference in terms of numbers, and (personal opinion/experience) the `patch` check seems to be quite noisy and opaque so I'd be in favor of considering loosening the tolerance substantially or even removing it. Conversely, I'd suggest tightening the tolerance for the `project` check that (again, IMO) is the one that we should try to enforce.

Even more in general one might argue that the concept of enforcing coverage is relatively moot for parts of the codebase with a single "designated owner" such as `watertap/ui` (i.e. where there's a 1:1 mapping between the person submitting a PR and the person who would later be responsible for fixing any potential bug)... all of this to say, if the coverage check fails but you feel that the code is well-covered, I personally have nothing against red-button-merging the PR.

dangunter:
So, I think the codecov patch summary doesn't communicate the threshold well; it simply mentions the target. That's a minor thing. But  I also  would be in favor of dropping the patch requirement and tightening the project requirement. We can keep patch reporting for people who want to see it. At a deeper level, I wonder if mypy (et al.) could replace some of our reliance on manual unit tests to "cover" all the code.

At any rate, this code is now even more tested .. can you look at it when you get a chance

andrewlee94:
@dangunter I think the issue is that we have it set so that coverage cannot decrease, rather than a simple >x%. This was useful in IDAES where we had really low test coverage to start and wanted to gradually push it upwards, but here we are already testing at a very high level.

lbianchi-lbl (approved):
I don't know much about the context but the changes LGTM.

Issue #676, opened 2022-08-18
WaterTAP costing improvements


TimBartholomew:
There are a few potential improvements that I've been thinking about for WaterTAP costing:

- [x] Moving the creation of *global* unit costing parameters (typically fixed variables) from the build_global_params in the costing package to the unit costing methods (Done in #835 for WaterTAPCosting Package, ~~TODO for ZeroOrderCosting Package~~) (Working on #911 I realized the ZO Costing package has *always* done this.)

The benefits would include:
- Avoiding the creation of the parameters if the unit is not costed on the flowsheet, currently all of the costing parameters get added regardless of whether the unit exists. The issue is that when someone displays the costing parameter block there are many unused parameters and is confusing for the user - this will only grow as we add more units.
- The unit costing method could create a block with the unit name to the costing package parameter block, this would help with the name space and organization of the parameters

- [ ] Moving the unit costing methods to their unit models (Started in #835 for WaterTAPCosting Package, DONE in #911 for ZeroOrderCosting Package, Completed in #1122)

The benefits would include:
- Clearer separation of the system costing and the unit costing, this also helps with documentation. The unit costing could have their own technical reference page.
- Making it easier for users to see what is required for creating a custom unit costing because each unit costing file is essentially an example.
- It would make it more clear to users if we have multiple costing methods for a unit, they could see multiple costing files and import them as needed as opposed to configuration arguments that I think is more convoluted

- [ ]  Documentation (e.g., #723) (Completed in #1122, moved to #1135)
- [ ]  #710 (Completed in #1122)
- [x] Allow user-settable `base_currency`, and hard-code all parameters to use a fixed year. (Done in #835)
- [x] Rename `Var` / `Param`s for flow cost (e.g., `electricity_base_cost`) to `{flow_type}_cost` (e.g., `electricity_cost`) using IDAES/idaes-pse#983; (#865)
- [x] Consider combining the two costing packages into a single package (moved to #1134)

andrewlee94:
As a comment regarding point 1: consider whether these parameters are something that should be varied globally (e.g. during sensitivity studies). If they are something that should be equal across all units, you really want them in a single, centralized location, otherwise the parameter can be varied locally for each unit resulting in inconsistencies.

bknueven:
[...]

PR #684, opened 2022-08-22
Ion exchange 0D single component model

kurbansitterley:
## Fixes/Resolves:

Replace PR #559 

## Summary/Motivation:

Replace PR #559 

- Single component Ion exchange 0D model
- test file tested on Windows machine
- Add IX costing method

### Legal Acknowledgement


TimBartholomew:
@lbibl in some of these tests, electrodialysis model tests are failing, but the files changed shouldn't be impacting them. So it seems that the ED model tests are not stable.

@bknueven in some of these tests, LSRRO multisweep test are failing, but the files changed shouldn't be impacting them. So it seems that that test is not stable.

Kurby is going to try to move the costing parameters and variables, which are the only thing that could be impacting those flowsheets.

bknueven:
> @bknueven in some of these tests, LSRRO multisweep test are failing, but the files changed shouldn't be impacting them. So it seems that that test is not stable.

For the LSRRO models, we made this decision to test the "4 corners" of the search space we explored. However, in retrospect I think that is a mistake, because it leads to somewhat ill-posed problems. E.g., optimizing a 4- or 5-stage LSRRO flowsheet against a low feed concentration leads to a problem with lots of local minima. We should probably be using test values that make operational sense. @adam-a-a thoughts? I'm opening an issue for this -- we also experienced this after making a different change -- the LSRRO flowsheet solved to a different local minima in one or more cases.

lbibl:
I can recognize the ED flowsheet (not the unit model) failure in certain OS the same one I tackled in the last PR.   I believe it is also caused by local minima of the problem itself.  My previous solution was only to tailor the scaling to let all platforms land on the same minimum.   The optimization problem in ED flowsheet tested at this stage is more arbitrary than carefully-designed, so I expect a better problem design to update the current version as a permanent solution.  For now,  I think we can 1) find out why the changes in the costing package caused the effect and change there, 2) tailor the scaling again as a temporary solution, or 3) disable the optimization test in the ED flowsheet temporarily.

lbibl:
I saw the updated costing package added 18 variables and 2 constraints to the optimization problem in the ED flowsheet, where the obj is LCOW.  I am not yet that familiar with the costing package.  Can someone tell if this is reasonable?

lbibl:
> I saw the updated costing package added 18 variables and 2 constraints to the optimization problem in the ED flowsheet, where the obj is LCOW. I am not yet that familiar with the costing package. Can someone tell if this is reasonable?

It seems Issue #676  is related to my question.

kurbansitterley:
> I saw the updated costing package added 18 variables and 2 constraints to the optimization problem in the ED flowsheet, where the obj is LCOW. I am not yet that familiar with the costing package. Can someone tell if this is reasonable?

IX costing added 18 variables so that tracks. Not totally sure where the two constraints could come from, but I did add two `defined_flows`.

bknueven:
> Kurby is going to try to move the costing parameters and variables, which are the only thing that could be impacting those flowsheets.

@TimBartholomew I think this is unnecessary with IDAES/idaes-pse#938, and it would be better to make such changes as part of #676.

lbibl:
PR #696 , once merged, is expected to resolve the error in the /flowsheet/electrodialysis/tests/test_electrodialysis_1stack.py.

adam-a-a (approved):
LGTM - I think any further modification can be handled in a subsequent PR, aside from getting the tests to pass, which I am assuming will end up being simple fixes.

bknueven (commented):
@kurbansitterley I know we want to merge this yesterday -- but any chance we could get a simple test for the costing code?

We've been including that with the unit model tests for ease with other units.

bknueven (approved):

PR #686, opened 2022-08-23
Lazy flow costing in WaterTAP Costing Package

bknueven:
## Summary/Motivation:
The costing package should only cost flows which appear on the flowsheet, not all those possible. Work towards resolving the issues with #684; complementary to IDAES/idaes-pse#938.

- The watertap costing package will now only track the flows for which `cost_flow` has been called, eliminating some unneeded variables (and constraints) from the costing package.
- Minor updates in full_treatment_train (including a single baseline update) for this change.

### Legal Acknowledgement


watertap/costing/watertap_costing_package.py
```diff
@@ -400,10 +400,44 @@ def build_global_params(self):
...
+        # is called for a flow. The `available_flows`
+        # are all the flows which can be costed with
+        # the WaterTAP costing package.
+        self.available_flows = {}
+        self.available_flows["electricity"] = self.electricity_base_cost
+        self.available_flows["NaOCl"] = self.naocl_cost / self.naocl_purity
+        self.available_flows["CaOH2"] = self.caoh2_cost / self.caoh2_purity
+        self.available_flows["steam"] = self.steam_unit_cost
+
+    def cost_flow(self, flow_expr, flow_type):
```
andrewlee94, line 511: Do we have unit tests that cover this feature?

watertap/examples/flowsheets/lsrro/tests/test_lssro_multi_sweep.py
```diff
@@ -23,7 +23,9 @@
 
 _this_file_path = os.path.dirname(os.path.abspath(__file__))
 
-_test_cases = list(range(1, 6))
+# TODO: consider changing back to 6
+# so we run 1 - 5 as opposed to 1 - 3
+_test_cases = list(range(1, 4))
```
adam-a-a: Why was this bit necessary? I'm guessing it is to avoid test failures in other PRs that we've occasionally identified but thought that your changes in the costing package would've been enough.

bknueven: These tests fail with this change to the costing package, because it makes very small changes to the overall problem, i.e., it removes a few variables which were fixed to 0 by a constraint.

bknueven:
~~Closing for now -- let's try to get IDAES/idaes-pse#938 in instead.~~

Re-opening. IDAES/idaes-pse#938 is merged but missed the 2.0.0a3 release.

watertap/costing/watertap_costing_package.py
bknueven: I have now added one in 33d086ef.

watertap/examples/flowsheets/lsrro/tests/test_lssro_multi_sweep.py
bknueven: Resolved in #701.

bknueven:
~~With #696 and #701, this is now a "nice to have".~~

Still seems to be needed for #684.

bknueven:
@lbibl can you have a look at the test failures here? Maybe we should revert #696 because it's not obvious it's helped, especially given Linux is passing.

lbibl:
> @lbibl can you have a look at the test failures here? Maybe we should revert #696 because it's not obvious it's helped, especially given Linux is passing.

For information, I saw this PR reduced the total var # in the ED flowsheet cost optimization by 6.  I assume this is a correct reflection from the changes made to the costing package.    Thanks @bknueven for making the temporary solution and  hope it'll work.  I will follow up with a better resolution on that file in the future.

lbibl (approved):

andrewlee94 (approved):

PR #689, opened 2022-08-25
Moving unit-costing parameters to their own block

bknueven:
## Summary/Motivation:
The WaterTAP costing package has too many parameters crowding the namespace. This puts the parameters for each unit on their own block.

- Reorganize costing parameters in WaterTAP Costing Package
- Touch a few affected tests

### Legal Acknowledgement

### DIFF
--- a/watertap/costing/watertap_costing_package.py
+++ b/watertap/costing/watertap_costing_package.py
@@ -166,7 +166,7 @@ class WaterTAPCostingData(FlowsheetCostingBlockData):
                 units=pyo.units.year**-1,
             )
             blk.reactor_cost = pyo.Var(
-                initialize=172.158,
+                initialize=202.346,
                 doc="UV reactor cost",
                 units=costing.base_currency / (pyo.units.m**3 / pyo.units.hr),
             )
--- a/watertap/costing/watertap_costing_package.py
+++ b/watertap/costing/watertap_costing_package.py
@@ -88,7 +88,7 @@ class WaterTAPCostingData(FlowsheetCostingBlockData):
         self.base_period = pyo.units.year
 
         # Build flowsheet level costing components
-        # This is package specific
+        # These are the global parameters
         self.utilization_factor = pyo.Var(
             initialize=0.9,
             doc="Plant capacity utilization [fraction of uptime]",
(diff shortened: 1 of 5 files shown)


andrewlee94 (commented):
One comment that we could put as a feature for the future, but looks good.

watertap/costing/watertap_costing_package.py
```diff
@@ -1140,10 +1232,10 @@ def cost_gac(blk):
             expr=blk.gac_makeup_cost
             == pyo.units.convert(
                 (
-                    blk.costing_package.gac_makeup_unit_cost
+                    blk.costing_package.gac.makeup_unit_cost
                     * (
-                        (1 - blk.costing_package.gac_regen_frac)
-                        * blk.unit_model.gac_mass_replace_rate
+                        (1 - blk.costing_package.gac.regen_frac)
+                        * blk.unit_model.gac.mass_replace_rate
```
hunterbarber:
```suggestion
                        * blk.unit_model.gac_mass_replace_rate
```

A unit_model variable with the prefix `gac_` got changed when find and replacing `costing.gac_` variables

bknueven:
> One comment that we could put as a feature for the future, but looks good.

@andrewlee94 I'm not seeing another comment from you on this PR. Curious what your thoughts are.

In line with #676, my next thing to-do was to add the mechanism to construct these blocks on-demand.

andrewlee94:
@bknueven I must have forgotten to submit the comment - I was just noting that I think we could easily set it up so that the unit-specific sub-methods are only triggered the first time a unit of that type is constructed. This would help keep down the number of unused parameters (and Blocks) required.

hunterbarber (approved):
Assuming the "add the mechanism to construct these blocks on-demand" will be in subsequent PRs, this first step of moving parameters looks good to me.

bknueven:
> Assuming the "add the mechanism to construct these blocks on-demand" will be in subsequent PRs, this first step of moving parameters looks good to me.

Yes, I will do that subsequently. This is ready to merge as-is, IMHO.

adam-a-a (approved):

PR #690, opened 2022-08-26
Reverse Osmosis example pump-as-turbine configuration 

arao53:
## Fixes/Resolves:

Replicates the pump-as-turbine configuration from the seawater reverse osmosis treatment train into a process-only flowsheet.

## Summary/Motivation:
Allows for the use of an alternative and simpler system configuration of reverse osmosis, separate from the full treatment train that is modeled in the `seawater_RO_desalination.py` case study

- Add option to specify `erd_type` as `pump_as_turbine` or `pressure_exchanger` (default)
- Modify `build`, `set_operating_conditions`, `initialize_system`, and `display` to consider `erd_type = pump_as_turbine`

### Legal Acknowledgement

### DIFF
--- a/watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
+++ b/watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
@@ -33,6 +33,7 @@ from idaes.models.unit_models.mixer import MomentumMixingType
 from idaes.core import UnitModelCostingBlock
 import idaes.core.util.scaling as iscale
 import idaes.logger as idaeslog
+from idaes.core.util.misc import StrEnum
 
 import watertap.property_models.NaCl_prop_pack as props
 from watertap.unit_models.reverse_osmosis_0D import (
@@ -47,7 +48,12 @@ from watertap.core.util.initialization import assert_degrees_of_freedom
 from watertap.costing import WaterTAPCosting
 
 
-def main(erd_type="pressure_exchanger"):
+class ERDtype(StrEnum):
+    pressure_exchanger = "pressure_exchanger"
+    pump_as_turbine = "pump_as_turbine"
+
+
+def main(erd_type=ERDtype.pressure_exchanger):
     # set up solver
     solver = get_solver()
 
@@ -63,7 +69,7 @@ def main(erd_type="pressure_exchanger"):
     print("\n***---Simulation results---***")
     display_system(m)
     display_design(m)
-    if erd_type == "pressure_exchanger":
+    if erd_type == ERDtype.pressure_exchanger:
         display_state(m)
     else:
         pass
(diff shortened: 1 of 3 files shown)


watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
```diff
@@ -40,58 +42,80 @@
...
-    display_design(m)
-    display_state(m)
+    return m
 
 
-def build():
+def build(erd_type="pressure_exchanger"):
     # flowsheet set up
     m = ConcreteModel()
+    m.erd_type = erd_type
```
adam-a-a: Why save this to the model?

adam-a-a: I think I see why you did this--to avoid having to pass args to subsequent functions. This is trivial but maybe assign this as `m.fs._erd_type = erd_type`.

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
```diff
@@ -40,58 +42,80 @@
...
-    m.fs.P2 = Pump(default={"property_package": m.fs.properties})
-    m.fs.M1 = Mixer(
-        default={
-            "property_package": m.fs.properties,
-            "momentum_mixing_type": MomentumMixingType.equality,  # booster pump will match pressure
-            "inlet_list": ["P1", "P2"],
-        }
-    )
+
+    if erd_type == "pressure_exchanger":
```
adam-a-a: Might be overkill, but maybe we could implement as an enumerated class

bknueven: Specifically, a StrEnum for ease: `idaes.core.util.misc.StrEnum`. Probably best practice on flowsheets, though we obviously don't do this everywhere else.

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
```diff
@@ -307,65 +350,78 @@ def initialize_system(m, solver=None):
...
-
-    # ---initialize pump 2---
-    propagate_state(m.fs.s09)
-    m.fs.P2.control_volume.properties_out[0].pressure.fix(
-        value(m.fs.P2.control_volume.properties_out[0].pressure)
-    )
-    m.fs.P2.initialize(optarg=optarg)
-    m.fs.P2.control_volume.properties_out[0].pressure.unfix()
+    if m.erd_type == "pressure_exchanger":
+        # ---initialize splitter and pressure exchanger---
```
bknueven: Should we just write a separate initialize function for each configuration, and call the appropriate function?

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
```diff
@@ -104,38 +128,44 @@ def build():
     m.fs.product = Product(default={"property_package": m.fs.properties})
     m.fs.disposal = Product(default={"property_package": m.fs.properties})
 
-    # costing
+    # costing for all configurations
     m.fs.P1.costing = UnitModelCostingBlock(
         default={"flowsheet_costing_block": m.fs.costing}
     )
```
bknueven: If we're going to create the costing blocks immediately after the unit, let's just do that for `P1` too.

bknueven (commented):
I'm a bit worried about how adding a configuration argument affects readability.

This used to be one of our simpler, but not trivial, examples. I've made a few suggestions on the readability front, I'm guessing there's probably more that could be done.

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
arao53: Now implemented the configuration choice using StrEnum.

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
arao53: This has been addressed by reorganizing the flowsheet to create/initialize the main pump, RO, and then an option for the ERD configuration. For initialization, the option now calls `initialize_pressure_exchanger` or `initialize_pump_as_turbine`.

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
```diff
@@ -146,24 +189,38 @@ def build():
...
+    m,
+    water_recovery=0.5,
+    over_pressure=0.3,
+    solver=None,
+):
+
     if solver is None:
         solver = get_solver()
+    else:
+        pass
```
bknueven:
```suggestion
```
Python doesn't require this, and in this case it hurts readability, IMHO.

This is true a few other places as well.

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
```diff
@@ -101,41 +106,79 @@ def build():
...
+        # add costing for ERD config
+        m.fs.ERD.costing = UnitModelCostingBlock(
+            default={"flowsheet_costing_block": m.fs.costing}
+        )
+    else:
+        raise NotImplementedError(
+            "erd_type was {}, but can only "
+            "be pressure_exchanger or pump_as_turbine"
+            "".format(erd_type)
+        )
```
bknueven: So, I half wonder if we move this to a function and call that function here and everywhere else you have a `pass`.

If somebody makes further modifications to the flowsheet, that would be safer.

bknueven (approved):
The readability is improved, thanks.

I have a few more nits, but they're not critical.

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
adam-a-a: I'll add that others prefer for this to be included for the sake of being explicit, but it doesn't make too much difference to me.

adam-a-a (approved):
LGTM

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
arao53: Added a function and test for this.

PR #694, opened 2022-08-26
Two-Stage Reverse Osmosis Flowsheet

MarcusHolly:
## Summary/Motivation:

This flowsheet creates a baseline case for 2-stage RO with energy recovery. The end goal is to eventually develop a 2-stage RO flowsheet with energy recovery that also utilizes a boron removal unit model. 

- Expands upon the following flowsheet example: https://github.com/watertap-org/watertap/blob/main/watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py

### Legal Acknowledgement

### DIFF
--- a/watertap/examples/flowsheets/RO_with_energy_recovery/two_stage_RO_with_energy_recovery.py
+++ b/watertap/examples/flowsheets/RO_with_energy_recovery/two_stage_RO_with_energy_recovery.py
@@ -42,6 +42,7 @@ from watertap.unit_models.reverse_osmosis_0D import (
 from watertap.unit_models.pressure_exchanger import PressureExchanger
 from watertap.unit_models.pressure_changer import Pump
 from watertap.core.util.initialization import assert_degrees_of_freedom
+from watertap.core.util.infeasible import print_infeasible_constraints, print_close_to_bounds, print_infeasible_bounds
 from watertap.costing import WaterTAPCosting
 
 
@@ -174,7 +175,7 @@ def build():
     iscale.set_scaling_factor(m.fs.P3.control_volume.work, 1e-3)
     iscale.set_scaling_factor(m.fs.RO2.area, 1e1)
 
-    # touch properties used in specifying and initializing the model (1st stage)
+    # touch properties used in specifying and initializing the model
     m.fs.feed.properties[0].flow_vol_phase["Liq"]
     m.fs.feed.properties[0].mass_frac_phase_comp["Liq", "NaCl"]
     m.fs.S1.mixed_state[0].mass_frac_phase_comp
@@ -193,8 +194,8 @@ def set_operating_conditions(m, water_recovery=0.5, over_pressure=0.3, solver=No
     # ---specifications---
     # feed
     # state variables
-    m.fs.feed.properties[0].pressure.fix(101325)  # stage 1 feed pressure [Pa]
-    m.fs.feed.properties[0].temperature.fix(273.15 + 25)  # stage 1 feed temperature [K]
+    m.fs.feed.properties[0].pressure.fix(101325)  # feed pressure [Pa]
+    m.fs.feed.properties[0].temperature.fix(273.15 + 25)  # feed temperature [K]
     # properties (cannot be fixed for initialization routines, must calculate the state variables)
     m.fs.feed.properties.calculate_state(
         var_args={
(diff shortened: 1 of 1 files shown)


MarcusHolly:
Currently running into a few errors with this flowsheet:

1.) Deprecation warning for the "default argument" and warnings about negative capital costs (I also run into these issue when running the original RO_with_energy recovery file)

2.) Solver failing to return an optimal solution - see my commented questions for why this may be

andrewlee94:
@MarcusHolly Regarding the warning about the "default" argument, that implies you are running a later version of IDAES than WaterTap currently tests against (not that this is a problem). At this stage, that is just a warning of an upcoming change, but if you can address it it would reduce the amount of effort required later.

Basically, where ever you did something like `m.fs = FlowsheetBlock(default={"dynamic": False})` you can now do `m.fs = FlowsheetBlock(dynamic=False)`. This will be the way to do thing in the IDAES November release.

MarcusHolly:
@adam-a-a 

Not sure why there are a bunch of equality constraints setting things equal to zero. It seems to be mostly affecting the outlet streams (PXR disposal, RO2 permeate, RO2 retentate). I've tried changing/muting a lot of stuff but nothing seems to affect these hidden constraints (or has very minimal effect).

Also, I forget exactly how to decipher what gets displayed by tee=True, but I think I remember that iteration numbers with an "r" next to them are not ideal (e.g. 7r - 91r)

andrewlee94:
@MarcusHolly Numbers with `r` next to them indicate that IPOPT has gone into restoration mode. For solving square problems, this is always a bad sign (for optimization problems, a small number of restoration steps can be OK).

From just reading your comment above (and very quickly at that), my first guess is that you have some badly scaled constraints, such as all those constraints setting things to zero. Try scaling all of those constraints by a factor or ~1e5 as a starting point,

MarcusHolly:
> @MarcusHolly Numbers with `r` next to them indicate that IPOPT has gone into restoration mode. For solving square problems, this is always a bad sign (for optimization problems, a small number of restoration steps can be OK).
> 
> From just reading your comment above (and very quickly at that), my first guess is that you have some badly scaled constraints, such as all those constraints setting things to zero. Try scaling all of those constraints by a factor or ~1e5 as a starting point,

Alright, I'll look into this. Thank you Andrew!

adam-a-a:
@MarcusHolly - Just seeing this now- Andrew hit on the important notes already, but-- which constraints were you referring to?

MarcusHolly:
> @MarcusHolly - Just seeing this now- Andrew hit on the important notes already, but-- which constraints were you referring to?
All of the ones listed below. I haven't really touched this file in a few days though, so maybe I'll be able to resolve some of these once I revisit it.

CONSTR fs.M1.pressure_equality_constraints[0.0,2]: 6.680745948475866 =/= 0.0
CONSTR fs.product.properties[0.0].eq_flow_vol_phase: 999.0 =/= 0.0
CONSTR fs.product.properties[0.0].eq_dens_mass_phase: -0.021460000000000038 =/= 0.0
CONSTR fs.s07_expanded.flow_mass_phase_comp_equality[0.0,Liq,H2O]: -0.47214554999999997 =/= 0.0
CONSTR fs.s07_expanded.flow_mass_phase_comp_equality[0.0,Liq,NaCl]: 0.05880687424011863 =/= 0.0
CONSTR fs.s11_expanded.flow_mass_phase_comp_equality[0.0,Liq,H2O]: -0.4721455499999999 =/= 0.0
CONSTR fs.s11_expanded.flow_mass_phase_comp_equality[0.0,Liq,NaCl]: -3.4836968742401178 =/= 0.0
CONSTR fs.s12_expanded.flow_mass_phase_comp_equality[0.0,Liq,H2O]: -0.4825 =/= 0.0
CONSTR fs.s12_expanded.flow_mass_phase_comp_equality[0.0,Liq,NaCl]: -3.4839922560081824 =/= 0.0
CONSTR fs.s13_expanded.flow_mass_phase_comp_equality[0.0,Liq,H2O]: -0.48249999999999993 =/= 0.0
CONSTR fs.s13_expanded.flow_mass_phase_comp_equality[0.0,Liq,NaCl]: -0.01600774399181798 =/= 0.0
CONSTR fs.s13_expanded.pressure_equality[0.0]: 6.557150373195833 =/= 0.0

adam-a-a:
This PR was meant to bring in a 2-pass RO system (i.e., permeate from stage 1 feeds into stage 2). Subsequently, we planned to insert a boron removal process (we have a unit for that) between stages. Later, we could use this as a baseline case against novel approaches that aim to remove boron from water.

However, this example has been low priority for some time, resulting in this stranded draft PR. I will also note the challenge in this example is likely due to the lower concentrations of the permeate, requiring some adjustment in scaling to converge to a solution. That said, I didn't probe this myself for quite some time, so it could also be a variable-bound related issue.

PR #696, opened 2022-08-29
Improve ED flowsheet stability 

lbibl:
## Summary/Motivation:

The previous /flowsheet/electrodialysis/electridialysis_1stack.py showed instability/inconsistency in solver outcomes under different OS or trivial changes from other files.  This was diagnosed to be caused by an inequality constraint in the optimization of that file leading to local minima convergence.  This PR provides a solution to this issue and improves its execution stability. 

-
-

### Legal Acknowledgement


watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack.py
```diff
@@ -237,9 +237,9 @@ def optimize_system(m, solver=None):
     m.fs.EDstack.cell_pair_num.setub(1000)
 
     # Set a treatment goal
-    # Example here is to reach a final product water containing NaCl < 1 g/L (from a 10 g/L feed)
+    # Example here is to reach a final product water containing NaCl = 1 g/L (from a 10 g/L feed)
     m.fs.eq_product_quality = Constraint(
-        expr=m.fs.product.properties[0].conc_mass_phase_comp["Liq", "Na_+"] <= 0.393
+        expr=m.fs.product.properties[0].conc_mass_phase_comp["Liq", "Na_+"] == 0.393
```
bknueven: Is it the case that any minima of this model should satisfy this constraint with equality? Put differently, if this constraint is satisfied with strict inequality (i.e.., the product NaCl mass concentration is strictly less than 0.393), then there is an improving direction/solution?

lbibl: Hi @bknueven , I may not yet know enough to fully understand your question.  I think converting this constraint to equality reduced a degree of freedom so that may have significantly transformed the problem including its minima profile?  Previously, with the non-strict inequality <=, the solution appeared to be different under different scaling on this constraint.  The == constraint made the solution pretty consistent across a range of scaling factors.  I tried to set a strict inequality to test your second question, but it seemed not allowing me to use any strict inequality with a ValueError.   Does this help clear your question?

bknueven: I guess the crux of my question is if this change makes sense striclty from a modeling perspective (forget about solving it, at least for a moment).

lbibl: Yes, at least for now.  This is a temporary solution we discussed in the Chemistry team meeting. So that constraint, by the problem, regulates the water product quality - earlier it says we want a  water containing <=1 g/L NaCl and this one says == 1g/L.  Although <= 1g/L is closer to a reality quality control, == 1 g/L let us do a more reliable optimization.  With the unit model bing refined, we'll continue to make the optimization problem more robust.

bknueven (approved):

TimBartholomew (approved):
LGTM

TimBartholomew:
Honestly the best approach is to just fix the variable to the value, instead of creating the equality

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack.py
adam-a-a: @lbibl I can't recall, but did you also try _not_ having any constraint for this during optimization and instead just set an upper bound on the variable? 

For example:
`m.fs.product.properties[0].conc_mass_phase_comp["Liq", "Na_+"].setub(1)`

lbibl: `setub` can pass locally on my machine.  Will that not be equivalent to writing inequality constraint as `<=`?  That still left 2 dof and I am not sure if it addresses the high sensitivity/ local minima issue .  I recommend merging this version . But I can also push the `setub` version to see if it passes all platforms (I think it will) and see if that causes issue upon costing package (which can only be tested in the IX PR).

adam-a-a: I think `setub` would be the way to go and I don't think it's the same as writing the inequality constraint. Sure, you could even throw up a draft PR up where you checkout from this branch, set the upper bound instead of constraint, and make the PR just to see how tests go now.

PR #697, opened 2022-08-29
Electrical carbon intensity in WaterTAP costing

arao53:
## Fixes/Resolves:
Accounts for the equivalent embedded carbon emissions associated with consuming electricity from an electric grid.

- `electrical_carbon_intensity` global parameter in WaterTAP costing.
- `add_specific_electrical_carbon_intensity` method normalizes the annual carbon emissions associated with electricity by the annual water production.
- Tested using the `RO_with_energy_recovery` flowsheet.

### Legal Acknowledgement

### DIFF
--- a/watertap/costing/watertap_costing_package.py
+++ b/watertap/costing/watertap_costing_package.py
@@ -119,7 +119,7 @@ class WaterTAPCostingData(FlowsheetCostingBlockData):
 
         self.electrical_carbon_intensity = pyo.Param(
             mutable=True,
-            initialize=0.1,
+            initialize=0.475,
             doc="Grid carbon intensity [kgCO2_eq/kWh]",
             units=pyo.units.kg / pyo.units.kWh,
         )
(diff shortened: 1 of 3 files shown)


bknueven (commented):
Should we specify that this is the *electrical* carbon intensity (e.g., `electrical_carbon_intensity` and `specific_electrical_carbon_intensity`.)

We have, and our bringing in more, chemical addition models which I would imagine have their own associated carbon intensity.

watertap/costing/watertap_costing_package.py
```diff
@@ -117,6 +117,13 @@ def build_global_params(self):
             units=self.base_currency / pyo.units.kWh,
         )
 
+        self.electrical_carbon_intensity = pyo.Param(
+            mutable=True,
+            initialize=0.1,
```
bknueven: Is this value notional? It seems a bit optimistic, at least according to what I can find googling around.

arao53: Thanks for the catch! The commit 4c03333 updates this value based on the [2019 IEA Global Energy & CO2 Status Report 2019](https://www.iea.org/reports/global-energy-co2-status-report-2019): ~0.475 kgCO2eq/kWh.

adam-a-a (approved):

bknueven (approved):

Issue #695, opened 2022-08-26
Standardize `ion_set` indexing and refine `Set` structure in the DSPMDE property package


lbibl:
Currently, in DSPMDE properties possessed only by ions are indexed by `ion_set | solute_set`.  The chem team agreed to standardize this to let ion properties indexed by`ion_set` only.  This is on a clear to do list. 

Secondly, the chem team discussed the potential refinement of the overall `Set` structure in DSPMDE.  The current version has `solute_set` being nonionic species separated from `ion_set`.  A better solution is to have `cation_set` and `anion_set`  IN `ion_set`; `ion_set` and `nonion_set`(or `molecule/neutral_set` to consider) IN `solute_set`; `solute_set` and `"H2O"` (`solvent_set`) IN `component_list`.   This can be done by directly declaring everything as `Set( )` and grouping an input dictionary into them by a loop in DSPMDE.  However, there are some considerations about being aligned with and whether to use the IDAES convention, on which side there are also some possible changes in the future.  The decision and action on this are thus on hold and tracked by this issue. 

TO DO:

- [x] Standardize ion property index ( PR #699 )
- [ ] Decide and act on refining the `Set()` structure

adam-a-a:
@lbibl Thanks for putting this together. Quick comment- I would opt for `neutral_set` with regard to solutes with no charge.

Additionally, we might want to rename this property package something else that doesn't mention "DSPMDE" explicitly. Although this property model started out as a quick creation of mine so that I could get the nanofiltration DSPM-DE model going, the prop model has become a bit more general and is being used by a few other units. A trivial change, but something like `ion_property_model` would seem fine to me. That is, no need to overthink the name.

lbibl:
> @lbibl Thanks for putting this together. Quick comment- I would opt for `neutral_set` with regard to solutes with no charge.
> 
> Additionally, we might want to rename this property package something else that doesn't mention "DSPMDE" explicitly. Although this property model started out as a quick creation of mine so that I could get the nanofiltration DSPM-DE model going, the prop model has become a bit more general and is being used by a few other units. A trivial change, but something like `ion_property_model` would seem fine to me. That is, no need to overthink the name.

I also thought of changing the file name, but I also would like a sense of historic flavor to keep the dspmde name (just to memorize how it was born :) ) .  If renaming, my first idea of it is `solution_property` or even `liquid_solution_property`, for solution is the basic term describing a mixture with one component (solvent) solvating others (solutes).  

[...]

PR #699, opened 2022-08-30
DSPMDE ion property refinement 

lbibl:
## Summary/Motivation:
This adds new electrical properties/methods into the DSPMDE ion property package and standards ion_set indexing discussed in issue #695 

- Added new electrical properties and methods for ions.
- Added corresponding tests for new properties
- Adapted scaling methods to the changes. 
- Changed all properties possessed by charged ions to be only indexed  by `ion_set`

### Legal Acknowledgement


lbibl:
@andrewlee94 and @bknueven :  may I draw you attention to this PR?  At this point, it fails only for MacOS and two of the errors there are concerning: one throws an infeasible solution and the other off calculations.  The failed file passes on my local Mac machine @ python 3.10.   While further tailoring the scaling factor is one possible resolution, I wonder if this significantly different solver performance of the Experimental MacOS by Github, from all other OS and my local Mac, can suggest something not quite right on the testing OS by GitHub action.   I would like to hear your suggestions before making painful efforts (already tried a few rounds) to hunt for the problem, which is even more inconvenient without having the tester in hand....

andrewlee94:
@lbibl There are two things that can cause a difference in solver behaviour across different platforms:

1. Limitations in the linear solver available - there is the possibility that the MacOS workflows are using MUMPS as the linear solver instead of MA27. You should check to see which linear solver is being used (it should be in the header of the IPOPT output).
2. Differences in the underlying linear algebra packages which result in small differences in the linear solver outputs. In a well scaled problem, these errors should be insignificant, but in problems with poor scaling (and especially unscaled very small numbers), these can and will result in significant differences in solver behaviour.

bknueven:
> 1. Limitations in the linear solver available - there is the possibility that the MacOS workflows are using MUMPS as the linear solver instead of MA27. You should check to see which linear solver is being used (it should be in the header of the IPOPT output).

This is the reason. @lbibl I made a few adjustments to the tests -- mostly to skip those that failed on my Intel Mac locally using MUMPS Ipopt. This is a fine solution we've used several other places in the code base.

lbibl:
> > 1. Limitations in the linear solver available - there is the possibility that the MacOS workflows are using MUMPS as the linear solver instead of MA27. You should check to see which linear solver is being used (it should be in the header of the IPOPT output).
> 
> This is the reason. @lbibl I made a few adjustments to the tests -- mostly to skip those that failed on my Intel Mac locally using MUMPS Ipopt. This is a fine solution we've used several other places in the code base.

Thanks @bknueven and @andrewlee94 .  Now it looks all passing. For information, I saw my local arm64 MacOS with ipopt 3.14.6 python 3.10 also uses NUMPS as the linear solver but it did _not_ fail as the MacOS workflow by Github action (which runs ipopt 3.14.4 and python 3.8 I think, I did not find information about its linear solver) on this PR.  Neither had I major issues in the past with this local setup.  That might suggest a difference either in the chip or ipopt/ python version I guess.

adam-a-a:
@lbibl I noticed that you remove 0 charge from tests where people specified components and their charge. My question is - if I try to specify my components and provide all the charge data for each, including those with 0, is there any issue with that? I didn't look thoroughly enough yet, but I am assuming you set the default value for charge to 0. I am just wondering why you deleted the 0s from the charge data input in some tests.

lbibl:
> @lbibl I noticed that you remove 0 charge from tests where people specified components and their charge. My question is - if I try to specify my components and provide all the charge data for each, including those with 0, is there any issue with that? I didn't look thoroughly enough yet, but I am assuming you set the default value for charge to 0. I am just wondering why you deleted the 0s from the charge data input in some tests.

@adam-a-a :  The change was made to let `charge_comp` indexed only by `ion_set`, which is established by taking nonzero-charged components from the `config.solute_list`.  This disallows the declaration of charge=0, as they are not in `ion_set`.    The same goes for all other properties possessed only by ions (nonzero charges).

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -257,18 +349,10 @@ def build(self):
...
-            mutable=True,
-            default=5.19e-8,  # default as Na+
-            initialize=self.config.electrical_mobility_data,
-            units=pyunits.meter**2 * pyunits.volt**-1 * pyunits.second**-1,
-            doc="Ion electrical mobility",
-        )
+
         # Ion charge
         self.charge_comp = Param(
-            self.ion_set | self.solute_set,
```
srikanthallu, line 271: @lbibl Maybe I am missing something, but is it sufficient to make charge_comp indexed on ion_set only when using to evaluate electrical properties. We need to check that user does not input ion sets with '0' charge. Can you please point me where you are raising errors when users does it by mistake ?

lbibl: What "mistake" do you mean?

adam-a-a: @lbibl I think @srikanthallu is referring to the idea of raising an exception in the case that a user tries entering 0 for the charge of an ion. 

If you were to have such an exception being raised, you'd also want to have a test to ensure that the exception actually gets raised when entering charge = 0 for an ion.

lbibl:
Note PR #711 expected to resolve the current Win64 failed tests; actual test pending.

watertap/property_models/ion_DSPMDE_prop_pack.py
lbibl: Test added.

lbibl:
@adam-a-a and all: the flowsheet improvement has resolved the earlier fails. Some new lines and tests were added to address given comments and cover more codes.  This is ready for reviewing.

srikanthallu (commented):
Changes LGTM, but when I run the test I see a test failure, `test_assert_electroneutrality_get_property` Does this happen on your end as well @lbibl ?

bknueven:
This maybe isn't for this PR, but when looking at it I noticed that `test_ion_DSPMDE_prop_pack` has two identically named functions: `test_property_ions` and `test_assert_electroneutrality_get_property`.

Because they're defined twice, Python overwrites the definition of the first with the definition of the second, so the first test is never executed.

I would strongly suggest reorganizing this test file into a test *class* for each model to avoid such namespace clashes between tests.

lbibl:
> Changes LGTM, but when I run the test I see a test failure, `test_assert_electroneutrality_get_property` Does this happen on your end as well @lbibl ?

@srikanthallu: there were no testing failures by GitHub Action and my local.  That might suggest some significant difference of your local environment from GitHub action .

lbibl:
> This maybe isn't for this PR, but when looking at it I noticed that `test_ion_DSPMDE_prop_pack` has two identically named functions: `test_property_ions` and `test_assert_electroneutrality_get_property`.
> 
> Because they're defined twice, Python overwrites the definition of the first with the definition of the second, so the first test is never executed.
> 
> I would strongly suggest reorganizing this test file into a test _class_ for each model to avoid such namespace clashes between tests.

Thanks for pointing out this @bknueven . @adam-a-a , just to draw your attention to this comment- we may consider a major revision on the test file  (as part of dspmde to-dos) according to Ben's suggestion.

bknueven:
> Thanks for pointing out this @bknueven . @adam-a-a , just to draw your attention to this comment- we may consider a major revision (as another to-do candidate for this growing property package) according to Ben's suggestion.

Mind -- nothing needs to change in the property package itself, just the tests.

lbibl:
> > Thanks for pointing out this @bknueven . @adam-a-a , just to draw your attention to this comment- we may consider a major revision (as another to-do candidate for this growing property package) according to Ben's suggestion.
> 
> Mind -- nothing needs to change in the property package itself, just the tests.

Yes! Thanks.  I think I got it correctly- in the test file, separating things into classes instead of too many functions in one class.

srikanthallu:
> 

If you define function names @bknueven mentioned, this failure will become evident:

```
============== short test summary info ========================
FAILED test_ion_DSPMDE_prop_pack.py::test_property_ions1 - AssertionError: Electroneutrality condition violated in fs.stream[0]. Ion concentrations should be adjusted to bring the result of 3.68E-05 closer towards 0.
```

lbibl:
> If you define function names @bknueven mentioned, this failure will become evident:
> 
> ```
> ============== short test summary info ========================
> FAILED test_ion_DSPMDE_prop_pack.py::test_property_ions1 - AssertionError: Electroneutrality condition violated in fs.stream[0]. Ion concentrations should be adjusted to bring the result of 3.68E-05 closer towards 0.
> ```

@srikanthallu : I don't think this error message is relevant to @bknueven 's comment.  This suggests the solver in your environment is computing things differently from GitHub Action, which does not show this error.

lbibl:
@srikanthallu sent to me another error message that is worth investigating.  I will try a simple solution to re-name one of the functions to make sure things were solved correctly.  Will give an update.

lbibl:
> @srikanthallu sent to me another error message that is worth investigating. I will try a simple solution to re-name one of the functions to make sure things were solved correctly. Will give an update.

Update:  with what @bknueven and @srikanthallu pointed out, I located an error that has been in the file before I made all changes. It eluded the past tests because of the duplicated function names but somehow was caught by Srikanth's local env (a helpful discrepancy there whatever it is). 

Briefly,  the following line, 

https://github.com/watertap-org/watertap/blob/dad677189d0b0a162ad91fbde78fd7d61c9234c2/watertap/property_models/tests/test_ion_DSPMDE_prop_pack.py#L161  

is failing (revealed by renaming that function; the wt-org main is quoted here) due to a too large charge balance calculated by the solver.  I extracted that part of codes and found the `assert_electroneutrality` gives correct outcome only if I, before calling it,  manually  touch `conc_mol_phase_comp` and run `stream.initialize()` , which were not in the testing file.  Adding these in the testing file led to slightly different calculation outcomes for all properties in that test session.  My insight is this points to some issue with the `assert_electroneutrality` function- it seems solving things a little differently and maybe lack a proper initialization before solving (not sure).  _With all considered, I propose to leave this issue to a separate PR in the future since it is not related to the changes made in this PR._  

@adam-a-a , again to draw your attention to this- I read through `assert_electroneutrality` function in dspmde but could not quickly come up with a fix.  May be you can have more insight quickly.  

Thanks to all for your input.

lbibl:
Reviewers: I realize that the ED1d refinement will accommodate changes in this PR so it's better to have this one merged before opening a PR for the ED1d refinement, which is ready. Appreciate your effort.

srikanthallu (approved):
Please list the open issues that we realized during the review.

lbibl:
> Please list the open issues that we realized during the review.

The issue is being fixed by PR #738 mentioned above.

watertap/property_models/ion_DSPMDE_prop_pack.py
```diff
@@ -1186,6 +1366,122 @@ def rule_pressure_osm_phase(b):
...
+                        Provide this configuration or use another "equiv_conductivity_calculation"
+                        configuration value to contruct the demanded variable(s))""".format(
+                            self.name
+                        )
+                    )
+                else:
+                    if len(self.params.ion_set) > 2:
+                        _log.warning(
+                            """ 
+                            Cautions should be taken to use a constant solution equivalent conductivity for a multi-electrolyte system.
```
adam-a-a:
```suggestion
                            Caution should be taken to use a constant solution equivalent conductivity for a multi-electrolyte system.
```

adam-a-a (approved):

PR #701, opened 2022-08-31
Updating LSRRO multisweep baselines

### LINKED ISSUE #685: Change some LSRRO baselines
PRs #570, #684, #686, and #679 show some LSRRO tests to be unstable.

I think this is because, as @adam-a-a mentioned [here](https://github.com/watertap-org/watertap/pull/570#issuecomment-1184921922), we can expect deviations when we have more stages than is strictly cost-optimal. From an optimization perspective, I agree -- a system with too many stages would likely have multiple local minima that Ipopt might converge to with slight perturbations in the problem.

The fix is to make sure (as best we are able) that our baseline tests are free of spurious local minima by not using operating conditions which are optimal for fewer stages.

bknueven:
## Summary/Motivation:
Some LSRRO tests are unstable because there a more stages than necessary. This removes coverage fro those cases because the flowsheets are known to have lots of local minima

- Ignore cases when the baseline has an LSRRO stage with area close to 1 m^2
- Remove 3, 4, and 5 stage tests for LSRRO multi-sweep because all baselines have a stage with area close to 1 m^2 or are infeasible.

### Legal Acknowledgement


watertap/examples/flowsheets/lsrro/tests/test_lssro_multi_sweep.py
```diff
@@ -11,38 +11,43 @@
...
     test = pd.read_csv(csv_test_file_name).astype(float).T.to_dict()
 
     assert len(baseline) == len(test)
 
-    for k in test:
-        assert pytest.approx(baseline[k], nan_ok=True, rel=1e-02, abs=1e-07) == test[k]
+    for k, base in baseline.items():
+        # Don't test those cases which have too many stages
+        for s in range(1, number_of_stages + 1):
+            if math.isclose(base[f"Membrane Area-Stage {s}"], 1.0, abs_tol=1e-4):
```
adam-a-a, line 47: Cool logic - just wondering, since we're now limiting the test cases to not span the full range of cases, is this check necessary?

bknueven: Yes -- it's now skipping 2-stage at 5 g/L and 30% recovery, which is still a not-very-stable point.

adam-a-a (approved):
LGTM - I'm assuming the tests are going to pass.

tarnold17 (approved):

PR #704, opened 2022-09-02
Add has_aop configuration for UV model

luohezhiming:
## Summary/Motivation:
Add `has_aop` configuration and corresponding equations

- add `has_aop` configuration
- add report
- add corresponding tests
- add corresponding doc

### Legal Acknowledgement


docs/technical_reference/unit_models/uv_aop.rst
```diff
@@ -98,16 +98,30 @@ When setting the `uv_dose_type` configuration option to `UVDoseType.calculated`,
    "Relative lamp output", ":math:`\frac{S}{S_0}`", "relative_lamp_output", None, None
    "Number of banks", ":math:`N_{bank}`", "num_of_banks", None, None
 
+When setting the ``has_aop`` configuration option to ``True``, there are 2 additional variables that must be fixed.
```
hunterbarber: This is 2 additional variables per number of species that are photolyzed. So 2j additional variables but it only works for 1 species currently, correct? I just want to make sure I am interpreting correctly.

docs/technical_reference/unit_models/uv_aop.rst
```diff
@@ -98,16 +98,30 @@ When setting the `uv_dose_type` configuration option to `UVDoseType.calculated`,
...
+.. csv-table::
+   :header: "Description", "Symbol", "Variable Name", "Index", "Units"
+
+   "Pseudo-first order rate constant for direct photolysis of component", ":math:`k_d`", "photolysis_rate_constant", "[p, j]", ":math:`\text{s}^{-1}`"
+   "Pseudo-first order rate constant for indirect photolysis of component", ":math:`k_i`", "reaction_rate_constant", "[p, j]", ":math:`\text{s}^{-1}`"
+   "Second-order reaction rate constant", ":math:`k_{OH}`", "second_order_reaction_rate_constant", "[p, j]", ":math:`\text{M}^{-1} \text{s}^{-1}`"
+   "Steady-state concentration of hydrogen peroxide", ":math:`c_{\text{H}_2\text{O}_2}`", "hydrogen_peroxide_conc", None, ":math:`\text{M}`"
+
+**Users must provide values for and 'fix' certain variables to solve the model with DOF=0. Thus, users should fix**
+    * either 'photolysis_rate_constant' or 'reaction_rate_constant',
```
hunterbarber, line 114: Firstly, this should now be removed from line 80 I believe since it is now only constructed when `has_aop=True`.

Secondly, would it make sense that when `has_aop=False`, `rate_constant` should be fixed. When `has_aop=True`, `rate_constant` should be unfixed and both `photolysis_rate_constant `or `reaction_rate_constant `should be fixed. How are these parameters obtained, and does this make sense?

hunterbarber (commented):
A couple clarifications for when switching between `has_aop `options

docs/technical_reference/unit_models/uv_aop.rst
luohezhiming: For the first point, you are correct line 80 need to be deleted.
For the second point, `rate_constant` does not need to be fixed or unfixed. The `rate_constant` can be derived from `inactivation_rate` and `uv_intensity` or directly fixed. So it is the same for `has_aop=True` as `has_aop=False`, either `rate_constant` or `inactivation_rate` need to be provided.

docs/technical_reference/unit_models/uv_aop.rst
luohezhiming: For multiple species, I think (j+1) additional variables need to be provided if describe precisely. `photolysis_rate_constant` or `reaction_rate_constant` has j, and only need 1 `hydrogen_peroxide_conc` or `second_order_rate_constant` (other can be derived). However, for here, I think it is easier for understanding to use 2 instead of (j+1), and the variable simply represents the notation listed below.

docs/technical_reference/unit_models/uv_aop.rst
hunterbarber: I'm on the same page when `has_aop=False` that "the `rate_constant` can be derived from `inactivation_rate` and `uv_intensity` or directly fixed." It also seems like the `inactivation_rate` is the governing variable in the mass transfer equation so this would be the direct specification.

I think that the `rate_constant` in this `eq_rate_constant` equation is fundamentally the `photolysis_rate_constant` though when `has_aop=False` and that is where there is some confusion when `has_aop=True`. Maybe having the `reaction_rate_constant.fix(0)` when `has_aop=False` is more clear, but I think you may of mentioned you used to have it this way? If this is true, then `rate_constant` which is the `overall_rate_constant` should never be fixed as only the `photolysis_rate_constant` or `reaction_rate_constant` are experimentally derived. 

Hopefully I am interpreting this correctly.

hunterbarber: So given the equations,
```
b.rate_constant[p, j] == b.uv_intensity * b.inactivation_rate[p, j]
b.rate_constant[p, j] == b.photolysis_rate_constant[p, j] + b.reaction_rate_constant[p, j]
```

Basically if for a theoretical system that,
```
when has_aop=False
     b.rate_constant[p, j]=10
```
I would suspect that `b.photolysis_rate_constant[p, j]=10` and `b.reaction_rate_constant[p, j]=0`

Then for the same theoretical system and introducing aop, if you keep the `b.rate_constant[p, j]=10` specification fixed, there is no mathematical way it is feasible and always should change to the independent term specifications of,
```
when has_aop=True
    b.photolysis_rate_constant[p, j]=10
    b.reaction_rate_constant[p, j]=5
```

luohezhiming: I finally decide to take your suggestion. Right now, both `photolysis_rate_constant` and `reaction_rate_constant` are default variables and `reaction_rate_constant` is fixed to 0 when `has_aop=False`.

watertap/unit_models/uv_aop.py
```diff
@@ -234,43 +236,26 @@ def build(self):
 
         units_meta = self.config.property_package.get_metadata().get_derived_units
 
-        # TODO: update IDAES such that solvent and solute lists are automatically created on the parameter block
```
hunterbarber, line 207: When you say this, you are saying in the NDMA_prop_pack correct? This is how ion_DSPMDE_prop_pack does it which may be a useful reference.

hunterbarber (approved):
only looking at the new `target_species `commit which looks good at first glance, only having one small comment for future work

watertap/unit_models/uv_aop.py
luohezhiming: This comment is deprecated and I copied from other files, I've deleted.

watertap/unit_models/uv_aop.py
```diff
@@ -527,23 +551,65 @@ def eq_uv_reactor_volume(b):
...
+        else:
+
+            @self.Constraint(
+                self.config.property_package.phase_list,
+                self.target_species,
+                doc="Fix pseudo-first order reaction rate constant to zero",
+            )
+            def eq_no_reaction_rate_constant(b, p, j):
+                b.reaction_rate_constant[p, j].fix(0)
+                return Constraint.Skip
```
adam-a-a: In this case, I wonder if we even need to the Constraint here. Maybe after `else`, we could just fix `reaction_rate_constant` to 0.
```suggestion
            self.reaction_rate_constant[p, j].fix(0)

```

bknueven: I agree -- or use `self.reaction_rate_constant[p, j] == 0`.

An empty constraint object rarely makes sense.

adam-a-a (approved):

PR #705, opened 2022-09-02
Objective in RO with Energy Recovery Flowsheet

arao53:
## Fixes/Resolves:
Through recent changes, the objective function was removed from the `set_operating_conditions` function.

## Summary/Motivation:
This PR adds the objective back to where it was originally and includes tests to verify that the `number of objectives == 1`.

- Adds objective function for RO with ERD flowsheet.
- Tests that the model has 1 objective function using `idaes.core.util.model_statistics.number_of_objectives(m)`.

### Legal Acknowledgement


bknueven:
Why put this in `set_operating_conditions`? If `optimize_set_up` is a bad place for it (where it is currently) then it should just be in `build`, IMHO

arao53:
@bknueven the objective and associated tests have been moved to `build` in b0fe48b

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
```diff
@@ -456,7 +456,6 @@ def initialize_pump_as_turbine(m, optarg):
 
 def optimize_set_up(m):
     # objective
```
bknueven:
```suggestion
```

bknueven (approved):
I'm still uncertain about the need for this change, but it doesn't seem like it would do any harm.

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
```diff
@@ -160,6 +160,7 @@ def build(erd_type=ERDtype.pressure_exchanger):
     m.fs.costing.add_specific_electrical_carbon_intensity(
         m.fs.product.properties[0].flow_vol
     )
+    m.fs.objective = Objective(expr=m.fs.costing.LCOW)
```
adam-a-a: Why put this here instead of where it originally was under `optimize_setup`? If it is only for the sake of checking the objective exists in tests, that could happen after the optimize_setup function is called in the test.

adam-a-a (changes requested):
I'm thinking the objective function should stay under the `optimize_setup `function, but the tests can stay (may need to move them to some place after optimize_setup is called though).

bknueven (approved):

agarciadiego (approved):
LGTM

tarnold17 (approved):

adam-a-a (approved):

PR #707, opened 2022-09-05
Update microbial battery groundwater treatment spreadsheet (AMO 1679)

tarnold17:
## Fixes/Resolves:
Updates to the groundwater treatment flowsheet. Some minor costing updates are included, and the display functions are updated to print out all of the information that is needed for reporting.

-
-

### Legal Acknowledgement


watertap/unit_models/zero_order/tests/test_microbial_battery_zo.py
```diff
@@ -266,6 +277,6 @@ def test_costing():
     assert degrees_of_freedom(m.fs.unit) == 0
     initialization_tester(m)
 
-    assert pytest.approx(4.37e-3, rel=1e-3) == value(m.fs.unit.costing.capital_cost)
+    assert pytest.approx(4320.05, rel=1e-3) == value(m.fs.unit.costing.capital_cost)
 
     assert m.fs.unit.electricity[0] in m.fs.costing._registered_flows["electricity"]
```
adam-a-a, line 271: Are there other registered flows that you should check for now?

watertap/core/zero_order_costing.py
```diff
@@ -2969,6 +2969,14 @@ def cost_microbial_battery(blk):
         blk.config.flowsheet_costing_block.cost_flow(
             blk.unit_model.electricity[t0], "electricity"
         )
+        blk.config.flowsheet_costing_block.cost_flow(
+            blk.unit_model.properties_in[t0].flow_mass_comp["filtration_media"],
+            "filtration_media",
+        )
+        blk.config.flowsheet_costing_block.cost_flow(
+            blk.unit_model.properties_byproduct[t0].flow_mass_comp["filtration_media"],
+            "filtration_media_disposal",
```
adam-a-a, line 2978: Should this be at the flowsheet level instead?

watertap/examples/flowsheets/case_studies/groundwater_treatment/groundwater_treatment.py
```diff
@@ -199,124 +348,111 @@ def add_costing(m):
...
     )
 
-    m.fs.costing.cost_spent_media_disposal = Expression(
-        expr=pyunits.convert(
-            m.fs.costing.spent_media_recovery_mass
-            * m.fs.costing.filtration_media_disposal_cost,
-            to_units=m.fs.costing.base_currency / m.fs.costing.base_period,
-        ),
-        doc="Disposal cost of spent filtration media per year (accounting for utilization factor)",
+    m.fs.costing.add_LCOW(m.fs.filtered_water.properties[0].flow_vol)
```
adam-a-a: Although the LCOW with respect to the product water flow rate will be very similar in this case, our standard go-to levelized cost metric should be the levelized cost with respect to influent flowrate.

watertap/examples/flowsheets/case_studies/groundwater_treatment/groundwater_treatment.py
```diff
@@ -199,124 +348,111 @@ def add_costing(m):
...
-        * pyunits.convert(
-            m.fs.feed.flow_mass_comp[0, "filtration_media"],
-            to_units=pyunits.kg / m.fs.costing.base_period,
-        ),
-        doc="Mass of fresh filtration media used per year (accounting for utilization factor)",
+    print("\n------------- Capital costs -------------")
+    DCC_normalized = value(
+        pyunits.convert(
+            (m.fs.pump.costing.capital_cost + m.fs.micbatt.costing.capital_cost)
+            / m.fs.costing.TIC
```
adam-a-a: Is this needed here? It's not clear whether we intended to apply a TIC factor, or if we just wanted a default value of TIC = 2 and would only apply in the case that the provided cost relationship called for its inclusion.

adam-a-a (commented):
Looks good- just a couple of minor comments/questions.

watertap/core/zero_order_costing.py
tarnold17: Are you referring to just the filtration media disposal cost? I don't think it hurts anything to have it here, but I'm open to considering an argument to the contrary. You mentioned to me the other day that you thought it made sense to include spent media disposal costs at the flowsheet level, since for this process it is conceivable that we might want to model something different downstream for the spent media other than just simply disposing of it. But I figure in that case, we could just set filtration_media_disposal_cost to $0/kg, and then add whatever downstream model units we wanted.

watertap/unit_models/zero_order/tests/test_microbial_battery_zo.py
tarnold17: Yes, probably. Good catch. I will look at Tim's metab example to get an idea of what other tests should be added.

watertap/examples/flowsheets/case_studies/groundwater_treatment/groundwater_treatment.py
tarnold17: Are you sure that influent flowrate is supposed to be our default basis for LCOW? I based a lot of these changes on what I saw in Tim's metab example, and in that flowsheet, product water is used as the basis for the LCOW calculation.

watertap/examples/flowsheets/case_studies/groundwater_treatment/groundwater_treatment.py
tarnold17: I copied this from `metab.py` - @TimBartholomew can you weigh in?

watertap/unit_models/zero_order/tests/test_microbial_battery_zo.py
tarnold17: I added a few additional tests on the latest commit. Let me know if you think I have missed anything that should be there.

watertap/examples/flowsheets/case_studies/groundwater_treatment/groundwater_treatment.py
adam-a-a: We've been using levelized cost of treatment (LCOT) to denote levelized cost per influent flow rate. I am sure that if we want a metric that is broadly applicable across all projects, levelized cost per influent flowrate will be the one of interest, at least for wastewater treatment applications.

Additionally, what we name the metric (i.e., LCOT or LCOW_feed) is fairly subjective. The important point here is that levelized cost per feed flowrate should always be relevant, and in some cases, levelized cost per product water flowrate may also be relevant. 

For this application, both metrics happen to be relevant.

watertap/examples/flowsheets/case_studies/groundwater_treatment/groundwater_treatment.py
adam-a-a: Nevermind- this is representative of the practical investment factor for the whole system. In the watertap costing package this is referred to as `factor_total_investment` and in this costing packing (zero order package) we use `TIC`. 

So we're good to go on this.

watertap/core/zero_order_costing.py
adam-a-a: > But I figure in that case, we could just set filtration_media_disposal_cost to $0/kg, and then add whatever downstream model units we wanted.

That works for me.

watertap/unit_models/zero_order/tests/test_microbial_battery_zo.py
adam-a-a: Looks good!

watertap/examples/flowsheets/case_studies/groundwater_treatment/groundwater_treatment.py
adam-a-a: Also note- if we are looking to the METAB flowsheet as the example- both LCOW and LCOT are expressions on the flowsheet (not to mention other levelized cost metrics relevant to that particular use case). 

And for the sake of clarity and being explicit while reporting, we can always say something like LCOW with respect to product water flowrate, LCOW with respect to influent flowrate. Alternatively we can just have LCOW and LCOT and be explicit about LCOT representing division by influent flow rate.

adam-a-a (approved):
LGTM - I think the only thing left to do is add an expression for LCOT (representing levelized cost with respect to influent flowrate) for inclusion in the report.

watertap/examples/flowsheets/case_studies/groundwater_treatment/groundwater_treatment.py
tarnold17: On my latest commit, I changed the LCOW metric to use feed flowrate as the basis. As you mentioned, it shouldn't really have any effect, because for this flowsheet, feed flowrate and outlet flowrate are essentially identical. But now it is following the preferred convention.

adam-a-a:
@tarnold17 I just realized that there is one more change that should be made here. The groundwater_treatment folder and its files should be moved to `watertap/examples/flowsheets/case_studies/wastewater_resource_recovery`

tarnold17:
> @tarnold17 I just realized that there is one more change that should be made here. The groundwater_treatment folder and its files should be moved to `watertap/examples/flowsheets/case_studies/wastewater_resource_recovery`

@adam-a-a This project isn't actually dealing with wastewater resource recovery, which is why I made a separate directory. Do you think it should live under `watertap/examples/flowsheets/case_studies/wastewater_resource_recovery` anyways?

adam-a-a:
> > @tarnold17 I just realized that there is one more change that should be made here. The groundwater_treatment folder and its files should be moved to `watertap/examples/flowsheets/case_studies/wastewater_resource_recovery`
> 
> @adam-a-a This project isn't actually dealing with wastewater resource recovery, which is why I made a separate directory. Do you think it should live under `watertap/examples/flowsheets/case_studies/wastewater_resource_recovery` anyways?

Yes, since it is part of that whole effort.

tarnold17:
> > > @tarnold17 I just realized that there is one more change that should be made here. The groundwater_treatment folder and its files should be moved to `watertap/examples/flowsheets/case_studies/wastewater_resource_recovery`
> > 
> > 
> > @adam-a-a This project isn't actually dealing with wastewater resource recovery, which is why I made a separate directory. Do you think it should live under `watertap/examples/flowsheets/case_studies/wastewater_resource_recovery` anyways?
> 
> Yes, since it is part of that whole effort.

Addressed in latest commit.

luohezhiming (approved):

PR #709, opened 2022-09-07
AMO 1595 costing updates

tarnold17:
## Fixes/Resolves:

Updates to the photothermal membrane and CANDO+P flowsheet. Some minor costing updates are included, and the display functions are updated to print out all of the information that is needed for reporting.

-
-

### Legal Acknowledgement


watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1595_photothermal_membrane_candoP/amo_1595.py
```diff
@@ -281,16 +295,30 @@ def add_costing(m):
...
+        doc="Levelized Cost of Treatment with respect to influent flowrate",
+    )
+
+    m.fs.costing.LCOT_with_revenue = Expression(
         expr=(
             m.fs.costing.total_capital_cost * m.fs.costing.capital_recovery_factor
             + m.fs.costing.total_operating_cost
             - m.fs.costing.value_bcp_recovery
             - m.fs.costing.value_N2O_byproduct
-            - m.fs.costing.value_water_byproduct
```
adam-a-a: I think that in this particular project, water recovery is more of a bonus, while the focus is placed on resource recovery. For that reason, I would lean towards keeping the value of water. Worst case scenario, the default sale price for water could be a very small number representative of 0, thus assuming no value for water but allowing the user to do a sensitivity analysis on that sale price.

tarnold17: Addressed in latest commit.

adam-a-a (approved):
Nice job!

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1595_photothermal_membrane_candoP/amo_1595_case_study.yaml
```diff
@@ -8,7 +8,7 @@ defined_flows:
     value: 0.5
     units: USD_2020/kg
   water:
-    value: 1.0
+    value: 1.0e-8
```
luohezhiming: Why changes this value? It seems very little cost of water

tarnold17: I discussed this with @adam-a-a - I want the default value to be 0, but if you put the value as exactly 0 or 0.0, there is a bug in pyomo which causes an error. So instead I set it to something infinitesimally small.

andrewlee94: @tarnold17 Is the error in Pyomo due to unit consistency checks failing? If so, could you add a comment to https://github.com/Pyomo/pyomo/issues/1627 in the hopes of getting Pyomo to address this.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1595_photothermal_membrane_candoP/amo_1595.py
```diff
@@ -264,10 +271,25 @@ def add_costing(m):
...
         ),
-        doc="Dollar value of byproduct nitrous oxide generated per year",
+        doc="Value of byproduct nitrous oxide generated per year",
     )
 
-    m.fs.costing.LCOW_bcp = Expression(
+    m.fs.costing.LCOW = Expression(
+        expr=(
+            m.fs.costing.total_capital_cost * m.fs.costing.capital_recovery_factor
+            + m.fs.costing.total_operating_cost
```
bknueven, line 280: Perhaps the costing package should provide an expression for this?

bknueven (approved):
Code looks reasonable; my only comment is for the costing package

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1595_photothermal_membrane_candoP/amo_1595_case_study.yaml
bknueven: It probably doesn't matter -- though if this is to be "zero" I would choose a smaller value (`1e-10` or lower)

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1595_photothermal_membrane_candoP/amo_1595.py
tarnold17: I'd be happy to add that to `zero_order_costing.py` as part of this PR - do you have a name you want to suggest?

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1595_photothermal_membrane_candoP/amo_1595_case_study.yaml
luohezhiming: If the water cost is 0, then why not directly delete this registered flow? Does it use in any place other than operating cost?

tarnold17: @andrewlee94 - done

@bknueven - done on latest commit

@luohezhiming - The intent is to keep the parameter available in case we want to run a sensitivity analysis in which the sale price of water is > 0. We just want the default value to be zero.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1595_photothermal_membrane_candoP/amo_1595.py
bknueven: What is this -- `total_annualized_cost`? Should probably be in both packages or in IDAES. Let's open an issue instead.

tarnold17: Created issue #710.

Issue #595, opened 2022-06-23
WaterTAP Documentation for Zero-order Models


adam-a-a:
Checklist for refining zero-order model documentation. Technically, we have models via WT3 and AMO, though our initial focus is on finalizing WT3 derived models.
**High Priority:**
- [x] Edit automated documentation script and revise costing description so that user can be directed to description of unit model cost methods (#1064 )
- [ ] Documentation for the workflow for zero order modeling flowsheets - how to, after September 2022

**Nice to have**
- [ ] Disclaimer single warning when loading data from YAML for zero order model data @bknueven 
- [ ] Add process subtypes to docs. Some models have them, while others don't. Further, certain subtypes may have additional variables and constraints that should also be incorporated (e.g., ion_exchange) - after September @adam-a-a 
- [ ] Refine constraint docstrings that are either missing in unit models or could use improvement - @kurbansitterley 
- [ ] Noticed the electricity methods page could use a little more explanation- this is separate from the automated unit model documentation but worth noting nevertheless, as most unit model docs provide a link to the electricity helper methods doc. Reevaluate all original documentation on zero order modeling framework.
- [ ] Should document valid ranges for unit costing methods, prioritizing units with WaterTAP-adapted cost formulations. For example, add column(s) for valid ranges relevant to costing methods for document automation script. Could also apply to technical performance and valid param/var ranges, but this would be going above and beyond allocated time/resources. - After September
- [ ] add images for each unit, and/or some introductory sentence on what each unit can typically be applied for.
- [ ] Remove no index directive for class documentation and change to what we are currently doing with documentation.
- [ ] DELETE the "Smp" unit
- [ ] as @andrewlee94 mentioned, there are some tools that can display Pyomo expressions in LaTeX (https://github.com/IDAES/idaes-pse/blob/main/idaes/core/util/expr_doc.py); could include constraints in symbolic form in documentation; J Eslick designed those tools so follow up with him.

**DONE**
- [x] Disclaimer - any data for zero order models are for specific conditions, here are default values that could be useful but users need to get values that are for their application. Short description of what zero order model is and the warning. @adam-a-a @wigging PR #921 #977 
- [x] Add a check or test to ensure that all zero order models have a zero order doc; this should account for any zero order models that get added without being entered in the excel file from which models are read in for automating document creation (@lbianchi-lbl will be looking into this) after September @adam-a-a @MarcusHolly 
PR #799  and PR #877 
- [x] Update electricity and costing methods for relevant units.
[...]

PR #722, opened 2022-09-12
Feed (ZO)

agarciadiego:
## Fixes/Resolves:

Feed ZO documentation in #595 

## Summary/Motivation:

Autogenerated doc for FeedZO

-
-

### Legal Acknowledgement


docs/technical_reference/unit_models/zero_order_unit_models/automate_rst_file_wt3.py
```diff
@@ -165,6 +165,81 @@ def grab_unit_components(unit_class):
     )
 
 
+def grab_unit_components_feed(unit_class):
+
+    m = ConcreteModel()
+    m.fs = FlowsheetBlock(default={"dynamic": False})
```
andrewlee94: We need to start transitioning away from using `default` in these cases as that is being deprecated in IDAES as we speak - i.e. after November this will cause exceptions to be raised, and we don;t want to be adding to the number of things we need to fix.

For reference, the new format is: `m.fs = FlowsheetBlock(dynamic=False)`

adam-a-a: @andrewlee94 if we reinstall/update our watertap environments, will this work? That is, is IDAES 2.0 set as a requirement yet in setup.py?

adam-a-a: Just checked- it looks like it is.

andrewlee94: As long as you are using some variant of IDAES v2.0 (e.g. 2.0.0.a2 which is the version used for the last WaterTap release). The simplest check is that if you see a deprecation warning when you do this, then the new version will also work.

docs/technical_reference/unit_models/zero_order_unit_models/feed_zo.rst
```diff
@@ -4,6 +4,7 @@ Feed  (ZO)
 Model Type
 ----------
 This unit model is formulated as a **None** model form.
```
adam-a-a: For the case of FeedZO, we could provide an alternative section title instead of Model Type and also get rid of that first sentence about the model form.

docs/technical_reference/unit_models/zero_order_unit_models/feed_zo.rst
```diff
@@ -13,6 +14,24 @@ Costing Method
...
+--------------------
+
+.. csv-table::
+   :header: "Description", "Variable Name", "Units"
+
+   "Volumetric flowrate in feed", "flow_vol", ":math:`m^3/s`"
+   "Component mass concentrations", "conc_mass_comp", ":math:`kg/m^3`"
+
+Additional Constraints
+----------------------
```
adam-a-a: For the other unit models, additional variables and constraints sections were written if they were not already covered by the "model type" helper functions (e.g., SIDO, PT, etc.), which have their own documentation. However in this case, FeedZO doesn't use any of those helper functions and inherits from IDAES' `Feed` block class. Essentially, FeedZO adds volumetric flowrate and mass concentration as _variables_, in connection with the zero-order property model's state variable that is mass flow rate.

Something that makes this useful is that the user could enter their feed flow rate (volumetric) and concentrations without having to figure out mass flowrates of each component in the feed, which could be less direct. We probably want to elaborate on that in the first section.

adam-a-a:
This is what the doc looks like now:
image


I think the Model Type section just needs a custom sentence or two describing what Feed ZO is based on (I think it inherits from IDAES Feed block but should double-check). Electricity consumption section can be removed too.

adam-a-a (approved):
LGTM

bknueven (approved):

PR #738, opened 2022-09-16
fix the two tests that were skipped

adam-a-a:
## Fixes/Resolves:
Resolves issue identified in PR #699  
## Summary/Motivation:
Fix tests in ion prop package

-
-

### Legal Acknowledgement


adam-a-a:
@lbibl Upon further review of the assert_electroneutrality method, I think that the stateblock was indeed solving. Happy to discuss with you to ensure there is no error within the method.

lbibl:
@adam-a-a : after our discussion accurately located the issue, I think I came up an easy solution to fix it, which is a "double solve" of the model in the `assert_neutrality` function and the first solve is equivalent to an `initialize`.  By this mean the model was solved at higher precision and can pass the common 1e-8 tol.  Please see if this looks good to you.

lbibl (approved):

srikanthallu (commented):
@adam-a-a , do you know why there are two tests that xfail ? Are we marking this anywhere ?

lbibl:
> @adam-a-a , do you know why there are two tests that xfail ? Are we marking this anywhere ?

I think that (xfails) is only associated with an experimental MacOS and should not cause concerns.  @bknueven might know more about this.

bknueven:
> > @adam-a-a , do you know why there are two tests that xfail ? Are we marking this anywhere ?
> 
> I think that (xfails) is only associated with an experimental MacOS and should not cause concerns. @bknueven might know more about this.

If you're seeing this on macOS its because these tests did not pass using Ipopt with MuMPS at some point. We should probably try removing the `@pytest.mark.requires_idaes_solver` to make sure they're really still necessary.

bknueven:
Ideally we'd also refactor this test (probably in a separate PR) to avoid such issues in the future.

bknueven (approved):

PR #740, opened 2022-09-16
Lazily constructing cost attribute blocks

bknueven:
## Summary/Motivation:
The number of typically unused parameters on the WaterTAP Costing block can be confusing to users. This PR would delay construction of those parameters until they are used.

- Add `LazyBlock` and `LazyBlockMixin` to facilitate on-demand block construction
- Use `LazyBlock`s for WaterTAPCosting parameters.

## Discussion:
- This may be too clever for our use-case -- the unit model costing methods could just call-out to construct the block themselves
- Conversely, if we do not implement something like this PR, then any user implementing their own costing method utilizing the global parameters needs to know to call the parameter block construction.
- We could "meet in the middle" and not have these blocks constructed on access, but instead via some method call.

### Legal Acknowledgement


bknueven:
As part of, or maybe before, the way flows cost attributes are created should be reconsidered with IDAES/idaes-pse#983.

bknueven:
Closing in favor of #835.

Issue #741, opened 2022-09-17
Fix example in how to guide


wigging:
The example given in the WaterTAP documentation for [How to run models in a Python script](https://watertap.readthedocs.io/en/latest/how_to_guides/how_to_run_models_in_a_py_script.html) does not run. The reverse osmosis model is imported as `ReverseOsmosis0` but should be imported as `ReverseOsmosis0D`.

wigging:
This was fixed with the merged pull request #742.

PR #742, opened 2022-09-17
Fix ReverseOsmosis0D model in documentation example

wigging:
## Fixes/Resolves:

Fixes issue #741

## Summary/Motivation:

The example given in the WaterTAP documentation for [How to run models in a Python script](https://watertap.readthedocs.io/en/latest/how_to_guides/how_to_run_models_in_a_py_script.html) does not run. The reverse osmosis model is imported as `ReverseOsmosis0` but should be imported as `ReverseOsmosis0D`. This pull request uses the proper import statement for the example.

- Replace `ReverseOsmosis0` with `ReverseOsmosis0D`
- Remove unnecessary spaces

### Legal Acknowledgement

### DIFF
--- a/docs/how_to_guides/how_to_run_models_in_a_py_script.rst
+++ b/docs/how_to_guides/how_to_run_models_in_a_py_script.rst
@@ -56,7 +56,7 @@ However, when writing a script, it is recommended (and in some cases, required; 
 Example: Python file with recommended structure
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-.. doctest::
+.. testcode::
 
    # Import concrete model from Pyomo
    from pyomo.environ import ConcreteModel
@@ -115,7 +115,7 @@ Example: Python file with recommended structure
 Example: the same code without recommended structure (may cause errors on Windows)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-.. doctest::
+.. code-block::
 
    # Import concrete model from Pyomo
    from pyomo.environ import ConcreteModel
--- a/docs/how_to_guides/how_to_run_models_in_a_py_script.rst
+++ b/docs/how_to_guides/how_to_run_models_in_a_py_script.rst
@@ -67,10 +67,10 @@ Example: Python file with recommended structure
    # Import utility tool for calculating scaling factors
    from idaes.core.util.scaling import calculate_scaling_factors
    # Import RO model
-   from watertap.unit_models.reverse_osmosis_0D import ReverseOsmosis0
-   # import the solver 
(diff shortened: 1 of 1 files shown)


bknueven:
This code being run as part of the doc test? We should do that as part of this PR.

wigging:
I don't think the example in the docs is run as part of the test suite; otherwise, the tests would have failed. I don't know how the documentation is tested but I'll investigate adding a test for this particular example.

bknueven:
> I don't think the example in the docs is run as part of the test suite; otherwise, the tests would have failed. I don't know how the documentation is tested but I'll investigate adding a test for this particular example.

Here's an example file: https://raw.githubusercontent.com/watertap-org/watertap/main/docs/how_to_guides/how_to_setup_simple_RO.rst

Note the use of the `.. doctest::` directive which can replace a code-block, and the `.. testsetup::` directive which is executed for the test but not shown in the documentation.

That said, it may be a bit complex with this particular set-up to actually run the code. But I think we can use `.. testcleanup::` directive to execute `main`

wigging:
@bknueven I added the `.. doctest::` directives.

docs/how_to_guides/how_to_run_models_in_a_py_script.rst
```diff
@@ -90,32 +90,32 @@ Example: Python file with recommended structure
...
+
        # Solve the model (using the tee=True option to display solver info)
-       results = solver.solve(m, tee=True)
+       solver.solve(m, tee=True)
+
 
-   # Call that function in the "__main__" for the script 
+   # Call that function in the "__main__" for the script
    if __name__ == "__main__":
        main()
```
bknueven, line 114:
```suggestion
.. testcleanup::

   # run `main`
   main()

```

This way `main` get executed as part of the test, and not just checked for syntax.

wigging: This gives an error about `NameError: name 'main' is not defined`.

bknueven: Yeah. That's because I misremembered above and we should be using the `.. testcode::` (and *not* doctest) directive for the main blocks.

However, this seems to be a pervasive issue in the documentation which should be addressed separately. Therefore, I suggest you revert 08a1d8c (as it compounds this issue) and I'll deal with that separately.

This code doesn't even work as-is (with your correction), so we need to fix it and probably a lot of other examples.

wigging: I used the `.. testcode::` directive for the corrected example and used `.. code-block::` for the example that may cause errors on Windows. With these changes everything is working fine.

bknueven: Yes, that will test fine for now. Of course, `main` is still not executed, so we're not seeing the errors from that.

But that is okay, I will try to address this later as part of another PR.

bknueven (approved):

wigging:
@bknueven Looks like two reviewers are needed to approve this pull request. Who should be the second reviewer?

adam-a-a (approved):

PR #743, opened 2022-09-19
AMO 1575 : Applications 1 & 4

MarcusHolly:
## Summary/Motivation:
AMO 1575 Applications 1 & 4 : High-Rate Contact Stabilization (HR-CS) & Magprex Struvite Precipitation

-
-

### Legal Acknowledgement


MarcusHolly:
@adam-a-a or @tarnold17 The results look pretty good in my opinion, but I'm a bit uncertain about the MgCl2 price result in the magprex flowsheet. It comes out to being zero since the pricing is based on the kg of struvite produced, and if you look at the flowrate of product struvite, it is very low (~1e-5). I'm just wondering if this is okay since I haven't seen other flowsheets that show a cost as $0.

watertap/unit_models/zero_order/magprex_zo.py
```diff
@@ -0,0 +1,75 @@
...
+    def build(self):
+        super().build()
+
+        self._tech_type = "magprex"
+
+        build_sido_reactive(self)
+        constant_intensity(self)
+
+        self.magnesium_chloride_dosage = Var(
+            units=pyunits.dimensionless,
```
adam-a-a, line 44: Why is this dimensionless? What units does it represent?

adam-a-a: I see-  kg-MgCl2/kg struvite

watertap/unit_models/zero_order/magprex_zo.py
```diff
@@ -0,0 +1,75 @@
...
+
+        @self.Constraint(
+            self.flowsheet().time,
+            doc="Constraint for magnesium chloride demand based on sludge flowrate.",
+        )
+        def MgCl2_demand(b, t):
+            return b.MgCl2_flowrate[t] == (
+                b.magnesium_chloride_dosage
+                * pyunits.convert(
+                    b.properties_byproduct[t].flow_mass_comp["phosphates"],
```
adam-a-a: Should this be byproduct flow rate of struvite instead?

watertap/unit_models/zero_order/magprex_zo.py
MarcusHolly: I based it off the electrochemical_nutrient_removal unit model: https://github.com/watertap-org/watertap/blob/main/watertap/unit_models/zero_order/electrochemical_nutrient_removal_zo.py

adam-a-a:
It looks like your tests are failing when you assert the value of very small numbers (e.g., 1e-15). Maybe you can try asserting such values to be 0 and use and absolute tolerance instead of a relative tolerance.

MarcusHolly:
> 

Yep, I've fixed the issue already. I'm just running all the tests locally first in case anything else goes wrong.

watertap/core/zero_order_costing.py
```diff
@@ -3079,6 +3203,53 @@ def cost_peracetic_acid(blk):
...
+            initialize=1,
+            units=blk.config.flowsheet_costing_block.base_currency,
+            bounds=(0, None),
+            doc="Capital cost of unit operation",
+        )
+
+        expr = pyo.units.convert(
+            ref_state * HRT * size_cost,
+            to_units=blk.config.flowsheet_costing_block.base_currency,
+        )
```
tarnold17, line 3313: In the four new costing functions you have written, the costing expression is `ref_state * HRT * size_cost`, where `ref_state`, `HRT`, and `size_cost` are all constants specified in the YAML. So then the cost is a constant. Unless there is a reason for doing this that I am not seeing, I think that in all of these costing expressions you want to replace `ref_state` with the actual volumetric flowrate from the model, e.g.

    expr = pyo.units.convert(
        blk.unit_model.properties_in[t0].flow_vol * HRT * size_cost,
        to_units=blk.config.flowsheet_costing_block.base_currency,
    )

Then, the `reference_state` entry can be removed from all of the YAML files.

MarcusHolly: Ok, that makes sense. So why do some of the models in zero_order_costing use these 3 parameters together. I was specifically looking at electrochemical_nutrient_removal since it has some similarities.

adam-a-a: @MarcusHolly I recall that @luohezhiming might have updated electrochemical_nutrient_removal by replacing state_ref with actual flowrate.

MarcusHolly: Currently I am using the existing ZO model for the clarifier, which still uses reference_state. Should I leave that as is or get rid of the reference_state and add a function for clarifiers into zero_order_costing?

tarnold17: As far as I can tell, electrochemical nutrient removal is the only model which is currently costed this way. I have no idea why it is like that. @adam-a-a, it looks to me like Chenyu has not updated it.

As for the ZO clarifier model, I don't see an associated costing method for it in `zero_order_costing.py`, so I'm not sure what to make of that....

MarcusHolly: @adam-a-a  or @tarnold17 The process also has FeCl3 (a coagulant) being fed into the clarifier. So is it better to make a zero_order_costing function that includes FeCl3 for clarifiers (and remove the reference state from its yaml) or just leave things as they are and ignore the FeCl3 costs? I don't see a reason why I would not choose to do the former.

tarnold17: Yeah I think it makes sense to calculate the FeCl3 costs. I probably wouldn't include FeCl3 as a component in the model though. Do you have any data that tells you how much FeCl3 is required? E.g., do you know if it is 1 g FeCl3 required per m3 of influent, or something like that?

MarcusHolly: Yea, they provide "lb FeCl3/MG", and I haven't included it as a component in the model. I'll be updating this PR as soon as I touch up on a few more things.

watertap/core/zero_order_costing.py
```diff
@@ -3032,6 +3010,229 @@ def cost_vfa_recovery(blk):
...
+        blk.config.flowsheet_costing_block.cost_flow(
+            blk.unit_model.polymer_demand[t0], "polymer"
+        )
+
+    def cost_clarifier(blk, number_of_parallel_units=1):
+        """
+        General method for costing clarifiers. Costing is carried out
+        using either the general_power_law form or the standard form which
+        computes HRT and sizing costs.
```
adam-a-a:
```suggestion
        computes HRT, sizing costs, and chemical input costs.
```

adam-a-a (approved):
LGTM - any modifications can be made in a subsequent PR when connecting to GUI.

watertap/unit_models/zero_order/tests/test_magprex_zo.py
```diff
@@ -0,0 +1,370 @@
...
+
+class TestMagprexZO:
+    @pytest.fixture(scope="class")
+    def model(self):
+        m = ConcreteModel()
+        m.db = Database()
+
+        m.fs = FlowsheetBlock(dynamic=False)
+        m.fs.params = WaterParameterBlock(
+            default={"solute_list": ["phosphates", "struvite"]}
```
tarnold17:
```suggestion
            solute_list=["phosphates", "struvite"]
```
I'm not sure how many other places you have used `default` in this PR, but you might consider removing them in order to prevent deprecation warnings from IDAES.

watertap/core/zero_order_costing.py
```diff
@@ -3032,6 +3010,229 @@ def cost_vfa_recovery(blk):
...
+            expr *= blk.config.flowsheet_costing_block.TIC
+
+        blk.capital_cost_constraint = pyo.Constraint(expr=blk.capital_cost == expr)
+
+        # Register flows
+        blk.config.flowsheet_costing_block.cost_flow(
+            blk.unit_model.electricity[t0], "electricity"
+        )
+        blk.config.flowsheet_costing_block.cost_flow(
+            blk.unit_model.MgCl2_flowrate[t0], "magnesium_chloride"
```
tarnold17, line 3060: You've added magnesium chloride as a registered flow in the costing for Magprex, but magnesium chloride isn't actually a component in the flowsheet. Instead, you have `"solute_list": ["phosphates", "struvite"]`. So then does adding magnesium chloride as a registered flow actually have any effect?

watertap/unit_models/zero_order/tests/test_magprex_zo.py
MarcusHolly: Good catch, this one must've slipped through the cracks, but I'll double check to see if I got all of them

watertap/core/zero_order_costing.py
MarcusHolly: From my understanding, MgCl2 is just a coagulant, so I didn't think it had to included as a component in the solute list. It doesn't have any effect on the flowsheet outside of the costing information (the cost of the MgCl2).

watertap/core/zero_order_costing.py
```diff
@@ -3032,6 +3010,229 @@ def cost_vfa_recovery(blk):
...
+                        number_of_parallel_units parallel units (default: 1)
+        """
+        # Get cost method for this technology
+        cost_method = _get_unit_cost_method(blk)
+        valid_methods = ["cost_power_law_flow", "cost_HRCS_clarifier"]
+        if cost_method == "cost_power_law_flow":
+            ZeroOrderCostingData.cost_power_law_flow(blk, number_of_parallel_units)
+        elif cost_method == "cost_HRCS_clarifier":
+            # NOTE: number of units does not matter for cost_HRCS_clarifier
+            #       as its a linear function of membrane area
```
tarnold17, line 3176:
```suggestion
            #       as it's a linear function of membrane area
```
The nittiest of nitpicks.

watertap/data/techno_economic/hrcs.yaml
```diff
@@ -0,0 +1,50 @@
...
+    oxidation:
+      key_reactant: cod
+      conversion: 0.21
+      stoichiometry:
+        cod:
+          conversion_ratio: -1
+        oxygen:
+          conversion_ratio: -1
+        carbon_dioxide:
+          conversion_ratio: 1
```
tarnold17, line 50: Another very nitty nitpick, but I recommend leaving a newline at the end of files as a general rule of thumb.

tarnold17 (approved):
You probably want to add an empty file `watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1575_hrcs/tests/__init__.py`. And I have two extremely nitpicky comments that could be ignored with no adverse effects. Great work here!

PR #744, opened 2022-09-20
Electrodialysis 1D refinement 

lbibl:
- Added nonohmic potential simulations 
- Added Nernst diffusion layer simulations 
- Improved scaling, constraints and initialization  
- Other improvements 

### Legal Acknowledgement


lbibl:
Please refer to PR #745 for an updated ED 1d documentation with key content (e.g., equations) underlying the refinement here.

lbibl:
> Please refer to PR #745 for an updated ED 1d documentation with key content (e.g., equations) underlying the refinement here.

Link to the generated doc: https://watertap--745.org.readthedocs.build/en/745/technical_reference/unit_models/electrodialysis_1D.html

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -103,11 +117,56 @@ class Electrodialysis1DData(UnitModelBlockData):
...
-            domain=In(["Constant_Current", "Constant_Voltage"]),
+            default=ElectricalOperationMode.Constant_Voltage,
+            domain=In(ElectricalOperationMode),
             description="The electrical operation mode. To be selected between Constant Current and Constant Voltage",
         ),
     )
+    CONFIG.declare(
+        "limiting_current_density_method",
+        ConfigValue(
+            default=LimitingCurrentDensityMethod.Constant_input,
```
srikanthallu:
```suggestion
            default=LimitingCurrentDensityMethod.Constant,
```

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -39,19 +39,33 @@
...
 import idaes.logger as idaeslog
+from enum import Enum
 
 __author__ = "Xiangyu Bi, Austin Ladshaw"
 
 _log = idaeslog.getLogger(__name__)
 
+
+class LimitingCurrentDensityMethod(Enum):
+    Constant_input = 0
```
srikanthallu:
```suggestion
    Constant = 0
```

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -453,13 +511,14 @@ def _make_performance(self):
...
-            bounds=(0, 1000),
-            units=pyunits.amp,
-            doc="Current across a cell-pair or stack",
-        )
+        if self.config.operation_mode == ElectricalOperationMode.Constant_Current:
+            self.current_applied = Var(
+                self.flowsheet().time,
+                initialize=1,
+                bounds=(0, 1000),
+                units=pyunits.amp,
```
srikanthallu, line 518: Is the constant current boundary condition usually defined in Amps or Amp/cm^2 ?

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -559,50 +655,159 @@ def eq_isothermal_concentrate(self, t, x):
...
-            )
 
         @self.Constraint(
             self.flowsheet().time,
             self.diluate.length_domain,
             doc="calcualte current density from the electrical input",
         )
         def eq_get_current_density(self, t, x):
-            if self.config.operation_mode == "Constant_Current":
+            if self.config.operation_mode == ElectricalOperationMode.Constant_Current:
```
srikanthallu, line 703: If "ElectricalOperationMode" is content current, we do not consider the case of "has_nonohmic_potential_membrane" and "has_Nernst_diffusion_layer" ?

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -559,50 +655,159 @@ def eq_isothermal_concentrate(self, t, x):
...
+                            * self.total_areal_resistance_x[t, x]
+                            == self.voltage_applied[t]
+                        )
 
         @self.Constraint(
             self.flowsheet().time,
             self.diluate.length_domain,
             doc="calcualte length_indexed voltage",
         )
         def eq_get_voltage_x(self, t, x):
```
srikanthallu, line 762: May be I am confused here, why is this constraint for "eq_get_voltage_x" repeated same as "eq_get_current_density" ?

srikanthallu (commented):
Overall looks good, very extensive refinement. Left minor comments, also there are few simple unit tests that are missing which can be added for any raised ConfigurationErrors in the model.

watertap/unit_models/electrodialysis_1D.py
lbibl: No , they are accounted for but do not affect the current density calculation under the "constant_current" mode, in which current density is fully determined by the input regardless of how charges are distributed. So it is reflected a uniform equation for all configurations.

watertap/unit_models/electrodialysis_1D.py
lbibl: They vary in the right-hand side of the equation as `voltage_applied` or `voltage_x`.  This constraint is essentially for getting the x-dependent voltage under the Const_Curr mode, while it simply says `v_x == v_applied`  in the other mode.  So this single constraint works for both modes.

watertap/unit_models/electrodialysis_1D.py
lbibl: They are equally usual I think.  Practically, I think they control current (ampere) and translate it to density (ampere m^-2) by accounting for the surface area.  Using current here save the user's effort to do the calculation.

watertap/unit_models/electrodialysis_1D.py
lbibl: After a second thought, I rename it to be `InitialValue` because the limiting current density isn't exactly constant over the cell length. Note is given as doc when declaring the config.

lbibl:
> Overall looks good, very extensive refinement. Left minor comments, also there are few simple unit tests that are missing which can be added for any raised ConfigurationErrors in the model.

The missed configerr hit is to raise an error when H2O is not in the component.  This isn't tested because the dspmde ion property pack has H2O  built as a component by default.  In so far as dspmde is used, this error cannot be triggered.  Since it is only 2 or 3 lines, I would leave it as is and test it when other property pack is used in the future.

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -103,11 +117,56 @@ class Electrodialysis1DData(UnitModelBlockData):
...
+            domain=In(LimitingCurrentDensityMethod),
+            description="Configuration for method to compute the limiting current density",
+            doc="""
+           **default** - ``LimitingCurrentDensityMethod.InitialValue``
+
+       .. csv-table::
+           :header: "Configuration Options", "Description"
+
+           "``LimitingCurrentDensityMethod.InitialValue``", "Limiting current is calculated from a single initial value of the feed solution tested by the user." 
+           "``DensityCalculation.seawater``", "Limiting current density is caculated from the empirical equation: TODO "
```
adam-a-a: Minor typo - need to update from densitycalculation

```suggestion
           "``DensityCalculation.seawater``", "Limiting current density is calculated from the empirical equation: TODO "
```

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -103,11 +117,56 @@ class Electrodialysis1DData(UnitModelBlockData):
...
+        ),
+    )
+
+    CONFIG.declare(
+        "limiting_current_density_data",
+        ConfigValue(
+            default=500,
+            description="Limiting current density data input",
+        ),
+    )
```
adam-a-a, line 149: Why set this as a config option when if could just be a variable in the model that user could fix?

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -255,15 +316,30 @@ def build(self):
...
+        add_object_reference(self, "ion_set", self.config.property_package.ion_set)
+
+        add_object_reference(
+            self, "cation_set", self.config.property_package.cation_set
+        )
+        add_object_reference(self, "anion_set", self.config.property_package.anion_set)
+        self.membrane_set = Set(
+            initialize=["cem", "aem"]
+        )  #   cem = Cation-Exchange Membrane aem = Anion-Exchange Membrane
+        self.electrode_side = Set(initialize=["cathode_left", "anode_right"])
```
adam-a-a: Just being nitpickish here, but how about `electrode_side_set`, i.e., ending the name with `_set` do be consistent with the other set naming convention you have in the model?

watertap/unit_models/electrodialysis_1D.py
lbibl: This config would only be used when the `InitialValue` method (the other config) is selected.  The actual "i_lim" is a x-dependent var in the 1-d model and is calculated from this single data from user input. Adding it as a var will cause larger changes in the model structure when other methods are added.   This way also makes the scaling easy.  So I feel it would be efficient to set it up like this.

adam-a-a: Fair enough!

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -778,6 +1070,497 @@ def eq_recovery_mass_H2O(self, t):
...
+            doc="Thickness of the diffusion layer",
+        )
+
+        @self.Constraint(
+            self.flowsheet().time,
+            self.diluate.length_domain,
+            doc="Calculate length-indexed limitting current density",
+        )
+        def eq_current_dens_lim_x(self, t, x):
+            if (
```
adam-a-a, line 1193: Should there be a conditional that raises an exception if anything other that InitialValue is provided?

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -778,6 +1070,497 @@ def eq_recovery_mass_H2O(self, t):
...
+            self.config.property_package.ion_set,
+            doc="Establish relationship between interfacial concentration polarization ratio and current density",
+        )
+        def eq_conc_polarization_ratio(self, mem, side, t, x, j):
+            if (mem == "cem" and side == "cathode_left") or (
+                mem == "aem" and side == "anode_right"
+            ):
+                return self.conc_mem_surf_mol_x[
+                    mem, side, t, x, j
+                ] / self.concentrate.properties[t, x].conc_mol_phase_comp["Liq", j] == (
```
adam-a-a, line 1225: Might it be safer to multiply the denominator to the other side rather than keep here in the denominator? The same things goes for your alternative constraint below where you divide by diluate on the LHS.

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -778,6 +1070,497 @@ def eq_recovery_mass_H2O(self, t):
...
+                    * (
+                        sum(
+                            self.diluate.properties[t, x].trans_num_phase_comp["Liq", j]
+                            / self.config.property_package.charge_comp[j]
+                            for j in self.cation_set
+                        )
+                        + sum(
+                            self.diluate.properties[t, x].trans_num_phase_comp["Liq", j]
+                            / self.config.property_package.charge_comp[j]
+                            for j in self.anion_set
```
adam-a-a, line 1256: out of curiosity, could this be reduced to a single summation across `self.ion_set`?

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -778,6 +1070,497 @@ def eq_recovery_mass_H2O(self, t):
...
+                * (
+                    sum(
+                        self.ion_trans_number_membrane[mem, j]
+                        / self.config.property_package.charge_comp[j]
+                        for j in self.cation_set
+                    )
+                    + sum(
+                        self.ion_trans_number_membrane[mem, j]
+                        / self.config.property_package.charge_comp[j]
+                        for j in self.anion_set
```
adam-a-a, line 1138: Can you reduce this to summing across ion_set?

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -778,6 +1070,497 @@ def eq_recovery_mass_H2O(self, t):
...
+                        self.conc_mem_surf_mol_x[mem, "cathode_left", t, x, j]
+                        for j in self.ion_set
+                    )
+                    / sum(
+                        self.conc_mem_surf_mol_x[mem, "anode_right", t, x, j]
+                        for j in self.ion_set
+                    )
+                )
+            )  # Note: the log argument is taken as *total ion concentration* while in the original equation as *electrolyte concentration*
+            # This is mathematically valid as the ratio eliminate the difference and electro-neutrality stands.
```
adam-a-a, line 1152:
```suggestion
            # This is mathematically valid as the ratio eliminate the difference and electro-neutrality stands.
```
Interesting - I _think_ I understand!

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -778,6 +1070,497 @@ def eq_recovery_mass_H2O(self, t):
...
+                        sum(
+                            self.diluate.properties[t, x].trans_num_phase_comp["Liq", j]
+                            / self.config.property_package.charge_comp[j]
+                            for j in self.cation_set
+                        )
+                        + sum(
+                            self.diluate.properties[t, x].trans_num_phase_comp["Liq", j]
+                            / self.config.property_package.charge_comp[j]
+                            for j in self.anion_set
+                        )
```
adam-a-a, line 1297: Same comment on reducing to ion_set summation

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -778,6 +1070,497 @@ def eq_recovery_mass_H2O(self, t):
...
+                        ** -1
+                        * sum(
+                            self.concentrate.properties[t, x].conc_mol_phase_comp[
+                                "Liq", j
+                            ]
+                            for j in self.ion_set
+                        )
+                        ** -1
+                    )
+                )
```
adam-a-a, line 1321: I am looking at the difference between what's in your if clause vs your else clause, and it's hard to tell due to the black formatting. From a glance, I am wondering why you can't condense this all rather than duplicating the expression as a whole

adam-a-a: Ah- concentrate and diluate concentrations and whether their sums are in the numerator or denominator-- seems to be the difference between the two log expressions. Maybe in your if-else, you could just have an alias for that ratio which changes based on the conditional, then you could just have the expression written out once with that alias in place.

lbibl: Thanks for the good suggestion.  Since this change will not actually alter the model structure or computational efficiency, plus I strongly think the overall constraint expressions could need major refinement in the next nonlinear optimization stage, can we leave this as is for now?

adam-a-a: Sure! This change wouldn't really do much except reduce lines of code, so need to incorporate now.

srikanthallu (approved):
LGTM

adam-a-a (approved):

PR #754, opened 2022-09-23
Fix use of `.. doctest::` directive in documentation

### LINKED ISSUE #746: Fix incorrect use of `.. doctest::` directive in documentation
In most places, `.. doctest::` should instead be `.. testcode::`.

Currently, lots of code-blocks in the documentation are not getting tested.

wigging:
## Fixes/Resolves:

Fixes issue #746 regarding use of the `.. doctest::` directive in example code in the documentation.

## Summary/Motivation:

The `.. doctest::` directive is not properly implemented in the documentation for code examples.

- Replace `.. doctest::` with `.. code-block::` or `.. testcode::` directives

### Legal Acknowledgement


wigging:
@bknueven I replaced the `.. doctest::` directives with `.. code-block::` for code examples in the documentation. Everything is in the How To Guides section. The `.. testcode::` directive doesn't make sense for some of the examples so I'm only applying it where applicable. However, I'm already running into issues when running doctest locally. For example, in `how_to_use_a_property_model.rst` I get the following exception from doctest:

```
Exception raised:
      File "/opt/miniconda3/envs/watertap/lib/python3.10/site-packages/idaes/core/base/property_base.py", line 965, in __getattr__
        if m[attr]["method"] is None:
    KeyError: 'molality_comp'

    During handling of the above exception, another exception occurred:

      File "/opt/miniconda3/envs/watertap/lib/python3.10/doctest.py", line 1346, in __run
        exec(compile(example.source, filename, "single",
      File "<doctest default[0]>", line 39, in <module>
        m.fs.state_block[0].molality_comp['NaCl']
      File "/opt/miniconda3/envs/watertap/lib/python3.10/site-packages/idaes/core/base/property_base.py", line 1011, in __getattr__
        raise PropertyNotSupportedError(
    idaes.core.util.exceptions.PropertyNotSupportedError: fs.state_block[0] package property metadata method does not contain a method for molality_comp. Please select a package which supports the necessary properties for your process.
```

But if I run the example in that document as a stand-alone script from the command line then I don't get any errors. Any idea why using `.. testcode::` would fail with doctest?

bknueven:
What's happening in 81ea0c9? It seems to undo a lot of the existing infrastructure we have set-up for doing documentation testing. This should be left in-place.

There's also a bunch of formatting changes which, while I'm supportive of, should happen in a separate PR to make this one easier to review.

wigging:
I removed all of the tests because some of them didn't actually test anything. I'm going through each code example and adding tests where it makes sense. This is a draft pull request so don't worry about the commits right now. But my question from above remains; I'm not getting the same results from the doctest extension compared to running code as a script.

bknueven:
> I removed all of the tests because some of them didn't actually test anything. I'm going through each code example and adding tests where it makes sense.

Yes -- several of the existing test were "check that this code executes without an Exception". That's still useful, especially in documentation, where API changes (either to WaterTAP or an upstream package) are liable to make our documentation slowly out-of-date. The current checks at least force the developer to make sure they updated the documentation if they make some breaking change.

Ideally *any* code snippet in the documentation would minimally have the "check that this code executes without an Exception". We should preserve this where we currently have it.

wigging:
I'm getting the following errors from the doctest tests:

- For the example code in `how_to_guides/how_to_use_a_property_model.rst`, there is an error about `molality_comp` method is not available.
- For the example code in `how_to_guides/how_to_setup_simple_RO.rst`, there is an exception about fs.unit degrees of freedom were not 0 at the beginning of initialization. DoF = 3.

Anyone have suggestions on how to fix these?

wigging:
Made various fixes and all the documentation tests are passing. I focused on the sections in the How To Guides. If there are other areas of the documentation that need testing, then it can be done in a separate pull request.

docs/how_to_guides/how_to_setup_simple_chemistry.rst
```diff
@@ -461,3 +459,10 @@ In the example code above, we show how to setup the thermo and reaction packages
 and place them into the `EquilibriumReactor` unit model, but do not go further.
 Additional instructions for setting up and solving unit models can be found at
 `IDAESWorkflow`_.
+
+.. testoutput::
+   :hide:
+   :options: +ELLIPSIS
+
+   WARNING: DEPRECATED: The default argument for the ProcessBlock class is
+       ...
```
adam-a-a, line 468: I'm guessing this tests that the deprecation warning are hidden, and this links the use of an ellipsis to what should be hidden? Just curious to confirm how this works given the unfamiliar (to me) syntax.

adam-a-a: Oh, the ellipsis is a wildcard actually. Also, posting these links for anyone else who's curious:
https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html#directive-testoutput
https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html#configuration

bknueven: Should we try to fix this particular deprecation warning? We'll probably need to soon regardless.

wigging: The deprecation warnings don't play well with doctest. I had to add the warning to the testoutput to get the test to pass. I can fix this in a future pull request after these warnings get fixed. But for now I think this is fine.

wigging:
This also fixes the syntax highlighting for the code examples in the How To Guides section.

docs/how_to_guides/how_to_setup_simple_chemistry.rst
lbianchi-lbl, line 468: As a note for later: this will need to be removed once we get rid of the deprecation warnings.

docs/how_to_guides/how_to_use_inherent_reactions.rst
```diff
@@ -289,15 +288,21 @@ code below.
...
     # At this point, you can 'fix' your inlet/outlet state conditions,
     #     setup scaling factors, initialize the model, then solve the model
     #     just as you would with any other IDAES flowsheet
+
+.. testoutput::
+   :hide:
+   :options: +ELLIPSIS
+
+   WARNING: DEPRECATED: The default argument for the ProcessBlock class is
+       ...
```
lbianchi-lbl, line 308: Same re: need to remove once deprecation warnings are addressed.

docs/how_to_guides/how_to_use_a_property_model.rst
```diff
@@ -55,16 +53,115 @@ users to model the chemical and physical properties of simple systems without th
...
+           None : 10000.0 : 5000000.0 : 50000000.0 :  True :  True : NonNegativeReals
+       dens_mass_phase : Mass density
+           Size=1, Index=fs.properties.phase_list, Units=kg/m**3
+           Key : Lower : Value   : Upper  : Fixed : Stale : Domain
+           Liq : 500.0 : 1021.46 : 2000.0 : False : False :  Reals
+       mass_frac_phase_comp : Mass fraction
+           Size=2, Index=fs.state_block[0].mass_frac_phase_comp_index
+           Key             : Lower : Value : Upper : Fixed : Stale : Domain
+            ('Liq', 'H2O') :   0.0 : 0.965 :  None : False : False :  Reals
+           ('Liq', 'NaCl') :   0.0 : 0.035 :  None : False : False :  Reals
```
lbianchi-lbl: As I understand, the expected output defined here will be compared verbatim (i.e. as a single multiline string), which would mean that the tests will fail if it the generated output differs from it by even a single decimal digit.

I remember having some discussions about the pros and cons of testing human-readable output, and as I remember the conclusion was that it leads to very brittle tests without any significant upside.

I'm not sure exactly how the `...` wildcard works, but perhaps we should limit the expected output specification to high-level information that we know will always be present (for this particular model), and use a wildcard for anything depending on actual numerical results (the validity of which is tested elsewhere).

lbianchi-lbl (commented):
@wigging Thanks for all the work on this, I think it'll be useful to have all the code snippets updated and tested properly.

I've just added a couple of notes for ourselves regarding the deprecation warnings (nothing that needs to be addressed for this PR), plus a question about possible test brittleness that I feel should be discussed before merging.

docs/how_to_guides/how_to_use_a_property_model.rst
lbianchi-lbl: (I've just realized that the more "digits-heavy" sections are not shown in the highlighted section, but are below it in the same `testoutput::` block)

docs/how_to_guides/how_to_setup_simple_chemistry.rst
wigging: Yes, this will need to be revised once the deprecation stuff is fixed.

docs/how_to_guides/how_to_use_inherent_reactions.rst
wigging: Yes, this will be revised once the deprecation warnings are fixed.

docs/how_to_guides/how_to_use_a_property_model.rst
wigging: The original example had `m.fs.state_block[0].display()` commented out. Should I just remove that from the example code which would eliminate the output?

wigging: The ellipsis `...` option basically matches any output. So for the output in this example, I can test certain lines and ignore other lines using the ellipsis.

wigging: This checks the first 12 lines of the output. The remaining lines are ignored in the test. Note that this only displays the first 12 lines in the documentation too.

```
.. testoutput::

   Block fs.state_block[0]

     Variables:
       flow_mass_phase_comp : Mass flow rate
           Size=2, Index=fs.state_block[0].flow_mass_phase_comp_index, Units=kg/s
           Key             : Lower : Value : Upper : Fixed : Stale : Domain
            ('Liq', 'H2O') :   0.0 : 0.965 :  None :  True :  True : NonNegativeReals
           ('Liq', 'NaCl') :   0.0 : 0.035 :  None :  True :  True : NonNegativeReals
       temperature : State temperature
           Size=1, Index=None, Units=K
           Key  : Lower  : Value  : Upper  : Fixed : Stale : Domain
           None : 273.15 : 298.15 : 373.15 :  True :  True : NonNegativeReals
       ...
```

bknueven: I think testing the output is okay, as long as the values do not come from some optimization solver.

In this case, I would be comfortable modifying the test to just check the output block in your comment here. The `Variables` in the first 12 lines are all fixed, so we're not likely to get burned by otherwise insignificant changes to the results.

lbianchi-lbl: I agree, I think this solution (showing the first N lines, or any portion of the output that's not likely to change) sounds like a good choice.

wigging: Ok, I'll implement the ellipsis approach and push up the changes.

bknueven (approved):
Definitely a descent direction!

lbianchi-lbl (approved):

PR #756, opened 2022-09-24
Report solver failures and change ModelExport keys

### LINKED ISSUE #755: Report solver failures
Currently solver failures may be hidden by not evaluating the status returned by Pyomo

dangunter:
## Fixes/Resolves:
Fixes #755, #773

## Summary/Motivation:
In addition to the issue above, this PR is being used to fix a semi-related bug for the keys used in the `model_objects` dict of the `FlowsheetExport` class. Previously they were UUIDs, but this means that any time the model is re-built and re-exported, or reloaded from storage, new UUIDs will be generated and they will not match any old keys. This causes a number of errors, but at any rate is solvable by simply making the key the full name of the Pyomo object in question.

- Look at solver status
- Raise exception if it did not find an optimal solution
- Change the key used in `FlowsheetExport.add()` to be the name of the object being added
- Explicitly clear the current saved objects when rebuilding and re-exporting a flowsheet
- Raise an exception if a duplicate object is added with `FlowsheetExport.add()`

### Legal Acknowledgement


watertap/ui/fsapi.py
```diff
@@ -301,7 +348,9 @@ def action_wrapper(**kwargs):
                         "constructor or call `add_action(Actions.export, <function>)` "
                         "on FlowsheetInterface instance."
                     )
-                # run_action will refuse to call the export action directly
+                # clear model_objects dict, since duplicates not allowed
+                self.fs_exp.model_objects = {}
```
lbianchi-lbl: Incredibly minor suggestion:

```suggestion
                self.fs_exp.model_objects.clear()
```

lbianchi-lbl (commented):
I've just before/after tested the ModelExport fix using the same tests being used for #773, and I confirm that it works. I think we can close #773, but maybe we should copy those tests to this PR since they seem to be sensitive to (at least some of) known and potential future ModelExport lifetime issues.

watertap/ui/fsapi.py
dangunter: Nice, didn't know that function existed.

dangunter:
Merged in test from Issue #773 (which this should also close) and minor suggestion on how to clear a dict.

lbianchi-lbl (approved):
This will go a long way to help `FlowsheetExporter` quit its unhealthy object hoarding habit.

bknueven (approved):

PR #758, opened 2022-09-26
AMO 1690 - unit models

tarnold17:
## Fixes/Resolves:

Unit models for AMO 1690.

-
-

### Legal Acknowledgement


watertap/core/zero_order_costing.py
```diff
@@ -3332,6 +3335,101 @@ def cost_struvite_classifier(blk):
...
+            expr *= blk.config.flowsheet_costing_block.TPEC
+        elif factor == "TIC":
+            expr *= blk.config.flowsheet_costing_block.TIC
+
+        blk.capital_cost_constraint = pyo.Constraint(expr=blk.capital_cost == expr)
+
+        # Register flows
+        blk.config.flowsheet_costing_block.cost_flow(
+            blk.unit_model.electricity[t0], "electricity"
+        )
```
adam-a-a, line 3433: My guess is that we weren't provided this information, but I could see a case where we'd want to account for heat demand too, e.g., when elevating fluid temp to 35 C on both sides of the membrane. 

However, if such info wasn't provided, can you just add a todo note here to consider adding heat as a registered flow?

watertap/data/techno_economic/anaerobic_digestion_reactive.yaml
```diff
@@ -0,0 +1,72 @@
...
+      conversion: 0.5
+      stoichiometry:
+        cod:
+          conversion_ratio: -1
+    tkn_reaction:
+      key_reactant: tkn
+      conversion: 0.5
+      stoichiometry:
+        tkn:
+          conversion_ratio: -1
```
adam-a-a, line 57: The good thing about this is that we can easily update these values/reactions as needed in the future, if given feedback from the team on more appropriate values. So we won't worry about how accurate these assumptions may or may not be for now (e.g., TKN includes ammoniacal nitrogen).

I think your comment covers this message well enough. =)

adam-a-a (approved):
LGTM - any further refinement can be settled in a subsequent PR. 

Potential points of refinement subject to team feedback: add heat duty, refine removal/reaction defaults in AD reactive.

Nice job!

watertap/core/zero_order_costing.py
tarnold17: The figure for membrane evaporator shows an optional heater for the inlet stream, and lists the inlet temperature as 37 C. Furthermore, there's a note in their report which says "The fluxes used in the analysis are for 35 C feed temperature; thus, centrate heating was not included [in TCA/LCA]." So I guess 35-37 C is low enough that the heat demand for the feed stream is close to negligible. But you are right that this is something we might want to consider in the future, so I have included a TODO note saying as much.

MarcusHolly (approved):
LGTM

PR #763, opened 2022-09-27
Add GUI for CANDOP

MarcusHolly:
## Summary/Motivation:
Adds GUI for CANDO_P flowsheet

-Adds GUI file

### Legal Acknowledgement


adam-a-a (approved):
LGTM - we might want to add the corresponding entry point to the setup.py file, especially if tests fail due to latest PR by @lbianchi-lbl.

lbianchi-lbl:
> LGTM - we might want to add the corresponding entry point to the setup.py file, especially if tests fail due to latest PR by @lbianchi-lbl.

Good point - at the moment, the only CI check that would fail if a UI module is created but not registered in `setup.py` is the Codecov project coverage one, and even then it's not guaranteed as it depends on whether the module has enough uncovered code to surpass the relaxed tolerance (configured at 1%, i.e. PRs are allowed to cause the project coverage to go down to `current_coverage - current_coverage * 0.01`).

Long story short: reviewers should ensure that each UI-enabling module is registered in `setup.py` (and that the corresponding tests all pass; see also #764).

bknueven (approved):

PR #767, opened 2022-09-28
Add GUI for Groundwater Treatment

MarcusHolly:
## Summary/Motivation:
Adds GUI for groundwater treatment flowsheet

-Adds GUI file
-Adds LCOW calculation to flowsheet

### Legal Acknowledgement


MarcusHolly:
I removed the revenue output section from the GUI because this process's only revenue source comes from the treated water. Should I add back the revenue section with this as the only revenue source and, if so, do I just pull the treated water selling price from another example?

adam-a-a (approved):

adam-a-a:
> I removed the revenue output section from the GUI because this process's only revenue source comes from the treated water. Should I add back the revenue section with this as the only revenue source and, if so, do I just pull the treated water selling price from another example?

I think you can leave it as is for now. In the future, there could be the case of recovering some of the nutrients in the feed to generate revenue, but that's beyond the scope of this PR.

bknueven (approved):

PR #769, opened 2022-09-28
Add GUI for HRCS

MarcusHolly:
## Summary/Motivation:
Adds GUI for high-rate contact stabilization (HRCS) flowsheet

- Adds GUI file

### Legal Acknowledgement


MarcusHolly:
A lot of comments/questions on this one:

---Comments---
A.) Calculations for total opex and capex costs are a bit different than usual because the mixer uses watertap_costing, so we have to combine the watertap costing method with the ZO costing method.

B.) COD is represented as mass flow rather than concentration in the inputs and outputs because we were given the COD feed conditions in tons/day. 

---Questions---
1.) Should I include a mixer unit model data section in the input tab because there are no parameters that the user can modify for a mixer. Is it better to just have a blank text box labelled "Mixer"?

2.) Should I include a mixer and/or separator cost data section in the input tab because. The mixer uses the watertap costing method (as opposed to the ZO costing method), so I suppose I could add another "system costing" section that lists out the parameters given here: https://github.com/watertap-org/watertap/blob/main/watertap/costing/watertap_costing_package.py. But I feel like this would be very confusing to the users. For the primary separator, it uses the default costing method (reference state and capital a & b parameters), which would also likely be very confusing for the users to modify.

3.) Should I include a revenue section in the output tab if the only revenue stream given is the cost of consuming FeCl3 (a coagulant) - see similar question here: https://github.com/watertap-org/watertap/pull/767.

adam-a-a (approved):

bknueven (approved):

PR #773, opened 2022-09-28
fsapi: Ensure references to parent components of exported Pyomo objects are kept valid

lbianchi-lbl:
## Fixes/Resolves:

- Errors showing up intermittently in `watertap-org/watertap-ui` REST API test suite as: `AttributeError: NoneType has no attribute "_units"` when converting units in `FlowsheetInterface.load()`

-
-

### Legal Acknowledgement

### DIFF
--- a/watertap/ui/fsapi.py
+++ b/watertap/ui/fsapi.py
@@ -205,6 +205,9 @@ class FlowsheetInterface:
                 self.add_action(getattr(Actions, name), arg)
             else:
                 raise ValueError(f"'do_{name}' argument is required")
+        # store various objects to prevent them being garbage-collected,
+        # which would result in "dead" weakrefs in Pyomo objects
+        self._objects_to_keep_alive = []
 
     def build(self, **kwargs):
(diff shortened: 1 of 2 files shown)


adam-a-a (approved):
LGTM

bknueven (approved):
@lbianchi-lbl it seems like we never clear the `_objects_to_keep_alive` list?

Does the `FlowsheetInterface` get cleared shortly thereafter? It's probably not a huge deal if these are kept alive forever.

lbianchi-lbl:
> @lbianchi-lbl it seems like we never clear the `_objects_to_keep_alive` list?
> 
> Does the `FlowsheetInterface` get cleared shortly thereafter? It's probably not a huge deal if these are kept alive forever.

@bknueven yes, that's a good point. At the time of creating this PR it was indeed the intended behavior but I think the fix @dangunter implemented in #756 is better long-term. Now that #756 has incorporated the tests from here, we can close this.

PR #788, opened 2022-09-30
Custom Ipopt options for LPs

### LINKED ISSUE #787: Improve performance of ipopt-watertap for linear programs
Ipopt has a few options which are best for linear programming problems; see: https://coin-or.github.io/Ipopt/OPTIONS.html and search for "LP".

Because we already do a huge amount of preprocessing in `ipopt-watertap`, it's trivial to apply these settings automatically if we're solving a LP.

bknueven:
## Summary/Motivation:
Ipopt has different recommended settings for LPs. Might be good to detect and use them automatically since we solve lots of ZO models.

- Set `alpha_for_y` option if the model is an LP
- Enables setting a `0` value in AMO 1690 flowsheet when `0` is the intent.
- Lot of updates to ZO tests. The main effect of this change is to push small values closer to 0, usually by an order of magnitude. However, there are a few exceptions where small values are getting larger, where the original implementer should weigh in:
  - `watertap/unit_models/zero_order/tests/test_centrifuge_zo.py` @MarcusHolly 
  - `watertap/unit_models/zero_order/tests/test_electrochemical_nutrient_removal_zo.py` @luohezhiming 
  - `watertap/unit_models/zero_order/tests/test_magprex_zo.py` @MarcusHolly 

### Legal Acknowledgement


tarnold17 (approved):

MarcusHolly (commented):
Not exactly sure what I should be weighing in on. I see that the values for magprex and the centrifuge are getting slightly larger rather than decreasing, but, at first glance, I have no clue why this would be the case. I assume this discrepancy won't cause any issues b/c the values are still quite close to zero. Would you like me to do some digging as to why this might be the case @bknueven ?

bknueven:
> Not exactly sure what I should be weighing in on. I see that the values for magprex and the centrifuge are getting slightly larger rather than decreasing, but, at first glance, I have no clue why this would be the case. I assume this discrepancy won't cause any issues b/c the values are still quite close to zero. Would you like me to do some digging as to why this might be the case @bknueven ?

First, this is not urgent. Second, the question is more "are you concerned with the new results?". If the answer is "no", that's good enough for me.

For many of these, we should probably be doing an absolute comparison against 0. But I'm not sure what the right threshold is, and I'm guessing for many the value isn't identically 0 (e.g., 99% removal).

MarcusHolly (approved):

Issue #777, opened 2022-09-29
GUI-Flowsheet Implementation Checklist


adam-a-a:
### **We need to connect the following AMO flowsheets to the GUI.**

- [x] 1581 #757 
- [x] 1611 #776 
- [x] 1617 #762 
- [x] 1664 #698 
- [x] 1595 #763 
- [x] 1679 #767 
- [x] 1614 #765 
- [x] 1690 #772 
- [x] 1598 #752 
- [x] 1575 #769 #775 #759 
- [x] 1503 #771 
- [x] 1669 #792

adam-a-a:
### Additional running to-do list for refinement:
### For release
- [x] Add all entry_points in setup.py
- [x] Add images for each flowsheet. Thus far, METAB has one, and dye desal has a temporary one. https://github.com/watertap-org/watertap/issues/782
- [x] Calculation of direct capital costs in flowsheets and/or GUI code may be erroneous if `cost_factor` is not set to `TIC` for corresponding unit yaml files. Fix this once the GUI PRs are in place. #793

lbianchi-lbl:
@adam-a-a thanks for putting the list together. I'll create a draft PR adding the entry points for the flowsheets that have been merged in already but don't have one yet.

MarcusHolly:
An additional item we can could add to the running to-do list is to verify whether the descriptions for variables in the output tab are visible when you hover over them (or maybe it just doesn't work for me).

lbianchi-lbl:
> An additional item we can could add to the running to-do list is to verify whether the descriptions for variables in the output tab are visible when you hover over them (or maybe it just doesn't work for me).

@MarcusHolly That's a good point. Without knowing the details, I'd say it sounds more like a purely UI-specific issue (unless e.g. within the same flowsheet, the description-on-hover works for some of the variables, but not all of them), so maybe opening an issue in watertap-org/watertap-ui would be more fitting.

PR #793, opened 2022-10-04
Updating GUI Cost Factors

MarcusHolly:
## Fixes/Resolves:

https://github.com/watertap-org/watertap/issues/777

- Updates costing factors for all unit models used by the GUIs from "None" to "TIC"
- Only updates the AMO subtype, where applicable

### Legal Acknowledgement


MarcusHolly:
Will this change negatively affect the costing in flowsheets for unit models that are also used outside of AMO projects (e.g. pump_electricity and primary_separator)?

watertap/unit_models/zero_order/tests/test_hydrothermal_gasification_zo.py
```diff
@@ -293,8 +293,6 @@ def test_costing():
     assert_units_consistent(m.fs)
     assert degrees_of_freedom(m.fs.unit) == 0
     initialization_tester(m)
-    results = solver.solve(m)
```
tarnold17, line 296: Removing these two lines does allow this test to pass. But I'm not sure that's what we want. If the call to `solver.solve(m)` is resulting in an error, I think we want the test to reveal that so we can find the underlying issue.

bknueven (approved):

watertap/unit_models/zero_order/tests/test_hydrothermal_gasification_zo.py
bknueven: I changed this to relax the termination condition instead. Not ideal, but better than removing the solve, IMHO.

tarnold17 (approved):

PR #798, opened 2022-10-05
Resolve near-silent errors when building flowsheet interfaces with Param objects

### LINKED ISSUE #796: fsapi: errors occurring in `FlowsheetInterface.build()` are logged but not propagated
https://github.com/watertap-org/watertap/blob/main/watertap/ui/fsapi.py#L274-L290

### LINKED ISSUE #797: Exporting Param objects in UI flowsheets causes other objects defined afterwards to be silently dropped
- This manifested itself in `dye_desalination`, likely as the (currently) only flowsheet where a `Param` is exported before other objects
- One of the symptoms can be seen in the GUI, where the `Input` and `Output` tabs are missing the `System Costs` category and all non-`Feed` categories, respectively
- The issue escaped detection in our test suite due to #796 

A PR that should resolve both the false-negative in the tests and the issue with `Param`s is upcoming.

image
image

lbianchi-lbl:
- [x] Restore re-raising exceptions within `FlowsheetInterface.build()`
- [x] Ensure more explicit validation errors are emitted
- [x] Add support for `Param` objects
- [x] Revisit logic for determining `is_readonly` based on `obj` (from @bknueven)
- [x] Fix various issues that were likely present but ignored because of the `FlowsheetExport.add()` error handling issues
  
image

### Legal Acknowledgement

### DIFF
--- a/watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1690/amo_1690_ui.py
+++ b/watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1690/amo_1690_ui.py
@@ -261,17 +261,6 @@ def export_variables(flowsheet=None, exports=None):
         is_output=False,
     )
     exports.add(
-        obj=fs.me.energy_electric_flow_vol_inlet,
-        name="Electricity intensity",
-        ui_units=pyunits.kWh / pyunits.m**3,
-        display_units="kWh/m3",
-        rounding=2,
-        description="Specific electricity intensity relating the intensity to the volume of product",
-        is_input=True,
-        input_category="Membrane evaporator",
-        is_output=False,
-    )
-    exports.add(
         obj=fs.me.removal_frac_mass_comp[0, "ammonium_as_nitrogen"],
         name="NH4 removal",
         ui_units=pyunits.dimensionless,
--- a/watertap/ui/fsapi.py
+++ b/watertap/ui/fsapi.py
@@ -8,7 +8,7 @@ __author__ = "Dan Gunter"
 import logging
 from collections import namedtuple
 from enum import Enum
-from typing import Callable, Optional, Dict, Union, TypeVar
+from typing import Any, Callable, Optional, Dict, Union, TypeVar
 from types import ModuleType
 from uuid import uuid4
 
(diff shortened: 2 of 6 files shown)


watertap/ui/fsapi.py
```diff
@@ -286,7 +339,7 @@ def build(self, **kwargs):
         try:
             self.run_action(Actions.build, **kwargs)
         except Exception as err:
-            _log.error(f"Building flowsheet: {err}")
+            raise RuntimeError(f"Building flowsheet: {err}") from err
```
bknueven, line 342: This syntax is new to me :-)

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1690/amo_1690_ui.py
```diff
@@ -260,17 +260,6 @@ def export_variables(flowsheet=None, exports=None):
         input_category="Membrane evaporator",
         is_output=False,
     )
-    exports.add(
```
adam-a-a, line 263: Why delete this?

bknueven (approved):
I don't know `pydantic` well enough to comment on that end of the code, but the `pyomo` parts look solid.

watertap/ui/fsapi.py
adam-a-a, line 342: Can we keep both?

adam-a-a (commented):
Looks good! I had a couple minor questions.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1690/amo_1690_ui.py
bknueven: I guess I should have asked here: this component was duplicated, and since the errors aren't suppressed the tests caught that.

lbianchi-lbl: The same fs component was being added twice. This by design should trigger an error when `add()` is called a second time with the same `obj`, but the error was being ignored by the unit tests because of the `_log.error()`-only handling.

watertap/ui/fsapi.py
bknueven: The `RuntimeError` will stop execution and print this message to stderr -- so I'm not sure why we'd want both

lbianchi-lbl: I'm not sure, but errors in `fsapi` might already be handled at the UI backend level, and therefore would be logged at the `WaterTAP-UI` application layer (@MichaelPesce will know more on this).

That said, I don't have anything against doing both the logging and the raising here, since I think the worst that can happen is a redundant log entry.

adam-a-a (approved):
LGTM

lbianchi-lbl:
Just as an example using `dye_desalination`, this is how the UI looks like after the changes in this PR:




(Compare with https://github.com/watertap-org/watertap/issues/797#issue-1398226678 for the "before").

lbianchi-lbl:
I don't understand why the non-admin merge is disabled, since all checks are shown to be passing (possibly a GitHub glitch?):
image

PR #799, opened 2022-10-06
ZO Documentation Check

MarcusHolly:
## Fixes/Resolves:

First point of https://github.com/watertap-org/watertap/issues/595

- Adds a check to ensure all zo models have an associated .rst  file
- Renames membrane_evaporator to membrane_evaporator_zo in zo unit models

### Legal Acknowledgement


docs/technical_reference/unit_models/zero_order_unit_models/automate_rst_file_wt3.py
```diff
@@ -173,7 +164,7 @@ def grab_unit_components_feed(unit_class):
     m.fs.props = WaterParameterBlock(
         solute_list=[
             "toc",
-            "tss",
+            "tkn" "tss",
             "cod",
             "tds",
             "nitrogen",
```
lbianchi-lbl: Without knowing the context, is it possible a comma is missing here?

```suggestion
            "tkn",
            "tss",
            "cod",
            "tds",
            "nitrogen",
```

docs/technical_reference/unit_models/zero_order_unit_models/gas_sparged_membrane_zo.rst
```diff
@@ -13,7 +13,8 @@ See documentation for :ref:`Helper Methods for Electricity Demand<electricity_me
 
 Costing Method
 --------------
-This unit does not include costing.
```
adam-a-a: I remember setting this one as an exception since it didn't include costing. We'll have to look at the automation script changes to see why this is different now.

docs/technical_reference/unit_models/zero_order_unit_models/automate_rst_file_wt3.py
```diff
@@ -279,6 +271,45 @@ def grab_unit_components_feed(unit_class):
...
+# rst file path
+rst_path = os.path.join(
+    os.path.dirname(os.path.abspath(__file__)),
+)
+rst_list = []
+exclude_rst_files = ["index.rst"]
+for f in os.listdir(rst_path):
+    filename = os.fsdecode(f)
+    if filename.endswith(".rst") and filename not in exclude_rst_files:
+        rst_list.append(filename[:-4])
```
lbianchi-lbl: This code should work; this is just a possible suggestion how to make it more compact (and possibly slightly more robust in e.g. how the path to the ZO Python modules is defined) using the `pathlib` module from the standard library:

```suggestion
# at the top of the file, add `from pathlib import Path`

# here `zo` is the imported module object for the ZO package; being a package, it has a `__path__` attribute
# containing (a list of) absolute path to the corresponding filesystem directory
ZO_unit_path = Path(zo.__path__[0])
ZO_unit_list = [
   path.stem
   for path in ZO_unit_path.glob("*.py")
   if path.name not in {"__init__.py"}
]
rst_path = Path(__file__).parent
rst_list = [
   path.stem
   for path in rst_path.glob("*.rst")
   if path.name not in {"index.rst"}
]
```

lbianchi-lbl (commented):
In general this looks good (to the extent that I'm able to say). I've left a question regarding a possible typo, and a possible suggestion on how to make the code more compact for the `.py`-`.rst` file match detection part being added by this PR.

docs/technical_reference/unit_models/zero_order_unit_models/automate_rst_file_wt3.py
MarcusHolly: Yes, thx for catching that!

docs/technical_reference/unit_models/zero_order_unit_models/automate_rst_file_wt3.py
MarcusHolly: Done - thx for the advice!

adam-a-a (commented):
Nice job- I have a couple of comments

docs/technical_reference/unit_models/zero_order_unit_models/anaerobic_digestion_oxidation_zo.rst
```diff
@@ -1,5 +1,5 @@
-Anaerobic Digestion (ZO)
-========================
+Anaerobic Digestion Oxidation (ZO)
```
adam-a-a: I thought I had added a title exception for this to leave out "Oxidation". Not sure how it came back, but can you check on this and see if we could get it back to omitting "Oxidation"?

docs/technical_reference/unit_models/zero_order_unit_models/automate_rst_file_wt3.py
```diff
@@ -279,6 +273,24 @@ def grab_unit_components_feed(unit_class):
...
+]
+# print(f"List of zo unit models: {ZO_unit_list}")
+# print(f"List of rst files: {rst_list}")
+
+# Check if lists are equal
+if sorted(ZO_unit_list) == sorted(rst_list):
+    print("Documentation has successfully been created for all ZO unit models")
+else:
+    difference = sorted(list(set(ZO_unit_list) - set(rst_list)))
+    print(f"The following ZO unit model files do not have an rst file: {difference}")
```
adam-a-a: The functionality seems good, but I had envisioned having a test file that would enforce the addition of new models. That is, if a developer adds a new ZO model but forgets to (a) include it in the excel file and (b) rerun the automation script, a test fails.

docs/technical_reference/unit_models/zero_order_unit_models/anaerobic_digestion_oxidation_zo.rst
MarcusHolly: Oh, yea I changed the name in the Excel file to explicitly distinguish it from Anaerobic Digestion Reactive. But I can remove the oxidation again if it's not necessary to distinguish the two.

docs/technical_reference/unit_models/zero_order_unit_models/automate_rst_file_wt3.py
MarcusHolly: I see - I'll work on creating this test file then.

docs/technical_reference/unit_models/zero_order_unit_models/gas_sparged_membrane_zo.rst
MarcusHolly: This will be another easy fix in the Excel file. The cost function was just changed from FALSE to cost_power_law_flow.

docs/technical_reference/unit_models/zero_order_unit_models/automate_rst_file_wt3.py
adam-a-a: The suggested change looks nice, but at least on my end, it looks like the suggested change wasn't committed. Did it work out when/if you tried it?

MarcusHolly: I didn't hit the "commit suggestion" button. I just typed it in manually (because I didn't want the comments in there). When I look in "Files Changed", this change has been made.

adam-a-a: Got it- it was just a matter of refreshing to the latest on GH for me. Thanks!

docs/technical_reference/unit_models/zero_order_unit_models/automate_rst_file_wt3.py
lbianchi-lbl: One thing I should mention (speaking from experience from a few different projects) is that setting up this type of test (i.e. spanning code and documentation) can be quite tricky. I'm happy to discuss more in detail, but in general, we'd want to make sure this test works reliably before adding it to the test suite. Depending on the planned timeline for this feature, it might make sense to consider introducing the test in a separate PR.

bknueven: I think this PR does plenty as-is. Let's leave the test for a follow-up, where we can validate it extensively and separately both locally and on GHA.

docs/technical_reference/unit_models/zero_order_unit_models/constructed_wetlands_zo.rst
```diff
@@ -8,8 +8,6 @@ See documentation for :ref:`single-input, single-output Helper Methods<siso_meth
 
 Electricity Consumption
 -----------------------
-This unit does not include energy consumption.
-
```
bknueven, line 12: Should these lines be removed? If so, I think the preceding two lines should also be removed.

bknueven (commented):
Looks good -- I have one minor comment.

I think we should merge this more-or-less as it is now and save testing the existence of documentation as a follow-up.

docs/technical_reference/unit_models/zero_order_unit_models/automate_rst_file_wt3.py
MarcusHolly: I agree that it may be best to leave the test for a follow-up PR, but I have a few questions about making the test file:

1.) Is there a place I can look to in order to get a better understanding of what all the pytests marks do?

2.) As of right now, the test file I made pretty much just rewrites the check that was originally in the automate_rst_file_wt3.py file, but it seems like it should be a bit more complicated than that? Should it also test that all of the expected entries are present in each list as opposed to just testing that these two lists are the same (similar to what has been done in https://github.com/watertap-org/watertap/blob/main/watertap/data/techno_economic/tests/test_unit_parameter_files.py)? I'm not sure what else there is to test...

docs/technical_reference/unit_models/zero_order_unit_models/constructed_wetlands_zo.rst
adam-a-a: I think for now we can just keep the section mentioning exclusion of energy consumption. Otherwise, we'd want to change the automation script to not write this section at all if there is no energy consumption to consider. Perhaps that could go in the follow-up that you mentioned.

docs/technical_reference/unit_models/zero_order_unit_models/automate_rst_file_wt3.py
adam-a-a: @bknueven @lbianchi-lbl might have some additional comments regarding this, but you could check out the IDAES wiki for more info on testing: https://github.com/IDAES/idaes-pse/wiki/Testing-Standards

I agree that pushing the test to a follow-up PR would be fine. Then, we could think about any additional complexities that we might want to account for in that test.

docs/technical_reference/unit_models/zero_order_unit_models/constructed_wetlands_zo.rst
bknueven: I would also rather keep these lines, FWIW. We just shouldn't remove them *and* keep the header.

If need, I'm happy to look at the documentation generation code and offer suggestion if this isn't already automated.

MarcusHolly: @adam-a-a I actually didn't change anything about constructed wetlands, so I think it just might not have been updated b4. But essentially this is happening because the entry for column G of the excel (energy) is just a "--", which is also what we have for the feed_zo. The latest commit should remove these 2 lines and the header. Everything should be up-to-date now!

MarcusHolly: Nvm, looks like changing the automation script is causing other issues... I'll just revert the last change.

bknueven:
It seems like there should be some moderate restructuring of the writer code to better support non-existent sections. Right now it's handled in a case-by-case as opposed to automated fashion.

bknueven (approved):

tarnold17 (approved):

PR #805, opened 2022-10-10
Move examples out of package directory

wigging:
## Fixes/Resolves:

Move the examples out of the `watertap` package to the top level of the repository since the example files are not needed for installation. This will decrease the installation size of the `watertap` package.

## Summary/Motivation:

When WaterTAP is installed with `pip install watertap`, the examples also get installed with the package. This adds about 3.5 MB to the `watertap` package. As you can see in my Conda environment, an examples directory is located inside the package at:

```
/opt/miniconda3/envs/watertap2/lib/python3.10/site-packages/watertap/examples
```

This occurs because the examples in the GitHub repository are located inside the package at `watertap/examples`. This directory should be moved out of the package directory and to the top level of the repository. This prevents the examples from getting installed with the package. The `watertap` package itself does not provide any mechanisms to directly run the examples so there's no need to install them with the package.

Also, the `tutorials` directory is moved into the examples folder so that all example files are in one location within the repository. The overall project structure is now:

```
watertap/
|-- docs/
|-- examples/
|    |-- chemistry/
|    |-- custom_model_demo/
|    |-- edb/
|    |-- flowsheets/
|    |-- tutorials/
|-- watertap/
|-- setup.py
```

- Move `watertap/watertap/examples` to the top level of the repository `watertap/examples`
- Update tests accordingly

### Legal Acknowledgement


lbianchi-lbl:
@wigging thanks for taking the time for proposing this. For context, the fact that the WaterTAP examples are part of the `watertap` Python package distribution is actually on purpose.

There are multiple reasons for this, but I'd say that the main one is being able to run and test the examples consistently across multiple environment and/or installation variants.

I do partially agree with the point being raised on "the `watertap` package itself [not providing] any mechanism to run the examples"... but I would say that this is only true if one excludes pytest, since `pytest --pyargs watertap.examples` will result in all the examples code that is exercised by the test suite to be run in any environment where `watertap` is available.

If the examples are not part of a Python package distribution, e.g. as a result of being moved to the root of this repository, we'd lose several pieces of functionality, including the ability to reliably import code defined in the `examples` modules (since it's not part of a package, we'd have to manually manage the parent directory of `examples` so that it's added to `sys.path`), as well as running the examples test suite after installing WaterTAP with a regular (i.e. `site-packages`) installation, which I would consider a key component of the CI and release pipeline.

If we goal is to make the codebase more modular and make the code in `examples` distributable independently from the rest of the `watertap` package (which is, as I understand, the key motivation for this PR), *and* be able to keep the capabilities described above, we could look into moving the examples into a separate repository, and create a separate Python package distribution so that one can run `pip install watertap-examples`. However, this approach comes with several tradeoffs (that, incidentally, the team is quite intimately familiar with, as it is the structure chosen for IDAES), which I think should be carefully considered in terms of both short- and long-term maintainability.

Hopefully this clarifies some of the motivations for the current repository structure and packaging choices (which admittedly are rarely if ever documented in a single, easy-to-access format), but please do not hesitate to reach out for any further question or clarification.

wigging:
@lbianchi-lbl Popular Python packages such as [NumPy](https://github.com/numpy/numpy), [SciPy](https://github.com/scipy/scipy), [Pandas](https://github.com/pandas-dev/pandas), [Matplotlib](https://github.com/matplotlib/matplotlib), [Seaborn](https://github.com/mwaskom/seaborn), [Django](https://github.com/django/django), [Flask](https://github.com/pallets/flask), and several others do not put the examples directory inside the package. These packages put examples in a separate GitHub repository or in a top level project directory. Examples are not part of the package's source code. If there is code in the WaterTAP examples that the watertap package relies on, then that code should not be in an "examples" directory or file. That code should be part of the package's source code. Can you show me a respectable Python package on GitHub that puts the examples inside the package directory?

wigging:
@lbianchi-lbl Also, you mentioned that the examples would need to be installed via `pip install watertap-examples` if they are moved to a separate repository.  Why would you need to pip install the examples? Once watertap is installed on the user's computer, there should be no additional steps needed to run the examples.

lbianchi-lbl:
From these latest comments, I suspect this might be a case of [Chesterton's fence](https://en.wiktionary.org/wiki/Chesterton%27s_fence), where (paraphrasing) most of the arguments in favor of a change come from an incomplete understanding of the existing situation.

In rough order from more concrete/technical to more general:

- If the main motivation for this PR is preventing the examples from being distributed together with the rest of the WaterTAP codebase, then the proposed implementation (as of commit c7caf1b) does not obtain this goal
  - Moving the `examples` directory to the root of this repository results in the examples being outside of the `watertap` *package*, but still inside the `watertap` *package distribution*
  - This can be tested by running `pip install .` from this PR's head branch in a fresh environment, which should result in both `watertap` and `examples` being copied in the environment's `site-packages` directory
  - Therefore, the only net effect would be to move the WaterTAP examples to a non-namespaced (and arguably ambiguously named) `examples` top-level package
- As one can see from the amounts of tests that are currently failing (and I suspect even more would occur if the current failures are addressed), being able to reliably access the examples code through the standard Python module import mechanism is an essential capability for the WaterTAP codebase
  - From my limited hands-on experience, off the top of my head I can think of the flowsheets being used for the UI, which therefore are directly tied to project deliverables
- The semantics of "examples" (i.e. what it means for a piece of code to be an "example" within a project's codebase), like many other concepts, is not universally defined, and is instead heavily dependent on context (both for the entire project, and within the project)
  - In other words, "examples" in a general-purpose, foundational project like Numpy, Pandas, Flask, and all the other examples mentioned can be very different from "examples" in a domain-specific, specialized project such as IDAES or, to an even greater degree, WaterTAP, even though the same word is used for both
  - More concretely, "examples" in IDAES/WaterTAP can comprise code that is both usable on its own (i.e. the user can copy that code and modify it to suit their use case), but also being used as-is to define more complex systems. One can of course wonder whether these "base" examples should be handled differently than the "final" examples, but the concrete requirement remains: the examples infrastructure must be able to accomodate all types of examples, as needed by the project
  - This is also the reason why I mentioned the case of the IDAES examples, not only as a concrete use case that many on the WaterTAP development team have experience with, but also as being much closer to WaterTAP's project structure and development workflow than any of the other Python projects mentioned, which I think should be taken into account when evaluating possible choices (and, frankly, seems IMO more informative than "respectability" as a decision criterion)

Zooming out a bit, as I mentioned in my previous comment, I'm not in principle opposed to the idea of the examples being in a separate repository. My opinion on this is that it's something we could consider, but would require a fair amount of work to set up, and might not be the most efficient choice in terms of continued maintenance. More in general, I think it's important to keep in mind that many decisions that might seem straightforward and/or purely technical are actually motivated by project management and long-term sustainability considerations. This is even more relevant for the specific case of WaterTAP, given its unique structure and the way it is positioned in the wider Pyomo/IDAES ecosytem.

---

In summary:

- Moving the examples from their current location in `watertap.examples` is something we *in principle* might want to consider
- As the examples are critical to several parts of the WaterTAP codebase, this migration would have a significant impact on the project, and should therefore be evaluated accordingly
- Specifically for this PR:
  - Given the above point, the upsides that the proposed changes would bring should be formulated more clearly, especially in terms of value-added for the project's deliverables and long-term maintenance
  - The current implementation has significant downsides that would need to be addressed thoroughly
- Given all of the above, I recommend moving any further discussion to a more general forum (such as the weekly dev call)

wigging:
I started discussion #809 about moving the examples directory.

@lbianchi-lbl Can you delete your comments here and move them to the discussion? This will keep everything in the same place and clean up the pull request comments.

ksbeattie:
No yet.

PR #806, opened 2022-10-11
Differential Parameter Sweep Sampling

k1nshuk:
## Fixes/Resolves:

Added a new Parameter Sweep Class to enable differential sampling using the existing sampling types.

## Summary/Motivation:

Enable better analysis.

- Creation of a new parameter sweep class `DifferentialParameterSweep`.
- Bug fixes in parameter sweep writer.

### Legal Acknowledgement


watertap/tools/parameter_sweep/differential_parameter_sweep.py
```diff
@@ -0,0 +1,390 @@
...
+        #                  "relative_ub" : For linear/geometric sampling,
+        #                  "relative_std_dev" : For normal sampling
+        #                 },
+        #     "B_comp" : {},
+        #     "Spacer_porosity"  : {},
+        # }
+
+        diff_sweep_param = {}
+        ctr = 0
+        for param, specs in differential_sweep_specs.items():
```
bknueven:
```suggestion
        for ctr, (param, specs) in enumerate(differential_sweep_specs.items()):
```
Let's use `enumerate` when we need an index

watertap/tools/parameter_sweep/differential_parameter_sweep.py
```diff
@@ -0,0 +1,390 @@
...
+                # Solve status
+                if key == "solve_successful":
+                    local_output_dict["solve_successful"].extend(item)
+                else:
+                    for subkey, subitem in item.items():
+                        local_output_dict[key][subkey]["value"] = np.concatenate(
+                            (local_output_dict[key][subkey]["value"], subitem["value"])
+                        )
+
+    def _create_global_output(self, local_output_dict):
```
bknueven: You should tweak the base class method instead of copy+pasting almost all its code.

`DifferentialParameterSweep` could pass in `req_num_samples=None` and a couple of checks against `None` in the base-case `_create_global_output` will enable the elimination of this method.

watertap/tools/parameter_sweep/differential_parameter_sweep.py
```diff
@@ -0,0 +1,390 @@
...
+                    lb = nominal_val * relative_lb
+                    ub = nominal_val * relative_ub
+                else:
+                    raise NotImplementedError
+                if specs["diff_sample_type"] == UniformSample:
+                    diff_sweep_param[param] = UniformSample(pyomo_object, lb, ub)
+                else:
+                    diff_sweep_param[param] = specs["diff_sample_type"](
+                        pyomo_object, lb, ub, self.config.num_diff_samples
+                    )
```
bknueven: The logic of this is a bit lost on me.

Don't `UniformSample` and `NormalSample` still need a number of samples parameter? Why can't `UniformSample` just use the logic on the last line?

watertap/tools/parameter_sweep/differential_parameter_sweep.py
```diff
@@ -0,0 +1,390 @@
...
+                std_dev = specs["relative_std_dev"]
+                diff_sweep_param[param] = NormalSample(
+                    pyomo_object, nominal_val, std_dev
+                )
+            else:
+                relative_lb = specs["relative_lb"]
+                relative_ub = specs["relative_ub"]
+                if specs["diff_mode"] == "sum":
+                    lb = nominal_val * (1 - relative_lb)
+                    ub = nominal_val * (1 + relative_lb)
```
bknueven:
```suggestion
                    ub = nominal_val * (1 + relative_ub)
```
copy+paste error

watertap/tools/parameter_sweep/differential_parameter_sweep.py
```diff
@@ -0,0 +1,390 @@
...
+        #     "Spacer_porosity"  : {},
+        # }
+
+        diff_sweep_param = {}
+        ctr = 0
+        for param, specs in differential_sweep_specs.items():
+            nominal_val = local_values[ctr]
+            pyomo_object = specs["pyomo_object"]
+            if specs["diff_sample_type"] == NormalSample:
+                std_dev = specs["relative_std_dev"]
```
bknueven: How is this the "relative" standard deviation?

watertap/tools/parameter_sweep/differential_parameter_sweep.py
bknueven: Isn't this more properly "percent"? "sum" might be:
```python
lb = nominal_val + relative_lb
ub = nominal_val + relative_ub
```

In this case, we might validate that `relative_lb < 0 < relative_ub`. (And for `"product"` that `relative_lb < 1 < relative_ub`.)

watertap/tools/parameter_sweep/differential_parameter_sweep.py
```diff
@@ -0,0 +1,390 @@
...
+            self._update_local_output_dict(
+                model,
+                sweep_params,
+                k,
+                local_values[k, :],
+                run_successful,
+                local_output_dict,
+            )
+
+            local_solve_successful_list.append(run_successful)
```
bknueven: This code is identical to the base class.

This case is a bit more complicated, but perhaps the easiest thing to do is have the base class `_do_param_sweep` call an optional method or callback inside its `for` loop with this class implements for "Step 2"

bknueven (commented):
I left a few review comments.

watertap/tools/parameter_sweep/differential_parameter_sweep.py
k1nshuk: NormalSample and UniformSample were created initially as multi-use objects. Somewhere, I accidentally introduced a different class constructor for the other sampling types. I was planning to fix that in another PR, but might as well fix that here. I'll make the others consistent with NormalSample and UniformSample.

watertap/tools/parameter_sweep/differential_parameter_sweep.py
k1nshuk: renamed `relative_std_dev` to just `std_dev` in commit de7822636e8fc41f91ef9371185d763c60d1f48a

watertap/tools/parameter_sweep/differential_parameter_sweep.py
k1nshuk: Yes, it is effectively a percentage I am open to using a different keyword that is more accurate. We would have to scale the values by 100 if we were to use the word percent.

I am not convinced that we need the above validation for the upper and lower bounds. Such a validation might be beneficial if we were using the same operator for computing the upper and lower bounds.

watertap/tools/parameter_sweep/differential_parameter_sweep.py
k1nshuk: Repurposed the base class to be able to handle this.

watertap/tools/parameter_sweep/differential_parameter_sweep.py
k1nshuk: Repurposed the base class method to handle this. Although, not sure if we should add assertions to catch incorrect keyword argument specifications for `ParameterSweep` and `RecursiveParameterSweep` for the config option `run_differential_sweep`. This may require us to write `__init__` functions for the two as, ideally, we would like to catch that when initializing the class. That being said, we could also check for that assertion within the `parameter_sweep` method.

watertap/tools/parameter_sweep/differential_parameter_sweep.py
k1nshuk: Updated all sampling types to have the same constructor where they accept num_samples in their argument. I'll change the documentation and move this from draft to a regular PR.

bknueven (approved):
Worked through a few things in person, now looks good.

avdudchenko (approved):
This looks good to me (and tested).

PR #807, opened 2022-10-11
Making Flowsheets More User-Friendly

### LINKED ISSUE #621: Making AMO flowsheets and multi-sweep files more user friendly
The following changes can be made in most, if not all, AMO flowsheets:
- [ ] Instead of `assert_optimal_termination` in main(), consider replacing with `check_optimal_termination` and more helpful error message for user
- [ ] Add more user-friendly usage notes for running multi-sweep from command line. Example:
https://github.com/watertap-org/watertap/blob/92f6ab39ef3cd4a350f31c4edcee0656831849f1/watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/suboxic_activated_sludge_process/multi_sweep.py#L83-L92

MarcusHolly:
## Fixes/Resolves:

Issue https://github.com/watertap-org/watertap/issues/621

- Replaces all instances of `assert_optimal_condition` with `check_optimal_condition` found under the flowsheet directory
- I could also tackle the 2nd point of the above issue with a little bit of guidance as to what needs to be added and where

### Legal Acknowledgement


watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1575_magprex/magprex.py
```diff
@@ -257,22 +257,6 @@ def add_costing(m):
...
-def solve(blk, solver=None, tee=False, check_termination=True):
-    if solver is None:
-        solver = get_solver()
-    results = solver.solve(blk, tee=tee)
-    if check_termination:
-        assert_optimal_termination(results)
-    return results
-
-
 def display_results(m):
```
tarnold17: In addition to `initialize_system` and `solve`, it looks like there also is a duplicate instance of `display_results`, so I recommend removing this one.

tarnold17 (approved):
Looks good, just one minor request.

bknueven (commented):
I know there's an issue for this -- but if this is really the behavior we want, then we should just drop the `assert_optimal_termination` lines.

`check_optimal_termination` returns a `bool` indicating success/failure, but if we do nothing with the return value it's essentially a no-op. So, what we're really saying here is that it's up to the calling function to check the returned results object.

I'm also not sure just eliminating the `assert_optimal_termination` makes the code more user-friendly. When the solver fails, if the calling code *does not* check the returned `results` object, then there will be (likely) a nonsensical answer stored in the Pyomo model. This is worse than crashing, because then there's been a failure but the code proceeded as-if nothing is wrong.

bknueven (changes requested):
To build on my earlier comment, the other half of Adam's ask in #621 was "a more friendly error message to the user". We should still stop execution, but with a standard error message about solver failure.

The naked assert that `assert_optimal_termination` does makes no mention of what went wrong.

MarcusHolly:
> To build on my earlier comment, the other half of Adam's ask in #621 was "a more friendly error message to the user". We should still stop execution, but with a standard error message about solver failure.
> 
> The naked assert that `assert_optimal_termination` does makes no mention of what went wrong.

You bring up some good points, so it sounds like we may need to discuss what the most effective course of action is in a future meeting. As for the "more friendly error message to the user", is this just going to be a comment/print statement in the code or something more involved? Either way, the content of the error message can also be discussed in a future meeting.

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
```diff
@@ -334,7 +338,7 @@ def calculate_operating_pressure(
     t.brine[0].pressure_osm_phase
     # solve state block
     results = solve_indexed_blocks(solver, [t.brine])
-    assert_optimal_termination(results)
+    check_optimal_termination(results)
```
bknueven: My one comment is that I'm still not a huge fan of these `check_optimal_terminations` where we do nothing with the result.

This is a no-op, so it might confuse someone reading this code as to why it's here in the first place.

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
```diff
@@ -344,7 +348,12 @@ def solve(blk, solver=None, tee=False, check_termination=True):
         solver = get_solver()
     results = solver.solve(blk, tee=tee)
     if check_termination:
-        assert_optimal_termination(results)
+        if not check_optimal_termination(results):
+            _log.warning(
+                "The solver failed to converge to an optimal solution."
+                "This suggests that the user provided infeasible inputs or that the model "
+                "is poorly scaled, poorly initialized, or degenerate. "
+            )
```
bknueven: I would think in cases like these we should get rid of the `check_termination` flag. It doesn't really serve a purpose now.

MarcusHolly: I get that the check_optimal_termination is a no-op, but isn't it still useful for the user to see the associated warning for when the solver fails. So would there be a way to display this error message w/o utilizing check/assert optimal termination?

andrewlee94: In this case, they wouldn't see it anyway, as the `assert_optimal_termination` would have failed first and thrown an `AssertionError`. So, it is not just a no-op, but completely redundant.

bknueven: This particular instance *isn't* a no-op though -- we take some action (sending a message to the logger) if `check_optimal_termination` returns `False`.

My comment here is we don't need the `check_termination` bool on line 350, or in this function. We should just always log the warning.

Perhaps we should write a helper function which will check the optimal termination and log a warning. It would be best if it could provide some information about where it was called though. Hmm.

MarcusHolly: I added a simple utility function that uses the check_optimal_termination. I'm not sure what adding "some information about where it was called" would look like, but let me know if I should expand on the utility function in some way.

watertap/core/util/optimal_termination.py
```diff
@@ -0,0 +1,27 @@
...
+# Laboratory, National Renewable Energy Laboratory, and National Energy
+# Technology Laboratory (subject to receipt of any required approvals from
+# the U.S. Dept. of Energy). All rights reserved.
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and license
+# information, respectively. These files are also available online at the URL
+# "https://github.com/watertap-org/watertap/"
+#
+###############################################################################
+
```
adam-a-a: I'd suggest renaming this utility module to something like more general than the function, but I suppose that can come about whenever we add more functions.

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
```diff
@@ -334,7 +339,6 @@ def calculate_operating_pressure(
     t.brine[0].pressure_osm_phase
     # solve state block
     results = solve_indexed_blocks(solver, [t.brine])
-    assert_optimal_termination(results)
```
adam-a-a: I'd think you'd want to keep this here in _this_ case since the purpose is to verify that this function returns a valid solution.

watertap/core/util/optimal_termination.py
```diff
@@ -0,0 +1,27 @@
...
+from pyomo.environ import check_optimal_termination
+import idaes.logger as idaeslog
+
+# Set up logger
+_log = idaeslog.getLogger(__name__)
+
+
+def optimal_termination(self):
+    if check_optimal_termination(self):
+        _log.warning(
```
adam-a-a: We might want to have this give either (1) a warning or (2) raise an exception while providing the same or similar message that you have here. It could be easy to miss the warning in some cases, so providing the option to raise an error in `optimal_termination` might make sense. That is, we should allow for _both_.

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
```diff
@@ -343,8 +347,7 @@ def solve(blk, solver=None, tee=False, check_termination=True):
     if solver is None:
         solver = get_solver()
     results = solver.solve(blk, tee=tee)
-    if check_termination:
-        assert_optimal_termination(results)
+    optimal_termination(results)
```
adam-a-a: If you add an argument to this function as suggested earlier, then you could set that argument to `True`, say, to raise an exception. 

For example:
`optimal_termination(results, raise_exception=check_termination)`

The additional argument is `raise_exception`, and if someone set `check_termination` True, it raises the error message and halts; otherwise, your message is logged as a warning when `check_termination` is False.

watertap/examples/flowsheets/case_studies/activated_sludge/ASM1_flowsheet.py
```diff
@@ -292,8 +297,7 @@ def function(unit):
     # Solve overall flowsheet to close recycle loop
     solver = get_solver(options={"bound_push": 1e-8})
     results = solver.solve(m)
-
-    pyo.assert_optimal_termination(results)
+    optimal_termination(results)
```
adam-a-a: Harping on about the warning or exception, I think raising an exception should be the default, while enabling the message to be logged as a warning would have to be requested by the user.

watertap/examples/flowsheets/case_studies/activated_sludge/ASM1_flowsheet.py
```diff
@@ -303,7 +307,6 @@ def function(unit):
 
     # Resolve with controls in place
     results = solver.solve(m, tee=True)
-    pyo.assert_optimal_termination(results)
```
adam-a-a, line 306: We don't want to be rid of this. Actually, the original issue sprouted from either @andrewlee94 and/or @bknueven 's feedback one of my AMO PRs that comprised a flowsheet. In the main function, we were just asserting optimal termination, and one or both of the gentlemen suggested something more friendly- i.e., explicitly telling the user that the solver failed to converge to a solution because of x, y, z. 

Here, we want to make sure the solution is not nonsensical because of a failure to converge and might not necessarily need to relay a user-friendly message (but we could, using a modified version of your new utility). Long story short-- we should not be completely removing assert_optimal_termination lines. We want some sort of check on the solution after every solve.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1575_magprex/magprex.py
```diff
@@ -55,16 +59,13 @@ def main():
     initialize_system(m)
 
     results = solve(m)
-    assert_optimal_termination(results)
```
adam-a-a, line 58: I'd think you'd want your new function here too. In these flowsheets, we were doing this to isolate failures, which can occur after any solve. Now that I'm looking at this, we might want another argument in optimal_termination that takes a string describing where the check (and failure) occurred.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1575_magprex/magprex.py
```diff
@@ -55,16 +59,13 @@ def main():
...
     display_results(m)
 
     add_costing(m)
     m.fs.costing.initialize()
 
     assert_degrees_of_freedom(m, 0)
     assert_units_consistent(m)
-    results = solve(m)
-    assert_optimal_termination(results)
```
adam-a-a, line 67: same as above

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1575_magprex/magprex.py
```diff
@@ -150,8 +151,7 @@ def solve(blk, solver=None, tee=False, check_termination=True):
     if solver is None:
         solver = get_solver()
     results = solver.solve(blk, tee=tee)
-    if check_termination:
-        assert_optimal_termination(results)
+    optimal_termination(results)
```
adam-a-a: same suggestion from before where check_termination was an argument

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1575_magprex/magprex.py
```diff
@@ -257,28 +257,6 @@ def add_costing(m):
...
-    results = solver.solve(blk, tee=tee)
-    if check_termination:
-        assert_optimal_termination(results)
-    return results
-
-
-def display_results(m):
-    unit_list = ["feed", "magprex", "centrifuge", "classifier"]
-    for u in unit_list:
-        m.fs.component(u).report()
```
adam-a-a: Why'd you remove this?

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1595_photothermal_membrane_candoP/amo_1595.py
```diff
@@ -166,8 +169,7 @@ def solve(blk, solver=None, tee=False, check_termination=True):
     if solver is None:
         solver = get_solver()
     results = solver.solve(blk, tee=tee)
-    if check_termination:
-        assert_optimal_termination(results)
```
adam-a-a, line 170: Same comments as before

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1690/amo_1690.py
```diff
@@ -162,8 +165,7 @@ def solve(blk, solver=None, tee=False, check_termination=True):
     if solver is None:
         solver = get_solver()
     results = solver.solve(blk, tee=tee)
-    if check_termination:
-        assert_optimal_termination(results)
+    optimal_termination(results)
```
adam-a-a: Same comments as before, applicable to all the AMO flowsheet changes

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack.py
```diff
@@ -224,8 +229,7 @@ def solve(blk, solver=None, tee=False, check_termination=True):
     if solver is None:
         solver = get_solver()
```
adam-a-a, line 229: Same feedback as earlier applies here

adam-a-a (changes requested):
My main suggestions are to add one or two arguments to the `optimal_termination` function:
1) an arg for the user to set whether to raise exception or simply log warning, with raising an exception being the default.
2) a string arg for the user to specify location in the code that is being checked, in case there are multiple solves and associated checks and the user would want to be able to easily determine where the problem is popping up.

I also had some other comments, including always checking termination after a solve, somehow, some way.

adam-a-a:
@MarcusHolly -Actually, I was just checking our `initialization` utility module since I recalled an example to point you to. It ends up (at first glance) that you may be able to just use `check_solve` from that utility, perhaps with some modification as needed to the error/warning message. Have a look at it and tell me what you think: https://github.com/watertap-org/watertap/blame/main/watertap/core/util/initialization.py#L29-L67

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1575_magprex/magprex.py
MarcusHolly: @adam-a-a These 3 functions were all accidentally duplicated (they were each present twice), so I just cut out the duplicate. I'll look more closely into all your comments later this afternoon, but I agree (at first glance) that the check_solve seems to have everything we are looking for. So I will probably just end up deleting the check_termination function I made.

watertap/core/util/initialization.py
```diff
@@ -46,19 +46,29 @@ def check_solve(results, checkpoint=None, logger=_log, fail_flag=False):
     """
     if check_optimal_termination(results):
         if checkpoint is None:
-            logger.info(f"Solve successful.")
+            logger.info(
+                "The solver failed to converge to an optimal solution."
+                "This suggests that the user provided infeasible inputs or that the model "
+                "is poorly scaled, poorly initialized, or degenerate."
```
adam-a-a: The logger message should indicate a successful solve occurred but you're replacing it with a message indicating a failure to converge.

watertap/core/util/initialization.py
```diff
@@ -46,19 +46,29 @@ def check_solve(results, checkpoint=None, logger=_log, fail_flag=False):
...
+            )
         else:
-            logger.info(f"{checkpoint} successful.")
+            logger.info(
+                f" The solver at the {checkpoint} step failed to converge to an optimal solution."
+                "This suggests that the user provided infeasible inputs or that the model "
+                "is poorly scaled, poorly initialized, or degenerate."
```
adam-a-a: Same goes here- this should indicate a successful solve

watertap/core/util/initialization.py
```diff
@@ -46,19 +46,29 @@ def check_solve(results, checkpoint=None, logger=_log, fail_flag=False):
...
+            )
     else:
         if checkpoint is None:
             msg = (
                 f"The solver failed to converge to an optimal solution. "
-                f"This suggests that the user provided infeasible inputs or that the model is poorly scaled."
+                "This suggests that the user provided infeasible inputs or that the model "
+                "is poorly scaled, poorly initialized, or degenerate."
```
adam-a-a, line 57: This is the correct place to put your refined message.

watertap/examples/flowsheets/case_studies/activated_sludge/ASM2D_flowsheet.py
```diff
@@ -409,8 +413,7 @@ def function(unit):
 
     # Resolve with controls in place
     results = solver.solve(m, tee=False)
-
-    pyo.assert_optimal_termination(results)
+    check_solve(results, checkpoint="controls in place", logger=_log, fail_flag=True)
```
adam-a-a:
```suggestion
    check_solve(results, checkpoint="resolve with controls in place", logger=_log, fail_flag=True)
```

watertap/examples/flowsheets/case_studies/municipal_treatment/municipal_treatment.py
```diff
@@ -146,7 +149,7 @@ def set_operating_conditions(m):
     m.fs.feed.conc_mass_comp[0, "tds"].fix(conc_mass_tds)
     m.fs.feed.conc_mass_comp[0, "tss"].fix(conc_mass_tss)
     m.fs.feed.conc_mass_comp[0, "toc"].fix(conc_mass_toc)
-    solve(m.fs.feed)
+    solve(m.fs.feed, checkpoint="set operating conditions")
```
adam-a-a:
```suggestion
    solve(m.fs.feed, checkpoint="solve feed block")
```

watertap/examples/flowsheets/case_studies/seawater_RO_desalination/seawater_RO_desalination.py
```diff
@@ -79,12 +81,12 @@ def solve_flowsheet(flowsheet=None):
     m = flowsheet.parent_block()  # UI block is 'm.fs' but funcs below use 'm'
     initialize_system(m)
     assert_degrees_of_freedom(m, 0)
-    solve(m)
+    solve(m, checkpoint="initialize system")
```
adam-a-a:
```suggestion
    solve(m, checkpoint="solve flowsheet after initializing system")
```

watertap/examples/flowsheets/case_studies/seawater_RO_desalination/seawater_RO_desalination.py
```diff
@@ -79,12 +81,12 @@ def solve_flowsheet(flowsheet=None):
...
     display_results(m)
     add_costing(m)
     initialize_costing(m)
     assert_degrees_of_freedom(m, 0)
-    solve(m)
+    solve(m, checkpoint="solve flowsheet")
```
adam-a-a:
```suggestion
    solve(m, checkpoint="solve flowsheet with costing")
```

watertap/examples/flowsheets/case_studies/seawater_RO_desalination/seawater_RO_desalination.py
```diff
@@ -97,14 +99,14 @@ def main(erd_type="pressure_exchanger"):
     initialize_system(m)
     assert_degrees_of_freedom(m, 0)
 
-    solve(m)
+    solve(m, checkpoint=f" initialize {erd_type} system")
```
adam-a-a:
```suggestion
    solve(m, checkpoint=f" solve flowsheet after initializing {erd_type} system")
```

watertap/examples/flowsheets/case_studies/seawater_RO_desalination/seawater_RO_desalination.py
```diff
@@ -97,14 +99,14 @@ def main(erd_type="pressure_exchanger"):
...
     display_results(m)
 
     add_costing(m)
     initialize_costing(m)
     assert_degrees_of_freedom(m, 0)
 
-    solve(m, tee=True)
+    solve(m, tee=True, checkpoint=f" solve {erd_type} flowsheet")
```
adam-a-a:
```suggestion
    solve(m, tee=True, checkpoint=f" solve {erd_type} flowsheet with costing")
```

watertap/examples/flowsheets/case_studies/seawater_RO_desalination/seawater_RO_desalination.py
```diff
@@ -368,7 +370,7 @@ def set_operating_conditions(m):
     m.fs.feed.flow_vol[0].fix(flow_vol)
     m.fs.feed.conc_mass_comp[0, "tds"].fix(conc_mass_tds)
     m.fs.feed.conc_mass_comp[0, "tss"].fix(conc_mass_tss)
-    solve(m.fs.feed)
+    solve(m.fs.feed, checkpoint="set operating conditions")
```
adam-a-a:
```suggestion
    solve(m.fs.feed, checkpoint="solve feed block")
```

adam-a-a:
@MarcusHolly I added a few suggestions for the messages to be displayed but did not apply them throughout all files where relevant. Please consider them and applying similar revisions to files that I didn't address directly, as needed.

TimBartholomew (approved):

bknueven (approved):
I think there are changes we should make to the `check_solve` function (including its name, potentially), but this PR is an improving direction.

adam-a-a (approved):

PR #808, opened 2022-10-11
Seawater Ion Generic Documentation

### LINKED ISSUE #269: Add documentation for seawater_ion_generic
`seawater_ion_generic` uses IDAES' generic property framework to construct an artificial seawater feed that comprises sodium, chloride, magnesium, calcium, and sulfate ions. Additionally, the config file uses the eNRTL model as the equation of state. Documentation should be added to the technical reference on this.

MarcusHolly:
## Fixes/Resolves:

Issue https://github.com/watertap-org/watertap/issues/269

- Adds documentation for the seawater_ion_generic property package (may be missing some information)

### Legal Acknowledgement


MarcusHolly:
I run into an error when writing "Cl_-" as shown below, so my current workaround is to write it as "Cl_1-"
image

bknueven:
> I run into an error when writing "Cl_-" as shown below, so my current workaround is to write it as "Cl_1-" image

Probably "escaping" the `_` or `-` will fix this, e.g., "Cl\\_-" or "Cl\_\\-"

bknueven:
The error on readthedocs is:
```
checking consistency... /home/docs/checkouts/readthedocs.org/user_builds/watertap/checkouts/808/docs/technical_reference/property_models/seawater_ion_generic.rst: WARNING: document isn't included in any toctree
```

You probably need to add a reference to `seawater_ion_generic` in this file:
https://github.com/watertap-org/watertap/blob/main/docs/technical_reference/property_models/index.rst

docs/technical_reference/property_models/seawater_ion_generic.rst
```diff
@@ -0,0 +1,87 @@
...
+   :header: "Description", "Equation"
+
+   "Component mole flowrate", ":math:`N_j = \frac{M_j}{MW_j}`"
+   "Component mole fraction", ":math:`y_j = \frac{N_j}{\sum_{j} N_j}`"
+
+
+
+Scaling
+-------
+This artificial seawater property package includes support for scaling, such as providing default or calculating scaling factors for all variables.
```
luohezhiming:
```suggestion
This artificial seawater property package includes support for scaling, such as default or user-supplied scaling factors for all variables.
```

docs/technical_reference/property_models/seawater_ion_generic.rst
```diff
@@ -0,0 +1,87 @@
...
+
+
+Relationships
+-------------
+.. csv-table::
+   :header: "Description", "Equation"
+
+   "Component mole flowrate", ":math:`N_j = \frac{M_j}{MW_j}`"
+   "Component mole fraction", ":math:`y_j = \frac{N_j}{\sum_{j} N_j}`"
+
```
luohezhiming, line 51: Probably add a note here to say `W_j` is molar mass of component j.

luohezhiming (commented):
LGTM. I only have some minor suggestions

docs/technical_reference/property_models/seawater_ion_generic.rst
MarcusHolly: I don't think this is necessary because none of the other documentation have a note like this, but perhaps this is something we should add to all the documentation?

luohezhiming: If no property package has this, it is OK not inlcuding the note.

luohezhiming (approved):

docs/technical_reference/property_models/seawater_ion_generic.rst
```diff
@@ -0,0 +1,87 @@
+Seawater Ion Generic Property Package
+=====================================
+
+This package implements property relationships for artificial seawater as provided in `Islam et al. (2021) <https://www.sciencedirect.com/science/article/abs/pii/S1383586621009412>`_.
```
adam-a-a, line 4: The main purpose of this model was to make use of IDAES' eNRTL model, which should be emphasized here. We then used the eNRTL model to predict mineral scaling in full treatment where RO was the primary desalination technology.

Another detail to mention: the formulation of this model is based what was formerly referred to as IDAES' generic property framework, which I think has now been renamed to something like the modular property framework.

adam-a-a: More specifically, `seawater_ion_generic` is really just a configuration file that gets passed into the GenericParameterBlock (generic property framework I mentioned earlier). You can look at the test for `seawater_ion_generic` to see what I mean. So perhaps clarification should be provided here to explain this. I am also wondering if `seawater_ion_generic` should be in a different folder at this point, but I guess that's a different discussion.

docs/technical_reference/property_models/seawater_ion_generic.rst
```diff
@@ -0,0 +1,87 @@
...
+State variables
+---------------
+.. csv-table::
+   :header: "Description", "Symbol", "Variable", "Index", "Units"
+
+   "Component molar flowrate", ":math:`N_j`", "flow_mol_phase_comp", "[p, j]", ":math:`\text{mole/s}`"
+   "Temperature", ":math:`T`", "temperature", "None", ":math:`\text{K}`"
+   "Pressure", ":math:`P`", "pressure", "None", ":math:`\text{Pa}`"
+
+Properties
```
adam-a-a, line 34: There are probably a bunch of parameters that we want list out here.

MarcusHolly: @adam-a-a Parameters as in more state variables (I think these are the only 3?) or properties? Looking at the property package file, there are a few things I missed, such as vol_mol_pure, vol_mol_liq_comp, and relative_permittivity_liq_comp. There's no metadata list like with the other prop packs, so I wasn't sure exactly what to include here. Should I still include properties like mass flow rate and mass frac even though this prop pack is formulated on a molar basis?

adam-a-a: @MarcusHolly this one is a bit trickier because it is really a configuration file for the GenerericParameterBlock. However, we might want to think about this a little more since this was written a while back and would need to revisit in more detail. One thing I do know is that we'd want to point to the eNRTL parameters (`Liq_tau`) in some fashion, which was the main purpose of the effort.

MarcusHolly: Added a few more properties, but we can revisit this later (after things have died down) to finish fleshing out this documentation.

tarnold17 (approved):

PR #817, opened 2022-10-13
Update idaes-pse prerelease requirement to watertap-org:2.0.0.dev4.watertap.22.10.13

lbianchi-lbl:
## Summary/Motivation:

- See how many things break when using the current IDAES upstream `main`.

- Set IDAES prerelease requirement to `idaes-pse @ https://github.com/watertap-org/idaes-pse/archive/2.0.0.dev4.watertap.22.10.13.zip`

### Legal Acknowledgement


bknueven:
@lbianchi-lbl @dangunter @adam-a-a I think I've resolved nearly all the failures on this PR *except* those related to the EDB. It needs an update similar to that in c47143a, but that seems like it would break the existing structure.

Given that I've never touched that code I don't think I'm the best person to address it.

bknueven:
@adam-a-a @luohezhiming there is one remaining failure in `swine_wwt_ui`. It looks to me like `waste_disposal` was intended to be costed, but an associated `cost_flow` is never called for it.

lbianchi-lbl:
@dangunter @bknueven In a very "monkey see [Perrys], monkey do" way, I've tried to add support for `eqn_type` in the EDB data files in a way that doesn't require changes to the validation schema: 931a2d25. I have no idea if this is the best way (or, indeed, any good at all), as my guiding criteria was "paste that snippet in `component.json` until `pytest -k edb` gives all green".

Let me know if you think this (or similar) approach can be acceptable (or, pragmatically, preferable to the alternatives of figuring out how to change the EDB schema or which EDB tests need to be skipped).

bknueven:
Thanks @lbianchi-lbl -- the changes in 931a2d2 look good to me.

~~I'm going to address the remaining failures in a separate forthcoming PR.~~
The fix was simpler than I thought, so I just added a commit here for it and reverted 6706dad.

bknueven (approved):

lbianchi-lbl:
@bknueven I forgot about the code snippets in the Sphinx docs, that are now failing because of the `Block(default={...})` API still being used. I'll take care of those; in the meantime, feel free to continue iterating on this branch (I see your latest commits about the Swine WWT UI flowsheet).

luohezhiming (approved):

PR #818, opened 2022-10-14
Adds InitializationError to Unit Models

### LINKED ISSUE #714: Ensure InitializationError is included for all unit models
Some unit models are missing an `InitializationError` exception in the case that initialization fails. This should be added anywhere that it is missing.

MarcusHolly:
## Fixes/Resolves:

Issue https://github.com/watertap-org/watertap/issues/714

- Adds a re-solve to the initialize_build function
- Adds an InitializationError error message to the initialize_build function (based on the InitializationError code written in reverse_osmosis_base.py)

### Legal Acknowledgement


watertap/unit_models/boron_removal.py
```diff
@@ -766,13 +772,23 @@ def initialize_build(
         # Solve unit
         with idaeslog.solver_log(solve_log, idaeslog.DEBUG) as slc:
             res = opt.solve(blk, tee=slc.tee)
+            # occasionally it might be worth retrying a solve
+            if not check_optimal_termination(res):
```
andrewlee94: I am not sure if this is a good idea or not - for one if you are seeing this sort of behaviour it tends to indicate a model issue that really should be fixed.

watertap/unit_models/boron_removal.py
```diff
@@ -766,13 +772,23 @@ def initialize_build(
...
+                res = opt.solve(blk, tee=slc.tee)
         init_log.info_high("Initialization Step 2 {}.".format(idaeslog.condition(res)))
 
         # ---------------------------------------------------------------------
         # Release Inlet state
         blk.control_volume.release_state(flags, outlvl + 1)
         init_log.info("Initialization Complete: {}".format(idaeslog.condition(res)))
 
+        if not check_optimal_termination(res):
+            if raise_on_failure:
```
andrewlee94: I am not sure if this is a good idea either - the whole purpose of the InitializationError is to force the user to recognize it and handle it. This makes it a bit too easy to suppress and ignore these,

watertap/unit_models/boron_removal.py
```diff
@@ -766,13 +772,23 @@ def initialize_build(
         # Solve unit
         with idaeslog.solver_log(solve_log, idaeslog.DEBUG) as slc:
             res = opt.solve(blk, tee=slc.tee)
+            # occasionally it might be worth retrying a solve
+            if not check_optimal_termination(res):
+                init_log.warning(
+                    f"Trouble solving unit model {blk.name}, trying one more time"
+                )
+                res = opt.solve(blk, tee=slc.tee)
```
bknueven: If we keep this everywhere, it might be nice to wrap it up in a utility.

bknueven (commented):
I have a minor comment about re-used code.

@andrewlee94 the RO units have these options for the LSRRO flowsheet. Even when the problem is well-posed, finding a reasonable set of initial conditions can be very tricky. The LSRRO flowsheet has a multi-pass initialization routine followed by using the sequential decomposition tool. Typically for that workflow on the initial passes the RO unit models are infeasible, but eventually become feasible as the initialization routine works.

All of the above said, I'm not certain we want to replicate it every unit model.

lbibl:
Just to voice my opinion:  I am with Andrew on questioning the necessity of a second-time solving.  If the solver "failed" in the first time,  it usually indicates an unsatisfactory initialization or fundamental errors in the equation system. Although occasionally a second time of solving provide a solution, which is usually the case that the first solving improved the initial values (for the second solve), this could be better handled by improving the initialization routine in the first place.  For other failure causes, a second solving may not help at all.

andrewlee94:
It is probably also worth commenting on why a second solve may find a solution when the first did not (some times the opposite is true to). This behaviour generally indicates an issue with scaling and/or initial conditions.

If you are seeing this behaviour, it generally indicates that the Jacobian scaling done (either by IPOPT or the IDAES implementation) is resulting in scaling factors large enough that IPOPT cannot confirm an optimal solution at the final step - it gets the right solution (or lose to it), but the scaling is so poor that it cannot meet the required tolerances. When the solver is rerun, the scaling is recalculated using the new values which results in better scaling allowing IPOPT to converge to the correct solution. Thus, you could avoid the need for the second solve by providing better initial scaling of the model, or a better set of initial guesses.

MarcusHolly:
> Just to voice my opinion: I am with Andrew on questioning the necessity of a second-time solving. If the solver "failed" in the first time, it usually indicates an unsatisfactory initialization or fundamental errors in the equation system. Although occasionally a second time of solving provide a solution, which is usually the case that the first solving improved the initial values (for the second solve), this could be better handled by improving the initialization routine in the first place. For other failure causes, a second solving may not help at all.

Alright, I'll get rid of the re-solves then.

watertap/unit_models/crystallizer.py
```diff
@@ -671,12 +670,11 @@ def initialize(
...
-        init_log.info("Initialization Complete: {}".format(idaeslog.condition(res)))
-
-        if not check_optimal_termination(res):
-            if raise_on_failure:
-                raise InitializationError(f"Unit model {blk.name} failed to initialize")
+        try:
+            blk.properties_in.release_state(flags, outlvl=outlvl)
+            init_log.info("Initialization Complete: {}".format(idaeslog.condition(res)))
+        except InitializationError:
+            raise InitializationError(f"Unit model {blk.name} failed to initialize")
```
adam-a-a: I'm not sure that we want the try/except here. I figured the user would wrap their `initialize() `step in a try/except to keep moving if an `InitializationError` were to occur.

The way this is implemented now, I think the `InitalizationError` might actually get skipped.

bknueven: I agree with Adam -- we should be checking the result of the `solve` method call do know whether or not to raise a `InitializationError`.

We should always release the state of the associated properties so that if the `InitializationError` is caught the unit has no more fixed variables than it did before the call to initialize.

andrewlee94: I will echo @bknueven's comment, and emphasize that if possible you should do the check after the release state step.

```
res = solver.solver(model)

release_state(model)

if not check_optimal_termination(model):
    raise InitializationError()
```

This way, the model should be left in its original state, allowing you to try to continue on (if desired). If you don't do this and try to ignore the error and solve the flowsheet, the problem will be over-specified due to the fixed inlet states.

watertap/unit_models/osmotically_assisted_reverse_osmosis_base.py
```diff
@@ -543,11 +543,8 @@ def initialize_build(
 
         init_log.info("Initialization Complete: {}".format(idaeslog.condition(res)))
 
-        if not check_optimal_termination(res):
-            if raise_on_failure:
-                raise InitializationError(
-                    f"Unit model {self.name} failed to initialize"
-                )
+        if not check_optimal_termination(res) and raise_on_failure:
```
bknueven, line 546: I would like to get rid of `raise_on_failure` here, but we need IDAES/idaes-pse#1008 (or something like it).

bknueven (approved):

tarnold17 (approved):
Looks good to me, although I think it would be advisable to have @andrewlee94 and/or @adam-a-a look over this again and approve before merging.

watertap/unit_models/osmotically_assisted_reverse_osmosis_base.py
andrewlee94: Is there a need for the "raise on failure here"?  It looks like this is the final step in the process so what is the case for not raising the error here rather than letting the next level handle it?

andrewlee94 (commented):
Only one comment regarding the need for the "raise on failure" argument in one model.

watertap/unit_models/osmotically_assisted_reverse_osmosis_base.py
bknueven: Yes, it's used in the LSRRO flowheet initialization here: https://github.com/watertap-org/watertap/blob/8fe2e562abd18dc7e16a951e6724039f67f1e911/watertap/examples/flowsheets/lsrro/lsrro.py#L716

In order to (straightforwardly) replace its usage with try/except I'd need the error to be properly handled on the IDAES side as well, hence https://github.com/IDAES/idaes-pse/pull/1008.

Keep in mind that the default for `raise_on_failure` is `False`, so the exception will be raised unless the user specifically request it not be.

andrewlee94: I don't think I am following here - for one, I can't actually see where `osmotically_assisted_reverse_osmosis_base.py` comes into that example.

However, the step to raise the `InitializationError` in that file is the last lines in the `initialize` method so I am not sure why whatever ends up calling that method would not be able to just wrap it in a `try/except`. As far as I can tell, all the steps to revert the model state have already been performed.

bknueven:
@andrewlee94 and I discussed the outstanding issue with the `raise_on_failure` argument and it will be addressed subsequently.

PR #819, opened 2022-10-15
Property Model Constraint Indexing

### LINKED ISSUE #605: More updates for property packages
It was pointed out in #564 that there are still some outstanding issues with standardization of properties. Below is a brief summary of additional things to address:

- [ ] Calculation of Schmidt number (`N_Sc`) in `_reverse_osmosis_base` needs to be indexed by component because it can be either NaCl or TDS or other depending on the package invoked
      - Connected to #876 
- [x] While all property names have been updated to have corresponding indices for phase/comp, this is still not reflected by the actual constraints. All constraints should have same indices as the variables they calculate. 
      - Connected to #309 
- [x] Several property packages have 'hard-coded' phase/comp names in their constraints. This should be changed to use `p` and `j`, for phase and component, respectively. Do not hard-code names of phases and comps if avoidable.

MarcusHolly:
## Summary/Motivation:
Preliminary file changes to make sure I'm making all the correct changes

- Update constraints
- One utility function that applies all transformation

### Legal Acknowledgement


watertap/property_models/NaCl_prop_pack.py
```diff
@@ -580,14 +582,16 @@ def _conc_mass_phase_comp(self):
             doc="Mass concentration",
         )
 
-        def rule_conc_mass_phase_comp(b, j):
+        def rule_conc_mass_phase_comp(b, p, j):
             return (
-                b.conc_mass_phase_comp["Liq", j]
-                == b.dens_mass_phase["Liq"] * b.mass_frac_phase_comp["Liq", j]
+                b.conc_mass_phase_comp[p, j]
+                == b.dens_mass_phase[p] * b.mass_frac_phase_comp[p, j]
```
MarcusHolly, line 588: If the rule for conc_mass_phase_comp references the dens_mass_phase, should the index for dens_mass_phase remain ["Liq"] or should it be p?

watertap/property_models/NaCl_prop_pack.py
```diff
@@ -702,9 +712,8 @@ def _osm_coeff(self):
 
         def rule_osm_coeff(b):
             return b.osm_coeff == (
-                b.params.osm_coeff_param["2"]
-                * b.mass_frac_phase_comp["Liq", "NaCl"] ** 2
-                + b.params.osm_coeff_param["1"] * b.mass_frac_phase_comp["Liq", "NaCl"]
+                b.params.osm_coeff_param["2"] * b.mass_frac_phase_comp[p, "NaCl"] ** 2
+                + b.params.osm_coeff_param["1"] * b.mass_frac_phase_comp[p, "NaCl"]
```
MarcusHolly, line 716: If the rule name does not have a "_phase" (e.g. it does not have self.params.phase_list as a variable) in its name, should I still replace the ["Liq"] with [p]?

MarcusHolly:
For the utility functions at the end, is the goal to combine the two functions into one or should the variables with just a phase index remain separate from variables with a phase and comp index? I'm aware I need to fix the failing tests...

MarcusHolly:
@TimBartholomew or @bknueven I am still a bit confused on how I should be modifying the utility functions for the property models, and it looks like this may be what's going most of these errors. So can one of you briefly re-explain how I am supposed to modify the utility functions?

PR #820, opened 2022-10-20
Remove unnecessary requirements file

wigging:
## Fixes/Resolves:

There are two requirements files in this repository. Only one is needed. This pull request removes the unneeded file.

## Summary/Motivation:

Use only one requirements.txt file for the project.

- Remove the requirements.txt file
- Rename the requirements-dev.txt to requirements.txt
- Update all filename references from requirements-dev.txt to requirements.txt

### Legal Acknowledgement


lbianchi-lbl (changes requested):
Thanks for taking the time to create this PR @wigging. This seems to rename `requirements-dev.txt` to `requirements.txt`. While from a purely technical standpoint the name of this file is ultimately arbitrary, the choice of having `requirements-dev.txt` was deliberate and motivated by the desire to keep the naming consistent across projects within the IDAES ecosystem. With this in mind, I'd recommend reverting the changes so that `requirements.txt` is removed and `requirements-dev.txt` is kept (this is what I originally meant with [my comment](https://github.com/watertap-org/watertap/discussions/815#discussioncomment-3874365) on the original discussion, but in hindsight I realize that the context as to why one name is preferable to the other was missing from there).

ksbeattie (approved):
LGTM

lbianchi-lbl (approved):

PR #821, opened 2022-10-20
Remove unused imports in watertap.core

wigging:
## Fixes/Resolves:

There are a lot of unused imports in the `watertap/core` sub package. This request removes those imports.

## Summary/Motivation:

Unused imports clutter up the code and are not needed for functionality.

- Remove unused imports

### Legal Acknowledgement


wigging:
Removing the `CONFIG_Template` import from `watertap.core.membrane_channel0d` causes several tests to fail (see error below). But `CONFIG_Template` is not used anywhere in `watertap.core.membrane_channel0d`. Can someone explain how this CONFIG_Template is being used by watertap or by idaes?

```
Error: lecting ... 2022-10-20 15:55:21 [ERROR] idaes.watertap.ui.fsapi: Cannot import module 'watertap.examples.flowsheets.case_studies.wastewater_resource_recovery.amo_1690.amo_1690_ui': cannot import name 'CONFIG_Template' from 'watertap.core.membrane_channel0d' 
```

andrewlee94:
It looks like `amo_1690_ui` is trying to import the `ConfigTemplate` via the membrane models, so it is probably needed there. Of course, `amo_1690_ui` could have imported directly from IDAES, but they might have wanted to do this to make it more accessible.

More generally, the Config Template is a generic configuration block for a typical IDAES model and defines the commonly expected arguments. I will note that IDAES itself doesn't make much use of it.

wigging:
@andrewlee94 I'm looking at `watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/amo_1690/amo_1690_ui.py` and don't see anything about a ConfigTemplate. Am I looking in the wrong place?

adam-a-a:
> OR]

@wigging Config_Template is used via `membrane_channel_base.py` which `membrane_channel_0D` inherits from.

EDIT: With that said, I don't see any use of it in amo_1690_ui either...

wigging:
@adam-a-a and @andrewlee94 Should I just submit an issue related to this `CONFIG_Template`? And if a fix is needed, then I can submit a separate pull request.

bknueven:
`membrane_channel_base.py` defines its down `CONFIG_Template` for the membrane channel, which is distinct from the IDAES control volume `CONFIG_Template`.

I import it in `membrane_channel_0d.py` as a convenience for the "user". If they want the `CONFIG_Template`, then its most natural to grab it from this module.

So, unlike most of the other extra imports in the code, this one is entirely intentional on my part.

TimBartholomew (approved):

bknueven (approved):

PR #831, opened 2022-10-28
OARO Unit Model Documentation

MarcusHolly:
## Summary/Motivation:
Creates documentation for the osmotically assisted reverse osmosis unit model

### Legal Acknowledgement


MarcusHolly:
One thing I noticed that may be missing from the OARO unit model files is the solute diffusion coefficient ("D" on page 3 of this literature reference: https://app.box.com/file/1038520214412). I created a variable for it in the documentation, but was unsure of what its variable name should be because I couldn't find it referenced anywhere in the unit model files.

MarcusHolly:
Also, is there a reason osmotic pressure wasn't listed as a variable in the RO documentation? If so, then I should probably remove it from this documentation as well.

 https://watertap--831.org.readthedocs.build/en/831/technical_reference/unit_models/reverse_osmosis_0D.html

adam-a-a:
> Also, is there a reason osmotic pressure wasn't listed as a variable in the RO documentation? If so, then I should probably remove it from this documentation as well.
> 
> 
> 
>  https://watertap--831.org.readthedocs.build/en/831/technical_reference/unit_models/reverse_osmosis_0D.htmlv

Likely because it is a property and is accounted for in the property model documentation

MarcusHolly:
> > Also, is there a reason osmotic pressure wasn't listed as a variable in the RO documentation? If so, then I should probably remove it from this documentation as well.
> > https://watertap--831.org.readthedocs.build/en/831/technical_reference/unit_models/reverse_osmosis_0D.htmlv
> 
> Likely because it is a property and is accounted for in the property model documentation

I see. I was just a bit confused because π was used in the equations, but it wasn't defined anywhere in the RO documentation. So is it better to expect people to find this in the property model documentation or should I Ieave it in the variables section as is? I suppose doing the latter would mean that we need to eventually go back and modify a lot of other documentation at some point, which would not be ideal.

adam-a-a:
> One thing I noticed that may be missing from the OARO unit model files is the solute diffusion coefficient ("D" on page 3 of this literature reference: https://app.box.com/file/1038520214412). I created a variable for it in the documentation, but was unsure of what its variable name should be because I couldn't find it referenced anywhere in the unit model files. 
> 
> 

This is a property that'd be covered in property model documentation. With that said, the property name is `diffus_phase_comp` if you still want describe it in this documentation

adam-a-a:
LGTM- any further modification could be incorporated in a subsequent PR.

adam-a-a (approved):

docs/technical_reference/unit_models/osmotically_assisted_reverse_osmosis_0D.rst
```diff
@@ -0,0 +1,198 @@
...
+On the other hand, configuring the OARO unit to calculate concentration polarization effects, mass transfer
+coefficient, and pressure drop would result in 7 additional degrees of freedom. In this case, in addition to the
+previously fixed variables, we typically fix the following variables to fully specify the unit:
+
+    * feed-spacer porosity
+    * feed-channel height
+    * feed-velocity
+    * permeate-space porosity
+    * permeate-channel height
+    * permeate-side velocity
```
luohezhiming: @adam-a-a Do we need to specify `permeate-side velocity`? I checked the test file, it seems only `feed-side velocity` is fixed.

luohezhiming (commented):
LGTM. I only have a question on the DOF when every configurations set calculated.

luohezhiming (approved):

PR #833, opened 2022-11-01
Add frictional pressure drop to electrodialysis 

lbibl:
To accommodate pumps in a ED flowsheet in the pipeline. 

-Added methods calculate fluid dimensionless quantities.
-Added multiple methods to calculate pressure drop in ED channels. 

### Legal Acknowledgement


lbibl:
@bknueven , could you take a quick look if it's ok to bypass the failed test by the NUMPS (MACOS) solver ?  I can also try arbitrarily tailoring it a little further.  Thanks.

bknueven:
> @bknueven , could you take a quick look if it's ok to bypass the failed test by the NUMPS (MACOS) solver ?  I can also try arbitrarily tailoring it a little further.  Thanks.

I think this is okay for this PR. I've noticed separately that the electrodialysis unit is a bit unstable -- we should save addressing both for a follow-up PR.

lbibl:
> > @bknueven , could you take a quick look if it's ok to bypass the failed test by the NUMPS (MACOS) solver ?  I can also try arbitrarily tailoring it a little further.  Thanks.
> 
> I think this is okay for this PR. I've noticed separately that the electrodialysis unit is a bit unstable -- we should save addressing both for a follow-up PR.

Thanks Ben.  I pushed another commit pending to be seen if passing all.  I agree and I am working on a new electrodialysis  flowsheet targeting a more carefully designed optimization.  The current one was made quite preliminarily and will be updated or replaced hopefully soon.

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -64,6 +64,25 @@ class ElectricalOperationMode(Enum):
...
+    none = 0
+    experimental = 1
+    Darcy_Weisbach = 2
+
+
+class FrictionFactorMethod(Enum):
+    defined = 0
+    Ponzio = 1
+    Gurreri = 2
+    Kuroda = 3
```
adam-a-a: Very nice!

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -112,6 +131,62 @@ class Electrodialysis1DData(UnitModelBlockData):
...
+            doc="""
+     **default** - ``FrictionFactorMethod.defined``
+
+       .. csv-table::
+           :header: "Configuration Options", "Description"
+
+           "``FrictionFactorMethod.defined``", "Friction factor is defined and fixed by users" 
+           "``FrictionFactorMethod.Ponzio``", "Frction factor evaluated based on Pazio's work"
+           "``FrictionFactorMethod.Gurreri``", "Frction factor evaluated based on Gurreri's work"
+           "``FrictionFactorMethod.Kuroda``", "Frction factor evaluated based on Kuroda's work"
```
adam-a-a: "i" missing in "Frction factor"

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -112,6 +131,62 @@ class Electrodialysis1DData(UnitModelBlockData):
...
+            description="Method to calculate the hydraulic diameter for a rectangular channel in ED",
+            doc="""
+     **default** - ``HydraulicDiameterMethod.conventional``
+
+       .. csv-table::
+           :header: "Configuration Options", "Description"
+           
+           "``HydraulicDiameterMethod.defined``", "Hydraulic diameter is defined and fixed by users" 
+           "``HydraulicDiameterMethod.conventional``", "Conventional method for a rectangular channel with spacer porosity considered" 
+           "``HydraulicDiameterMethod.spacer_specific_area_known``", "A method for spacer-filled channel requiring the spapcer specific area data"
```
```suggestion
           "``HydraulicDiameterMethod.spacer_specific_area_known``", "A method for spacer-filled channel requiring the spacer specific area data"
```

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -112,6 +131,62 @@ class Electrodialysis1DData(UnitModelBlockData):
...
+            default=HydraulicDiameterMethod.conventional,
+            domain=In(HydraulicDiameterMethod),
+            description="Method to calculate the hydraulic diameter for a rectangular channel in ED",
+            doc="""
+     **default** - ``HydraulicDiameterMethod.conventional``
+
+       .. csv-table::
+           :header: "Configuration Options", "Description"
+           
+           "``HydraulicDiameterMethod.defined``", "Hydraulic diameter is defined and fixed by users"
```
```suggestion
           "``HydraulicDiameterMethod.fixed``", "Hydraulic diameter is defined and fixed by users" 
```
Would be more in line with what we did for other models (i.e., RO)

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -439,11 +521,17 @@ def build(self):
             units=pyunits.meter,
             doc="The width of the electrodialysis cell, parallel to the current/ion flux direction",
         )
-        self.spacer_thickness = Var(
+        self.channel_height = Var(
```
adam-a-a, line 500: Should the corresponding change be made in 0D?

watertap/unit_models/electrodialysis_1D.py
adam-a-a: Should the new options also be in 0D?

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -64,6 +64,25 @@ class ElectricalOperationMode(Enum):
...
+    defined = 0
+    Ponzio = 1
+    Gurreri = 2
+    Kuroda = 3
+
+
+class HydraulicDiameterMethod(Enum):
+    defined = 0
+    spacer_specific_area_known = 1
+    conventional = 2
```
adam-a-a: Adding the corresponding model changes into the documentation would be good.

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -1556,6 +1690,170 @@ def eq_dl_thickness(self, mem, side, t, x):
...
+                    self.config.hydraulic_diameter_method
+                    == HydraulicDiameterMethod.conventional
+                ):
+                    return (
+                        self.hydraulic_diameter
+                        == 2
+                        * self.channel_height
+                        * self.cell_width
+                        * self.spacer_porosity
+                        * (self.channel_height + self.cell_width) ** -1
```
adam-a-a, line 1733: I think this _might_ be incorrect. Hydraulic diameter is 4* cross-sectional area divided by the wetted perimeter, accounting for the spacer. It looks like you're accounting for spacer porosity in the area but not for the wetted perimeter.

adam-a-a: For a secondary reference, you can look at how I incorporated `dh` for the RO model.

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -1556,6 +1690,170 @@ def eq_dl_thickness(self, mem, side, t, x):
...
+                    return (
+                        self.hydraulic_diameter
+                        == 4
+                        * self.spacer_porosity
+                        * (
+                            2 * self.channel_height**-1
+                            + (1 - self.spacer_porosity) * self.spacer_specific_area
+                        )
+                        ** -1
+                    )
```
adam-a-a, line 1748: We might not need this _if_ my assertion above is correct.

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -1556,6 +1690,170 @@ def eq_dl_thickness(self, mem, side, t, x):
...
+
+        @self.Constraint(
+            doc="To calculate Re",
+        )
+        def eq_Re(self):
+
+            return (
+                self.Re
```
```suggestion
                self.N_Re
```
For dimensionless numbers, the IDAES standard name should start with `N_`

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -1556,6 +1690,170 @@ def eq_dl_thickness(self, mem, side, t, x):
...
+                * self.visc_d**-1
+            )
+
+        @self.Constraint(
+            doc="To calculate Sc",
+        )
+        def eq_Sc(self):
+
+            return (
+                self.Sc
```
```suggestion
                self.N_Sc
```

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -1556,6 +1690,170 @@ def eq_dl_thickness(self, mem, side, t, x):
...
+                * self.dens_mass**-1
+                * self.general_mass_diffusivity**-1
+            )
+
+        @self.Constraint(
+            doc="To calculate Sc",
+        )
+        def eq_Sh(self):
+
+            return self.Sh == 0.29 * self.Re**0.5 * self.Sc**0.33
```
adam-a-a: same as above about naming of dimensionless numbers.. also look at RO as a cross-ref

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -1556,6 +1690,170 @@ def eq_dl_thickness(self, mem, side, t, x):
...
+
+    def _pressure_drop_calculation(self):
+        self.pressure_drop = Var(
+            self.flowsheet().time,
+            initialize=1e4,
+            units=pyunits.pascal * pyunits.meter**-1,
+            doc="pressure drop per unit of length in a single ED channel",
```
adam-a-a: So, you're using a ControlVolume, which should have a config option called `has_pressure_change`. When set True, a `deltaP` variable is created. Although this may work, and might even be clearer, for the sake of consistency, you probably should be using the ControlVolume to do this (e.g., as RO model does).

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -1556,6 +1690,170 @@ def eq_dl_thickness(self, mem, side, t, x):
...
+                        == FrictionFactorMethod.Gurreri
+                    ):
+                        return (
+                            self.friction_factor
+                            == 4 * 50.6 * self.spacer_porosity**-7.06 * self.Re**-1
+                        )
+                    elif (
+                        self.config.friction_factor_method
+                        == FrictionFactorMethod.Kuroda
+                    ):
```
adam-a-a: I really like that you added more options for calculating the friction factor, which is something we might want to do in RO later on. Just a comment of admiration =)

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -1556,6 +1690,170 @@ def eq_dl_thickness(self, mem, side, t, x):
...
+                )
+                def eq_friction_factor(self):
+                    if (
+                        self.config.friction_factor_method
+                        == FrictionFactorMethod.Ponzio
+                    ):
+                        if value(self.Re) < 61:
+                            return self.friction_factor == 1400 * self.Re**-1
+                        else:
+                            return self.friction_factor == 104.5 * self.Re**-0.37
```
adam-a-a: Oh, this _could_ be problematic if Reynolds number isn't fixed by the user. Do we want to consider using something like an "expression if" (`Expr_IF`)?

adam-a-a (changes requested):
I have a few comments, questions, and recommended changes.

watertap/unit_models/electrodialysis_1D.py
lbibl: Yes if we want (I say this just because it looks like 1D will be selected for most use cases).  I am thinking a separate PR for 0D to mirror changes in 1D,  after the 1D unit works for the new flowsheet under development to our satisfaction.

watertap/unit_models/electrodialysis_1D.py
lbibl: Same here about mirroring the changes to 0d.

watertap/unit_models/electrodialysis_1D.py
lbibl: Yes. To do in this quarter.

watertap/unit_models/electrodialysis_1D.py
lbibl: Get your point Adam.  I carefully reviewed literature on this and actually had the same question to yours- but this equation was used by a few papers. So my take on this is that the hydraulic diameter for a porous media gonna carry a higher degree of error/uncertainty than regular channels (e.g., pipes) so any equation can only do so much.  The other method I had is corresponding to the method in RO with a modification to reflect the original derivation of that equation.  I can explain this more in tomorrow's meeting.

watertap/unit_models/electrodialysis_1D.py
lbibl: This would be the equation from which is derived the equation you used in RO.  I traced the source of the equation and this is the original propose.  Will explain more tomorrow.

watertap/unit_models/electrodialysis_1D.py
lbibl: Thanks! Just worked this out... I can push the new adds to this PR or in my next PR.

watertap/unit_models/electrodialysis_1D.py
lbibl: Thank you!  This might reflect somewhat my hard time in choosing things =)

watertap/unit_models/electrodialysis_1D.py
lbibl: > I have a few comments, questions, and recommended changes.

Thanks Adam!  I just replied to some of your comments.  I will make commits to address the rest.

lbibl:
> I have a few comments, questions, and recommended changes.

Thanks Adam! I just replied to some of your comments. I will make commits to address the rest.

watertap/unit_models/electrodialysis_1D.py
adam-a-a: We definitely should for consistency. I agree that this can be done in a subsequent PR.

watertap/unit_models/electrodialysis_1D.py
adam-a-a: I am guessing that you tracked this down to Schock & Miquel, 1987. Right--I recall the originally proposed equation and opted for the implementation that you see in the RO model to avoid the necessity for a specific area specification. Instead, the way it is in RO, you would just need spacer porosity, channel height, and width.

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -1556,6 +1690,170 @@ def eq_dl_thickness(self, mem, side, t, x):
...
+                return (
+                    self.pressure_drop[t]
+                    == self.dens_mass
+                    * self.friction_factor
+                    * self.velocity_D[0, 0] ** 2
+                    * 0.5
+                    * self.hydraulic_diameter**-1
+                )
+
+            if self.config.friction_factor_method == FrictionFactorMethod.defined:
```
```suggestion
            if self.config.friction_factor_method == FrictionFactorMethod.fixed:
```

watertap/unit_models/electrodialysis_1D.py
adam-a-a: Alternatively, we could consider skipping this one all together if we want to avoid the complication. Is this one tested by the way?

bknueven: I just want to echo Adam's comment: as-is this is a modeling error. If we need the conditional I think expr_if is an option.

watertap/unit_models/electrodialysis_1D.py
lbibl: Upon some discussion with Tim, decisions related to friction factor and hydraulic diameter methods are (1) taking the method needing a `express_if` out to avoid complicatedness, (2) keep the specific surface area (S_v) d_h method as a more general equation and I can note somewhere (e.g., documentation) that under certain circumstances S_v can be approximated by 8/h, which leads to the eq used in RO.

watertap/unit_models/electrodialysis_1D.py
lbibl: Resolution is made to remove this method without hurting the model's performance to avoid complicatedness.  Thanks for pointing out this mistake, which I did not realize...

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -1556,6 +1691,165 @@ def eq_dl_thickness(self, mem, side, t, x):
                     * self.current_dens_lim_x[t, x] ** -1
                 )
 
+    def _get_fluid_dimensionless_quantities(self):
+        self.general_mass_diffusivity = Var(
```
adam-a-a: Should this be diffus_phase_comp in prop package?

adam-a-a: Ah, "not individual ions". Maybe change naming to line up more closely with `diffus_phase_comp`?

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -1556,6 +1691,165 @@ def eq_dl_thickness(self, mem, side, t, x):
...
+                    self.config.hydraulic_diameter_method
+                    == HydraulicDiameterMethod.conventional
+                ):
+                    return (
+                        self.hydraulic_diameter
+                        == 2
+                        * self.channel_height
+                        * self.cell_width
+                        * self.spacer_porosity
+                        * (self.channel_height + self.cell_width) ** -1
```
adam-a-a, line 1733: I think we should remove this and only use the one below.

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -1556,6 +1691,165 @@ def eq_dl_thickness(self, mem, side, t, x):
...
+        def eq_Sh(self):
+
+            return self.N_Sh == 0.29 * self.N_Re**0.5 * self.N_Sc**0.33
+
+    def _pressure_drop_calculation(self):
+        self.pressure_drop = Var(
+            self.flowsheet().time,
+            initialize=1e4,
+            units=pyunits.pascal * pyunits.meter**-1,
+            doc="pressure drop per unit of length",
```
adam-a-a, line 1784: Isn't `deltaP`going to handle this? `deltaP` in 1D corresponds to dp_dx

watertap/unit_models/electrodialysis_1D.py
lbibl: I kept this as an option  for now because several papers used this equation. Maybe I can use this equation in some cases for comparison but I can also remove it in a later PR if ok.

watertap/unit_models/electrodialysis_1D.py
lbibl: Technically it can.  But since currently I have all friction calculation wrapped in a separate function plus I have two control volumes, it'd be cleanest to do it in this way, so like passing an intermediate var to two control volumes as controlled by a configuration.  I would keep this as is now.

watertap/unit_models/electrodialysis_1D.py
lbibl: done.

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -572,6 +633,20 @@ def build(self):
...
+            units=pyunits.meter * pyunits.second**-1,
+            doc="Linear velocity of flow",
+        )
+        self.velocity_C = Var(
+            self.flowsheet().time,
+            self.diluate.length_domain,
+            initialize=0.01,
+            units=pyunits.meter * pyunits.second**-1,
+            doc="Linear velocity of flow",
+        )
```
TimBartholomew: rename to be more specific `velocity_diluate` and `velocity_concentrate`

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -1556,6 +1691,162 @@ def eq_dl_thickness(self, mem, side, t, x):
...
+            initialize=10,
+            units=pyunits.dimensionless,
+            doc="friction factor of the channel fluid",
+        )
+
+        if self.config.pressure_drop_method == PressureDropMethod.experimental:
+            _log.warning(
+                "Do not forget to FIX the experimental pressure drop value in [Pa/m]!"
+            )
+        else:  # PressureDropMethod.Darcy_Weisbach is used
```
TimBartholomew, line 1802: It might be better to explicitly check that Darcy_Weisbach is specified with an elif, and leave the else to raise an error. Unless something above this already checks that the configuration arguments are valid

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -2019,6 +2385,62 @@ def calculate_scaling_factors(self):
...
+                )
+        if hasattr(self, "eq_pressure_drop_total"):
+            for i, c in self.eq_pressure_drop_total.items():
+                iscale.constraint_scaling_transform(
+                    c, iscale.get_scaling_factor(self.pressure_drop_total)
+                )
+        if hasattr(self, "eq_friction_factor"):
+            iscale.constraint_scaling_transform(
+                self.eq_friction_factor, iscale.get_scaling_factor(self.friction_factor)
+            )
```
TimBartholomew: @bknueven can correct me if I'm wrong, but I believe we let our jacobian autoscaling scale all of the constraints in unit models so we don't use `constraint_scaling_transform` in unit model files

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -1556,6 +1691,162 @@ def eq_dl_thickness(self, mem, side, t, x):
...
+        def eq_Sh(self):
+
+            return self.N_Sh == 0.29 * self.N_Re**0.5 * self.N_Sc**0.33
+
+    def _pressure_drop_calculation(self):
+        self.pressure_drop = Var(
+            self.flowsheet().time,
+            initialize=1e4,
+            units=pyunits.pascal * pyunits.meter**-1,
+            doc="pressure drop per unit of length",
```
TimBartholomew, line 1784: Currently you have a single pressure drop (pressure/length) for the whole module, our RO model calculates this locally for each node (I don't think this PR should necessarily change the calculation to each node, but it could be considered for future PRs, or we should justify why ED shouldn't). 

Additionally, this pressure drop is only calculated based on diluate velocity, but you created a variable for the concentrate velocity. Is this a mistake? For our OARO model we calculate the pressure drop separately for each channel (feed and permeate side), it may make sense for ED as well.

watertap/unit_models/electrodialysis_1D.py
bknueven: In general, we shouldn't over-transform the constraints. It probably doesn't hurt much, however.

One place where you *might* still want and need to do this is if the absolute constraint tolerance we specify for Ipopt, `1e-08`, is either too loose (in which case scale the constraint up) or too tight (in which case scale the constraint down). So, if we expect the values these constraints are calculating are either very small or very large it is probably worth doing.

watertap/unit_models/electrodialysis_1D.py
lbibl: Thanks @TimBartholomew .  Yes, the current model assumes a constant velocity when calculating N_Re and friction pressure, although the actual velocity has ~4% change when water migration is molded. The error caused by this change for friction pressure will be even smaller because v is squared in its calculation.  I accept this approximation by the judging this error level is much insignificant compared to errors in calculating others, e.g., hydraulic diameter, and friction factor (which can an error of >10%).  I can note this in the later documentation update . 

Another thing to consider is that,  if velocity change is fully considered in accounting for pressure change, I think the correct way is to apply a Navier-Stokes or Bernoullie's equation to expressing energy and momentum conservations of the flow.  The velocity change in my case corresponds to a pressure change.   I haven't seen the two conservation equations be implemented in IDAES control-volumes and I think the approximation as-is is good for the purpose of the current model. 

Upon accepting the approximation, there is no difference between the flow in diluate and concentrate channels in the current ED setup because of identical channel dimensions.   All the friction pressure in the model is taking ` velocity_diluate[ 0, 0]` (which is equivalent to `velocity_concentrate[0,0]`)  and passed to both channels.  The same strategy is used for control volume `length_domain`. `velocity_concentrate` is indeed calculated and unused for other things, but I think it's ok to let them hanging there as a calculation output.  Inspecting velocities might be useful in the future somewhere. 

In summary, my recommendation is to accept the constant velocity approximation for calculating pressure drop thus far.  Let me know if you have more thoughts.   Thanks for your comment that made me think it through more.

watertap/unit_models/electrodialysis_1D.py
lbibl: Yes, I think the configuration arguments is well bounded to make the `else` condition Darcy_Weisbach only.

watertap/unit_models/electrodialysis_1D.py
lbibl: Done.

watertap/unit_models/electrodialysis_1D.py
lbibl: Thanks for the clarification.   I take this as it's ok to keep things as they are now.

watertap/unit_models/electrodialysis_1D.py
srikanthallu: @lbibl, this approximation has to be documented. Can you please open another ticket for this issue. If RO is including the pressure drop for each channel, ED should be be able to model this as well.

srikanthallu (approved):

watertap/unit_models/electrodialysis_1D.py
lbibl: Thanks for reviewing, @srikanthallu .  I will make sure updating the decimation to reflect key contents.  To clarify, pressure drops for both channels are accounted in the current.

TimBartholomew (approved):
LGTM, although I think we may want to make this more similar to RO membrane channels in the future. It may even be possible to generalize the membrane channel control volume and apply it to the ED model.

adam-a-a (approved):

PR #834, opened 2022-11-02
Initial commit on BPE equation in seawater property package

Zhuoran29:
## Fixes/Resolves:

N.A.

## Summary/Motivation:
Add boiling point elevation in seawater property package for the use in unit models

- Add BPE to seawater property package
- Add test for BPE
- Add documentation for BPE

### Legal Acknowledgement

### DIFF
--- a/watertap/property_models/seawater_prop_pack.py
+++ b/watertap/property_models/seawater_prop_pack.py
@@ -1462,7 +1462,9 @@ class SeawaterStateBlockData(StateBlockData):
 
         def rule_boiling_point_elevation_phase(
             b,
+            p,
         ):  # boiling point elevation of seawater from eq. 36 in Sharqawy et al. (2010)
+
             t = b.temperature - 273.15 * pyunits.K
             s = b.mass_frac_phase_comp["Liq", "TDS"]
             A = b.params.bpe_A0 + b.params.bpe_A1 * t + b.params.bpe_A2 * t**2
@@ -1470,7 +1472,7 @@ class SeawaterStateBlockData(StateBlockData):
             return b.boiling_point_elevation_phase["Liq"] == A * s**2 + B * s
 
         self.eq_boiling_point_elevation_phase = Constraint(
-            rule=rule_boiling_point_elevation_phase
+            self.params.phase_list, rule=rule_boiling_point_elevation_phase
         )
 
     # -----------------------------------------------------------------------------
--- a/watertap/property_models/seawater_prop_pack.py
+++ b/watertap/property_models/seawater_prop_pack.py
@@ -705,7 +705,9 @@ class SeawaterParameterData(PhysicalParameterBlock):
                 "therm_cond_phase": {"method": "_therm_cond_phase"},
                 "dh_vap_mass": {"method": "_dh_vap_mass"},
                 "diffus_phase_comp": {"method": "_diffus_phase_comp"},
-                "bpe_phase": {"method": "_bpe_phase"},
+                "boiling_point_elevation_phase": {
+                    "method": "_boiling_point_elevation_phase"
+                },
             }
         )
         # TODO: add diffusivity variable and constraint since it is needed when calculating mass transfer coefficient in
@@ -1449,8 +1451,8 @@ class SeawaterStateBlockData(StateBlockData):
 
         self.eq_dh_vap_mass = Constraint(rule=rule_dh_vap_mass)
 
-    def _bpe_phase(self):
-        self.bpe_phase = Var(
+    def _boiling_point_elevation_phase(self):
+        self.boiling_point_elevation_phase = Var(
             self.params.phase_list,
             initialize=5e-1,
             bounds=(0, 1e9),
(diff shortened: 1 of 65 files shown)


watertap/property_models/seawater_prop_pack.py
```diff
@@ -625,7 +625,7 @@ def build(self):
         self.bpe_A0 = Var(
             within=Reals,
             initialize=17.95,
-            units=t_inv_units,
+            units=pyunits.K,
```
andrewlee94, line 628: Whilst it doesn't really matter here, it is generally good practice to take the units from the property package metadata rather than hard-coding it here.

Note also that IDAES recently changed how it handles metadata for units - whilst the old approach still works, the new (improved) API is:

```
units_metadata = self.get_metadata().default_units

temperature_units = units_metadata.TEMPERATURE
```

Note this this should support autocompletion in most IDEs.

watertap/property_models/seawater_prop_pack.py
```diff
@@ -704,6 +705,7 @@ def define_metadata(cls, obj):
                 "therm_cond_phase": {"method": "_therm_cond_phase"},
                 "dh_vap_mass": {"method": "_dh_vap_mass"},
                 "diffus_phase_comp": {"method": "_diffus_phase_comp"},
+                "bpe_phase": {"method": "_bpe_phase"},
```
andrewlee94: I would suggest a more descriptive name for this: `boiling_point_elevation`. It is longer and more cumbersome, but it is much more informative to someone trying to use the property package.

watertap/property_models/seawater_prop_pack.py
```diff
@@ -1447,6 +1449,26 @@ def rule_dh_vap_mass(
 
         self.eq_dh_vap_mass = Constraint(rule=rule_dh_vap_mass)
 
+    def _bpe_phase(self):
+        self.bpe_phase = Var(
+            self.params.phase_list,
+            initialize=5e-1,
+            bounds=(0, 1e9),
+            units=pyunits.K,
```
andrewlee94, line 1459: As before regarding units of measurement,

watertap/property_models/seawater_prop_pack.py
```diff
@@ -1447,6 +1449,26 @@ def rule_dh_vap_mass(
...
+        def rule_bpe_phase(
+            b,
+        ):  # boiling point elevation of seawater from eq. 36 in Sharqawy et al. (2010)
+            t = b.temperature - 273.15 * pyunits.K
+            s = b.mass_frac_phase_comp["Liq", "TDS"]
+            A = b.params.bpe_A0 + b.params.bpe_A1 * t + b.params.bpe_A2 * t**2
+            B = b.params.bpe_B0 + b.params.bpe_B1 * t + b.params.bpe_B2 * t**2
+            return b.bpe_phase["Liq"] == A * s**2 + B * s
+
+        self.eq_bpe_phase = Constraint(rule=rule_bpe_phase)
```
andrewlee94: Due to some underlying issues in Pyomo, there can be cases where construction of properties on demand fails on the first attempt (due to order of operations). These instances are rare, but to protect against this you need to wrap all of this in a `try/except` block:

```
def _bpe_phase(self):
    try:
        # Build the variable and constraints
    except AttributeError:
        self.del_component(self.bpe_phase)
        self.del_component(self.eq_bpe_phase)
```

andrewlee94 (commented):
Looks good for the most part - my only comments are on little technicalities (which probably don't matter for WaterTap but I thought I should make you aware of).

watertap/property_models/seawater_prop_pack.py
Zhuoran29: Thanks a lot for the detailed suggestions, Andrew. One quick question regarding the metadata of units, currently the seawater property package uses the old approach to specify the units for all other property variables, do you think I should update them as well?

andrewlee94: It would be nice, but hardly necessary. So, if you have the time it would be good to do (and serve as an example of how to do things in the future), but if there are more pressing things to do it should not hold up this PR.

Zhuoran29: Thank you, it makes sense.

watertap/property_models/seawater_prop_pack.py
adam-a-a: We would need to apply this block to all other on-demand variables/constraints for consistency then. I think that is beyond the scope of this PR and can be reserved for a subsequent PR.

watertap/property_models/seawater_prop_pack.py
adam-a-a: I agree. I also wonder if this needs `_phase` in the name. I think not, unless there is an IDAES standard that _does_ include `phase` as a qualifier.

Reviewing the IDAES standards (https://idaes-pse.readthedocs.io/en/1.4.0/standards.html), it seems that we have no restriction to abide by, so perhaps just `boiling_point_elevation` would be fine without the phase qualifier and corresponding index.

andrewlee94: I did ask this question myself - for this application it is fairly obvious that there is only one BPE value. However, there are cases where you might have two liquid phases with different BPE values - these are probably not on the immediate horizon for WaterTap, but it might be worth thinking ahead a little in this case.

adam-a-a (commented):
LGTM - I think we just need to rename to `boiling_point_elevation` and remove the phase index from the variable and constraint before approving.

watertap/property_models/seawater_prop_pack.py
adam-a-a: Ah, I didn't think of that!

adam-a-a (commented):
Actually, @andrewlee94 's comment made me change my mind. Going back on what I said in my last comment, let's keep the phase qualifier and index! The only change I think we need then is to change `bpe` to `boiling_point_elevation`.

adam-a-a:
@Zhuoran29 - I think you accidentally merged the lt-med branch back in to this one, so now the tests are failing and the lt-med files are here. We'll want to roll that back.

Zhuoran29:
@adam-a-a Yes, I'm trying to revert the merge before I request the reviews again.

adam-a-a (approved):
LGTM - I think the other suggestions by @andrewlee94 are changes that need to be broadly applied across this property model (and others) and is beyond the scope of this PR.

watertap/property_models/seawater_prop_pack.py
Zhuoran29: The variable name has been updated to `boiling_point_elevation_phase`
Also, I noticed that for most of the variables with a phase index, the scaled constraints are not indexed (except for osmotic pressure: `pressure_osm_phase`, which has a unique way of transforming constraints). We might want to unify the transforming approach for the same type of variables as well in a subsequent PR.

@andrewlee94 For this PR I think it makes sense to just implement the BPE equation, and in a different PR we can update the approaches you mentioned. Thanks again for those insights.
@adam-a-a Thanks for helping resolve the accident merge issue.

tarnold17 (approved):

watertap/property_models/seawater_prop_pack.py
Zhuoran29: Reserved for future commit.

PR #835, opened 2022-11-02
Reorganzing the WaterTAP Costing Package

bknueven:
## Summary/Motivation:
The WaterTAP costing package was too big. This breaks out all its costing methods and associated parameters into their own files.

- [x] Separate out costing methods and associated parameters from the watertap costing package into their own modules
- [x] Add basic utility costing methods and helpers to `watertap/costing/util.py`
- [x] Test associated lazy construction techniques for robustness

## Questions:
- Are the costing methods in the "right place"? We could put them closer to or even in the unit model modules.
- Does the interface make sense?
   - Here is a simple example: https://github.com/watertap-org/watertap/blob/17c4acc53133ec289af36033bf592e1bec76e3e0/watertap/costing/units/energy_recovery_device.py
   - This is a complex example: https://github.com/watertap-org/watertap/blob/17c4acc53133ec289af36033bf592e1bec76e3e0/watertap/costing/units/ion_exchange.py

### Legal Acknowledgement


watertap/unit_models/tests/test_ion_exchange_0D.py
```diff
@@ -930,13 +930,13 @@ def obj_rule(m):
         results = solver.solve(m, tee=False)
         assert_optimal_termination(results)
 
-        assert pytest.approx(440372.278, rel=1e-5) == value(
+        assert pytest.approx(453807.598, rel=1e-5) == value(
```
TimBartholomew, line 928: Why are the ion exchange costs changing? Was a bug fixed?

TimBartholomew (approved):
LGTM - one minor comment

watertap/unit_models/tests/test_ion_exchange_0D.py
bknueven: Yes, look at f3473c3. There were some uncaught issue with units.

TimBartholomew:
@OOAmusat @lbibl this seems like a big PR, however, it is essentially moving the costing methods for each unit out of the WaterTAP costing package into a separate file. Could one of you review this, you don't need to go through it line by line. Just see one example of the unit model costing methods now interacts with the costing package

OOAmusat (approved):
LGTM

lbibl (approved):
Looks good and a better organization.

PR #836, opened 2022-11-04
Remove idaes solver directives

bknueven:
## Summary/Motivation:
We're skipping several test which pass without an IDAES solver.

- Remove `requires_idaes_solver` where not needed

### Legal Acknowledgement

### DIFF
--- a/watertap/examples/flowsheets/full_treatment_train/flowsheet_components/chemistry/tests/test_posttreatment.py
+++ b/watertap/examples/flowsheets/full_treatment_train/flowsheet_components/chemistry/tests/test_posttreatment.py
@@ -69,7 +69,6 @@ def test_ideal_naocl_chlorination():
 
 
 @pytest.mark.component
-@pytest.mark.requires_idaes_solver
 def test_ideal_naocl_chlorination_full_block():
     model = run_chlorination_block_example(fix_free_chlorine=True)
     assert model.fs.ideal_naocl_mixer_unit.dosing_rate.value == pytest.approx(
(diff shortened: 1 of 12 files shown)


hunterbarber (approved):
This revision is good.

TimBartholomew:
I support this, what are the next steps to address the ion exchange test failures?

bknueven:
> I support this, what are the next steps to address the ion exchange test failures?

Something changed between when I first opened this PR and when I merged in main 12 days ago. I've merged in main again to see if there's been a subsequent change in ion exchange (or anywhere else).

The easiest answer is "maintain the `requires_idaes_solver` directive on the failing tests". I'm not sure it's worth the effort to do more, though I generally think that a failure with the conda Ipopt is a sign of a potentially unstable model, at least in WaterTAP.

kurbansitterley:
I am not sure what to do about the differences in solutions between Windows and Mac machines. Recall that part of the reason it took so long to merge in the IX model is because of this exact issue. The PR motivation is to remove idaes solver "where not needed". My view is that it is needed in this case.

bknueven:
> I am not sure what to do about the differences in solutions between Windows and Mac machines. Recall that part of the reason it took so long to merge in the IX model is because of this exact issue. The PR motivation is to remove idaes solver "where not needed". My view is that it is needed in this case.

The difference is in the stricter initialization procedures we recently added. Before the unit was failing to initialize on macOS for one of the tests (and possibly on Linux/Windows too). We probably need to revisit the Ion Exchange initialization routine at some point.

That said, that's not what this PR is about, so I've just re-added the `requires_idaes_solver` directive in a few places.

andrewlee94 (approved):

PR #837, opened 2022-11-07
Property Package Constraint Indexing & Utility Function Update

### LINKED ISSUE #309: Add utility for scaling property constraints
Current property packages scale their property constraints with loops that are differentiated based on how the constraint is indexed (i.e. no index, indexed by phase, indexed by component, and indexed by phase and component).

A better approach would be if all constraints for a property are indexed exactly like the variable, then a single for loop could easily apply the transformation.

MarcusHolly:
## Fixes/Resolves:
- Issue #309

- Update the utility function of various property packages to apply all the transformations in a single loop
- All constraints for a property are indexed exactly like the variable 
- cryst_prop_pack was intentionally left untouched

### Legal Acknowledgement

### DIFF
--- a/watertap/core/util/scaling.py
+++ b/watertap/core/util/scaling.py
@@ -21,7 +21,7 @@ def transform_property_constraints(self):
         if metadata_dic["method"] is not None and self.is_property_constructed(var_str):
             var = getattr(self, var_str)
             if not isinstance(var, Var):
-                continue  # properties that are vars do not have constraints
+                continue  # properties that are not vars do not have constraints
             con = getattr(self, "eq_" + var_str)
             for ind, c in con.items():
                 sf = iscale.get_scaling_factor(var[ind], default=1, warning=True)
--- a/watertap/core/util/scaling.py
+++ b/watertap/core/util/scaling.py
@@ -11,7 +11,7 @@
 #
 ###############################################################################
 
-from pyomo.environ import Expression, Param
+import pyomo.environ as pyo
 import idaes.core.util.scaling as iscale
 
 
@@ -20,7 +20,7 @@ def transform_property_constraints(self):
         var_str = metadata_dic["name"]
         if metadata_dic["method"] is not None and self.is_property_constructed(var_str):
             var = getattr(self, var_str)
-            if not isinstance(var, Var):
+            if not isinstance(var, pyo.Var):
                 continue  # properties that are not vars do not have constraints
             con = getattr(self, "eq_" + var_str)
             for ind, c in con.items():
--- a/watertap/core/util/scaling.py
+++ b/watertap/core/util/scaling.py
@@ -1,0 +1,30 @@
+###############################################################################
+# WaterTAP Copyright (c) 2021, The Regents of the University of California,
+# through Lawrence Berkeley National Laboratory, Oak Ridge National
+# Laboratory, National Renewable Energy Laboratory, and National Energy
+# Technology Laboratory (subject to receipt of any required approvals from
+# the U.S. Dept. of Energy). All rights reserved.
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and license
+# information, respectively. These files are also available online at the URL
+# "https://github.com/watertap-org/watertap/"
+#
+###############################################################################
+
+from pyomo.environ import Expression, Param
+import idaes.core.util.scaling as iscale
+
+
+def transform_property_constraints(self):
+    for metadata_dic in self.params.get_metadata().properties.values():
+        var_str = metadata_dic["name"]
+        if metadata_dic["method"] is not None and self.is_property_constructed(var_str):
+            var = getattr(self, var_str)
+            if isinstance(var, Expression):
+                continue  # properties that are expressions do not have constraints
+            if isinstance(var, Param):
+                continue  # properties that are parameters do not have constraints
+            con = getattr(self, "eq_" + var_str)
+            for ind, c in con.items():
+                sf = iscale.get_scaling_factor(var[ind], default=1, warning=True)
+                iscale.constraint_scaling_transform(c, sf)
--- a/watertap/core/util/scaling.py
+++ b/watertap/core/util/scaling.py
@@ -20,10 +20,8 @@ def transform_property_constraints(self):
         var_str = metadata_dic["name"]
         if metadata_dic["method"] is not None and self.is_property_constructed(var_str):
             var = getattr(self, var_str)
-            if isinstance(var, Expression):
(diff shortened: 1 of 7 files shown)


MarcusHolly:
For the DSPMDE property package, I had to add a list of exclusions to the utility function at the end because none of those variables have a rule or constraint equation associated with them even though they are listed in the metadata. Is this a sufficient workaround or is there a better method?

andrewlee94:
@MarcusHolly The issue with the metadata would at first seem to be on of two issues: either you are using the metadata for something it really wasn't intended for, or those properties should not be in the metadata at all.

On a related note however, IDAES is working to improve the metadata so that we will be able to explicitly declare what proprieties are supported (separate from if they have a build-on-demand method). This is still WIP right now, but should be ready in a few weeks: https://github.com/IDAES/idaes-pse/pull/995

MarcusHolly:
> @MarcusHolly The issue with the metadata would at first seem to be on of two issues: either you are using the metadata for something it really wasn't intended for, or those properties should not be in the metadata at all.
> 
> On a related note however, IDAES is working to improve the metadata so that we will be able to explicitly declare what proprieties are supported (separate from if they have a build-on-demand method). This is still WIP right now, but should be ready in a few weeks: [IDAES/idaes-pse#995](https://github.com/IDAES/idaes-pse/pull/995)

Ok, I'm leaning more towards that these should not be listed in the metadata, but I'd like so more opinions on that first.

andrewlee94:
@MarcusHolly It definitely looks like those properties should not be in the metadata - they list a method which does not seem to appear in the class (unless there is some inheritance which I did not check).

MarcusHolly:
@andrewlee94 Well the fix won't be as simple as just deleting these parameters from the metadata because they are still directly used in the constraint equations and, by just deleting these parameters from the metadata, the following error emerges: `idaes.core.util.exceptions.PropertyNotSupportedError: fs.state_block[0] package property metadata method does not contain a method for mw_comp. Please select a package which supports the necessary properties for your process.
`

I'm not quite sure how to go about addressing this, especially because none of the other property models have parameters that are defined using the `add_object_reference` approach.

andrewlee94:
@MarcusHolly For some reason they didn't show up when I first did a search for them - I see them now. There should be kept in the metadata as they are valid properties.

However, I suspect I see the issue now too. Firstly, at least some of those are `Params` and do not have constraints associated with them - line 1972 should be `if isinstance(var, [Expression, Param]):`.

The second issue is that the method at the end makes some assumptions about the constraints associated with each property variable - namely there is only one and it has a standard name. My guess is that some of the those properties do not follow this convention.

MarcusHolly:
This PR should be ready for review now.

watertap/property_models/NDMA_prop_pack.py
```diff
@@ -532,13 +538,15 @@ def _mole_frac_phase_comp(self):
...
-                "Liq", j
-            ] / sum(b.flow_mol_phase_comp["Liq", j] for j in b.params.component_list)
+        def rule_mole_frac_phase_comp(b, p, j):
+            return b.mole_frac_phase_comp[p, j] == b.flow_mol_phase_comp[p, j] / sum(
+                b.flow_mol_phase_comp[p, j] for j in b.params.component_list
+            )
 
         self.eq_mole_frac_phase_comp = Constraint(
-            self.params.component_list, rule=rule_mole_frac_phase_comp
+            self.params.phase_list,
```
andrewlee94, line 549: This is more best practice than necessary but it is generally better to use `phase_component_set` in these cases - it doesn't really affect this property package but in more complex property packages this will ensure that non invalid phase-component pairs end up in the indexing set.

There are a lot of places where this could be applied.

andrewlee94 (commented):
I had one comment on "best practices" - it doesn't really affect these packages as there is only one phase but it would be good to use better practices to set a good example for the future.

watertap/property_models/NDMA_prop_pack.py
MarcusHolly: So basically this would require structuring all the property packages like cryst_prop_pack? I think this would require a decent bit of effort, so maybe this should be a separate PR? Also, I think it twould requiring hard-coding some phases and components as constraints (e.g. ["Liq"], ["NaCl"]), which is frequently done in the cryst_prop_pack. I think Tim had said we wanted to try and move away from that notation?

watertap/property_models/NDMA_prop_pack.py
```diff
@@ -753,27 +761,16 @@ def calculate_scaling_factors(self):
...
+            if metadata_dic["method"] is not None and self.is_property_constructed(
+                var_str
+            ):
+                var = getattr(self, var_str)
+                if isinstance(var, Expression):
+                    continue  # properties that are expressions do not have constraints
+                con = getattr(self, "eq_" + var_str)
+                for ind, c in con.items():
+                    sf = iscale.get_scaling_factor(var[ind], default=1, warning=True)
                     iscale.constraint_scaling_transform(c, sf)
```
TimBartholomew: I believe this can become a utility function: `transform_property_constraints`, in which all of the property packages that you modified in this PR can use

This could be located in `watertap/core/util` with a new file called `scaling.py`

TimBartholomew (commented):
one minor comment to make the constraint transform a utility

watertap/property_models/NDMA_prop_pack.py
andrewlee94: FYI: IDAES is working on a new API for scaling models which we hope will replace these methods (and automate a lot of this as well).

MarcusHolly: I assume the scaling.py file is small enough that it does not need a test file?

watertap/core/util/scaling.py
```diff
@@ -0,0 +1,30 @@
...
+
+def transform_property_constraints(self):
+    for metadata_dic in self.params.get_metadata().properties.values():
+        var_str = metadata_dic["name"]
+        if metadata_dic["method"] is not None and self.is_property_constructed(var_str):
+            var = getattr(self, var_str)
+            if isinstance(var, Expression):
+                continue  # properties that are expressions do not have constraints
+            if isinstance(var, Param):
+                continue  # properties that are parameters do not have constraints
```
bknueven: Should we just be asking if the thing we have is a `Var` and continuing if not? E.g.,
```suggestion
            if not isinstance(var, Var):
                continue  # properties that are vars do not have constraints
```

bknueven (approved):
Looks good -- I just have a minor comment.

TimBartholomew (approved):
LGTM, I added Ben's commit and then slightly edited the comment

Issue #832, opened 2022-10-28
Reference for citing watertap repository


wigging:
### Description

Provide a reference to the watertap repository in the README and in the documentation. This reference should contain a DOI along with other relevant information for citing software. The reference can be used in publications, reports, and presentations to cite the watertap codebase instead of using the GitHub link to the repository.

### Motivation

Other than using the link to the watertap repository, there is no way to cite the codebase. Open-source projects typically provide a reference so users can cite the software in their research. Here are some projects that provide a reference and DOI for their code:
- [Scientific Computational Imaging Code (SCICO)](https://github.com/lanl/scico), see the JOSS badge 
- [SALib](https://github.com/SALib/SALib), see the README
- [NumPy](https://github.com/numpy/numpy), see the DOI badge

### Possible Implementation

- [x] Add a reference in the README and online documentation.

wigging:
A reference to watertap is available on OSTI at https://www.osti.gov/biblio/1785311 and provides a DOI. But the version is given as 1.0 whereas the current version for the repository is 0.6. Can someone correct the version number on OSTI?

PR #840, opened 2022-11-10
Add reference text and citation file

wigging:
## Fixes/Resolves:

Resolves issue #832 by providing reference text for the project.

## Summary/Motivation:

Adds a citation file which is supported by GitHub to reference software. Text that can be used to reference the WaterTAP project is also added to the README and documentation.

- Add a citation file
- Add reference text to README and documentation

### Legal Acknowledgement


wigging:
See [here](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files) for an overview of CITATION files on GitHub.

adam-a-a (approved):
LGTM!

lbianchi-lbl (changes requested):
I think this is a fine idea, but I'm concerned about keeping the list of authors up-to-date in the long run if individual names are listed. For the same reason, I also see the fact that the same list is repeated twice in non-machine-readable files (in `README.md` and `docs/index.rst`), which would have to be manually synced to the `CITATION.cff`.

As a "cheap" way to bypass both problems, if there are no strong reasons or preferences against it, I suggest we consider using a more general `WaterTAP contributors` (using the appropriate schema for the format) instead of listing the names.

wigging:
@lbianchi-lbl See latest commit.

lbianchi-lbl (approved):
Looks good, thanks!

PR #841, opened 2022-11-10
OARO Modeling & Initialization

bknueven:
## Summary/Motivation:
Updates to OARO model to facilitate its use in flowsheets.

- remove redundant constraint on temperature and change how the isothermal assumption is modeled
- rewrite OARO initialization and update membrane channel initialization for permeate side
- update OARO tests
- add logging information to RO tests

### Legal Acknowledgement

### DIFF
--- a/watertap/unit_models/osmotically_assisted_reverse_osmosis_base.py
+++ b/watertap/unit_models/osmotically_assisted_reverse_osmosis_base.py
@@ -568,23 +568,21 @@ class OsmoticallyAssistedReverseOsmosisBaseData(
         # Create solver
         opt = get_solver(solver, optarg)
 
-        # Solve unit *without* flux equation
-        self.eq_flux_mass.deactivate()
+        # Solve unit
         with idaeslog.solver_log(solve_log, idaeslog.DEBUG) as slc:
             res = opt.solve(self, tee=slc.tee)
+            # occasionally it might be worth retrying a solve
+            if not check_optimal_termination(res):
+                init_log.warning(
+                    f"Trouble solving unit model {self.name}, trying one more time"
+                )
+                res = opt.solve(self, tee=slc.tee)
         init_log.info_high(f"Initialization Step 2 {idaeslog.condition(res)}")
-
-        # Solve unit *with* flux equation
-        self.eq_flux_mass.activate()
-        with idaeslog.solver_log(solve_log, idaeslog.DEBUG) as slc:
-            res = opt.solve(self, tee=slc.tee)
-        init_log.info_high(f"Initialization Step 3 {idaeslog.condition(res)}")
 
         # release inlet state, in case this error is caught
         self.permeate_side.release_state(permeate_flags, outlvl)
         self.feed_side.release_state(feed_flags, outlvl)
         self.eq_permeate_isothermal.activate()
-        print("DOF after release state", degrees_of_freedom(self))
 
         init_log.info(f"Initialization Complete: {idaeslog.condition(res)}")
 
--- a/watertap/unit_models/osmotically_assisted_reverse_osmosis_base.py
+++ b/watertap/unit_models/osmotically_assisted_reverse_osmosis_base.py
@@ -116,7 +116,7 @@ class OsmoticallyAssistedReverseOsmosisBaseData(
         )
 
         # Add constraint for equal temperature between bulk and interface
-        self.feed_side.add_isothermal_conditions()
+        self.feed_side.add_interface_isothermal_conditions()
 
         # Add constraint for volumetric flow equality between interface and bulk
         self.feed_side.add_extensive_flow_to_interface()
@@ -157,7 +157,7 @@ class OsmoticallyAssistedReverseOsmosisBaseData(
         )
 
         # Add constraint for equal temperature between bulk and interface
-        self.permeate_side.add_isothermal_conditions()
+        self.permeate_side.add_interface_isothermal_conditions()
 
         # Add constraint for volumetric flow equality between interface and bulk
         self.permeate_side.add_extensive_flow_to_interface()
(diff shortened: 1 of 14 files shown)


watertap/unit_models/osmotically_assisted_reverse_osmosis_base.py
```diff
@@ -193,9 +195,18 @@ def build(self):
...
+                b.permeate_side.properties[t, x].temperature
+                == 0.5
+                * b.feed_side.properties[
+                    t, b.feed_side.length_domain.first()
+                ].temperature
+                + 0.5
+                * b.permeate_side.properties[
+                    t, b.permeate_side.length_domain.last()
+                ].temperature
             )
```
luohezhiming, line 210: For a 1D OARO model, we have temperature at `permeate_in` set as our value, and temperature at other place set as an average temperature of feed inlet and permeate inlet. I know it is our decision to make this dummy temperature model, and actually we always have to set equivalent temperature for solving. And right now we don't have the 1D OARO model imported, so it is OK for now. But maybe for future work, we could at least assign a temperature value based on the discrete point for the upcoming 1D model.

watertap/unit_models/osmotically_assisted_reverse_osmosis_base.py
```diff
@@ -268,8 +285,12 @@ def eq_recovery_mass_phase_comp(b, t, j):
...
             return b.rejection_phase_comp[t, "Liq", j] == 1 - (
-                b.permeate_side.properties_out[t].conc_mass_phase_comp["Liq", j]
-                / b.feed_side.properties_in[t].conc_mass_phase_comp["Liq", j]
+                b.permeate_side.properties[t, b.first_element].conc_mass_phase_comp[
+                    "Liq", j
+                ]
+                / b.feed_side.properties[t, b.first_element].conc_mass_phase_comp[
+                    "Liq", j
+                ]
             )
```
luohezhiming, line 294: I just want to confirm do we still have `propterties_in` and `properties_out` for` permeate_side`? Not so clear why we want to use `propterties` instead.

watertap/unit_models/osmotically_assisted_reverse_osmosis_base.py
```diff
@@ -507,15 +534,50 @@ def initialize_build(
...
-        source_flags = self.feed_side.initialize(
-            state_args=state_args,
+        init_log = idaeslog.getInitLogger(self.name, outlvl, tag="unit")
+        solve_log = idaeslog.getSolveLogger(self.name, outlvl, tag="unit")
+
+        # The model is isothermal, but the initialize methods
+        # (or the user) will fix the permeate-side temperature.
+        # Therefore, for initialization we need to deactivate
+        # the isothermal constraint but check that it will
+        # be enforced.
```
luohezhiming: I think these are previous comments, our current method does not need to deactivate the constraint, right?

watertap/unit_models/tests/test_osmotically_assisted_reverse_osmosis_0D.py
```diff
@@ -494,17 +493,14 @@ def test_CP_calculation_with_kf_fixed(self):
...
         m.fs.unit.structural_parameter.fix(300e-6)
 
-        perm_flow_mass = 0.5
-        perm_mass_frac_NaCl = 0.005
-        perm_mass_frac_H2O = 1 - perm_mass_frac_NaCl
-
-        m.fs.unit.permeate_outlet.flow_mass_phase_comp[0, "Liq", "NaCl"].fix(
-            perm_flow_mass * perm_mass_frac_NaCl
+        m.fs.unit.permeate_inlet.flow_mass_phase_comp[0, "Liq", "H2O"].fix(
+            0.251513390470481
```
luohezhiming, line 497: Are these values the same as previous concentration setting?

luohezhiming (commented):
It looks good to me. I have a few minor questions.

watertap/unit_models/tests/test_osmotically_assisted_reverse_osmosis_0D.py
bknueven: Yes, just the inlet and outlet are switched.

watertap/unit_models/osmotically_assisted_reverse_osmosis_base.py
bknueven:
```suggestion
```
Right.

bknueven: Removed in 86da522

watertap/unit_models/osmotically_assisted_reverse_osmosis_base.py
bknueven: Yes -- this expression is equivalent to what its replacing in the 0D model. This is the "base" class so we really should be expressing these constraints in the common format for 1D and 0D.

watertap/unit_models/osmotically_assisted_reverse_osmosis_base.py
bknueven: I'm not sure I understand your point -- this constraint as-expressed should work for the 1D model.

luohezhiming: I mean, for the 1D model. The temperature at permeate side apart from inlet is set as a same value (average of feed side and permeate side inlet). It might make more sense if we assign a value it node `i` as the average of `i-1` and `i+1`. I know it does not matter bacause we always have to set the same teperature otherwise we fail the initialization. But probably we could reconsider it later.

luohezhiming (approved):

watertap/unit_models/osmotically_assisted_reverse_osmosis_base.py
bknueven: Hmm maybe. This is an implementation of what @TimBartholomew suggested yesterday -- set all the nodes besides the inlet equal to the average.

I'm not sure it matters either way.

TimBartholomew (approved):

PR #845, opened 2022-11-14
Isothermal constraint

### LINKED ISSUE #649: Isothermal configuration arguments
Almost all models in WaterTAP should have an isothermal configuration argument. Where the energy balance on the control volume is skipped and instead the energy balance is just represented with making the outlet temperatures equal to the inlet temperature.

The only models this won't apply to are processes where temperature change is an inherent part of the process like evaporative processes and some chemical reactions.

bknueven:
## Fixes/Resolves: #649 

- Add custom CV with isothermal assumption
- Unit Models updated:
   - [x] Boron removal
   - [x] Coag and floc
   - [x] Electrodialysis 0D
   - [x] Electrodialysis 1D
   - [x] GAC
   - [x] Nanofiltration ZO
   - [x] Pressure Exchanger

### Legal Acknowledgement


watertap/unit_models/nanofiltration_ZO.py
```diff
@@ -442,15 +464,6 @@ def eq_recovery_mass_phase_comp(b, t, j):
...
 
-        @self.Constraint(
-            self.flowsheet().config.time, doc="Isothermal assumption for permeate"
-        )
-        def eq_permeate_isothermal(b, t):
-            return (
-                b.feed_side.properties_in[t].temperature
-                == b.properties_permeate[t].temperature
-            )
-
```
bknueven: You don't want to remove this one -- it's connecting the feed and permeate sides. This change is just about connected the feed side inlet and outlet temperatures.

watertap/unit_models/nanofiltration_ZO.py
```diff
@@ -678,9 +688,6 @@ def calculate_scaling_factors(self):
             sf = iscale.get_scaling_factor(self.rejection_phase_comp[ind])
             iscale.constraint_scaling_transform(c, sf)
 
-        for t, c in self.eq_permeate_isothermal.items():
-            sf = iscale.get_scaling_factor(self.feed_side.properties_in[t].temperature)
-            iscale.constraint_scaling_transform(c, sf)
```
bknueven: Please retain this (see above).

watertap/unit_models/gac.py
```diff
@@ -244,16 +285,6 @@ def eq_isothermal(b, t):
...
-        @self.Constraint(
-            self.flowsheet().config.time,
-            doc="Isothermal assumption for absorbed contaminant",
-        )
-        def eq_isothermal_adsorbed_contam(b, t):
-            return (
-                b.process_flow.properties_in[t].temperature
-                == b.adsorbed_contam[t].temperature
-            )
-
```
bknueven: This should be retained.

bknueven (commented):
@MarcusHolly there was a few places where the isothermal constraint was crossing the control volume. I've restored those and now I'm confident the test should pass.

watertap/core/control_volume_isothermal.py
```diff
@@ -0,0 +1,95 @@
...
+                b.properties[t, b.length_domain.first()].temperature
+                == b.properties[t, x].temperature
+            )
+
+    def calculate_scaling_factors(self):
+        super().calculate_scaling_factors()
+        if hasattr(self, "isothermal_assumption_eq"):
+            for (t, x), c in self.isothermal_assumption_eq.items():
+                sf = iscale.get_scaling_factor(
+                    self.properties[t, x].temperature, default=1e2, warning=True
```
lbibl: Is the default sf=1e2 for temperature too larger?

MarcusHolly: Oh, yea this should probably be 1e-2, right? Thanks for pointing that out.

bknueven: ~~Temperature is measured in Kelvins, and all our processes have to operate well above 200K. `1e2` seems appropriate?~~

Right, `1e-2`.

watertap/core/control_volume_isothermal.py
```diff
@@ -0,0 +1,95 @@
...
+        )
+        def isothermal_assumption_eq(b, t):
+            return b.properties_out[t].temperature == b.properties_in[t].temperature
+
+    def calculate_scaling_factors(self):
+        super().calculate_scaling_factors()
+        if hasattr(self, "isothermal_assumption_eq"):
+            for t, c in self.isothermal_assumption_eq.items():
+                sf = iscale.get_scaling_factor(
+                    self.properties_in[t].temperature, default=1e2, warning=True
```
bknueven:
```suggestion
                    self.properties_in[t].temperature, default=1e-2, warning=True
```

watertap/core/control_volume_isothermal.py
bknueven:
```suggestion
                    self.properties[t, x].temperature, default=1e-2, warning=True
```

bknueven: Fixed in fb5a25d.

watertap/core/control_volume_isothermal.py
```diff
@@ -0,0 +1,95 @@
...
+            self.length_domain,
+            doc="Isothermal assumption for 1D control volume",
+        )
+        def isothermal_assumption_eq(b, t, x):
+            if x == b.length_domain.first():
+                return Constraint.Skip
+            return (
+                b.properties[t, b.length_domain.first()].temperature
+                == b.properties[t, x].temperature
+            )
```
TimBartholomew, line 86: I know this would be a separate PR, but creating this WaterTAP control volume has me thinking... What if we make a WaterTAP control volume that has a configuration argument for whether it is 0D or 1D and it automatically includes the references for the 0D so that it quacks like a 1D model with `properties[t, x]` instead of just `properties_in[t]` and `properties_out[t]`

watertap/unit_models/boron_removal.py
```diff
@@ -120,23 +120,34 @@ class BoronRemovalData(InitializationMixin, UnitModelBlockData):
...
-    # NOTE: This option is temporarily disabled
-    '''
-    CONFIG.declare("energy_balance_type", ConfigValue(
-        default=EnergyBalanceType.useDefault,
-        domain=In(EnergyBalanceType),
-        description="Energy balance construction flag",
-        doc="""Indicates what type of energy balance should be constructed,
-    **default** - EnergyBalanceType.useDefault.
+    CONFIG.declare(
+        "isothermal",
```
TimBartholomew: I think following IDAES conventions this might be better as `is_isothermal`

TimBartholomew (commented):
2 comments - one just an idea for a separate effort, the other minor naming

watertap/core/control_volume_isothermal.py
andrewlee94: My only caution here is to think through what this means downstream and what confusions this might create. IDAES deliberately kept the two separate and with different APIs so that it was clear to the end-user which was which, even if it meant they had to understand a bit more.

bknueven: I think there's a trade-off here between keeping separate, understandable APIs while also allowing code-sharing between 0D unit models and their 1D equivalent. I would also add that our 0D-model still allow you to access the 0D-type API.

I would be in favor of enabling a 1D-quacking 0DCV, with the caveat that for models for which we only have a 0D version of we stick to the 0D API in the implementation.

lbibl: Question: is "1D-quacking 0DCV" referring to a 1D-cv with 1 finite element?

bknueven: No -- it has (for a flow-forward model) `properties_in[t]` mapped to `properties[0,t]` and `properties_out[t]` mapped to `properties[1,t]` using references.

This allows for the expression of constraints common to both a 0D and 1D version of a model, see, e.g., https://github.com/watertap-org/watertap/blob/d2382933c88859b6e4f6ae7f791ca187f30a3a33/watertap/unit_models/reverse_osmosis_base.py#L346-L372

watertap/unit_models/boron_removal.py
bknueven: Done in 242f87c.

TimBartholomew:
@bknueven the test failure is for the dye desalination case study, I think this came up before in another PR or was it always this PR? Do we have any insight into the issues here?

bknueven:
> @bknueven the test failure is for the dye desalination case study, I think this came up before in another PR or was it always this PR? Do we have any insight into the issues here?

It's been an issue in this PR, and I think it's come up in others before.

I don't have any insight because I cannot reproduce the failure locally, even after creating a fresh conda environment on my Ubuntu VM. I would be happy if anybody else saw the same failure at the Linux GHA runners.

bknueven:
> @bknueven the test failure is for the dye desalination case study, I think this came up before in another PR or was it always this PR? Do we have any insight into the issues here?

@TimBartholomew I managed to track this down. One of the steps in the dye desalination initialization converged for me locally, but only after 2000+ iterations. Fortunately, the fix was fairly simple: 8c8798e.

TimBartholomew (approved):
LGTM

TimBartholomew:
@lbibl can you review/approve?

lbibl (approved):
Good to me.

Issue #650, opened 2022-07-29
Enthalpy within property models needs a pressure term


TimBartholomew:
Currently the enthalpy in the seawater and NaCl property packages does not include pressure effects. A pressure term should be included in the specific enthalpy that is relative to some reference pressure. E.g., https://www.sciencedirect.com/science/article/pii/S0011916416300807

The inclusion of the pressure term might impact our isothermal pump model.

PR #853, opened 2022-11-21
Adds Pressure Effects Into Specific Enthalpy

MarcusHolly:
## Fixes/Resolves:

Issue https://github.com/watertap-org/watertap/issues/650

### Legal Acknowledgement


MarcusHolly:
Ignore the changes to the water prop pack... those were just accidentally carried over from a previous PR.

watertap/property_models/NaCl_prop_pack.py
```diff
@@ -749,26 +824,37 @@ def _enth_mass_phase(self):
...
         def rule_enth_mass_phase(
             b,
-        ):  # specific enthalpy, eq. 55 and 43 in Sharqawy  # TODO: remove enthalpy when all units can be isothermal
+        ):  # specific enthalpy, eq. 25 and 26 in Nayar et al. (2016)  # TODO: remove enthalpy when all units can be isothermal
             t = (
                 b.temperature - 273.15 * pyunits.K
             )  # temperature in degC, but pyunits in K
-            S = b.mass_frac_phase_comp["Liq", "NaCl"]
-            h_w = (
+            S = b.mass_frac_phase_comp["Liq", "TDS"]
```
bknueven:
```suggestion
            S = b.mass_frac_phase_comp["Liq", "NaCl"]
```

Not sure why the index changed here...

watertap/property_models/NaCl_prop_pack.py
```diff
@@ -749,26 +824,37 @@ def _enth_mass_phase(self):
...
         def rule_enth_mass_phase(
             b,
-        ):  # specific enthalpy, eq. 55 and 43 in Sharqawy  # TODO: remove enthalpy when all units can be isothermal
+        ):  # specific enthalpy, eq. 25 and 26 in Nayar et al. (2016)  # TODO: remove enthalpy when all units can be isothermal
             t = (
                 b.temperature - 273.15 * pyunits.K
             )  # temperature in degC, but pyunits in K
             S = b.mass_frac_phase_comp["Liq", "NaCl"]
-            h_w = (
+            P = b.pressure - 0.101 * pyunits.MPa
```
bknueven:
```suggestion
            P = b.pressure - 1.01e+5 * pyunits.Pa
```

watertap/property_models/seawater_prop_pack.py
```diff
@@ -1280,18 +1304,16 @@ def _enth_mass_phase(self):
...
             )  # temperature in degC, but pyunits in K
             S = b.mass_frac_phase_comp["Liq", "TDS"]
-            h_w = (
-                b.params.enth_mass_param_A1
-                + b.params.enth_mass_param_A2 * t
-                + b.params.enth_mass_param_A3 * t**2
-                + b.params.enth_mass_param_A4 * t**3
-            )
-            h_sw = h_w - S * (
+            P = b.pressure - 0.101 * pyunits.MPa
```
bknueven:
```suggestion
            P = b.pressure - 1.01e+5 * pyunits.Pa
```

watertap/property_models/seawater_prop_pack.py
```diff
@@ -1280,18 +1304,16 @@ def _enth_mass_phase(self):
...
             S = b.mass_frac_phase_comp["Liq", "TDS"]
-            h_w = (
-                b.params.enth_mass_param_A1
-                + b.params.enth_mass_param_A2 * t
-                + b.params.enth_mass_param_A3 * t**2
-                + b.params.enth_mass_param_A4 * t**3
-            )
-            h_sw = h_w - S * (
+            P = b.pressure - 1.01e+5 * pyunits.Pa
+            h_w = 141.355 + 4202.07 * t - 0.535 * t**2 + 0.004 * t**3
```
bknueven: Maybe we should rename the `h_w` parameters a `enth_mass_param_C*` instead of hard-coding them?

MarcusHolly:
@TimBartholomew The pressure correction term (everything to the right of (P-Po)) in the following equation is very different than the density of seawater. The density of seawater is 1026 kg/m3, while "1/pressure correction term" comes out to be  0.0011 assuming a temperature of 25C (t = 25) and a `mass_frac_phase_comp["Liq", "TDS"]` of 0.035 (S = 35 ppt).

image

MarcusHolly (commented):
This could just be a misunderstanding on my part, but doesn't this enthalpy equation having temperature in Celsius pose a pretty big issue? I would imagine the relationship doesn't hold true by simply converting temperature to Kelvin. Additionally, it seems like pyomo doesn't support units in Celsius nor have I seen Celsius used anywhere in WaterTAP. This oversight might also explain why the enthalpy value at the end of the test_seawater_prop_pack jumped so much (although most of the other enthalpy values are relatively close to their previous values)
image

andrewlee94:
So long as you replace `t` with `(T-273.15*pyo.units.K)` this will still work, but you do have to do the explicit conversion to Celcius; otherwise, you will indeed see erroneous results.

Also, Pyomo does not support Celcius due to issues with actual and differential temperatures. I.e., with an actual temperature in Celcius `T(K) = T(C) +273.15` but if it is a differential temperature `dT(C) = T1(C0 - T2(C)` then `dT(K) = dT(C)` - there is no way to know from the units alone whether you need the 273.15K conversion factor.

MarcusHolly:
@andrewlee94 Alright, that makes sense. And yes, t is specified as `(T-273.15*pyo.units.K)`.

MarcusHolly:
So far I have implemented the pressure effects into the enthalpy relationships for the NaCl prop pack and the seawater prop pack. The latter appears to be causing a lot of errors tho. I have addressed most, if not all, of the simple assertion errors, but I am still a bit uncertain if these failures are all expected or if there is a mistake/oversight in the implementation that is causing these failures. There appear to be three main types of failures:

1.) Solver is failing due to the termination being infeasible and/or the maxIterations being exceeded. I have been able to resolve a small number of these failures by slightly adjusting scaling in the seawater_prop_pack but am unsure how to go about addressing most of the others.

2.) DOF failures with the test_flowsheets file. This is my first time working with these files and it seems like they're purpose is just to optimize and verify the accuracy of the LCOW at certain system recoveries (except for the NF flowsheets). That being said, I don't even know where to start in terms of addressing all the DOF failures. So hopefully fixing errors upstream will resolve these issues...

3.) Assertion errors with the pressure changer unit model. The issue here is that the enthalpy flow of the inlet is not equal to the enthalpy flow of the outlet. This failure is not surprising since this PR is literally introducing pressure effects into the enthalpy calculations, but I am unsure how to go about addressing these failures. It seems like these failing tests might just need to be scrapped or, at the very least, rewritten to account for the pressure effects.

MarcusHolly:
Update to above comment: Changing the scaling in the seawater prop pack has resolved a majority of issues. The ones remaining now are related to the pressure changer and evaporator unit models. The latter fails to initialize and the former likely needs to have its test reworked for the reason mentioned above.

MarcusHolly:
Not sure what to do about these numpy errors... everything runs locally for me w/o failures.

watertap/property_models/seawater_prop_pack.py
```diff
@@ -1290,28 +1339,47 @@ def _enth_mass_phase(self):
...
+                + b.params.enth_mass_param_B4 * s**3
                 + b.params.enth_mass_param_B5 * t
                 + b.params.enth_mass_param_B6 * t**2
                 + b.params.enth_mass_param_B7 * t**3
-                + b.params.enth_mass_param_B8 * S * t
-                + b.params.enth_mass_param_B9 * S**2 * t
-                + b.params.enth_mass_param_B10 * S * t**2
+                + b.params.enth_mass_param_B8 * s * t
+                + b.params.enth_mass_param_B9 * s**2 * t
+                + b.params.enth_mass_param_B10 * s * t**2
```
bknueven: I think the "s" here is still `kg/kg`, so should stay `S`.

watertap/property_models/seawater_prop_pack.py
```diff
@@ -1290,28 +1339,47 @@ def _enth_mass_phase(self):
...
+            )
+            h_sw = h_sw0 + P_MPa * (
+                b.params.enth_mass_param_A1
+                + b.params.enth_mass_param_A2 * t
+                + b.params.enth_mass_param_A3 * t**2
+                + b.params.enth_mass_param_A4 * t**3
+                + S
```
bknueven: I think this "s" is in `g/kg`, so should be `s` instead.

Might be better to use different names: `S_kg_kg` and `S_g_kg` instead of just "S".

lbianchi-lbl:
> Not sure what to do about these numpy errors... everything runs locally for me w/o failures.

@MarcusHolly I believe these might be due to the recent release of NumPy 1.24, which introduced a number of "expired deprecations" (https://numpy.org/devdocs/release/1.24.0-notes.html#expired-deprecations), including, I guess, removing `np.bool` (which seems to be the cause of the errors in CI). I suspect this is also why you're not getting the same errors locally, if an earlier version of `numpy` is installed in your local environment (you can check by running e.g. `pip show numpy`).

Since this is not directly caused by changes in this PR, you could get the tests to pass by adding a constraint on the numpy requirement in `setup.py` under `install_requires`, i.e. `numpy < 1.24`. The constraint can be removed together with the deprecated code in a separate, dedicated PR.

watertap/property_models/tests/test_seawater_prop_pack.py
```diff
@@ -37,12 +37,13 @@ def configure(self):
         self.scaling_args = {
             ("flow_mass_phase_comp", ("Liq", "H2O")): 1,
             ("flow_mass_phase_comp", ("Liq", "TDS")): 1e2,
+            ("enth_mass_phase", ("Liq")): 1e-4,
```
bknueven: Probably isn't needed anymore?

bknueven (commented):
One final comment

watertap/property_models/tests/test_seawater_prop_pack.py
MarcusHolly: Yea, you're right.

watertap/examples/flowsheets/full_treatment_train/analysis/tests/test_flowsheets.py
```diff
@@ -81,9 +81,9 @@ def test_flowsheet_softening():
 def test_flowsheet_softening_two_stage():
     desal_kwargs = flowsheet_two_stage.desal_kwargs
     m = flowsheet_softening_two_stage.optimize_flowsheet(
-        system_recovery=0.80, **desal_kwargs
+        system_recovery=0.70, **desal_kwargs
     )
-    assert value(m.fs.costing.LCOW) == pytest.approx(0.9277, rel=1e-3)
+    assert value(m.fs.costing.LCOW) == pytest.approx(0.549963, rel=1e-3)
```
bknueven: I lied about only having one more -- is this change also still needed?

MarcusHolly: Nope - thanks for being so thorough with your review!

bknueven (approved):
Looks good!

bknueven:
@TimBartholomew this is ready for your review.

watertap/property_models/seawater_prop_pack.py
```diff
@@ -323,93 +323,142 @@ def build(self):
...
             within=Reals,
-            initialize=7.826e3,
+            initialize=7.82607e3,
             units=enth_mass_units * t_inv_units,
             doc="Specific enthalpy parameter B5",
         )
         self.enth_mass_param_B6 = Var(
             within=Reals,
-            initialize=-4.417e1,
+            initialize=-4.41733,
```
TimBartholomew, line 410: Whoa, so the previous parameter was off by an order of magnitude? It used to be -4.417e1 and now its -4.41733.

watertap/property_models/seawater_prop_pack.py
```diff
@@ -323,93 +323,142 @@ def build(self):
...
             within=Reals,
-            initialize=2.778e4,
+            initialize=2.77846e4,
             units=enth_mass_units * t_inv_units,
             doc="Specific enthalpy parameter B9",
         )
         self.enth_mass_param_B10 = Var(
             within=Reals,
-            initialize=9.728e1,
+            initialize=9.72801,
```
TimBartholomew, line 434: Same thing here. Was the previous parameter off by an order of magnitude? Used to be 9.728e1 and now is 9.2780.

TimBartholomew (commented):
@MarcusHolly did you simulate the properties you modified across a range of conditions and match it with the literature source?

watertap/property_models/seawater_prop_pack.py
MarcusHolly: I haven't checked the literature source for the older values to see if they were input incorrectly, but the values I have here are correct according to the new literature source.
image

watertap/property_models/seawater_prop_pack.py
MarcusHolly: See above comment

MarcusHolly:
> @MarcusHolly did you simulate the properties you modified across a range of conditions and match it with the literature source?

Nope - how should I go about doing this?

TimBartholomew:
@MarcusHolly I believe Lienhard's group publishes figures for each of the properties. So you can recreate the figure by simulating a stateblock and vary things like salinity, temperature, and pressure. You can't vary all three at the same time and make a 2D plot, but you can see what the literature source kept constant and what it varied and the bounds of their plot. Then you can make the same figure and see if they are roughly in alignment. This will mainly catch large mistakes that effect the properties by more than 5-10%.

watertap/property_models/seawater_prop_pack.py
MarcusHolly: These are the parameter values for the old source, so it looks like they were input correctly and these two values (B6 and B10) just differ by an order of magnitude for some reason.
image

MarcusHolly:
> @MarcusHolly I believe Lienhard's group publishes figures for each of the properties. So you can recreate the figure by simulating a stateblock and vary things like salinity, temperature, and pressure. You can't vary all three at the same time and make a 2D plot, but you can see what the literature source kept constant and what it varied and the bounds of their plot. Then you can make the same figure and see if they are roughly in alignment. This will mainly catch large mistakes that effect the properties by more than 5-10%.

Left = Generated from a simulation file; right = literature source
image

I've only graphed half of the points here, but I think it's enough to see that the results are very similar. @TimBartholomew Let me know if a more formal comparison is necessary or if you would like to see the other half of the data points generated.

andrewlee94:
@MarcusHolly Speaking from experience in IDAES, I would recommend a formal test covering all the data points. You can digitize the literature plots to get the data in a more useful form, and then iterate over all the inputs, solve the model and compare to the literature values as an integration test.

MarcusHolly:
@andrewlee94 Alright, that makes sense. Would you happen to have an example of how to digitize the literature plot and how the integration test is set up?

andrewlee94:
@MarcusHolly See these:

https://github.com/watertap-org/watertap/pull/853#issuecomment-1373757206
https://apps.automeris.io/wpd/

TimBartholomew (approved):

PR #860, opened 2022-11-30
Reorganize sets in multicomponent aq solution proper pack 

### LINKED ISSUE #695: Standardize `ion_set` indexing and refine `Set` structure in the DSPMDE property package

lbibl:
## Summary/Motivation:
See Issue #695 

- Reorganized sets as this: `cation_set` and `anion_set`  \in `ion_set`; `ion_set` and `neutral_set` \in `solute_set`; `solute_set` and `solvent_set` in `component_list`.  This is actually consistent to the `components` module in IDAES as far as I see. 
  

### Legal Acknowledgement


watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -277,37 +277,44 @@ def build(self):
         # components
         self.H2O = Solvent()
 
-        # blank sets
-        self.cation_set = Set()
-        self.anion_set = Set()
-        self.solute_set = Set()
-        self.ion_set = Set()
```
adam-a-a, line 284: @lbibl Can you remind me why `ion_set` was excluded? I think it's because `ion_set` is already accounted for via IDAES, but I'm forgetting. I only ask because I anticipate that resolving the failure related to the NF DSPMDE model will require both an ion_set and solute_set. It could be easily handled without any modification here, but it would be convenient to just have it defined here. 

And if it is still the case the `ion_set` is already generated via IDAES, then I might suggest adding a comment here saying so as a reminder.

lbibl: Good catch.  The reason is that in IDAES the `Anion` and `Cation` class inherit `Ion` class.  It is current coded in the way that when you put an attribute in `Anion` (or `Cation`), you have to take it out from the `Ion` class, i.e., you cannot have a component in both `Anion` and `Ion`.   Because of this, IDAES documented that "Users are encouraged to use the Anion and Cation classes instead of the generic Ion class, as these validate that sign of the charge configuration argument." (https://idaes-pse.readthedocs.io/en/stable/reference_guides/core/comp.html#ion-class).  This design could be improved from IDAES but this is what is currently.   I still added an ion_set so we have something to call `anion_set|cation_set`.  This is the only way to do this as far as I see.

lbibl: `solute_set` in this PR is equivalent to `ion_set|solute_set` previously.  The previous `solute_set` is equivalent to `neutral_set` in this PR.

lbibl:
@adam-a-a : a challenge for me to debug this failure is that the error does not reproduce on my local  Mac and Linux env- the `nano filtration_DSPMDE`  passes the test like a boss there.  If you look into the reported failure by GH: https://github.com/watertap-org/watertap/actions/runs/3650225701/jobs/6165944727#step:11:5728 , the failure refers to ipopt output "internal_error", and ipopt doc says "An unknown internal error has occurred. Please notify the authors of Ipopt via the mailing list." for this output, which suggest it being an error out of immediate explanation.   

I wonder if you or someone else can reproduce this error locally.   @bknueven , I again would like to draw your curiosity to this solver error :)

This PR is not holding off other works for the time being so no urgency here in my view.

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
```diff
@@ -1121,11 +1121,8 @@ def test_parameter_block_comparison(model4):
 
     assert isinstance(m_ion.fs.properties.solute_set, Set)
     assert isinstance(m_generic.fs.properties.solute_set, Set)
-    assert len(m_ion.fs.properties.solute_set) == len(
-        m_generic.fs.properties.solute_set
-    )
```
adam-a-a, line 1126: Ah, the intention of this test was to verify compatibility between this prop package and the (formerly known as) generic property framework in IDAES. I am guessing you had to remove this since you redefined how the sets are organized, and so things might not line up anymore with the generic property framework.

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
```diff
@@ -1121,11 +1121,8 @@ def test_parameter_block_comparison(model4):
...
     for j in m_ion.fs.properties.solute_set:
-        assert j in ["E"]
+        assert j in ["A", "B", "C", "D", "E"]
```
adam-a-a, line 1142: I think you should add tests to verify your redefinition of the sets.

adam-a-a (changes requested):
Although there's an issue to be resolved with the DSPM-DE model and this PR, I also think you could add some tests to check the sets that you defined. 

I also noticed that this prop package will no longer be interchangeable with the Generic Property framework in IDAES, which was something that @aladshaw3 added tests to ensure. For example, I don't think IDAES' definition of solute_set matches with your redefined solute_set. Just something to be aware of and think about

adam-a-a:
I see you mentioned the correction of a scaling method glitch, but I couldn't find that in your changes.

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
lbibl: That is correct.  My key points on this are (1) I didn't see the point that we want a property package interchangeable to an existing one, (2) idaes's generic p.p. does not use and is inconsistent to its `component.py` module, where `Solute` is the parent class of `Ion`, (3) it's a common knowledge that the solutes of a solution can embrace ionic species if they are solvated.  Therefore, I am in favor of making this change, although I will also be ok with not making the change and disregarding this PR.

lbibl:
> I see you mentioned the correction of a scaling method glitch, but I couldn't find that in your changes.

This has been a while.  I vaguely recall I caught small mistype in a scaling method, something like a mis-index or an inaccurate variable use.  I suspect I carried that fix in another PR so it is not reflected here.   I will delete that description .

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
lbibl: L1107-1125 now test all indexing to be correct.

lbibl:
> Although there's an issue to be resolved with the DSPM-DE model and this PR, I also think you could add some tests to check the sets that you defined.
> 
> I also noticed that this prop package will no longer be interchangeable with the Generic Property framework in IDAES, which was something that @aladshaw3 added tests to ensure. For example, I don't think IDAES' definition of solute_set matches with your redefined solute_set. Just something to be aware of and think about

Responded at L1126 and 1125.

aladshaw3:
@lbibl @adam-a-a   

**Regarding the interoperability of the generic prop package and this prop package:**

The main driving-force behind my initial effort there was to ensure that the Electrolyte Database in MongoDB could be used for auto-generation of the necessary packages and configurations for the chemical-process (and other related) unit models. However, if we are moving to a paradigm where "most" unit models are associated with their own property package, then it may not be as critical to maintain that interoperability.

This may also depend on how frequently folks are actually using the Electrolyte Database. If it is not really being used and maintained, then it may not matter. You could just use the Electrolyte Database for the "simplest" chemical processes (such as pH control, precipitation/dissolution, recarbonization, etc.) and use these custom property packages for both physical and chemical processes (such as the electrodialysis and NF DSPMDE model).

lbibl:
This PR now passes all tests and ready for final reviews.  More notes follow in the related issue(s) (#916, #695,  #860 ).

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -272,41 +272,49 @@ def build(self):
...
-                        str(j),
+                        j,
                         Anion(charge=self.config.charge[j], _electrolyte=True),
                     )
-                    self.component_list.add(str(j))
-                    self.ion_set.add(str(j))
+                    self.ion_set.add(j)
             else:
-                self.add_component(str(j), Solute())
+                self.neutral_set.add(j)
```
bknueven: This logic can be expressed without calling `del_component`.

lbibl: @bknueven , logically I thought the same as you did, but it turned out I had to use the `del_component` to first take the `j` off the Solute() class or the `solute_set` and then add it into the `Cation()` or `Anion()` class.  I think this is a control by Pyomo block and here all sets are children of the same block.  I did deleted a useless `del_component` line and added a test for an error message in the recent push.

hunterbarber:
@lbibl To confirm, the only changes in this PR are reflected in the green line?

image

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -272,41 +272,48 @@ def build(self):
...
+        self.neutral_set = Set(dimen=1)  # Components with charge =0
+        self.ion_set = Set(dimen=1)  # All Ion Components (cations + anions)
 
+        # Group components into different sets
         for j in self.config.solute_list:
+            if j == "H2O":
+                raise ConfigurationError(
+                    "'H2O'is reserved as the default solvent and cannot be a solute."
+                )
+            self.add_component(j, Solute())
```
bknueven, line 295:
```suggestion
```

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -272,41 +272,48 @@ def build(self):
...
             if j in self.config.charge:
                 if self.config.charge[j] == 0:
                     raise ConfigurationError(
                         "The charge property should not be assigned to the neutral component: {}".format(
                             j
                         )
                     )
-                elif self.config.charge[j] > 0:
+                if self.config.charge[j] > 0:
+                    self.del_component(j)
```
bknueven, line 307:
```suggestion
```

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -272,41 +272,48 @@ def build(self):
...
                     self.add_component(
-                        str(j),
+                        j,
                         Cation(charge=self.config.charge[j], _electrolyte=True),
                     )
-                    self.component_list.add(str(j))
-                    self.ion_set.add(str(j))
+                    self.ion_set.add(j)
                 else:
+                    self.del_component(j)
```
bknueven, line 315:
```suggestion
```

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -272,41 +272,48 @@ def build(self):
...
-                        str(j),
+                        j,
                         Anion(charge=self.config.charge[j], _electrolyte=True),
                     )
-                    self.component_list.add(str(j))
-                    self.ion_set.add(str(j))
+                    self.ion_set.add(j)
             else:
-                self.add_component(str(j), Solute())
+                self.neutral_set.add(j)
```
bknueven, line 322:
```suggestion
                self.add_component(j, Solute())
                self.neutral_set.add(j)
```

lbibl:
> @lbibl To confirm, the only changes in this PR are reflected in the green line?
> 
> image

Yes.  And the current main version does not have the neutral_set and places solute_set in the place of neutral in this diagram.  Nice picture and maybe we should put a diagram like this in a subsequent doc update.

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -272,41 +272,48 @@ def build(self):
...
+            self.add_component(j, Solute())
             if j in self.config.charge:
                 if self.config.charge[j] == 0:
                     raise ConfigurationError(
                         "The charge property should not be assigned to the neutral component: {}".format(
                             j
                         )
                     )
-                elif self.config.charge[j] > 0:
+                if self.config.charge[j] > 0:
```
bknueven, line 303:
```suggestion
                elif self.config.charge[j] > 0:
```
The logic was clearer with this as an elif. Alternatively, we might have `self.config.charge[j] > 0`, `self.config.charge[j] < 0` else raise the configuration error.

bknueven (commented):
@lbibl I'm a bit confused by your earlier comment. Why can we not apply my suggestions from this review?

watertap/property_models/multicomp_aq_sol_prop_pack.py
lbibl: @bknueven , I double-checked this way and it doesn't make it work (actually having a fail in building the component_list).  I think the trick is in this file: https://github.com/IDAES/idaes-pse/blob/main/idaes/core/base/components.py . I see two reasons for me having to add j in `Solute()` before going to Cation and Anion: 1) this can trigger it being added into the `component_list` by calling a `_add_to_component_list` function, 2) this can let `Anion()` and `Cation()` also registered in `Solute()`.  The current main version actually uses the logic you wrote but it exclude the ions out of `Solute()`.  Let me know if you find I mistake anything.  Thanks. 

After all these work spent on this simple thing, I might feel ok to say this could be an example of using an IDAES base made things more complicated.

bknueven: @andrewlee94 are we doing something wrong here? It seems strange that the API would require us to add a component just to delete it, in order to get some side effect.

Even if we're not using the IDAES API as intended, I wonder if we should implement a different work-around. My preference would be to directly code-in what we want to happen as opposed to relying on the side effect of `add_component` followed by calling `del_component`, which to be just looks like a silly refactor away.

@lbibl if we do decide to leave this as-is, it needs a comment in the code similar to what you say in your comment here.

lbibl: Commented notes for this are added.  

One obvious profit of using IDAES components is to have the `is_solute()` fun working for these sets, although `is_solute()` can be simply replace by `in solute_set` wherever needed. 

At this time I would recommend as-is because 1) the "silly refactor" doesn't really cost anything and 2) of saving time.

lbibl: Another (best in my view) resolution is to make changes in https://github.com/IDAES/idaes-pse/blob/main/idaes/core/base/components.py  to let `Cation()` and `Anion()`, and maybe `Ion()` too, possess methods to add their members into the corresponding sets of their parent classes  (Solute for all, Ion for the other two).  This could be @andrewlee94 's call.

bknueven: > At this time I would recommend as-is because 1) the "silly refactor" doesn't really cost anything and 2) of saving time.

I wasn't being clear: when I look at this code my first thought is, "why do we add a component only to delete it"? So we need comments (which you've added) to explain that. Otherwise somebody else will come along with the same thought in this comment and it won't work.

I don't really like things as they are, but if we're using IDAES correctly I can live with it.

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -272,41 +272,51 @@ def build(self):
...
+        # Group components into different sets
         for j in self.config.solute_list:
+            if j == "H2O":
+                raise ConfigurationError(
+                    "'H2O'is reserved as the default solvent and cannot be a solute."
+                )
+            self.add_component(j, Solute())
+            # This triggers the addition of j into component_list and solute_set
+            # by idaes.core.base.components, although a "del_component" for j will
+            # be called if it goes to cation_ or anion_set (below).
```
bknueven:
```suggestion
            # This triggers the addition of j into component_list and solute_set
            # by idaes.core.base.components, although a "del_component" for j will
            # be called if it goes to cation_ or anion_set (below).
            self.add_component(j, Solute())
```

I feel like it's convention to add the comment just before the line called ...

lbibl: Changed the format and also reworded the note for better clarity.

watertap/property_models/multicomp_aq_sol_prop_pack.py
andrewlee94: @bknueven @lbibl As far as I can tell (and I am having a little trouble following everything here), the issue is that you are using the IDAES `Component` objects, but don't agree with/want their definition of the component sets. So in short, if you don't want the IDAES set definitions you should not be using the IDAES `Component` objects.

Of course, this would mean you lose the functionality of the `Component` objects (or would need to replicate this), but that is the trade off you would need in this case.

bknueven: @lbibl is the definition of component the same throughout WaterTAP, and hence divergent from the IDAES definitions?

andrewlee94: @bknueven I am fairly certain most of the WaterTap property packages use the standard IDAES `Components`, and thus component sets - for one the eNRTL implementation uses the IDAES modular properties (which is what the `Components` were originally written for).

lbibl: > @bknueven @lbibl As far as I can tell (and I am having a little trouble following everything here), the issue is that you are using the IDAES `Component` objects, but don't agree with/want their definition of the component sets. So in short, if you don't want the IDAES set definitions you should not be using the IDAES `Component` objects.
> 
> Of course, this would mean you lose the functionality of the `Component` objects (or would need to replicate this), but that is the trade off you would need in this case.

If we put this PR aside for a moment,  I saw an internal inconsistency in IDAES's components.py.  `Cation` and `Anion` there inherited `Solute` and running `is_solute()` on Cation or Anion member returns True.  But the two subclasses did not add their members in the `solute_set` nor the component_list .

lbibl: > @lbibl is the definition of component the same throughout WaterTAP, and hence divergent from the IDAES definitions?

I didn't originally write the original version of this property package, neither the rest of property packages in Watertap.  I can only answer this PR does not change the definition of component from its original version of this property package.  I thought it's consistent (I could be wrong) to IDAES, because the components.py has been used from the first in this specific property package,  but always with some band-aid to walk around the inconsistency I described above.

andrewlee94: @lbibl The problem is that your definition is fundamentally inconsistent with the one that IDAES uses - that is why there is an issue. It should be noted the "solute_set" in IDAES was always intended to be "non-ionic solutes" but we could never think of a good name for it - the reason that ions are not added to `solute_set` is so that we can clearly identify sub-types.

Further, the ionic `Components` were only intended to be used in limited cases where the property package would decide what to do with them (hence why there are not added to `solute_set` or `component_list`). Basically, they were not designed to be used the way you are (for better or worse).

lbibl: Hi @andrewlee94 , thanks for your clarification.  I heard all the message.  Just to repeat one thing (to make sure you didn't miss):  running `is_solute()` on `Ion`, `Cation`  and `Anion` returns `True` although they are not in `solute_set`.   I think this could cause confusion or mistakes when using this file.

andrewlee94: Yes - that is correct, and is because of the naming of `solute_set` - as I said it really should have a different name to indicate that it is "non-ionic solutes".

lbibl: To make my conclusion over this discussion, 1) I myself recommend keeping this PR as-is because it functions as we planned; 2) with all information shared,  I'll follow reviewers' recommendation/decision on alternative fate of this PR .

bknueven: I'm not opposed to adopting our own definition of these sets in principle, if it's what makes sense for WaterTAP. I would also be happy to mock up our own components which deviate from those in IDAES.

I do think that having different definitions is different property packages will ultimately lead to confusion. WaterTAP should at least be internally consistent, and ideally consistent with IDAES.

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
```diff
@@ -104,9 +104,22 @@ def test_parameter_block(model):
     assert isinstance(model.fs.properties.solvent_set, Set)
     for j in model.fs.properties.solvent_set:
         assert j in ["H2O"]
+    # solute set and ion set are identical when all components in component_list are ions given hierarchy
+    assert isinstance(model.fs.properties.solute_set, Set)
+    for j in model.fs.properties.solute_set:
+        assert j in ["A", "B", "C", "D"]
     assert isinstance(model.fs.properties.ion_set, Set)
-    for j in model.fs.properties.ion_set | model.fs.properties.solute_set:
```
hunterbarber, line 108: @lbibl I think this specific testing was outdated with new set hierarchy, although it still passed because `ion_set == solute_set` in this specific example, meaning the union of the two was still passing. Double check these additional tests for components being organized by the appropriate set when not explicitly specified in `type` argument.

adam-a-a (approved):
LGTM - glad you found the issue.

watertap/property_models/multicomp_aq_sol_prop_pack.py
adam-a-a: I approved the PR but I do share @bknueven's concerns

adam-a-a: It seems to me that a corresponding modification should be considered in IDAES' generic property framework (forgetting what the newest name for it is) to maintain consistency where possible, if it wouldn't break a million things in IDAES. For instance, renaming solute_set as neutral_set and then defining solute_set to include both ions and neutral solutes--maybe @lbibl can post an issue in IDAES regarding that. Thoughts?

andrewlee94: I am not sure how easy it would be to change things on the IDAES side, plus IDAES will not have the resources to do this anytime soon (but it is open-source so WaterTap can always propose a fix if it is important enough).

However, "neutral_set" is probably not a great name, as that could include solvents as well (which are generally neutral). Also, simpler models may wish to declare ions as general solutes in case where they don't need more complex aqueous phase behaviours.

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
lbibl: Thanks @hunterbarber . I think this is fine to change and keep as is.  Test on all sets were covered by the module starting L1090 to ensure all sets are functioning correctly.

hunterbarber (approved):
@lbibl I would say create an issue or note regarding the need for WaterTAP components or a comparable change discussed previously if you think that is something that should still be addressed. But in its current state, looks good.

watertap/property_models/multicomp_aq_sol_prop_pack.py
lbibl: My honest thought is time has been overly used on a simple matter.  With all considered, my recommendation at this time is to either keep things as they are, (as far as I see, this PR does not cause inconsistency in WaterTap but only make things clearer and easier) , or write a WaterTap Component file (which is fairly short) to be consistent with this PR.

lbibl: I don't think the key is about naming- but putting ions in solute in components. I don't see a valid reason to let ions give `Is_solute() == True` but `in solute_set == False`.

PR #861, opened 2022-12-01
Additional limiting current density methods for ED_1D

lbibl:
- Adds empirical and theoretical methods to calculate limiting current density. 
- Better scaling methods. 

### Legal Acknowledgement


watertap/unit_models/electrodialysis_1D.py
```diff
@@ -1703,14 +1770,23 @@ def _get_fluid_dimensionless_quantities(self):
         )
         self.hydraulic_diameter = Var(initialize=1e-3, units=pyunits.meter)
         self.N_Re = Var(
-            initialize=100, units=pyunits.dimensionless, doc="Reynolds Number"
-        )
+            initialize=50,
+            bounds=(0, 2000),
```
TimBartholomew: Sometimes RO modules can operate at Reynolds numbers around 2000, I'm not sure about ED but I would think it could be the same. Generally, this arbitrary upper bound would be looser (i.e. a larger value) just so this doesn't get hit and make the model infeasible. Very minor comment though

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -1789,18 +1865,24 @@ def _pressure_drop_calculation(self):
...
-            self._get_fluid_dimensionless_quantities()
+            if not (
+                self.config.has_Nernst_diffusion_layer
+                and self.config.limiting_current_density_method
+                == LimitingCurrentDensityMethod.Theoretical
+            ):
+                self._get_fluid_dimensionless_quantities()
+            self.friction_factor = Var(
+                initialize=10,
+                bounds=(0, 10000),  # arbitrary upper bound
```
TimBartholomew: Alternatively, you could use `None` instead of an arbitrary upper bound value

TimBartholomew (approved):
LGTM, although I don't have the expertise to know if the limiting current density calculations are correct or are the right approach. I just had a minor comment on the "arbitrary upper bounds"

watertap/unit_models/electrodialysis_1D.py
lbibl: Made the upper bound two orders of magnitude larger for now.  Will keep an eye on this when doing optimization.

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -1787,18 +1863,24 @@ def _pressure_drop_calculation(self):
...
                 "Do not forget to FIX the experimental pressure drop value in [Pa/m]!"
             )
         else:  # PressureDropMethod.Darcy_Weisbach is used
-            self._get_fluid_dimensionless_quantities()
+            if not (
+                self.config.has_Nernst_diffusion_layer
+                and self.config.limiting_current_density_method
+                == LimitingCurrentDensityMethod.Theoretical
+            ):
+                self._get_fluid_dimensionless_quantities()
```
bknueven, line 1877: I'm surprised `black` doesn't force you to add a new line here -- might be good for readability.

bknueven (approved):
I would prefer we remove unneeded upper bounds.

watertap/unit_models/electrodialysis_1D.py
lbibl: Removed upper bounds and reformatted here as suggested.

lbianchi-lbl:
@lbibl I've tried re-building the build from the RTD control panel. This works with the same commit hash, but the specific build associated with the PR check is not working for some reason. Can you try to create and push an empty commit?

```sh
git commit --allow-empty -m "Empty commit to trigger CI"
git push
```

lbibl:
> @lbibl I've tried re-building the build from the RTD control panel. This works with the same commit hash, but the specific build associated with the PR check is not working for some reason. Can you try to create and push an empty commit?
> 
> ```shell
> git commit --allow-empty -m "Empty commit to trigger CI"
> git push
> ```

It looks still failing.   The same failure is in the other PR of mine too.  #870

lbianchi-lbl:
@lbibl I'm not sure what's going on but it looks like it might be a malfunctioning in GitHub and/or RTD. I'll temporarily disable the requirement for the RTD build to pass so that the PRs can be merged.

PR #872, opened 2022-12-14
Updating Sherwood Number Calculation

### LINKED ISSUE #867: Support additional Sherwood number calculation
### Description

Add recently published relationship for sherwood number calculation: https://doi.org/10.1016/j.seppur.2022.122121

### Motivation

NAWI funded a project to analyze concentration polarization and they found that the sherwood number calculation we use from Guillen and Hoek (2009) can be pretty inaccurate. This NAWI project came up with another correlation and it would be good to add it as an option, or potentially replace Guillen and Hoek's relationship.

### Possible Implementation

- Replace the Sherwood number calculation in the membrane channel class
- Or add a configuration argument where users can select which relationship they want to use

### Additional Context

This new Sherwood number calculation includes another variable (distance from the RO module entrance), it may make the model harder to solve. We should assess the stability of this formulation.

MarcusHolly:
## Fixes/Resolves:

Issue #867 

## Summary/Motivation:
Replace the Guillen and Hoek (2009) sherwood number relationship with NAWI's new relationship in  https://doi.org/10.1016/j.seppur.2022.122121

### Legal Acknowledgement


MarcusHolly:
There is a failure with the lssro paper analysis... I believe that one of the following is the issue:

1.) My implementation of the config arguments for the sherwood_number_eq in the membrane_channel_base is not correct

2.) My implementation in the membrane_channel_base is correct, but I'm not actually applying the old sherwood equation to the lssro file, so the test_lssro_paper_analysis is failing because the new sherwood equation is causing the degree of freedom check to fail (would need to specify channel_length, which is the new variable added to the sherwood equation)

MarcusHolly:
Two issues with this PR:

1.) One of the tests are failing only for linux, and I'm not sure why the linux values are slightly offset from the windows values. The solution is probably to adjust scaling somewhere...

2.) I don't see why so many values are changing in the lssro paper analysis baselines since these files use the old sherwood number relationship, so I would assume that nothing should change. This might be because I needed to specify the channel length for the RO unit model in the lssro.py, but this variable is only used to compute the new sherwood number relationship and should have little/no effect on the other results. Alternatively, it might be the case that the new sherwood number relationship is being applied behind the scenes rather than the old one...

watertap/core/membrane_channel_base.py
```diff
@@ -580,6 +627,14 @@ def _add_calculated_pressure_change_mass_transfer_components(self):
...
         )
 
+        self.channel_length = Var(
+            initialize=1,
+            bounds=(1e-2, 1e2),
+            domain=NonNegativeReals,
+            units=units_meta("length"),
+            doc="membrane-channel length",
+        )
+
```
bknueven: We already define `length` -- this seems redundant.

watertap/core/membrane_channel_base.py
```diff
@@ -74,6 +74,16 @@ class PressureChangeType(Enum):
...
 
 
+class SherwoodNumberEq(Enum):
+    """
+    new: Sherwood number relationship depicted in https://doi.org/10.1016/j.seppur.2022.122121v
+    old: Sherwood number relationship depicted in Guillen and Hoek (2009)
+    """
+
+    new = auto()
+    old = auto()
```
bknueven: We should think about better names.

watertap/core/membrane_channel_base.py
```diff
@@ -74,6 +74,16 @@ class PressureChangeType(Enum):
     calculated = auto()
 
 
+class SherwoodNumberEq(Enum):
+    """
+    new: Sherwood number relationship depicted in https://doi.org/10.1016/j.seppur.2022.122121v
```
bknueven: Maybe say eq. (18) here?

watertap/core/membrane_channel_base.py
```diff
@@ -519,17 +551,32 @@ def eq_K(b, t, x, j):
...
+            @self.Constraint(
+                self.flowsheet().config.time,
+                self.length_domain,
+                self.config.property_package.solute_set,
+                doc="Sherwood number",
             )
+            def eq_N_Sh_comp(b, t, x, j):
+                return b.N_Sh_comp[t, x, j] == 2.401 * (
+                    (b.N_Re[t, x] * b.N_Sc_comp[t, x, j]) ** 0.297
+                ) * ((b.channel_length / b.dh) ** -0.279)
```
bknueven: Reading the original source, it seems like _x_ is the current position *in* the channel. So `length` should be multiplied by `x` here, I think. Probably also need to consider the backwards case carefully.

@adam-a-a thoughts?

watertap/core/membrane_channel_base.py
```diff
@@ -519,17 +551,32 @@ def eq_K(b, t, x, j):
...
-            self.config.property_package.solute_set,
-            doc="Sherwood number",
-        )
-        def eq_N_Sh_comp(b, t, x, j):
-            return (
-                b.N_Sh_comp[t, x, j]
-                == 0.46 * (b.N_Re[t, x] * b.N_Sc_comp[t, x, j]) ** 0.36
+        if sherwood_number_eq == SherwoodNumberEq.new:
+
+            @self.Constraint(
```
bknueven, line 576: I'm a bit baffled that this doesn't show as covered. Let me try merging in `main`...

bknueven (changes requested):
I have a few comments. Mainly:
1. I think we're defining `length` twice for the channel. Maybe in a separate PR we should rename `channel_height` to just `height` for consistency
2. `new` and `old` are non-descriptive names. We should find better ones
3. I think we need to consider `x` in the calculation. If the direction of flow is "backwards" it may need to be substituted for `(1-x)`

I'd like @adam-a-a to weigh in for domain expertise.

watertap/core/membrane_channel_base.py
bknueven: We need to pass the `sherwood_number_eq` option into this function from the unit model. That's why this line is not covered.

watertap/core/membrane_channel_base.py
```diff
@@ -449,8 +485,17 @@ def eq_cp_modulus(b, t, x, j):
...
+                SherwoodNumberEq.length_dependent,
+                SherwoodNumberEq.length_independent,
+            ):
+                raise ConfigurationError(
+                    f"Unrecognized sherwood_number_eq {sherwood_number_eq}"
+                )
+
             if mass_transfer_coefficient == MassTransferCoefficient.calculated:
                 self._add_calculated_mass_transfer_coefficient()
+                self._add_sherwood_number()
```
MarcusHolly: @bknueven Do I need to pass in the sherwood number like this or can I just set it like `sherwood_number_eq=SherwoodNumberEq.length_independent` and get rid of the _add_sherwood_number function?

Either way, is it still necessary to keep the `sherwood_number_eq=SherwoodNumberEq.length_independent` in the add_calculated_mass_transfer_coefficient (line 527)?

bknueven: You don't need to have an `_add_sherwood_number` method. You just need to pass the value of `sherwood_number_eq` into `_add_calculated_mass_transfer_coefficient` and add the appropriate sherwood number equation there (like you were doing before).

watertap/unit_models/tests/test_osmotically_assisted_reverse_osmosis_0D.py
```diff
@@ -56,6 +56,7 @@
     MassTransferCoefficient,
     PressureChangeType,
 )
+from watertap.core.membrane_channel_base import SherwoodNumberEq
```
bknueven: Should we add the `SherwoodNumberEq` to watertap/core/\_\_init\_\_.py (like `PressureChangeType`)?

watertap/core/membrane_channel_base.py
```diff
@@ -519,16 +562,36 @@ def eq_K(b, t, x, j):
...
+            @self.Constraint(
+                self.flowsheet().config.time,
+                self.length_domain,
+                self.config.property_package.solute_set,
+                doc="Sherwood number",
+            )
+            def eq_N_Sh_comp_length_dependent(b, t, x, j):
+                return b.N_Sh_comp[t, x, j] == 2.401 * (
+                    (b.N_Re[t, x] * b.N_Sc_comp[t, x, j]) ** 0.297
+                ) * ((b.length / b.dh) ** -0.279)
```
bknueven: I still have the same concern as before: https://github.com/watertap-org/watertap/pull/872#discussion_r1082782345.

Here is the equation in the reference (https://www.sciencedirect.com/science/article/pii/S1383586622016768)
Screenshot 2023-02-01 at 8 43 19 AM

In the Nomenclature section the channel length is represented by _L_. So I think _x_ here is the "location along the membrane", e.g., `x`. Unfortunately this isn't stated in the nomenclature, but is at the bottom left on page 4:
Screenshot 2023-02-01 at 8 50 10 AM

Further, if my supposition is correct and we need to change this equation, then we need to carefully consider the "backwards" case, e.g., when `x` starts at `1` and goes to `0`.

bknueven (commented):
@adam-a-a we could use your expertise to vet the implementation of the length-dependent Sherwood number.

watertap/core/membrane_channel_base.py
adam-a-a: @bknueven I'm confused about this for the backwards case because the Sherwood number is not the only length-dependent variable. Would we then need to revisit the local friction factor and other length-dependent variables in the backwards case? Maybe it's only a concern here because x is actually in the constraint, unlike (most of?) our other length-dependent variables. 

In any case, @MarcusHolly instead of `b.length` in the numerator, it should probably be something like `x * b.length` since x will be some value in the normalized length domain (0 to 1) and multiplying by length would yield the actual location. We can discuss today during our group call.

bknueven: > Maybe it's only a concern here because x is actually in the constraint, unlike (most of?) our other length-dependent variables. 

I think this is correct, but it's probably worth double checking to make sure it's the only one that would explicitly have `x` in the constraint.

bknueven:
```suggestion
            def eq_N_Sh_comp_length_dependent(b, t, x, j):
                if b.flow_direction == FlowDirection.forward:
                    x_ = x
                else:
                    x_ = (1-x)
                if x_ == 0:
                    return Constraint.skip
                return b.N_Sh_comp[t, x, j] == 2.401 * (
                    (b.N_Re[t, x] * b.N_Sc_comp[t, x, j]) ** 0.297
                ) * (((x_ / b.length) / b.dh) ** -0.279)
```

adam-a-a:
@zacharybinger This PR may be of interest to you. We left this alone for a while, and I am leaning towards closing this draft, but the intended change is relevant to some of the mods we might be bringing in for RO. Ultimately, we were trying to bring in a new Sherwood correlation based on a recent CFD study supported through NAWI. From rough recollection, this difficulty came down to 0D vs 1D cases and how to treat `x` in the correlation (x being the location at some length of the module). Specifically, I seem to recall the inlet (x=0) for 0D formulation being the issue. We had ideas on how to address but never settled on/implemented a final solution.

Just want to put this on your radar before I close it.

adam-a-a:
While we would still be interested in incorporating this correlation, this draft has been around for some time. We can reopen a new PR whenever we get around to attacking this issue again.

PR #873, opened 2022-12-15
Electrodialysis flowsheet with pumps and flow recirculation 

lbibl:
- This adds a new flowsheet example of electrodialysis.  
- This new flowsheet delivers a "feed-and-bleed" mode of ED stack, operated with pumps, etc., with components of unit simulation and cost optimization.  

### Legal Acknowledgement


watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
```diff
@@ -0,0 +1,495 @@
...
+        momentum_mixing_type=MomentumMixingType.none,
+        inlet_list=["from_feed", "from_conc_out"],
+    )
+
+    m.fs.pump0 = Pump(property_package=m.fs.properties)
+    m.fs.pump0.del_component("ratioP")
+    m.fs.pump0.del_component("ratioP_calculation")
+    m.fs.pump1 = Pump(property_package=m.fs.properties)
+    m.fs.pump1.del_component("ratioP")
+    m.fs.pump1.del_component("ratioP_calculation")
```
bknueven: This is a bit weird -- should we change the `Pump` unit?

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
```diff
@@ -0,0 +1,495 @@
...
+    m.fs.mix0.initialize(optarg=optarg)
+    m.fs.costing.initialize()
+
+    iscale.calculate_scaling_factors(m)
+    print("BADLY SCALED VARS & CONSTRAINS")
+    badly_scaled_var_values = {
+        var.name: val for (var, val) in iscale.badly_scaled_var_generator(m)
+    }
+    for j, k in badly_scaled_var_values.items():
+        print(j, ":", k)
```
bknueven: I'm not sure this should be part of this PR.

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
```diff
@@ -0,0 +1,495 @@
...
+            value(m.fs.disp.inlet.pressure[0]),
+            value(m.fs.disp.inlet.temperature[0]),
+        ),
+    }
+    pt_table = pd.DataFrame(data=pt_dict, index=["Pressure (Pa)", "Temperature (K)"])
+    print(pt_table)
+
+
+if __name__ == "__main__":
+    main()
```
bknueven:
```suggestion
    m = main()
```
So you can access this if running inside a notebook.

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
```diff
@@ -0,0 +1,495 @@
...
+    m.fs.mix0.from_feed_state[0].flow_vol_phase[...]
+    m.fs.mix0.from_conc_out_state[0].flow_vol_phase[...]
+    m.fs.pump0.control_volume.properties_in[0].flow_vol_phase[...]
+    m.fs.pump0.control_volume.properties_in[0].flow_vol_phase[...]
+    m.fs.pump1.control_volume.properties_in[0].flow_vol_phase[...]
+    m.fs.pump1.control_volume.properties_in[0].flow_vol_phase[...]
+    m.fs.prod.properties[0].flow_vol_phase[...]
+    m.fs.disp.properties[0].flow_vol_phase[...]
+    m.fs.EDstack.diluate.properties[...].flow_vol_phase[...]
+    m.fs.EDstack.concentrate.properties[...].flow_vol_phase[...]
```
bknueven: I saw something similar in #875 .. why are we doing this exactly? The lazy property creation should do this for you, and calculating additional properties not strictly needed to determine a feasible/optimal solution can cause additional numerical difficulties for Ipopt (or any solver).

bknueven (commented):
Looks pretty good. I have a few minor comments.

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
lbibl: > I saw something similar in #875 .. why are we doing this exactly? The lazy property creation should do this for you, and calculating additional properties not strictly needed to determine a feasible/optimal solution can cause additional numerical difficulties for Ipopt (or any solver).

Hi Ben, could you explain what is "lazy property creation"?  So these are all properties that are later needed and touching them here seems to provide  some pre-calculation to set  better initial values.  I don't think they cause difficulties here but probably removing them will also be fine.

bknueven: Basically they're only created after they're accessed the first time. If there is some benefit to doing this here as opposed to later on, then maybe there's something that needs to be done when we create the properties initially.

I'm guessing you do this because the initial values for the properties are better aligned. I wonder if the property constructor methods should calculate the value based on the existing variable values, e.g., call `calculate_variable_from_constraint` to get an initial value for the property.

@andrewlee94 I'm sure there's probably a good reason this isn't the standard thing to do. Thoughts?

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
lbibl: I del the var and its corresponding constraint because I don't really need them for my model.  Keeping them will not cause an error but create unnecessary computation. I would keep this as is now.

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
lbibl: Upon discussion with @TimBartholomew , I removed unnecessary touching of variables and kept minimum needed for reporting outcome. I recommend accepting this change for now, while we can still linger on Ben's comment above and make subsequent changes if needed.

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
lbibl: accepted.

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
lbibl: Deleted this part.

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
bknueven: So, I would recommend just deactivating the constraint instead, e.g., `m.fs.pump0.ratioP_calculation.deactivate()`. If the pump has underlying changes I think that'll be more robust.

If the variable `ratioP` is not used elsewhere, then it won't get included in the model when its handed to Ipopt.

lbibl: Recommendation accepted.

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
```diff
@@ -0,0 +1,460 @@
...
+    # ---specifications---
+    # Here is simulated a scenario of a defined EDstack and
+    # specific water recovery and product salinity.
+    m.fs.feed.properties[0].pressure.fix(101325)
+    m.fs.feed.properties[0].temperature.fix(298.15)
+    m.fs.pump1.control_volume.properties_in[0].pressure.fix(101325)
+    m.fs.pump1.efficiency_pump.fix(0.8)
+    m.fs.pump0.efficiency_pump.fix(0.8)
+
+    # m.fs.pump1.control_volume.properties_in[0].temperature.fix(298.15)
```
adam-a-a: commented code

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
```diff
@@ -0,0 +1,460 @@
...
+    m.fs.prod.properties[0].pressure.fix(101325)  #
+    # m.fs.prod.properties[0].temperature.fix(298.15)
```
adam-a-a: commented code

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
```diff
@@ -0,0 +1,460 @@
...
+        source=m.fs.EDstack.outlet_concentrate, destination=m.fs.sepa1.inlet
+    )
+    m.fs.arc6 = Arc(source=m.fs.sepa1.to_disp, destination=m.fs.disp.inlet)
+    m.fs.arc7 = Arc(source=m.fs.sepa1.to_conc_in1, destination=m.fs.mix0.from_conc_out)
+    TransformationFactory("network.expand_arcs").apply_to(m)
+
+    return m
+
+
+def condition_base(m):
```
adam-a-a, line 262: Not critical, but several flowsheets (not necessarily all yet), use `set_operating_conditions` as the function name. It might be nice to avoid vastly different naming across flowsheet example.

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
```diff
@@ -0,0 +1,460 @@
...
+
+    # add extra variables and constraints
+    m.fs.recovery_vol_H2O = Var(
+        initialize=0.5,
+        bounds=(0, 1),
+        domain=NonNegativeReals,
+        units=pyunits.dimensionless,
+        doc="flowsheet level water recovery calculated by volumeric flow rate",
+    )
+    m.fs.mem_area = Var(
```
adam-a-a, line 189: Any reason why this created at the flowsheet level rather than the unit model?

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
```diff
@@ -0,0 +1,460 @@
...
+        bounds=(0, 1e3),
+        units=pyunits.meter**2,
+        doc="Total membrane area for cem (or aem) in one stack",
+    )
+    m.fs.prod_salinity = Var(
+        initialize=1, bounds=(0, 1000), units=pyunits.kg * pyunits.meter**-3
+    )
+    m.fs.disp_salinity = Var(
+        initialize=1, bounds=(0, 1e6), units=pyunits.kg * pyunits.meter**-3
+    )
```
adam-a-a: should add doc strings and spell out the names (e.g., `product_salinity`)

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
```diff
@@ -0,0 +1,460 @@
...
+    m.fs.mix0 = Mixer(
+        property_package=m.fs.properties,
+        energy_mixing_type=MixingType.none,
+        momentum_mixing_type=MomentumMixingType.none,
+        inlet_list=["from_feed", "from_conc_out"],
+    )
+
+    m.fs.pump0 = Pump(property_package=m.fs.properties)
+    m.fs.pump0.ratioP_calculation.deactivate()
```
adam-a-a, line 140: Why do you deactivate this?

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
```diff
@@ -0,0 +1,460 @@
...
+        property_package=m.fs.properties,
+        operation_mode=ElectricalOperationMode.Constant_Voltage,
+        finite_elements=20,
+        has_pressure_change=True,
+        has_nonohmic_potential_membrane=True,
+        has_Nernst_diffusion_layer=True,
+        limiting_current_density_method=LimitingCurrentDensityMethod.Theoretical,
+        pressure_drop_method=PressureDropMethod.Darcy_Weisbach,
+        hydraulic_diameter_method=HydraulicDiameterMethod.spacer_specific_area_known,
+        friction_factor_method=FrictionFactorMethod.Gurreri,
```
adam-a-a, line 155: nice work

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
adam-a-a: Also, why do you need these variables? I might be mistaken, but can't you just touch `conc_mass_phase_comp` on the corresponding Product blocks?

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
```diff
@@ -0,0 +1,460 @@
...
+            for j in m.fs.properties.ion_set | m.fs.properties.solute_set
+        )
+    )
+
+    m.fs.eq_mem_area = Constraint(
+        expr=m.fs.mem_area
+        == m.fs.EDstack.cell_width
+        * m.fs.EDstack.cell_length
+        * m.fs.EDstack.cell_pair_num
+    )
```
adam-a-a, line 237: Given this constraint, this reinforces my previous question on why this is at the flowsheet level instead of the unit level.

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
```diff
@@ -0,0 +1,460 @@
...
+    propagate_state(m.fs.arc7)
+    propagate_state(m.fs.arc3b, direction="backward")
+    propagate_state(m.fs.arc2)
+    m.fs.mix0.initialize(optarg=optarg)
+    m.fs.costing.initialize()
+
+    iscale.calculate_scaling_factors(m)
+
+
+def display_model_metrics(m):
```
adam-a-a, line 380: Curious why you don't use the unit model's report method within this function. It may not cover everything but should cover some key unit-level metrics at least (if you included in the unit model)

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
```diff
@@ -0,0 +1,460 @@
...
+    LimitingCurrentDensityMethod,
+)
+from watertap.unit_models.electrodialysis_1D import Electrodialysis1D
+from watertap.costing.watertap_costing_package import WaterTAPCosting
+from watertap.property_models.multicomp_aq_sol_prop_pack import MCASParameterBlock
+
+__author__ = "Xiangyu Bi"
+_log = idaeslogger.getLogger(__name__)
+
+
```
adam-a-a, line 57: Nice to have: a doc string telling the user what this flowsheet is intended to represent.

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
lbibl: This is a bit different that it leaves two dof to be flexibly controlled.  I need this to efficiently optimize an objective.  I would keep this name to avoid the confusion.

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
lbibl: The unit does not need this var for anything but the flowsheet need it for reporting data.

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
lbibl: This is not used so it's a good practice to pass less loading to the solver wherever possible.

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
lbibl: It was like you suggested before.  But I found when I add up the concentration to express a molecular salinity, doing it in this way directs the solver to minimize the error of the sum of two species.  In others words, minimizing two individual errors may not minimize the sum of the two if that makes sense.

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
lbibl: done.

lbibl:
@adam-a-a I've responded to or addressed your comments.

TimBartholomew (approved):

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
adam-a-a: Which 2 dof?

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
adam-a-a: I still think you should have `area` on the unit model; that would make it consistent with other unit models that we have. Even if you can get away without it, I'd rather try to keep some level of consistency where possible.

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
adam-a-a: And why is that needed to optimize? Can you add some more clarification?

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
```diff
@@ -0,0 +1,474 @@
...
+    pm_table = pd.DataFrame(
+        data=[
+            value(m.fs.recovery_vol_H2O),
+            value(m.fs.mem_area),
+            value(m.fs.EDstack.cell_pair_num),
+            value(m.fs.EDstack.cell_length),
+            value(m.fs.EDstack.voltage_applied[0]),
+            value(m.fs.costing.specific_energy_consumption),
+            value(m.fs.costing.LCOW),
+        ],
```
adam-a-a, line 424: For ED-related metrics, any reason why you don't use `report()` anywhere?

adam-a-a (commented):
LGTM - just a couple simple comments to look over.

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
lbibl: I consider an optimal model structure more important than "some level of consistency".  The `area` isn't an essential variable member for ED model but maybe it is for other models.  Adding it means adding a reducible dimension to the equation matrix.  The solver will take resource to reduce the dimension.  When building complex flowsheet connecting multiple unit models, maintaining each unit in its cleanest model structure will be in favor for minimizing numerical difficulties in all aspects.   Secondary things like this are more flexibly handled at the flowsheet level than being embedded in unit models.   Please see if this change your opinion.  If not, I will need more time to make this change in a couple of files.

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
lbibl: The `report()` was added in the last revision.

watertap/examples/flowsheets/electrodialysis/electrodialysis_1stack_conc_recirc.py
bknueven: I think we should either consider (1) making area an `Expression` on the unit model or (2) making it an `Expression` on the flowsheet.

I suspect that goes for all of these quantities. No reason for a variable just for reporting purposes.

bknueven (approved):
I have comments, and I don't necessarily disagree with what Adam is saying.

But they're all relatively minor. Let's go ahead and merge this.

lbibl:
> I have comments, and I don't necessarily disagree with what Adam is saying.
> 
> But they're all relatively minor. Let's go ahead and merge this.

Thanks @bknueven .   I do feel `Expression` is the best way to go.  Honestly I didn't use this much so it skipped my opinions.   I will consider making some modification later.

PR #875, opened 2022-12-20
Flowsheet demonstration of IonExchange0D

kurbansitterley:
## Fixes/Resolves:

Simulation and optimization example for IonExchange0D unit model in flowsheet

## Summary/Motivation:

Create flowsheet demonstration file to be used in lab report for IX model. 

- add ``ion_exchange_demo.py`` to flowsheet examples
- add test file

### Legal Acknowledgement


watertap/examples/flowsheets/ion_exchange/ion_exchange_demo.py
```diff
@@ -0,0 +1,333 @@
...
+    solver.solve(ix)
+    # num_col = np.ceil(ix.number_columns())  # To eliminate fractional number of columns
+    # bed_depth = ix.bed_depth()
+    # ix.bed_depth.fix(bed_depth)
+    # ix.number_columns.fix(num_col)
+
+
+def get_ion_config(ions):
+
+    if not isinstance(ions, list):
```
bknueven:
```suggestion
    if not isinstance(ions, (list, tuple)):
```
Just to make this a bit more robust ..

bknueven: You might also ask instead
```suggestion
    if isinstance(ions, str):
```
If we expect the ion list to be expressed as a list of strings.

watertap/examples/flowsheets/ion_exchange/ion_exchange_demo.py
```diff
@@ -0,0 +1,333 @@
...
+    num_col = np.ceil(ix.number_columns())  # To eliminate fractional number of columns
+    bed_depth = ix.bed_depth()
+    ix.bed_depth.fix(bed_depth)
+    ix.number_columns.fix(num_col)
+    check_dof(m)
+    results = solver.solve(m)
+    assert_optimal_termination(results)
+    print(f"\nDOF = {degrees_of_freedom(m)}")
+    print(f"Model solve {results.solver.termination_condition.swapcase()}")
+    display_results(m)
```
bknueven, line 73:
```suggestion
    display_results(m)
    
    return m
```

watertap/examples/flowsheets/ion_exchange/ion_exchange_demo.py
```diff
@@ -0,0 +1,333 @@
...
+        print(
+            f'{f"Conc. Out [{ion}, mg/L]":<40s}{pyunits.convert(prop_out.conc_mass_phase_comp[liq, ion], to_units=pyunits.mg/pyunits.L)():<40.3e}{"mg/L":<40s}'
+        )
+        print(
+            f'{f"Conc. Regen [{ion}, mg/L]":<40s}{pyunits.convert(prop_regen.conc_mass_phase_comp[liq, ion], to_units=pyunits.mg/pyunits.L)():<40.3e}{"mg/L":<40s}'
+        )
+
+
+if __name__ == "__main__":
+    main()
```
bknueven:
```suggestion
    m = main()
```

This is just a suggestion (along with `main` returning `m`), which can make it a bit nicer to interact with the flowsheet in the python notebook. (E.g., you can execute `%run ion_exchange_demo.py` and have the model available in your environment.)

watertap/examples/flowsheets/ion_exchange/tests/test_ion_exchange_demo.py
```diff
@@ -0,0 +1,210 @@
...
+            1.10981, abs=1e-4
+        )
+        assert value(m.fs.costing.specific_energy_consumption) == pytest.approx(
+            0.032097, abs=0.0001
+        )
+        assert value(m.fs.costing.LCOW) == pytest.approx(0.160, abs=0.001)
+
+    @pytest.mark.unit
+    def test_main_fun(self):
+        ixf.main()
```
bknueven: If you have `main` return the model, you could also spot check the same values you do immediately above.

watertap/examples/flowsheets/ion_exchange/ion_exchange_demo.py
```diff
@@ -0,0 +1,333 @@
...
+from idaes.core.util.scaling import set_scaling_factor, calculate_scaling_factors
+from idaes.core.util.initialization import propagate_state
+from idaes.models.unit_models import Product, Feed
+
+from watertap.core.util.initialization import check_dof
+from watertap.property_models.multicomp_aq_sol_prop_pack import MCASParameterBlock
+from watertap.unit_models.ion_exchange_0D import IonExchange0D
+from watertap.costing import WaterTAPCosting
+
+import numpy as np
```
bknueven: Since we only use `ceil`, from `numpy`, we should just use the equivalent in `math` (which is built-in).
E.g.,
```python
import math

num_col = math.ceil(ix.number_columns())
```

watertap/examples/flowsheets/ion_exchange/tests/test_ion_exchange_demo.py
```diff
@@ -0,0 +1,210 @@
...
+        assert value(m.fs.product.properties[0].flow_vol_phase["Liq"]) == pytest.approx(
+            0.049901566, abs=1e-6
+        )
+        assert value(m.fs.regen.properties[0].flow_vol_phase["Liq"]) == pytest.approx(
+            0.016666, abs=1e-6
+        )
+        ix = m.fs.ion_exchange
+        num_col = np.ceil(
+            ix.number_columns()
+        )  # To eliminate fractional number of columns
```
bknueven:
```suggestion
        # To eliminate fractional number of columns
        num_col = math.ceil(ix.number_columns())  
```
Similar to my comment above in the demo.

watertap/examples/flowsheets/ion_exchange/ion_exchange_demo.py
```diff
@@ -0,0 +1,333 @@
...
+    ix.dimensionless_time.unfix()
+    # Can optimize around different design variables, e.g., bed_depth, service_flow_rate (or combinations of these)
+    # Here demonstrates optimization around column design
+    ix.number_columns.unfix()
+    ix.bed_depth.unfix()
+    solver.solve(ix)
+    # num_col = np.ceil(ix.number_columns())  # To eliminate fractional number of columns
+    # bed_depth = ix.bed_depth()
+    # ix.bed_depth.fix(bed_depth)
+    # ix.number_columns.fix(num_col)
```
bknueven: Since this is a demo, maybe it makes sense to keep this commented code? Not sure.

bknueven (commented):
Looks pretty good. I just have a few minor comments.

watertap/examples/flowsheets/ion_exchange/ion_exchange_demo.py
```diff
@@ -0,0 +1,333 @@
...
+    for ion in ions:
+        ion_config["solute_list"].append(ion)
+        ion_config["diffusivity_data"][("Liq", ion)] = diff_data[ion]
+        ion_config["mw_data"][ion] = mw_data[ion]
+        ion_config["charge"][ion] = charge_data[ion]
+    return ion_config
+
+
+def display_results(m):
+
```
hunterbarber, line 363: How different is this than the `.report()`? If this truly is to serve as a demo what is the value in custom reporting if there is a report options internal to WaterTAP and the IX unit model?

watertap/examples/flowsheets/ion_exchange/ion_exchange_demo.py
```diff
@@ -0,0 +1,333 @@
...
+    check_dof(m)
+    results = solver.solve(m)
+    assert_optimal_termination(results)
+    print(f"\nDOF = {degrees_of_freedom(m)}")
+    print(f"Model solve {results.solver.termination_condition.swapcase()}")
+    display_results(m)
+
+    optimize_system(m)
+    ix = m.fs.ion_exchange
+    num_col = np.ceil(ix.number_columns())  # To eliminate fractional number of columns
```
hunterbarber:
```suggestion
    # To eliminate fractional number of columns
    num_col = np.ceil(ix.number_columns()) 
```

watertap/examples/flowsheets/ion_exchange/tests/test_ion_exchange_demo.py
```diff
@@ -0,0 +1,210 @@
...
+        bed_depth = ix.bed_depth()
+        ix.bed_depth.fix(bed_depth)
+        ix.number_columns.fix(num_col)
```
hunterbarber, line 194: Did you ensure the ceil of the solution is truly the optimum and no weird concavity in the optimization exists? Without pure integer optimization, if the optimization solved to 4.2, 4 could actually be more optimum than 5. This may be possible with the objective function as LCOW, although not possible for something like minimizing a sizing parameter.

hunterbarber (commented):
a few comments that may be relevant

watertap/examples/flowsheets/ion_exchange/ion_exchange_demo.py
adam-a-a: We could check what the ion exchange unit model has in its `_get_performance_contents` method to use `.report()`. With that said, we have been using `display_results` across most flowsheets just to display flowsheet level results that might go beyond a particular unit model. However, we probably could use `.report()` to display the IX model's outputs under this function.

kurbansitterley: IX unit model has both of those methods, but they need to be updated. Will be quick, but I will have to make a separate PR for it. 

Also a downside to ``.report()`` is it doesn't display costing data, which I wanted to include in the demo here.

watertap/examples/flowsheets/ion_exchange/tests/test_ion_exchange_demo.py
kurbansitterley: I have done no real statistics based checking on this outside of some I made during the initial model development that showed minimal impact to LCOW when using floor or ceil, and I figured this was acceptable given the lack of integer optimization for our solver (and that I would leave it to the user to pick one or the other, or neither if they are mostly looking for costing results rather than design and so would be ok with fractional column values). 

I also found that the model is generally pushing the bounds of the column sizing variables and using floor can lead to less reliable solves because you are trying to cram the same amount of resin into fewer columns, thus the columns must be bigger. So I always use ceil... but yeah the design aspect of this model is not ideal for that reason.

bknueven: #885 is related -- where we have a parameter we might want to be an integer variable in certain contexts.

watertap/examples/flowsheets/ion_exchange/tests/test_ion_exchange_demo.py
kurbansitterley: switched to ``math``

watertap/examples/flowsheets/ion_exchange/ion_exchange_demo.py
kurbansitterley: Incorporated these suggestions.

watertap/examples/flowsheets/ion_exchange/tests/test_ion_exchange_demo.py
hunterbarber: Yea, this is definitely larger than the scope of this. I think this is fine to pass and have #885 to capture this for the future.

hunterbarber (approved):
Changes since last review are all minor, LGTM.

watertap/examples/flowsheets/ion_exchange/ion_exchange_demo.py
```diff
@@ -0,0 +1,424 @@
...
+    # To adjust solution to fixed-pattern to achieve desired effluent, must unfix dimensionless_time.
+    ix.dimensionless_time.unfix()
+    # Can optimize around different design variables, e.g., bed_depth, service_flow_rate (or combinations of these)
+    # Here demonstrates optimization around column design
+    ix.number_columns.unfix()
+    ix.bed_depth.unfix()
+    solver.solve(ix)
+
+
+def get_ion_config(ions):
```
avdudchenko, line 329: We might want to add commetnts here on what parameters/data needed.

avdudchenko (approved):
This look clear enough to me.

PR #877, opened 2022-12-30
ZO Documentation Test

MarcusHolly:
## Fixes/Resolves:

1st point of Issue #595 

### Legal Acknowledgement

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -86,7 +86,10 @@ setup(
         "Programming Language :: Python :: 3 :: Only",
     ],
     keywords="water systems, chemical engineering, process modeling, filtration, desalination, nawi",
-    packages=find_namespace_packages(),
+    # just include watertap and everything under it
+    packages=find_namespace_packages(
+        include=("watertap*",),
+    ),
     python_requires=">=3.7",
     install_requires=[
         # primary requirements for unit and property models
--- a/watertap/unit_models/zero_order/tests/test_zo_documentation.py
+++ b/watertap/unit_models/zero_order/tests/test_zo_documentation.py
@@ -54,7 +54,11 @@ def generate_model_list(path, extension, exclude):
 
 
 @pytest.mark.integration
-@pytest.mark.skipif(not os.path.exists(doc_path), reason="No docs directory")
+doc_path = Path(doc_path).resolve()
+
+...
+
+@pytest.mark.skipif(not os.path.exists(doc_path / "index.rst"), reason="No RST files found")
 def test_lists_match():
 
     rst_list = generate_model_list(rst_path, ".rst", ("index.rst",))
--- a/watertap/unit_models/zero_order/tests/test_zo_documentation.py
+++ b/watertap/unit_models/zero_order/tests/test_zo_documentation.py
@@ -59,12 +59,37 @@ def test_rst_path():
         "zero_order_unit_models",
     )
 
+    for file in os.listdir(rst_path):
+        filename = os.fsdecode(file)
+        if filename.endswith(".rst") and filename not in exclude_rst_files:
+            file_path = os.path.join(
+                os.path.dirname(os.path.abspath(__file__)),
+                "..",
+                "..",
+                "..",
+                "..",
+                "docs",
+                "technical_reference",
+                "unit_models",
+                "zero_order_unit_models",
+                f"{file}",
+            )
+            assert os.path.isfile(file_path) == True
+
 
 @pytest.mark.unit
 def test_py_path():
     assert os.path.exists(py_path) == True
 
     assert py_path == os.path.join(os.path.dirname(os.path.abspath(__file__)), "..")
+
+    for file in os.listdir(py_path):
+        filename = os.fsdecode(file)
+        if filename.endswith(".py") and filename not in exclude_py_files:
+            file_path = os.path.join(
+                os.path.dirname(os.path.abspath(__file__)), "..", f"{file}"
+            )
+            assert os.path.isfile(file_path) == True
 
 
 @pytest.mark.integration
--- a/watertap/unit_models/zero_order/tests/test_zo_documentation.py
+++ b/watertap/unit_models/zero_order/tests/test_zo_documentation.py
@@ -16,85 +16,44 @@ Tests that documentation exists for all zero-order unit models
 import pytest
 import os
 
-rst_path = os.path.join(
+doc_path = os.path.join(
     os.path.dirname(os.path.abspath(__file__)),
     "..",
     "..",
     "..",
     "..",
     "docs",
+)
+
+rst_path = os.path.join(
+    doc_path,
     "technical_reference",
     "unit_models",
     "zero_order_unit_models",
 )
-rst_list = []
-exclude_rst_files = ["index.rst"]
-for file in os.listdir(rst_path):
-    filename = os.fsdecode(file)
-    if filename.endswith(".rst") and filename not in exclude_rst_files:
-        rst_list.append(filename[:-4])
-
 py_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), "..")
-py_list = []
-exclude_py_files = ["__init__.py"]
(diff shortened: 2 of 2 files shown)


MarcusHolly:
This test is very simple for now, but I'm not sure what other things should be tested besides the two lists matching. That is to say, if the two lists match, then there shouldn't be any other issues pertaining to the documentation.

adam-a-a:
Looking at the test failures very briefly, it seems in a couple of the checks, the rst file list is turning up empty, which leads me to believe that the way you navigate the docs directory may not work as intended. Perhaps you could try adding one/two extra checks that assert the length of each list -- these would just be additional checks to provide more clarity to the developer if ever there were an issue like this in the future, supplementing the list comparison assertion that you already have. However, this doesn't solve your problem with the test failure...

adam-a-a:
Actually- it looks like the "user mode" checks are the ones failing. I am unsure why. @lbianchi-lbl - any thoughts? Is the docs directory path nonexistent when installed as a user? @MarcusHolly you could also add checks to assert that the file paths exist

adam-a-a:
Small suggestion- I'd put the length of list checks before the check to compare lists.

watertap/unit_models/zero_order/tests/test_zo_documentation.py
```diff
@@ -0,0 +1,100 @@
...
+            file_path = os.path.join(
+                os.path.dirname(os.path.abspath(__file__)), "..", f"{file}"
+            )
+            assert os.path.isfile(file_path) == True
+
+
+@pytest.mark.integration
+def test_lists_match():
+    assert len(py_list) == len(rst_list)
+
```
bknueven:
```suggestion
```
I would get rid of this assertion -- the one below will help you debug a missing file, but this one won't be as helpful.

bknueven:
On the specifics of the test failures -- I am surprised the test `test_rst_path` passed on the user-mode builds.

I looked into it, and after installing latest `watertap` release via `pip` in a fresh environment, I *do* see the `docs` directory!
```bash
(watertap_test) bknueven-35255s:zero_order_unit_models bknueven$ pwd
/Users/bknueven/opt/anaconda3/envs/watertap_test/lib/python3.9/site-packages/docs/technical_reference/unit_models/zero_order_unit_models
(watertap_test) bknueven-35255s:zero_order_unit_models bknueven$ ls
__pycache__			automate_rst_file_wt3.py	dummy_sidor_data.yaml
```

I also see our `tutorials` directory and the `bin` directory. Since I think these should be excluded (and end up at the top-level site-packages directory, not inside `watertap`), I think we should address that issue first *before* attempting to fix this PR for user mode tests.

MarcusHolly:
> On the specifics of the test failures -- I am surprised the test `test_rst_path` passed on the user-mode builds.
> 
> I looked into it, and after installing latest `watertap` release via `pip` in a fresh environment, I _do_ see the `docs` directory!
> 
> ```shell
> (watertap_test) bknueven-35255s:zero_order_unit_models bknueven$ pwd
> /Users/bknueven/opt/anaconda3/envs/watertap_test/lib/python3.9/site-packages/docs/technical_reference/unit_models/zero_order_unit_models
> (watertap_test) bknueven-35255s:zero_order_unit_models bknueven$ ls
> __pycache__			automate_rst_file_wt3.py	dummy_sidor_data.yaml
> ```
> 
> I also see our `tutorials` directory and the `bin` directory. Since I think these should be excluded (and end up at the top-level site-packages directory, not inside `watertap`), I think we should address that issue first _before_ attempting to fix this PR for user mode tests.

Alright, thanks for looking at this @bknueven , and your explanation makes sense. I agree that your issue should be addressed first before coming back to fix this PR. So it seems like, after your issue is addressed, the user-mode tests will still fail. Is it better to just bypass these tests or is there a way to/ is it reasonable to install the docs directory directly within the test file (only if it doesnt already exist - which would be the case in user-mode) such that all the tests pass?

As for why the test_rst_path is passing, isn't it because the path itself exists (docs/technical_reference/unit_models/zero_order_unit_models), but the directory only contains the 3 files in your screenshot, none of which are rst files? I would guess that after your issue is addressed, this test would fail because the docs directory will be excluded from the build.

setup.py
```diff
@@ -86,7 +86,10 @@
         "Programming Language :: Python :: 3 :: Only",
     ],
     keywords="water systems, chemical engineering, process modeling, filtration, desalination, nawi",
-    packages=find_namespace_packages(),
+    # just include watertap and everything under it
+    packages=find_namespace_packages(
+        include=("watertap*",),
+    ),
```
lbianchi-lbl, line 92: Unless there's a strong reason to keep this, we should consider reverting this change before merging.

watertap/unit_models/zero_order/tests/test_zo_documentation.py
```diff
@@ -0,0 +1,64 @@
...
+                path,
+                file,
+            )
+            assert os.path.isfile(file_path)
+            file_list.append(filename[: -len(extension)])
+    return file_list
+
+
+@pytest.mark.integration
+@pytest.mark.skipif(not os.path.exists(doc_path), reason="No docs directory")
```
lbianchi-lbl: I think the reason why the tests are failing is that `doc_path` exists (so the test doesn't get skipped), but `rst_path` doesn't (so the test fails). Therefore, it might work with this change:

```suggestion
@pytest.mark.skipif(not os.path.exists(rst_path), reason="Directory with .rst source files for ZO models docs not found")
```

lbianchi-lbl (changes requested):
I've left a suggestion on a possible way to address the test failures in the non-editable ("user mode") CI environment.

Apart from that, I think the current implementation is OK, but we might want to revisit it to make it more robust sooner rather than later. More in general, this test has two aspects that make it tricky:

- Matching between the Sphinx docs and the Python code (including e.g. finding the location of the `docs` directory, that might or might not be there depending on the installation mode)
- Finding all ZO models, which in turns requires listing all Python modules within a given package

Both of these would be useful to have for other tests beyond this one, so it might make sense to merge this (as soon as it works reliably) and then refactor it together with the other related tests in a separate PR.

setup.py
lbianchi-lbl: To clarify:

- I think the point being made in #879 is valid, and we should modify the `setup.py` settings so that files and directories other than `watertap` don't end up in the installation
- However, since this change would impact other things beyond the original scope of this PR (and I'm planning to rework our distribution setup soon), it might make sense to deal with it separately
- That said, ultimately I don't have strong preferences and I'm reasonably confident that our tests would catch any possible regression introduced by this change, so I'd be OK if we wanted to merge it as part of this PR

watertap/unit_models/zero_order/tests/test_zo_documentation.py
lbianchi-lbl: On second thought, reading https://github.com/watertap-org/watertap/pull/877#issuecomment-1371266240 and https://github.com/watertap-org/watertap/pull/877#issuecomment-1371293292, I don't think this would work: the directory would exist, but without any `.rst` file.

Testing for the presence of at least one `.rst` file should work:

```suggestion
doc_path = Path(doc_path).resolve()

...

@pytest.mark.skipif(not os.path.exists(doc_path / "index.rst"), reason="No RST files found")
```

However, this starts to look a bit too brittle for my tastes...

setup.py
MarcusHolly: Although it may look a bit brittle, all of the tests are passing now. As for this change, I'll let @bknueven comment on if there's a strong reason to keep it.

lbianchi-lbl: Yeah, I agree, I don't think there's any harm in merging this as-is (including the change in `setup.py`).

lbianchi-lbl (approved):
As far as I'm concerned, we can merge this as it is and then refactor the shared functionality in a separate PR.

bknueven (approved):

PR #880, opened 2023-01-04
Add OARO costing method

luohezhiming:
## Fixes/Resolves:

## Summary/Motivation:
Add OARO costing method and optimization for flowsheet

- add OARO costing method
- add cost optimization for OARO flowsheet

### Legal Acknowledgement


watertap/costing/units/osmotically_assisted_reverse_osmosis.py
```diff
@@ -0,0 +1,82 @@
...
+@register_costing_parameter_block(
+    build_rule=build_osmotically_assisted_reverse_osmosis_cost_param_block,
+    parameter_block_name="osmotically_assisted_reverse_osmosis",
+)
+def cost_high_pressure_osmotically_assisted_reverse_osmosis(blk):
+    cost_membrane(
+        blk,
+        blk.costing_package.osmotically_assisted_reverse_osmosis.high_pressure_membrane_cost,
+        blk.costing_package.osmotically_assisted_reverse_osmosis.factor_membrane_replacement,
+    )
```
MarcusHolly, line 86: LGTM - just one question.

This is my first time looking at the new costing method, but is there a reason lines 78-82 can't just be substituted directly into line 65? I don't see why it's necessary to make a separate function for the high_pressure case, so it might just be better to condense everything into one function.

luohezhiming: I followed up the original RO module, the unit model costs in our current framework only depend on membrance cost and replacement factor. They take two cases (normal and high pressure) into consideration when implementing the flowsheet. For this part, it does not matter to create a seprate function to represent high pressure case, maybe they consider some potential modifications of details for it. I believe it does not harm to keep this function.

MarcusHolly (approved):

watertap/costing/units/osmotically_assisted_reverse_osmosis.py
bknueven: @MarcusHolly this is so you can just specify the method `cost_high_pressure_osmotically_assisted_reverse_osmosis` directly as the costing method. E.g., you can do
```python
m.fs.unit.costing = UnitModelCostingBlock(
    flowsheet_costing_block=m.fs.costing
    costing_method=cost_high_pressure_osmotically_assisted_reverse_osmosis,
)
```

Of course, you can also do
```python
m.fs.unit.costing = UnitModelCostingBlock(
    flowsheet_costing_block=m.fs.costing
    costing_method_arguments={"oaro_type":OAROType.high_pressure},
)
```
though its a bit more verbose. We opt to allow a user to use either method for maximum flexibility.

bknueven (commented):
The `initialize_system` function in the oaro flowsheet should call `m.fs.costing.initialize()` after the units are initialized. See, e.g., line 812 of the LSSRO flowsheet: https://github.com/watertap-org/watertap/blob/fbf53a48473fb5241337052f814458e35426ffac/watertap/examples/flowsheets/lsrro/lsrro.py#L782-L812

watertap/costing/units/osmotically_assisted_reverse_osmosis.py
```diff
@@ -0,0 +1,82 @@
...
+    blk.membrane_cost = pyo.Var(
+        initialize=30,
+        doc="Membrane cost",
+        units=pyo.units.USD_2018 / (pyo.units.meter**2),
+    )
+    blk.high_pressure_membrane_cost = pyo.Var(
+        initialize=50,
+        doc="Membrane cost",
+        units=pyo.units.USD_2018 / (pyo.units.meter**2),
+    )
```
bknueven, line 45: Do we have a source of these numbers, or are the just notional? If there is a source, we should include it.

luohezhiming:
> The `initialize_system` function in the oaro flowsheet should call `m.fs.costing.initialize()` after the units are initialized. See, e.g., line 812 of the LSSRO flowsheet:
> 
> https://github.com/watertap-org/watertap/blob/fbf53a48473fb5241337052f814458e35426ffac/watertap/examples/flowsheets/lsrro/lsrro.py#L782-L812

Thanks for pointing out this, I've added costing initialization after all units are initialized.

watertap/costing/units/osmotically_assisted_reverse_osmosis.py
luohezhiming: We are following previous paper of Tim and Adam, they provided values for the OARO modules. @adam-a-a can you indicate me which paper is better for reference?

luohezhiming: @bknueven , I added citations for the source of these costing parameter values

bknueven (approved):

PR #889, opened 2023-01-14
Selective oil permeation model development

tarnold17:
## Summary/Motivation:
Add selective oil permeation unit model and property package

### Legal Acknowledgement


watertap/examples/SOP/SOP_prop_pack.py
```diff
@@ -0,0 +1,543 @@
...
+                "temperature": {"method": None},
+                "pressure": {"method": None},
+                "visc_d_phase_comp": {"method": "_visc_d_phase_comp"},
+                "dens_mass_phase_comp": {"method": "_dens_mass_phase_comp"},
+                "mass_frac_phase_comp": {"method": "_mass_frac_phase_comp"},
+                "flow_vol_phase_comp": {"method": "_flow_vol_phase_comp"},
+                "flow_vol_phase": {"method": "_flow_vol_phase"},
+                "flow_mass_phase": {"method": "_flow_mass_phase"},
+                "dens_mass_phase": {"method": "_dens_mass_phase"},
+                "vol_frac_phase_comp": {"method": "_vol_frac_phase_comp"},
```
adam-a-a: This looks like one of those properties that you'll need to throw into `obj.define_custom_properties`. See here for an example: https://github.com/watertap-org/watertap/pull/937/files

Note, this might not be the only property that you'll need to move out of this dict and move into `define_custom_properties`

watertap/examples/SOP/demo_SOP.py
```diff
@@ -0,0 +1,132 @@
+from pyomo.environ import (
```
adam-a-a: I am not sure how necessary this file is since it is essentially a copy of the test file. Maybe it could subsequently be built upon, e.g., by adding a feed pump at the beginning. We should consider whether we want to remove this file from this PR or followup with another PR to bolster this example. Not critical though

adam-a-a: A trivial comment: I think we should extend file names to `selective_oil_permeation` rather than SOP

tarnold17: I am on board with removing `demo_SOP.py` and using "selective_oil_permeation" instead of "SOP" in all names.

Do you think we should also get rid of `demo_SOP_prop_pack.py`?

watertap/examples/SOP/demo_SOP_prop_pack.py
```diff
@@ -0,0 +1,103 @@
+from pyomo.environ import ConcreteModel, assert_optimal_termination, value
```
adam-a-a: same comment as above

watertap/property_models/SOP_prop_pack.py
```diff
@@ -0,0 +1,551 @@
...
+            units=units.Pa,
+            doc="State pressure",
+        )
+
+    # -----------------------------------------------------------------------------
+    # Property Methods
+    # these property methods build variables and constraints on demand (i.e. only when asked
+    # for by a user or unit model)
+    def _visc_d_phase_comp(self):
+        self.visc_d_phase_comp = Param(
```
adam-a-a: I suppose we could get away with this, but for consistency purposes, we typically build the params on the parameter block above, then here, we'd do an `add_object_reference`. You can check out the multi_aq_sol_prop_pack and see an example or two of that (check it out for viscosity since the case for diffusivity may be a bit more complicated to read through)

watertap/property_models/SOP_prop_pack.py
```diff
@@ -0,0 +1,551 @@
...
+            initialize=1e-3,
+            units=units.kg * units.m**-1 * units.s**-1,
+            doc="Dynamic viscosity",
+        )
+
+        self.visc_d_phase_comp["Liq", "H2O"] = 1e-3
+        self.visc_d_phase_comp["Liq", "oil"] = 3.5e-3
+
+    def _dens_mass_phase_comp(self):
+        self.dens_mass_phase_comp = Param(
```
adam-a-a: same comment as above

watertap/property_models/SOP_prop_pack.py
```diff
@@ -0,0 +1,551 @@
...
+
+        self.eq_flow_mass_phase = Constraint(
+            self.params.phase_list, rule=rule_flow_mass_phase
+        )
+
+    def _dens_mass_phase(self):
+        self.dens_mass_phase = Var(
+            self.params.phase_list,
+            initialize=1e3,
+            bounds=(1e2, 2e3),
```
adam-a-a: not sure, but it could be a safer bet to relax this upper bound

watertap/property_models/SOP_prop_pack.py
```diff
@@ -0,0 +1,551 @@
...
+        self.eq_dens_mass_phase = Constraint(
+            self.params.phase_list, rule=rule_dens_mass_phase
+        )
+
+    def _vol_frac_phase_comp(self):
+        self.vol_frac_phase_comp = Var(
+            self.params.phase_list,
+            self.params.component_list,
+            initialize=0.1,
+            bounds=(None, None),
```
adam-a-a: Technically, bounds are 0 to 1, right?

watertap/property_models/SOP_prop_pack.py
```diff
@@ -0,0 +1,551 @@
...
+            self.params.component_list,
+            rule=rule_vol_frac_phase_comp,
+        )
+
+    def _conc_mass_phase_comp(self):
+        self.conc_mass_phase_comp = Var(
+            self.params.phase_list,
+            self.params.component_list,
+            initialize=1,
+            bounds=(None, None),
```
adam-a-a:
```suggestion
            domain=NonNegativeReals,
```
another way to write this

adam-a-a: actually I'm forgetting what the lower bound is for NonNegativeReals (i.e., 0 or None)

watertap/unit_models/selective_oil_permeation.py
```diff
@@ -0,0 +1,669 @@
...
+
+    def build(self):
+        # Call UnitModel.build to setup dynamics
+        super().build()
+
+        self.scaling_factor = Suffix(direction=Suffix.EXPORT)
+        units_meta = self.config.property_package.get_metadata().get_derived_units
+
+        # Add unit parameters
+        self.pore_diameter = Param(
```
adam-a-a, line 192: Why not make this a variable?

watertap/unit_models/selective_oil_permeation.py
```diff
@@ -0,0 +1,669 @@
...
+            mutable=True,
+            initialize=4e-8,
+            units=units_meta("length"),
+            doc="Average membrane pore diameter",
+        )
+
+        self.porosity = Param(
```
adam-a-a, line 199: Same comment- why not make it a var?

watertap/unit_models/selective_oil_permeation.py
```diff
@@ -0,0 +1,669 @@
...
+            initialize=150,
+            units=units.dimensionless,
+            doc="Permeability constant",
+        )
+
+        self.module_diameter = Param(
+            mutable=True,
+            initialize=0.064,
+            units=units_meta("length"),
+            doc="Module diameter",
```
adam-a-a, line 224: All of these can be variables that get fixed by the user after creating an instance of the unit model

watertap/unit_models/selective_oil_permeation.py
```diff
@@ -0,0 +1,669 @@
...
+            units=units_meta("length") * units_meta("time") ** -1,
+            doc="Oil volumetric flux if the feed was pure oil",
+        )
+
+        self.effective_area_ratio = Var(
+            self.flowsheet().config.time,
+            initialize=0.1,
+            bounds=(0.0, 1.0),
+            units=units.dimensionless,
+            doc="Effective area ratio",
```
adam-a-a: maybe a more descriptive doc would be good

watertap/unit_models/selective_oil_permeation.py
```diff
@@ -0,0 +1,669 @@
...
+            doc="Average transmembrane pressure",
+        )
+
+        self.liquid_velocity = Var(
+            self.flowsheet().config.time,
+            initialize=0.01,
+            bounds=(0.0, 1e3),
+            domain=NonNegativeReals,
+            units=units_meta("length") * units_meta("time") ** -1,
+            doc="Shell side liquid velocity at module inlet",
```
adam-a-a, line 348: I'd think that liquid_velocity could be calculated at the inlet and outlet, but this is set for inlet only. Maybe expand to include inlet and outlet?

watertap/unit_models/selective_oil_permeation.py
```diff
@@ -0,0 +1,669 @@
...
+        ):
+            self.deltaP = Reference(self.feed_side.deltaP)
+
+        @self.Constraint(
+            self.flowsheet().config.time,
+            doc="Average transmembrane pressure",
+        )
+        def eq_pressure_transmemb(b, t):
+            return (
+                b.pressure_transmemb[t]
```
adam-a-a:
```suggestion
                b.pressure_transmemb_avg[t]
```

adam-a-a: I'd add a qualifier here to denote that this is an average

watertap/unit_models/selective_oil_permeation.py
```diff
@@ -0,0 +1,669 @@
...
+            self.flowsheet().config.time,
+            self.config.property_package.phase_list,
+            self.config.property_package.component_list,
+            doc="Mass transfer term",
+        )
+        def eq_mass_transfer_term(b, t, p, j):
+            if j == "H2O":
+                self.feed_side.mass_transfer_term[t, p, "H2O"].fix(
+                    0
+                )  # no water transfer
```
adam-a-a, line 421: seems very ideal to assume _no_ water transfer

watertap/unit_models/selective_oil_permeation.py
```diff
@@ -0,0 +1,669 @@
...
+        )
+        def eq_permeate_isothermal(b, t):
+            return (
+                b.feed_side.properties_in[t].temperature
+                == b.properties_permeate[t].temperature
+            )
+
+        @self.Constraint(
+            self.flowsheet().config.time,
+            doc="Shell side iquid velocity",
```
adam-a-a:
```suggestion
            doc="Shell side liquid velocity",
```

watertap/unit_models/selective_oil_permeation.py
```diff
@@ -0,0 +1,669 @@
...
+        )
+        def eq_liquid_velocity(b, t):
+            return (
+                b.feed_side.properties_in[t].flow_vol_phase["Liq"]
+                == b.liquid_velocity[t] * 0.25 * Constants.pi * b.module_diameter**2
```
adam-a-a: If `liquid_velocity` doesn't get modified to include inlet and outlet conditions, I'd rename to `liquid_velocity_in`

watertap/unit_models/selective_oil_permeation.py
```diff
@@ -0,0 +1,669 @@
...
+        expr_dict = {}
+        var_dict["Membrane area"] = self.area
+        var_dict["Avg. transmembr. pressure"] = self.pressure_transmemb[time_point]
+        if hasattr(self, "deltaP"):
+            var_dict["Pressure drop"] = self.deltaP[time_point]
+        var_dict["Oil volumetric flux"] = self.flux_vol_oil[time_point]
+        var_dict["Pure oil volumetric flux"] = self.flux_vol_oil_pure[time_point]
+        var_dict["Effective area ratio"] = self.effective_area_ratio[time_point]
+        var_dict["Oil recovery"] = self.recovery_frac_oil[time_point]
+        var_dict["Shell side liquid velocity"] = self.liquid_velocity[time_point]
```
adam-a-a: Unless you change this to be velocity at inlet _and_ outlet, I'd specify this as "Shell side liquid velocity in"

adam-a-a (commented):
Nice job! I had a few comments for you to review but nothing too critical. Have a look, and I suspect after a couple tweaks, this should be ready to merge.

watertap/examples/SOP/demo_SOP.py
adam-a-a: I think so. If we are going to have an example, I think it should be a flowsheet which makes use of the prop model as well as other units, such as pumps, as well as a Feed and Product block. Alternatively, we could morph these demo files into something on the documentation side, serving as a how-to or tutorial perhaps

adam-a-a: But that could be for a future PR, if we decide to delve into this further. For this PR, I'd say remove both demo files. You should still be able to access them if ever you wanted to revamp them.

watertap/examples/SOP/demo_SOP_prop_pack.py
adam-a-a: I simply meant to delete this demo file too.

watertap/property_models/SOP_prop_pack.py
andrewlee94: Sometimes you can get away with this and get better solver behaviour (e.g. mole fractions). However, you need to ensure there is an implicit upper bound in place (e.g., if the sum of the fractions must equal one and the lower bound is zero, then all fractions have an implicit upper bound of 1).

watertap/property_models/SOP_prop_pack.py
andrewlee94: NonNegativeReals has an effective lower bound of 0 (although it is set separately from bounds in Pyomo). PsotiveReals has effective lower bound of epsilon.

watertap/property_models/SOP_prop_pack.py
```diff
@@ -0,0 +1,551 @@
...
+            self.params.phase_list,
+            self.params.component_list,
+            rule=rule_flow_vol_phase_comp,
+        )
+
+    def _flow_vol_phase(self):
+        self.flow_vol_phase = Var(
+            self.params.phase_list,
+            initialize=1e-3,
+            bounds=(None, None),
```
MarcusHolly:
```suggestion
            bounds=(0, None),
```
Edit: Just saw Andrew's response to a similar comment, so feel free to ignore these if there is an implicit upper bound in place

watertap/property_models/SOP_prop_pack.py
```diff
@@ -0,0 +1,551 @@
...
+
+        self.eq_flow_vol_phase = Constraint(
+            self.params.phase_list, rule=rule_flow_vol_phase
+        )
+
+    def _flow_mass_phase(self):
+        self.flow_mass_phase = Var(
+            self.params.phase_list,
+            initialize=1e-3,
+            bounds=(None, None),
```
MarcusHolly:
```suggestion
            bounds=(0, None),
```

watertap/property_models/SOP_prop_pack.py
```diff
@@ -0,0 +1,551 @@
...
+        super(SopStateBlockData, self).build()
+
+        self.scaling_factor = Suffix(direction=Suffix.EXPORT)
+
+        # first we create the state variables
+        self.flow_mass_phase_comp = Var(
+            self.params.phase_list,
+            self.params.component_list,
+            initialize={("Liq", "H2O"): 1, ("Liq", "oil"): 0.01},
+            bounds=(1e-15, 100),
```
MarcusHolly:
```suggestion
            bounds=(0, None),
```
I think these are the bounds that are traditionally used for flow_mass_phase_comp - unless there is a reason why these should be different

adam-a-a: I'll add here that, to my recollection, @bknueven 's tweaking of ipopt settings via ipopt-watertap, the default solver that we all use in WaterTAP now, helped enable us to use 0 instead of very small numbers (with regard to the lower bound). With regard to the upper bound, I agree that None is the better way. A user could always reset the upper bound to some desired value, subject to his/her use-case.

bknueven: We should use 0 if we mean 0 -- that is also consistent with the rest of the codebase.

watertap/property_models/SOP_prop_pack.py
tarnold17: @adam-a-a can you explain what you mean by "safer" in this context? And by "relax", do you mean to just make the upper bound higher, or to remove it altogether?

watertap/unit_models/selective_oil_permeation.py
```diff
@@ -0,0 +1,669 @@
...
+        super().build()
+
+        self.scaling_factor = Suffix(direction=Suffix.EXPORT)
+        units_meta = self.config.property_package.get_metadata().get_derived_units
+
+        # Add unit parameters
+        self.pore_diameter = Param(
+            mutable=True,
+            initialize=4e-8,
+            units=units_meta("length"),
```
MarcusHolly, line 195: It seems like our unit models aren't very consistent on this, but what's the advantage of using `units_meta` over `pyunits.m`? I assume it just allows the user to choose their own units, but what happens if they fail to specify units? Does it just default to `pyunits.m`? Ig what I'm getting at here is should we keep how we specify units consistent among unit models (it may not be worth the effort) and, if so, which alternative is better?

watertap/unit_models/selective_oil_permeation.py
tarnold17: Probably, but that is what the UT folks have told us to assume. I think this is the assumption we want to stick with for now.

TimBartholomew:
Hi guys, I feel like we should have a meeting to discuss what our objective is for developing the SOP model. I'll send out a poll via Slack.

watertap/unit_models/selective_oil_permeation.py
```diff
@@ -0,0 +1,669 @@
...
+        init_log.info("Initialization Complete: {}".format(idaeslog.condition(res)))
+
+        if not check_optimal_termination(res):
+            raise InitializationError(f"Unit model {self.name} failed to initialize")
+
+    def _get_performance_contents(self, time_point=0):
+        var_dict = {}
+        expr_dict = {}
+        var_dict["Membrane area"] = self.area
+        var_dict["Avg. transmembr. pressure"] = self.pressure_transmemb[time_point]
```
MarcusHolly:
```suggestion
        var_dict["Average transmembrane pressure"] = self.pressure_transmemb[time_point]
```
Is it worth typing this out fully?

watertap/unit_models/selective_oil_permeation.py
andrewlee94: `units_meta` takes its units from whatever property package is being used (also note there is a new, better API for this), whilst using Pyomo units picks a specific unit to use (and thus might need unit conversion latter).

watertap/unit_models/selective_oil_permeation.py
tarnold17: All of these `Param`s are things that we don't think it is likely that we will need/want to optimize over. So I opted to include them in the model as mutable `Param`s instead of as `Var`s. I figure that we can convert any of these parameters to variables later, if and when it is needed.

watertap/examples/SOP/SOP_prop_pack.py
tarnold17: I think this comment is no longer relevant. I'm going to resolve, but please reopen if I am wrong.

watertap/unit_models/selective_oil_permeation.py
tarnold17: For this model, the liquid velocity only at the inlet is needed - inlet liquid velocity is used in the correlation for effective area ratio. We could similarly calculate the outlet liquid velocity, but it's not needed for anything, so I haven't bothered to implement it.

watertap/examples/SOP/demo_SOP.py
tarnold17: Addressed in latest push

watertap/property_models/SOP_prop_pack.py
tarnold17: I changed the bounds to be `(0, None)`.

watertap/property_models/SOP_prop_pack.py
tarnold17: I changed the bounds to be `(0, 1)`.

watertap/property_models/SOP_prop_pack.py
tarnold17: I changed the bounds to be `(0, 1)` and the domain to be `NonNegativeReals`. Perhaps that's redundant, but it shouldn't hurt anything.

watertap/property_models/SOP_prop_pack.py
tarnold17: I took the suggestion and changed the bounds to `(0, None)`.

watertap/property_models/SOP_prop_pack.py
tarnold17: I changed the bounds to `(0, None)` as per Marcus' suggestion.

MarcusHolly (approved):
LGTM - looks like the only unresolved comment is the one about making things `Var` or `Param`. Personally, I would list them as variables, but perhaps leaving them as mutable parameters is fine for now. Regardless, I don't think this is something worth holding up the PR over.

adam-a-a (approved):
LGTM- @tarnold17 can you just make an issue reminding us of @MarcusHolly point that comments regarding vars or params weren't addressed?

watertap/unit_models/selective_oil_permeation.py
```diff
@@ -0,0 +1,674 @@
...
+from watertap.core import ControlVolume0DBlock, InitializationMixin
+
+
+_log = idaeslog.getLogger(__name__)
+
+
+@declare_process_block_class("SelectiveOilPermeation")
+class SelectiveOilPermeationData(InitializationMixin, UnitModelBlockData):
+    """
+    Selective oil permeation model.
```
adam-a-a: Would be good to add the reference used here

adam-a-a:
```suggestion
    Selective oil permeation model based on A.Y. Mercelat, C.M. Cooper, K.A. Kinney, F. Seibert, L.E. Katz, Mechanisms for Direct Separation of Oil from Water with Hydrophobic Hollow Fiber Membrane Contactors, ACS EST Eng. 1 (2021) 1074–1083. https://doi.org/10.1021/acsestengg.1c00055.

```

tarnold17: @adam-a-a thanks for taking care of this

PR #893, opened 2023-01-20
Adding Friction Factor Config Option

### LINKED ISSUE #890: Adding friction factor correlations for spiral wound membranes
### Description

Add a configuration argument to our membrane models that allows the selection of friction factor correlation.

### Motivation

Currently, our reverse osmosis and related membrane models use a friction factor correlation that is actually specific to a flat-sheet membrane module with specific spacer and channel geometry, from work by Guillen & Hoek. This relationship was utilized in @TimBartholomew 's paper(s) on OARO and general modeling of membrane modules; it was also used in @adam-a-a 's paper on OARO and other osmotically mediated approaches. This is the reason why we kept this friction factor correlation around until now.

However, it would make sense to add alternatives for estimating friction factor, especially ones that are meant to be more general. I propose starting with the classic relationship for spiral wound membranes based on work by Schock & Miquel, 1987. There are other key papers in the literature that we could also look to, such as, but not limited to, work by Da Costa et al, circa 1994.

### Possible Implementation

We can add a configuration argument to our membrane channel base control volume that allows specification of friction factor correlation.

### Additional Context

The classic relationship for spiral wound membranes is simply 

`f = 6.23 * Re**-0.3`

MarcusHolly:
## Fixes/Resolves:

Issue #890 

### Legal Acknowledgement


watertap/unit_models/tests/test_osmotically_assisted_reverse_osmosis_0D.py
```diff
@@ -216,6 +217,25 @@ def test_option_pressure_change_calculated():
...
+        has_pressure_change=True,
+        concentration_polarization_type=ConcentrationPolarizationType.calculated,
+        mass_transfer_coefficient=MassTransferCoefficient.calculated,
+        pressure_change_type=PressureChangeType.calculated,
+        friction_factor=FrictionFactor.spiral_wound,
+    )
+
+    assert m.fs.unit.config.friction_factor == FrictionFactor.spiral_wound
+    assert isinstance(m.fs.unit.feed_side.velocity, Var)
+    assert isinstance(m.fs.unit.feed_side.friction_factor_darcy, Var)
```
bknueven: You should test instead for the existence of the different constraint. I think you'll find that the option isn't begin propagated correctly. (Which is probably also an issue on #872...)

bknueven (changes requested):

watertap/unit_models/tests/test_osmotically_assisted_reverse_osmosis_0D.py
```diff
@@ -216,6 +217,24 @@ def test_option_pressure_change_calculated():
...
+        property_package=m.fs.properties,
+        has_pressure_change=True,
+        concentration_polarization_type=ConcentrationPolarizationType.calculated,
+        mass_transfer_coefficient=MassTransferCoefficient.calculated,
+        pressure_change_type=PressureChangeType.calculated,
+        friction_factor=FrictionFactor.spiral_wound,
+    )
+
+    assert m.fs.unit.config.friction_factor == FrictionFactor.spiral_wound
+    assert isinstance(m.fs.unit.eq_friction_factor_spiral_wound, Constraint)
```
MarcusHolly: My guess is that I just need to edit this line to `m.fs.???.eq_friction_factor_spiral_wound` since I don't think this constraint actually exists on the unit model (only in the membrane channel base), but I'm not sure what the "???" should be.

EDIT: Changing it to `m.fs.feed_side.eq_friction_factor_spiral_wound` has the following error: 
 AttributeError: '_ScalarMembraneChannel0DBlock' object has no attribute 'eq_friction_factor_spiral_wound'

watertap/unit_models/tests/test_reverse_osmosis_0D.py
```diff
@@ -193,6 +195,25 @@ def test_option_pressure_change_calculated():
...
+        concentration_polarization_type=ConcentrationPolarizationType.calculated,
+        mass_transfer_coefficient=MassTransferCoefficient.calculated,
+        pressure_change_type=PressureChangeType.calculated,
+        friction_factor=FrictionFactor.spiral_wound,
+    )
+
+    assert m.fs.unit.config.friction_factor == FrictionFactor.spiral_wound
+    assert isinstance(m.fs.unit.feed_side.velocity, Var)
+    assert isinstance(m.fs.unit.feed_side.eq_friction_factor_spiral_wound, Constraint)
+
```
adam-a-a, line 217: I think you should do the same check for the other side of the membrane since this is OARO (counterflow operation).

adam-a-a: Additionally, it'd be nice to solve the model and check the solutions for friction factor values

adam-a-a (changes requested):
Looks good- just a couple suggested tests that I think you can add.

watertap/unit_models/tests/test_reverse_osmosis_0D.py
MarcusHolly: Thanks for the suggestions. I assume you meant to point to the oaro test file and not the RO0D test file (the variable & constraint in question do not exist on the permeate side of RO0D or RO1D)

watertap/unit_models/tests/test_osmotically_assisted_reverse_osmosis_0D.py
```diff
@@ -216,6 +217,29 @@ def test_option_pressure_change_calculated():
...
+        pressure_change_type=PressureChangeType.calculated,
+        friction_factor=FrictionFactor.spiral_wound,
+    )
+
+    assert m.fs.unit.config.friction_factor == FrictionFactor.spiral_wound
+    assert isinstance(m.fs.unit.feed_side.velocity, Var)
+    assert isinstance(m.fs.unit.feed_side.eq_friction_factor_spiral_wound, Constraint)
+    assert isinstance(m.fs.unit.permeate_side.velocity, Var)
+    assert isinstance(
+        m.fs.unit.permeate_side.eq_friction_factor_spiral_wound, Constraint
```
bknueven: @adam-a-a @MarcusHolly the new OARO test function does check the permeate side ...

bknueven (approved):
@adam-a-a I think your comments have been addressed.

watertap/unit_models/tests/test_osmotically_assisted_reverse_osmosis_0D.py
MarcusHolly: Yep, I was just saying that the code Adam pointed to in his comment was for the 0D RO file (and not OARO like his comment suggested). Anyways, everything should be good to go now

watertap/unit_models/tests/test_reverse_osmosis_1D.py
```diff
@@ -881,6 +904,259 @@ def testReverseOsmosis_cp_calculated_kf_calculated(self):
             m.fs.unit.mixed_permeate[0].flow_mass_phase_comp["Liq", "NaCl"]
         )
 
+    @pytest.mark.component
+    def testRO_Pdrop_calculation(self):
+        m = ConcreteModel()
```
luohezhiming: A little bit unsure why this test is added in this PR? Don't we include it for other tests or anything changed?

MarcusHolly: Good catch - I  added it in this PR because I thought it was missing. 0D RO and 0D OARO both have a test called Pdrop_calculation that makes sure the system works when PressureChangeType is set to calculated, but `def RO_frame` actually covers that in this file.

luohezhiming: Yeah, so no need for this test, other things look good to me!

luohezhiming (approved):
Good work!

adam-a-a (approved):
LGTM- the only thing I thought of changing was the constraint name to be the same regardless of the config option selected (e.g., `eq_friction_factor`). However, that isn't critical so I won't hold this up any longer.

PR #894, opened 2023-01-20
Adds Pressure Effect into Water Specific Enthalpy

MarcusHolly:
## Summary/Motivation:
These changes are very similar to those made in PR #853 for the seawater prop pack, but in this case, the salinity terms in rule_enth_mass_phase are ignored (set equal to 0).

Reference for this new relationship (Table 9): https://www.sciencedirect.com/science/article/abs/pii/S0011916416300807

### Legal Acknowledgement


MarcusHolly:
The evaporator is failing to initialize...

I've gotten it to work by changing `m.fs.evaporator.delta_temperature_in.fix(30)` to `m.fs.evaporator.delta_temperature_in.fix(15)` in test_evaporator.py, but I'm not sure if this change makes sense.

After making this change in test_evaporator.py and test_mvc.py, the condenser in the test_mvc.py fails to initialize, and I've been unable to get this working (although I haven't tried for too long).

watertap/property_models/water_prop_pack.py
```diff
@@ -778,25 +773,30 @@ def _enth_mass_phase(self):
...
             )
-
-            if phase == "Liq":
-                return b.enth_mass_phase[phase] == h_w
-            else:  # phase == 'Vap'
-                # dh_vap_w = b.params.dh_vap_w_param_0 + b.params.dh_vap_w_param_1 * t + b.params.dh_vap_w_param_2 * t ** 2 \
-                #           + b.params.dh_vap_w_param_3 * t ** 3 + b.params.dh_vap_w_param_4 * t ** 4
-                # h_vap = h_w + dh_vap_w
-                return b.enth_mass_phase[phase] == h_w + b.dh_vap_mass
+            return b.enth_mass_phase[p] == h_w_P
```
TimBartholomew: I believe this should be:
`b.enth_mass_phase[p] == h_w_P + b.dh_vap_mass`

That should increase the specific enthalpy of the vapor by an order of magnitude.

TimBartholomew (approved):
LGTM

TimBartholomew (changes requested):
Oh I just realized, this change should be updated in the documentation

MarcusHolly:
> Oh I just realized, this change should be updated in the documentation

@TimBartholomew Oh yea, ig that talk from the last dev call is coming in handy. I'll try to be more mindful about this going forward. A potential quick fix to this would be to add a section to the PR header (Fixes, Summary/Motivation, Proposed Changes, etc.)  that asks if any documentation needs to be updated due to changes made in the PR. Just to pseudo-force people to think about documentation whenever they create a PR.

MarcusHolly:
I also just realized there is no documentation for the water prop pack... is this because it's similar enough to the seawater prop pack?

TimBartholomew:
@MarcusHolly ah yeah, ok let's merge this then but make a separate issue for creating documentation for the water property package. Basically that should reference the seawater property package, but then add the documentation for the properties that are specific for this one - mostly the vapor properties

TimBartholomew (approved):

carson-tucker (approved):

PR #896, opened 2023-01-23
Moving GAC properties to MCAS, `DiffusivityCalculation` as an MCAS argument

hunterbarber:
## Fixes/Resolves:
Adds options for diffusivity calculations
Documentation for changes in #901 

02/14 update: during resolving the above, more resolution is made to (1) unify the diffusion_phase_comp method to be all relying on a `diffuse_phase_comp = Var()`,  (2) make all diffuse_phase_comp be indexed by the keys of `diffusivity_data`, (3) keep a `diffuse_phase_comp_param = Param()` for unit models to access it from the property package directly.  The update allows only a subset of the all `solute_list` to have `diffus_phase_comp` constructed, and when this is the case, a warning will notify the user which components have or have not this property. 

## Summary/Motivation:
GAC uses the Hayduk-Laudie correlation for diffusivity of nonelectrolytes which was not in prior property package calculations
The method of switching between Param and Var for `diffus_phase_comp` based on the enumeration option is to not have to change previously established tests in this PR

- Default diffusivity option to set `diffus_phase_comp` as a Param from `diffusivity_data` Param
- Hayduk-Laudie equation as diffusivity option to calculated `diffus_phase_comp` as a Var and adding molar volume Param with accompanying `molar_volume_data`
- GAC `visc_d_phase `call from MCAS
- GAC `dens_mass_phase `call from MCAS

### Legal Acknowledgement


hunterbarber:
Tests will be failing. 
@lbibl I'm just double checking with you that the "Missing a valid diffusivity_data configuration to use EinsteinRelation to compute the "elec_mobility_phase_comp" configuration error in `test_elec_properties_errormsg` may now be redundant with the new method of diffusivity

hunterbarber:
> Tests will be failing. @lbibl I'm just double checking with you that the "Missing a valid diffusivity_data configuration to use EinsteinRelation to compute the "elec_mobility_phase_comp" configuration error in `test_elec_properties_errormsg` may now be redundant with the new method of diffusivity

Actually, the MCAS tests all passed, I may have skipped the build on the model when running locally to get an error. The error message may still be redundant, but it should be passing.

hunterbarber:
Making a note, the density of GAC tests used previously was 997 as opposed to 1000. This value is changed in this PR but the results assertions are not. The small difference did not exceed the 1e-3 error tolerance of the tests.

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -1173,8 +1208,65 @@ def rule_visc_k_phase(b, p):
...
+                            j, self.name
+                        )
+                    )
+                else:
+                    return (
+                        b.diffus_phase_comp[p, j]
+                        == self.params.config.diffusivity_data[p, j]
+                        * pyunits.m**2
+                        * pyunits.s**-1
+                    )
```
adam-a-a: What if we just keep the `add_object_reference` line here for the case when `DiffusivityCalculation` is none?

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -1173,8 +1208,65 @@ def rule_visc_k_phase(b, p):
...
+                    # self.params.config.diffus_calculation == DiffusivityCalculation.hayduklaudie
+                    diffus_coeff_inv_units = pyunits.s * pyunits.m**-2
+                    visc_solvent_inv_units = pyunits.Pa**-1 * pyunits.s**-1
+                    molar_volume_inv_units = pyunits.mol * pyunits.m**-3
+                    return (
+                        b.diffus_phase_comp[p, j] * 1e4 * diffus_coeff_inv_units
+                    ) * ((b.visc_d_phase[p] * 1e3 * visc_solvent_inv_units) ** 1.14) * (
+                        (b.molar_volume_comp[p, j] * 1e6 * molar_volume_inv_units)
+                        ** 0.589
+                    ) == 13.26e-5
```
adam-a-a: Maybe these hard-coded parameter values should be parameters or variables that are subsequently fixed (similar to how we handle coefficients in nacl and seawater prop models). Not completely sure but something to consider

adam-a-a: Should also note the reference above this

lbibl:
> Tests will be failing.
> @lbibl I'm just double checking with you that the "Missing a valid diffusivity_data configuration to use EinsteinRelation to compute the "elec_mobility_phase_comp" configuration error in `test_elec_properties_errormsg` may now be redundant with the new method of diffusivity

Were you saying the new method may create a situation where diffusivity is not referenced from the diffusivity_data?  The EinstainRelation method for electrical mobility requires an ion diffusivity input.    Nonetheless, it looks like the method you are adding is for nonelectrolytes only, so it should not be messed up with the Electrical properties that are possessed only by ions and the code should assure this.

hunterbarber:
> diffusivity

@lbibl There was some discussion on whether ionic diffusivity and molecular (nonelectrolyte) diffusivity should be the same or not. I think it was decided that since an ion can only have an ionic diffusivity and a nonelectrolyte can only have a molecular diffusivity, and they both are in reference to the diffusion coefficient in the aqueous phase and completely exclusive, they should both utilize `diffus_mass_comp`. I'll wait for your thoughts on this before further discussing the error reporting.

lbibl:
> > diffusivity
> 
> @lbibl There was some discussion on whether ionic diffusivity and molecular (nonelectrolyte) diffusivity should be the same or not. I think it was decided that since an ion can only have an ionic diffusivity and a nonelectrolyte can only have a molecular diffusivity, and they both are in reference to the diffusion coefficient in the aqueous phase and completely exclusive, they should both utilize `diffus_mass_comp`. I'll wait for your thoughts on this before further discussing the error reporting.

It's inaccurate to say "ion can only have an ionic diffusivity".   I see a missing point on this matter, which is the complexity of the diffusion of a "multi-component" system.   Diffusion is essentially the mechanism of the Brownian motion phenomenon, and is described through some models, the simplest of which is Fick's law where a single diffusivity (diffusion constant) of a SINGLE particle (ion, molecule, or others) is determined.  This said, Fick's law is not applicable to a muti-component system and it is also mechanistically incorrect to simply add up "D*detlaC" to express diffusion mass transfer of multiple components in a system.  This can be a core issue for us to move forward with multi-component systems, and underlies one previous comment I made that diffusivity-related quantities (e.g., Schmidt number) should not be indexed by components.  A common diffusion model for a multiple-component system is the Maxwell-Stefan model (https://en.wikipedia.org/wiki/Maxwell–Stefan_diffusion).  This model is comprehensive but unfortunately (and realistically for many theories) uneasy to use due to the difficulty of getting the Maxwell-Stefan diffusion constant there.  For a bi-component system, there is a simpler equation to relate the individual particle's diffusivity to a general/molecular diffusivity, but when it comes to there or more components, I am concerned the pure theoretical approach will be quite limited and one can only rely on empirical/experimental numbers.

lbibl:
Aside from this concern, for what you are doing here, I don't see the urgency of solving this problem.  Since you are just adding a method for non-electrolyte, you can just add a control to let the method only applicable to non electrolyte, current the solute_set or maybe, in the future, the neutral_set.  Another reason to make the indexing accurate.

hunterbarber:
> Aside from this concern, for what you are doing here, I don't see the urgency of solving this problem. Since you are just adding a method for non-electrolyte, you can just add a control to let the method only applicable to non electrolyte, current the solute_set or maybe, in the future, the neutral_set. Another reason to make the indexing accurate.

I think this is a good idea, I can let you know once some revisions are made. Thanks.

adam-a-a:
@lbibl I see your concern, but we should also consider that this property model was originally meant for the DSPM-DE model, and in that model, we wanted bulk diffusivity values of individual components, which can be ions or neutral solutes. In that exception, I wouldn't see harm in allowing the user to either provide bulk diffusivity values (as we do already) and/or estimate such values in the absence of immediate data. That doesn't necessarily mean that the diffusivity values of the property model are meant to account for interactions between all other components. 

Overall, I think the way to handle this is to explain in documentation that calculated values _would not_ account for "diffusion mass transfer of multiple components in a system." Thanks for the bringing up this point. I think we should definitely be clear about this with caveats in the documentation and maybe even through logger warnings.

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -1186,8 +1229,42 @@ def rule_visc_k_phase(b, p):
...
+                units=pyunits.m**2 * pyunits.s**-1,
+                doc="Molecular diffusivity of nonelectrolytes",
+            )
+
+            def rule_diffus_phase_comp(b, p, j):
+                # (Hayduk & Laudie, 1974)
+                diffus_coeff_inv_units = pyunits.s * pyunits.m**-2
+                visc_solvent_inv_units = pyunits.Pa**-1 * pyunits.s**-1
+                molar_volume_inv_units = pyunits.mol * pyunits.m**-3
+                return (b.diffus_phase_comp[p, j] * 1e4 * diffus_coeff_inv_units) * (
```
kurbansitterley: You might get push back on having hard-coded parameters here.

kurbansitterley:
@lbibl and @adam-a-a would be nice to have a review/decision on these changes since MCAS is used for both GAC and IX and @hunterbarber and I should be using the same version of the prop pack for our PFAS analysis.

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -1186,8 +1229,43 @@ def rule_visc_k_phase(b, p):
...
+                ) * (
+                    (b.molar_volume_comp[p, j] * 1e6 * molar_volume_inv_units) ** 0.589
+                ) == 13.26e-5
+
+            self.eq_diffus_phase_comp = Constraint(
+                self.params.phase_list,
+                self.params.solute_set,
+                rule=rule_diffus_phase_comp,
+            )
+        # TODO: add warning for use of ionic vs molecular diffusivity when ion+solute set > 1
```
lbibl: A few insights: 1) I doubt the units are treated accurately here.  A var already has a unit when it is created and it seems you multiplied the original unit with another unit.   Maybe using the unit conversion function is the correct way to go.  Also when a var has a power in the eq, its unit will carry the power too, so you may need to account for the units of all coefficients in the equation.  To assure everything is correct, you may wanna run an `assert_units_consistent` in the test.

lbibl: 2) I quickly searched this Hayduk laudie equation you are implementing-  I saw some inconsistency in wrt the unit of "molar volume" or "molal volume" in different places, with units of cm^3/mol or cm^3/(mol g).  Would you make sure the accuracy of what is used ?   This equations seem to be an empirical correlation upon 87 substances (claimed in their original paper); would this mean you can only apply it for these 87 compounds? I am curious how useful is this method in practice.

lbibl: Overall, I would probably always encourage the use of a constant diffusion constant input at this time, but I wouldn't object having this method in place.  Just be very careful with the unit issue and precise documentation I think.

hunterbarber: @lbibl I added a dof and units check to the MCAS test in the latest commit. 

With the HL correlation, the coefficient units are effectively meaningless because the coefficient values were determined w.r.t. the specific units shown in the image here. I wanted to keep the hard coded coefficient values the same as the HL correlation for clarity.
image
Therefore, I multiply the Var/Param (and units with the Var/Param) within the correlation by the inverse of the correct units and any necessary unit conversion to create dimensionless multiplicands and ensure the output is correct.

It's correct that only some molar volume values are tabulated but you can also estimate molar volume from the LaBas method using elements/functional groups of any species. This is actively being done in the pfas_treatment_analysis branch as ongoing work and the order of magnitude for the diffus_phase_comp is at least correct. The lack of data availability for PFAS species was the motivation for having a systematic method for estimating unknown diffusivities.

hunterbarber: @lbibl there's a new implementation if you can check it at some point

lbibl:
@hunterbarber : I found the previous method design wouldn't allow the case where a mixture of electrolyte and non-electrolyte is passed to the property package.  I hope you wouldn't mind me pushing some improvement directly.  The change I made and tested let the property to use your method (h_l) to calculate D for nonelectrolyte and still retrieve D data from the `diffusion_data` configure, when h_l method is chosen.  This also cleared a TODO note you posted.  This does NOT modify any of the method you coded.  Please take a look and let me know if you find any issue.  I will look more into your changes in the method later.

hunterbarber:
> @hunterbarber : I found the previous method design wouldn't allow the case where a mixture of electrolyte and non-electrolyte is passed to the property package. I hope you wouldn't mind me pushing some improvement directly. The change I made and tested let the property to use your method (h_l) to calculate D for nonelectrolyte and still retrieve D data from the `diffusion_data` configure, when h_l method is chosen. This also cleared a TODO note you posted. This does NOT modify any of the method you coded. Please take a look and let me know if you find any issue. I will look more into your changes in the method later.

@lbibl  Thanks for collaborating on this, and yes you were correct about the previous short-term implementation. The only thing I would say is right now technically having `molar_volume_data` provided and specifying `DiffusivityCalculation.HaydukLaudie` is redundant from a coding standpoint. `molar_volume_data` supersedes the option by exclusively applying for species with `molar_volume_comp['liq', j]` to solve via HaydukLaudie, meaning it could be coded without the option all together and being implied by providing `molar_volume_data`. Though `molar_volume_data` will likely be used for other properties later, I think it is actually better this way.

Also, does using diffus_phase_comp_nonhl as a temp variable make it persist in later model statistics? 

I'll adjust the GAC tests when a get a chance as well, I think it's just the case sensitive naming convention that's causing tests to fail. The case sensitive naming in configuration options is inconsistent across codes from what I've seen.

watertap/property_models/multicomp_aq_sol_prop_pack.py
hunterbarber: @lbibl Another niche scenario that I'm curious of your input.

The GAC model is single species adsorption, denoted with a target_species argument in the unit model. In the past version when DiffusivityCalculation.HaydukLaudie, the diffus_phase_comp was only indexed with solute_set (and not ion_set). Therefore, background ions did not have a calculated diffusivity and did not require data. 

Now with accepting fixed and calculated by hl diffusivities when DiffusivityCalculation.HaydukLaudie, data **must** be provided for background compounds even though not required by the GAC model.

I added arbitrary diffusivities for the background components in the GAC tests. They are never used by the model but now required for the property package requirements to be fulfilled. I can see either just documenting this clearly or adding a warning that _no diffusivity data was provided for species  j, value is defaulted to 1e-5_ or something similar.

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -330,14 +362,41 @@ def build(self):
...
+            )
+        elif self.config.diffus_calculation == DiffusivityCalculation.HaydukLaudie:
+            hl_ind = [i[1] for i in self.config.molar_volume_data.keys()]
+            nonhl_ind = [i[1] for i in self.config.diffusivity_data.keys()]
+            if not (hl_ind + nonhl_ind) == self.ion_set | self.solute_set:
+                raise ConfigurationError(
+                    "Insufficient data is provided to determine diffusivity of all solute components. "
+                    "Check the molar_volume_data and diffusivity_data configuration."
+                )
+            self.diffus_phase_comp_nonhl = Param(
```
adam-a-a: Why give this a different name?

adam-a-a: Nevermind- I see you later specify the variable in the on-demand method.

andrewlee94: A warning for this however - IDAES v2.0 will be a lot more strict about objecting to non-standard naming of properties. You will have to explicitly declare non-standard properties as custom properties once you update to v2.0.

See line 2175 for an example: https://github.com/IDAES/idaes-pse/blob/main/idaes/models/properties/general_helmholtz/helmholtz_functions.py

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -330,14 +362,41 @@ def build(self):
...
+                self.ion_set | self.solute_set,
+                mutable=True,
+                default=1e-9,
+                initialize=self.config.diffusivity_data,
+                units=pyunits.m**2 * pyunits.s**-1,
+                doc="Bulk diffusivity of solute components",
+            )
+        elif self.config.diffus_calculation == DiffusivityCalculation.HaydukLaudie:
+            hl_ind = [i[1] for i in self.config.molar_volume_data.keys()]
+            nonhl_ind = [i[1] for i in self.config.diffusivity_data.keys()]
```
adam-a-a: Add comments above each to spell out what you're doing (assuming hl_ind gives you components in solute_set and nonhl_ind give ion_set).

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -330,14 +362,41 @@ def build(self):
...
+            self.solute_set,
             mutable=True,
-            default=1e-9,
-            initialize=self.config.diffusivity_data,
-            units=pyunits.m**2 * pyunits.s**-1,
-            doc="Bulk diffusivity of ion",
+            default=1e-5,
+            initialize=self.config.molar_volume_data,
+            units=pyunits.m**3 / pyunits.mol,
+            doc="pure component density of solutes",
```
adam-a-a: Pure component density or Molar volume?

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -1186,8 +1246,81 @@ def rule_visc_k_phase(b, p):
...
         add_object_reference(self, "radius_stokes_comp", self.params.radius_stokes_comp)
 
+    def _molar_volume_comp(self):
+        add_object_reference(self, "molar_volume_comp", self.params.molar_volume_comp)
+
     def _diffus_phase_comp(self):
-        add_object_reference(self, "diffus_phase_comp", self.params.diffus_phase_comp)
+        if self.params.config.diffus_calculation == DiffusivityCalculation.none:
+            add_object_reference(
+                self, "diffus_phase_comp", self.params.diffus_phase_comp
```
adam-a-a: In the future, we might want to consider making diffusivity (and the other parameters that are meant to be populated by user data) a variable that is immediately fixed after declaration, setting things up for parameter estimation later on if desired. 

That is beyond the scope of this PR.

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -1186,8 +1246,81 @@ def rule_visc_k_phase(b, p):
...
+            self.params.config.diffus_calculation == DiffusivityCalculation.HaydukLaudie
+        ):
+            hl_ind = [i[1] for i in self.params.config.molar_volume_data.keys()]
+            nonhl_ind = [i[1] for i in self.params.config.diffusivity_data.keys()]
+            self.diffus_phase_comp = Var(
+                self.params.phase_list,
+                self.params.solute_set | self.params.ion_set,
+                initialize=self.params.config.diffusivity_data,
+                units=pyunits.m**2 * pyunits.s**-1,
+                doc="mass diffusivity of neutral and ionic speices",
```
adam-a-a:
```suggestion
                doc="mass diffusivity of neutral and ionic species",
```

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
```diff
@@ -1250,11 +1251,6 @@ def model6():
             ("Liq", "N"): 1.5e-9,
         }
     }
-    dic_elec_mob = {
-        "elec_mobility_data": {"Na_+": 5.19e-8, "Cl_-": 7.92e-8},
-    }
-    dic_transnum = {"trans_num_data": {("Liq", "Na_+"): 0.4, ("Liq", "Cl_-"): 0.6}}
-    dic_equivcond = {"equiv_conductivity_phase_data": {"Liq": 0.01}}
```
adam-a-a, line 1272: Why remove?

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
```diff
@@ -1341,3 +1337,89 @@ def test_elec_properties_errormsg(model6):
...
+
+@pytest.mark.unit
+def test_diffus_properties_errormsg():
+    m = ConcreteModel()
+    m.fs = FlowsheetBlock(dynamic=False)
+    with pytest.raises(
+        ConfigurationError,
+        match=(
+            "Insufficient data is provided to determine diffusivity of all solute components. "
+            "Check the molar_volume_data and diffusivity_data configuration."
```
adam-a-a: I think it would be beneficial for this to be more descriptive. For example, the user needs to add diffusivity data. Can you make this more digestible for the novice user he may not know what this means exactly without digging?

adam-a-a: Also should diffusivity data only be supplied for ions? What happens if the user provides diffusivity data for all components and tries to use this calc method? It doesn't seem logical that one would want to do that, but it could happen. What error crops up when that happens?

watertap/unit_models/tests/test_gac.py
```diff
@@ -524,7 +535,16 @@ def gac_frame_multi(self):
                 "BGAN": 0.1,
             },
             charge={"BGCAT": 1, "BGAN": -2},
+            diffus_calculation=DiffusivityCalculation.HaydukLaudie,
+            molar_volume_data={("Liq", "TCE"): 9.81e-5},
+            diffusivity_data={
+                ("Liq", "BGSOL"): 1e-5,
```
adam-a-a, line 544: So the user could provide diffusivity data for neutral solutes to initialize essentially. Is that right?

adam-a-a (changes requested):
I had a couple comments/questions to address.

watertap/property_models/multicomp_aq_sol_prop_pack.py
bknueven: I would also make the check more explicit, e.g., check `hl_ind` against `solute_set` and `nonhl_ind` against `ion_set`. Not sure intermediate lists are needed either.

What is stored in `i[0]`?

watertap/property_models/multicomp_aq_sol_prop_pack.py
hunterbarber:
```suggestion
            doc="molar volume of solutes",
```

watertap/property_models/multicomp_aq_sol_prop_pack.py
lbibl: We were discussing around a situation where we may allow NOT all components in the given `solute_list` to possess a `diffus_phase_comp` property.  I am having about-ready  commits to realize this but have one pending decision to make-  for those which won't have a diffusivity data, do we not build the property at all or have then lingering with default values.  I am in favor of not building them to avoid confusions or potential mistakes.  Maybe we can try deciding this in the dev meeting if a few min of discussion would be enough.

lbibl: `I[0]` is `"Liq"` (of course ... :))

>

watertap/property_models/multicomp_aq_sol_prop_pack.py
lbibl: I am in favor of this and should be able to make it happen quickly in the next push. we are making it be `var` in the calculation method already.

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
lbibl: Sorry, these were some redundant vars that were not used due to directly pasting by me in the past...

watertap/property_models/multicomp_aq_sol_prop_pack.py
andrewlee94: I would recommend caution on this - whilst it makes sense if you know what is going on, if makes things harder for downstream models as they may not know which components have a diffusivity and then they need to know what to do if one does not exist.

watertap/unit_models/tests/test_gac.py
lbibl: That'll be overwritten by the calculation method since the config above.  But if we set config as `none`, neutral species can get values from `diffusivity_dt`

adam-a-a: `diffusivity_dt`?

adam-a-a: I guess my point is the following should be clear to the user: 
(1) What exactly am I intended to enter when choosing this calculation method?
(2) What are the repercussions of not entering my data the "typical" way? For example, do I get a warning that my value will be overwritten?
(3) What was my error if I did enter something incorrectly?

I know we raise one exception already. Should there be others? If not, should there be warnings so that the user is aware of what's happening? Before, things were more straight forward since we were just supplying values for our components as needed. Now, it could be a mix of supplied values and calculated, so there could be unintended consequences that we should make sure to cover.

watertap/property_models/multicomp_aq_sol_prop_pack.py
adam-a-a: Agreed. Since it'll likely require changes across more files (tests), I figured this incremental change would be a good stepping stone for the time being.

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -330,14 +362,14 @@ def build(self):
             units=pyunits.m,
             doc="Stokes radius of solute",
         )
-        self.diffus_phase_comp = Param(
-            self.phase_list,
-            self.ion_set | self.solute_set,
+        self.molar_volume_comp = Param(
+            ["Liq"],
```
lbibl: Consider changing this to `phase_list`

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -1186,8 +1227,133 @@ def rule_visc_k_phase(b, p):
...
+                        (
+                            pyunits.convert(
+                                b.molar_volume_comp[p, j],
+                                to_units=pyunits.cm**3 * pyunits.mol**-1,
+                            )
+                            * molar_volume_inv_units
+                        )
+                        ** b.hl_molar_volume_coeff
+                    ) == b.hl_diffus_cont
+
```
lbibl, line 1352: So basically, you cancelled all units by timing their reciprocals.  Would an easier way to do this  be using a `value()` to save some computation?   This approach is mathematically correct. The alternative can be accounting for the units of all multiplier and power numbers to make sure units are balanced, which would be consistent to the rest of constraints or idaes designs.  I am fine with either way.

watertap/unit_models/tests/test_gac.py
lbibl: (1) Dict of diffusivity_data and/or molar_volume_data with correct indices.  This should be a documentation job.  (2) Yes in the current version. (3) Typical "enter something incorrectly" would be calling some component's `diffus_phase_comp` that is not constructed.  Error massage in this case is clearly as "Index '('Liq', 'xxx')' is not valid for indexed component ...", invoked from the `_validate_index` function in `indexed_component.py` in promo.   

When the `DiffusivityCalculation.HaydukLaudie` is selected and `diffusivity_data` is provided for some components, a warning will show the latter will be overridden by the former.

watertap/property_models/multicomp_aq_sol_prop_pack.py
lbibl: Other than a second thought on the DiffusivityCalculation.HaydukLaudie's method unit treatment for @hunterbarber to consider, I think this PR is good to go.

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
lbibl: These lines are removed in the update.

watertap/property_models/multicomp_aq_sol_prop_pack.py
lbibl: All is carried in this PR now.

watertap/property_models/multicomp_aq_sol_prop_pack.py
lbibl: Via our meeting discussion, we decided to index `diffus_phase_comp` by what is provided in the data and calling those whose data is unavailable will trigger an indexing error.  This is implemented in the current version.

watertap/property_models/multicomp_aq_sol_prop_pack.py
hunterbarber, line 366:
```suggestion
            self.phase_list,
```
Current use of molar volume in H-L is liquid only but I checked and other phases are valid so I agree.

hunterbarber: @lbibl I double checked and technically can be of any phase so it would be good for future usage, though the use in the H-L correlation for diffusivity is "molar volume at normal boiling point for solute." So then if changing the indexing to include `phase_comp`, would you think the naming convention across usage should also change to `molar_volume_phase_comp`? Note that the GAC model does call `properties_in[0].diffus_phase_comp["Liq", j]` so it should only call for the liquid input.

lbibl: @hunterbarber You are right on this. For the time being It's probably trival because the property pack only has one phase "Liq" anyways.  Many vars in this file used to be hard-coded with "Liq" but at some point we discussed and decided to generalize the `phase_list` index for future's adaptability as you said.  I think changing it to `molar_volume_phase_comp` (and use `phase_list` as index) is good for aligning it with others like `diffuse_phase_comp`.  If not this way, you may just delete the ["Liq"]  here to make it a single index.

lbibl:
@hunterbarber :  just to draw your attention to #860 .  #860 is now free of fails and I expect it to be merged soon.  I recommend this PR go after #860 since we'll need some changes in regards to the `solute_set` after #860 is settled and making the changes here would be easier I think.  I also invited you to review #860.

watertap/property_models/multicomp_aq_sol_prop_pack.py
hunterbarber: Addressed in commit 2d72f93e

watertap/property_models/multicomp_aq_sol_prop_pack.py
hunterbarber: @lbibl I did not use `value()` when the H-L equation was implemented in the unit model because it ran the risk of incorrect calculations when `units_meta` was not as expected. However, the property packages have hardcoded units which the unit models directly adopt, so this doesn't risk the issue. Still, I think this method has less ambiguity for what units are correct for the parameters used in the equation, and holds true if for some reason the units defined at the var declaration are ever changed.

hunterbarber:
To update, this is just waiting on [#860](https://github.com/watertap-org/watertap/pull/860). No other changes are pending to address until then.

hunterbarber:
> To update, this is just waiting on [#860](https://github.com/watertap-org/watertap/pull/860). No other changes are pending to address until then.

This has been merged, this PR is now ready for review.

lbibl (approved):
Codes look good to me.

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -1199,8 +1240,135 @@ def rule_visc_k_phase(b, p):
...
+                    "No diffusivity_data or molar_volume_data is provided for {}; "
+                    "there will be no diffus_phase_comp properties for these components.".format(
+                        missing_diffus_ind
+                    )
+                )  # warning for components with neither diffusivity_data nor molar_volume_data entry.
+            common_ind = [i for i in self.mv_dt_ind if i in self.diffus_dt_ind]
+            _log.warning(
+                "Both diffusivity_data and molar_volume_data are provided for {}; "
+                "the diffus_phase_comp property will be calculated based on their molar_volume_data "
+                "for these components becuase the HaydukLadie method is selected.".format(
```
adam-a-a:
```suggestion
                "for these components because the Hayduk-Ladie method is selected.".format(
```

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -1199,8 +1240,135 @@ def rule_visc_k_phase(b, p):
...
+                initialize=1e-9,
+                units=pyunits.m**2 * pyunits.s**-1,
+                doc="mass diffusivity of solute components",
+            )
+        elif (
+            self.params.config.diffus_calculation == DiffusivityCalculation.HaydukLaudie
+        ):
+            missing_diffus_ind = [
+                i
+                for i in (self.params.ion_set | self.params.solute_set)
```
adam-a-a:
```suggestion
                for i in self.params.solute_set
```
Since PR #860 went through, I believe this can work instead.

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -370,6 +402,14 @@ def build(self):
             units=pyunits.dimensionless,
             doc="Ion charge",
         )
+        self.diffus_phase_comp_param = Param(
```
adam-a-a: Do we actually need this param? It looks like it isn't used anywhere, except when referencing diffusivity in unit models, which I think doesn't have to be the case. The reason is that you create the `diffus_phase_comp` variable and fix it when `DiffusivityCalculation` is `None`.

Originally, I said that we could later on update `diffus_phase_comp` to be a variable that gets fixed instead of a parameter, but line 1329 already does the job. So, I would suggest removing this Param, unless i am missing a good reason for its inclusion.

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -1199,8 +1240,135 @@ def rule_visc_k_phase(b, p):
...
 
+    def _molar_volume_phase_comp(self):
+        add_object_reference(
+            self, "molar_volume_phase_comp", self.params.molar_volume_phase_comp
+        )
+
     def _diffus_phase_comp(self):
-        add_object_reference(self, "diffus_phase_comp", self.params.diffus_phase_comp)
+        # Retrieve two components sets from the molar_volume_data and diffusivity_data configurations
+        self.mv_dt_ind = Set(
```
adam-a-a:
```suggestion
        self.molar_volume_data_indices = Set(
```
It's longer and seems less wieldy, but it is also more readable more users/developers and we try to take that route whenever possible.

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -1199,8 +1240,135 @@ def rule_visc_k_phase(b, p):
...
+            initialize=[i[1] for i in self.params.config.molar_volume_data.keys()]
+        )
+        self.diffus_dt_ind = Set(
```
adam-a-a:
```suggestion
        self.diffus_data_indices = Set(
```

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -1199,8 +1240,135 @@ def rule_visc_k_phase(b, p):
...
+            initialize=[i[1] for i in self.params.config.diffusivity_data.keys()]
+        )
+        if self.params.config.diffus_calculation == DiffusivityCalculation.none:
+            missing_diffus_ind = [
+                i
+                for i in (self.params.ion_set | self.params.solute_set)
```
adam-a-a:
```suggestion
                for i in self.params.solute_set
```

watertap/property_models/multicomp_aq_sol_prop_pack.py
lbibl: Initially I removed-  The absence of it created some errors where a unit calls this param by `properties.params.diffus_phase_comp` directly (as against calling it by some state_block).  One example is in the nanofiltration file (where this param is retained, while I modified those in the ED files).   So I added this param back and thought it might not be a bad idea to have it staying out of state blocks for easy calling when needed.

watertap/property_models/multicomp_aq_sol_prop_pack.py
lbibl: yes, I forgot to change this.  I'll push a commit.

watertap/property_models/multicomp_aq_sol_prop_pack.py
lbibl: The `-` was not written to be consistent to the name in the Method Class, where it is named HaydukLadie, I would keep them consistent regardless of having - or not

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -1199,8 +1240,135 @@ def rule_visc_k_phase(b, p):
...
+                    )
+                )  # warning for components with neither diffusivity_data nor molar_volume_data entry.
+            common_ind = [i for i in self.mv_dt_ind if i in self.diffus_dt_ind]
+            _log.warning(
+                "Both diffusivity_data and molar_volume_data are provided for {}; "
+                "the diffus_phase_comp property will be calculated based on their molar_volume_data "
+                "for these components becuase the HaydukLadie method is selected.".format(
+                    common_ind
+                )
+            )  # warning for components whose diffusivity_data will be overriden by the HaydukLaudie method.
```
adam-a-a:
```suggestion
            )  # warning for components whose diffusivity_data will be overridden by the HaydukLaudie method.
```

watertap/property_models/multicomp_aq_sol_prop_pack.py
adam-a-a: Hmmm... If that is the case, I would prefer just declaring `diffus_phase_comp` as a variable here and fixing it, when DiffusivityCalculation is None. Then in the method, you can add object reference and/or do Constraint.Skip when calculation type is none (or something like that--didn't think it through).

watertap/property_models/multicomp_aq_sol_prop_pack.py
adam-a-a:
```suggestion
                "for these components becuase the HaydukLaudie method is selected.".format(
```
Fair enough- still needs a "u" though.

watertap/property_models/multicomp_aq_sol_prop_pack.py
lbibl: That is already implemented.  That's why the param now has a different name.  If you wanna remove the param, could you go ahead to modify the nanofiltration file to let the diffusivity called by the accurate `state_block` where applicable?   You can also simply delete `diffus_phase_comp_param` along as way.

lbibl: Honestly I like the way it is so sometimes a unit can call a constant diffusivity without bothering to decide which state_block to put it on.  Such a situation is often encountered.   If we remove this param completely, I don't see why we keep all the other parameters especially if "consistency" is so liked.

adam-a-a: > That is already implemented. That's why the param now has a different name. If you wanna remove the param, could you go ahead to modify the nanofiltration file to let the diffusivity called by the accurate `state_block` where applicable? You can also simply delete `diffus_phase_comp_param` along as way.

I meant for it to be implemented here on the parameter block, not where it is now.

adam-a-a: The way I am suggesting would allow the user to just call the `params.diffus_phase_comp` instead of pointing to a state block _or_ pointing to `params.diffus_phase_comp_param`. In a subsequent PR, we can convert other parameters to variables that get fixed on the parameter block, but I didn't want to hold up the PR much longer.

adam-a-a (approved):
I added some commits to make the last set of changes that I had in mind. I am approving the PR but won't merge it until at least @hunterbarber or @lbibl have a look and are fine with what I did.

EDIT: @bknueven just in case, please hold on enabling automerge =)

lbibl:
Latest modifications by @adam-a-a let components not in `diffusivity_data` or `molar_volume_data` list obtain a `diffus_phase_comp` value of 1e-9 by default under method none, posing an error risk. It is different from the decision we early made .  I recommend [7933717](https://github.com/watertap-org/watertap/pull/896/commits/7933717b35d428ec3d852c14091a54b5aa4ff6fe)  but fine with the latest version.

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
```diff
@@ -1310,26 +1309,30 @@ def model6():
 @pytest.mark.unit
 def test_elec_properties_errormsg(model6):
```
lbibl: Why does this test need to be changed/removed? I think there should be no interference to electrical property tests by the changes on diffusivity.

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
```diff
@@ -1310,26 +1309,30 @@ def model6():
...
+    #         solute_list=["Na_+", "Cl_-", "N"],
+    #         mw_data={"H2O": 0.018, "Na_+": 0.023, "Cl_-": 0.0355, "N": 0.01},
+    #         charge={"N": 0, "Na_+": 1, "Cl_-": -1},
+    #     )
     m[0].fs.properties = MCASParameterBlock(
         solute_list=["Na_+", "Cl_-", "N"],
         mw_data={"H2O": 0.018, "Na_+": 0.023, "Cl_-": 0.0355, "N": 0.01},
         charge={"Na_+": 1, "Cl_-": -1},
         elec_mobility_calculation=ElectricalMobilityCalculation.EinsteinRelation,
+        diffus_calculation=DiffusivityCalculation.HaydukLaudie,
```
lbibl: I don't see why this is added, only if you are writing a new test  but what is about the original test?

lbibl:
I am concerned with the modifications made in this test:  @pytest.mark.unit
def test_elec_properties_errormsg(model6) L1310.   I don't see interferences of changes in this PR to this test.  I would clarify this before merging.

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
adam-a-a: Well, first let me explain why I included a default value for diffus_phase_comp when DiffusivityCalculation is None. Before, as a parameter, we could get away with the user not providing a value, but when I add_object_reference within the diffus_phase_comp method for the stateblock, an error will occur. Thus, I decided for calculation type of None, it wouldn't hurt to assign a default and provide a logger warning. I get why we think that might be risky, but I'd prefer this way rather than having separately named params/vars to point to when choosing between diffus_phase_comp on stateblock or parameter block.

However, I preserved the functionality behind Hayduk-Laudie. So with no diffus calculation, this error message below will not be raised because default values are provided in absence of a user-provided value (again, because it is not a fixed var on parameter block that is also referenced in the associated on-demand method). 

By specifying Hayduk-Laudie, the error message _will_ be raised, since no defaults need be supplied.

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
adam-a-a: Oh, I forgot to uncomment this. I was probing issues with test after when trying to figure out why the error message was no longer raised. Nice catch

adam-a-a:
> Latest modifications by @adam-a-a let components not in `diffusivity_data` or `molar_volume_data` list obtain a `diffus_phase_comp` value of 1e-9 by default under method none, posing an error risk. It is different from the decision we early made . I recommend [7933717](https://github.com/watertap-org/watertap/pull/896/commits/7933717b35d428ec3d852c14091a54b5aa4ff6fe) but fine with the latest version.

Actually, that is only true for when DiffusivityCalculation is None and thus when `diffusivity_data` does not include values for solutes. The `molar_volume_data` list has no relevance to this calculation type of None. In fact, this is in line with the previous implementation which is now on main. That is, on main, `diffus_phase_comp` was simply a Param with a default value of 1e-9. This replicates that representation and provides a logger warning explaining such. I find it no less risky than overwriting user provided values with the Hayduk-Laudie method.

Overall, I want to preserve `diffus_phase_comp` on both the parameter block and stateblock when DiffusivityCalculation is None.

adam-a-a:
> I am concerned with the modifications made in this test: @pytest.mark.unit def test_elec_properties_errormsg(model6) L1310. I don't see interferences of changes in this PR to this test. I would clarify this before merging.

Thanks- I just forgot to bring the test back after debugging. I've fixed this.

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
lbibl: I don't think that is how the electrical method is written- error message was controlled by `diffusivity_data` in the configuration rather than any parameter or vars.  The error should show in any methods to assure the correctness of the electrical property calculation.

lbibl:
> Overall, I want to preserve `diffus_phase_comp` on both the parameter block and stateblock when DiffusivityCalculation is None.

The above is true in [7933717](https://github.com/watertap-org/watertap/pull/896/commits/7933717b35d428ec3d852c14091a54b5aa4ff6fe) .   I checked that no error to use the same name for param and var in that commit, although no harm or some benefit to distinguish them.  I think the question is whether to revert a decision we made earlier.  I recommend no but ok with yes.

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
lbibl: > I get why we think that might be risky, but I'd prefer this way rather than having separately named params/vars to point to when choosing between diffus_phase_comp on stateblock or parameter block.

1) I don't think naming is more important than avoiding a risk here; 2) the name could be changed to the same in the previous version; 3) we decided to eliminate the risk and was done in the previous version; 4) diffusivity is critical, e.g., is related to electrical migration properties- it had occurred that the default values messed up some calculations in ED models that were later caught by me. I imagine this can easily happen to users so I still recommend eliminating it.

adam-a-a:
At this point, I'm going to say we should go with what we have now. I did a bit of general cleanup in my commits as well that I don't think we should revert. If issues arise later, we can always raise an issue and fix things. However, the code seems fine at the moment. I'll pause on merging until @hunterbarber sees the changes, but I don't want to hold up the PFAS effort any longer, which depends on this.

lbibl:
> > I am concerned with the modifications made in this test: @pytest.mark.unit def test_elec_properties_errormsg(model6) L1310. I don't see interferences of changes in this PR to this test. I would clarify this before merging.
> 
> Thanks- I just forgot to bring the test back after debugging. I've fixed this.

The original functionality of the quoted test remains altered due to the addition of L1326.  The test was written to assure a series of electrical properties are correctly calculated for ionic/electrolyte components while `HaydukLaudie` is only applicable to non-electrolyte components.  Adding L1326 essentially nullifies the test and probably suggests some electrical property calculation methods were broken.

hunterbarber:
> At this point, I'm going to say we should go with what we have now. I did a bit of general cleanup in my commits as well that I don't think we should revert. If issues arise later, we can always raise an issue and fix things. However, the code seems fine at the moment. I'll pause on merging until @hunterbarber sees the changes, but I don't want to hold up the PFAS effort any longer, which depends on this.

@adam-a-a All the changes seem good to me, I think it is good to merge. Keep me updated if subsequent PR's come out to further this effort or the entirety of the declaration of param/var needs more input.

adam-a-a:
@lbibl even though this was ready to go, I wanted to make all parties happy. In my latest commits, I remove the default value assignments for diffusivity, reverted your elec prop error message test, and revised the _diffus_phase_comp method to make it work. Also noticed our scaling utility needed a conditional since I do not include `eq_diffus_phase_comp` for no calculation of diffusivity (only add_object_reference)

lbibl:
On my local, the following codes hit an error at L 2142 in multicomp_aq_sol_prop_pack.py  , with message "property metadata method does not contain a method for eq_diffus_phase_comp" 

`solver = get_solver
m = ConcreteModel()
    solute_list=["A", "B", "C", "D", "E", "F", "G", "H"],
    charge={"E": 1, "F": -1},
    diffus_calculation=DiffusivityCalculation.none,
    molar_volume_data={
        ("Liq", "A"): 96e-6,  # tested for benzene
        ("Liq", "B"): 100e-6,  # arbitrary
        ("Liq", "C"): 60e-6,  # arbitrary
        ("Liq", "D"): 200e-6,  # arbitrary
    },
    diffusivity_data={("Liq", "D"): 1e-11, ("Liq", "E"): 1.33e-9, ("Liq", "F"): 2.03e-9},
)
m.fs.properties.visc_d_phase["Liq"] = 1.0e-3

m.fs.sb = m.fs.properties.build_state_block([0], defined_state=True)
iscale.calculate_scaling_factors(m.fs)`

adam-a-a:
> On my local, the following codes hit an error at L 2142 in multicomp_aq_sol_prop_pack.py , with message "property metadata method does not contain a method for eq_diffus_phase_comp"
> 
> `solver = get_solver m = ConcreteModel() m.fs = FlowsheetBlock(dynamic=False) m.fs.properties = MCASParameterBlock( solute_list=["A", "B", "C", "D", "E", "F", "G", "H"], charge={"E": 1, "F": -1}, diffus_calculation=DiffusivityCalculation.none, molar_volume_data={ ("Liq", "A"): 96e-6, # tested for benzene ("Liq", "B"): 100e-6, # arbitrary ("Liq", "C"): 60e-6, # arbitrary ("Liq", "D"): 200e-6, # arbitrary }, diffusivity_data={("Liq", "D"): 1e-11, ("Liq", "E"): 1.33e-9, ("Liq", "F"): 2.03e-9}, ) m.fs.properties.visc_d_phase["Liq"] = 1.0e-3
> 
> m.fs.sb = m.fs.properties.build_state_block([0], defined_state=True) iscale.calculate_scaling_factors(m.fs)`

I've added this as a test to ensure tests pass. Once the checks pass, I think it's time to finally merge. =)

PR #900, opened 2023-01-24
Fix report() for nanofiltration_0D

### LINKED ISSUE #899: Nanofiltration 0D stream labels
### Description

The streams for the nanofiltration 0D unit are inlet, permeate, and retentate -- but this has not been specified as part of the stream table, leading to an error when trying to get a report on the NF unit in a flowsheet.
Ex) when saying m.fs.NF.report() the following error comes up: 
ConfigurationError: Unit model fs.NF does not have the standard Port names (inlet and outlet). Please contact the unit model developer to develop a unit specific stream table.

### Expected Behavior

_No response_

### Steps to Reproduce

_No response_

### Environment

- OS:
- Python:
- WaterTAP:

### Anything Else?

Adam suggested that this would be the solution:
After the _get_performance_contents method in the NF 0D unit model, there should also be another method to get the stream table which provides conditions at inlet and outlet:

def _get_stream_table_contents(self, time_point=0):
    return create_stream_table_dataframe(
        {
            "Feed Inlet": self.inlet,
            "Feed Outlet": self.retentate,
            "Permeate Outlet": self.permeate,
        },
        time_point=time_point,
    )

adam-a-a:
## Summary/Motivation:
An error was reporting regarding use of the `report()` method for `nanofiltration_0D`.

This PR resolves that issue.

- add `_get_stream_table_contents` method to `nanofiltration_0D`
- add test to check report runs without error

### Legal Acknowledgement


MarcusHolly (approved):
I'll admit that I am a bit curious as to why/how the report() wasn't working in the first place, but everything you did here seems pretty straightforward. After adding a new line to the end of the test file, everything LGTM.

adam-a-a:
@MarcusHolly the `report` method relies on `_get_performance_contents()` and `_get_stream_table_contents()` both of which need to be methods in the unit model class. The unit model was missing the latter, so when state variables were supposed to be reported for each port, an error arose.

adam-a-a:
@MarcusHolly also, since there was no test that simply ran the report method, the error was able to slip by. We probably should have this same simple test for all unit models and at least have placeholder methods in unit models that are missing the 2 aforementioned methods.

agarciadiego (approved):

MarcusHolly (approved):

tarnold17 (approved):

PR #903, opened 2023-01-25
Adding density parameter to MCAS when `DensityCalculation.constant`

hunterbarber:
## Fixes/Resolves:
DensityCalculation.constant allowed for solution density to be constant, but only at 1000 kg/m3. Adding a parameter to allow the constant density to be manipulated from 1000.

- add dens_mass_const as param

### Legal Acknowledgement

### DIFF
--- a/watertap/property_models/multicomp_aq_sol_prop_pack.py
+++ b/watertap/property_models/multicomp_aq_sol_prop_pack.py
@@ -194,7 +194,7 @@ class MCASParameterData(PhysicalParameterBlock):
        .. csv-table::
            :header: "Configuration Options", "Description"
 
-           "``DensityCalculation.constant``", "Solution density assumed constant at 1000 kg/m3"
+           "``DensityCalculation.constant``", "Solution density assumed constant at 1000 kg/m3 by default in dens_mass_const parameter"
            "``DensityCalculation.seawater``", "Solution density based on correlation for seawater (TDS)"
            "``DensityCalculation.laliberte``", "Solution density based on mixing correlation from Laliberte"
        """,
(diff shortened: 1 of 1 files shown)


adam-a-a (approved):
LGTM

MarcusHolly (approved):
LGTM

adam-a-a:
@hunterbarber - just realized one small addition here- can you modify the documentation to reflect this change?

adam-a-a (changes requested):
I know I approved initially, but I realized we should include this modification in the documentation now rather than pushing it off, especially since it's a small change.

hunterbarber:
> I know I approved initially, but I realized we should include this modification in the documentation now rather than pushing it off, especially since it's a small change.

From what I've seen Params() typically aren't reflected in the documentation at all, let me double check and get back to you.

hunterbarber:
@adam-a-a Other than this being captured by the class automation I don't see any misinformation in the docs, or any mention of the old method. The only thing I did see when actually working on the other changes was that `dens_mass_w_phase` is in the docs but not in the actual code but should be `dens_mass_solvent`(?), but that's a different conversation.

hunterbarber:
> @adam-a-a Other than this being captured by the class automation I don't see any misinformation in the docs, or any mention of the old method. The only thing I did see when actually working on the other changes was that `dens_mass_w_phase` is in the docs but not in the actual code but should be `dens_mass_solvent`(?), but that's a different conversation.

image

hunterbarber:
> @adam-a-a Other than this being captured by the class automation I don't see any misinformation in the docs, or any mention of the old method. The only thing I did see when actually working on the other changes was that `dens_mass_w_phase` is in the docs but not in the actual code but should be `dens_mass_solvent`(?), but that's a different conversation.

Also that .lalibert actually isn't currently supported.

adam-a-a:
@hunterbarber I hadn't thoroughly looked at the current state of the MCAS documentation. Originally, we didn't have any documentation for it and wanted to spin something out quickly. Looking at it now though, its order deviates from other documentation (e.g., class documentation usually goes at the end), and it looks like it could use a fair amount of improvement. That should be reserved for another PR though. Can you make an issue with a running to-do list for improving the MCAS documentation (or add to the to-do list if an issue already exists for it)?

hunterbarber:
> @hunterbarber I hadn't thoroughly looked at the current state of the MCAS documentation. Originally, we didn't have any documentation for it and wanted to spin something out quickly. Looking at it now though, its order deviates from other documentation (e.g., class documentation usually goes at the end), and it looks like it could use a fair amount of improvement. That should be reserved for another PR though. Can you make an issue with a running to-do list for improving the MCAS documentation (or add to the to-do list if an issue already exists for it)?

It's up in #905, please add anything else you may have caught.

adam-a-a (approved):
LGTM

Issue #907, opened 2023-01-27
Database does not utilize path to user provided YAML file


wigging:
### Description

The `dbpath` parameter for the `Database` class is supposed to let the user define the path to a YAML file. But this path is ignored and watertap uses the database YAML files provided in watertap which are located at `watertap/data/techno_economic/`.

```python
from pyomo.environ import ConcreteModel
from watertap.core.wt_database import Database

model = ConcreteModel()
model.db = Database(dbpath=".")
``` 

### Expected Behavior

When the user provides the path to a YAML file with `Database(dbpath="path/to/myfile.yaml")`, then the model should use the parameters defined in that file.

### Steps to Reproduce

Parameters for a zero-order model are defined in the YAML file shown below.

```yaml
default:
  energy_electric_flow_vol_inlet:
    value: 0.050718512
    units: kWh/m^3
  capital_cost:
    basis: flow_vol
    cost_factor: None
    reference_state:
      value: 4732.0
      units: m^3/hr
    capital_a_parameter:
      value: 12.17829669e6
      units: USD_2014
    capital_b_parameter:
      value: 0.5862
      units: dimensionless
  recovery_frac_mass_H2O:
    value: 0.99
    units: dimensionless
    reference:
  default_removal_frac_mass_comp:
    value: 0
    units: dimensionless
  removal_frac_mass_comp:
    nonvolatile_toc:
      value: 0.2
      units: dimensionless
      constituent_longform: Nonvolatile TOC
    toc:
      value: 0.2
      units: dimensionless
      constituent_longform: Total Organic Carbon (TOC)
    tss:
      value: 0.97
      units: dimensionless
      constituent_longform: Total Suspended Solids (TSS)
```

The YAML file is located in the same directory as the example shown below.

```python
from pyomo.environ import ConcreteModel
from idaes.core import FlowsheetBlock
from idaes.core.solvers import get_solver
from watertap.core.wt_database import Database
from watertap.core.zero_order_properties import WaterParameterBlock
from watertap.unit_models.zero_order import DualMediaFiltrationZO

def main():

    model = ConcreteModel()
    model.db = Database(dbpath=".")

    model.fs = FlowsheetBlock(dynamic=False)
    model.fs.params = WaterParameterBlock(solute_list=["nonvolatile_toc", "toc", "tss"])

    model.fs.unit = DualMediaFiltrationZO(property_package=model.fs.params, database=model.db)
[...]

PR #912, opened 2023-02-02
Fix database path for user provided YAML file

wigging:
## Fixes/Resolves:

Fixes issue #907.

## Summary/Motivation:

The Database class has an optional argument where the user can provide the path to a YAML file for use with a zero-order model.

```python

model = ConcreteModel()
model.db = Database(dbpath="./myfile.yaml")
```

This path variable does not account for the name of the YAML file; therefore, the default database file in `watertap/data/techno_economic`/ is always used for the associated model. This pull request fixes the Database class to actually use the YAML file at the given path.

- Use a `self._filename` attribute for the `Database` class to define the `technology` variable. This will be set when `dbpath` is given; otherwise, the default database file is used for the associated zero-order model.

### Legal Acknowledgement

### DIFF
--- a/watertap/core/tests/test_wt_database.py
+++ b/watertap/core/tests/test_wt_database.py
@@ -27,8 +27,12 @@ def test_default_path():
 
 @pytest.mark.unit
 def test_invalid_path():
-    with pytest.raises(OSError):
+    with pytest.raises(OSError) as excinfo:
         Database(dbpath="./some/foo.yaml")
+    assert (
+        "Could not find requested path some. Please check that this path exists."
+        in str(excinfo.value)
+    )
 
 
 @pytest.mark.unit
--- a/watertap/core/tests/test_wt_database.py
+++ b/watertap/core/tests/test_wt_database.py
@@ -14,7 +14,7 @@
 Tests for WaterTAP database wrapper
 """
 import pytest
-import os
+from pathlib import Path
 
 from watertap.core.wt_database import Database
 
@@ -22,39 +22,20 @@ from watertap.core.wt_database import Database
 @pytest.mark.unit
 def test_default_path():
     db = Database()
-
-    assert os.path.normpath(db._dbpath) == os.path.normpath(
-        os.path.join(
-            os.path.dirname(os.path.abspath(__file__)),
-            "..",
-            "..",
-            "data",
-            "techno_economic",
-        )
-    )
+    assert db._dbpath == Path(__file__).parent.parent.parent / "data/techno_economic"
 
 
 @pytest.mark.unit
 def test_invalid_path():
-    with pytest.raises(
-        OSError,
-        match="Could not find requested path . Please " "check that this path exists.",
-    ):
-        Database(dbpath="foo.yaml")
+    with pytest.raises(OSError):
+        Database(dbpath="./some/foo.yaml")
 
 
 @pytest.mark.unit
 def test_custom_path():
     # Pick a path we know will exist, even if it isn't a data folder
-    db = Database(
-        dbpath=os.path.join(
-            os.path.dirname(os.path.abspath(__file__)), "..", "..", "core"
-        )
-    )
-
-    assert db._dbpath == os.path.join(
-        os.path.dirname(os.path.abspath(__file__)), "..", ".."
-    )
+    db = Database(dbpath=Path(__file__).parent.parent / "core")
+    assert db._dbpath == Path(__file__).parent.parent
 
 
 class TestDatabase:
--- a/watertap/core/tests/test_wt_database.py
+++ b/watertap/core/tests/test_wt_database.py
@@ -38,10 +38,9 @@ def test_default_path():
 def test_invalid_path():
     with pytest.raises(
         OSError,
-        match="Could not find requested path foo. Please "
-        "check that this path exists.",
+        match="Could not find requested path . Please " "check that this path exists.",
     ):
-        Database(dbpath="foo")
+        Database(dbpath="foo.yaml")
 
 
 @pytest.mark.unit
@@ -54,7 +53,7 @@ def test_custom_path():
     )
 
     assert db._dbpath == os.path.join(
-        os.path.dirname(os.path.abspath(__file__)), "..", "..", "core"
+        os.path.dirname(os.path.abspath(__file__)), "..", ".."
     )
 
 
--- a/watertap/core/tests/test_zero_order_sido_reactive.py
+++ b/watertap/core/tests/test_zero_order_sido_reactive.py
@@ -48,7 +48,7 @@ from watertap.core.zero_order_sido_reactive import (
 
 solver = get_solver()
 
-local_path = os.path.dirname(os.path.abspath(__file__))
+local_path = os.path.dirname(os.path.abspath(__file__)) + "/test_sidor_data.yaml"
 
 
 @declare_process_block_class("DerivedSIDOR")
--- a/watertap/core/wt_database.py
+++ b/watertap/core/wt_database.py
@@ -27,10 +27,13 @@ class Database:
     associated with zero-order models in WaterTap.
 
     Args:
-        dbpath - (optional) path to database folder containing yaml files
+        dbpath - (optional) Path to a user provided YAML database file. For
+        example, the path could be ~/Desktop/myproject/mydata.yaml on a Mac.
+        If a path is not provided, then a default YAML database file is used
+        for the associated zero-order model.
 
     Returns:
-        an instance of a Database object linked to the provided database
+        An instance of a Database object linked to the provided database.
     """
 
     def __init__(self, dbpath=None):
@@ -44,7 +47,8 @@ class Database:
                 "techno_economic",
             )
         else:
-            self._dbpath = dbpath
+            self._dbpath = os.path.dirname(dbpath)
+            self._filename = os.path.basename(dbpath).split(".")[0]
 
             # Confirm valid path
             if not os.path.isdir(self._dbpath):
@@ -196,7 +200,10 @@ class Database:
             # If data is already in cached files, return
             return self._cached_files[technology]
         else:
-            # Else load data from required file
+            # If user provided a YAML file, use the filename as the technology
+            if hasattr(self, "_filename"):
+                technology = self._filename
+            # Load from database YAML file or from user provided YAML file
             try:
                 with open(os.path.join(self._dbpath, technology + ".yaml"), "r") as f:
                     lines = f.read()
--- a/watertap/core/wt_database.py
+++ b/watertap/core/wt_database.py
@@ -190,24 +190,25 @@ class Database:
         return self._component_list
 
     def _get_technology(self, technology):
+        # If data is already in cached files, then return that data
         if technology in self._cached_files:
-            # If data is already in cached files, return
             return self._cached_files[technology]
-        else:
-            # If user provided a YAML file, use the filename as the technology
-            if hasattr(self, "_filename"):
-                technology = self._filename
-            # Load from database YAML file or from user provided YAML file
-            try:
-                lines = (self._dbpath / technology).with_suffix(".yaml").read_text()
-            except OSError:
-                raise KeyError(f"Could not find entry for {technology} in database.")
 
-            fdata = yaml.load(lines, yaml.Loader)
+        # If user provided a YAML file, use the filename as the technology
+        if hasattr(self, "_filename"):
+            technology = self._filename
 
-            # Store data in cache and return
-            self._cached_files[technology] = fdata
-            return fdata
+        # Load from database YAML file or from user provided YAML file
+        try:
+            lines = (self._dbpath / technology).with_suffix(".yaml").read_text()
+        except OSError:
+            raise KeyError(f"Could not find entry for {technology} in database.")
+
+        fdata = yaml.load(lines, yaml.Loader)
+
+        # Store data in cache and return
+        self._cached_files[technology] = fdata
+        return fdata
 
     def _load_component_list(self):
         """
--- a/watertap/core/wt_database.py
+++ b/watertap/core/wt_database.py
@@ -14,9 +14,9 @@
 This module contains the base class for interacting with WaterTAP data files
 with zero-order model parameter data.
 """
-import os
 import yaml
 from copy import deepcopy
+from pathlib import Path
 
 
 class Database:
@@ -40,18 +40,14 @@ class Database:
         self._cached_files = {}
 
         if dbpath is None:
-            self._dbpath = os.path.join(
-                os.path.dirname(os.path.abspath(__file__)),
-                "..",
-                "data",
-                "techno_economic",
-            )
+            self._dbpath = Path(__file__).parent.parent / "data/techno_economic"
         else:
-            self._dbpath = os.path.dirname(dbpath)
-            self._filename = os.path.basename(dbpath).split(".")[0]
+            path = Path(dbpath)
+            self._dbpath = path.parent
+            self._filename = path.stem
 
             # Confirm valid path
-            if not os.path.isdir(self._dbpath):
+            if not self._dbpath.is_dir():
                 raise OSError(
                     f"Could not find requested path {self._dbpath}. Please "
                     f"check that this path exists."
@@ -80,9 +76,7 @@ class Database:
         else:
             # Else load data from required file
             try:
-                with open(os.path.join(self._dbpath, "water_sources.yaml"), "r") as f:
-                    lines = f.read()
-                    f.close()
+                lines = (self._dbpath / "water_sources.yaml").read_text()
             except OSError:
                 raise KeyError("Could not find water_sources.yaml in database.")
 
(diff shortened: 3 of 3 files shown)


watertap/core/wt_database.py
```diff
@@ -44,7 +47,8 @@ def __init__(self, dbpath=None):
                 "techno_economic",
             )
         else:
-            self._dbpath = dbpath
+            self._dbpath = os.path.dirname(dbpath)
+            self._filename = os.path.basename(dbpath).split(".")[0]
```
lbianchi-lbl: Suggestion using `pathlib.Path`:

```suggestion
            dbpath = Path(dbpath).resolve()
            self._dbpath = dbpath.parent
            self._filename = dbpath.stem
```

watertap/core/wt_database.py
```diff
@@ -196,7 +200,10 @@ def _get_technology(self, technology):
             # If data is already in cached files, return
             return self._cached_files[technology]
         else:
-            # Else load data from required file
+            # If user provided a YAML file, use the filename as the technology
+            if hasattr(self, "_filename"):
+                technology = self._filename
+            # Load from database YAML file or from user provided YAML file
             try:
                 with open(os.path.join(self._dbpath, technology + ".yaml"), "r") as f:
```
lbianchi-lbl: Suggestion: the contents of the `try` block can be simplified by e.g.:

```py
technology_file_path = (self._dbpath / technology).with_suffix(".yaml")
lines = technology_file_path.read_text()
```

watertap/core/tests/test_wt_database.py
```diff
@@ -54,7 +53,7 @@ def test_custom_path():
     )
 
     assert db._dbpath == os.path.join(
-        os.path.dirname(os.path.abspath(__file__)), "..", "..", "core"
+        os.path.dirname(os.path.abspath(__file__)), "..", ".."
```
lbianchi-lbl: Clarification: if the purpose of this test is just to check that `db._dbpath` is processed correctly using any arbitrary path, why not to use `Path(__file__).resolve()` (i.e. the absolute path to the test file itself)? Barring very exotic corner cases, this would be guaranteed to exist almost by definition.

(Also, feel free to disregard if this implementation detail predates this PR and is unrelated to the other changes).

watertap/core/tests/test_zero_order_sido_reactive.py
```diff
@@ -48,7 +48,7 @@
 
 solver = get_solver()
 
-local_path = os.path.dirname(os.path.abspath(__file__))
+local_path = os.path.dirname(os.path.abspath(__file__)) + "/test_sidor_data.yaml"
```
lbianchi-lbl, line 51: Suggestion: using `pathlib.Path`:

```suggestion
local_path = Path(__file__).resolve().parent / "test_sidor_data.yaml"
```

watertap/core/wt_database.py
```diff
@@ -196,7 +200,10 @@ def _get_technology(self, technology):
             # If data is already in cached files, return
             return self._cached_files[technology]
         else:
```
lbianchi-lbl: Suggestion: the `if` block above contains a `return` statement, so the `else` is not necessary.

lbianchi-lbl (commented):
I've left a few comments with suggestion on the implementation, but nothing show-stopping.

watertap/core/wt_database.py
wigging: I did not use `pathlib` because `os` was already used in this file. Is there any reason why `os` was used instead of `pathlib` for this class?

wigging: Are there other areas of watertap that would expect an `os` path instead of a `pathlib` path?

watertap/core/wt_database.py
wigging: This assumes using `pathlib` instead of `os`. See my previous comment about using `pathlib` instead of `os` for paths.

watertap/core/tests/test_wt_database.py
wigging: All of the original code in this pull request predates me. So I don't know why it's pointing to `core`.

watertap/core/tests/test_zero_order_sido_reactive.py
wigging: See previous comment about using `pathlib` vs `os`.

watertap/core/wt_database.py
wigging: This whole function is badly written. I just made the changes needed to get the path functionality working.

watertap/core/wt_database.py
bknueven: > I did not use `pathlib` because `os` was already used in this file. Is there any reason why `os` was used instead of `pathlib` for this class?

I don't think so. We should switch to pathlib where/when convenient.

watertap/core/wt_database.py
wigging: Are there any examples of `self._cached_files` being used?

watertap/core/wt_database.py
wigging: Ok, I replaced os with pathlib.

watertap/core/wt_database.py
wigging: Done.

watertap/core/tests/test_wt_database.py
wigging: Implemented this with pathlib instead of os.

watertap/core/tests/test_zero_order_sido_reactive.py
wigging: Done.

watertap/core/tests/test_wt_database.py
```diff
@@ -14,48 +14,28 @@
...
-    )
+    assert db._dbpath == Path(__file__).parent.parent.parent / "data/techno_economic"
 
 
 @pytest.mark.unit
 def test_invalid_path():
-    with pytest.raises(
-        OSError,
-        match="Could not find requested path foo. Please "
-        "check that this path exists.",
```
bknueven, line 42: We generally like having these match statements so we can be sure the "right" error was raised. But perhaps not strictly needed.

bknueven (approved):

watertap/core/tests/test_wt_database.py
wigging: Pushed up a commit to address this.

watertap/core/wt_database.py
wigging: I removed the `else` statement but left the `self._cached_files` since I'm not sure if these are used elsewhere in the code.

bknueven (commented):
Something worth mentioning is I think this breaks the original intent of supplying your own `dbpath`: that you have a whole set of `yaml` files just like in `data/techno_economic/`. Instead you have to supply individual file names.

Maybe this makes more sense for our workflow, where maybe people will come with their own custom units. But in that case shouldn't they just use a separate "Database" object with the custom units?

wigging:
@bknueven I don't know the original intentions for this database class. But based on the tests (there are no examples that I'm aware of), the `dbpath` is being used to parse a single YAML file. I don't see any tests or examples that demonstrate using the path to parse several YAML files in a directory.

wigging:
Looks like @andrewlee94 is the creator of the `Database` class. Andrew, what is the purpose of the `dbpath` argument? From looking at the tests, it is used to parse a single YAML file. I'm not aware of any examples where it is used to parse many YAML files at once in a directory. Or maybe I missed something, I don't know. Any clarification on this would be helpful.

andrewlee94:
It has been a very long time since I wrote this, but I believe it was intended to be a path to a folder containing a set of YAML files.

bknueven:
> @bknueven I don't know the original intentions for this database class. But based on the tests (there are no examples that I'm aware of), the `dbpath` is being used to parse a single YAML file. I don't see any tests or examples that demonstrate using the path to parse several YAML files in a directory.

The default `dbpath` is this whole directory: https://github.com/watertap-org/watertap/tree/main/watertap/data/techno_economic. This test points to path, not a filename: https://github.com/watertap-org/watertap/blob/85e1f521141dd52c10c560d52b651e22c4832311/watertap/core/tests/test_wt_database.py#L48-L58

Here is an example flowsheet: https://github.com/watertap-org/watertap/blob/main/watertap/examples/flowsheets/case_studies/municipal_treatment/municipal_treatment.py. Note only one `Database` object, lots of unit models.

 I'm going to revisit #907.

bknueven (changes requested):
After thinking more carefully about #907, this code works as intended.

I think there's a separate issue of *how* we associate filenames with unit models, but I don't think requiring a separate database for each unit model is the way to go. This will rapidly become annoying if a user has an entire set of yaml files.

It might be that better documentation is needed.

wigging:
Would something like the following suffice:

```python
class Database:
    def __init__(self, dbpath=None, yamlfile=None):
        # blah blah blah
```

The user could provide the path to their YAML file like this:

```python
model = ConcreteModel()
model.db = Database(yamlfile="./myparams/dual_media_filter.yaml")
```

Actually, is there any scenario when `dbpath` would point any where other than `watertap/data/techno_economic/`?

bknueven:
> The user could provide the path to their YAML file like this:
> ```python
> model = ConcreteModel()
> model.db = Database(yamlfile="./myparams/dual_media_filter.yaml")
> ```

What I don't like about this is it turns the `Database` into a manager for a specific yaml file as opposed to a manager for a directory of yaml files. There's also the issue with `water_sources`.


I think the thing that makes more sense is for the *unit model* to tell the database where its yaml file is, e.g.,
```python
model = ConcreteModel()
model.db = Database(dbpath=".")

model.fs = FlowsheetBlock(dynamic=False)
model.fs.params = WaterParameterBlock(solute_list=["nonvolatile_toc", "toc", "tss"])

model.fs.unit = DualMediaFiltrationZO(property_package=model.fs.params, database=model.db, yaml_file="./myparams/dual_media_filter.yaml")
```

This would allow the user to bind an instance of a ZO unit model class to their own YAML file, which seems like a good use-case: the user can customize a single unit in the process easily while taking the defaults for all the other units.

wigging:
So what's the point of having the user define `dbpath` for the `Database`? Why would a user ever want to modify that path? 

If the user did something like shown below then why define a path in two different places?

```python
model = ConcreteModel()
model.db = Database(dbpath=".")

model.fs = FlowsheetBlock(dynamic=False)
model.fs.params = WaterParameterBlock(solute_list=["nonvolatile_toc", "toc", "tss"])

model.fs.unit = DualMediaFiltrationZO(property_package=model.fs.params, database=model.db, yaml_file="./myparams/dual_media_filter.yaml")
```

If a user provides the YAML file, then why does a database instance need to be created? Can this be done:

```python
model = ConcreteModel()

model.fs = FlowsheetBlock(dynamic=False)
model.fs.params = WaterParameterBlock(solute_list=["nonvolatile_toc", "toc", "tss"])

model.fs.unit = DualMediaFiltrationZO(property_package=model.fs.params, yaml_file="./myparams/dual_media_filter.yaml")
```

bknueven:
> So what's the point of having the user define `dbpath` for the `Database`? Why would a user ever want to modify that path? 


Copy the contents of `watertap/data/techno_economic` to your own directory, make the modifications you like, and point the `dbpath` to the new directory. We could probably define a simple function to "set up" the database files in a desired location.

wigging:
I'm so confused. Why would I copy the entire database directory of YAML files to some other location? That directory is part of watertap.

bknueven:
So you can easily set the database parameters to your own values.

wigging:
I installed watertap in editable mode with `pip install -e`. I don't need to copy any of the database files. I can edit them directly. Anyway, that is not the point of this pull request.

andrewlee94:
@wigging The point is that there is no single set of universal parameters for this - different projects/applications might have different sets of parameters and a single user might need to work on multiple projects. Thus, there needs to be a way for users to select from different parameter sets.

wigging:
@andrewlee94 Why use a YAML file in `watertap/data/techno_economic` to define the parameter values for these models? Why not set those values in the model itself?

andrewlee94:
@wigging There is nothing to prevent the user from doing that, but the whole zero-order model thing grew out of WaterTAP3 where it was intended to be a very hands-off experience for low-level users. Plus, these parameters tend to be case specific, but there are often many unit instances in a single case - the goal was to have a central location for defining the global parameters, but have a way that they could be easily swapped out for a different set (note that we generally expected the person creating the datasets to have a greater knowledge of how things worked).

wigging:
@andrewlee94 Can you point me to the documentation on how to setup and run a zero-order model? I've been looking at the tests inside of watertap as a guide. There's a section on [zero-order unit models](https://watertap.readthedocs.io/en/stable/technical_reference/unit_models/zero_order_unit_models/index.html) in the docs but it just describes each model, there's nothing about actually setting up the model and running it.

andrewlee94:
This is the only example I could find in a quick search: https://github.com/watertap-org/watertap/blob/main/watertap/examples/flowsheets/case_studies/municipal_treatment/municipal_treatment.py

However, I am not the one to ask about that - I was responsible for putting together the basic infrastructure but not putting it to use.

wigging:
@andrewlee94 Yeah, I saw these scripts which demonstrate using zero-order models within a flowsheet environment. But I was wondering if there is documentation regarding running an individual zero-oder model. So apparently there is no documentation. Just tests and scripts buried inside of the watertap source code.

adam-a-a:
> @andrewlee94 Yeah, I saw these scripts which demonstrate using zero-order models within a flowsheet environment. But I was wondering if there is documentation regarding running an individual zero-oder model. So apparently there is no documentation. Just tests and scripts buried inside of the watertap source code.

The only other documentation we have on zero-order modeling is [here](https://watertap.readthedocs.io/en/stable/technical_reference/core/index.html). As far as putting documentation together on the workflow of putting all the pieces together for creating and running zero-order models, that's still a subtask in a master issue we have that relates to this: #595. Notably, we haven't had this as high-priority since we were in the weeds for quite some time with other issues of importance. So yes, we typically follow test files as examples to go by, and for developing and running individual unit models, we'd usually just meet and go over it with each other (or with someone who was interested in using ZO models). We still need someone to step up and get some documentation in on ZO modeling development/usage workflows. Interested? =)

adam-a-a:
@wigging  So I thought about this more, and it seems to me that there are a couple different aspects we need to consider here. You should be able to navigate to some other db path aside from `data/technoeconomic`. Fair enough. I can see the case where someone might want to have their own directory of yaml files to work from. 

However, for your purposes, and please correct me if I'm wrong, it seemed like you wanted to use a unit model that already existed and to supply your own yaml file to populate that unit model. From the developer perspective, the way I would suggest doing this is by adding a subtype to the existing yaml attributed to the unit model you're working with. That is, add to the existing yaml file, then specify your subtype name when using the unit model (if you're unsure how to do this due to lack of documentation, happy to have a call and talk through). On the other hand, if you are adamant about the idea that a user shouldn't have to dig into our yaml files and modify anything (seems reasonable that we shouldn't force users to do that), then I do think it makes sense for the custom yaml file to be specified at the unit level, overriding usage of the Database instance for that particular unit.

As @andrewlee94 mentioned, this _all_ sprouted from an abrupt need to review nearly 100 simple models from another repo, cleanup and standardize their implementation, and migrate them into WaterTAP, all in a fairly short time. This might be why we didn't thoroughly consider other use cases that might come about. Once we accomplished the migration of those models, we were content with moving on to more important tasks, like working on detailed, predictive models which can be of far greater use. All that said, it is fairly easy to use the ZO modeling "framework," once you know how things connect. So, if the concern is implementing your use case, I would think we could do that easily (i.e., simulating a unit or flowsheet with your own specified parameter values). If the main concern is having a workflow for a novice user (not developer) to be able to do their own thing outside of the existing "database" and workflow that we've used nominally in the past (i.e., change the workflow so that the user can have more options/control over how they provide YAML data), then I wonder if the end justifies the means. How far do we take this?

watertap/core/wt_database.py
```diff
@@ -192,23 +190,25 @@ def _return_component_list(self):
...
-                with open(os.path.join(self._dbpath, technology + ".yaml"), "r") as f:
-                    lines = f.read()
-                    f.close()
-            except OSError:
-                raise KeyError(f"Could not find entry for {technology} in database.")
 
-            fdata = yaml.load(lines, yaml.Loader)
+        # If user provided a YAML file, use the filename as the technology
+        if hasattr(self, "_filename"):
+            technology = self._filename
```
adam-a-a, line 199: Shouldn't we have a test that includes a Database instance along with a unit model that makes use of this user-provided file path? I would want to see if anything breaks when using the "typical" sequence of methods (e.g., `load_parameters_from_database`, `get_unit_operation_parameters`) that use in our tests for ZO unit models and/or flowsheets.

wigging:
I'm going to close this based on the discussion during Wednesday's meeting. Instead of restructuring the watertap code, I'll submit a pull request to improve the zero-order model documentation to make it clear how to properly use a zero-order model. We can discuss options for improving the zero-order models at a later date.

PR #913, opened 2023-02-02
Add initial setup for static checks using Pylint

### LINKED ISSUE #620: Add static checking using Pylint
Partially inspired by #609, but useful also for other things.

Following @ksbeattie's (IIRC) suggestion, I think the best thing would be starting from an opt-in strategy, i.e. disable all checks by default, and enable the subset of checks that we want to enforce one at a time.

Full list of checks/possible errors (somewhat confusingly, called "messages" in Pylint): https://pylint.pycqa.org/en/latest/user_guide/messages/messages_overview.html

lbianchi-lbl:
- Add minimal working Pylint configuration with all messages (individual checks) disabled
- Add CI check running Pylint on the entire `watertap` package
- "Unlock" one or more checks for issues that can be fixed quickly as a start

### Legal Acknowledgement

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -42,6 +42,23 @@ jobs:
         run: |
           black --check .
 
+  pylint:
+    name: Code linting (pylint)
+    runs-on: ubuntu-latest
+    needs: [code-formatting]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.8'
+      - name: Install dev dependencies
+        run: |
+          pip install --progress-bar off -r requirements-dev.txt
+          pip list
+      - name: Run pylint
+        run: |
+          pylint watertap
+
   tests:
     name: Tests (py${{ matrix.python-version }}/${{ matrix.os }})
     runs-on: ${{ matrix.os-version }}
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -62,7 +62,6 @@ jobs:
   tests:
     name: Tests (py${{ matrix.python-version }}/${{ matrix.os }})
     runs-on: ${{ matrix.os-version }}
-    if: "false"
 # TODO: Grab cloud database password from encrypted repo secrets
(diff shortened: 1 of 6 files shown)


bknueven (approved):

wigging:
There are several linters available for Python code. Why did you want to use pylint? And please don't say you chose it because that's what IDAES uses.

bknueven:
> There are several linters available for Python code. Why did you want to use pylint? 

If you have a concrete suggestion, please say so. I thought we had settled this yesterday.

> And please don't say you chose it because that's what IDAES uses.

This is actually a very good reason. There's a huge amount of overlap between the two development teams, and utilizing different tools on different projects increases the mental overhead for those people. So if we're going to diverge I think we need a compelling reason.

wigging:
@bknueven I don't think we ever discussed other linters. I just assumed ya'll had already chosen pylint since that was the topic of discussion. I'm just wondering if there was a particular reason for using pylint for watertap. I'm familiar with flake8 and use it on several other work projects. On those projects, we chose flake8 because pylint was slow and it was too picky about errors/warnings. There was less to configure with flake8. And regarding linters, I've heard good things about [ruff](https://github.com/charliermarsh/ruff), so it might something to consider too.

.github/workflows/checks.yml
```diff
@@ -42,6 +42,23 @@ jobs:
...
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.8'
+      - name: Install dev dependencies
+        run: |
+          pip install --progress-bar off -r requirements-dev.txt
+          pip list
+      - name: Run pylint
+        run: |
+          pylint watertap
```
wigging, line 60: For the first implementation of this, can you just run pylint on a sub-directory of watertap? For example, do something like `pylint watertap/core` instead of linting the whole package. When linting passes in watertap/core then add another sub-directory and so on. Eventually, you can change this back to `pylint watertap` when all the sub-directories have passed.

ksbeattie (approved):
LGTM. We can now come up with a plan on how to start turning up the heat, bit by bit.

Issue #914, opened 2023-02-03
Enable additional Pylint static checks


lbianchi-lbl:
### Description

Once #913 is in, we should start a coordinated effort to enable additional Pylint static checks.

### Motivation

- Improve "static coverage" by reducing the number of false-negatives (i.e. issues that could be caught by a static-analysis tool such as Pylint, but are or might end up in the codebase since the corresponding check is not enforced)

### Possible Implementation

I have in mind something along the lines of:

- Determine what checks ("messages" in Pylint parlance) would be worth implementing, using as a reference:
  - https://pylint.readthedocs.io/en/latest/user_guide/messages/messages_overview.html
- Run `pylint` on the current codebase with those checks enabled

Then, for each check/message:

- Estimate the effort needed to address the corresponding issue in the entire codebase (i.e. creating a "clean slate" so that the check can be enforced on incoming code in PRs)
- Implement the fixes "atomically", according to the following "check-driven" workflow:
  - Open a PR
  - Enable one (or a small number of) check by adding its message string in the `enable` list within the `[tool.pylint."messages control"]` stanza of `pyproject.toml`
  - Commit the changes to `pyproject.toml`
  - Verify that the Pylint check fails
  - Address the issues by modifying the Python code
  - Commit the changes
  - Verify that the Pylint check passes
  - Merge the PR

ksbeattie:
Once #995 is merged we can enable that check.  But then the pylint magic in idaes should be brought over here.

PR #917, opened 2023-02-05
Pylint: enable all errors passing on main

bknueven:
## Fixes/Resolves:
Part of adding Pyllint checks #914

- Enables checking pylint Error messages compatible with main
- Drive-by fixes a few easy-to-fix Error messages

### Legal Acknowledgement


pyproject.toml
```diff
@@ -2,7 +2,19 @@
 py-version = "3.8"
 
 [tool.pylint."messages control"]
```
wigging, line 4: Does it matter if this is lower case or upper case? The pylint docs use `[tool.pylint.'MESSAGES CONTROL']`.

bknueven: https://github.com/PyCQA/pylint/blob/main/examples/pyproject.toml#L344

It seems like it doesn't matter. The existing configuration also clearly worked locally for me as well as on the github runner.

lbianchi-lbl (approved):
Looks good!

tarnold17 (approved):

PR #921, opened 2023-02-10
Add documentation for running zero-order models

wigging:
## Fixes/Resolves:

This adds some documentation in the How To Guides section for running a zero-order model. This also adds notes regarding the YAML file name associated with a zero-order model and how to use a local YAML database file.

## Summary/Motivation:

There is currently no documentation about running a zero-order model or providing a user defined YAML file. This request adds some documentation to hopefully clarify how to use a zero-order model.

- Add documentation in the How To Guides section for running a zero-order model
- Add documentation regarding the use of a local YAML file and proper naming of that file for use with a zero-order model

### Legal Acknowledgement

### DIFF
--- a/docs/technical_reference/unit_models/zero_order_unit_models/index.rst
+++ b/docs/technical_reference/unit_models/zero_order_unit_models/index.rst
@@ -1,5 +1,8 @@
 Zero-Order Unit Models
 ======================
+
+The YAML database files for the zero-order models are located in :code:`watertap/data/techno_economic/`. The YAML file name should match the associated model name without the ZO letters and use snake case style. For example, the YAML file for the :code:`DualMediaFiltrationZO()` zero-order model is named :code:`dual_media_filtration.yaml`.
+
 .. toctree::
    :maxdepth: 1
 
--- a/docs/technical_reference/unit_models/zero_order_unit_models/index.rst
+++ b/docs/technical_reference/unit_models/zero_order_unit_models/index.rst
@@ -1,7 +1,7 @@
 Zero-Order Unit Models
 ======================
 
-The YAML database files for the zero-order models are located in :code:`watertap/data/techno_economic/`. The YAML file name should match the associated model name without the ZO letters and use snake case style. For example, the YAML file for the :code:`DualMediaFiltrationZO()` zero-order model is named :code:`dual_media_filtration.yaml`.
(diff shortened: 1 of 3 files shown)


ksbeattie:
@wigging if this is ready for review, can you add some reviewers?

lbianchi-lbl:
For the reviewers' convenience, I think this is the preview version of the content being added: https://watertap--921.org.readthedocs.build/en/921/how_to_guides/how_to_run_zero_order_model.html

hunterbarber (commented):
@wigging Looks good, I only have two recommendations. I would move "The output from running the example is also shown below" to line 42, between the two codeblocks. Given someone is at the starting point of looking how to run ZO models, there may not even be the experience there to know the difference in input and output formatting. Additionally, one line of inline commenting before each section in the code, such as # building model, # specifying model variables, etc. consistent with the other examples in IDAES/WaterTAP may be good for this same inexperience reason.

wigging:
@hunterbarber Ok, I'll add more comments and descriptions.

docs/technical_reference/unit_models/zero_order_unit_models/index.rst
```diff
@@ -1,5 +1,8 @@
 Zero-Order Unit Models
 ======================
+
+The YAML database files for the zero-order models are located in :code:`watertap/data/techno_economic/`. The YAML file name should match the associated model name without the ZO letters and use snake case style. For example, the YAML file for the :code:`DualMediaFiltrationZO()` zero-order model is named :code:`dual_media_filtration.yaml`.
+
```
adam-a-a, line 5: This might be a good place to tack on a disclaimer about the data, basically pointing out that they were mainly meant to serve as default values and should be taken with a grain of salt. We could also mention here that users are intended to supply their own data, after they get the hang of running models with defaults.

adam-a-a: This could be done here or in a subsequent PR. I just wanted to make note of it here before forgetting about it.

tarnold17 (approved):
I see there are a few minor requests that should be addressed, but other than that, this looks good to me.

wigging:
I made some changes based on everyone's comments. Let me know if anything else needed.

docs/technical_reference/unit_models/zero_order_unit_models/index.rst
wigging: Made some changes to address this. See recent commit.

MarcusHolly (approved):
LGTM

hunterbarber (approved):
Changes look good

PR #933, opened 2023-02-22
Translator block ADM1-ASM1

agarciadiego:
## Summary/Motivation:
Translator block ADM1-ASM1

- Includes a translator unit model to specific ADM1/ASM1 property packages

### Legal Acknowledgement


agarciadiego:
@MarcusHolly @luohezhiming Here is the first look of the implementation of the translator block. It may have some changes but I think it would be good for you guys to see what constraints/conversions are needed to do this for ASM2D

watertap/unit_models/translator_adm1_asm1.py
```diff
@@ -0,0 +1,436 @@
...
+            doc="Nitrogen in amino acids and proteins [kmole N/kg COD]",
+        )
+        self.N_bac = Param(
+            initialize=0.08 / 14,
+            units=pyunits.kmol * pyunits.kg**-1,
+            mutable=True,
+            doc="Nitrogen content in bacteria [kmole N/kg COD]",
+        )
+
+        self.i_ec = Param(
```
agarciadiego: @adam-a-a this may belong in the properties block. Any thoughts?

MarcusHolly: Where is this value from? I didn't see it in the Rosen & Jeppsson paper...unless it's just the numerator of N_I

watertap/unit_models/translator_adm1_asm1.py
```diff
@@ -0,0 +1,438 @@
...
+            )
+
+        @self.Constraint(
+            self.flowsheet().time,
+            doc="Equality S_ND equation",
+        )
+        def eq_Snd_conc(blk, t):
+            return blk.properties_out[t].conc_mass_comp["S_ND"] == (
+                blk.properties_in[t].conc_mass_comp["S_I"] * blk.N_I
+            ) + (blk.properties_in[t].conc_mass_comp["S_I"] * blk.N_aa)
```
MarcusHolly:
```suggestion
            ) + (blk.properties_in[t].conc_mass_comp["S_aa"] * blk.N_aa)
```

watertap/unit_models/translator_adm1_asm1.py
```diff
@@ -0,0 +1,438 @@
...
+                        + blk.properties_in[t].conc_mass_comp["X_aa"]
+                        + blk.properties_in[t].conc_mass_comp["X_fa"]
+                        + blk.properties_in[t].conc_mass_comp["X_c4"]
+                        + blk.properties_in[t].conc_mass_comp["X_pro"]
+                        + blk.properties_in[t].conc_mass_comp["X_ac"]
+                        + blk.properties_in[t].conc_mass_comp["X_h2"]
+                    )
+                )
+                + (blk.properties_in[t].conc_mass_comp["X_I"] * blk.N_I)
+                + (blk.properties_in[t].conc_mass_comp["X_c"] * blk.N_I)
```
MarcusHolly: Is there a reason `N_I` is used here over the `N_xc` parameter?

watertap/unit_models/translator_adm1_asm1.py
agarciadiego: from Copp simulation manual pp6 called i_XP

watertap/unit_models/translator_adm1_asm1.py
agarciadiego: Changed. However, we should evaluate if this parameters live here

watertap/unit_models/translator_adm1_asm1.py
adam-a-a: Although we discussed internally, I'll post the consensus here: we're deciding to _not_ move reaction parameters to the property parameter block and will instead allow for the rxn inlet and outlet of the translator block be specified in config options. We can revise later on if we somehow decide that this would be a bad route to take.

watertap/unit_models/translator_adm1_asm1.py
```diff
@@ -0,0 +1,463 @@
...
+# Solutions of Sandia, LLC, Carnegie Mellon University, West Virginia University
+# Research Corporation, et al.  All rights reserved.
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and
+# license information.
+#################################################################################
+"""
+Modified translator block.
+
+This is copied from the Generic template for a translator block.
```
adam-a-a: Why not inherit from the translator block and build on top of it? I would imagine it would eliminate some unnecessary (duplicated) lines of code, though it wouldn't be necessary for functionality purposes I suppose.

watertap/unit_models/translator_adm1_asm1.py
```diff
@@ -0,0 +1,463 @@
...
+_log = idaeslog.getLogger(__name__)
+
+
+@declare_process_block_class("Translator_ADM1_ASM1")
+class TranslatorData(UnitModelBlockData):
+    """
+    Standard Translator Block Class
+    """
+
+    CONFIG = ConfigBlock()
```
adam-a-a: If we inherited from the Translator block class, I think we wouldn't need to include _most_ of these config options--only the newly added ones.

watertap/unit_models/translator_adm1_asm1.py
```diff
@@ -0,0 +1,463 @@
...
+            self.flowsheet().time,
+            doc="Material properties in outgoing stream",
+            defined_state=self.config.outlet_state_defined,
+            has_phase_equilibrium=False,
+            **self.config.outlet_property_package_args
+        )
+
+        # Add ports
+        self.add_port(name="inlet", block=self.properties_in, doc="Inlet Port")
+        self.add_port(name="outlet", block=self.properties_out, doc="Outlet Port")
```
adam-a-a: This would also be eliminated by inheriting from TranslatorData (?)

watertap/unit_models/translator_adm1_asm1.py
```diff
@@ -0,0 +1,463 @@
...
+        self.properties_out.initialize(
+            outlvl=outlvl,
+            optarg=optarg,
+            solver=solver,
+            state_args=state_args_out,
+        )
+
+        if degrees_of_freedom(self) == 0:
+            with idaeslog.solver_log(init_log, idaeslog.DEBUG) as slc:
+                res = opt.solve(self, tee=slc.tee)
```
adam-a-a: Should check the solver for optimal termination and if not optimal, raise initialization error (but raising the error should come after release_state)

watertap/unit_models/translator_adm1_asm1.py
```diff
@@ -0,0 +1,463 @@
...
+
+            init_log.info("Initialization Complete {}.".format(idaeslog.condition(res)))
+        else:
+            init_log.warning(
```
adam-a-a: I guess this was just taken from Translator template. I wonder if this was an oversight in IDAES or not. Usually, we raise exceptions for something like this, and I am wondering if IDAES' Translator's `initialize_build` routine intentionally relaxes our typical restrictions for some reason. I am leaning towards this just being an oversight, but maybe this is a question for @andrewlee94

watertap/examples/flowsheets/case_studies/anaerobic_digestor/Translator_BSM2.py
```diff
@@ -0,0 +1,132 @@
...
+# Framework (IDAES IP) was produced under the DOE Institute for the
+# Design of Advanced Energy Systems (IDAES), and is copyright (c) 2018-2021
+# by the software owners: The Regents of the University of California, through
+# Lawrence Berkeley National Laboratory,  National Technology & Engineering
+# Solutions of Sandia, LLC, Carnegie Mellon University, West Virginia University
+# Research Corporation, et al.  All rights reserved.
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and
+# license information.
+#################################################################################
```
adam-a-a: Should use the WaterTAP copyright header

watertap/unit_models/translator_adm1_asm1.py
```diff
@@ -0,0 +1,463 @@
...
+# Framework (IDAES IP) was produced under the DOE Institute for the
+# Design of Advanced Energy Systems (IDAES), and is copyright (c) 2018-2021
+# by the software owners: The Regents of the University of California, through
+# Lawrence Berkeley National Laboratory,  National Technology & Engineering
+# Solutions of Sandia, LLC, Carnegie Mellon University, West Virginia University
+# Research Corporation, et al.  All rights reserved.
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and
+# license information.
+#################################################################################
```
adam-a-a: Replace with WT copyright header

watertap/unit_models/translator_adm1_asm1.py
```diff
@@ -0,0 +1,463 @@
...
+"""
+Modified translator block.
```
adam-a-a:
```suggestion
Translator block representing the ADM1/ASM1 interface.
```

watertap/unit_models/translator_adm1_asm1.py
```diff
@@ -0,0 +1,463 @@
...
+            self.flowsheet().time,
+            doc="Equality X_I equation",
+        )
+        def eq_XI_conc(blk, t):
+            return (
+                blk.properties_out[t].conc_mass_comp["X_I"]
+                == blk.properties_in[t].conc_mass_comp["X_I"]
+            )
+
+        self.readily_biodegradable = Set(
```
adam-a-a: Should we have another set for components with unchanged concentrations (e.g., `X_I` and others where concentration in = concentration out)?

Then, we could eliminate several lines of code and have something like
```
        @self.Constraint(
            self.flowsheet().time,
            self.unchanged_concentrations
            doc="Equality X_I equation",
        )
        def eq_unchanged_concentrations(blk, t, c):
            return (
                blk.properties_out[t].conc_mass_comp[c]
                == blk.properties_in[t].conc_mass_comp[c]
            )
```

It wouldn't change functionality but might make the code more readable.

watertap/unit_models/translator_adm1_asm1.py
```diff
@@ -0,0 +1,463 @@
...
+            has_phase_equilibrium=False,
+            **self.config.outlet_property_package_args
+        )
+
+        # Add ports
+        self.add_port(name="inlet", block=self.properties_in, doc="Inlet Port")
+        self.add_port(name="outlet", block=self.properties_out, doc="Outlet Port")
+
+        mw_n = 14.006 * pyunits.kg / pyunits.kmol
+        mw_c = 12.011 * pyunits.kg / pyunits.kmol
```
adam-a-a: Do we use these numbers in the corresponding property/reaction models for ASM1/ADM1? I only ask because I seem to recall us using something like 14 instead of 14.006 and 12 instead of 12.011 (I could be wrong.)

Maybe it isn't critical, but it could lead to inconsistencies that we'd want to avoid.

watertap/unit_models/translator_adm1_asm1.py
andrewlee94: This code predates the existence of `InitializationError`, and just has not been updated since.

docs/technical_reference/unit_models/translator_adm1_asm1.rst
```diff
@@ -0,0 +1,89 @@
...
+Introduction
+------------
+
+A link is required to translate between biological based and physical or chemical mediated processes
+to develop whole-plant modeling of wastewater treatment. This model mediates the interaction between
+the Anaerobic Digestor Model 1 (ADM1) and the Activated Sludge Model 1 (ASM1).
+
+The model relies on the following key assumption:
+
+   * supports only liquid phase
```
MarcusHolly: This might be a bit too obvious (even for a list of assumptions), but we could add that the inlet property package must be ADM1 and the outlet must be ASM1

MarcusHolly (approved):
LGTM besides one minor typo

docs/technical_reference/unit_models/translator_adm1_asm1.rst
```diff
@@ -0,0 +1,90 @@
...
+   * supports only ADM1 to ASM1 translationS
```
MarcusHolly:
```suggestion
   * supports only ADM1 to ASM1 translations
```

luohezhiming (approved):

watertap/examples/flowsheets/case_studies/anaerobic_digestor/Translator_BSM2.py
```diff
@@ -0,0 +1,131 @@
+###############################################################################
```
adam-a-a: I'm uncertain about the need for this file. It seems like the test has what is needed, so maybe we don't need this example. Any reasoning that I'm missing?

watertap/unit_models/translator_adm1_asm1.py
```diff
@@ -0,0 +1,338 @@
...
+# Technology Laboratory (subject to receipt of any required approvals from
+# the U.S. Dept. of Energy). All rights reserved.
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and license
+# information, respectively. These files are also available online at the URL
+# "https://github.com/watertap-org/watertap/"
+#
+###############################################################################
+"""
+Translator block representing the ADM1/ASM1 interface.
```
adam-a-a: Although this is formulated in a similar fashion to a unit model, I think this translator block belongs in the property_models folder. Ideally, we would have something like
`property_models/wastewater/translators/translator_adm1_asm1.py`
`property_models/wastewater/anaerobic_digestion/ADM1_files...`
`property_models/wastewater/activated_sludge/ASM_files...`

docs/technical_reference/unit_models/index.rst
```diff
@@ -19,5 +19,6 @@ Unit Models
    gac
    osmotically_assisted_reverse_osmosis_0D
    anaerobic_digestor
+   translator_adm1_asm1
```
adam-a-a: I think this should be under property models instead of unit models. See comment below.

watertap/unit_models/translator_adm1_asm1.py
```diff
@@ -0,0 +1,338 @@
...
+
+        @self.Constraint(
+            self.flowsheet().time,
+            self.zero_flow_components,
+            doc="Components with no flow equation",
+        )
+        def return_zero_flow_comp(blk, t, i):
+            return (
+                blk.properties_out[t].conc_mass_comp[i]
+                == 1e-6 * pyunits.kg / pyunits.m**3
```
adam-a-a: Is 1e-6 small enough? Just wondering of the possibility of any trace concentrations that could end up being on the same order of magnitude or smaller. It may not matter but figured I'd ask.

watertap/unit_models/tests/test_translator_adm1_asm1.py
```diff
@@ -0,0 +1,303 @@
...
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and license
+# information, respectively. These files are also available online at the URL
+# "https://github.com/watertap-org/watertap/"
+#
+###############################################################################
+"""
+Tests for Translator ADM1-ASM1 unit model.
+
+Verified against approximated results results from:
```
adam-a-a:
```suggestion
Verified against approximated results from:
```

watertap/unit_models/tests/test_translator_adm1_asm1.py
```diff
@@ -0,0 +1,303 @@
...
+
+    @pytest.mark.unit
+    def test_dof(self, asmadm):
+        assert degrees_of_freedom(asmadm) == 0
+
+    @pytest.mark.solver
+    @pytest.mark.skipif(solver is None, reason="Solver not available")
+    @pytest.mark.component
+    def test_initialize(self, asmadm):
+        initialization_tester(asmadm, optarg={"bound_push": 1e-8})
```
adam-a-a: Is the bound_push necessary?

adam-a-a:
```suggestion
        initialization_tester(asmadm)
```
I checked, and it isn't.

watertap/unit_models/tests/test_translator_adm1_asm1.py
```diff
@@ -0,0 +1,303 @@
...
+
+    @pytest.mark.solver
+    @pytest.mark.skipif(solver is None, reason="Solver not available")
+    @pytest.mark.component
+    def test_solve(self, asmadm):
+        solver = get_solver(options={"bound_push": 1e-8})
```
adam-a-a:
```suggestion
        solver = get_solver()
```

watertap/unit_models/tests/test_translator_adm1_asm1.py
```diff
@@ -0,0 +1,303 @@
...
+        assert pytest.approx(1.322, abs=1e-2) == value(
+            asmadm.fs.unit.outlet.conc_mass_comp[0, "S_NH"]
+        )
+        assert pytest.approx(0.008, abs=1e-2) == value(
+            asmadm.fs.unit.outlet.conc_mass_comp[0, "S_ND"]
+        )
+        assert pytest.approx(0.251, abs=1e-2) == value(
+            asmadm.fs.unit.outlet.conc_mass_comp[0, "X_ND"]
+        )
+        assert pytest.approx(0.095061, abs=1e-2) == value(
```
adam-a-a: Shouldn't we also check the `zero_flow_components`?

watertap/unit_models/tests/test_translator_adm1_asm1.py
```diff
@@ -0,0 +1,303 @@
...
+            asmadm.fs.unit.outlet.pressure[0]
+        )
+        assert pytest.approx(308.15, abs=1e-2) == value(
+            asmadm.fs.unit.outlet.temperature[0]
+        )
+        assert pytest.approx(0.1308, abs=1e-2) == value(
+            asmadm.fs.unit.outlet.conc_mass_comp[0, "S_I"]
+        )
+        assert pytest.approx(0.2585, abs=1e-2) == value(
+            asmadm.fs.unit.outlet.conc_mass_comp[0, "S_S"]
```
adam-a-a: We typically used relative tolerances and reserve absolute tolerances for checking 0 values. Additionally, preferable to check relative tolerance of 1e-3 to 1e-5, if we're being consistent with other testing on WaterTAP.

adam-a-a (changes requested):
Some suggested changes, mostly with where things live. The most important part seems to be good to go though.

watertap/unit_models/translator_adm1_asm1.py
agarciadiego: @adam-a-a one of the only reasons I would avoid this change is because of how the translator blocks are used in the flowsheet. They are used like unit models, even having streams come in and out.

MarcusHolly: I agree with Alejandro here - I think it makes more sense to have this under unit models since translators are functioning like one, and I think this would be the first place someone would look for this translator block. IDAES also has their translator block under the unit_model folder. It seems like the main advantage of putting everything in the property models folder is that we can take advantage of the anaerobic_digestion and activated_sludge sub-folders. So if we are worried about cluttering the unit_model folder with a bunch of BSM2-related files, we could create sub-folder(s) here.

watertap/unit_models/tests/test_translator_adm1_asm1.py
agarciadiego: We are testing them ("X_BH, "X_BA”, "X_P, "S_O, "S_NO")

watertap/examples/flowsheets/case_studies/anaerobic_digestor/Translator_BSM2.py
agarciadiego: I was showing the implementation of the translator block for users, I will not include it in this PR and add it when we do the PR on the other translator to show users

docs/technical_reference/unit_models/translator_adm1_asm1.rst
```diff
@@ -0,0 +1,90 @@
+ADM1 to ASM1 Translator
+=======================
+
+Introduction
+------------
+
+A link is required to translate between biological based and physical or chemical mediated processes
```
adam-a-a:
```suggestion
A link is required to translate between biological and physically- or chemically-mediated processes
```

watertap/unit_models/tests/test_translator_adm1_asm1.py
```diff
@@ -0,0 +1,303 @@
+###############################################################################
+# WaterTAP Copyright (c) 2021, The Regents of the University of California,
```
adam-a-a: Update to latest copyright header: https://github.com/watertap-org/watertap/pull/936

watertap/unit_models/translator_adm1_asm1.py
adam-a-a: Fair enough! I suppose we can consider any future sub-folders in a subsequent PR.

watertap/unit_models/translator_adm1_asm1.py
```diff
@@ -0,0 +1,338 @@
...
+
+
+# Set up logger
+_log = idaeslog.getLogger(__name__)
+
+
+@declare_process_block_class("Translator_ADM1_ASM1")
+class TranslatorDataAA(TranslatorData):
+    """
+    Standard Translator Block Class
```
adam-a-a: Update this doc string or maybe replace with doc string from up top

watertap/unit_models/translator_adm1_asm1.py
```diff
@@ -0,0 +1,338 @@
...
+
+__author__ = "Alejandro Garciadiego, Andrew Lee"
+
+
+# Set up logger
+_log = idaeslog.getLogger(__name__)
+
+
+@declare_process_block_class("Translator_ADM1_ASM1")
+class TranslatorDataAA(TranslatorData):
```
adam-a-a:
```suggestion
class TranslatorDataADM1ASM1(TranslatorData):
```

docs/technical_reference/unit_models/translators/index.rst
```diff
@@ -0,0 +1,7 @@
+Tranlsators
```
adam-a-a:
```suggestion
Translators
```

docs/technical_reference/unit_models/translators/translator_adm1_asm1.rst
```diff
@@ -0,0 +1,90 @@
...
+=======================
+
+Introduction
+------------
+
+A link is required to translate between biological and physically- or chemically-mediated processes
+to develop whole-plant modeling of wastewater treatment. This model mediates the interaction between
+the Anaerobic Digestor Model 1 (ADM1) and the Activated Sludge Model 1 (ASM1).
+
+The model relies on the following key assumption:
```
adam-a-a:
```suggestion
The model relies on the following key assumptions:
```

adam-a-a (approved):
Nice job!

PR #937, opened 2023-02-28
Update IDAES to 2.0.0rc0

lbianchi-lbl:
## Summary/Motivation:

- Start testing WaterTAP with Feb release of IDAES

### Legal Acknowledgement


bknueven (approved):

watertap/property_models/NaCl_prop_pack.py
```diff
@@ -186,9 +186,14 @@ def define_metadata(cls, obj):
...
-                "osm_coeff": {"method": "_osm_coeff"},
                 "pressure_osm_phase": {"method": "_pressure_osm_phase"},
                 "enth_mass_phase": {"method": "_enth_mass_phase"},
+            }
+        )
+
+        obj.define_custom_properties(
+            {
+                "osm_coeff": {"method": "_osm_coeff"},
                 "enth_flow": {"method": "_enth_flow"},
```
andrewlee94, line 196: This looks like a case where we should add the property to the IDAES property sets - an issue (or PR) would be welcome.

watertap/property_models/anaerobic_digestion/adm1_properties.py
```diff
@@ -135,6 +135,10 @@ def define_metadata(cls, obj):
                 "pressure": {"method": None},
                 "temperature": {"method": None},
                 "conc_mass_comp": {"method": None},
+            }
+        )
+        obj.define_custom_properties(
+            {
                 "anions": {"method": None},
```
andrewlee94, line 141: Not really part of this PR, but these don't look like they are "properties"

watertap/property_models/anaerobic_digestion/adm1_properties_vapor.py
```diff
@@ -109,6 +109,10 @@ def define_metadata(cls, obj):
                 "pressure": {"method": None},
                 "temperature": {"method": None},
                 "conc_mass_comp": {"method": None},
+            }
+        )
+        obj.define_custom_properties(
+            {
                 "p_sat": {"method": None},
```
andrewlee94, line 115: Minor comment, but the standard name is `"pressure_sat"`

andrewlee94 (approved):

watertap/property_models/NaCl_prop_pack.py
adam-a-a: Added an issue: https://github.com/IDAES/idaes-pse/issues/1113

watertap/property_models/anaerobic_digestion/adm1_properties.py
adam-a-a: I believe the actual properties used for these would either be `conc_mass_phase_comp` or `conc_mol_phase_comp`. @agarciadiego had some reasoning for why it was done this way, but I am forgetting what the reason was.

watertap/property_models/anaerobic_digestion/adm1_properties_vapor.py
adam-a-a: @bknueven made an issue and I referenced the issue in our master issue which includes efforts related to ADM1

adam-a-a (approved):
LGTM - I think I noticed all the changes as a result of IDAES modifications although I don't know what the repercussions would be without said changes (e.g., exception raised when not using `define_custom_properties`? Advantage of using `list_required_properties` as opposed to what we had before?)

PR #943, opened 2023-03-03
Refine Scaling Utility

### LINKED ISSUE #942: Refine scaling utility
I think that the `transform_property_constraints` utility would benefit from a doc string explaining what it is used for (despite it seeming obvious to us) and also, I see an issue with hard-coding a check for `"eq_" + var_str`. Does IDAES v2 enforce that property constraints be named that way? If not, then one could use this utility intending to scale constraints, but then they'd miss any constraints that don't follow this naming convention.

https://github.com/watertap-org/watertap/blob/31edc1a26a0eb94caacefbd5e04473ef4fab1bb3/watertap/core/util/scaling.py#L18-L30

### LINKED ISSUE #979: Refine testing for scaling utility function
### Description

The test file should directly use the utility function rather than hard-coding it. Additionally, we should consider developing a dummy metadata set rather than importing a property package. These suggested changes exist as To-Do items in the test file itself.

### Motivation

The purpose of this test file is to test `transform_property_constraints`, so, ideally, this function should be directly used as opposed to hard-coding it. Using dummy variables should also make it easier to test for failures since the imported property package should not be setting off any failures. 

### Possible Implementation

_No response_

### Additional Context

_No response_

MarcusHolly:
## Fixes/Resolves:
Issue #942 

- Adds a sentence describing what the utility function does
- Adds an else conditional that displays a logger warning if a property constraint is not scaled

### Legal Acknowledgement


bknueven (approved):

watertap/core/util/scaling.py
```diff
@@ -27,3 +32,10 @@ def transform_property_constraints(self):
                 for ind, c in con.items():
                     sf = iscale.get_scaling_factor(var[ind], default=1, warning=True)
                     iscale.constraint_scaling_transform(c, sf)
+            else:
+                msg = (
+                    f"One or more property constraints are not scaled. "
+                    "This suggests that the user defined a property in metadata but failed to "
+                    "follow the the naming convention for its constraint: 'eq_' + name in metadata."
+                )
+                _log.warning(msg)
```
adam-a-a:
```suggestion
            else:
                msg = (
                    f"If there was a property constraint written for the variable, {var}, that constraint was not scaled. "
                    "The transform_property_constraints tool expects constraints to have the following naming convention: 'eq_' + '{var_str}'. This suggests that the user may have defined a property in metadata but failed to "
                    "follow the naming convention for its constraint. If there is no property constraint associated with the {var_str} variable, this warning can be ignored."
                )
                _log.warning(msg)
```

Something like this, since in some cases, there is no constraint- just a reference to a parameter.

adam-a-a: You may have to fix the f-string for it to work. Also we should have some test file for this utility (see `test_initialization` to compare with test for `watertap.core.util.initialization`).

adam-a-a (changes requested):
Have a suggested change and should also add a test file for this

MarcusHolly:
> Have a suggested change and should also add a test file for this

@adam-a-a For the test file, should I be importing an existing property package and testing that or should I create a new mini property package in this test file that has its own (made-up) metadata and constraints? Either way, I've tried setting up both but I'm not exactly sure how to go about doing this, so maybe this should be a separate PR (I could make an issue for it) if we don't want to hold this one up for the March release.

adam-a-a:
> > Have a suggested change and should also add a test file for this
> 
> @adam-a-a For the test file, should I be importing an existing property package and testing that or should I create a new mini property package in this test file that has its own (made-up) metadata and constraints? Either way, I've tried setting up both but I'm not exactly sure how to go about doing this, so maybe this should be a separate PR (I could make an issue for it) if we don't want to hold this one up for the March release.

@MarcusHolly I would say you could do the latter- dummy metadata/constraints. I think we should try to get at least one or two simple tests in for the release. Maybe we should check that the logging warning displays as expected. See here for an example of using the caplog fixture that pytest has: https://github.com/IDAES/idaes-pse/blob/90f762afda2134e290b8fa1c78aa64fee9049954/idaes/generic_models/properties/core/eos/tests/test_enrtl.py#L188-L207

I never used caplog myself but looked it up here: https://docs.pytest.org/en/latest/how-to/logging.html

adam-a-a:
Actually, since we want something quick for the sake of having some testing vs no testing, it may be easier to just use a property package.

MarcusHolly:
> Actually, since we want something quick for the sake of having some testing vs no testing, it may be easier to just use a property package.

@adam-a-a Added a test file which uses a property package, but there are a few To-Do items I listed. Mainly, I couldn't get the `transform_property_constraints(self)`, which I realize is the entire purpose of this file... I've played around with it a decent bit, but I always get a `AttributeError` stating the object has no attribute 'params'. So the current workaround is to hard-code the utility function and test it that way, which obviously isn't ideal.

The second To-Do issue is forcing the logger warning to display. If I just use a property package, this warning will not appear since all the constraints are scaled, and I'm not sure if it's even possible to append a dummy variable into the metadata that would cause this warning to display. This may be more straightforward after addressing the first To-Do.

adam-a-a (approved):
LGTM- @MarcusHolly if we don't have one already, can you add an issue reminding us of these ToDo notes with a summary of what would need to be addressed?

MarcusHolly:
> LGTM- @MarcusHolly if we don't have one already, can you add an issue reminding us of these ToDo notes with a summary of what would need to be addressed?

Yep, will do.

PR #944, opened 2023-03-03
ED documentation update 

### LINKED ISSUE #939: Clarify multi-component capability of ED in documentation
### Description

Add a warning or note to ED documentation explaining limitations of trying to model treatment of a multi-component solution.

### Motivation

One of our academic colleagues pointed out how the documentation implies that multiple components can be handled by the ED unit. However, even though one could pass multiple components to ED without errors occurring, the mathematical representation of component diffusivities and transport numbers is lacking in the literature and would need to be provided.

### Possible Implementation

Add warning in ED doc

### Additional Context

_No response_

lbibl:
### Legal Acknowledgement


MarcusHolly (approved):
LGTM - not sure if we want to add this disclaimer as a note or a warning text bubble just so that it stands out a bit more and is harder to ignore. Regardless, I think the content of the message is good.

hunterbarber (approved):
Per what Marcus said, we don't have a neat wrapper to be consistent across all documentation so do what you will with the formatting, but content is good.

PR #946, opened 2023-03-03
Reorganize model diagnostics

bknueven:
## Summary/Motivation:
Anticipate adding more model diagnostic tools in the near term; having a single place will make understanding what we have easier.

Reorganize existing model diagnostics into a single folder.

### Legal Acknowledgement


watertap/core/util/model_diagnostics/ipopt_initialization.py
```diff
@@ -0,0 +1,135 @@
...
+#################################################################################
+
+import pyomo.environ as pyo
+from idaes.core.util.scaling import get_scaling_factor, __none_left_mult
+from idaes.core.solvers import get_solver
+
+
+def assert_no_initialization_perturbation(blk, optarg=None, solver=None):
+    """
+    Assert that IPOPT will *not* move the initialization
```
adam-a-a, line 20: Can you explain example use cases for the novel developer/user on when this tool would come in handy? When do I use this? How does it help me?

watertap/core/util/model_diagnostics/ipopt_initialization.py
```diff
@@ -0,0 +1,135 @@
...
+        print(
+            f"IPOPT will move scaled initial value for variable {v.name} from {val:e} to {result:e}"
+        )
+
+
+def generate_initialization_perturbation(
+    blk, bound_push=1e-2, bound_frac=1e-2, bound_relax_factor=1e-8, user_scaling=False
+):
+    """
+    Generate the initialization perturbations performed by IPOPT for a given Block
```
adam-a-a, line 90: Same as comment above- I essentially want to tease out the watered down explanation of where these tools can be most useful so that we can later convey this in documentation to the general user.

bknueven: This function attempts to determine if the initial value provided by the user/Pyomo will be changed prior to the first iteration of Ipopt. This can cause issues within our general sequential initialization approach. I don't think this module currently reports what `ipopt-watertap` does, however. And maybe `ipopt-watertap` address this issue to a sufficient extent that this entire module can be removed.

I agree that this module needs better documentation; as-is, this PR just reorganizes existing code. I was holding off refinements for subsequent PRs to ease the review burden.

adam-a-a: Do we have an issue related to model diagnostics? Maybe it would be could to have a task list and some discussion around next steps.

adam-a-a (approved):
LGTM - I just had one more question/comment. Since this is mostly reorganization and might be reconsidered if not applicable to ipopt-watertap, I don't see any reason to hold this up.

watertap/core/util/model_diagnostics/ipopt_initialization.py
bknueven: Good idea -- I opened #955 as a start. Feel free to add / edit / discuss there.

lbibl (approved):

PR #948, opened 2023-03-03
Add modified ADM1 model with P extension

luohezhiming:
## Fixes/Resolves:

-
-

### Legal Acknowledgement


watertap/property_models/anaerobic_digestion/adm1_properties_P.py
```diff
@@ -0,0 +1,453 @@
+#################################################################################
```
adam-a-a: I figured we would discuss how to implement this first, but I guess this is just a draft. Have you thought about inheriting from the existing adm1 class? If there is a lot of the same code, maybe we don't need to copy paste everything.

MarcusHolly: @adam-a-a The only parts of the code that should be the same in the end are the lists of parameters (although there are quite a few additions and removals to these list). So perhaps it is worth inheriting these from ADM1. But to play devil's advocate, it'd make the code easier to follow if all the parameters were listed out in this file as well.

adam-a-a:
@luohezhiming @MarcusHolly Can you put snapshots of the detailed IPOPT output?

watertap/property_models/anaerobic_digestion/modified_adm1_reactions.py
```diff
@@ -0,0 +1,2564 @@
...
+            return (
+                self.conc_mol_nh4
+                == self.conc_mass_comp_ref["S_IN"]
+                / (14 * pyo.units.kg / pyo.units.kmol)
+                - self.conc_mol_nh3
+            )
+
+        self.concentration_of_nh4 = pyo.Constraint(
+            rule=concentration_of_nh4_rule,
+            doc="constraint concentration of pro-",
```
adam-a-a: update doc string

watertap/property_models/anaerobic_digestion/modified_adm1_reactions.py
```diff
@@ -0,0 +1,2564 @@
...
+                - self.conc_mass_bu / (160 * pyo.units.kg / pyo.units.kmol)
+                - self.conc_mass_va / (208 * pyo.units.kg / pyo.units.kmol)
+                - self.S_OH
+                - self.state_ref.anions
+                == 0
+            )
+
+        self.S_H_cons = pyo.Constraint(
+            rule=S_H_rule,
+            doc="constraint concentration of pro-",
```
adam-a-a: update doc string

watertap/property_models/anaerobic_digestion/modified_adm1_reactions.py
```diff
@@ -0,0 +1,2564 @@
...
+            return self.S_OH == self.KW / self.S_H
+
+        self.S_OH_cons = pyo.Constraint(
+            rule=S_OH_rule,
+            doc="constraint concentration of OH",
+        )
+
+        def S_H_rule(self):
+            return (
+                self.state_ref.cations
```
adam-a-a, line 2102: So, now that potassium and magnesium are tracked components, shouldn't they be incorporated into this constraint? Perhaps we need to double-check the reference to see how these components are handled. Overall, I would think that they need to be tied to `cations` and `anions`, at least indirectly (i.e., included in this constraint) if not directly (e.g., add constraint that relates potassium and magnesium concentrations to the total cation and anion concentrations).

adam-a-a:
@MarcusHolly @luohezhiming - it appears that scaling was the issue (or at least part of it), so I fixed things just enough to get tests to pass--mainly to get initialization and solve passing. I would consider some of this a temporary fix. A more permanent solution with better scaling should follow this, but I anticipate that it'll take us some time to improve scaling not only in this modified ADM1 package, but also in the original ADM1 package, anaerobic digestor unit, and possibly the ASM models too. I imagine that'll all have to come in a series of subsequent PRs, so perhaps we just need to do more cleanup for this PR, make sure the implementation is correct and complete, and ensure that the regression tests line up with values expected from the literature.

MarcusHolly:
> @MarcusHolly @luohezhiming - it appears that scaling was the issue (or at least part of it), so I fixed things just enough to get tests to pass--mainly to get initialization and solve passing. I would consider some of this a temporary fix. A more permanent solution with better scaling should follow this, but I anticipate that it'll take us some time to improve scaling not only in this modified ADM1 package, but also in the original ADM1 package, anaerobic digestor unit, and possibly the ASM models too. I imagine that'll all have to come in a series of subsequent PRs, so perhaps we just need to do more cleanup for this PR, make sure the implementation is correct and complete, and ensure that the regression tests line up with values expected from the literature.

@adam-a-a Sounds like a plan. Thanks for helping out with this!

luohezhiming:
> @MarcusHolly @luohezhiming - it appears that scaling was the issue (or at least part of it), so I fixed things just enough to get tests to pass--mainly to get initialization and solve passing. I would consider some of this a temporary fix. A more permanent solution with better scaling should follow this, but I anticipate that it'll take us some time to improve scaling not only in this modified ADM1 package, but also in the original ADM1 package, anaerobic digestor unit, and possibly the ASM models too. I imagine that'll all have to come in a series of subsequent PRs, so perhaps we just need to do more cleanup for this PR, make sure the implementation is correct and complete, and ensure that the regression tests line up with values expected from the literature.

Thank you very much for digging into this, Adam! I am glad all tests passed for nor and we could plan to have follow-up PRs for finally resolving the scaling issue

watertap/property_models/anaerobic_digestion/modified_adm1_reactions.py
MarcusHolly: @luohezhiming I've addressed this in my latest commits, so feel free to look over/ask questions about my implementation.

MarcusHolly:
I've tried re-running the MacOS setup test, but it still fails. Is it fine if this test is just skipped?

watertap/property_models/anaerobic_digestion/modified_adm1_reactions.py
```diff
@@ -0,0 +1,2604 @@
...
+            )
+
+        self.concentration_of_Mg = pyo.Constraint(
+            rule=concentration_of_Mg_rule,
+            doc="constraint concentration of Mg",
+        )
+
+        def concentration_of_K_rule(self):
+            return self.conc_mol_Mg == self.conc_mass_comp_ref["X_PP"] / (
+                300.41 * pyo.units.kg / pyo.units.kmol
```
adam-a-a: This seems to be a copy of the previous constraint.

MarcusHolly: Thx for catching this - the expressions are equivalent, but I did forget to replace `conc_mol_Mg` with `conc_mol_K`

adam-a-a: What molecule does that molecular weight correspond to? I figure X_PP represents polyphosphates but unsure what the MW assumption specifically ties to.

MarcusHolly: Screenshot from Excel located here in Box: https://app.box.com/file/1153207546703
image

adam-a-a:
> I've tried re-running the MacOS setup test, but it still fails. Is it fine if this test is just skipped?

So, I think it depends on what is failing in the test. Originally, it was just a matter of absolute tolerance. However, after latest changes, it appears to be more of a scaling issue (I suspect) that now throws off the AD unit's initialization. It'd be better to try to resolve that before we resort to skipping. That said, it'd be helpful for a Mac user to tinker with this. @bknueven could you take a look at this? I am sure you will find that there is substantial room for improvement in scaling and initialization for these models, but for now, we're just looking to hit the _bare minimum requirement_ to get an initial implementation passing tests and merged. Following that, we'd have subsequent PRs to address scaling and initialization.

watertap/property_models/anaerobic_digestion/modified_adm1_reactions.py
adam-a-a: @MarcusHolly How were K and Mg handled in the literature? I would assume they'd account for those species as I suggested and as you have now implemented; I just want a sanity check on how well our implementation reflects the Flores et al. reference.

adam-a-a:
Another question/comment that I want @MarcusHolly and @luohezhiming to think about ahead of our meeting today: our slides showing ADM1->ASM2D and ASM2D ->ADM1 give a table that lists struvite (MgNH4PO4) as well as calcium phosphate (Ca3(PO4)2), both of which will be key for our 2336 team collaboration. It seems that the current implementation hasn't accounted for those species yet (but I may have missed that). If so, how much more would it take to incorporate those species? Think about it, and let's discuss later today =)

watertap/property_models/anaerobic_digestion/modified_adm1_reactions.py
adam-a-a: > > I've tried re-running the MacOS setup test, but it still fails. Is it fine if this test is just skipped?
> 
> So, I think it depends on what is failing in the test. Originally, it was just a matter of absolute tolerance. However, after latest changes, it appears to be more of a scaling issue (I suspect) that now throws off the AD unit's initialization. It'd be better to try to resolve that before we resort to skipping. That said, it'd be helpful for a Mac user to tinker with this. @bknueven could you take a look at this? I am sure you will find that there is substantial room for improvement in scaling and initialization for these models, but for now, we're just looking to hit the _bare minimum requirement_ to get an initial implementation passing tests and merged. Following that, we'd have subsequent PRs to address scaling and initialization.

The other approach we can take is to skip the Mac test for now, merge this PR, get the associated translator block PR approved and merged, then come back and refine scaling, etc. 

However, please see my inquiry about additional species inclusion (struvite and calcium phosphate).

MarcusHolly: I think your question is essentially where is this equation coming from? I just modified the equation presented in the reference for the regular adm1 reaction file: 
image
Our reference (from what I can tell) does not spell out how K and Mg are handled. If it does, it'd probably be in one of the MATLAB files.

MarcusHolly:
> Another question/comment that I want @MarcusHolly and @luohezhiming to think about ahead of our meeting today: our slides showing ADM1->ASM2D and ASM2D ->ADM1 give a table that lists struvite (MgNH4PO4) as well as calcium phosphate (Ca3(PO4)2), both of which will be key for our 2336 team collaboration. It seems that the current implementation hasn't accounted for those species yet (but I may have missed that). If so, how much more would it take to incorporate those species? Think about it, and let's discuss later today =)

The quick answer is that we don't currently account for these species. However, they are accounted for in the Precipitation Extension in the Flores-Alsina paper: https://app.box.com/file/1153202083720?s=he1whq9a66fvxj8y39puqsqijzuy7sbr

I haven't looked too closely at how much effort this would require, so I'll try to look into it before the meeting.

watertap/property_models/anaerobic_digestion/modified_adm1_reactions.py
MarcusHolly: > > > I've tried re-running the MacOS setup test, but it still fails. Is it fine if this test is just skipped?
> > 
> > 
> > So, I think it depends on what is failing in the test. Originally, it was just a matter of absolute tolerance. However, after latest changes, it appears to be more of a scaling issue (I suspect) that now throws off the AD unit's initialization. It'd be better to try to resolve that before we resort to skipping. That said, it'd be helpful for a Mac user to tinker with this. @bknueven could you take a look at this? I am sure you will find that there is substantial room for improvement in scaling and initialization for these models, but for now, we're just looking to hit the _bare minimum requirement_ to get an initial implementation passing tests and merged. Following that, we'd have subsequent PRs to address scaling and initialization.
> 
> The other approach we can take is to skip the Mac test for now, merge this PR, get the associated translator block PR approved and merged, then come back and refine scaling, etc.
> 
> However, please see my inquiry about additional species inclusion (struvite and calcium phosphate).

This is probably something we'll talk about more during today's meeting, but there is a complication with the Translator block PR. Our current implementation of ASM2d does not include S_IC, S_K, or S_Mg terms. So we will either need to update our current ASM2d implementation or add a "modified_asm2d". I'd probably lean towards the latter. I'm currently looking for literature sources on this, but my first instinct is to use the ASM GitHub created by the same authors: https://github.com/wwtmodels/Activated-Sludge-Models. My immediate concern with this is that everything appears to be inside MATLAB files this time (there's no Excel with results)

bknueven:
For this PR we should just mark the failing tests with `requires_idaes_solver`. We should be removing that GHA check soon anyways (see #970).

watertap/property_models/anaerobic_digestion/tests/test_modified_adm1_reaction.py
```diff
@@ -0,0 +1,771 @@
...
+    def test_rxn_rate(self, model):
+        assert isinstance(model.rxns[1].reaction_rate, Var)
+        assert len(model.rxns[1].reaction_rate) == 25
+        assert isinstance(model.rxns[1].rate_expression, Constraint)
+        assert len(model.rxns[1].rate_expression) == 25
+
+    @pytest.mark.unit
+    def test_get_reaction_rate_basis(self, model):
+        assert model.rxns[1].get_reaction_rate_basis() == MaterialFlowBasis.mass
+
```
adam-a-a, line 508:
```suggestion

@pytest.mark.requires_idaes_solver
```

MarcusHolly:
@adam-a-a I've avoided changing some of the pytest marks from component to unit because this test file follows the same conventions as other similar test files for adm1, asm1, and asm2d. So if these tests really are incorrectly marked, then maybe we/I should just fix them all in one-go (in a separate PR) rather than just updating this test file.

adam-a-a (approved):
LGTM- this model needs additions anyway that will take place in subsequent PRs

MarcusHolly (approved):

PR #953, opened 2023-03-09
Add electrocoagulation zero-order model

### LINKED ISSUE #739: Electrocoagulation model development
- [x] Implement Phenomenological model for EC
- [x] Implement Costing for EC
- [x] EC Documentation
- [ ] Demonstration of EC in flowsheet

kurbansitterley:
## Fixes/Resolves:

Adds EC ZO model to WaterTAP.  Resolves #739

## Summary/Motivation:
Add the EC model provided by NMSU via the watertap-seto subproject to main WaterTAP.

-Adds ZO Electrocoagulation model (in unit_models/zero_order) and test file, yaml

### Legal Acknowledgement


kurbansitterley:
@srikanthallu @wigging @lbibl @adam-a-a @avdudchenko There are two EC models in this draft PR complete with test files. I ended up doing this because it was cumbersome to reconcile the differences between the OG EC ZO model with those from the watertap-seto project whilst still including both CONFIG options for calculating cell over potential using WaterTAP's ZO model framework, particularly dealing with the yaml file, subtypes, subtypes of subtypes...

So to incorporate those changes, I just made a full detailed unit model that uses the MCAS property package. My preference would be to keep the detailed model and build off of it as EC model development progresses, but I will leave it to the community of developers here to make the final decision since I don't know what direction that development will take.

wigging:
The description says "Adds one of two EC models to WaterTAP" but I see two models in this pull request:
- `watertap/unit_models/electrocoagulation.py`
- `watertap/unit_models/zero_order/electrocoagulation_zo.py`

Since the branch for this pull request is named `ec_zo_model` I would expect to just see the zero-order model `electrocoagulation_zo.py` as part of this pull request. In the Changes Proposed section you mention the two models, but in the main description you state that this pull request adds one of the two models. So is the purpose of this pull request to submit two models or just one model? Does the zero-order model depend on something from the unit model `electrocoagulation.py`? If these are indeed two separate models, I would expect separate pull requests to be made for each model. One pull request for the `unit_models/electrocoagulation.py` and another pull request for the `unit_models/zero_order/electrocoagulation_zo.py`.

srikanthallu:
So @wigging , after looking at the PR - we need to consolidate better as both of them are ZO models. I think what @kurbansitterley is saying is that SETO ZO model does not conform with waterTAP ZO configs due to cell over potential calculations.

adam-a-a:
I _think_  the consensus was to keep the ZO formulation around for the time being and consider adding either a subtype and/or config option to toggle between options for, e.g., energy consumption calculation. This can be applied to dealing with the various overpotential assumptions too. For now, I'd suggest not getting hung up over whether any of this technically fits within the "zero-order" categorization. We can always refactor later once the big picture becomes more apparent and new predictive relationships are subsequently acquired.

kurbansitterley:
The consensus was confusing to me, so in the interest of moving forward I am just going to keep the ZO model here and I still have the detailed model on my branch. I think the detailed model will be better for using Water Tectonics data but we will just deal with the need as it comes up.

wigging:
> The consensus was confusing to me, so in the interest of moving forward I am just going to keep the ZO model here and I still have the detailed model on my branch. I think the detailed model will be better for using Water Tectonics data but we will just deal with the need as it comes up.

@kurbansitterley Yes, keep the zero-order model here. Just submit that model (`unit_models/zero_order/electrocoagulation_zo.py`) for this pull request. You can do a separate pull request for the more detailed model after we get this one done.

watertap/unit_models/zero_order/electrocoagulation_zo.py
```diff
@@ -0,0 +1,611 @@
...
+
+
+class ElectrodeMaterial(StrEnum):
+    aluminum = "aluminum"
+    iron = "iron"
+
+
+class ReactorMaterial(StrEnum):
+    pvc = "pvc"
+    stainless_steel = "stainless_steel"
```
wigging: Is ReactorMaterial being used by this model? Maybe I'm missing it, but I don't see where this is actually utilized elsewhere in the code.

watertap/unit_models/zero_order/electrocoagulation_zo.py
```diff
@@ -0,0 +1,611 @@
...
+    )
+
+    CONFIG.declare(
+        "reactor_material",
+        ConfigValue(
+            default="pvc",
+            domain=In(ReactorMaterial),
+            description="Reactor material",
+        ),
+    )
```
wigging, line 79: Related to my earlier comment, it doesn't look like reactor_material is used in this model.

kurbansitterley: You are correct, I noticed that too. I will be adding it shortly.

wigging:
I get the following warning:

```
[WARNING] idaes.core.base.costing_base: flow_expr is an expression with a lower bound of less than zero. Costing requires that all flows have a lower bound equal to or greater than zero to avoid negative costs.
```

when I run the example shown below. Anyone have a suggestion on how to fix this? How do I set the lower bound of the flows?

I went through the code for `ElectrocoagulationZOData` class and bounds are set for everything so I don't know why this warning occurs.

```python
from pyomo.environ import value

from idaes.core import FlowsheetBlock
from idaes.core.solvers import get_solver
from idaes.core import UnitModelCostingBlock

from watertap.core.zero_order_properties import WaterParameterBlock
from watertap.unit_models.zero_order import ElectrocoagulationZO
from watertap.core.zero_order_costing import ZeroOrderCosting

def main():

    model = ConcreteModel()
    model.db = Database()

    model.fs = FlowsheetBlock(dynamic=False)
    model.fs.params = WaterParameterBlock(solute_list=["tds", "tss", "toc"])

    model.fs.unit = ElectrocoagulationZO(property_package=model.fs.params, database=model.db)
    model.fs.unit.inlet.flow_mass_comp[0, "H2O"].fix(43.8)
    model.fs.unit.inlet.flow_mass_comp[0, "toc"].fix(0.004599)
    model.fs.unit.inlet.flow_mass_comp[0, "tss"].fix(0.5527998)
    model.fs.unit.inlet.flow_mass_comp[0, "tds"].fix(5.256)

    model.fs.unit.load_parameters_from_database(use_default_removal=True)

    model.fs.costing = ZeroOrderCosting()

    ec = model.fs.unit
    ec.costing = UnitModelCostingBlock(flowsheet_costing_block=model.fs.costing)

    model.fs.costing.cost_process()
    model.fs.costing.add_LCOW(ec.properties_treated[0].flow_vol)
    model.fs.costing.add_electricity_intensity(ec.properties_treated[0].flow_vol)

    solver = get_solver()
    results = solver.solve(model)

    print(results)
    model.fs.unit.report()

    print("\nLCOW = ", value(model.fs.costing.LCOW))

if __name__ == "__main__":
    main()
```

srikanthallu (approved):
LGTM

wigging (approved):

watertap/unit_models/zero_order/electrocoagulation_zo.py
```diff
@@ -0,0 +1,655 @@
...
+            within=PositiveReals,
+            doc="Current required per reactor",
+        )
+
+        self.tds_to_cond_conversion = Param(
+            initialize=5e3,
+            mutable=True,
+            units=(pyunits.mg * pyunits.m) / (pyunits.liter * pyunits.S),
+            within=PositiveReals,
+            doc="Conersion factor for mg/L TDS to S/m",
```
adam-a-a:
```suggestion
            doc="Conversion factor for mg/L TDS to S/m",
```

watertap/unit_models/zero_order/electrocoagulation_zo.py
```diff
@@ -0,0 +1,655 @@
...
+            bounds=(1, 2000),
+            units=pyunits.ampere / pyunits.m**2,
+            doc="Current density",
+        )
+
+        self.power_required = Var(
+            initialize=1,
+            bounds=(0, None),
+            units=pyunits.watt,
+            doc="Power requireds",
```
adam-a-a:
```suggestion
            doc="Power required",
```

PR #954, opened 2023-03-09
GAC model overhaul

### LINKED ISSUE #868: Corrections and advancement of GAC model (March)
### Description

Itemize potential improvements to the GAC model with varying priority (generally listed high -> low priority):
- [x] call molecular diffusion coefficient from prop pack #896
- [x] embed molar volume calculation #896 
- [x] update documentation for new naming conventions (e.g. MCAS prop pack)
- [x] check in=out for inert (not in `target_species`)
- [x] correct referencing to `min_elap_time` when solving for mtz during ss assumptions
- [x] consolidate report() output
- [x] potentially remove `particle_porosity`, `particle_dens_sol`, and `eq_particle_porosity`
- [x] dynamic sf for gac.equil_conc as f(inlet_conc)
- [x] add ConfigurationError error handling

### Motivation

List includes changes for both clarity and model rigor

### Possible Implementation

_No response_

### Additional Context

Migrating more detailed and intensive tasks from #683 to a new issue to close and push small changes
Ongoing list, anticipated March release

hunterbarber:
## Fixes/Resolves:
Major fix to the methodology for the steady state approximation method that would fail when (length of the mass transfer zone) > (bed length)
Also included full model overhaul for overall model stability and clarity of code

## Summary/Motivation:
Challenges with GAC model stability

- changing testing to match changes
- changing documentation to match changes
- changing variable naming and inline documentation for clarity
- Reordered declaration of variables and constraints to more logically group caclulation steps (e.g., sizing dimensions grouped together, steady state variables grouped together, etc.)
- removed degenerate equations
- removed `particle_porosity` and `particle_dens_sol` as required variables as they were only present in an isolated equation and unnecessarily added to the dof
- declares `target_species` and `inert_species` sets globally (I could use some input on having `target_species` config argument as a list or str, declaring as a python set led to errors of converting an unordered python set to an ordered pyomo Set())
- new steady state approximation methodology (introduces `finite_elements_ss_approximation`, and `ele_` variables)
- Improved variable initialized values
- improved state block initialization by initializing variables to 0 that are later fixed to 0 (`gac_removed.flow_mol_phase_comp['Liq','H20']`)
- Cleaned up reporting in `.report()`

### Legal Acknowledgement


hunterbarber:
Will update out of draft when this is ready for review, but still need to update the documentation as of now.

watertap/unit_models/gac.py
```diff
@@ -61,9 +63,8 @@ class SurfaceDiffusionCoefficientType(Enum):
 @declare_process_block_class("GAC")
 class GACData(InitializationMixin, UnitModelBlockData):
     """
-    Initial Granular Activated Carbon Model -
-    currently should be used for only with ion_DSPMDE_prop_pack with
-    a single solute and single solvent as water
+    Empirical CPHSDM Granular Activated Carbon Model -
```
adam-a-a:
```suggestion
    Empirical Constant-Pattern-Homogeneous-Surface-Diffusion Model (CPHSDM) for Granular Activated Carbon -
```

watertap/unit_models/gac.py
```diff
@@ -270,475 +299,394 @@ def build(self):
...
+            doc="bed length",
         )
-
         self.bed_diameter = Var(
             initialize=1,
             bounds=(0, None),
             domain=NonNegativeReals,
             units=units_meta("length"),
-            doc="Adsorber bed diameter, valid if considering only a single adsorber",
+            doc="bed diameter, valid if considering only a single adsorber",
```
adam-a-a, line 414: And for adsorbers in parallel, assuming each has this diameter?

watertap/unit_models/gac.py
```diff
@@ -270,475 +299,394 @@ def build(self):
...
             domain=NonNegativeReals,
-            units=units_meta("mass"),
-            doc="Mass of fresh GAC in the adsorber bed",
+            units=units_meta("length") ** 3,
+            doc="bed volume",
         )
-
-        self.velocity_sup = Var(
-            initialize=1,
+        self.ebct = Var(
```
adam-a-a, line 430: not going to be picky with naming conventions in this PR, but it is generally preferable to be explicit and spell names out for better readability for users (albeit we might expect a user in this field of research to be familiar with ebct).

watertap/unit_models/gac.py
```diff
@@ -760,209 +708,303 @@ def eq_dg(b, t, j):
...
         @self.Constraint(
-            doc="Minimum fluid residence time in the adsorber bed to achieve a constant pattern solution"
+            doc="minimum fluid residence time in the bed to achieve a constant pattern solution"
         )
-        def eq_min_res_time_cps(b):
-            return b.min_res_time == b.bed_voidage * b.min_ebct
-
-        @self.Constraint(doc="Residence time")
-        def eq_res_time(b):
-            return b.res_time == b.bed_voidage * b.ebct
```
adam-a-a, line 790: Was this where you found your degenerate constraint?

watertap/unit_models/gac.py
```diff
@@ -760,209 +708,303 @@ def eq_dg(b, t, j):
...
             return b.min_ebct * (1 - b.bed_voidage) * b.kf == b.min_N_St * (
                 b.particle_dia / 2
             )
 
+        @self.Constraint(doc="throughput based on empirical 5-parameter regression")
+        def eq_throughput(b):
+            return b.throughput == b.b0 + b.b1 * (
+                b.conc_ratio_replace**b.b2
+            ) + b.b3 / (1.01 - (b.conc_ratio_replace**b.b4))
+
```
adam-a-a, line 779: Going through this quickly, but I see this constraint pop up twice, except one is applied to discretized elements and the other is this one, which probably works out. I haven't connected all the dots between constraints though. After doing this model overhaul, did you check for degeneracy again after solving successfully?

watertap/unit_models/tests/test_gac.py
```diff
@@ -52,11 +53,16 @@
...
 from watertap.costing import WaterTAPCosting
 
-
 __author__ = "Hunter Barber"
 
 solver = get_solver()
 
+# inputs for badly_scaled_var_generator used across test frames
+sv_large = 1e2
+sv_small = 1e-2
```
adam-a-a, line 62: Seems more stringent than necessary, but that's ok. Did you have to be this stringent for things to work?

watertap/unit_models/tests/test_gac.py
```diff
@@ -52,11 +53,16 @@
...
+sv_zero = 1e-8
+
+print(isinstance({"species": "TCE"}, list))
```
adam-a-a: This could be that Github is hiding some code that I'm missing, but why is this here? Seems out of place.

watertap/unit_models/tests/test_gac.py
```diff
@@ -223,9 +226,19 @@ def test_simplified_solution(self, gac_frame_simplified):
 
         # Approx data pulled from graph in Hand, 1984 at ~30 days
         # 30 days adjusted to actual solution to account for web plot data extraction error within reason
-        assert (
-            pytest.approx(29.563, rel=1e-3) == value(ms.fs.unit.elap_time) / 24 / 3600
+        # values calculated by hand and match those reported in Hand, 1984
```
adam-a-a, line 227: no pun intended?

watertap/unit_models/tests/test_gac.py
```diff
@@ -242,12 +255,12 @@ def gac_frame_robust(self):
...
-        mr.fs.properties.dens_mass_const = 1000
-
+        mr.fs.properties.dens_mass_const = 999.7
         mr.fs.unit = GAC(
             property_package=mr.fs.properties,
-            film_transfer_coefficient_type="calculated",
-            surface_diffusion_coefficient_type="calculated",
+            film_transfer_coefficient_type="fixed",
+            surface_diffusion_coefficient_type="fixed",
+            finite_elements_ss_approximation=9,
```
adam-a-a, line 261: Did you notice any instability when setting number of elements too high/too low?

watertap/unit_models/tests/test_gac.py
```diff
@@ -671,63 +687,141 @@ def test_multi_solve(self, gac_frame_multi):
...
     def test_multi_solution(self, gac_frame_multi):
         mm = gac_frame_multi
 
-        # values calculated independently and near to those reported in Crittenden, 2012
-        assert pytest.approx(1.139, rel=1e-3) == value(mm.fs.unit.mass_throughput)
-        assert pytest.approx(12830000, rel=1e-3) == value(mm.fs.unit.elap_time)
-        assert pytest.approx(0.8333, rel=1e-3) == value(mm.fs.unit.bed_length)
-        assert pytest.approx(10.68, rel=1e-3) == value(mm.fs.unit.bed_area)
+        # only checking for variables new to configuration options
+        assert pytest.approx(2.473, rel=1e-3) == value(mm.fs.unit.N_Re)
```
adam-a-a, line 697: That's a very low value for Reynolds number... do we expect Re values to be that low in GAC?

adam-a-a: Just scanned an excerpt from [Crittenden et al., 1987](https://www.jstor.org/stable/41290795), that mentions a range of 0.08 to 150, so I guess the answer could be yes. Notably that range is specific to the correlation used for film transfer coefficient in that paper.

docs/technical_reference/unit_models/gac.rst
```diff
@@ -324,5 +332,8 @@ Journal of Environmental Engineering, 110(2), 440-456.
 Crittenden, J., Rhodes, R., Hand, D., Howe, K., & Tchobanoglous, G. (2012). MWHs Water Treatment. Principles and Design.
 EditorialJohn Wiley & Sons.
```
adam-a-a:
```suggestion
John Wiley & Sons.
```

docs/technical_reference/unit_models/gac.rst
```diff
@@ -314,7 +322,7 @@ Code Documentation
 -------------------
 
 * :mod:`watertap.unit_models.gac`
-* :meth:`watertap.costing.watertap_costing_package.WaterTAPCostingData.cost_gac`
+* :meth:`watertap.costing.watertap_costing_package.units.gac`
```
adam-a-a, line 325: You're probably aware, but this link to the costing method isn't clickable -probably because we don't have documentation for the costing package yet. That's fine for now- just pointing it out just in case you weren't aware.

watertap/unit_models/gac.py
```diff
@@ -760,209 +708,303 @@ def eq_dg(b, t, j):
...
+                doc="shape correction factor",
             )
 
-            @self.Constraint(doc="Reynolds number calculation")
-            def eq_reynolds_number(b):
+            @self.Constraint(
+                self.flowsheet().config.time,
+                doc="Reynolds number calculation",
+            )
+            def eq_reynolds_number(b, t):
```
adam-a-a, line 978: I noticed a slight variation for calculating Re number in another paper by Crittenden et al., 1987. 
image
I fairly certain they are equivalent but just want to make sure. Your formulation includes particle diameter instead of 2*R, and you don't divide by bed voidage since that is accounted for by interstitial velocity (superficial velocity divided by bed voidage). Is that correct?

adam-a-a (approved):
Great job! I had a few comments and questions for you to review, but I'm also approving.

watertap/unit_models/gac.py
hunterbarber: This is just for ease of fixing diameter as a variable when matching data to a single column's performance just because bed length and diameter are typically given, used in _most_ validation. This clarifier is here just because if considering multiple in parallel, the **volume** variable equals the sum of the volumes for the beds operating in parallel, so diameter then isn't directly relevant. This diameter variable isn't touched with respect to any dimensionless parameters or performance, just for the single column dimension insight. This reasoning is touched on in the documentation as well.

watertap/unit_models/gac.py
hunterbarber: There were two issues. One was effectively a circular reference in relating length. volume, cross sectional area, volumetric flowrate, velocity, etc. in a set of equations. The other was an isolated constraint involving the solid mass density of the GAC. Although the DOF would be zero, when providing logical variables to be fixed, the isolated constraint would have no response/ infinite solutions in the optimizer due to 2 unknown, 1 isolated constraint.

watertap/unit_models/tests/test_gac.py
hunterbarber: Didn't even notice haha

watertap/unit_models/tests/test_gac.py
hunterbarber: 9 was selected here because it gave increments closest to 0.1 which is what I had the exact solutions to in an offline spreadsheet to validate the results,. It is stable throughout, just adds 4 variables every 1 more element. The default is set to lower because lower tends to be a good enough approximation.

watertap/unit_models/tests/test_gac.py
hunterbarber: Up to 25 minute EBCT, definitely expecting laminar and low Re. Also only constructed when using that correlation, and not used in the performance equations.

docs/technical_reference/unit_models/gac.rst
hunterbarber: fixed in #960 when I noticed it ☺, just waiting for this or that to merge so I can fix the conflicts in the doc file

watertap/unit_models/gac.py
hunterbarber: R is adsorbent radius in that so particle diameter is equivalent.

Reynolds number for packed beds is inconsistent across reference usage. Basically some multiply by particle sphericity as a correction factor, divide by bed voidage, or divide by (1-bed voidage). All with relatively comparable results. Here, I used the expression defined in Crittenden, 2012 (same as the one you show) since the majority of the CPHSDM is referenced from there as well. And yes the velocity relation is correct, so it is superficial velocity divided by bed voidage here.

watertap/unit_models/tests/test_gac.py
hunterbarber: Nope, this is an accident. Removing it.

watertap/unit_models/tests/test_gac.py
hunterbarber: The model works fine under the defaults! Something I changed, in addition to having dynamic scaling factors for variables as a function of the default sf for `H2O` and `target_species` molar flow from the property package, dynamic scaling factors according to inlet volumetric flow because it was more convenient for variables such as bed volume. So I used tight bounds when testing these scaling factors just because I wanted to make sure I was actually doing it right and the default of 1e4 (or whatever the default is) is fine for edge cases, but I wanted to be more nearer to 1 for a traditional process scale such as this.

watertap/unit_models/gac.py
hunterbarber: Good news is I just double checked and DegeneracyHunter says it is in the clear, but there is a more efficient way to implement this, though slightly challenging.

So in the discretization of this equation, this is basically the end point of the steps. So this solution is used to determine the step size given a fixed number of discretized elements. I played around with changing the indices (which are already a little convoluted) and declaring a rule that fixing the matching element_[n] case to the original case, but it turned out to be more difficult than expected. If I remember correctly, it would be solving the performance equations for elements i throguh (n-1) but solve the linspace utility for element n because of how the variables would be fixed. I think it's worth revisiting in subsequent PR's just because the amount of changes in this. Let me know your thoughts.

watertap/unit_models/gac.py
hunterbarber: I used this sparingly for EBCT and SPDFR just because the black formatting when using min_empty_bed_contact_time and surface_to_pore_diffusion_flux_ratio gets pretty illegible because of the length. I think for EBCT, it's such a fundamental concept in the field, like having to write out plug flow reactor, that I would think it's fine. I'm not opposed to changing it to maintain a convention though.

hunterbarber:
Note, changes to watertap/core/initialization_mixin.py pushed to main causes 1 test to fail.

watertap/unit_models/gac.py
```diff
@@ -760,243 +708,347 @@ def eq_dg(b, t, j):
...
-            # TODO check N_Re formulation for packed beds
+
             self.N_Re = Var(
-                initialize=1,
+                initialize=10,
                 bounds=(0, None),
                 domain=NonNegativeReals,
                 units=pyunits.dimensionless,
-                doc="Reynolds number, correlation using Schmidt number valid in Re < 2e4",
+                doc="Reynolds number, correlations using Reynolds number valid in Re < 2e4",
```
kurbansitterley, line 957: Any reason to set the bounds for ``N_Re`` (and other dimensionless variables) to the validated ranges? I have done this in IX to make sure it stays in the assumptions for the model, but I'm not sure if its the best approach. I won't let it hold up the PR.

kurbansitterley (approved):
LGTM- I had one question on dimensionless variable bounds but won't let it hold this up.

watertap/unit_models/gac.py
hunterbarber: @kurbansitterley This is the same logic as yours, basically N_Re should be laminar for the correlations to be valid. I see it from this perspective though. If a user is inputting variables into the model for a design and flow becomes turbulent one of two things will happen:

- **N_Re is not bounded:** the model will solve but N_Re will be out of the assumptions range, it's on the user to identify the poor conditions
- **N_Re is bounded:** the model will not solve due to a variable at bounds error, but requiring debugging techniques to find this out

This may be troublesome and more tricky to debug when the model isn't solving because it's ambiguous if the error is with N_Re hitting a bound or something else with the solve. Could use more input here though.

andrewlee94: I personally tend to use the fitted range for the bounds relating to correlations, as this is the range over which we know the correlation is valid. If someone tries to exceed the bound, then there is no guarantee the result is correct even if the solver finds an "optimal" solution, so I tend to err towards a definitive failure rather than providing a potentially invalid answer. Users can always relax bounds later (at their own risk).

IDAES is trying to work on better diagnostics tools this year which would help users identify the source of an infeasible model.

PR #960, opened 2023-03-13
Adding costing options for pressure or gravity GAC contactors

### LINKED ISSUE #886: Need for GAC costing based on subclasses
### Description

Add multiple costing correlations, or simply multi contactor cost correlations, for GAC unit models to reflect differences in design configurations.

### Motivation

Need to review the magnitude and trend of cost differences in subclasses before moving forward with any changes.

### Possible Implementation

_No response_

### Additional Context

Currently only costed as pressure contactors, which is likely the higher cost of the design subclasses and would slightly overestimate costs of other types. Designs are typically made based on system size and land use limitations, so may be valuable to have discretely different costing methods, but current is likely accurate enough for slight error if used appropriately.

hunterbarber:
## Fixes/Resolves:
Directly addressed issue #886

## Summary/Motivation:
Distinction is relevant at small vs large scale systems

- [x] Add costing options for gravity basins into watertap_costing_package
- [x] Re-add costing option for gravity basins into ZO_costing
- [x] Add approximate energy consumptions now relevant with gravity basins
- [x] Add tests
- [x] Change docs

### Legal Acknowledgement


watertap/examples/flowsheets/case_studies/municipal_treatment/tests/test_municipal_treatment.py
```diff
@@ -63,7 +63,7 @@ def test_municipal_treatment():
         m.fs.recharge_pump.properties[0].flow_mass_comp["tss"]
     ) == pytest.approx(1.656e-6, rel=1e-3)
 
-    assert value(m.fs.costing.LCOW) == pytest.approx(5.0367e-7, rel=1e-3)  # in M$/m**3
+    assert value(m.fs.costing.LCOW) == pytest.approx(4.9602e-7, rel=1e-3)  # in M$/m**3
```
hunterbarber, line 66: I confirmed new fs.gac cost matches offline calculations

docs/technical_reference/unit_models/gac.rst
```diff
@@ -235,26 +235,43 @@ The following parameters are constructed when applying the GAC costing method in
...
+are regressed from higher detailed costing methods in (U.S. EPA, 2021).
 
 Costing GAC contactors is defaulted to purchasing 1 operational and 1 redundant contactor for alternating operation. For large systems this may be a poor
 assumption considering vessel sizing and achieving pseudo-steady state.  The number of contactors input by the user should justify reasonable
 (commercially available) dimensions of identical modular contactors in parallel. When costing several operational vessels, the area reported
-in the unit model should be interpreted as the sum of the areas across all operating GAC contactors.
+in the unit model should be interpreted as the sum of the areas across all operating GAC contactors. The costing
+parameters may be selected from either steel pressure-fed vessels or concrete gravity-fed basins by the
+``surface_diffusion_coefficient_type`` argument. Note this only effects costing calculations are volume dimensions
+within the model are only calculated assuming a cylindrical bed.
```
MarcusHolly:
```suggestion
``surface_diffusion_coefficient_type`` argument. Note this only affects costing calculations and volume dimensions
within the model are only calculated assuming a cylindrical bed.
```
Everything LGTM, just a small grammatical issue here

MarcusHolly (approved):
LGTM

watertap/costing/units/gac.py
```diff
@@ -19,106 +21,162 @@
...
+    blk.grav_energy_coeff_1 = pyo.Var(
+        initialize=0.132403,
+        units=pyo.units.kW * (pyo.units.m**3) ** -1,
+        doc="GAC concrete gravity contactor polynomial energy consumption coefficient 1",
+    )
+    blk.grav_energy_coeff_2 = pyo.Var(
+        initialize=-1.41512e-5,
+        units=pyo.units.kW * (pyo.units.m**3) ** -2,
+        doc="GAC concrete gravity contactor polynomial energy consumption coefficient 2",
+    )
```
adam-a-a, line 162: Are we fixing all of these coefficients somewhere? Also, it looks like regardless of the contactor type specified, all of the cost variables will be created for both pressure and gravity-fed contactors. Is that right?

adam-a-a (commented):
One thing that isn't clear to me is why all the cost coefficients are specified as variables but are not fixed. Do you check the degrees of freedom before initializing?

adam-a-a:
> One thing that isn't clear to me is why all the cost coefficients are specified as variables but are not fixed. Do you check the degrees of freedom before initializing?

Just realized that the variables get fixed in `register_costing_parameter_block`. Even so, I figure you should try initializing in your tests and checking DOF = 0 before initialization.

hunterbarber:
> > One thing that isn't clear to me is why all the cost coefficients are specified as variables but are not fixed. Do you check the degrees of freedom before initializing?
> 
> Just realized that the variables get fixed in `register_costing_parameter_block`. Even so, I figure you should try initializing in your tests and checking DOF = 0 before initialization.

Addressed in [842cc77](https://github.com/watertap-org/watertap/pull/960/commits/842cc77631004b5c02960f95246f7bdd4e11e849). Though an interesting thing I found is that you can't run `initialization_tester` twice on the same frame. It creates a dummy constraint that cannot be written a second time due to it already existing.

andrewlee94:
@hunterbarber That sounds like it might be a bug - the initialization tester should be cleaning those dummy constraints up. If it isn't, could you please post an issue on IDAES with a minimal example replicating the issue (or better yet a PR with a fix).

hunterbarber:
> @hunterbarber That sounds like it might be a bug - the initialization tester should be cleaning those dummy constraints up. If it isn't, could you please post an issue on IDAES with a minimal example replicating the issue (or better yet a PR with a fix).

@andrewlee94 added in [idaes-pse/issues/1141](https://github.com/IDAES/idaes-pse/issues/1141), let me know if you need more details.

andrewlee94:
@hunterbarber I think I see the issue - it is not the dummy constraint itself that is the issue, but an implicit indexing set that is created along with it (and thus not cleaned up as it was created implicitly).

hunterbarber:
> @hunterbarber I think I see the issue - it is not the dummy constraint itself that is the issue, but an implicit indexing set that is created along with it (and thus not cleaned up as it was created implicitly).

@andrewlee94 That is an error in the function then correct, not the usage? I had thought I had posted the error the first time I got it, but I lost my draft jumping between workspaces and forgot the exact wording in the error. So, this was just replicating it quickly with what's written in the issue.

andrewlee94:
@hunterbarber Yes, it is an error in the function - the traceback was enough to figure out what the problem was.

andrewlee94:
@hunterbarber I just opened a PR with the fix: https://github.com/IDAES/idaes-pse/pull/1142.

adam-a-a (approved):
LGTM- @hunterbarber can you make an issue to remind us of the cost params for two contactor types being constructed? We can later decide if we want to change or leave it alone.

hunterbarber:
@adam-a-a issue created in #975 to discuss revision

PR #966, opened 2023-03-16
Added functionality to unfix input variables

### LINKED ISSUE #28: Consider a DOF display that updates on input tab
I think it would be nice to have the degrees of freedom (DOF) displayed on the input tab. Later, if we enable users to fix/unfix certain variables of their choosing, the "Degrees of Freedom" entry would display the latest value of DOF as the user toggles "fixed" on or off for particular vars.

MichaelPesce:
## Fixes/Resolves:
Previously, all input variables could only be fixed to one value.

## Summary/Motivation:
Allow unfixing of variables

- Add attributes to flowsheet interface model object: fixed, lb, ub, dof
- Update load function to ensure these additional attributes get updated when an update occurs

### Legal Acknowledgement


watertap/ui/fsapi.py
```diff
@@ -508,6 +540,21 @@ def export_values(self):
...
+                    tmp = pyo.Var(initialize=mo.obj.ub, units=u.get_units(mo.obj))
+                    tmp.construct()
+                    mo.ub = pyo.value(u.convert(tmp, to_units=mo.ui_units))
+                if mo.obj.lb is None:
+                    mo.lb = mo.obj.lb
+                else:
+                    tmp = pyo.Var(initialize=mo.obj.lb, units=u.get_units(mo.obj))
+                    tmp.construct()
+                    mo.lb = pyo.value(u.convert(tmp, to_units=mo.ui_units))
+            except Exception as e:
```
lbianchi-lbl: I'd suggest to use any (and ideally all) of:

- A more specific exception type
- A comment explaining under which circumstances exceptions can be raised
- Depending on whether the exceptions are rare or part of the ordinary control flow, a logging message of the appropriate level

lbianchi-lbl (commented):
I've found one generic silent `except` that could use some clarification, but other than that it looks good.

adam-a-a (approved):
LGTM - I'm not sure if we'd need any additional testing, but it looks like nothing is breaking yet.

watertap/ui/fsapi.py
MichaelPesce: That `Except` was basically operating as an `if` statement (really the `else` of the `if` statement), as it was getting triggered when the Pyomo object was an expression, because expressions don't have bounds. I just pushed a change to replace that `try/except` with an `if` statement, which checks if the object is an instance of any of the Pyomo Expression classes before trying to set the bounds.

lbianchi-lbl (approved):

PR #977, opened 2023-03-24
Electrocoagulation_ZO: Automatic rst generation 

adam-a-a:
## Fixes/Resolves:
NA
## Summary/Motivation:
Some initial documentation was manually written for the `electrocoagulation_zo` model in PR #953. However, all the ZO documentation was previously generated automatically by a script that I wrote, and running it again once a new ZO model is added can generate a more extensive rst file that lists out additional constraints and variables tacked on to the model (additional with respect to the vars/constraints created when using the "helper functions", e.g., `build_sido`).

**NOTE: running the automation script also made adjustments to certain rst files where doc strings were changed in the respective unit models. Additionally, `hr` was replaced with `h` for some reason.** 

## Changes proposed in this PR:
- main focus: generate more extensive rst file for electrocoagulation using the automation script workflow
- add entry for electrocoagulation to the table in `docs\technical_reference\unit_models\zero_order_unit_models\WT3_unit_classification_for_doc.xlsx`
- automation script builds rst files based on entries in the aformentioned xlsx file
- miscellaneous changes to other rst files that happened automatically, but shouldn't be focused on too much

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #978, opened 2023-03-24
Remove redundant constant_intensity function call

adam-a-a:
## Fixes/Resolves:
NA
## Summary/Motivation:
While working on PR #977, I noticed that `clarifier_zo` was giving me a pyomo warning indicating a modeling error, attributed to creating the same variables more than once. This PR removes a duplicate function call that ultimately builds the same variables twice.

- remove extra `constant_intensity()` call from `clarifier_zo`

### Legal Acknowledgement


watertap/unit_models/zero_order/clarifier_zo.py
```diff
@@ -38,7 +38,6 @@ def build(self):
         self._tech_type = "clarifier"
 
         build_sido(self)
-        constant_intensity(self)
```
MarcusHolly, line 41: My guess is that you should've kept this line and deleted line 46 instead.

tarnold17 (approved):

watertap/unit_models/zero_order/clarifier_zo.py
adam-a-a: I think you're right.

MarcusHolly (approved):

PR #982, opened 2023-03-27
BSM2 initial model

agarciadiego:
## Summary/Motivation:

Adds BSM2 initial model, and plotting capabilities

-
-

### Legal Acknowledgement


watertap/examples/flowsheets/case_studies/activated_sludge/ASM1_flowsheet.py
```diff
@@ -115,7 +115,7 @@ def build_flowsheet():
     m.fs.MX6 = Mixer(property_package=m.fs.props, inlet_list=["clarifier", "reactor"])
     # Product Blocks
     m.fs.Treated = Product(property_package=m.fs.props)
-    m.fs.Sludge = Product(property_package=m.fs.props)
```
adam-a-a, line 118: I suppose you removed this since the sludge will be the AD inlet in our subset flowsheet. We could later consider having the option of adding this Product block for sludge for someone interested in analyzing the AS process in isolation, or skipping this in the case where we extend to the "full" flowsheet.

watertap/property_models/anaerobic_digestion/adm1_reactions.py
```diff
@@ -1216,7 +1216,7 @@ def _rxn_rate(self):
         self.reaction_rate = pyo.Var(
             self.params.rate_reaction_idx,
             initialize=self.rates,
-            bounds=(1e-9, 1e-4),
+            bounds=(1e-10, 1e-4),
```
adam-a-a, line 1219: I might suggest we modify these bounds similar to what I did in the PR on modified ADM1 (#948)

watertap/property_models/anaerobic_digestion/adm1_reactions.py
```diff
@@ -1325,7 +1325,7 @@ def Dissociation_rule(self, t):
         # Equation from [2]
         def CO2_acid_base_equilibrium_rule(self, t):
             return (
-                self.K_a_co2
+                self.K_a_co2 * 1e-2
```
adam-a-a, line 1328: Why add this factor? Is it a correction for something overlooked? If so, we would need to consider it in PR #948.

adam-a-a: nevermind-- after looking a few lines down, I saw that you apply this on the right hand-side too. I do believe I added some scaling for this constraint in #948 as it was one of the problematic ones with regard to scaling.

watertap/unit_models/anaerobic_digestor.py
```diff
@@ -884,11 +884,11 @@ def initialize_build(
...
-        if degrees_of_freedom(self) != 0:
-            raise InitializationError(
-                f"{self.name} degrees of freedom were not 0 at the beginning "
-                f"of initialization. DoF = {degrees_of_freedom(self)}"
-            )
+        # if degrees_of_freedom(self) != 0:
+        #     raise InitializationError(
+        #         f"{self.name} degrees of freedom were not 0 at the beginning "
+        #         f"of initialization. DoF = {degrees_of_freedom(self)}"
+        #     )
```
adam-a-a, line 916: Similarly, I made tweaks in #948 that should be useful for the AD unit initialization (just a start though). Furthermore, we can add a try/except to the sensitivity run so that we simply allow the solve to continue even after a failed init,

watertap/examples/flowsheets/case_studies/activated_sludge/ASM1_flowsheet.py
agarciadiego: I understand the idea when making this flowsheets "blocks" to be added . Maybe the deletion of the product stream should be done on the full flowsheet instead of the ASM flowsheet. And in the case of the full BSM2 flowsheet this product stream should be added after the dewatering unit model in the BSM2.

watertap/property_models/anaerobic_digestion/adm1_reactions.py
agarciadiego: Agree, however I think we should take a closer look after the thickener unit model is added to see the effects on concentration to understand how the bounds would look like

watertap/property_models/anaerobic_digestion/adm1_reactions.py
agarciadiego: agree, I was trying out changing the scaling here.

watertap/unit_models/anaerobic_digestor.py
agarciadiego: Yes, the main issue here is that for the initialization procedure I used the model has 0 DOF, even thought the reactor still has some DOF. And this if statement makes the flowsheet fail. I was considering adding an option in the configuration to accommodate for both

adam-a-a:
@agarciadiego I think we can close this PR as it would still be accessible to anyone to review for whatever reason. I initially wanted it up so that the whole team could see and collaborate on the previous quarter's analysis milestone. Do you agree on closing?

agarciadiego:
Close with #1038

PR #986, opened 2023-03-30
Initial implementation of parallel manager for MPI and single process…

shelman:
… runner

## Fixes/Resolves:

Initial implementation of ParallelManager, with working code for MPI and "Dummy MPI" when running under a single process. In a following PR I'll plan to remove MPI from the parameter sweep code entirely.

This doesn't contain tests yet, as we discussed. If the API looks good to you then I can add unit tests and update the PR before merging it in.

### Legal Acknowledgement

### DIFF
--- a/watertap/tools/parallel/parallel_manager.py
+++ b/watertap/tools/parallel/parallel_manager.py
@@ -21,19 +21,17 @@ class ParallelManager(ABC):
         if parallel_manager_class is not None:
             return parallel_manager_class()
 
-        if mpi4py_available and ParallelManager.is_running_under_mpi():
+        if mpi4py_available and ParallelManager.has_mpi_peer_processes():
             return MPIParallelManager(mpi4py)
 
         return SingleProcessParallelManager()
 
     @classmethod
-    def is_running_under_mpi(cls):
+    def has_mpi_peer_processes(cls):
         """
-        Returns whether the process was run as part of an MPI group.
+        Returns whether the process was run as part of an MPI group with > 1 processes.
         """
-        # see https://www.open-mpi.org/faq/?category=running#mpi-environmental-variables.
-        # this is one of the environment variables that MPI sets.
-        return os.getenv("OMPI_COMM_WORLD_SIZE") is not None
+        return mpi4py_available and mpi4py.MPI.COMM_WORLD.Get_size() > 1
 
     @abstractmethod
     def is_root_process(self):
--- a/watertap/tools/parallel/parallel_manager.py
+++ b/watertap/tools/parallel/parallel_manager.py
@@ -5,99 +5,92 @@ import numpy
 
 from pyomo.common.dependencies import attempt_import
 
+mpi4py, mpi4py_available = attempt_import("mpi4py")
+
 
 class ParallelManager(ABC):
-    """
-    Create and return an instance of a ParallelManager, based on the libraries available in the
-    runtime environment.
-
-    Allows an optional python class to be passed in as parallel_manager_class. If so, this class
-    is instantiated and returned rather than checking the local environment.
-    """
-
     @classmethod
     def create_parallel_manager(cls, parallel_manager_class=None):
+        """
+        Create and return an instance of a ParallelManager, based on the libraries available in the
+        runtime environment.
+
+        Allows an optional python class to be passed in as parallel_manager_class. If so, this class
+        is instantiated and returned rather than checking the local environment.
+        """
         if parallel_manager_class is not None:
             return parallel_manager_class()
 
-        mpi4py, mpi4py_available = attempt_import("mpi4py")
         if mpi4py_available and ParallelManager.is_running_under_mpi():
             return MPIParallelManager(mpi4py)
 
         return SingleProcessParallelManager()
 
-    """
-    Returns whether the process was run as part of an MPI group.
-    """
-
     @classmethod
     def is_running_under_mpi(cls):
+        """
+        Returns whether the process was run as part of an MPI group.
+        """
         # see https://www.open-mpi.org/faq/?category=running#mpi-environmental-variables.
         # this is one of the environment variables that MPI sets.
         return os.getenv("OMPI_COMM_WORLD_SIZE") is not None
 
-    """
-    Return whether the current process should be considered as the root of the parallel group it
-    is a part of.
-    """
-
     @abstractmethod
     def is_root_process(self):
+        """
+        Return whether the current process should be considered as the root of the parallel group it
+        is a part of.
+        """
         raise NotImplementedError
-
-    """
-    Synchronize a piece of data with all peer processes. The data parameter is either:
-    - (if root) the source of truth that will be broadcast to all processes
-    - (if not root) an empty buffer that will hold the data sent from root once sync_data returns
-    """
 
     @abstractmethod
     def sync_data(self, data):
+        """
+        Synchronize a piece of data with all peer processes. The data parameter is either:
+        - (if root) the source of truth that will be broadcast to all processes
+        - (if not root) an empty buffer that will hold the data sent from root once sync_data returns
+        """
         raise NotImplementedError
-
-    """
-    Scatter a function out to a set of child processes, to be run
-    for a list of parameters.
-    - all_parameter_combos is a list, where each item represents the
-    parameters for a single run.
-    - fn is a function to be run, and should accept a single parameter. 
-    It will be called for each item in the all_parameter_combos list.
-    """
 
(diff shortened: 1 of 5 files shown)


watertap/tools/parallel/parallel_manager.py
```diff
@@ -0,0 +1,155 @@
...
+    Allows an optional python class to be passed in as parallel_manager_class. If so, this class
+    is instantiated and returned rather than checking the local environment.
+    """
+
+    @classmethod
+    def create_parallel_manager(cls, parallel_manager_class=None):
+        if parallel_manager_class is not None:
+            return parallel_manager_class()
+
+        mpi4py, mpi4py_available = attempt_import("mpi4py")
```
bknueven: There's no harm in putting this at the top of the file. The pyomo `attempt_import` function is pretty slick and will delay the import anyways.

watertap/tools/parallel/parallel_manager.py
```diff
@@ -0,0 +1,155 @@
...
+        return SingleProcessParallelManager()
+
+    """
+    Returns whether the process was run as part of an MPI group.
+    """
+
+    @classmethod
+    def is_running_under_mpi(cls):
+        # see https://www.open-mpi.org/faq/?category=running#mpi-environmental-variables.
+        # this is one of the environment variables that MPI sets.
```
bknueven: Can you put the docstring under the method? It's more consistent with our style.
```suggestion
    @classmethod
    def is_running_under_mpi(cls):
        """
        Returns whether the process was run as part of an MPI group.
        """
        # see https://www.open-mpi.org/faq/?category=running#mpi-environmental-variables.
        # this is one of the environment variables that MPI sets.
```

bknueven: Similar throughout.

watertap/tools/parallel/parallel_manager.py
```diff
@@ -0,0 +1,155 @@
...
+
+"""
+Class representing the results of one process's run of an optimization routine.
+"""
+
+
+class LocalResults:
+    def __init__(self, parameters, results):
+        self.parameters = parameters
+        self.results = results
```
bknueven: Should this and `GatherResults` just be a data class?   https://docs.python.org/3/library/dataclasses.html

watertap/tools/parameter_sweep/parameter_sweep.py
```diff
@@ -574,26 +597,76 @@ class ParameterSweep(_ParameterSweepBase):
...
+                combined_outputs[i] = np.append(combined_outputs[i], output["value"][i])
+
+        return np.asarray(combined_outputs)
 
     def parameter_sweep(
         self,
         model,
         sweep_params,
-        outputs=None,
+        combined_outputs=None,
```
bknueven, line 716: You might need to carry this change through in other places in watertap...

bknueven (commented):
Looks pretty good -- there's a `pylint` error to track down.

Main comment I have is about putting the docstrings immediately under the method/class they're documenting. Everything else is optional.

ksbeattie:
@shelman, @bknueven any news on this?  Still possible for the June release?

watertap/tools/parallel/parallel_manager.py
shelman: I think since data classes use type annotations and we don't have those in watertap (yet?) it might throw people off. I can revisit if these classes get more complex though.

bknueven: That's fine, this is a soft suggestion.

To you other point, I would happily accept PRs with type annotations. Obviously not required, though perhaps the larger team should have that discussion.

watertap/tools/parameter_sweep/parameter_sweep.py
shelman: I'll take a look and make sure anything that needs the update has it.

watertap/tools/parameter_sweep/parameter_sweep.py
```diff
@@ -192,74 +197,94 @@ def _publish_updates(self, iteration, solve_status, solve_time):
...
+        # over from the root process.
+        if self.parallel_manager.is_root_process():
+            global_combo_array = self._create_global_combo_array(d, sampling_type)
+        else:
+            global_combo_array = self._create_empty_global_combo_array(
+                d, sampling_type, num_samples
+            )
+
+        # make sure all processes running in parallel have an identical copy of the data
+        self.parallel_manager.sync_data(global_combo_array)
```
k1nshuk, line 287: Low priority, but we should change `sync_data` to `broadcast_data` or something along those line.

watertap/tools/parallel/parallel_manager.py
```diff
@@ -0,0 +1,150 @@
...
+        return SingleProcessParallelManager()
+
+    @classmethod
+    def is_running_under_mpi(cls):
+        """
+        Returns whether the process was run as part of an MPI group.
+        """
+        # see https://www.open-mpi.org/faq/?category=running#mpi-environmental-variables.
+        # this is one of the environment variables that MPI sets.
+        return os.getenv("OMPI_COMM_WORLD_SIZE") is not None
```
bknueven: This will work if we are using OpenMPI, but maybe not MPICH. I will look into this.

bknueven:
```suggestion
        from mpi4py import MPI
        return MPI.COMM_WORLD.Get_size() > 1
```
This is not MPI-implementation dependent, and if we're just running in MPI with a single rank we can just use the `SingleProcessParallelManager` anyways.

bknueven (changes requested):
Please change the one small thing regarding MPI detection. Otherwise it looks good!

watertap/tools/parallel/parallel_manager.py
shelman: Updated in the latest commit! Good catch.

watertap/tools/parallel/parallel_manager.py
```diff
@@ -0,0 +1,148 @@
...
+        Create and return an instance of a ParallelManager, based on the libraries available in the
+        runtime environment.
+
+        Allows an optional python class to be passed in as parallel_manager_class. If so, this class
+        is instantiated and returned rather than checking the local environment.
+        """
+        if parallel_manager_class is not None:
+            return parallel_manager_class()
+
+        if mpi4py_available and ParallelManager.has_mpi_peer_processes():
```
bknueven:
```suggestion
        if ParallelManager.has_mpi_peer_processes():
```
I guess we needn't check twice.

bknueven (approved):

watertap/tools/parameter_sweep/parameter_sweep_differential.py
```diff
@@ -246,13 +247,14 @@ def _run_differential_sweep(self, model, local_value):
             reinitialize_function=self.config.reinitialize_function,
             reinitialize_kwargs=self.config.reinitialize_kwargs,
             reinitialize_before_sweep=self.config.reinitialize_before_sweep,
+            parallel_manager_class=SingleProcessParallelManager,
             comm=DummyCOMM,
```
k1nshuk, line 251: Is the `DummyCOMM` still needed given that SingleProcessParallelManager exists?

k1nshuk (changes requested):

watertap/tools/parameter_sweep/parameter_sweep_differential.py
shelman: It is for the moment, unfortunately - since not all explicit uses of the `comm` object are pulled out of parameter sweep. We'll be able to remove it once I get rid of the other uses.

watertap/tools/parameter_sweep/parameter_sweep.py
shelman: Should I change for this PR? Or should we go back over the API later and adjust naming?

k1nshuk (approved):

watertap/tools/parameter_sweep/parameter_sweep.py
k1nshuk: This can happen in a subsequent PR.

watertap/tools/parameter_sweep/parameter_sweep_differential.py
k1nshuk: @shelman Do you intend to make this change in the PR or a subsequent one?

shelman: A subsequent one!

PR #992, opened 2023-03-31
Move seawater ion parameter block out of examples

wigging:
## Fixes/Resolves:

The seawater ion parameter block (`seawater_ion_prop_pack.py`) that is located in the examples directory is being used for tests within the watertap package. This makes the examples a dependency of the watertap package. To remove this dependency, the seawater ion parameter block was moved to the `watertap.property_models` sub package.

## Summary/Motivation:

Examples for a Python package should be independent of that package. Tests and source code for that package should not rely on example code. If the examples for a package are deleted, the package and its tests should still work. Ideally, examples should not be inside the source directory of a package but I'll fix that later.

- Remove `seawater_ion_prop_pack.py` from examples directory
- Add `seawater_ion_prop_pack.py` to `watertap.property_models`
- Fix tests that are affected by the above changes

### Legal Acknowledgement


TimBartholomew:
I think its a good idea to move this property package out of the examples because it is being used in tests for unit models.

TimBartholomew:
However, what this points to is that we need a generic/flexible simple property package that has mass-based state variables. Similar to the selective oil permeation property package (although that has its differences in having separate properties for oil and water like density and viscosity, which shouldn't be include in a generic/flexible prop pack).

Options include 1) making a mass based property package like the multicomponent aqueous solution, or 2) making the multicomponent aqueous solution support mass based state variables.

adam-a-a (approved):
LGTM

srikanthallu (approved):

PR #995, opened 2023-03-31
Remove more unused imports

### LINKED ISSUE #994: Remove more unused imports
### Description

Using `flake8 --statistics --select=F401 --exclude=__init__.py .` gives 223 warnings about unused imports in watertap. I'll go through the package and remove these unused imports.

### Expected Behavior

Expect Python files to only import what is needed for the code in that file.

### Steps to Reproduce

Use the flake8 tool from within the watertap directory:

```
cd watertap
flake8 --statistics --select=F401 --exclude=__init__.py .
```

### Environment

- OS: macOS 12.6.3
- Python: 3.10
- WaterTAP: latest version

### Anything Else?

_No response_

wigging:
## Fixes/Resolves:

Resolves issue #994.

## Summary/Motivation:

Remove unused imports throughout the watertap package. After this pull request is merged, the GitHub Actions should be updated to check future pull requests for unused imports.

- Remove unused imports

### Legal Acknowledgement

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,4 +16,5 @@ disable = [
 # for a list of all messages
 enable = [
     "unnecessary-pass",
+    "unused-import",
 ]
--- a/watertap/core/membrane_channel0d.py
+++ b/watertap/core/membrane_channel0d.py
@@ -28,6 +28,7 @@ import idaes.logger as idaeslog
 from watertap.core.membrane_channel_base import (
     MembraneChannelMixin,
     PressureChangeType,
+    CONFIG_Template,
 )
 
 
--- a/watertap/core/membrane_channel0d.py
+++ b/watertap/core/membrane_channel0d.py
@@ -28,7 +28,6 @@ import idaes.logger as idaeslog
 from watertap.core.membrane_channel_base import (
     MembraneChannelMixin,
     PressureChangeType,
-    CONFIG_Template,
 )
 
 
--- a/watertap/core/membrane_channel0d.py
+++ b/watertap/core/membrane_channel0d.py
@@ -28,7 +28,10 @@ import idaes.logger as idaeslog
 from watertap.core.membrane_channel_base import (
     MembraneChannelMixin,
     PressureChangeType,
+    CONFIG_Template as Base_CONFIG_Template,
 )
+
+CONFIG_Template = Base_CONFIG_Template()
 
 
 @declare_process_block_class("MembraneChannel0DBlock")
--- a/watertap/core/membrane_channel1d.py
+++ b/watertap/core/membrane_channel1d.py
@@ -10,13 +10,8 @@
 # "https://github.com/watertap-org/watertap/"
 #################################################################################
 
-from copy import deepcopy
-
 from pyomo.common.config import ConfigValue, In
-from pyomo.environ import (
-    Constraint,
-    Set,
-)
+from pyomo.environ import Set
 from idaes.core import (
     declare_process_block_class,
     DistributedVars,
--- a/watertap/core/tests/test_zero_order_costing.py
+++ b/watertap/core/tests/test_zero_order_costing.py
@@ -26,7 +26,7 @@ from pyomo.environ import (
     Var,
 )
 from pyomo.util.check_units import assert_units_consistent
-from pyomo.common.config import ConfigBlock, ConfigValue
+from pyomo.common.config import ConfigValue
 
 from idaes.core import FlowsheetBlock, declare_process_block_class
 from idaes.core.solvers import get_solver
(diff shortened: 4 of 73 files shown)


wigging:
The `CONFIG_Template` that is imported in `watertap/core/membrane_channel0d.py` is unused in that file. However, if `CONFIG_Template` is removed then it breaks everything in watertap. It appears to be imported in the `membrane_channel0d.py` and then imported from that file into other files. `CONFIG_Template` is created in the `membrane_channel_base.py` so why isn't it imported from that file throughout the package?

wigging:
The `ConcentrationPolarizationType` that is imported in `watertap/unit_models/reverse_osmosis_0D.py` is not used in that file. However, `ConcentrationPolarizationType` is imported from `watertap/unit_models/reverse_osmosis_0D.py` by other files. Why is this not imported from `watertap/core/membrane_channel_base.py` where it is originally defined?

adam-a-a:
> `ConcentrationPolarizationType`
This was probably an oversight from when the RO 0D and 1D models were refactored with the new creation of the membrane_channel control volumes. I think you can alter the import to take straight from membrane_channel_base

adam-a-a:
> The `ConcentrationPolarizationType` that is imported in `watertap/unit_models/reverse_osmosis_0D.py` is not used in that file. However, `ConcentrationPolarizationType` is imported from `watertap/unit_models/reverse_osmosis_0D.py` by other files. Why is this not imported from `watertap/core/membrane_channel_base.py` where it is originally defined?

@wigging This was probably an oversight from when the RO 0D and 1D models were refactored with the new creation of the membrane_channel control volumes. I think you can alter the import to take straight from membrane_channel_base

bknueven:
> > The `ConcentrationPolarizationType` that is imported in `watertap/unit_models/reverse_osmosis_0D.py` is not used in that file. However, `ConcentrationPolarizationType` is imported from `watertap/unit_models/reverse_osmosis_0D.py` by other files. Why is this not imported from `watertap/core/membrane_channel_base.py` where it is originally defined?
> 
> @wigging This was probably an oversight from when the RO 0D and 1D models were refactored with the new creation of the membrane_channel control volumes. I think you can alter the import to take straight from membrane_channel_base

I would not be in favor of doing this -- I think the correct API has these things imported from the RO module. Users using the RO module don't need to know about `watertap.core.membrane_channel_base`.

wigging:
@bknueven The `ConcentrationPolarizationType` is imported from `watertap.core` not from `watertap.core.membrane_channel_base`. It is defined in `watertap.core.membrane_channel_base` but it is imported from core because of how it's handled in `__init__.py`.

adam-a-a:
Hmmm.. I see @bknueven 's point though. So in that case, it wasn't an oversight. For the user's sake, it would make more sense for a user to import `ConcentrationPolarizationType` from the RO unit (or at least have the option to).

bknueven:
> Hmmm.. I see @bknueven 's point though. So in that case, it wasn't an oversight. For the user's sake, it would make more sense for a user to import `ConcentrationPolarizationType` from the RO unit (or at least have the option to).

Yes, the current behavior is intentional.

All the "option types" for the RO unit should be imported from a single location, IMHO.

wigging:
@bknueven The enums `ConcentrationPolarizationType`, `MassTransferCoefficient`, `PressureChangeType`, and `FrictionFactor` are setup in watertap so they can be imported from `watertap.core`. It looks like this follows the IDAES approach where things like `MaterialBalanceType` and `FlowDirection` are imported from `idaes.core`.

bknueven:
Let's do the following
1. Move `watertap/unit_models/reverse_osmosis_*.py` to new directory `watertap/unit_models/reverse_osmosis/`
2. Put the enums for reverse osmosis into `watertap/unit_models/reverse_osmosis/__init__.py` along side the unit model classes.

Then the API will be reasonably consistent between our various unit models. I would also be happy to leave the unused imports alone in these modules -- they do serve a specific purpose.

wigging:
@bknueven I reverted the last two commits since the purpose of this pull request is to fix unused imports (not the API). To appease the linter and keep the original API, I added a few lines to the reverse osmosis unit models. See the latest commit. The use of the enum types and where they are located in the package should be handled better in watertap but I don't want to tackle that in this pull request.

adam-a-a (approved):
LGTM

wigging:
@lbianchi-lbl Is anything else needed for this pull request?

lbianchi-lbl:
> @lbianchi-lbl Is anything else needed for this pull request?

Nope, thanks for bringing this over the finish line!

lbianchi-lbl (approved):

Issue #934, opened 2023-02-23
Implementing Benchmark Simulation Model No. 2


adam-a-a:
### Description

Our goal is to create a flowsheet for the Benchmark Simulation Model No. 2 (BSM2) which is a standard used in the wastewater treatment industry. BSM2 represents a full wastewater treatment plant (WWTP) that includes the activated sludge process as well as an anaerobic digestor, along with other units. 

**The ultimate goal is to utilize this detailed, plant-scale model, and prioritized BSM2-related model enhancements that would be required, to facilitate detailed technology comparisons via cost optimization for innovations proposed by teams under the IEDO 2336 effort.**

### Motivation

This will provide a detailed baseline for wastewater treatment processes and could be used for cost optimization studies.

### Running TO-DO:
- [ ] #985 
- [ ] Create a modified ASM2D that would be compatible with precipitation PCM extension for ADM1
- [ ] #1318 
- [ ] Add TSS, COD, BOD5, Kjeldahl nitrogen to ADM1
- [ ] Add TSS, ISS, and/or VSS calculations directly to ASM/ADM models that are missing them and phase out ASM config option on thickener/dewaterer  (**started with modified ASM2D and modified ADM1 in #1078 but needs additions on ASM1, ASM2d and ADM1**)
- [ ] Review pH refinements in ASM and ADM models and incorporate into modified versions
- [ ] Secondary clarifier unit model - Takács model
- [ ] primary clarifier unit model - BSM implementation of model

### Questions , points of discussion, and minor issues
- [ ] Implement variation of ASM1 --> ADM1 translator block that includes **modified** Copp Interface
- [ ] storage tank unit model?
- [ ] Modify auxiliary units to be compatible with other ASM models #1078 - does primary clarifier in true implementation need this?

- [ ] Address simplification in modified ASM2d where S_IC does not account for carbonate and dissolved CO2 (Serralta 2004)
- [ ] make a baseline activated sludge reactor unit model?
- [ ] Track hydrogen sulfide (H2S) in ADM such that `Z_h2s` != 0?
- [ ] storage tank unit model?

### Tasks complete by Sept 2025
- [x] Delete simple modified ASM2d model? #1531 
- [x] Add TSS, COD, BOD5, Kjeldahl nitrogen to ASM2d #1492
- [x] Refinement of scaling in ASM2D #1533 
- [x] Track aggregate metrics throughout BSM2 flowsheet (pH, P components, N components) - see Solon 2017 #1244 

[...]

PR #997, opened 2023-04-03
Minor Refinements to ADM1

### LINKED ISSUE #941: Update `p_sat` name in ADM1 Properties Vapor to `pressure_sat`
Minor comment, but the standard name is `"pressure_sat"`

_Originally posted by @andrewlee94 in https://github.com/watertap-org/watertap/pull/937#discussion_r1123758808_

MarcusHolly:
## Fixes/Resolves:

Issue #934 

- `p_sat` has been updated to the standard name of `pressure_sat`

### Legal Acknowledgement


MarcusHolly:
@adam-a-a The `p_sat` -> `pressure_sat` issue might be more of a headache than it's worth. If you simply rename instances of `p_sat` to `pressure_sat`, you'll run into the following error:

 `idaes.core.util.exceptions.PropertyPackageError: A property with the name pressure_sat already exists. Please use update_property method if you wish to update an existing property's metadata.`

I think this is because you are trying to define a custom property that already exists. However, if you choose to get rid of the `define_custom_properties` in `adm1_properties_vapor` and add `pressure_sat` under `add_properties`, this seemingly minor change has a lot of effects downstream. Namely the number of unused variables in the `test_anaerobic_digestor` goes from 0 to 4 and many values in `test_solution` change in this file and in others. Obviously this is not supposed to happen so I'm not sure where to go from here. Maybe it's best if we just leave it as `p_sat`...

andrewlee94:
@MarcusHolly Changing the metadata definition in itself should have no effect on the variables and constraints in the system - this implies you are doing something unexpected somewhere else. Whilst it might be a pain, this suggests that there is something that should be looked into here.

Have you looked at what the unused variables being reported are? That would be the first step in working out what is happening.

MarcusHolly:
> @MarcusHolly Changing the metadata definition in itself should have no effect on the variables and constraints in the system - this implies you are doing something unexpected somewhere else. Whilst it might be a pain, this suggests that there is something that should be looked into here.
> 
> Have you looked at what the unused variables being reported are? That would be the first step in working out what is happening.

These are the variables that are currently unused. I'll try to look more into these b/c it's not immediately obvious to me why these concentrations are unused while others are.

image

andrewlee94:
@MarcusHolly It is also unclear as to how some of those would relate to saturation pressure. It might be due to the CH4 (which I assume does have a saturation pressure), and that the others are calculated directly from there. However, this does not explain why they would be unused - there should still be some constraints using those variables.

MarcusHolly:
So I think that these variables actually are unused. The `adm1_reactions.py` has a `Ci_dict`, which lists the carbon content of various components, and these entries are used exclusively in calculating S_IC in the Petersen matrix. It seems like every entry other than these 4 is used somewhere in the Petersen matrix, but that raises the question of why this failure is only appearing after making the `p_sat` to `pressure_sat` change (as Andrew elaborated on above)

andrewlee94:
@MarcusHolly That is something that needs to be answered - if they are really meant to be unused but aren't, then we need to know what is using them in the current version and why.

MarcusHolly:
> @MarcusHolly That is something that needs to be answered - if they are really meant to be unused but aren't, then we need to know what is using them in the current version and why.

Agreed -  I'll look into this, but @agarciadiego may be a better fit since he worked on this directly. Not sure if he's back from his break yet tho.

MarcusHolly:
I believed I figured out what the issue is - I'll make a separate PR so as not to conflate it with the issues being addressed in this one.

adam-a-a:
> I believed I figured out what the issue is - I'll make a separate PR so as not to conflate it with the issues being addressed in this one.

What do you think it is?

watertap/property_models/anaerobic_digestion/adm1_reactions.py
```diff
@@ -1110,15 +1110,6 @@ def build(self):
             ("R19", "Liq", "X_I"): 0,
         }
 
-        s_ic_rxns = ["R5", "R6", "R10", "R11", "R12"]
-
-        for R in s_ic_rxns:
-            self.rate_reaction_stoichiometry[R, "Liq", "S_IC"] = -sum(
-                self.Ci[S] * self.rate_reaction_stoichiometry[R, "Liq", S] * mw_c
-                for S in Ci_dict.keys()
```
adam-a-a: If you delete this, you might not use those Ci values, which I guess are showing up as unused vars.

MarcusHolly: Yep - that's exactly what it is

MarcusHolly: So what I'm working on now (in a separate PR) is removing this for loop (this is something Alejandro, Chenyu, and I discussed earlier) and making sure the Petersen matrix entries for S_IC are being computed correctly. For example, there are some places where the value for `Ci["X_su"]` is being hard-coded rather than called from the dictionary

MarcusHolly: This shouldn't be a functional change in any way. It's essentially just incorporating this for loop into the matrix directly - which is consistent with what Chenyu and I did for modified ADM1

adam-a-a: I see. I put this for loop in in the initial formulation of ADM1 (https://github.com/watertap-org/watertap/pull/627/commits/bf942b39808f267e44130afc58f39fa40b1e82c7) thinking it would be more efficient and cleaner. Can you remind me why this was even consider an issue? Is it more an issue of consistency?

MarcusHolly: After working on the new PR (and discussing with Alejandro behind-the-scenes), I see why you put in the for loop and it might make more sense to just keep it in. But yes, the original issue Chenyu and I saw with it was just consistency. So I think the solution Alejandro and I reached is that we should just delete the 4 keys in Ci_dict that are unused.

adam-a-a: Wait- but are they actually unused, or are they just being used under a different parameter/variable name?

MarcusHolly:
@adam-a-a  @agarciadiego So in summary, this is currently what the PR does:

1.) Delete the `define_custom_properties` and instead replace `p_sat` with `pressure_sat` under `add_properties` in `adm1_properties_vapor.py` 

2.) Go through the adm1 files & replace instances of `p_sat` with `pressure_sat`

3.) Delete the four unused variables in the Ci_dict: X_su, S_ch4, X_aa, and X_pro

There's no need for the other PR I was working on anymore, but Alejandro is planning on resolving the issue concerning how the first change is affecting the test values.

watertap/property_models/anaerobic_digestion/adm1_reactions.py
MarcusHolly: The latter, I believe

MarcusHolly: Since some of those parameters have the same value

agarciadiego: The change affected a tested value of the matrix (which shouldn't be affected by pressure in any way)

watertap/property_models/anaerobic_digestion/adm1_reactions.py
```diff
@@ -117,15 +117,11 @@ def build(self):
             "S_bu": 0.0250,
             "S_pro": 0.0267867187876688,
             "S_ac": 0.0312221602404522,
-            "S_ch4": 0.0156255859594734,
```
adam-a-a: So, these deleted factors--where are they used in the Petersen matrix under a different parameter/var? The reason I ask is because I don't think we should have majority of carbon content parameters defined here, while the four values being deleted are being used somewhere, only under a different attribute or local variable. Does that make sense?

adam-a-a (approved):
LGTM

agarciadiego (approved):
LGTM

PR #998, opened 2023-04-04
IonExchange0D Model Overhaul

kurbansitterley:
## Fixes/Resolves:

Improves ion exchange model stability, adds CONFIG options for isotherm.

## Summary/Motivation:

IonExchange0D has long suffered from stability problems that made it difficult to use and caused the tests to fail. As part of the PFAS analysis we are doing, I have done an overhaul of the entire model and this PR (hopefully) addresses many of the issues of the current IX unit model while (mostly) leaving the costing model intact. I have tested the results of the current model with the model included in this PR, and they are in strong agreement.

In addition (and also a result of the PFAS analysis), I have added new CONFIG options to the model: 
- The ability to use both Langmuir and Freundlich isotherms for equilibrium concentrations.  Equilibrium data for IX is presented for both isotherms, so this will expand the usable data for the model.

Adds CONFIG option to use either Langmuir or Freundlich isotherms for equilibrium calculation (``config.isotherm``) and associated class ``IsothermType``

Elimination of ``properties_regen`` in favor of using ``ControlVolume0D`` and tracking the mass of contaminant absorbed (rather than also trying to calculate the regeneration flow rate -- I believe this was a significant part of the reason why the previous model was difficult to get consistent solves). Here I use a very similar approach to that in our GAC model ``gac.py``.

Conversion of several Vars to mutable Params (or Expressions), including:
- ``separation_factor``
- ``bed_vol``
- ``col_vol_per``
- ``rate_coeff``
- variables and constraints relating to durations of different parts of IX cycle (e.g., ``t_bw``, ``t_regen``) and their flow rates (e.g., ``bw_flow``, ``rinse_flow``)
- pumping power variables (e.g, ``main_pump_power``)
- ``holdup``
- many column/bed design variables and constraints
- ``HTU``
- elimination of many constraints that are handled by ``ControlVolume0D``

Improvement of initialization routine.

Improvement of scaling factors.

Updated unit model test file, demo, and demo test file.

Changes and improvements to IX costing to accommodate unit model changes

Still TODO:
- [x] finish incorporating Freundlich equation
- [x] write tests for Freundlich isotherm
- [x] update documentation

### Legal Acknowledgement


bknueven:
@kurbansitterley big picture question: why the conversion from `Var`s to `Param`s?

Generally we leave things as `Var`s, even if they're really parameters, because that enables an advanced user to fit the parameters to their own data easily utilizing parameter estimation techniques, e.g., using [`parmest`](https://pyomo.readthedocs.io/en/stable/contributed_packages/parmest/index.html).

watertap/examples/flowsheets/ion_exchange/tests/test_ion_exchange_demo.py
```diff
@@ -194,28 +189,33 @@ def test_optimization(self, ix_0D):
         results = solver.solve(m)
         assert_optimal_termination(results)
         assert degrees_of_freedom(m) == 0
-        assert value(m.fs.ion_exchange.number_columns) == 5
-        assert value(m.fs.ion_exchange.bed_depth) == pytest.approx(1.24567, abs=1e-4)
+        assert value(m.fs.ion_exchange.number_columns) == 6
+        assert value(m.fs.ion_exchange.bed_depth) == pytest.approx(1.622805, abs=1e-4)
+        assert value(m.fs.ion_exchange.t_breakthru) == pytest.approx(
+            133415.81129, abs=1e-4
```
bknueven:
```suggestion
            133415.81129, rel=1e-6
```

watertap/examples/flowsheets/ion_exchange/tests/test_ion_exchange_demo.py
```diff
@@ -194,28 +189,33 @@ def test_optimization(self, ix_0D):
...
     def test_main_fun(self):
         m = ixf.main()
 
         assert degrees_of_freedom(m) == 0
-        assert value(m.fs.ion_exchange.number_columns) == 5
-        assert value(m.fs.ion_exchange.bed_depth) == pytest.approx(1.24567, abs=1e-4)
+        assert value(m.fs.ion_exchange.number_columns) == 6
+        assert value(m.fs.ion_exchange.bed_depth) == pytest.approx(1.622805, abs=1e-4)
+        assert value(m.fs.ion_exchange.t_breakthru) == pytest.approx(
+            133415.81129, abs=1e-4
```
bknueven:
```suggestion
            133415.81129, rel=1e-6
```

bknueven (commented):
My suggestions should fix the test failures on Linux. No need to check that number to 10 significant figures.

kurbansitterley:
> @kurbansitterley big picture question: why the conversion from `Var`s to `Param`s?
> 
> Generally we leave things as `Var`s, even if they're really parameters, because that enables an advanced user to fit the parameters to their own data easily utilizing parameter estimation techniques, e.g., using [`parmest`](https://pyomo.readthedocs.io/en/stable/contributed_packages/parmest/index.html).

The short answer to this is that many of the ``Var``s I changed to ``Param``s``/Expression``s because it made the model much easier to solve and they had minimal impact on costing outcomes. In essence, there was nothing to really optimize. Here I am mostly thinking of variables relating to the regeneration/backwashing/rinsing steps of the IX cycle that were previously used as part of the mass balance, but I have simplified all of that by using the ``ControlVolume0D`` and those variables are now used to calculate pumping powers, which is a small part of the overall operating cost.

Other parameters that are part of regressions (e.g., ``p_drop_A``, ``bed_expansion_frac_A``) I fit using excel and I think would be way overkill to use ``parmest`` to fit the same parameters. A user could still get a backwashing bed expansion curve from a manufacturer and fit the data in excel and then set the Params in WaterTAP. 

However, I am not opposed to changing them back to ``Var``s.

kurbansitterley:
@hunterbarber @adam-a-a @bknueven I am marking this ready for review. Documentation still needs to be updated with Freundlich isotherm approach, but I will be out the next week so any thoughts on getting the model in now and documentation in a separate PR in the interest of making sure the model is ready for the June release?

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -425,117 +582,25 @@ def build(self):
...
             units=pyunits.s,
-            doc="Cycle time",
+            doc="Resin contact time",
         )
 
-        self.t_contact = Var(
+        self.ebct = Var(
             initialize=520,
             bounds=(120, None),
             units=pyunits.s,
```
hunterbarber, line 583: GAC although typically runs with an ebct of 300-600 seconds, for the RSSCT runs with ebcts << 100 seconds. The challenge is to balance the model stability with the niche scenarios that the bounds will have to be relaxed, but I'm assuming these are better for model stability.

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -599,539 +650,668 @@ def build(self):
...
+                doc="Fluid mass transfer coefficient",
+            )
 
-        self.bed_expansion_frac = Var(
-            initialize=0.5,
-            bounds=(0.4, 1.2),
-            units=pyunits.dimensionless,
-            doc="Fraction of bed depth increase during backwashing",
-        )
+        if self.config.isotherm == IsothermType.freundlich:
```
hunterbarber, line 711: Given this is a massive model, would you ever think it would be valuable to modular some aspects? Say isotherms available in different folders with the same structure, and a generalized backwash function call, or all of these too specific to the scenario? Not saying you specifically relative to this model doing it, but just for general ease of use for future models and advancements?

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -599,539 +650,668 @@ def build(self):
...
 
-        self.bed_expansion_h = Var(
-            initialize=0.5,
-            units=pyunits.m,
-            bounds=(0, None),
-            doc="Additional column sidewall height required for bed expansion",
-        )
+            self.num_traps = 5  # TODO: make CONFIG option
+            self.trap_disc = range(self.num_traps + 1)
+            self.trap_index = self.trap_disc[1:]
```
hunterbarber, line 715: We should probably also try to standardize our use of the trapezoid rule for use in adsorption breakthroughs referencing the trapezoid rule implementation in the DAE package (I think), not for this PR though.

hunterbarber (commented):
Costing changes seem straightforward to me. With the `total_installed_cost_factor`, is this not something by the overarching WaterTAP costing package, similar to labor fees and land fees? I'm not sure whether this factor is directly related to the installation of specifically IX, but it feels like one of the fees that should be applied to the entire `m.fs` and not the isolated units.

Since there's a lot of changes in this, ranging from large to small, I'll probably do another thorough review once I see the documentation that states what the most significant changes are. With that I don't have any drastic changes, so I would say go ahead with the documentation. I may be more nitpicky on a thorough review once that is up.

kurbansitterley:
> Costing changes seem straightforward to me. With the `total_installed_cost_factor`, is this not something by the overarching WaterTAP costing package, similar to labor fees and land fees? I'm not sure whether this factor is directly related to the installation of specifically IX, but it feels like one of the fees that should be applied to the entire `m.fs` and not the isolated units.

This factor is not related specifically to IX, but I do think it is something we could include at the global costing level and apply to unit models as needed (this is how we did it for WT3). It is just a costing factor and the idea is that it compensates for the unaccounted for installed cost of the system. My justification for using it here is that I took costing from the EPA WBS model and didn't come close to accounting for all the costs in their approach. I am not opposed to taking it out, but I think it would make our IX costing results appear low compared to other IX costing estimates (and making this comparison is another reason I used to justify its inclusion).

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: You're right on both accounts- the lower ``ebct`` bound is lower than you might see for suggested operation of IX units. However, I wanted to be able to push the upper bounds of ``service_flow_rate`` to >15 BV/hr (my initial ub) to better align with the ranges I had seen published, and lowering the EBCT lb helped in stability for that.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: Yes it is something I thought about doing for this overhaul, but thought I might get some push back... at least for this version though, I opted to keep it all contained in one file. Ultimately, tweaking backwashing/rinsing step parameters have very small impacts to cost (and no impact to performance), so I thought about excluding them entirely (this is also why I changed them from Vars to Params and Expressions- there's no real need to optimize anything there at this point). 

However, there are important aspects of the IX regeneration process that are excluded from this model. If/when we get to wanting to model IX regen brine, that is definitely something that could be modularized. 

But I do agree the model is huge and sifting through the code to find constraints/vars is tedious. I am not opposed to separating them out if we think it is valuable at this point.

adam-a-a:
> > Costing changes seem straightforward to me. With the `total_installed_cost_factor`, is this not something by the overarching WaterTAP costing package, similar to labor fees and land fees? I'm not sure whether this factor is directly related to the installation of specifically IX, but it feels like one of the fees that should be applied to the entire `m.fs` and not the isolated units.
> 
> This factor is not related specifically to IX, but I do think it is something we could include at the global costing level and apply to unit models as needed (this is how we did it for WT3). It is just a costing factor and the idea is that it compensates for the unaccounted for installed cost of the system. My justification for using it here is that I took costing from the EPA WBS model and didn't come close to accounting for all the costs in their approach. I am not opposed to taking it out, but I think it would make our IX costing results appear low compared to other IX costing estimates (and making this comparison is another reason I used to justify its inclusion).

I think @hunterbarber was referring to `factor_total_investment` which is already in the watertap costing package and used in a few examples (and in the zero order costing package, I think it was referred to as `TIC`).

watertap/unit_models/ion_exchange_0D.py
adam-a-a: Based on @hunterbarber 's comment stating "RSSCT runs with ebcts << 100 seconds," I interpreted the comment to mean that your lower bound on ebct could/should be lower than 120 seconds, so long as it doesn't somehow affect model stability for your use case.

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -141,60 +203,129 @@ class IonExchangeODData(InitializationMixin, UnitModelBlockData):
...
+        Environ Sci Technol, 21(6), 573-580. doi:10.1021/es00160a008
 
-        # Michaud, C.F. (2013)
-        # Hydrodynamic Design, Part 8: Flow Through Ion Exchange Beds
-        # Water Conditioning & Purification Magazine (WC&P)
-        # https://wcponline.com/2013/08/06/hydrodynamic-design-part-8-flow-ion-exchange-beds/
+        Croll, H. C., Adelman, M. J., Chow, S. J., Schwab, K. J., Capelle, R., Oppenheimer, J., & Jacangelo, J. G. (2023). 
+        Fundamental kinetic constants for breakthrough of per- and polyfluoroalkyl substances at varying empty bed contact times: 
+        Theoretical analysis and pilot scale demonstration. 
+        Chemical Engineering Journal, 464. doi:10.1016/j.cej.2023.142587
```
adam-a-a: This should go in a doc string towards beginning of the file describing the model.

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -141,60 +203,129 @@ class IonExchangeODData(InitializationMixin, UnitModelBlockData):
...
         self.target_ion_set = Set(
             initialize=[target_ion]
         )  # create set for future development of multi-component model
+        inerts = comps - self.target_ion_set
 
         if "+" in target_ion:
             self.ion_exchange_type = IonExchangeType.cation
-            self.regen_chem = RegenerantChem.HCl
         if "-" in target_ion:
             self.ion_exchange_type = IonExchangeType.anion
```
adam-a-a: This seems a little questionable since the user would need to know to have a + or - in the target_ion string. Let's say that wouldn't be hard to do. I'd think we still need a way to enforce this. Do you have a check for this?

Alternatively, assuming you're still using MCAS for this, you could set your solute_list and provide charge for each ion. Then, target ion could be set and you could use charge to determine` ion_exchange_type`. For example,

```
if self.config.property_package.charge_comp[target_ion] > 0:
    self.ion_exchange_type = IonExchangeType.cation
elif self.config.property_package.charge_comp[target_ion] < 0:
    self.ion_exchange_type = IonExchangeType.anion
else:
    raise Exception    
```

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -544,36 +609,22 @@ def build(self):
...
             units=pyunits.m / pyunits.s,
-            doc="Interstitial velocity",
-        )
-
-        self.holdup = Var(
-            initialize=100,
-            bounds=(90, 250),  # Inamuddin/Luqman
-            units=pyunits.dimensionless,
-            doc="Holdup percent",
+            doc="Interstitial velocityv through bed",
```
adam-a-a:
```suggestion
            doc="Interstitial velocity through bed",
```

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -544,36 +609,22 @@ def build(self):
...
-        self.pressure_drop = Var(
-            initialize=14,
-            units=pyunits.psi,
-            bounds=(0, 25),
-            doc="Pressure drop across column",  # max pressure drop is 25 psi, MWH
-        )
-
         # ====== Dimensionless variables ====== #
 
         self.Re = Var(
```
adam-a-a:
```suggestion
        self.N_Re = Var(
```

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -584,7 +635,7 @@ def build(self):
         )
 
         self.Sc = Var(
```
adam-a-a:
```suggestion
        self.N_Sc = Var(
```

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -599,539 +650,668 @@ def build(self):
 
         self.Pe_p = Var(
```
adam-a-a:
```suggestion
        self.N_Pe_particle = Var(
```

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -599,539 +650,668 @@ def build(self):
...
             initialize=0.1,
-            # bounds=(0.01, 0.8),
             units=pyunits.dimensionless,
             doc="Peclet particle number",
         )
 
         self.Pe_bed = Var(
```
adam-a-a:
```suggestion
        self.N_Pe_bed = Var(
```

lbianchi-lbl:
The review comments should be addressed before merging this.

adam-a-a (changes requested):
Some comments to be addressed and likely another round of reviewing

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: modified all dimensionless numbers to match this format

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: modified all dimensionless numbers to match this format

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: modified all dimensionless numbers to match this format

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: modified all dimensionless numbers to match this format

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: fixed in changes

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: I pretty much took exactly your suggested implementation here.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: Moved to top of file.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: Lowered LB.

docs/technical_reference/unit_models/ion_exchange_0D.rst
```diff
@@ -36,288 +80,437 @@ The model provides three ports (Pyomo notation in parenthesis):
...
+   "Bed capacity parameter", ":math:`A`", "``bed_capacity_param``", None, ":math:`\text{dimensionless}`"
+   "Bed volumes at breakthrough", ":math:`BV`", "``bv``", "None", ":math:`\text{dimensionless}`"
+   "Bed volumes at 50% influent conc.", ":math:`BV_{50}`", "``bv_50``", "None", ":math:`\text{dimensionless}`"
+   "Kinetic fitting parameter", ":math:`r`", "``kinetic_param``", "None", ":math:`\text{dimensionless}`"
+   "Mass transfer coefficient", ":math:`k_T`", "``mass_transfer_coeff``", "None", ":math:`\text{s}^{-1}`"
+   "Concentration at breakthrough", ":math:`C_{b}`", "``c_breakthru``", "``target_ion_set``", ":math:`\text{kg/}\text{m}^3`"
+   "Average relative breakthrough concentration at breakthrough time", ":math:`X_{avg}`", "``c_norm_avg``", "None", ":math:`\text{dimensionless}`"
+   "Relative breakthrough conc. for trapezoids", ":math:`X_{trap,k}`", "``c_traps``", "``k``", ":math:`\text{dimensionless}`"
+   "Breakthrough times for trapezoids", ":math:`t_{trap,k}`", "``tb_traps``", "``k``", ":math:`\text{s}`"
+   "Area of trapezoids", ":math:`A_{trap,k}`", "``traps``", "``k``", ":math:`\text{dimensionless}`"
```
hunterbarber, line 219: So you're only using trapezoid rule for the Freundlich isotherm? Average concentration over the breakthrough time is still calculated using mass removed in the Langmuir?

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -118,19 +166,76 @@ class IonExchangeODData(InitializationMixin, UnitModelBlockData):
...
+            domain=str,
+            description="Designates targeted species for removal",
+        ),
     )
 
     CONFIG.declare(
         "regenerant",
         ConfigValue(
-            default=RegenerantChem.HCl,
+            default=RegenerantChem.none,
```
hunterbarber: Since `none` always defaults to NaCl, what was the motivation for changing this, and having a conditional that acts as a 'default' in line 283?

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -141,85 +246,103 @@ class IonExchangeODData(InitializationMixin, UnitModelBlockData):
...
+        comps = self.config.property_package.component_list
         target_ion = self.config.target_ion
+
         self.target_ion_set = Set(
             initialize=[target_ion]
         )  # create set for future development of multi-component model
+        inerts = comps - self.target_ion_set
 
-        if "+" in target_ion:
+        if len(self.target_ion_set) > 1:
```
hunterbarber, line 269: I have some testing on this set for GAC if we need to up the coverage to test these config errors, https://github.com/watertap-org/watertap/blob/main/watertap/unit_models/tests/test_gac.py#L647

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -141,85 +246,103 @@ class IonExchangeODData(InitializationMixin, UnitModelBlockData):
...
+        self.distributor_h = Param(
+            initialize=0.5,
+            mutable=True,
+            units=pyunits.m,
+            doc="Distributor height",  # Perry's
         )
 
-        self.holdup_exp = Param(
-            initialize=0.28, units=pyunits.dimensionless, doc="Holdup equation exponent"
+        self.p_drop_psi_to_m = Param(
```
hunterbarber: You know this works, right? 
```
    pyunits.convert(
        1*pyunits.psi/(998*pyunits.kg*pyunits.m**-3)/(9.81*pyunits.m*pyunits.second**-2),
                       to_units=pyunits.m)
```
I confirmed its ~0.703 using P=rho*g*h, which I think you just revert these calculations later. You are not actually doing a conversion, but instead using the pressure head equation, which isn't clear from this. I would say for conciseness this should be revised.

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -242,20 +365,27 @@ def build(self):
         # Wilson and Geankoplis, Ind. Eng. Chem. Fundam., 5, 9 (1966)
 
         self.Sh_A = Param(
-            initialize=1.09,
+            initialize=2.4,
```
hunterbarber, line 357: Are these changed because of incorrect units in the past? Not clear

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -411,11 +560,11 @@ def build(self):
...
-        self.col_vol_per = Var(
-            initialize=10,
-            bounds=(0.02, 70),
-            units=pyunits.m**3,
-            doc="Column volume",
+        self.col_height_to_diam_ratio = Var(
+            initialize=1,
+            bounds=(1, 100),
+            units=pyunits.dimensionless,
+            doc="Min ratio of bed depth to diameter",
```
hunterbarber, line 556:
```suggestion
            doc="For stability, bounded bed depth to diameter ratio",
```

watertap/unit_models/ion_exchange_0D.py
hunterbarber: On this review, the model is still huge.

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -544,594 +601,688 @@ def build(self):
...
+                initialize=0.5,
+                bounds=(0, None),
+                units=pyunits.kg / pyunits.m**3,
+                doc="Breakthrough concentration of target ion",
+            )
 
-        self.rinse_pump_power = Var(
-            initialize=0.1, units=pyunits.kW, doc="Rinse pump power"
-        )
+            self.freundlich_n = Var(
```
hunterbarber, line 767: You and I should probably be explicit here, at least in the doc string for n vs 1/n and 'exponent'

hunterbarber (commented):
Got through some but not all of it with this review

docs/technical_reference/unit_models/ion_exchange_0D.rst
kurbansitterley: That's correct. Mass removed by the resin divided by the breakthrough time is the CV mass transfer coefficient for Langmuir. There is a TODO somewhere in there for applying the trapezoid rule for the Langmuir config...

However, the "breakthrough curve" produced using the Langmuir config is essentially a square wave. I think this results from the approach used to solve the differential equations used as the model basis. I haven't fully investigated it but I don't think applying the trapezoid rule to the Langmuir config would be as straightforward.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: This isn't _totally_ for stability, more to ensure you don't end up with pancake-shaped columns. But with a LB = 1, they can end up like cubes, which isn't ideal but I left it for stability.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: I changed the regression I was using for Sh, that is why the number changed.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: Yes, I took guidance from GAC here because it is _not_ inverted ``n``

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: -_- ... yes of course I knew that. 

Addressed in my last commit. Turns out you don't need density or gravity at all.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: Trying to figure out and... I have no idea why I did this. Possibly so I could change the regen chem without having to re run the model when I was doing some sensitivity or something. Anyway I took it out.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: So I meant to take this conditional out because I don't think it is possible for ``len(self.target_ion_set) > 1``. The only component that gets put in ``target_ion_set`` is ``target_ion`` which is set via CONFIG. Since the domain is ``str``, it seems anything that you pass as ``target_ion`` gets coerced to a ``str`` regardless. So if you pass ``["Ca_2+". "Na_+"]``, ``target_ion_set`` ends up with a single member ``'["Ca_2+". "Na_+"]'``, which then triggers a ``KeyError`` on the line 273 because ``'["Ca_2+". "Na_+"]'`` is ``not valid for indexed component fs.properties.charge_comp``. 

So I don't think this conditional is doing anything, but I am open to suggestions for how to implement the check needed. I see in GAC you have the domain as a list for ``target_species``... maybe that is the way to go?

watertap/unit_models/ion_exchange_0D.py
hunterbarber: Stability of optimization, at least so a posed optimization doesn't push toward a pancake. I was always conflicted on whether to include this or not in my models. It adds another equation which may challenge computational expense, but often in optimization I want to fix this length to diameter ratio to ~8 and optimize the vessel sizing this way. 

This is just conversational, and I wish there was a clear answer and standard on this across models.

watertap/costing/units/ion_exchange.py
```diff
@@ -221,7 +230,9 @@ def cost_ion_exchange(blk):
     """
     make_capital_cost_var(blk)
     make_fixed_operating_cost_var(blk)
+    ion_exchange_params = blk.costing_package.ion_exchange
     # Conversions to use units from cost equations in reference
+    tot_num_col = blk.unit_model.number_columns + blk.unit_model.number_columns_redund
```
adam-a-a, line 235: It would be nice to have some of these as expressions for the user to access directly in the model results. For example, total number of columns seems like something that could be of interest to a user and would fit well under the `_get_performance_contents` method in the IX unit model.

watertap/examples/flowsheets/ion_exchange/ion_exchange_demo.py
```diff
@@ -393,10 +303,10 @@ def display_results(m):
         f'{"TOTAL Capital Cost":<40s}{f"${ix.costing.capital_cost():<39,.2f}"}{"$":<40s}'
```
adam-a-a, line 303: Do you use the `report` method at all for IX? It could be added here potentially.

watertap/examples/flowsheets/ion_exchange/tests/test_ion_exchange_demo.py
```diff
@@ -92,80 +90,66 @@ def test_build_model(self, ix_0D):
...
+            56759.75759, rel=1e-3
         )
         assert value(m.fs.ion_exchange.partition_ratio) == pytest.approx(
-            235.99899, rel=1e-6
+            235.99899, rel=1e-3
         )
 
         assert value(m.fs.costing.specific_energy_consumption) == pytest.approx(
-            0.057352, abs=0.001
+            0.057245, abs=0.001
```
adam-a-a: Any reason to use absolute tolerance over relative? Typically see absolute tolerance for zero value checks and relative tolerance for non-zero values.

watertap/examples/flowsheets/ion_exchange/tests/test_ion_exchange_demo.py
```diff
@@ -92,80 +90,66 @@ def test_build_model(self, ix_0D):
...
         )
-        assert value(m.fs.costing.LCOW) == pytest.approx(0.219, abs=0.01)
+        assert value(m.fs.costing.LCOW) == pytest.approx(0.222437, abs=1e-3)
```
adam-a-a: BTW= do you `assert_units_consistent` anywhere in this test?

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -35,22 +39,60 @@
...
+Vassilis J. Inglezakis and Stavros G. Poulopoulos
+Adsorption, Ion Exchange and Catalysis: Design of Operations and Environmental Applications (2006).
+doi.org/10.1016/B978-0-444-52783-7.X5000-9
+
+Michaud, C.F. (2013)
+Hydrodynamic Design, Part 8: Flow Through Ion Exchange Beds
+Water Conditioning & Purification Magazine (WC&P)
+https://wcponline.com/2013/08/06/hydrodynamic-design-part-8-flow-ion-exchange-beds/
+
+Clark, R. M. (1987).
```
adam-a-a, line 78: You might've done this, but did you include the Clark ref in the rst file?

watertap/unit_models/ion_exchange_0D.py
adam-a-a: Maybe a list of strings as the expectation?

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -287,51 +409,83 @@ def build(self):
...
+            doc="Number of redundant columns for ion exchange process",
         )
+        # ==========VARIABLES==========
+        # COMMON TO LANGMUIR + FREUNDLICH
 
         self.resin_diam = Var(
             initialize=7e-4,
-            bounds=(5e-4, 1.5e-3),
-            units=pyunits.m,  # Perry's
+            bounds=(5e-4, 1.5e-3),  # Perry's
```
adam-a-a, line 487: FYI- @bknueven often suggests moving these inline comments above lines of code to prevent funkiness from black formatting.

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -387,12 +525,12 @@ def build(self):
...
 
         self.bed_depth = Var(
-            initialize=1, bounds=(0, 2), units=pyunits.m, doc="Bed depth"  # EPA-WBS
+            initialize=1, bounds=(0, 3.0), units=pyunits.m, doc="Bed depth"  # EPA-WBS
         )
 
         self.bed_porosity = Var(
             initialize=0.5,
-            bounds=(0.45, 0.65),
+            bounds=(0.3, 0.8),
```
adam-a-a, line 533: Did you observe effects on model stability without bounds? Do problems occur if you expand upper bound to be very close to 1? I understand it may not be practical for use cases but just curious.

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -425,117 +563,25 @@ def build(self):
...
-            doc="Contact time",
+            doc="Resin contact time",
         )
 
-        self.t_waste = Var(
-            initialize=1800,
-            bounds=(1, None),
+        self.ebct = Var(
+            initialize=520,
+            bounds=(90, None),
```
adam-a-a, line 582: Does this lower bound impact model stability?

watertap/unit_models/tests/test_ion_exchange_0D.py
```diff
@@ -598,331 +442,482 @@ def test_build(self, IX_frame_with_inert):
...
-        H2O_mass_frac = 1 - sum(x for x in ix_prop_in.values())
-        H2O_mol_comp_flow = (
-            H2O_mass_frac
-            * pyunits.kg
-            / pyunits.kg
-            * mass_flow_in
-            / m.fs.unit.config.property_package.mw_comp["H2O"]
+        m.fs.ix = ix = IonExchange0D(**ix_config)
+
+        # c0 = pyunits.convert(c0, to_units=pyunits.kg / pyunits.m**3)
```
adam-a-a:
```suggestion
```

adam-a-a (approved):
Other than adding `assert_units_consistent` to your demo file test and some minor cleanup, I don't have any pressing changes to suggest. Considering this PR is important for our PFAS effort, I also don't want to hold things up.  Refinement can always happen in a later PR as needed. 

Great job on this model! This looks like it took a lot of effort.

watertap/costing/units/ion_exchange.py
```diff
@@ -328,24 +340,30 @@ def cost_ion_exchange(blk):
...
                 + blk.capital_cost_backwash_tank
                 + blk.capital_cost_regen_tank
             )
-            * TIC,
+            * ion_exchange_params.total_installed_cost_factor,
             to_units=blk.costing_package.base_currency,
         )
     )
     if blk.unit_model.config.hazardous_waste:
+        blk.regen_dens = 1000 * pyo.units.kg / pyo.units.m**3
```
hunterbarber, line 352: Does this need to be accessible as a parameter?

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -544,594 +590,664 @@ def build(self):
...
+        )
+        def eq_isobaric_regen_stream(b):
+            return prop_in.pressure == regen.pressure
 
-        @self.Constraint(doc="Temperature conservation")
-        def eq_temp_conservation(b):
-            return b.properties_in[0].temperature == b.properties_out[0].temperature
+        for j in inerts:
+            self.process_flow.mass_transfer_term[:, "Liq", j].fix(0)
+            self.regeneration_stream[0].get_material_flow_terms("Liq", j).fix(0)
```
hunterbarber, line 988: Also be aware that we may have to change this soon when MCAS updates to allow for mass or molar flow in material flow terms.

watertap/examples/flowsheets/ion_exchange/tests/test_ion_exchange_demo.py
hunterbarber: Also seeing use of both 1e-3 and 1e-2, is that order of magnitude actually used to pass tests?

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -141,85 +246,92 @@ class IonExchangeODData(InitializationMixin, UnitModelBlockData):
...
-        # https://wcponline.com/2013/08/06/hydrodynamic-design-part-8-flow-ion-exchange-beds/
-
-        ion_set = self.config.property_package.ion_set
+        comps = self.config.property_package.component_list
         target_ion = self.config.target_ion
+
         self.target_ion_set = Set(
             initialize=[target_ion]
         )  # create set for future development of multi-component model
+        inerts = comps - self.target_ion_set
```
hunterbarber, line 267: Although it is future proofing, your domain in line 226 is still a string. So technically this `target_ion_set` is always initialized to a list of size 1 that contains a string. You could convert over line 226 to a list, but it is completely unnecessary until the model supports multicomponent anyway.

hunterbarber (approved):
Of all the lingering comments, I don't think any are model breaking. A lot of smaller formatting things, all of which can be handled soon or in subsequent PRs. So, I'm also approving with some of these comments available to be addressed soon.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: I have never looked at model stability related to bed porosity. This is something I assume the user would provide, and I expanded the bounds here to be >inclusive of some values I saw while looking at PFAS-specific resins.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: Yes, Clark (1987) is in the rst file.

watertap/examples/flowsheets/ion_exchange/tests/test_ion_exchange_demo.py
kurbansitterley: Yup, for all three frames: lines 243, 520, 821.

watertap/examples/flowsheets/ion_exchange/tests/test_ion_exchange_demo.py
kurbansitterley: Ah no good point, should be relative. Careless copy/paste on my part, I will change.

watertap/examples/flowsheets/ion_exchange/ion_exchange_demo.py
kurbansitterley: I typically don't because I want different units than the Vars are in, so I have made my own reporting functions.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: I don't think so (but I haven't really checked), this lower bound is more about having a lower limit on the EBCT so that when optimizing, the model doesn't force water through the column as fast as possible, which would break some of the assumptions of the modeling approach.

watertap/costing/units/ion_exchange.py
kurbansitterley: Its not a bad idea. I imagine you could have some pretty dense regen solutions, but unclear how much it would impact this calculation (hazardous waste disposal), which in my experience is a pretty small part of the costs when included.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: Since we are fixing it to zero, wouldn't it not really matter for these lines? And/or couldn't it be changed via CONFIG? I guess I am not really familiar with the MCAS updates.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: Yes right now you could use a list but it would coerce the list to a str anyway... but using a list in target_ion config would perpetuate other errors anyway so I didn't change it. I also dug into this after I had finished prepping the PR, but if you want this change to get it merged I will follow what you have in GAC.

hunterbarber: No, I think consistency isn't relevant until multicomponent adsorption is actually supported.

watertap/unit_models/ion_exchange_0D.py
hunterbarber: Yea, in this specific case it doesn't matter, but it will for the adsorbed species .

watertap/examples/flowsheets/ion_exchange/tests/test_ion_exchange_demo.py
kurbansitterley: Oops I missed this was the demo test file... updating now

PR #1006, opened 2023-04-17
Adding basic Electrolyzer unit model

hunterbarber:
## Fixes/Resolves:
New unit model for electrolyzer.

## Summary/Motivation:
New unit model for electrolyzer to simulate integrated treatment-train chemical generation

- Added electrolyzer unit model
- Added electrolyzer cost model
- Added electrolyzer basic testing
- Added electrolyzer documentation

### Legal Acknowledgement


hunterbarber:
@avdudchenko @lbibl @TimBartholomew I've added the documentation in the most recent pushes, also find them here with the [direct link](https://watertap--1006.org.readthedocs.build/en/1006/technical_reference/unit_models/electrolyzer.html). I made the docs in hopes that there's enough substance to gather what the fundamental assumptions are, so we can assess if we would like to remove any for the first merge of the model.

watertap/unit_models/electrolyzer.py
```diff
@@ -0,0 +1,731 @@
...
+            self.config.property_package.phase_list,
+            self.config.property_package.component_list,
+            doc="ion permeation through the membrane",
+        )
+        def eq_membrane_permeation(b, t, p, j):
+            return (
+                b.anolyte.mass_transfer_term[t, p, j]
+                == -b.electron_flow * b.membrane_stoich[p, j]
+            )
+
```
lbibl, line 486: I am trying to line up the codes with the current documentation.  could you write out the full mass/material balance eq in the doc, as N_in - N_out + N_r = 0 ?  If you have balance for both ion and molecule species, I think the stoich chef. should appear in the balance eq ?

lbibl:
By this point I read the documentation and the unit model codes.  I commented for a clarity of the material balance.  Some other thoughts are (1) I speculate a basic ohm's law should exist in the model to relate voltage to current, and (2) from the doc, it seems to me that the model will have equal surface area of electrodes and membrane, of which I am not certain.

watertap/unit_models/electrolyzer.py
hunterbarber: I agree that it wasn't exactly intuitive so I added the following to the documentation.
image

hunterbarber:
> By this point I read the documentation and the unit model codes. I commented for a clarity of the material balance. Some other thoughts are (1) I speculate a basic ohm's law should exist in the model to relate voltage to current, and (2) from the doc, it seems to me that the model will have equal surface area of electrodes and membrane, of which I am not certain.

Thanks for taking a look. The method of sizing electrode area is something I haven't found a clear metric on and hence why I have it simplified at the moment.  This and cost of anode and cathode materials are some things that needs updated but it's just a matter of finding the appropriate literature. I believe I have both listed as TODO's in-line.

docs/technical_reference/unit_models/electrolyzer.rst
```diff
@@ -0,0 +1,245 @@
...
+   "power efficiency", ":math:`\theta_{power}`", "efficiency_power", "None", ":math:`\text{dimensionless}`"
+   "molar flow of species j across the membrane from anolyte to catholyte", ":math:`\dot{n}_{mem,j}`", "mass_transfer_term", "[t, p, j]", ":math:`\frac{mol}{s}`"
+   "molar generation of species j by electrolysis at the anode", ":math:`\dot{n}_{ano,j}`", "custom_reaction_anode", "[t, j]", ":math:`\frac{mol}{s}`"
+   "molar generation of species j by electrolysis at the cathode", ":math:`\dot{n}_{cat,j}`", "custom_reaction_cathode", "[t, j]", ":math:`\frac{mol}{s}`"
+
+| :math:`^a` Variable intended to be move to a interchangeable component block, callable to the base electrolyzer model
+| :math:`^b` Value is normalized to 1 electron in electrolysis stoichiometry
+
+
+For non-fixed efficiencies, electrochemical potentials, and other relevant variables, custom constraints  may be constructed at the flowsheet level. These may allow for predicative performance as a function of temperature, concentration, overpotential, and other variables.
```
srikanthallu: you mean predictive performance.

watertap/unit_models/electrolyzer.py
```diff
@@ -0,0 +1,853 @@
...
+            bounds=(0, None),
+            domain=NonNegativeReals,
+            units=units_meta("current") * units_meta("length") ** -2,
+            doc="membrane current density",
+        )
+        self.membrane_ion_transport_number = Var(
+            self.config.property_package.phase_list,
+            self.config.property_package.component_list,
+            initialize=0,
+            bounds=(None, None),
```
srikanthallu, line 277: Is there a reason for not bounding the transport_number ?

docs/technical_reference/unit_models/electrolyzer.rst
```diff
@@ -0,0 +1,245 @@
...
+
+    .. math::
+
+        & \dot{n}_{in,j}-\dot{n}_{out,j}+\frac{\left(\varepsilon_{ano,j}-\varepsilon_{mem,j}\right)I\theta_{current}}{F}=0 \\\\
+
+Cathode:
+
+    .. math::
+
+        & \dot{n}_{in,j}-\dot{n}_{out,j}+\frac{\left(\varepsilon_{cat,j}+\varepsilon_{mem,j}\right)I\theta_{current}}{F}=0 \\\\
```
srikanthallu: \varepsilon_{mem,j} is missing from the nomenclature/variables definition

srikanthallu (commented):
Left some comments, will get back to it again.

watertap/unit_models/electrolyzer.py
hunterbarber: So in a standard simulation it would be between 0 and 1 where the sum of all transport numbers equals 1. However, since it is coded w.r.t. transport from anolyte to catholyte (cation exchange membrane), negative values could simulate back migration across the membrane of a CEM, or moreover model a AEM, so for now it is unbounded.

srikanthallu (approved):
LGTM for initial 0D model.

docs/technical_reference/unit_models/electrolyzer.rst
```diff
@@ -0,0 +1,245 @@
+Electrolyzer
+============
+.. note::
+
+  The development of the electrolyzer model is ongoing.
+
+This is a simplified electrolyzer unit model used to approximate electrolysis performance.  With the current build, the model is simulated under the following assumptions:
+   * simulation of this unit model is only supported with the Multi-Component Aqueous Solution (MCAS) property package
+   * supports liquid phase only, vapor components are modeled in the liquid phase
```
lbibl: Recommend deleting "vapor comp are... liquid phase".  By physical chemistry language, when gas molecules are in a liquid phase, they are solutes like others.  When speaking of "gas phase",  it means a matter condition with uniform physical properties.  The second half of the sentence may unnecessarily confuse about what is "gas components" referring to (phase or molecules).

docs/technical_reference/unit_models/electrolyzer.rst
```diff
@@ -0,0 +1,245 @@
...
+
+.. currentmodule:: watertap.unit_models.electrolyzer
+
+Introduction
+------------
+This model was primarily motivated to simulate the chlor-alkali membrane electrolysis process, a conventional electrolyzer for the production of chlorine gas and hydroxide (Bommaraju, 2015) (Kent, 2007). The model has been demonstrated for the chlor-alkali configuration in the electrolyzer testing file and may be generalizable to other electrolysis mechanisms, but has not been validated for generalized processes.
+
+Given a membrane electrolyzer, the catholyte and anolyte are separated by an ion exchange membrane. This provides two distinct control volumes to perform model calculations. The basis for determining electrolyzer performance is accounting for the Faradaic conversion of species with respect to the electrolysis reactions at the cathode and anode. Faradaic conversion considers equating the supplied electrical current to the amount of electrons available for electrochemical reaction by Faraday's law. The calculation of Faradaic conversion is described in "electrons passed between anode and cathode contributing to reactions" found in the :ref:`Equations <electrolyzer_equations>`.
+
```
lbibl, line 29: "Faraday's law" is also known as the basic electromagnetism law where induction is related to magnetic flux, although I can tell what it refers to here.  Maybe point it to some equation for clarity.

docs/technical_reference/unit_models/electrolyzer.rst
```diff
@@ -0,0 +1,245 @@
...
+   * membrane current density, :math:`i_{mem}`
+   * anode current density, :math:`i_{ano}`
+   * anode overpotential, :math:`\eta_{ano}`
+   * cathode current density, :math:`i_{cat}`
+   * cathode overpotential, :math:`\eta_{cat}`
+   * current, :math:`I`
+   * current efficiency, :math:`\theta_{current}`
+   * voltage efficiency, :math:`\theta_{voltage}` *or* resistance, :math:`R`
+
```
lbibl, line 46: I think there are redundant "efficiency" parameters here (considering they are all fixed so the redundancy is likely neglected).  Try clarify what is each efficiency accounting for?  For example, the current efficiency may account for all "side" electron consumption except desired electrolysis; this makes sense to parameterize since the model is not writing those side effects out.  However, for the voltage, you have already written out over potentials, ohmic potential drops, and thermo potential (you'll probably need a membrane potential"; with these you've already parameterized all "side" potential drops and what is the efficiency for?

docs/technical_reference/unit_models/electrolyzer.rst
```diff
@@ -0,0 +1,245 @@
...
+The following variables should then be fixed:
+
+.. code-block::
+
+   # membrane properties
+   m.fs.electrolyzer.membrane_ion_transport_number["Liq", "NA+"].fix(1)
+
+   # anode properties
+   # Cl- --> 0.5 Cl2 + e-
+   m.fs.electrolyzer.anode_electrochem_potential.fix(1.21)
```
lbibl, line 70: Are these thermo potential used anywhere?  There is no Nernst equation to account for the thermo simulation.

docs/technical_reference/unit_models/electrolyzer.rst
```diff
@@ -0,0 +1,245 @@
...
+
+        & H_2O+e^-\to{\small\frac{1}{2}}H_2+OH^-\hspace{4em}E=-0.99V \\\\
+
+The following variables should then be fixed:
+
+.. code-block::
+
+   # membrane properties
+   m.fs.electrolyzer.membrane_ion_transport_number["Liq", "NA+"].fix(1)
+
```
lbibl, line 67: If all ions are given a t_mem, those of anions should be fixed at 0.

docs/technical_reference/unit_models/electrolyzer.rst
```diff
@@ -0,0 +1,245 @@
...
+   # H20 + e- --> 0.5 H2 + OH-
+   m.fs.electrolyzer.cathode_electrochem_potential.fix(-0.99)
+   m.fs.electrolyzer.cathode_stoich["Liq", "H2O"].fix(-1)
+   m.fs.electrolyzer.cathode_stoich["Liq", "H2-v"].fix(0.5)
+   m.fs.electrolyzer.cathode_stoich["Liq", "OH-"].fix(1)
+
+Model Structure
+------------------
+The electrolyzer model consists of 2 ``ControlVolume0DBlocks``, one for the anolyte and another for the catholyte. Currently, the generation of species via electrolysis reactions is handled by the ``custom_molar_term`` within the ``ControlVolume0DBlock``. Coupling this custom molar term with the Faradaic conversion, the material balance may be written by the following and calculated within the ``ControlVolume0DBlock``. Here, membrane flux is handled by relating the stoichiometric coefficient required to satisfy the charge balance. The direction of membrane flow is calculated from anode to cathode in the unit model, lending to the signage of the membrane stoichiometric coefficient shown.
+
```
lbibl, line 86: These language isn't quite convention electrochemical statement and a little hard to get by me. For example, membrane flux of what component? what is "stoichiometric coefficient required to satisfy the charge balance."? what is "membrane flow"?       Some overall suggestions are 1) in a single model, you should have one unified flux direction and assign it to be positive, 2) be very careful with the current direction, which may be the same or opposite to the flux direction depending on the charge carrier's +/-; 3) herein you may simply state the stoich coef. is for the mass change on the electrode surface and the t_mem is for the mass transfer across the membrane.

docs/technical_reference/unit_models/electrolyzer.rst
```diff
@@ -0,0 +1,245 @@
...
+    .. math::
+
+        & \dot{n}_{in,j}-\dot{n}_{out,j}+\frac{\left(\varepsilon_{ano,j}-t_{mem,j}\right)I\theta_{current}}{F}=0 \\\\
+
+Cathode:
+
+    .. math::
+
+        & \dot{n}_{in,j}-\dot{n}_{out,j}+\frac{\left(\varepsilon_{cat,j}+t_{mem,j}\right)I\theta_{current}}{F}=0 \\\\
+
```
lbibl, line 98: The last comment covers through these lines.

docs/technical_reference/unit_models/electrolyzer.rst
```diff
@@ -0,0 +1,245 @@
...
+.. _electrolyzer_equations:
+
+Equations
+-----------
+.. csv-table::
+   :header: "Description", "Equation"
+
+   "membrane current density", ":math:`I = i_{mem}A_{mem}`"
+   "ion permeation through the membrane", ":math:`\dot{n}_{mem,j} = -t_{mem,j}\dot{n}_{e^-}`"
```
lbibl, line 171: I haven't read through the codes in detail.  But I think this equation overlaps the mass balance eq, where you've considered the flux across membrane as the mass_transfer term I assume.

lbibl (commented):
Some more feedback at the model description level reflected by the documentation.  Consider adding a diagram.

docs/technical_reference/unit_models/electrolyzer.rst
hunterbarber: Yes, all variables not overwritten with a nonzero value are fixed to zero within the build of the model.

docs/technical_reference/unit_models/electrolyzer.rst
hunterbarber: Overlaps isn't necessarily correct, but my reasoning is that, there technically is no variable constructed for a few variables at the unit model level. Rather, they are calculated by a variable within the control volume. I don't think we have a convention for these cases, so here I denote a _n_ as a flow rate term, but it matches the `mass_transfer_term`. However since `mass_transfer_term` is not exclusively membrane transfer, I think it may also be confusing to use some universal symbol as such.

I think this is a matter of is the documentation used to help users understand the model at the surface, or should it explain line by line and discuss every touched function, i.e., `ControlVolume0D` usage, or not. I'm going to make some changes based on your comments here once I figure out how to present it with more clarity. Open to suggestions.

docs/technical_reference/unit_models/electrolyzer.rst
hunterbarber: I understand the concern here, revised to "vapor-liquid phase equilibrium is not calculated." Trying to figure the best shorthand way to describe that Cl2 (v.) (gas molecule in vapor phase) and Cl2 (aq.) (dissolved gas molecule in aqueous phase) are indistinguishable in the current state and exist only in the `['Liq']` index

docs/technical_reference/unit_models/electrolyzer.rst
hunterbarber: Clarified to Faraday's law of electrolysis to differentiate from the Faraday's law of electromagnetism, and added this clarification to the corresponding equation in the equation section.

docs/technical_reference/unit_models/electrolyzer.rst
hunterbarber: @lbibl This is a good catch. It should be unnecessary given the new addition of `eq_voltage_cell`. The only value to this is if a user doesn't know the anode and cathode overpotentials and/or the resistance. Then a 'redundant' equation can replace

```
b.voltage_cell == (
                b.voltage_reversible
                + b.anode_overpotential
                + b.cathode_overpotential
                + b.current * b.resistance
            )
```
with
```
b.voltage_reversible == b.voltage_cell * b.efficiency_voltage
```
where 
```
b.efficiency_voltage = (
                + b.anode_overpotential
                + b.cathode_overpotential
                + b.current * b.resistance
)
```
but wants to be assumed at a single value because the individual values are unknown. Good for general analysis of low fidelity models.

Do you think this should just be removed or added as a config option?

docs/technical_reference/unit_models/electrolyzer.rst
hunterbarber: No temperature dependence is included in this first model, basically assuming all input parameters are fixed at the operating temperature and the operating temperature is held constant.

docs/technical_reference/unit_models/electrolyzer.rst
hunterbarber: @avdudchenko also could use some input based on the intended analysis around the model

lbibl: I recommend removing the voltage efficiency.

docs/technical_reference/unit_models/electrolyzer.rst
lbibl: So I guess what I called "thermo potential" is the "electrochemical potential here.  I think a better name should be standard reduction potential.   I am not sure if it is a good idea to separate two over potentials if this is supposed to be a simpler basic model. They are somewhat redundant variables if taking arbitrary values.  But not a technical mistake as well.

docs/technical_reference/unit_models/electrolyzer.rst
lbibl: If the electron flow serving as a intermediate variable is what you meant, it's technically ok.  I am not sure about this statement "mass_transfer_term is not exclusively membrane transfer"- so if your control vol is 3-D, membrane in 2-d is not in your c.v.. The mass_transfer_term of the c.v.  is exactly the membrane transfer because it is a 2-d flux in this case.

docs/technical_reference/unit_models/electrolyzer.rst
lbibl: Make sure this is documented if it hasn't been.

docs/technical_reference/unit_models/electrolyzer.rst
hunterbarber: I believe I've seen the same value also named electromotive force potential and electric potential, so I think any of the above are okay. So long as it's clear on what to input, which these values are established for this reaction within marginal deviations.

Also I kept them separate because technically the reactions in the anolyte and catholyte are determined by what the stoichiometry inputs are. Therefore, if a different reaction is occurring on the anode, then it is clear that the cathode potential is still the same, although the anode and overall are now different.

docs/technical_reference/unit_models/electrolyzer.rst
```diff
@@ -0,0 +1,245 @@
+Electrolyzer
+============
+.. note::
+
+  The development of the electrolyzer model is ongoing.
+
+This is a simplified electrolyzer unit model used to approximate electrolysis performance.  With the current build, the model is simulated under the following assumptions:
+   * simulation of this unit model is only supported with the Multi-Component Aqueous Solution (MCAS) property package
```
adam-a-a: Would be convenient and helpful for the user to provide a hyperlink to MCAS documentation here

docs/technical_reference/unit_models/electrolyzer.rst
```diff
@@ -0,0 +1,245 @@
...
+   * supports liquid phase only, vapor components are modeled in the liquid phase
+   * supports steady-state only
+   * assumes isothermal conditions and performance is not temperature dependent
```
adam-a-a:
```suggestion
   * assumes isothermal conditions and performance is not temperature-dependent
```

docs/technical_reference/unit_models/electrolyzer.rst
adam-a-a: I am realizing now that other unit models are also missing this, but I think it'd be beneficial to include the constraint name as a column for the table for users to more easily compare the documentation with model code (similar to how we provide variable_name in the variables table).

docs/technical_reference/unit_models/electrolyzer.rst
```diff
@@ -0,0 +1,245 @@
...
+
+   "membrane capital cost", ":math:`C_{membrane}`", "membrane_cost", ":math:`$`"
+   "anode capital cost", ":math:`C_{anode}`", "anode_cost", ":math:`$`"
+   "cathode capital cost", ":math:`C_{cathode}`", "cathode_cost", ":math:`$`"
+   "membrane replacement cost", ":math:`C_{memreplace}`", "membrane_replacement_cost", ":math:`\frac{$}{yr}`"
+
+Capital Cost Calculations
++++++++++++++++++++++++++
+
+Capital costs are contributing to the majority of material costs for the anode cathode and membrane. Each material cost is calculated individually then summed (O’Brien, 2005).
```
adam-a-a:
```suggestion
Capital costs contribute to the majority of material costs for the anode cathode and membrane. Each material cost is calculated individually then summed (O’Brien, 2005).
```

docs/technical_reference/unit_models/electrolyzer.rst
```diff
@@ -0,0 +1,245 @@
...
+
+        & C_{membrane} = c_{membrane}A_{membrane} \\\\
+        & C_{anode} = c_{anode}A_{anode} \\\\
+        & C_{cathode} = c_{cathode}A_{cathode} \\\\
+        & C_{cap,total} = \frac{C_{membrane}+C_{anode}+C_{cathode}}{f_{memreplace}}
+
+Operating Cost Calculations
++++++++++++++++++++++++++++
+
+Operating costs for the electrolyzer are the electricity requirements and membrane replacement costs. Electricity is costed using ``cost_flow`` applied to the ``power`` variable on the unit model. Currently, replacement costs for the anode and cathode are not considered in the costing function.
```
adam-a-a, line 227: Thinking about this a little more, if we cost the flow by power, which I presume is in `kW`, would this be accurate if we were to encounter a non-hourly timescale? What I mean is, at an hourly temporal resolution, power (kW) and energy (kWh) would match, but that wouldn't hold true at other temporal scales.

watertap/costing/units/electrolyzer.py
```diff
@@ -0,0 +1,139 @@
...
+        == pyo.units.convert(
+            blk.costing_package.electrolyzer.factor_membrane_replacement
+            * blk.costing_package.electrolyzer.membrane_unit_cost
+            * blk.unit_model.membrane_area,
+            to_units=blk.costing_package.base_currency
+            / blk.costing_package.base_period,
+        )
+    )
+    blk.costing_package.cost_flow(
+        pyo.units.convert(blk.unit_model.power, to_units=pyo.units.kW),
```
adam-a-a, line 134: The other thing to consider is that electricity costs are typically in units of `$/kWh` so for the units to work out, if we want to be precise about it, electricity should technically be in units of kWh.

watertap/unit_models/electrolyzer.py
```diff
@@ -0,0 +1,853 @@
...
+    CONFIG = ConfigBlock()
+
+    CONFIG.declare(
+        "dynamic",
+        ConfigValue(
+            domain=In([False]),
+            default=False,
+            description="Dynamic model flag - must be False",
+            doc="""Indicates whether this model will be dynamic or not,
+    **default** = False. The filtration unit does not support dynamic
```
adam-a-a:
```suggestion
    **default** = False. The electrolyzer unit does not support dynamic
```

watertap/unit_models/electrolyzer.py
```diff
@@ -0,0 +1,853 @@
...
+        ),
+    )
+    CONFIG.declare(
+        "has_holdup",
+        ConfigValue(
+            default=False,
+            domain=In([False]),
+            description="Holdup construction flag - must be False",
+            doc="""Indicates whether holdup terms should be constructed or not.
+    **default** - False. The filtration unit does not have defined volume, thus
```
adam-a-a:
```suggestion
    **default** - False. The electrolyzer unit does not have defined volume, thus
```

watertap/costing/units/electrolyzer.py
hunterbarber: @adam-a-a I'm trying to track down where the units could potentially already be handled. Like...

```
     @self.Constraint(doc="Aggregation constraint for fixed O&M costs")
        def aggregate_fixed_operating_cost_constraint(blk):
            e = 0
            for u in self._registered_unit_costing:
                # Allow for units that might only have a subset of cost Vars
                if hasattr(u, "fixed_operating_cost"):
                    e += pyo.units.convert(
                        u.fixed_operating_cost, to_units=c_units / t_units
                    )
```

...wouldn't ($any/kWh)(kW) to units ($base/yr) be handled automatically?

adam-a-a: You're right- I went back as well and was reminded of how "defined flows" are accounted for in the IDAES costing framework. I made the point in the PR related to this one, which was that this is sufficient for fixed load operation, and that's all we're considering for the time being. Thus, the calculation to arrive at $/hour of electricity cost and converting to annual basis works out--it's just that in practice, you'd multiply price by the amount of energy consumed in given period of time, which is why I hesitated.

In summary, this is good to go.

docs/technical_reference/unit_models/electrolyzer.rst
hunterbarber: @adam-a-a I would say maybe make it an issue. I don't want to change it here and be inconsistent with all other documentation. If this were to become a convention across unit models, maybe it would be best to have a helper function or harness to start these tables, and then we can do it for all unit models instead of this one.

adam-a-a: @hunterbarber sounds completely reasonable. I had some sort of helper function to do this for all our zero-order models, and there is a module in IDAES that could be used to display expressions in latex, if we wanted to go a step further.

adam-a-a (approved):
LGTM

adam-a-a:
@lbibl I think any subsequent issues can be addressed in one or more follow-up PRs. Do you agree?

hunterbarber:
> @lbibl I think any subsequent issues can be addressed in one or more follow-up PRs. Do you agree?

And once the analysis starts, there's bound to be some need for edits.

lbibl (approved):

Issue #950, opened 2023-03-07
Diffusivity warning always reported even when on-demand diffusivity property is not constructed


hunterbarber:
### Description

Diffusivity warning always reported even when on-demand diffusivity property is not constructed. The following error is always logging,

`            # warning for components with no diffusivity_data entry
            if not missing_diffus_ind == []:
                _log.warning(
                    f"Diffusivity data was not provided for {missing_diffus_ind}. "
                )
`

### Expected Behavior

Only log warning when diffusivity data is called, `self.config.diffus_calculation == DiffusivityCalculation.none`, AND `if not missing_diffus_ind == []`

hunterbarber:
Took this off the March release because it's lower priority than other items for March. but should be a small fix

hunterbarber:
Note to also look at the density parameters when looking at the diffusivity parameters. These are persistently constructed even when unnecessary.

hunterbarber:
logging the warnings
```
[WARNING] idaes.watertap.property_models.multicomp_aq_sol_prop_pack: Diffusivity data was not provided for ['DCE'].
```
even when diffusivity isn't constructed on-demand, and
```
[WARNING] idaes.watertap.core.util.scaling: If there was a property constraint written for the variable, 
fs.properties.diffus_phase_comp, that constraint was not scaled. The transform_property_constraints tool expects
constraints to have the following naming convention: 'eq_' + 'diffus_phase_comp'. This suggests that the user may have
defined a property in metadata but failed to follow the naming convention for its constraint. If there is no property 
constraint associated with the diffus_phase_comp, this warning can be ignored.
```
This is obviously a precautionary warning but is reported even when `diffus_phase_comp` isn't constructed and for each state block

hunterbarber:
> logging the warnings
> 
> ```
> [WARNING] idaes.watertap.property_models.multicomp_aq_sol_prop_pack: Diffusivity data was not provided for ['DCE'].
> ```
> 
> even when diffusivity isn't constructed on-demand, and
> 
> ```
> [WARNING] idaes.watertap.core.util.scaling: If there was a property constraint written for the variable, 
> fs.properties.diffus_phase_comp, that constraint was not scaled. The transform_property_constraints tool expects
> constraints to have the following naming convention: 'eq_' + 'diffus_phase_comp'. This suggests that the user may have
> defined a property in metadata but failed to follow the naming convention for its constraint. If there is no property 
[...]

PR #1007, opened 2023-04-17
Removing unwanted code from `test_gac.py`

hunterbarber:
## Fixes/Resolves:
Found a duplicated line of code and a print statement in the `test_gac.py` file. Corrected this and made other code formatting changes. All formatting changes are solely for readability and do not change the functionality at all.

## Summary/Motivation:
Code cleanup

- Remove duplicated statements
- Remove print statement
- General cleanup to reduce lines of code and improve readability with black
- Consolidate import list

### Legal Acknowledgement

### DIFF
--- a/watertap/unit_models/tests/test_gac.py
+++ b/watertap/unit_models/tests/test_gac.py
@@ -184,7 +184,7 @@ class TestGACSimplified:
 
         # Approx data pulled from graph in Hand, 1984 at ~30 days
         # 30 days adjusted to actual solution to account for web plot data extraction error within reason
-        # pyo.values calculated by hand and match those reported in Hand, 1984
+        # values calculated by hand and match those reported in Hand, 1984
         assert pytest.approx(0.0005178, rel=1e-3) == pyo.value(ms.fs.unit.equil_conc)
         assert pytest.approx(19780, rel=1e-3) == pyo.value(ms.fs.unit.dg)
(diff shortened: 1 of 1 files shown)


watertap/unit_models/tests/test_gac.py
```diff
@@ -160,34 +131,30 @@ def test_simplified_build(self, gac_frame_simplified):
...
         for port_str in port_lst:
             port = getattr(ms.fs.unit, port_str)
-            assert len(port.vars) == 3  # number of state variables for property package
+            assert len(port.vars) == 3
             assert isinstance(port, Port)
 
         # test statistics
-        assert number_variables(ms) == 100
-        assert number_total_constraints(ms) == 64
-        assert number_unused_variables(ms) == 11  # dens parameters from properties
```
MarcusHolly, line 169: It might be worth keeping the comment here (and on line 353) since it may be helpful for a developer that works with this code in the future and is wondering why there are so many unused variables.

MarcusHolly (approved):
One small comment/suggestion, but everything else LGTM.

watertap/unit_models/tests/test_gac.py
hunterbarber: Issue #950 is concerning diffusivity parameters (and warnings) that are constructed even when unused in this same MCAS property pack. I'll just add a note to this issue to also look at these density parameters to see if they have to be constructed in the first place and re-add the note then if needed.

adam-a-a (approved):

PR #849, opened 2022-11-17
Mvc example

carson-tucker:
## Fixes/Resolves:

Example MVC flowsheet. 

## Summary/Motivation:
Provide users a sample flowsheet for the MVC model. 

## Changes proposed in this PR:
- MVC flowsheet
- Adds evaporator, compressor, and heat exchanger costing

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #1009, opened 2023-04-19
MVC example flowsheet

carson-tucker:
## Fixes/Resolves:
An example flowsheet for single-stage MVC with preheaters. 

## Summary/Motivation:
Provides users with an example of an MVC flowsheet. 

- Added single-stage MVC flowsheet
- Added costing for evaporator, compressor, and heat exchangers

### Legal Acknowledgement


adam-a-a:
Note to reviewers: the older draft PR for this was #849. Should we close it out?

TimBartholomew (approved):

watertap/examples/flowsheets/mvc/tests/test_mvc_single_stage.py
```diff
@@ -0,0 +1,341 @@
...
+
+        initialize_system(m)
+
+        # mass flows in evaporator
+        assert value(
+            m.fs.evaporator.properties_vapor[0].flow_mass_phase_comp["Vap", "H2O"]
+        ) == pytest.approx(22.222, rel=1e-3)
+        # assert value(
+        #     m.fs.evaporator.properties_vapor[0].flow_mass_phase_comp["Liq", "H2O"]
+        # ) == pytest.approx(1e-8, rel=1e-3)
```
adam-a-a: Commented code- you can delete, or if this gave you trouble, change from relative tolerance to absolute tolerance since this is meant to indicate 0, like so:
```suggestion
        #     m.fs.evaporator.properties_vapor[0].flow_mass_phase_comp["Liq", "H2O"]
        # ) == pytest.approx(1e-8, abs=1e-3)
```

watertap/examples/flowsheets/mvc/tests/test_mvc_single_stage.py
```diff
@@ -0,0 +1,341 @@
...
+        m = mvc_single_stage
+
+        scale_costs(m)
+        fix_outlet_pressures(m)
+
+        assert degrees_of_freedom(m) == 1
+
+        m.fs.objective = Objective(expr=m.fs.Q_ext[0])
+        solver = get_solver()
+        results = solve(m, solver=solver)
```
adam-a-a: You should add an `assert_optimal_termination(results)` after this line, and wherever else you have a solve. This might show up in the tests as the reason for the failures on Linux (i.e., not that there are different solutions between Windows and Linux but rather that on Windows, you had optimal termination, while on Linux, you did not).

adam-a-a (changes requested):
LGTM- but Linux tests are failing. See my comments about checking for optimal termination.

watertap/examples/flowsheets/mvc/mvc_single_stage.py
```diff
@@ -0,0 +1,745 @@
...
+)
+from pyomo.network import Arc
+
+import pyomo.environ as pyo
+import pyomo.util.infeasible as infeas
+from idaes.core import FlowsheetBlock
+from idaes.core.solvers import get_solver
+from idaes.core.util.model_statistics import degrees_of_freedom
+from idaes.core.util.initialization import propagate_state
+from idaes.core.util.model_diagnostics import DegeneracyHunter
```
adam-a-a: We typically remove utilities like this from models/flowsheets before merging into the main codebase.
```suggestion
```

watertap/examples/flowsheets/mvc/mvc_single_stage.py
```diff
@@ -0,0 +1,745 @@
...
+    initialize_system(m)
+    scale_costs(
+        m
+    )  # rescale costs after initialization because scaling depends on flow rates
+    fix_outlet_pressures(m)  # outlet pressure are initially unfixed for initialization
+
+    # set up for minimizing Q_ext in first solve
+    print(
+        "DOF after initialization: ", degrees_of_freedom(m)
+    )  # should be 1 because Q_ext is unfixed
```
adam-a-a: Moving the comment one line above the print line should mitigate the funky formatting by black.
```suggestion
    # should be 1 because Q_ext is unfixed
    print(
        "DOF after initialization: ", degrees_of_freedom(m)
    )  
```

watertap/examples/flowsheets/mvc/mvc_single_stage.py
```diff
@@ -0,0 +1,745 @@
...
+    print("Termination condition: ", results.solver.termination_condition)
+    display_metrics(m)
+    display_design(m)
+
+    print("\n***---Second solve - optimization results---***")
+    m.fs.Q_ext[0].fix(0)  # no longer want external heating in evaporator
+    del m.fs.objective
+    set_up_optimization(m)
+    results = solve(m, solver=solver, tee=False)
+    print("Termination condition: ", results.solver.termination_condition)
```
adam-a-a, line 83: You should probably change to assert optimal termination to more clearly alert the user to a failed solve.

watertap/examples/flowsheets/mvc/mvc_single_stage.py
```diff
@@ -0,0 +1,745 @@
...
+        delta_temperature_callback=delta_temperature_chen_callback,
+        flow_pattern=HeatExchangerFlowPattern.countercurrent,
+    )
+    # Set lower bound of approach temperatures
+    m.fs.hx_distillate.delta_temperature_in.setlb(0)
+    m.fs.hx_distillate.delta_temperature_out.setlb(0)
+    m.fs.hx_distillate.area.setlb(10)
+    add_pressure_drop_to_hx(m.fs.hx_distillate, m.fs.config.time)
+
+    m.fs.hx_brine = HeatExchanger(
```
adam-a-a, line 121: Correct me if I'm wrong, but I believe this is the 0D HX. IDAES also has 1D. My guess is that we don't think it would add much value to go to 1D, but curious if you considered it are experimented with it.

watertap/examples/flowsheets/mvc/mvc_single_stage.py
```diff
@@ -0,0 +1,745 @@
...
+        flowsheet_costing_block=m.fs.costing
+    )
+
+    m.fs.costing.cost_process()
+    m.fs.costing.add_annual_water_production(m.fs.distillate.properties[0].flow_vol)
+    m.fs.costing.add_LCOW(m.fs.distillate.properties[0].flow_vol)
+    m.fs.costing.add_specific_energy_consumption(m.fs.distillate.properties[0].flow_vol)
+
+
+def add_pressure_drop_to_hx(hx_blk, time_point):
```
adam-a-a: Interesting- I wonder if IDAES should include pressure drop in HX units.

andrewlee94: @adam-a-a IDAES does support pressure change in HX units - each side has a `has_pressure_change` option.

watertap/examples/flowsheets/mvc/mvc_single_stage.py
```diff
@@ -0,0 +1,745 @@
...
+        == hx_blk.cold.properties_out[0].pressure + hx_blk.cold.deltaP[0]
+    )
+    hx_blk.hot.pressure_drop = Constraint(
+        expr=hx_blk.hot.properties_in[0].pressure
+        == hx_blk.hot.properties_out[0].pressure + hx_blk.hot.deltaP[0]
+    )
+
+
+def set_operating_conditions(m):
+    m.fs.costing.base_currency = pyo.units.USD_2020
```
adam-a-a: This line probably belongs in `add_costing`. Also, I think 2020 USD might be a little skewed due to the pandemic so considering an alternate year might be of interest for analysis purposes.

watertap/examples/flowsheets/mvc/mvc_single_stage.py
adam-a-a: Ah, true. I had quickly jumped to the 0D HX and searched for "has_pressure" or "deltaP" and nothing turned up, but the control volumes are set up in the `heater` unit, which is where you'll find the `has_pressure_change` option and also see that `deltaP` is accounted for in the model. 

So then- is there a good reason to have this `add_pressure_drop_to_hx` that I'm missing?

TimBartholomew:
@bknueven Carson is having a hard time addressing the linux test failures because they pass on her laptop. Usually when we see this I believe the problem is that the model is not stable and either needs to be scaled better or reformulated. However, I'm confused because this model has been used to generate results for a paper with large parameter sweeps and it reliably solves on windows.

Do you have any recommendations for her? Do you think you could take a look?

watertap/examples/flowsheets/mvc/mvc_single_stage.py
carson-tucker: Thank you for pointing this out. I added the config argument for "has_pressure_change" and removed the "add_pressure_drop_to_hx" function.

adam-a-a (approved):

PR #1010, opened 2023-04-20
Dewatering and thickener unit models BSM2

agarciadiego:
## Summary/Motivation:
Unit model necessary for BSM2 flowsheet. Addresses point in Issue #934 

- Include dewatering unit model
- Include thickener unit model

### Legal Acknowledgement


MarcusHolly (approved):
LGTM - I will add the documentation under a separate PR.

watertap/unit_models/dewatering.py
```diff
@@ -0,0 +1,187 @@
...
+# Enumerate options for balances
+class SplittingType(Enum):
+    """
+    Enum of supported material split types.
+    """
+
+    totalFlow = 1
+    phaseFlow = 2
+    componentFlow = 3
+    phaseComponentFlow = 4
```
adam-a-a: I think we can import `SplittingType` from IDAES. Any reason not to?

watertap/unit_models/dewatering.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+@declare_process_block_class("Dewatering_Unit")
+class DewateringUnit(SeparatorData):
+    """
+    Dewatering unit block for BSM2
+    """
+
```
adam-a-a, line 54: We can do this and avoid having the user even having to deal with these arguments
```suggestion
    CONFIG = SeparatorData.CONFIG()
    CONFIG.outlet_list = ["underflow", "overflow"]
    CONFIG.split_basis=SplittingType.componentFlow
    
```

watertap/unit_models/dewatering.py
```diff
@@ -0,0 +1,187 @@
...
+        # Call UnitModel.build to setup dynamics
+        super(DewateringUnit, self).build()
+
+        if "underflow" and "overflow" not in self.config.outlet_list:
+            raise ConfigurationError(
+                "{} encountered unrecognised "
+                "outlet_list. This should not "
+                "occur - please use overflow "
+                "and underflow as outlets.".format(self.name)
+            )
```
adam-a-a, line 77: I suppose we could keep this if we incorporate my change above, in case an advanced user knew they could modify the config's outlet list to whatever they want. This would catch that case.

watertap/unit_models/dewatering.py
```diff
@@ -0,0 +1,187 @@
...
+
+        self.p_dewat = Param(
```
adam-a-a, line 79: I would shoot for more elaborate naming schemes. My guess without referring back to the literature is that you went with matching symbols in the literature. While that seems reasonable for the popular ASM and ADM models, I wonder if it might be more useful to offer up our own names in these unit cases for improved clarity. This is a broad comment regarding any parameter/variable names, when I compare those names with the associated doc string and question their alignment.

watertap/unit_models/dewatering.py
```diff
@@ -0,0 +1,187 @@
...
+        @self.Constraint(
+            self.flowsheet().time,
+            self.non_particulate_components,
+            doc="soluble fraction",
+        )
+        def non_particulate_components(blk, t, i):
+            return blk.split_fraction[t, "overflow", i] == 1 - blk.f_q_du[t]
+
+    def _get_performance_contents(self, time_point=0):
+        if hasattr(self, "split_fraction"):
```
adam-a-a: Is there a case where `split_fraction` wouldn't be an attribute?

watertap/unit_models/tests/test_thickener_unit.py
```diff
@@ -0,0 +1,279 @@
...
+    @pytest.mark.solver
+    @pytest.mark.skipif(solver is None, reason="Solver not available")
+    @pytest.mark.component
+    def test_conservation(self, tu):
+        assert (
+            abs(
+                value(
+                    tu.fs.unit.inlet.flow_vol[0] * tu.fs.props.dens_mass
+                    - tu.fs.unit.overflow.flow_vol[0] * tu.fs.props.dens_mass
+                    - tu.fs.unit.underflow.flow_vol[0] * tu.fs.props.dens_mass
```
adam-a-a, line 255: should we also look at Q*C balances? In reality, the underflow density would be different than the overflow although I don't know if we are given any expressions to approximate anything better than a single assumed solution density equal to that of water.

watertap/unit_models/thickener.py
```diff
@@ -0,0 +1,187 @@
...
+class SplittingType(Enum):
+    """
+    Enum of supported material split types.
+    """
+
+    totalFlow = 1
+    phaseFlow = 2
+    componentFlow = 3
+    phaseComponentFlow = 4
+
```
adam-a-a: Same as above -should be able to just import SplittingType from IDAES.

watertap/unit_models/thickener.py
```diff
@@ -0,0 +1,187 @@
...
+
+@declare_process_block_class("Thickener_Unit")
+class ThickenerUnit(SeparatorData):
+    """
+    Thickener unit block for BSM2
+    """
+
```
adam-a-a, line 55: Can add the config settings her as I suggested for dewatering unit

watertap/unit_models/tests/test_thickener_unit.py
agarciadiego: I agree, however the property package is built with the assumption that the density is the one ok water

watertap/unit_models/tests/test_dewatering_unit.py
```diff
@@ -0,0 +1,271 @@
...
+    def du(self):
+        m = ConcreteModel()
+        m.fs = FlowsheetBlock(dynamic=False)
+
+        m.fs.props = ASM1ParameterBlock()
+
+        m.fs.unit = Dewatering_Unit(
+            property_package=m.fs.props,
+            outlet_list=["underflow", "overflow"],
+            split_basis=SplittingType.componentFlow,
```
adam-a-a: I think with my suggested changes, these would no longer require specification

watertap/unit_models/tests/test_dewatering_unit.py
```diff
@@ -0,0 +1,271 @@
...
+        )
```
adam-a-a:
```suggestion
        m.fs.unit = Dewatering_Unit(
            property_package=m.fs.props,
        )
```

watertap/unit_models/dewatering.py
agarciadiego: I think if we are following ASM and ADM convention for names of our property, reaction package, etc. following naming convention from BSM2 paper would avoid confusion for users specially if they are following the paper we added as references.

watertap/unit_models/tests/test_thickener_unit.py
```diff
@@ -0,0 +1,283 @@
...
+# National Renewable Energy Laboratory, and National Energy Technology
+# Laboratory (subject to receipt of any required approvals from the U.S. Dept.
+# of Energy). All rights reserved.
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and license
+# information, respectively. These files are also available online at the URL
+# "https://github.com/watertap-org/watertap/"
+#################################################################################
+"""
+Tests for thickener unnit example.
```
MarcusHolly:
```suggestion
Tests for thickener unit example.
```
The same typo exists in the dewatering test file.

watertap/unit_models/dewatering.py
```diff
@@ -0,0 +1,177 @@
...
+)
+
+__author__ = "Alejandro Garciadiego"
+
+
+# Set up logger
+_log = idaeslog.getLogger(__name__)
+
+
+@declare_process_block_class("Dewatering_Unit")
```
adam-a-a:
```suggestion
@declare_process_block_class("DewateringUnit")
```

adam-a-a: Splitting hairs, but this keeps naming more consistent with other models.

watertap/unit_models/thickener.py
```diff
@@ -0,0 +1,180 @@
...
+)
+
+__author__ = "Alejandro Garciadiego"
+
+
+# Set up logger
+_log = idaeslog.getLogger(__name__)
+
+
+@declare_process_block_class("Thickener_Unit")
```
adam-a-a:
```suggestion
@declare_process_block_class("Thickener")
```

watertap/unit_models/dewatering.py
adam-a-a: Fair enough - the documentation will probably help things.

watertap/unit_models/dewatering.py
```diff
@@ -0,0 +1,177 @@
...
+            return blk.split_fraction[t, "overflow", i] == 1 - blk.TSS_rem
+
+        @self.Constraint(
+            self.flowsheet().time,
+            self.non_particulate_components,
+            doc="soluble fraction",
+        )
+        def non_particulate_components(blk, t, i):
+            return blk.split_fraction[t, "overflow", i] == 1 - blk.f_q_du[t]
+
```
adam-a-a, line 143: So, for both units, I'm aware that you didn't add `initialize_build` and `calcualte_scaling_factors` methods, likely because the separator handles those. I still wonder though if we should include them here anyway, even if, for example, we just have loggers setup for initialization and run `super().calculate_scaling_factors()`. @andrewlee94 curious to hear your thoughts.

andrewlee94: If you inherit from another model or base class, then any methods you do not overload will use that of the parent. For initialization this might be OK as long as the additions in the new model are relatively simple. However, for scaling, the new model is adding new variables and constraints (otherwise it wouldn't be a new model), and the parent's scaling methods won't touch these leaving part of the model unscaled.

On a related note, be aware that IDAES plans to rebuild both initialization and scaling this year. The new scaling API is already in place if you want to start looking at it: https://idaes-pse.readthedocs.io/en/latest/reference_guides/initialization/index.html

adam-a-a: Right, I figured we could technically get away with it since the Separator's initialization and calculate_scaling_factors are essentially being used here, but that it might just be good practice and make for better readability/consistency between models for [new] developers. Furthermore, there are probably some unscaled vars that are added here but simple enough that additional scaling doesn't seem necessary at this point. I think we could put a pin in it for now and revisit if issues arise when analyzing the BSM2 flowsheet later.

And thanks for the note on the new scaling API.

watertap/unit_models/tests/test_dewatering_unit.py
```diff
@@ -0,0 +1,276 @@
...
+    @pytest.mark.component
+    def test_initialize(self, du):
+
+        iscale.calculate_scaling_factors(du)
+        initialization_tester(du)
+
+    @pytest.mark.solver
+    @pytest.mark.skipif(solver is None, reason="Solver not available")
+    @pytest.mark.component
+    def test_solve(self, du):
```
adam-a-a, line 191: There should typically be a test to check for badly scaled variables here. I checked, and a few will crop up. We can be lenient about it and make a note to revisit in our master issue, considering the whole effort here for the BSM2 flowsheet, or you could add the typical test to check for badly scaled vars and eliminate any that exist by adding scaling. Though the badly scaled vars are probably attributed to zero components, it's hard to say whether this will end up impacting model stability at the flowsheet level later on.

adam-a-a: Actually, through our off-GH discussion, you made it clear that the badly scaled vars stem from the lack of scaling on the ASM1 property model, which we have noted in our master issue (#934)

adam-a-a (approved):
LGTM

PR #1015, opened 2023-04-26
Consolidating parameters in GAC costing model

### LINKED ISSUE #975: Consolidate amount of costing parameter variables in GAC costing
### Description

With the new configuration option in #960, costing parameters of the unused option are constructed but unused. 

### Motivation

Effort to provide concise model code and reporting in `.display()`.

### Possible Implementation

- Index `Var()` based on config options since equations are of the same form
- If conditional to construct `Var()` based on which config option is selected

### Additional Context

_No response_

hunterbarber:
## Fixes/Resolves:
In response to issue #975

## Summary/Motivation:
The original costing model constructed 20 costing parameters, 11 of which were used in calculations dependent on the configuration argument. All parameters had their own Pyomo variables are therefore cluttered the code and reporting in `.display()`.

- Create utility `_unpack_data_dict` to transform nested dictionaries into readable initialization objects by `pyo.Var()` (currently located within `watertap/costing/units/gac.py`)
- Convert individual Pyomo variables to be more concise
- With the given implementation, the need for rules based on configuration options are unnecessary since the variables are indexed by the configuration argument, therefore remove these

### Legal Acknowledgement


hunterbarber:
@adam-a-a Take a look and let me know what you think for this formatting of parameters before I update the documentation and request reviews

watertap/costing/units/gac.py
```diff
@@ -22,142 +22,99 @@
 
 
 class ContactorType(StrEnum):
-    pressure = "pressure"
-    gravity = "gravity"
+    pressure = "pres"
+    gravity = "grav"
```
bknueven: Why make this change? I think it was better before ...

watertap/unit_models/tests/test_gac.py
```diff
@@ -418,7 +419,7 @@ def test_robust_costing_gravity(self, gac_frame_robust):
 
         mr_grav.fs.unit.costing = UnitModelCostingBlock(
             flowsheet_costing_block=mr_grav.fs.costing,
-            costing_method_arguments={"contactor_type": "gravity"},
+            costing_method_arguments={"contactor_type": ContactorType.gravity},
```
bknueven: The whole point of the `StrEnum` is you don't have to import these enums everywhere to use them. As an example I like the prior code better.

bknueven (commented):
I have a few minor comments. I would like to merge this soon-ish so I can finally finish #676.

watertap/unit_models/tests/test_gac.py
hunterbarber: This is fine to change back, I made it just to have my tests passing as I was playing around with changing names of the `StrEnum` options.

watertap/costing/units/gac.py
hunterbarber: I made this change just to have equal string lengths in the main code as I was building dictionaries and it would make it visually consistent. Bit of an extreme formatting thing. This also would be fine to change back.

adam-a-a: I'd prefer to change it back too for the sake of clarity.

hunterbarber:
> I have a few minor comments. I would like to merge this soon-ish so I can finally finish #676.

@bknueven If you think this is a formatting improvement this can come out of draft, I'll revert those naming changes as well. The question I have is is this the best option of these two to consolidate unused costing parameters in this case?

1. (Current change in this PR) All the Pyomo variables that are constructed are used, however these variables are now indexed where only the index corresponding to the specified configuration option is used. Technically, the other index of the variable still contains unused parameters, they just don't overflow the reporting.
2. Add a conditional to only construct the variables when the configuration option is selected. This would increase the code length slightly (from the current change in this PR) but it would remove the unused parameters entirely.

bknueven:
> > I have a few minor comments. I would like to merge this soon-ish so I can finally finish #676.
> 
> @bknueven If you think this is a formatting improvement this can come out of draft, I'll revert those naming changes as well. The question I have is is this the best option of these two to consolidate unused costing parameters in this case?
> 
>     1. (Current change in this PR) All the Pyomo variables that are constructed are used, however these variables are now indexed where only the index corresponding to the specified configuration option is used. Technically, the other index of the variable still contains unused parameters, they just don't overflow the reporting.
> 
>     2. Add a conditional to only construct the variables when the configuration option is selected. This would increase the code length slightly (from the current change in this PR) but it would remove the unused parameters entirely.

Thanks for the explanation. I prefer option (2) for clarity when printing the model and examining it in code. IDAES/WaterTAP conditionally build variables / constraints all the time and option (2) conforms better to that convention.

hunterbarber:
> > > I have a few minor comments. I would like to merge this soon-ish so I can finally finish #676.
> > 
> > 
> > @bknueven If you think this is a formatting improvement this can come out of draft, I'll revert those naming changes as well. The question I have is is this the best option of these two to consolidate unused costing parameters in this case?
> > ```
> > 1. (Current change in this PR) All the Pyomo variables that are constructed are used, however these variables are now indexed where only the index corresponding to the specified configuration option is used. Technically, the other index of the variable still contains unused parameters, they just don't overflow the reporting.
> > 
> > 2. Add a conditional to only construct the variables when the configuration option is selected. This would increase the code length slightly (from the current change in this PR) but it would remove the unused parameters entirely.
> > ```
> 
> Thanks for the explanation. I prefer option (2) for clarity when printing the model and examining it in code. IDAES/WaterTAP conditionally build variables / constraints all the time and option (2) conforms better to that convention.

Sure, I'll implement that and combine with the existing improvements. I'm out the next two work days, but it can be done next week.

watertap/costing/units/gac.py
```diff
@@ -22,142 +22,99 @@
...
+    energy_consumption_coeff_data = {
+        "pres": {0: 8.09926e-4, 1: 8.70577e-4, 2: 0},
+        "grav": {0: 0.123782, 1: 0.132403, 2: -1.41512e-5},
+    }
+
+    blk.contactor_cost_coeff = pyo.Var(
+        blk.contactor_type_list,
+        range(4),
+        initialize=_unpack_data_dict(contactor_cost_coeff_data),
+        units=pyo.units.dimensionless,  # USD_2020 embedded in equation
```
adam-a-a: The USD_2020 comment can go above to avoid any funky black formatting that may arise

watertap/costing/units/gac.py
```diff
@@ -22,142 +22,99 @@
...
+        "grav": {0: 38846.9, 1: 0.490571},
+    }
+    energy_consumption_coeff_data = {
+        "pres": {0: 8.09926e-4, 1: 8.70577e-4, 2: 0},
+        "grav": {0: 0.123782, 1: 0.132403, 2: -1.41512e-5},
+    }
+
+    blk.contactor_cost_coeff = pyo.Var(
+        blk.contactor_type_list,
+        range(4),
```
adam-a-a: Maybe this should be in terms of the length of the associated dictionary.

hunterbarber:
> > > > I have a few minor comments. I would like to merge this soon-ish so I can finally finish #676.
> > > 
> > > 
> > > @bknueven If you think this is a formatting improvement this can come out of draft, I'll revert those naming changes as well. The question I have is is this the best option of these two to consolidate unused costing parameters in this case?
> > > ```
> > > 1. (Current change in this PR) All the Pyomo variables that are constructed are used, however these variables are now indexed where only the index corresponding to the specified configuration option is used. Technically, the other index of the variable still contains unused parameters, they just don't overflow the reporting.
> > > 
> > > 2. Add a conditional to only construct the variables when the configuration option is selected. This would increase the code length slightly (from the current change in this PR) but it would remove the unused parameters entirely.
> > > ```
> > 
> > 
> > Thanks for the explanation. I prefer option (2) for clarity when printing the model and examining it in code. IDAES/WaterTAP conditionally build variables / constraints all the time and option (2) conforms better to that convention.
> 
> Sure, I'll implement that and combine with the existing improvements. I'm out the next two work days, but it can be done next week.

@bknueven @adam-a-a  The current file conforms to not constructing any unused variables for all indices. Had to do a little niche method in `cost_gac` to have it working as intended, but the in-line comments should cover the reasoning. I double-checked the variables built on the model and the tests, so everything should be good as long as this niche method isn't flawed in some way. Just let me know if you find something. (With the okay on this method I also have to change the docs before its merged.)

watertap/costing/units/gac.py
```diff
@@ -178,6 +123,50 @@ def cost_gac(blk, contactor_type=ContactorType.pressure):
...
+        blk.costing_package.gac.other_cost_param,
+        blk.costing_package.gac.energy_consumption_coeff,
+    ]
+
+    # refix variables to appropriate costing parameters
+    for index in range(len(cost_param_var_list)):
+        param_data = cost_param_data_list[index]
+        param_var = cost_param_var_list[index]
+        for param_index in param_var.keys():
+            param_var[param_index].fix(param_data[param_index])
```
bknueven: It's a small thing, but you could use `ComponentMap` here to map the var to the coeff data.

Alternatively, you could create a similar `tuple`, e.g., `[( blk.costing_package.gac.adsorbent_unit_cost_coeff, adsorbent_unit_cost_coeff_data), ...]`.

Or, we could use `zip` to iterate over the list simultaneously. This is my least favorite option. Ideally the var would be close to the coeff_data in the code.

watertap/costing/units/gac.py
```diff
@@ -28,136 +28,81 @@ class ContactorType(StrEnum):
...
-        initialize=0.552207,
+
+    # ---------------------------------------------------------------------
+    # correlation parameter data
+
+    # dummy data that is refixed in cost_gac based on the ContactorType
+    contactor_cost_coeff_dummy = {0: 10000, 1: 1000, 2: -10, 3: 0.1}
+    adsorbent_unit_cost_coeff_dummy = {0: 1, 1: -1e-5}
+    other_cost_param_dummy = {0: 10000, 1: 0.1}
+    energy_consumption_coeff_dummy = {0: 1e-3, 1: 1e-3, 2: 0}
```
bknueven: I would consider moving these temporary dictionaries close to where they're used.

watertap/costing/units/gac.py
```diff
@@ -28,136 +28,81 @@ class ContactorType(StrEnum):
...
+
+    # USD_2020 embedded in equation
+    blk.contactor_cost_coeff = pyo.Var(
+        range(len(contactor_cost_coeff_dummy)),
```
bknueven: You can just pass in the dictionary directly, or use `.keys()` if you want to be explicit, to set the indices. E.g.,
```suggestion
        contactor_cost_coeff_dummy,
```

bknueven (approved):
Looks good; I have a few Python stylistic suggestions

watertap/costing/units/gac.py
hunterbarber: I thought the same thing originally with the tuple, but figured it wouldn't actually consolidate anything because black would end up formatting it in a similar way. I will look into the `ComponentMap` briefly though.

bknueven: Following up from above. Maybe this is better than the `ComponentMap` idea because it's temporary.

```suggestion
    # iterable matching coeff_data to vars
    cost_params = (
        (adsorbent_unit_cost_coeff_data, blk.costing_package.gac.adsorbent_unit_cost_coeff),
        (contactor_cost_coeff_data, blk.costing_package.gac.contactor_cost_coeff),
        (other_cost_param_data, blk.costing_package.gac.other_cost_param),
        (energy_consumption_coeff_data, blk.costing_package.gac.energy_consumption_coeff),
    )

    # refix variables to appropriate costing parameters
    for data, var in cost_params:
        for index, var in param_var.items():
            var.fix(data[index])
```

bknueven: The point with the `tuples` is that it's easier to catch mistakes because the two pieces of information are together in the code. Same with `ComponentMap`.

hunterbarber:
> Looks good; I have a few Python stylistic suggestions

@bknueven Made sense, and I applied those suggestions in the latest commit, thanks. As soon as these docs to reflect these changes are passing (there's been GitHub issues today so it may wait on that) this should be merged.

docs/technical_reference/unit_models/gac.rst
```diff
@@ -245,33 +245,27 @@ The following parameters are constructed when applying the GAC costing method in
...
-   "GAC adsorbent exponential cost parameter coefficient (U.S. EPA, 2021)", ":math:`y_1`", "adsorbent_unit_cost_exp_coeff ", "-1.25311e-5", ":math:`kg^{-1}`"
-   "GAC pressure other cost power law coefficient", ":math:`z_{p0}`", "pres_other_cost_coeff", "16660.7", ":math:`$/\left( m^3 \right)^{z_1}`"
-   "GAC pressure other cost power law exponent", ":math:`z_{p1}`", "pres_other_cost_exp", "0.552207", ":math:`\text{dimensionless}`"
-   "GAC gravity other cost power law coefficient", ":math:`z_{g0}`", "grav_other_cost_coeff", "38846.9", ":math:`$/\left( m^3 \right)^{z_1}`"
-   "GAC gravity other cost power law exponent", ":math:`z_{g1}`", "grav_other_cost_exp", "0.490571", ":math:`\text{dimensionless}`"
    "Fraction of spent GAC adsorbent that can be regenerated for reuse", ":math:`f_{regen}`", "regen_frac", "0.70", ":math:`\text{dimensionless}`"
+   "Reference maximum value of GAC mass needed for initial charge where economy of scale no longer discounts the unit price  (U.S. EPA, 2021)", ":math:`M_{GAC}^{ref}`", "bed_mass_gac_max_ref", "18143.7", ":math:`kg`"
```
adam-a-a, line 249: Can you add the linked reference to the EPA source around the end of the doc?

adam-a-a (approved):
LGTM

docs/technical_reference/unit_models/gac.rst
hunterbarber: Sure, just added the link to the end of the ref.

hunterbarber:
@bknueven @adam-a-a Thanks. This is ready to merge.

PR #1016, opened 2023-04-27
Removing unnecessary warnings as a result of `diffus_phase_comp` in MCAS prop pack

### LINKED ISSUE #950: Diffusivity warning always reported even when on-demand diffusivity property is not constructed

hunterbarber:
## Fixes/Resolves:
Addresses #950

## Summary/Motivation:
The warnings logged were not necessary as the variable in question is not even constructed. Since the warnings don't contribute to the pass/fail of tests, it's only really noticed when running` pytest -s -vv` but does not effect test actions.

- Alter MCAS code to only log warnings when necessary

### Legal Acknowledgement

### DIFF
--- a/watertap/property_models/multicomp_aq_sol_prop_pack.py
+++ b/watertap/property_models/multicomp_aq_sol_prop_pack.py
@@ -400,17 +400,6 @@ class MCASParameterData(PhysicalParameterBlock):
         )
 
         if self.config.diffus_calculation == DiffusivityCalculation.none:
-            # Retrieve component string names from user-provided diffusivity_data configuration
-            diffus_data_indices = [i[1] for i in self.config.diffusivity_data.keys()]
-            # identify component string names that were not specified in diffusivity_data
-            missing_diffus_ind = [
-                i for i in self.solute_set if i not in diffus_data_indices
-            ]
-            # warning for components with no diffusivity_data entry
-            if not missing_diffus_ind == []:
-                _log.warning(
-                    f"Diffusivity data was not provided for {missing_diffus_ind}. "
-                )
 
             self.diffus_phase_comp = Var(
                 self.phase_list,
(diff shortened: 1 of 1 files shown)


adam-a-a (approved):
LGTM

hunterbarber:
The `dens_mass_param`s are used in two different property constraints, so it is less straightforward to implement a fix to remove them from the unused parameter list in this PR. This is good to go with this small change only then.

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -1259,20 +1248,19 @@ def _molar_volume_phase_comp(self):
...
+        # Retrieve component string names from molar_volume_data configuration
+        molar_volume_data_indices = {
+            i[1] for i in self.params.config.molar_volume_data.keys()
+        }
+        missing_diffus_ind = [
+            i
+            for i in self.params.solute_set
+            if i not in (molar_volume_data_indices | diffus_data_indices)
+        ]
+
```
lbibl, line 1262: I don't really get why to move these indices outside the method because they were supposed to be useful only under that specific H-L method.

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -400,17 +400,6 @@ def build(self):
...
-            diffus_data_indices = [i[1] for i in self.config.diffusivity_data.keys()]
-            # identify component string names that were not specified in diffusivity_data
-            missing_diffus_ind = [
-                i for i in self.solute_set if i not in diffus_data_indices
-            ]
-            # warning for components with no diffusivity_data entry
-            if not missing_diffus_ind == []:
-                _log.warning(
-                    f"Diffusivity data was not provided for {missing_diffus_ind}. "
-                )
```
lbibl, line 403: The original warnings were for mistaken default values being taken when a user neglect to provide correct diffusivity data.   I understand you are trying to get rid of "unnecessary" warnings when diffusivity is not needed. I am not sure if it worths the risk of not having it though.   Without testing it out,  will `diffus_phase_comp` be constructed for all members in `solute_list` including those without `diffusivity_data`?  it looks so to me because the `diffus_p_c` is indexed by `solute_set` and if that is the case, my recommendation is keeping the warning.  I'd leave the decision for you or others to make.

watertap/property_models/multicomp_aq_sol_prop_pack.py
hunterbarber: Moving the indices are actually consolidating them. If you review the original code, find that the indices are constructed on line 403 (when `self.config.diffus_calculation == DiffusivityCalculation.none`) to log the warning and line 1263 (when `self.config.diffus_calculation == DiffusivityCalculation.HaydukLaudie`) to build the actual H-L calculation.

watertap/property_models/multicomp_aq_sol_prop_pack.py
hunterbarber: The warning is kept exactly in the same implementation, just moved to the method `_diffus_phase_comp`. This ensures the warning only logs when the `diffus_phase_comp` variables are constructed (on demand). It corrected the issue where the warning was stating that diffusivity data was not provided for a flowsheet that the `diffus_phase_comp` is never constructed or used.

hunterbarber:
@lbibl just re-pinging you so you know there are responses for this.

lbibl (approved):

hunterbarber:
@adam-a-a I only re-requested to get this merged, thanks

PR #1020, opened 2023-05-02
ElectroNP ZO model with UnitModelBlockData base

luohezhiming:
## Fixes/Resolves:

Add a electroNP ZO model compatible with ASM2d

### Legal Acknowledgement


watertap/costing/units/electroNP.py
```diff
@@ -0,0 +1,89 @@
...
+                blk.unit_model.MgCl2_flowrate[t0],
+                to_units=pyo.units.kg / pyo.units.hr,
+            ),
+            "magnesium chloride",
+        )
+
+
+def cost_electroNP_capital(blk, HRT, sizing_cost):
+    """
+    Generic function for costing a UV system.
```
MarcusHolly: Should UV be ElectroNP?

watertap/unit_models/electroNP_ZO.py
```diff
@@ -0,0 +1,494 @@
...
+        @self.Constraint(
+            self.flowsheet().time,
+            self.config.property_package.solute_set,
+            doc="Default solute removal equations",
+        )
+        def default_solute_removal_equation(b, t, j):
+            if hasattr(self.config.property_package.solute_set, "S_PO4"):
+                return b.removal_frac_mass_comp[t, "S_PO4"] == 0.98
+            elif hasattr(self.config.property_package.solute_set, "S_NH4"):
+                return b.removal_frac_mass_comp[t, "S_NH4"] == 0.3
```
MarcusHolly: Is it worth considering S_N2 and S_NO3 here as well or are we just assuming that most nitrogen will be present in the form of "Ammonium plus ammonia nitrogen"? I know we discussed this in the last meeting but I forget what we ultimately decided to do (it might've just been this).

luohezhiming: I think using S_NH4 is OK since the data sent by the team uses NH3

MarcusHolly: Alright, then I agree.

watertap/unit_models/electroNP_ZO.py
```diff
@@ -0,0 +1,494 @@
...
+            ].get_material_flow_terms(
+                p, j
+            )
+
+        # Solute concentration of treated stream
+        @self.Constraint(
+            self.flowsheet().time,
+            self.config.property_package.phase_list,
+            self.config.property_package.solute_set,
+            doc="Constraint for solute concentration in treated " "stream.",
```
MarcusHolly:
```suggestion
            doc="Constraint for solute concentration in treated stream.",
```

watertap/unit_models/tests/test_electroNP_ZO.py
```diff
@@ -0,0 +1,263 @@
...
+                    * m.fs.unit.treated.conc_mass_comp[0, j]
+                    - m.fs.unit.byproduct.flow_vol[0]
+                    * m.fs.unit.byproduct.conc_mass_comp[0, j]
+                )
+            )
+
+    @pytest.mark.solver
+    @pytest.mark.skipif(solver is None, reason="Solver not available")
+    @pytest.mark.component
+    def test_solution(self, ElectroNP_frame):
```
MarcusHolly, line 167: Is it worth listing out the solution for the byproduct stream too?

luohezhiming: Since we don't have any byproduct, so I don't think we need to list them all.

MarcusHolly: I thought the byproduct stream would include the S_NH4 and S_PO4 that got removed?

adam-a-a: I agree with @MarcusHolly. The treated stream would be fed back to the headworks of the plant or right before the activated sludge process, while the byproduct stream would comprise the recovered nutrients.

watertap/unit_models/electroNP_ZO.py
```diff
@@ -0,0 +1,494 @@
...
+
+        # Add performance variables
+        self.recovery_frac_mass_H2O = Var(
+            self.flowsheet().time,
+            initialize=0.8,
+            domain=NonNegativeReals,
+            units=pyunits.dimensionless,
+            bounds=(0.0, 1.0000001),
+            doc="Mass recovery fraction of water in the treated stream",
+        )
```
adam-a-a: Now that we're making this its own unit, separate from the ZO model framework, I wonder if we even need this recovery variable here. If included, I'd think the value would be very close to 1.

watertap/unit_models/electroNP_ZO.py
```diff
@@ -0,0 +1,494 @@
...
+            ) + b.properties_byproduct[
+                t
+            ].get_material_flow_terms(
+                "Liq", "H2O"
+            )
+
+        # default water recovery
+        @self.Constraint(self.flowsheet().time, doc="Default water recovery equation")
+        def default_water_recovery_equation(b, t):
+            return b.recovery_frac_mass_H2O[t] == 1
```
adam-a-a: I would just set the recovery_frac_mass_H2O variable to an initial value and then fix the variable within this build instead of having this constraint.

watertap/unit_models/electroNP_ZO.py
```diff
@@ -0,0 +1,494 @@
...
+
+        # Default solute concentration
+        @self.Constraint(
+            self.flowsheet().time,
+            self.config.property_package.solute_set,
+            doc="Default solute removal equations",
+        )
+        def default_solute_removal_equation(b, t, j):
+            if hasattr(self.config.property_package.solute_set, "S_PO4"):
+                return b.removal_frac_mass_comp[t, "S_PO4"] == 0.98
```
adam-a-a: I think I would maybe have parameters defined for the values, then fix removal fractions directly to those parameter values. The user could change the parameter values as desired and have them fixed.

watertap/unit_models/electroNP_ZO.py
```diff
@@ -0,0 +1,494 @@
...
+        def MgCl2_demand(b, t):
+            return b.MgCl2_flowrate[t] == (
+                b.magnesium_chloride_dosage
+                * pyunits.convert(
+                    b.properties_treated[t].get_material_flow_terms("Liq", "S_PO4"),
+                    to_units=pyunits.kg / pyunits.hour,
+                )
+            )
+
+    def initialize(
```
adam-a-a:
```suggestion
    def initialize_build(
```

watertap/unit_models/tests/test_electroNP_ZO.py
```diff
@@ -0,0 +1,263 @@
...
+solver = get_solver()
+
+
+class TestElectroNP:
+    @pytest.fixture(scope="class")
+    def ElectroNP_frame(self):
+        m = ConcreteModel()
+        m.fs = FlowsheetBlock(dynamic=False)
+
+        m.fs.properties = ModifiedASM2dParameterBlock(
```
adam-a-a, line 66: Hmmm.... I just thought about this, and there is a chance, due to how we've compartmentalized the property and reaction models, that some expressions/additional variables/additional constraints implemented in the reaction block _could_ be desired in the unit model (but not sure yet since I didn't think about it deeply). Maybe this isn't an issue yet, and my concern is probably more relevant to our ADM1 formulation rather than ASM2D, but let's keep this in mind since we expect to make additions to the modified ASM2d property/reaction package.

watertap/unit_models/electroNP_ZO.py
adam-a-a: We should probably make a note to go over this with electroN-P team and get their take on whether we should be summing soluble nitrogen (e..g., NH4 + NH2 + NO3 as nitrogen) or if this should suffice.

watertap/unit_models/electroNP_ZO.py
```diff
@@ -0,0 +1,494 @@
...
+
+        for (t, p, j), v in self.solute_removal_equation.items():
+            iscale.constraint_scaling_transform(
+                v,
+                iscale.get_scaling_factor(
+                    self.properties_in[t].get_material_flow_terms(p, j),
+                    default=1,
+                    warning=True,
+                    hint=" for solute removal",
+                ),
```
adam-a-a, line 497: might not need these scaling transforms if we fix the variables and not build as constraints.

watertap/unit_models/tests/test_electroNP_ZO.py
```diff
@@ -0,0 +1,263 @@
...
+        m.fs.unit.inlet.conc_mass_comp[0, "X_MeP"].fix(EPS * units.mg / units.liter)
+        m.fs.unit.inlet.conc_mass_comp[0, "X_TSS"].fix(EPS * units.mg / units.liter)
+        m.fs.unit.inlet.conc_mass_comp[0, "S_K"].fix(EPS * units.mg / units.liter)
+        m.fs.unit.inlet.conc_mass_comp[0, "S_Mg"].fix(EPS * units.mg / units.liter)
+
+        # Alkalinity was givien in mg/L based on C
+        m.fs.unit.inlet.alkalinity[0].fix(61 / 12 * units.mmol / units.liter)
+
+        # Unit option
+        m.fs.unit.energy_electric_flow_mass.fix(0.044 * units.kWh / units.kg)
```
adam-a-a, line 103: Something just came to mind- we can probably later convert this model to not only focus on ZO formulation. Instead, we could consider having config options would would select between fixed performance or predictive performance from surrogates. In ZO config, removal fractions, etc., would just be fixed to variables, while in predictive mode, the constraint would be activated and the surrogate would be inserted there. Just food for thought.

watertap/unit_models/electroNP_ZO.py
luohezhiming: I choose to fix the recovery to 1.

watertap/unit_models/electroNP_ZO.py
luohezhiming: Yeah, modified.

watertap/unit_models/electroNP_ZO.py
luohezhiming: I add `removal_P` and `removal_N` as two mutable parameters.

watertap/unit_models/tests/test_electroNP_ZO.py
luohezhiming: I added byproduct test for `S_PO4` and `S_NH4`, but to be noticed, since we set the water recovery as 1, so the byproduct stream is assigned to a calculation error value extremely close to 0. So the byproduct concentration will be extremely high and does not make much sense.

watertap/unit_models/tests/test_electroNP_ZO.py
luohezhiming: Yeah, this is a good point, some additional variables can be added to the unit model instead of properties model. Just for modified ASM2D package, we might need to keep the additonal_solute config for now.

watertap/unit_models/electroNP_ZO.py
luohezhiming: Added a note to remind us after we talk with electroN-P team.

watertap/unit_models/electroNP_ZO.py
luohezhiming: I still keep the variables for now.

watertap/unit_models/tests/test_electroNP_ZO.py
luohezhiming: It is a good idea, we could upgrade this model to be a surrogate modle with config option to choose, and I believe I can save a lot of time for building some basic vairables. I still name it as electroNP_ZO for now but would consider to rename it when we add the surrogate model.

watertap/unit_models/tests/test_electroNP_ZO.py
MarcusHolly: @luohezhiming Maybe just add an `assert value(m.fs.unit.byproduct.flow_vol[0])` so that it's more clear why the byproduct values are so high?

MarcusHolly (approved):
Made one more small comment, but otherwise LGTM

agarciadiego (approved):

PR #1025, opened 2023-05-11
Move examples out of package

wigging:
## Fixes/Resolves:

Ignore this Keith.

-
-

### Legal Acknowledgement


dangunter:
Hey Gavin, I wouldn't want you to spend too much valuable time on this without more discussion in the larger team. I think we already discussed this and decided to leave the examples where they are. Has this decision changed, and if so, why?

wigging:
@dangunter Examples do not belong in the source directory of a Python package. Period. End of discussion.

dangunter:
@wigging you are not leading this project or its software engineering. So if you would like to end the discussion, that is fine. But the decision is then to keep the examples where they are. period.

PR #1027, opened 2023-05-15
Add code demo from NAWI spring meeting

adam-a-a:
## Summary/Motivation:
Create a code demo via a Jupyter notebook for the NF DSPMDE model.
Also, bring in @avdudchenko's flowsheet example for NF with bypass and export to GUI.

**EDIT: removing flowsheet examples and ui code and only keeping notebook in this PR.

- add jupyter notebook
- incorporate `total_hardness` into MCAS property model
- silence some warnings the popped up in the Jupyter notebook

### Legal Acknowledgement


watertap/examples/flowsheets/nf_dspmde/nf.py
```diff
@@ -0,0 +1,463 @@
...
+    comp_list = stream.properties[0.0].component_list
+    # print("comp_list", comp_list)
+    stream.total_hardness = Var(
+        initialize=0.5,
+        bounds=(0, 10000),
+        domain=NonNegativeReals,
+        units=pyunits.mg / pyunits.kg,
+        doc="Water Hardness as CO3",
+    )
+    stream.total_hardness.unfix()
```
adam-a-a: Why do you need to unfix? After creating the Var, it should already be unfixed. Is there a special reason for this?

watertap/examples/flowsheets/nf_dspmde/nf.py
```diff
@@ -0,0 +1,473 @@
...
+def add_objective(m):
+    if m.find_component("fs.LCOW_objective") is None:
+        m.fs.LCOW_objective = Objective(expr=m.fs.costing.LCOW)
+
+
+def optimize(m, solver, **kwargs):
+    # add_objective(m)
+    print("Optimizing with {} DOFS".format(degrees_of_freedom(m)))
+    result = solver.solve(m, tee=True)
+    return result
```
MichaelPesce:
```suggestion
def optimize(m, solver=None, **kwargs):
    if solver is None:
        solver = get_solver()
    # add_objective(m)
    print("Optimizing with {} DOFS".format(degrees_of_freedom(m)))
    result = solver.solve(m, tee=True)
    return result
```

MichaelPesce: the parameter_sweep function requires this optimize function as an argument, but I'm not sure if I can send in the solver with it. so can we check for solver and initialize it inside this function in the case it's not provided?

watertap/examples/flowsheets/nf_dspmde/nf.py
avdudchenko: No real good reason. Its actually legacy as I copy paste same code to add a var and include the fix line...

watertap/examples/flowsheets/nf_dspmde/nf_with_bypass.py
```diff
@@ -0,0 +1,190 @@
...
+    m.fs.NF.product.initialize(optarg=solver.options)
+    m.fs.NF.retentate.initialize(optarg=solver.options)
+
+
+def unfix_opt_vars(m):
+    nf.unfix_opt_vars(m)
+    m.fs.by_pass_splitter.split_fraction[0, "bypass"].unfix()
+
+
+def optimize(m, solver):
```
MichaelPesce:
```suggestion
def optimize(m, solver=None):
```

watertap/core/util/scaling.py
```diff
@@ -24,20 +24,30 @@ def transform_property_constraints(self):
         var_str = p.name
         if p.method is not None and self.is_property_constructed(var_str):
             var = getattr(self, var_str)
+            msg = (
```
adam-a-a, line 27: @MarcusHolly I modified your utility function to not show logger warnings for cases where a variable is fixed.

adam-a-a:
@avdudchenko @MichaelPesce @dangunter   - the NF with bypass flowsheet is failing the Linux tests which indicates a problem with scaling since Windows tests are passing. 

The main goal of this PR is to merge in the code demo (Jupyter notebook) from the Spring meeting. The NF and NF+bypass flowsheets are really just there for export to the GUI, but the GUI aspect of the demo will require other PRs that bring in everything that was shown in the demo, from what I understand. 

I propose to remove the NF with bypass flowsheet from this PR and make it a separate PR, where the issue can be resolved there. Any additional GUI-related PRs that seek to incorporate demo of the NF with bypass flowsheet would also depend on that PR. Any objections?

avdudchenko:
> 

Lets try to fix the scaling issue. I will take a pass at it tomorrow. if I cant' get it to work we can go from there.

adam-a-a:
> > 
> 
> Lets try to fix the scaling issue. I will take a pass at it tomorrow. if I cant' get it to work we can go from there.

OK- the only obstacle for me was that I have Windows and had trouble trying to resolve on Linux. I do have WSL and tried resolving there, where I did manage to get the flowsheet test to pass locally, but for some odd reason, the checks still fail here. I think my version of Linux may be old, so updating now and going to retry. If you can run on Linux, then would be happy for you to take a shot at it tomorrow.

adam-a-a:
@bknueven I removed what I considered to be the problematic flowsheet and associated files and moved to #1062 . However, not surprisingly, the macos check is failing for the regular old nf flowsheet. 

I can do a couple things.
1) move the nf flowsheet along to that separate PR (#1062)
2) add the requires_idaes_solver mark
3) give you a chance to look at this before I proceed with either (1) or (2)

I guess the 4th option could be for me to try to resolve the issue, but I'd rather focus time on more pressing matters.

bknueven:
> @bknueven I removed what I considered to be the problematic flowsheet and associated files and moved to #1062 . However, not surprisingly, the macos check is failing for the regular old nf flowsheet. 
> 
> 
> 
> I can do a couple things.
> 
> 1) move the nf flowsheet along to that separate PR (#1062)
> 
> 2) add the requires_idaes_solver mark
> 
> 3) give you a chance to look at this before I proceed with either (1) or (2)
> 
> 
> 
> I guess the 4th option could be for me to try to resolve the issue, but I'd rather focus time on more pressing matters.

Let's move forward with (1) or (2). I don't have an opinion on which is preferable.

adam-a-a:
@bknueven skipping the macos test then.

@dangunter @MichaelPesce not sure how heavily this would affect GUI users subsequently.

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -2132,6 +2164,10 @@ def calculate_scaling_factors(self):
                 )
                 iscale.set_scaling_factor(self.ionic_strength_molal, sf)
 
+        if self.is_property_constructed("total_hardness"):
+            if iscale.get_scaling_factor(self.total_hardness) is None:
+                sf = 1 / value(self.total_hardness)
```
bknueven, line 2169: Not a huge fan of this...

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -1498,12 +1495,11 @@ def initialize_build(
                 self._automate_rescale_variables()
         # ---------------------------------------------------------------------
         # Solve unit attempt 1
-        # self.eq_solute_solvent_flux.deactivate()
```
bknueven, line 1501: As an aside, it's interesting that this approach doesn't seem to work for this model. We use something similar in OARO, which was inspired by the IDAES heat exchanger initialization.

bknueven (approved):
I left some comments, but I don't think they should be addressed in this PR.

watertap/unit_models/nanofiltration_DSPMDE_0D.py
adam-a-a: It may be helpful for the model; I don't remember rechecking after the fact. This was something that I had in place when I was still debugging the model at a more fundamental level. Hard to recall, but once I found the real issues with the model, I think I found that this was no longer needed, commented it out, but never deleted. We could try it out again to see if it helps with stability, but would prefer to reserve for another PR.

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -1684,6 +1690,33 @@ def rule_elec_cond_phase(b, p):
...
+            units=pyunits.mg / pyunits.L,
+            doc="total hardness as CaCO3",
+        )
+
+        def rule_total_hardness(b):
+            return b.total_hardness == pyunits.convert(
+                sum(
+                    b.flow_mol_phase_comp["Liq", j]
+                    / b.flow_vol_phase["Liq"]
+                    * 100.0869
```
hunterbarber, line 1707: What is the 100.0869 parameter here? converting to CaCO3 equivalent?

Also curious of the decision to have the in-model units for hardness as milligrams, non SI and inconsistent with other mass use within the model. Even if it's just for reporting, wouldn't it be better to convert directly in the `.report()` function, in case this value is ever needed for other model util.

hunterbarber (approved):
Only had one comment, but also only viewed the raw text of the Jupyter Notebook, so if there's a way to view it in the proper display I can take another look in case there's something I missed.

PR #1031, opened 2023-05-17
New ASM2d Model

MarcusHolly:
## Summary/Motivation:
Adds property & reaction packages for the ASM2d as described by Flores-Alsina (https://www.sciencedirect.com/science/article/pii/S0043135415301160). This implementation makes fewer simplifying assumptions than our current Modified ASM2d implementation.

- X_TSS, X_MeOH, & X_MeP terms have been removed
- S_ALK (and any associated alkalinity terms) has been deleted and replaced with S_IC
- S_K and S_Mg added as solutes
- Updated Petersen Matrix based on the Flores-Alsina Excel file: https://app.box.com/file/1220462875801?s=xu1j32kfnb5m5xodatkvi61mvy6f9h6c
- Updated rate expressions based on the Flores-Alsina c-code: https://app.box.com/file/1220458811803
- Updated/added new parameter values based on Flores-Alsina MATLAB code: https://app.box.com/file/1220460272233
- Minor typo fixes in existing ASM files

### Legal Acknowledgement


MarcusHolly:
A few notes/questions on this:

1.) The file and class names will probably need to be updated, but I was unsure of what we wanted to do with the simplified Modified ASM2d model. Do we want to delete the simple model and just use its naming conventions here or should we keep both and have a different name for this model (not just "Modified ASM2d")
EDIT: We've decided to keep the current naming convention for now and will update it when the simplified ASM2d model is eventually replaced. 

2.) There is one known issue - one of the tests is failing because the CSTR fails to initialize. Before I dug into this, I wanted to know exactly what relationship EXPOsan has with the ASM2d test files. Is this just a source that we were validating against? Do I have access to it? Should we even be using the EXPOsan data/tests for this model because it doesn't appear to account for all of the Flores-Alsina modifications.
EDIT: Resolved by changing the inputs for S_K and S_Mg from EPS (1e-9)

3.) In the base ASM2d Petersen matrix, a few of the entries for S_N2 and S_NO3 were multiplied by mw_N/40kg/kmol. Why was this done and should I do something similar in this implementation?
EDIT: This is not necessary in this implementation since the c code does not suggest that any conversion is necessary.

agarciadiego (approved):

watertap/property_models/activated_sludge/new_modified_asm2d_properties.py
```diff
@@ -0,0 +1,382 @@
...
+@declare_process_block_class("NewASM2dParameterBlock")
+class NewASM2dParameterData(PhysicalParameterBlock):
+    """
+    Property Parameter Block Class
+    """
+
+    CONFIG = PhysicalParameterBlock.CONFIG()
+
+    CONFIG.declare(
+        "decay_switch",
```
adam-a-a: Where does this get utilized in the model?

MarcusHolly: A few of the rate expressions change depending on whether this is on or off. It is used in R9, R15, R16, R17, and R19.

MarcusHolly: @adam-a-a Are you just suggesting this should be a config option under the reaction package rather than the property package?

adam-a-a: @MarcusHolly Actually, I just searched for the config option and didn't see it in the rxn package due to the file being truncated by Github. Since you mention it,  it does make sense for it to be a config option for the rxn package. Are there any drawbacks to making the change? At first glance it seems like it doesn't make a difference functionally, but it would make more sense logically.

watertap/property_models/activated_sludge/new_modified_asm2d_properties.py
```diff
@@ -0,0 +1,351 @@
...
+        # Particulate species
+        self.X_AUT = Solute(doc="Autotrophic nitrifying organisms. [kg COD/m^3]")
+        self.X_H = Solute(doc="Heterotrophic organisms. [kg COD/m^3]")
+        self.X_I = Solute(doc="Inert particulate organic material. [kg COD/m^3]")
+        self.X_PAO = Solute(doc="Phosphate-accumulating organisms. [kg COD/m^3]")
+        self.X_PHA = Solute(
+            doc="A cell internal storage product of phosphorus-accumulating organisms, primarily comprising poly-hydroxy-alkanoates (PHA). [kg COD/m^3]"
+        )
+        self.X_PP = Solute(doc="Poly-phosphate. [kg P/m^3]")
+        self.X_S = Solute(doc="Slowly biodegradable substrates. [kg COD/m^3]")
```
adam-a-a: As a sanity check- do we expect this modification to comprise 18 components (excluding H2O)?

watertap/property_models/activated_sludge/new_modified_asm2d_reactions.py
```diff
@@ -0,0 +1,1601 @@
...
+        self.reaction_rate = pyo.Var(
+            self.params.rate_reaction_idx,
+            initialize=0,
+            doc="Rate of reaction",
+            units=pyo.units.kg / pyo.units.m**3 / pyo.units.s,
+        )
+
+        try:
+            # TODO: Refer to Flores-Alsina c code to account for sulfur in rate expressions
+            def rate_expression_rule(b, r, decay_switch=DecaySwitch.on):
```
adam-a-a: Does `decay_switch` need to be an arg here, or can we just check the config in relevant rate equations?

watertap/property_models/activated_sludge/tests/test_new_modified_asm2d_reaction.py
```diff
@@ -0,0 +1,851 @@
...
+        )
+        assert value(model.fs.R1.outlet.conc_mass_comp[0, "X_PP"]) == pytest.approx(
+            0, abs=1e-4
+        )
+        assert value(model.fs.R1.outlet.conc_mass_comp[0, "X_S"]) == pytest.approx(
+            123.44e-3, rel=1e-4
+        )
+
+
+class TestAerobic15C:
```
adam-a-a: Does 15C correspond to something in the literature or SI material specifically?

adam-a-a (commented):
I had a few questions/comments, but everything looks great to me. A couple more comments:

We should probably add an integration test, similar to what we have for ASM1 and conventional ASM2D testing, but this can just be added to the to-do list rather than handling in this PR.

With regard to naming, though seemingly trivial, I would opt to name our first modified version something like `simple_modified_asm2d` (and eventually phase out when we no longer need it anymore) and name this one `modified_asm2d`. @luohezhiming Does that sound good to you since you deal with the simple version now?

adam-a-a:
@MarcusHolly one more question- when would we use the decay_switch config option? Or rather, why would we opt to choose switch on vs switch off? Just want to get an idea of its significance and usage.

watertap/property_models/activated_sludge/new_modified_asm2d_properties.py
MarcusHolly: Yep, base ASM2d has 19 components, but we removed 4 (X_TSS, X_MeOH, X_MeP, S_ALK) and added 3 (S_K, S_Mg, S_IC). These components are the first 18 listed in the Gujer matrix, and as we discussed, other components such as sulfur and iron can be added later.

watertap/property_models/activated_sludge/new_modified_asm2d_reactions.py
MarcusHolly: I think it needs to be an arg here because, otherwise, you'll run into a `NameError: name 'decay_switch' is not defined`

watertap/property_models/activated_sludge/tests/test_new_modified_asm2d_reaction.py
MarcusHolly: Not sure - just copied this test from the other ASM2d test files, which I believe are based on EXPOsan. Although the naming is a bit odd because the temperature in all of these tests is 25C, not 15C. So I'm curious where this 15C came from as well.

watertap/property_models/activated_sludge/new_modified_asm2d_reactions.py
adam-a-a: You should point to where the config option lives though.

MarcusHolly: What do you mean by that?

MarcusHolly:
> @MarcusHolly one more question- when would we use the decay_switch config option? Or rather, why would we opt to choose switch on vs switch off? Just want to get an idea of its significance and usage.

Great question - I was wondering the same thing lol. Maybe this will be one of the questions we email Flores-Alsina et. al about.

MarcusHolly:
> I had a few questions/comments, but everything looks great to me. A couple more comments:
> 
> We should probably add an integration test, similar to what we have for ASM1 and conventional ASM2D testing, but this can just be added to the to-do list rather than handling in this PR.
> 
> With regard to naming, though seemingly trivial, I would opt to name our first modified version something like `simple_modified_asm2d` (and eventually phase out when we no longer need it anymore) and name this one `modified_asm2d`. @luohezhiming Does that sound good to you since you deal with the simple version now?

What's an integration test again?

luohezhiming:
> I had a few questions/comments, but everything looks great to me. A couple more comments:
> 
> We should probably add an integration test, similar to what we have for ASM1 and conventional ASM2D testing, but this can just be added to the to-do list rather than handling in this PR.
> 
> With regard to naming, though seemingly trivial, I would opt to name our first modified version something like `simple_modified_asm2d` (and eventually phase out when we no longer need it anymore) and name this one `modified_asm2d`. @luohezhiming Does that sound good to you since you deal with the simple version now?

@adam-a-a Yeah, it sounds good to me. Alternatively, we could name the previous version `asm2d_with_additional_solutes` because it only adds an option to add inert component (not limited to `S_K `and `S_Mg`), but as addressed in this new model, `S_K `and `S_Mg` should be added as reactants.

watertap/property_models/activated_sludge/modified_asm2d_reactions.py
```diff
@@ -10,81 +10,1592 @@
...
+    def _rxn_rate(self):
+        self.reaction_rate = pyo.Var(
+            self.params.rate_reaction_idx,
+            initialize=0,
+            doc="Rate of reaction",
+            units=pyo.units.kg / pyo.units.m**3 / pyo.units.s,
+        )
+
+        try:
+            # TODO: Refer to Flores-Alsina c code to account for sulfur in rate expressions
```
luohezhiming, line 1044: Since `S_K` and `S_Mg` are added as reactants, shouldn't they be accounted in these rate expressions? Or they are just zero-order reactants?

MarcusHolly: S_K and S_Mg were not implemented in their version of the rate expression, so I assume they have no effect (zero-order reactant). If you want to take a look at their code, there's a Box link in the PR header. It might be a bit overwhelming to look at at first, but the rate equations are written as `proc## = rate expression` and the terms corresponding to S_K and S_Mg are `xtemp[16]` and `xtemp[17]`, respectively. Neither of which are included in any of the rate expressions.

luohezhiming: OK. That make sense!

agarciadiego (approved):

adam-a-a:
In this case, we could check that this modified ASM2D model works within a flowsheet comprising reactors, as done in the existing integration test that we have for the original asm2D model.
> 
> I had a few questions/comments, but everything looks great to me. A couple more comments:
> 
> > 
> 
> > We should probably add an integration test, similar to what we have for ASM1 and conventional ASM2D testing, but this can just be added to the to-do list rather than handling in this PR.
> 
> > 
> 
> > With regard to naming, though seemingly trivial, I would opt to name our first modified version something like `simple_modified_asm2d` (and eventually phase out when we no longer need it anymore) and name this one `modified_asm2d`. @luohezhiming Does that sound good to you since you deal with the simple version now?
> 
> 
> 
> What's an integration test again?

watertap/property_models/activated_sludge/new_modified_asm2d_reactions.py
adam-a-a: The same way we deal with config options on other models, e.g., 'self.config.decay_switch', though it may not be as straightforward as 'self.config'.

MarcusHolly:
> In this case, we could check that this modified ASM2D model works within a flowsheet comprising reactors, as done in the existing integration test that we have for the original asm2D model.
> 
> > I had a few questions/comments, but everything looks great to me. A couple more comments:
> > > 
> > 
> > 
> > > We should probably add an integration test, similar to what we have for ASM1 and conventional ASM2D testing, but this can just be added to the to-do list rather than handling in this PR.
> > 
> > 
> > > 
> > 
> > 
> > > With regard to naming, though seemingly trivial, I would opt to name our first modified version something like `simple_modified_asm2d` (and eventually phase out when we no longer need it anymore) and name this one `modified_asm2d`. @luohezhiming Does that sound good to you since you deal with the simple version now?
> > 
> > 
> > What's an integration test again?

@adam-a-a I believe this has already been done - all the tests in `test_modified_asm2d_reaction` are based on the tests in `test_asm2d_reaction`

luohezhiming (approved):

PR #1035, opened 2023-05-19
Add a custom parameter sweep function that implements multiprocessing for local parallel sweeps  

avdudchenko:
## Summary/Motivation:
Current parameter tool requires use of MPI, which does not reliably work in many environments. This adds a custom function that can be used by the parameter tool to run sweeps in parallel using the native python multiprocessing package. 

- small change to how custom parameters are passed in param sweep tool. Instead of unpacking custom_do_param_sweep_kwargs into custom_do_param_sweep function, the param sweep class is passed into the function itself, this enables custom function dev to access all options/kwargs with in param sweep tool.  
-

### Legal Acknowledgement


avdudchenko:
So I think I will need to revert this to ray.io, I found just like with MPI on winodws when useing multiprocessing there are many issues with memory and paging files (e.g. after a few runs you get this error ImportError: DLL load failed while importing _socket: The paging file is too small for this operation to complete.) This is related to how multiprocessing and MPI load windows DLLs and there is not much you can do to solve it except increasing paging file size(which still might not be a good solution).

avdudchenko:
@k1nshuk @bknueven I have run a lot of tests on multiprocessing implementation and found issues with memory errors similar to  MPI, its likely my system is an outlier as I can run 10+ workers, but non the less i find ray.io being more reliable. I did major refactor of how the parallel code works to make it more versatile and included ray.io import (it also includes analysistools that I will put up as another PR later as part of an effort to bring analysis tools into watertap). It would be nice if ray.io instance could live on watertap, even if it's not tested and I can maintain it etc, until the parallel processor project comes online and proves to be better then ray in my use cases.

bknueven:
A lot of this functionality has been built into #986, which already supports MPI and multiprocessing simultaneously. We can also include ray.io as part of the new parallelization paradigm, which we were planning to start on once #986 is merged.

avdudchenko:
> A lot of this functionality has been built into #986, which already supports MPI and multiprocessing simultaneously. We can also include ray.io as part of the new parallelization paradigm, which we were planning to start on once #986 is merged.

That is fair, although the objective of this PR is to provide an example of how to setup custom functions for paramater sweep tool.  I think it would be a worthwhile addition, once I remove duplicate code, and its likely ones 986 is merged it then some of the options would need to change. 
 
986 needs quiet a bit of work and I left comments in there.

PR #1045, opened 2023-05-26
Add costing method for anaerobic digestor

luohezhiming:
## Fixes/Resolves:

## Summary/Motivation:
Costing method is based on `anaerobic_digestion_reactive_zo.py` (original WT3 model), and the electricity parameter is taken from Maravalias' data

-
-

### Legal Acknowledgement


watertap/costing/units/anaerobic_digestor.py
```diff
@@ -0,0 +1,92 @@
...
+)
+
+
+def build_anaerobic_digestor_cost_param_block(blk):
+
+    blk.capital_a_parameter = pyo.Var(
+        initialize=19.3552312e6,
+        doc="A parameter for capital cost",
+        units=pyo.units.USD_2012,
+    )
```
MarcusHolly, line 27: Maybe leave a note here mentioning that these values came from the anaerobic_digestion_oxidation.yaml?

MarcusHolly (approved):
One small comment - but otherwise LGTM

watertap/costing/units/anaerobic_digestor.py
luohezhiming: Yeah, it's good to have a reference and I've added a comment.

agarciadiego (approved):

watertap/costing/units/anaerobic_digestor.py
adam-a-a: It'd be even better if we noted the original source rather than the yaml

watertap/costing/units/anaerobic_digestor.py
```diff
@@ -0,0 +1,93 @@
...
+        blk.costing_package.cost_flow(
+            pyo.units.convert(
+                blk.unit_model.electricity[t0],
+                to_units=pyo.units.kW,
+            ),
+            "electricity",
+        )
+
+
+def cost_anaerobic_digestor_capital(
```
adam-a-a, line 66: We should could eventually consider and account for any typical o&m costs, but this is a start.

luohezhiming: I added the electricity cost in the unit model files.

watertap/unit_models/anaerobic_digestor.py
```diff
@@ -462,6 +462,19 @@ def build(self):
             mutable=True,
             doc="Gas-liquid transfer coefficient",
         )
+        self.electricity = Var(
+            self.flowsheet().time,
+            units=pyunits.kW,
```
adam-a-a, line 467: This is technically electric _power_ rather than energy if using units kW. I know we weren't strict about this in the past, but figured I'd mention now.

adam-a-a: I'd personally prefer calling this `power_consumption_electric`. Added the electric qualifier in case there is any thermal input that we'd consider later

EDIT: doesn't have to be named that. Could also be something like `electricity_consumption`, but I like putting power to signify that units are kW as opposed to kWh

watertap/costing/units/anaerobic_digestor.py
MarcusHolly: It seems like the original source of this data is the "NREL Waste-to-Energy Model". I got this info from cell A890 in the VAR tab of this Excel Kurby sent me earlier: https://app.box.com/file/1177021287137.

watertap/unit_models/anaerobic_digestor.py
adam-a-a: The other thing to consider is that ADs produce gas that can in turn be used to produce electricity, and I'd rather now confuse the user about whether this is electricity production or consumption.

watertap/unit_models/anaerobic_digestor.py
```diff
@@ -462,6 +462,19 @@ def build(self):
...
+            bounds=(0, None),
+            doc="Electricity consumption of unit",
+        )
+        # The value is taken from Maravelias' data
+        self.energy_electric_flow_mass = Param(
```
adam-a-a:
```suggestion
        self.energy_electric_flow_vol_inlet = Param(
```

watertap/unit_models/anaerobic_digestor.py
luohezhiming: In the Maravalia's data, they provide an electricity cost with respect to AD and at least for this value I added, I believe it is electricity consumption

adam-a-a (approved):
LGTM- I just have some suggested changes and comments for you to consider, but nothing that affects the functionality of the code!

watertap/unit_models/anaerobic_digestor.py
adam-a-a: Sure- I just meant we should rename from `electricity` to something more specific like `power_consumption_electric` or `electricity_consumption`.

watertap/costing/units/anaerobic_digestor.py
adam-a-a: That's good and should be good enough for this PR, but later, we might consider other costs such as heating and other relevant o&m.

watertap/costing/units/anaerobic_digestor.py
luohezhiming: Thanks, @MarcusHolly. I've updated the reference

watertap/unit_models/anaerobic_digestor.py
luohezhiming: > Sure- I just meant we should rename from `electricity` to something more specific like `power_consumption_electric` or `electricity_consumption`.

I change it to `electricity_consumption`

PR #1053, opened 2023-06-02
EC ZO cost update

MuktaHardikar:
## Fixes/Resolves:

- The cost of power supply is updated to be a linear function total power (W) required (with a zero intercept).
- A coefficient is included to double the cost of electrode material and therefore account for manufacturing cost
- Test file cost values were updated to reflect the above changes.

### Legal Acknowledgement

### DIFF
--- a/watertap/data/techno_economic/electrocoagulation.yaml
+++ b/watertap/data/techno_economic/electrocoagulation.yaml
@@ -57,10 +57,10 @@ default:
       value: 0.5523
       units: dimensionless
       reference: Sharma (2010) - doi.org/10.1016/j.envadv.2020.100029 
-    ec_power_supply_base:
-      value: 260000
-      units: USD_2020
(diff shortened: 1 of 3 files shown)


kurbansitterley (approved):
LGTM

hunterbarber (approved):
Small changes to the costing look fine. I would only double check that the LCOW reduction by 33% is expected and more accurate.

kurbansitterley:
> I would only double check that the LCOW reduction by 33% is expected and more accurate.

@hunterbarber The motivation for these changes was because industry partners said some aspects of our costing approach were way too high (power supply) and some were too low (electrode costs). So in theory these costs are more in line with what you might expect. Previously power supply costs were based on the number of power supplies needed ($260k base cost) and now it is purely related to the power needed. That brought estimated costs down a bunch.

PR #1054, opened 2023-06-06
NaCl Property Pack with Temperature Dependence 

savannahsakhai:
## Summary/Motivation:
New property model for NaCl with temperature dependence. 

- Adds NaCl with temperature dependence property model
- Adds testing
- Add documentation file

### Legal Acknowledgement


watertap/property_models/NaCl_T_dep_prop_pack.py
```diff
@@ -0,0 +1,1791 @@
...
+# National Renewable Energy Laboratory, and National Energy Technology
+# Laboratory (subject to receipt of any required approvals from the U.S. Dept.
+# of Energy). All rights reserved.
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and license
+# information, respectively. These files are also available online at the URL
+# "https://github.com/watertap-org/watertap/"
+#################################################################################
+"""
+Initial property package for H2O-NaCl system
```
adam-a-a:
```suggestion
Initial property package for H2O-NaCl system with temperature dependence
```

watertap/property_models/NaCl_T_dep_prop_pack.py
```diff
@@ -0,0 +1,1791 @@
...
+
+
+@declare_process_block_class("NaClParameterBlock")
+class NaClParameterData(PhysicalParameterBlock):
+    CONFIG = PhysicalParameterBlock.CONFIG()
+
+    def __init__(self, component):
+        super().__init__(component)
+        self.enth_param_A = None
+
```
adam-a-a: I think you can delete this.

watertap/property_models/NaCl_T_dep_prop_pack.py
```diff
@@ -0,0 +1,1791 @@
...
+        # solubility, 0-450 C
+        # Sparrow 2003, Eq 5: Xsat = param_0 + param_1 * T + param_2 * T**2
+        solubility_param_dict = {"0": 0.2628, "1": 62.75e-6, "2": 1.084e-6}
+        self.solubility_param = Var(
+            solubility_param_dict.keys(),
+            domain=Reals,
+            initialize=solubility_param_dict,
+            units=pyunits.dimensionless,
+            doc="Mass density parameters",
```
adam-a-a:
```suggestion
            doc="Solubility parameters",
```

watertap/property_models/NaCl_T_dep_prop_pack.py
```diff
@@ -0,0 +1,1791 @@
...
+        )
+        self.vap_pressure_E2_param = Var(
+            vap_pressure_E2_param_dict.keys(),
+            domain=Reals,
+            initialize=vap_pressure_E2_param_dict,
+            units=pyunits.MPa,
+            doc="Vapor pressure parameters E2",
+        )
+
+        # could add entropy if needed from Sparrow 2003, Eq. 9
```
adam-a-a:
```suggestion
        # TODO: could add entropy if needed from Sparrow 2003, Eq. 9
```

watertap/property_models/NaCl_T_dep_prop_pack.py
```diff
@@ -0,0 +1,1791 @@
...
+        self.th_cond_param_D = Var(
+            th_cond_D_param_dict.keys(),
+            domain=Reals,
+            initialize=th_cond_D_param_dict,
+            units=pyunits.W / (pyunits.m * pyunits.K),
+            doc="Thermal conductivity parameter D",
+        )
+
+        # viscosity, 0-150 salinity (g/kg), 20-180 T (C)
+        # Qasuem et al. 2021 (IAPWS-2008 correlation)
```
adam-a-a:
```suggestion
        # Qasem et al. 2021 (IAPWS-2008 correlation)
```

watertap/property_models/NaCl_T_dep_prop_pack.py
```diff
@@ -0,0 +1,1791 @@
...
+                "cp_mass_phase": {"method": "_cp_mass_phase"},
+            }
+        )
+        obj.define_custom_properties(
+            {
+                "osm_coeff": {"method": "_osm_coeff"},
+                "enth_flow": {"method": "_enth_flow"},
+                "solubility": {"method": "_solubility"},
+                "th_cond_phase": {"method": "_th_cond_phase"},
+                "vapor_pressure": {"method": "_vapor_pressure"},
```
adam-a-a: Can double check naming conventions in IDAES to see if anything can be refined here. Also, for consistency, may want to add qualifiers (i.e. `solubility_comp` where the component is NaCl, even though it is the only solute)

andrewlee94: 1. Anything that is indexed should have the indexing set(s) as a suffix
2. I think this is `therm_cond_phase`
3. `pressure_sat`

savannahsakhai: Thank you for bringing this to my attention! I was unaware of the naming conventions in the IDAES documentation. The naming is now updated.

watertap/property_models/NaCl_T_dep_prop_pack.py
savannahsakhai: This was an out of date reference. I went through and updated the references throughout the file. I had forgotten to do so after my most recent regression.

watertap/property_models/NaCl_T_dep_prop_pack.py
```diff
@@ -0,0 +1,1797 @@
...
+            bounds=(0.0, 5),
+            units=pyunits.MPa,
+            doc="Vapor Pressure",
+        )
+
+        def rule_pressure_sat(b):  # vapor pressure, eq. 6 in Sparrow 2003
+            t = (b.temperature - 273.15 * pyunits.K) / pyunits.K
+            T = value(t)
+            x = b.mass_frac_phase_comp["Liq", "NaCl"]
+            if (T >= 0) and (T <= 150):
```
adam-a-a: I would be careful with this since temperature is a variable, and during a solve, the value would be changing with each iteration. One could argue that temperature can typically be fixed, but since temp is indeed a variable and can't be guaranteed to always be fixed, I don't think you should use if-else conditionals.

Instead, you can try an `Expr_if`, which essentially should do what you intended the if-else conditional to do for you but as an `Expression`. The caveat is that instability can supposedly occur when the variable approaches the bound (e.g., T =150), but I haven't encountered this issue myself.

andrewlee94: You should generally never do this in a rule - the rule is only evaluated once at the time of model construction, and so will be based on the initial value of T only (which is likely a default value at this point). Also note that it is generally a very bad idea to use `value` inside rule.

As @adam-a-a said, you can use an `Expr_If` if a conditional is required or a smooth approximation of the if. However, you should always ask yourself if the `if` even matters - i.e. in the actual application will the `if` ever trigger?

savannahsakhai: I played around with this today. The `Expr_if` did work for what I had wanted it to do, but as you said (@adam-a-a), it did have difficulty with T = 150.

However, after giving your question (@andrewlee94) some thought, I don't think the `if` matters. I am unaware of applications which go above 150 C. Please correct me if I am forgetting an example which does.

savannahsakhai: I updated the pressure_sat with new parameters from a regression using new data which will avoid the need for an if statement while also encompassing the full temperature range up to 300 C.

watertap/property_models/NaCl_T_dep_prop_pack.py
```diff
@@ -1255,9 +1190,9 @@ def rule_solubility_comp(b, j):
 
     def _pressure_sat(self):
         self.pressure_sat = Var(
-            initialize=0.01,
-            bounds=(0.0, 2),
-            units=pyunits.MPa,
+            initialize=1e3,
+            bounds=(1, 1e8),
```
adam-a-a, line 1193: Are these bounds attributed to the valid range of the fitted relationship? Or are these fairly arbitrary?

watertap/property_models/tests/test_NaCl_T_dep_prop_pack.py
```diff
@@ -64,7 +64,7 @@ def configure(self):
             ("cp_mass_phase", "Liq"): 4000,
             ("solubility_comp", "NaCl"): 0.265,
             ("therm_cond_phase", "Liq"): 0.6015,
-            ("pressure_sat", None): 0.00293,
+            ("pressure_sat", None): 1767.56,
```
adam-a-a: Looks like a substantial change in the vapor pressure result. 

I just scrolled up and noticed the previous relationship was based on MPa, and I suppose since you fit a new relationship of your own, we would expect some variation from the previous relationship's results. Have you compared results  between each relationship?

watertap/property_models/NaCl_T_dep_prop_pack.py
savannahsakhai: These are fairly arbitrary. I have been using the seawater prop pack as a guide for how tight or loose to set bounds because I was having some difficulty when attempting to set up an example using a unit model. With the bounds more closely aligned to the seawater prop pack bounds, I was able to get a running example with RO.

watertap/property_models/tests/test_NaCl_T_dep_prop_pack.py
savannahsakhai: I had noticed that at these lower temperatures/salinities there was some greater deviation. I am going over the regression again now to ensure there is not an error there.

savannahsakhai: In looking into this further, I think the best course forward is to use the relationship from the Sparrow 2003 paper with temperature range, 0-150 C. (So, go back to the original implementation with units of MPa.)

At lower values of temperature (under 100 C), the vapor pressure's rate of change is low, but after that, the vapor pressure quickly increases with increasing temperatures. This has made the regression difficult to get an appropriate fit. I have tried manipulating the degree of polynomial and fitting only sections of the data at a time, but I am not getting better results with my regression. 

The Sparrow relationship more closely aligns with the experimental data than the regression I implemented here.

docs/technical_reference/property_models/NaCl_T_dep.rst
```diff
@@ -0,0 +1,120 @@
...
+
+   * 1e-2 for temperature
+   * 1e-6 for pressure
+   * 1e-3 for mass density
+   * 1e3 for dynamic viscosity
+   * 1e9 for solute diffusivity
+   * 1 for the osmotic coefficient
+   * 1e-2 for the specific enthalpy
+   * 1e-4 for the specific heat capacity
+   * 1e2 for vapor pressure
```
adam-a-a: revisit this considering the changes in relationships and requested changes from MPa to Pa

TimBartholomew:
Ok let's go back to Sparrow for vapor pressure, but change the units to Pa

TimBartholomew (approved):
LGTM - my understanding is that this has been validated and tested with the MVC model. Any changes we want can be noted in future issues and addressed in future PRs.

watertap/property_models/NaCl_T_dep_prop_pack.py
```diff
@@ -0,0 +1,1708 @@
...
+        )
+        self.enth_param_E = Var(
+            enth_E_param_dict.keys(),
+            domain=Reals,
+            initialize=enth_E_param_dict,
+            units=pyunits.J / pyunits.kg,
+            doc="Specific enthalpy parameter E",
+        )
+
+        # vapor pressure parameters, 0-150 C
```
adam-a-a, line 259: I wonder if this comment should also be above the constraint for pressure_sat and/or in the documentation

adam-a-a (approved):
LGTM

PR #1057, opened 2023-06-14
Adding costing method for rectifier capital cost and upscaled electricity flow

### LINKED ISSUE #1011: Best implementation for costing an AC to DC conversion for electrified processes
### Description

Electrocoagulation, electrodialysis, and electrolysis mainly, but also other unit models, will require costing for rectification equipment and other associated costs. It is unclear on whether this should be done on a model-to-model basis or have a universal method when this utility is needed.

### Motivation

Several electrochemical processes will need this conversion and costing estimation, so the objective is to have a unified method to do this across models.

### Possible Implementation

- cost AC and DC as separate electricity flows, add method for efficiency of this conversion and other costs within the `watertap_costing_package` directly
- add rectification equipment unit model(?) block that converts current, feeds it to `cost_electricity_flow`, and may have it's own costing correlation as a unit model
- check IDAES and DISPATCHES for any existing implementation of this and standardize for WaterTAP

hunterbarber:
## Fixes/Resolves:
Adds new method for costing rectifier as a util similar to `cost_membrane`

## Summary/Motivation:
Many electrified units require this, and therefore is best implemented as a global function

### Legal Acknowledgement


watertap/costing/util.py
```diff
@@ -100,6 +100,57 @@ def cost_membrane(blk, membrane_cost, factor_membrane_replacement):
...
+    by solar or some other DC power generator, this method should not be used. Assumes the unit_model has an `power`
+    variable or parameter.
+
+    Args:
+        ac_dc_conversion_efficiency - Efficiency of the conversion from AC to DC current
+    """
+
+    # create variables on cost block
+    make_capital_cost_var(blk)
+    blk.ac_dc_conversion_efficiency = pyo.Expression(expr=ac_dc_conversion_efficiency)
```
adam-a-a: add doc string for consistency

watertap/costing/util.py
```diff
@@ -100,6 +100,57 @@ def cost_membrane(blk, membrane_cost, factor_membrane_replacement):
     )
 
 
+def cost_rectifier(blk, ac_dc_conversion_efficiency=0.90):
+    """
+    Method to cost rectifiers required for electrified process units that use DC current. Note that if power is supplied
+    by solar or some other DC power generator, this method should not be used. Assumes the unit_model has an `power`
+    variable or parameter.
```
adam-a-a: "Note that if power is supplied by solar or some other DC power generator" -- should we just say that this should only be used for AC power sources and is meant to convert from AC to DC? That might add clarity.

watertap/costing/util.py
```diff
@@ -100,6 +100,57 @@ def cost_membrane(blk, membrane_cost, factor_membrane_replacement):
...
+    # refix variables to appropriate costing parameters
+    for index, var in blk.rectifier_cost_coeff.items():
+        var.fix(rectifier_cost_coeff[index])
+
+    # calculate capital cost
+    blk.capital_cost_constraint = pyo.Constraint(
+        expr=blk.capital_cost
+        == blk.rectifier_cost_coeff[1]
+        + (blk.rectifier_cost_coeff[0] * blk.ac_power)
+        * (pyo.units.USD_2020 * pyo.units.kW**-1)
```
adam-a-a: Comment above says USD_2022 embedded in eq, but USD_2020 implemented in this constraint. Can you clarify the discrepancy?

adam-a-a: Also, not clear how the units are working out. Can you add `assert_units_consistent(m)` to your test?

watertap/costing/tests/test_util_other.py
```diff
@@ -0,0 +1,81 @@
...
+from idaes.core import (
+    FlowsheetBlock,
+    UnitModelBlock,
+    UnitModelCostingBlock,
+)
+from idaes.core.solvers import get_solver
+from watertap.costing.watertap_costing_package import WaterTAPCosting
+from watertap.costing.util import cost_rectifier, register_costing_parameter_block
+
+
```
adam-a-a: Maybe you made this separate file for the sake of clarity, but I think this should go in `test_util`

watertap/costing/util.py
hunterbarber: So the data is recent and true to USD_2022, but the current support on currency units is only up to 2020 I believe. I was looking up the frequency of updating pyomo/IDAES units handling but couldn't find anything. Not sure how to handle this.

adam-a-a: Last comment on this (I think)--should we be converting to `base_currency` here? If not, then why not?

hunterbarber: Also units were implemented in the wrong place, correcting that now.

hunterbarber: I did find they're available up until 2021 in `idaes\core\base\costing_base.py`

hunterbarber: @adam-a-a In the short term I switched to assuming the `USD_2021` conversion is applicable for the costing parameters. Then I converted the `base_currency` of the test to `USD_2021` to keep the capital cost value the same for the assertion. I also had to then consider the adjusted the electricity price using the CEPCI's to ensure the cost calculation is still as expected. Will push in a second.

hunterbarber:
I'll mark this as ready to review for now, but it is not implemented in any cost methods specific to unit ops. It is only tested as an additional cost constructed at the flowsheet level.

If this version is the one to merge, the implementation of this method across electrified unit costing would either be in a subsequent PR(s) or added as a rolling implementation as those relevant cost models undergo other changes.

avdudchenko (approved):
LGTM

lbianchi-lbl:
A comment by @adam-a-a on #1006 applies here as well.

TimBartholomew (approved):

adam-a-a (approved):
LGTM - the potential issue I was referring to had to do with my anticipation of later moving to multiperiod modeling, where, say, the temporal resolution could be at 5 minutes. In such a case, I was questioning the validity of just applying power*electricity_cost, which would work out for an hourly resolution, but not higher temporal resolution. 

I think this is just a broader question to probe later on, once we move out of steady-state territory, and is outside the scope of this PR though.

hunterbarber:
> LGTM - the potential issue I was referring to had to do with my anticipation of later moving to multiperiod modeling, where, say, the temporal resolution could be at 5 minutes. In such a case, I was questioning the validity of just applying power*electricity_cost, which would work out for an hourly resolution, but not higher temporal resolution.
> 
> I think this is just a broader question to probe later on, once we move out of steady-state territory, and is outside the scope of this PR though.

Ah yea, when you said it, I was really only considering the hourly to annual conversion, and I agree that probably a larger overhaul across all unit models will be necessary when the push to dynamic comes.

hunterbarber:
@lbibl letting you know that auto-merge is on for this, so you can pull this directly from main later today.

PR #1059, opened 2023-06-14
Initial implementation of ConcurrentFuturesParallelManager

shelman:
## Fixes/Resolves:

The CFPM is only used if the new `number_of_subprocesses` config option is given to the ParameterSweep object. The changes should be entirely non-breaking and ready to merge.

I added a bit of unit test coverage for the CFPM by parametrizing one of the tests. If this API makes sense I'll add more ParallelManager-specific unit tests in a subsequent PR.

### Legal Acknowledgement


watertap/tools/parallel/mpi_parallel_manager.py
```diff
@@ -0,0 +1,62 @@
+import numpy
```
shelman, line 13: The code in this file is mostly unchanged, just split into its own file.

watertap/tools/parallel/single_process_parallel_manager.py
```diff
@@ -0,0 +1,35 @@
+from watertap.tools.parallel.parallel_manager import ParallelManager, run_sweep
```
shelman: This code is unchanged, just moved to its own file.

avdudchenko:
The current implementation will not work, as you can not pickle an ideas/watertap model, which is the assumption. You can take a look at pr https://github.com/watertap-org/watertap/pull/1035 that handles this in multiprocessing/ray implementation.  (you can test this by trying to run the RO_with_erd example with multiprocessing, you will need to set it up with class as in example below.) To deal with pickle issue, you will need to pass the model build function into each worker. Ideally each worker should be set up to build, and initialize model and stay alive receiving request to solve a problem (as i did in PR1035).  Do not expect to_json/from_json to work reliably or correctly, this is due to waterTap models not properly handling user scaling and resetting or modifying it upon call to iscale.calculate_scailing_factors() (as well as other strange behaviors that seem to leave models in an unsolvable/unstable state). (I left that in PR1035, but actually don't use it, and use my own tool which is not part of watertap atm). 

Beyond this, it looks like the current implementation mimics MPI, in that it is not asynchronous. The point of using native parallel implementation over MPI is to enable user to run parameter sweep on their own machines that have few logical cores and would significantly benefit form async operation, as in many cases when running parameter sweep tool there can be as many as 50% of points not being solved (skipped using probe function for example). (For example sweeping across WR% and salinity, half points might be in range where NaCl precipitates and you would not want to try solving) (PR1035 also shows how that could  be handled). 

`Traceback (most recent call last):
  File "monte_carlo_sampling_RO_ERD.py", line 128, in <module>
    run_parameter_sweep("output/monte_carlo_results.csv")
  File "monte_carlo_sampling_RO_ERD.py", line 115, in run_parameter_sweep
    global_results_arr, _ = ps.parameter_sweep(
  File "d:\onedrive\nawi_work\analysis\watertap\water_tap_forks\watertap_parallel\watertap\tools\parameter_sweep\parameter_sweep.py", line 761, in parameter_sweep
    all_results = parallel_manager.gather()
  File "d:\onedrive\nawi_work\analysis\watertap\water_tap_forks\watertap_parallel\watertap\tools\parallel\concurrent_futures_parallel_manager.py", line 80, in gather
    results.append(LocalResults(process_number, params, future.result()))
  File "D:\miniconda\envs\watertap-p\lib\concurrent\futures\_base.py", line 437, in result
    return self.__get_result()
  File "D:\miniconda\envs\watertap-p\lib\concurrent\futures\_base.py", line 389, in __get_result
    raise self._exception
  File "D:\miniconda\envs\watertap-p\lib\multiprocessing\queues.py", line 239, in _feed
    obj = _ForkingPickler.dumps(obj)
  File "D:\miniconda\envs\watertap-p\lib\multiprocessing\reduction.py", line 51, in dumps
    cls(buf, protocol).dump(obj)
_pickle.PicklingError: Can't pickle <class 'idaes.core.base.process_block._ScalarFlowsheetBlock'>: attribute lookup _ScalarFlowsheetBlock on idaes.core.base.process_block failed`


https://github.com/watertap-org/watertap/blob/main/watertap/examples/flowsheets/RO_with_energy_recovery/monte_carlo_sampling_RO_ERD.py
modify the sweep function as follows to reproduce the above error. 

`    # Run the parameter sweep
    ps = ParameterSweep(
        csv_results_file_name=csv_results_file_name,
        h5_results_file_name=h5_results_file_name,
        optimize_function=optimize,
        optimize_kwargs={"solver": solver, "check_termination": False},
        number_of_subprocesses=20,
    )
    global_results_arr, _ = ps.parameter_sweep(
        m,
        sweep_params,
        outputs,
        num_samples=num_samples,
        seed=seed,
    )
`

lbianchi-lbl:
@andrewlee94 mentions that IDAES/Pyomo models are supposed to be pickle-able, so the error is not expected and should be looked at as a possible bug.

avdudchenko:
> @andrewlee94 mentions that IDAES/Pyomo models are supposed to be pickle-able, so the error is not expected and should be looked at as a possible bug.

So pyomo models are, but i never could pickle the IDAES model. If that is solved could make these implementations much simpler.

watertap/tools/parallel/mpi_parallel_manager.py
```diff
@@ -0,0 +1,74 @@
...
+    def results_from_local_tree(self, results):
+        return [
+            next(
+                (
+                    result
+                    for result in results
+                    if result.process_number == self.get_rank()
+                )
+            )
+        ]
```
bknueven: This code confuses me a bit -- what are we returning here? The first time you come across a `result` with `process_number == self.get_rank()`, contained in a list with only that element?

shelman:
I just put up an update to this PR that allows a function `rebuild_common_sweep_args_fn` to be passed into the ParallelManager's `scatter` function in order to rebuild the arguments in each subprocess, getting around the issue of having to pickle the model. The local updates to `monte_carlo_sampling_RO_ERD.py` now work for me with multiple subprocesses specified.

I spoke with @k1nshuk and @bknueven and the consensus is that a smarter queueing system for the tasks makes sense but can be implemented in a later change, especially since this PR is already fairly large.

watertap/tools/parallel/mpi_parallel_manager.py
shelman: Yes - the idea is to return any results that were run on this specific process. It's used for writing the separate results to their output files using the ParameterSweepWriter, at the end of the parameter_sweep function.

bknueven: Doesn't 
```suggestion
        return [
                    result
                    for result in results
                    if result.process_number == self.get_rank()
        ]
```
accomplish the same thing?

shelman: Yes, good catch. Updated in latest patch.

watertap/examples/flowsheets/RO_with_energy_recovery/monte_carlo_sampling_RO_ERD.py
```diff
@@ -101,6 +100,56 @@ def run_parameter_sweep(
     outputs["EC"] = m.fs.costing.specific_energy_consumption
     outputs["LCOW"] = m.fs.costing.LCOW
 
+    ParameterSweep._assign_variable_names(m, outputs)
```
bknueven: Why are we calling an private class method here?

watertap/examples/flowsheets/RO_with_energy_recovery/monte_carlo_sampling_RO_ERD.py
```diff
@@ -101,6 +100,56 @@ def run_parameter_sweep(
...
+        additional_kwargs["number_of_subprocesses"] = number_of_subprocesses
+        additional_kwargs["rebuild_sweep_args_fn"] = build_sweep_args_minus_solver
+        additional_kwargs["rebuild_sweep_args_kwargs"] = {
+            "use_LHS": use_LHS,
+            "read_model_defauls_from_file": read_model_defauls_from_file,
+            "read_sweep_params_from_file": read_sweep_params_from_file,
+            "sweep_params_fname": sweep_params_fname,
+            "defaults_fname": defaults_fname,
+        }
+
```
bknueven: This example is now much more complicated, which really gives me pause...

bknueven (commented):
I think we should discuss the need for `rebuild_sweep_args_*`. It points to something in the current API which is not quite right.

watertap/examples/flowsheets/RO_with_energy_recovery/monte_carlo_sampling_RO_ERD.py
shelman: I did this because it needs to happen after the outputs are built, which happens inside the rebuilding function. I'll see if it can be cleaned up or removed.

shelman:
Patch up with the changes discussed offline. The parallel manager's `scatter()` function now accepts more generic `do_build` and `do_execute` arguments, similar to the code in `ui/fsapi.py`. The more specific code for rebuilding the model and sweep params is now in the parameter sweep file, and passed to the parallel manager according to the api.

docs/how_to_guides/how_to_use_parameter_sweep.rst
```diff
@@ -99,9 +99,16 @@ The file `outputs_results.csv` contains the `sweep_param` values and `outputs` v
...
+to the parameter_sweep function will be callable functions that return the appropriate objects. Passing the objects
+in directly still works but is deprecated. 
+
 .. testcode::
 
-    parameter_sweep(m, sweep_params, outputs, csv_results_file_name='outputs_results.csv', h5_results_file_name='outputs_results.h5')
+    build_model = lambda: m
+    build_sweep_params = lambda model: sweep_params
+    build_outputs = lambda model, sweep_params: outputs
+    parameter_sweep(build_model, build_sweep_params, build_outputs, csv_results_file_name='outputs_results.csv', h5_results_file_name='outputs_results.h5')
```
bknueven: We probably need to just update the documentation directly to reflect the new behavior -- could be a subsequent PR.

I don't want to leave this hack in the documentation which gets around the deprecation warning but still has the same problem as passing in the model directly.

shelman:
Patch up with changes:
- Fixed pylint issue.
- Updated doctest to ignore the deprecation output rather than hack around it.

bknueven (approved):

watertap/tools/parallel/concurrent_futures_parallel_manager.py
```diff
@@ -0,0 +1,105 @@
...
+        self.actual_number_of_subprocesses = None
+
+        # Future -> (process number, parameters). Used to keep track of the process number and parameters for
+        # all in-progress futures
+        self.running_futures = dict()
+
+    def is_root_process(self):
+        return True
+
+    def get_rank(self):
```
k1nshuk, line 35: We need to add simple tests here

watertap/tools/parallel/concurrent_futures_parallel_manager.py
```diff
@@ -0,0 +1,105 @@
...
+        return self.ROOT_PROCESS_RANK
+
+    def number_of_worker_processes(self):
+        if self.actual_number_of_subprocesses is None:
+            return self.max_number_of_subprocesses
+        return self.actual_number_of_subprocesses
```
k1nshuk, line 41: Same, here

watertap/tools/parameter_sweep/parameter_sweep.py
```diff
@@ -146,20 +147,31 @@ class _ParameterSweepBase(ABC):
...
 
-        self.parallel_manager = ParallelManager.create_parallel_manager(
-            options.pop("parallel_manager_class", None)
-        )
-
         self.comm = options.pop("comm", MPI.COMM_WORLD)
-        self.rank = self.comm.Get_rank()
+        parallel_manager_class = options.pop("parallel_manager_class", None)
 
+        self.rank = self.comm.Get_rank()
```
k1nshuk, line 167: Any reason why this is not covered by the parallel manager?

watertap/tools/parameter_sweep/parameter_sweep.py
```diff
@@ -708,63 +748,139 @@ def _combine_outputs(self, gathered_results):
 
         return np.asarray(combined_outputs)
 
+    """
```
k1nshuk, line 751: Is this docstring for `run_scatter_gather`?

k1nshuk (commented):
@shelman Looks good to merge in once I hear back from you on the comments.
@avdudchenko Any additional feedback?

avdudchenko (approved):
This looks good to me after setting up examples and tests. I suggest updated ERD example with parallelfutures example, but can be left for next PR.

bknueven:
@shelman please look at @k1nshuk's and @avdudchenko's comments an plan to address them in the next PR.

PR #1058, opened 2023-06-14
New ADM1/ASM2d Translator

MarcusHolly:
## Summary/Motivation:
Creates a translator block for the new ASM2d model and renames the translator block for the old, modified ASM2d model to `translator_adm1_simple_asm2d`

## Changes proposed in this PR:
- Creates translator block and test file for new ASM2d model
- Renames translator block and test file for old ASM2d model
- Updates imports in ElectroN-P so that it still uses the old translator block

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #1063, opened 2023-06-20
ADM1/ASM2d Translator Doc Update

MarcusHolly:
## Summary/Motivation:

Updates the documentation for the ADM1 -> ASM2d translator block based on changes made in #1058 

### Legal Acknowledgement


docs/technical_reference/unit_models/translators/translator_adm1_asm2d.rst
```diff
@@ -46,17 +46,16 @@ Sets
    "Time", ":math:`t`", "[0]"
    "Inlet/outlet", ":math:`x`", "['in', 'out']"
    "Phases", ":math:`p`", "['Liq']"
-   "Inlet Components", ":math:`j`", "['H2O', 'S_su', 'S_aa', 'S_fa', 'S_va', 'S_bu', 'S_pro', 'S_ac','S_h2','S_ch4','S_IC','S_IN','S_IP','S_I','X_ch','X_pr','X_li','X_su','X_aa','X_fa','X_c4','X_pro','X_ac','X_h2','X_I','X_PHA','X_PP','X_PAO','S_K','S_Mg']"
+   "Inlet Components", ":math:`j`", "['H2O', 'S_su', 'S_aa', 'S_fa', 'S_va', 'S_bu', 'S_pro', 'S_ac', 'S_h2', 'S_ch4', 'S_IC', 'S_IN', 'S_IP', 'S_I', 'X_ch', 'X_pr', 'X_li', 'X_su', 'X_aa', 'X_fa', 'X_c4', 'X_pro', 'X_ac', 'X_h2', 'X_I', 'X_PHA', 'X_PP', 'X_PAO', 'S_K', 'S_Mg']"
    "Ion", ":math:`j`", "['S_cat', 'S_an'] \  :sup:`1`"
-   "Outlet Components", ":math:`j`", "['H2O', 'S_A','S_F','S_I','S_N2','S_NH4','S_NO3','S_O2','S_PO4','S_ALK','X_AUT','X_H','X_I','X_MeOH','X_MeP','X_PAO','X_PHA','X_PP','X_S','X_TSS'] \  :sup:`2`"
+   "Outlet Components", ":math:`j`", "['H2O', 'S_A', 'S_F', 'S_I', 'S_N2', 'S_NH4', 'S_NO3', 'S_O2', 'S_PO4', 'S_K', 'S_Mg', 'S_IC', 'X_AUT', 'X_H', 'X_I', 'X_PAO', 'X_PHA', 'X_PP', 'X_S']"
```
adam-a-a: This table is a little confusing since some of the symbols do not carry through in any of the constraints listed below (e.g., k and m symbols) and the fact that we use _j_ for subsets of the component set was also a little confusing. I think this should be cleaned up to better align with the associated code, clean up sets in the code that might not be necessary, _or_ a bit of both.

MarcusHolly: @adam-a-a Hmmm, so would it be better to just list the inlet and outlet components as sets in the table? I agree that we probably shouldn't include sets like Readily Biodegradable COD (k) or Slowly Biodegradable COD (m) if we are not going to explicitly reference them elsewhere in the documentation. Or would you prefer that we make use of these sets in the documentation? For example, rather than typing out the equations for all of the "Unchanged Components", I could just give it a new symbol (let's say "z" for now) and state that the equation is z_out = z_in. Similar equations could be written to make use of the currently unused sets.

As for cleaning up the code itself, are you suggesting that we don't need to define sets like "readily_biodegradable" or "unchanged_components"? I feel like it makes sense to define these sets since it makes it easier to tell why those variables are being used in particular equations rather than just hard-coding everything in.

adam-a-a: @MarcusHolly Per your last point, I meant this to be more of a question of whether it still makes sense to have those separate sets in the code. Since it sounds like it is helpful to have those, then I would say no need to adjust the code. Instead, I would just want to either (a) consolidate the sets shown in the doc if they are not represented anywhere else in the doc (i.e., in the constraints) or (b) make use of sets in the doc as you suggested. I think (b) would be preferable since I anticipate that this approach would more directly align with the code and could be held side-by-side with it to provide helpful insight.

lbianchi-lbl:
Some more small changes are planned by @MarcusHolly but it should make the release.

luohezhiming (approved):

adam-a-a (approved):

PR #1069, opened 2023-06-28
add rectifier cost to electrodialysis 

lbibl:
## Summary/Motivation:

This adds the rectifier cost to ED unit and a one of its flowsheet.  
- In the electrodialysis costing,  a method argument is added to control `has_rectifier` cost or not
- When the method is called, the `rectifier_cost` is executed within the electrodialysis costing. 
- Added new cost testing on the unit and modified a flowsheet to reflect rectifier cost.  

### Legal Acknowledgement


lbibl:
Reviewers:  because rectifier electricity is bound to ED power calculation, I chose to embed the rectifier_cost in the ED_cost as a part of its electricity flow and have it controlled by an arg. This is the best way to minimize mistakes afaik.  I am fairly new to the costing package structures so I need your quality control over these changes.  Particularly, I hope to have your examination over the value changes in corresponding costing tests and see if they make sense to what is expected by the added `rectifier_cost` method.  One puzzling thing to me is why, upon the addition of a rectifier cost,  the `specific_energy_consumption` increased (expected) but the `LCOW` decreased in a conditioned simulation- see this in L69-71 in test_electrodialysis_1stack_conc_recirc.py

hunterbarber:
Note, investigating how the rectifier cost method constructs the `blk.capital_cost` variable

watertap/costing/units/rectifier.py
```diff
@@ -0,0 +1,89 @@
...
+# through Lawrence Berkeley National Laboratory, Oak Ridge National Laboratory,
+# National Renewable Energy Laboratory, and National Energy Technology
+# Laboratory (subject to receipt of any required approvals from the U.S. Dept.
+# of Energy). All rights reserved.
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and license
+# information, respectively. These files are also available online at the URL
+# "https://github.com/watertap-org/watertap/"
+#################################################################################
+
```
adam-a-a, line 12: What was the key takeaway point on why we'd want to make this a separate "unit" instead of keeping this in costing/util? I saw a lot of back and forth on Slack, but it would be good to get a clear summary.

watertap/costing/units/rectifier.py
```diff
@@ -0,0 +1,89 @@
...
+        initialize=100,
+        domain=pyo.NonNegativeReals,
+        units=pyo.units.kW,
+        doc="Unit AC power",
+    )
+
+    # use unit.power variable in conversion with efficiency
+    blk.power_conversion = pyo.Constraint(
+        expr=blk.ac_power * blk.costing_package.rectifier.ac_dc_conversion_efficiency
+        == pyo.units.convert(blk.unit_model.power, to_units=pyo.units.kW)
```
adam-a-a, line 59: Assuming we maintain this approach, which I am not immediately crazy about, let's at least rename this to `dc_power` for clarity.

adam-a-a (changes requested):
My guess is that rectifier costing was reorganized so that after solving the ED flowsheet, one could grab the unit capital and operating costs for each unit more easily; i.e., the user could more easily obtain rectifier costs for reporting purposes. 

However, the implementation seems a bit inconsistent in comparison to how costing is dealt with in other cases in WaterTAP, and I do think that consistency matters. 

I don't think there is a necessity to merge this in for the June release, and I think we should iterate a little bit more before trying to cram this into the main code base. 

If we end up keeping these changes, this is one of those cases where documentation should be included upfront.

watertap/costing/units/rectifier.py
hunterbarber: So the current version [0b0e7ec](https://github.com/watertap-org/watertap/pull/1069/commits/0b0e7ec94e390f6e45c815da7e76a919c028e5a0) is structured the way it is because it was the minimal amount of refactoring to not have this as the bottleneck for the ED analysis. 

The main discussion that then will have subsequent challenges to address in either implementation is:
1. Should the rectifier `captial_cost` exist on its own unit model costing block? If it exists on its own unit model costing block, it must also have a registered unit model block.
2.  If it exists as a component of `electrodialysis.costing.captial_cost = (electrodialysis.costing.electrodialysis_capital_cost +  electrodialysis.costing.rectifier_capital_cost)`, is it a clear cost breakdown? Say two electrified unit ops are operating in parallel, supplied by the same rectifier. This costing method implies one half of the rectifier is being costed on each electrified unit costing block.
 
If making this analogous to a pump that supplies multiple membrane processes operating in parallel, then 1 is the consistent way. The pump is a distinctly different cost that supplies something the membrane processes. If making this analogous to the membrane costs of a membrane unit op, then 2 is the consistent way. The membrane is a core component of the unit op, and they are costed as the same unit.

lbibl:
Team has discussed to linger longer on how to design the rectifier cost and separating the ED needs on my local only at this point.  This PR is therefore invalidated and closed.

PR #1075, opened 2023-07-06
Add additional comments for valid temp ranges for NaCl_T_dep_prop_pack

### LINKED ISSUE #1070: Note valid ranges for NaCl w/temp dependency
I wonder if this comment should also be above the constraint for pressure_sat and/or in the documentation

_Originally posted by @adam-a-a in https://github.com/watertap-org/watertap/pull/1054#discussion_r1247137351_

This is probably an issue that applies to all property models, but in this case I was explicitly referring to vapor pressure only being valid from 0-150C in the nacl with temp dependency package.

We should be clear about this in documentation and in comments in the code.

savannahsakhai:
## Fixes/Resolves:
 Fixes #1070 (Note valid ranges for NaCl w/temp dependency)

## Summary/Motivation:
There are existing comments on the temperature validity ranges of each property given above their parameters. This will extend the comments to also appear above their constraint. 

- add additional comments listing the valid temperature ranges for different properties above their constraint
-

### Legal Acknowledgement

### DIFF
--- a/watertap/property_models/NaCl_T_dep_prop_pack.py
+++ b/watertap/property_models/NaCl_T_dep_prop_pack.py
@@ -1036,7 +1036,7 @@ class NaClStateBlockData(StateBlockData):
             doc="Mass density",
         )
 
-        def rule_dens_mass_phase(b, p):  # density, eq. 7 in Sparrow 2003, 0-300 C
+        def rule_dens_mass_phase(b, p):  # Sparrow 2003, Eq. 7, 0-300 C
             t = (b.temperature - 273.15 * pyunits.K) / pyunits.K
             param_vec = [
                 b.params.dens_mass_param_A,
--- a/watertap/property_models/NaCl_T_dep_prop_pack.py
+++ b/watertap/property_models/NaCl_T_dep_prop_pack.py
@@ -1036,7 +1036,7 @@ class NaClStateBlockData(StateBlockData):
             doc="Mass density",
         )
 
-        def rule_dens_mass_phase(b, p):  # density, eq. 7 in Sparrow 2003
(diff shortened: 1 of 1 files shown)


hunterbarber (approved):
The in-line comments are fine to approve. 

I was thinking about it and would assume hard bounding the variables would be bad and I'm approving that it is not in place. I was going to suggest logging a warning that when a property constructed is outside of the bounds to say "A variable which has been constructed is outside of the validated bounds, property will be solved but is an extrapolation. Results should be evaluated" or something similar. But I think that since all of the properties are over different validity it would be such a long conditional that it would be hard to test and may be more trouble than it is worth. As long as it is in the documentation.

watertap/property_models/NaCl_T_dep_prop_pack.py
```diff
@@ -1295,7 +1295,9 @@ def _therm_cond_phase(self):
             doc="Thermal Conductivity",
         )
 
-        def rule_therm_cond_phase(b, p):
+        def rule_therm_cond_phase(
+            b, p
+        ):  # Regressed from Zaytsev & Aseev (1992), 0-155 C
```
bknueven:
```suggestion
        # Regressed from Zaytsev & Aseev (1992), 0-155 C
        def rule_therm_cond_phase(b, p):  
```

You didn't create the original issue, but we should get in the habit of putting comments *above* the code they're documenting, not to the right of it.

Otherwise `black` will make multi-line code out of otherwise single lines like here. I would encourage this change in every comment this PR touches.

savannahsakhai: Thank you for pointing this out to me. I believe everything should be reformatted now.

adam-a-a (approved):
LGTM

PR #1077, opened 2023-07-07
Add sweep analysis for electroNP flowsheet

luohezhiming:
## Fixes/Resolves:

- add sweep analysis
- revise costing data

### Legal Acknowledgement


watertap/costing/watertap_costing_package.py
```diff
@@ -184,6 +184,16 @@ def build_global_params(self):
         )
         self.add_defined_flow("magnesium chloride", self.magnesium_chloride_cost)
```
```suggestion
        self.add_defined_flow("magnesium chloride", self.magnesium_chloride_cost)
```
On second thought, shouldn't this go under the electronp costing and not in the global param block for WT costing?

watertap/costing/watertap_costing_package.py
```diff
@@ -184,6 +184,16 @@ def build_global_params(self):
...
+        self.phosphorus_recovery_revenue = pyo.Param(
+            mutable=True,
+            initialize=-0.07,
+            doc="Phosphorus recovery revenue",
+            units=pyo.units.USD_2020 / pyo.units.kg,
+        )
+        self.add_defined_flow(
+            "phosphorus salt product", self.phosphorus_recovery_revenue
+        )
+
```
adam-a-a: same thing here; this should be moved to electronp costing

adam-a-a: Otherwise this defined flow will be added for other flowsheets that don't have anything to do with phosphorus recovery

watertap/costing/watertap_costing_package.py
luohezhiming: This is the chemical addition registered flow for operating cost, it is similar to electricity and based on convention, it is added in the global param block. For the unit block, we call it by `cost_MgCl2_flow`

watertap/examples/flowsheets/case_studies/electroNP/electroNP_flowsheet.py
```diff
@@ -177,125 +182,95 @@ def build_flowsheet():
...
+    total_operating_cost = value(
+        pyunits.convert(
+            m.fs.costing.total_operating_cost, to_units=pyunits.USD_2018 / pyunits.year
+        )
+    )
+    print(f"Total Operating Cost: {total_operating_cost:.3f} $/year")
+
+    opex_fraction = total_operating_cost / (
+        total_operating_cost + value(m.fs.costing.total_capital_cost)
+    )
```
adam-a-a: This doesn't make sense to include to me, and I actually removed it from the report.  Also, should the units not work out? That is, $/yr + $ ?

luohezhiming: This is from our previous 0D flowsheet, we try to report it as an annually Opex proportion

watertap/costing/watertap_costing_package.py
luohezhiming: I moved it to unit model for now

watertap/costing/watertap_costing_package.py
luohezhiming: I moved these variables to unit model

watertap/examples/flowsheets/case_studies/electroNP/electroNP_flowsheet.py
luohezhiming: I updated the Opex fraction as opex/LCOW

watertap/costing/units/electroNP.py
```diff
@@ -25,17 +25,36 @@ def build_electroNP_cost_param_block(blk):
...
+    costing = blk.parent_block()
+    blk.magnesium_chloride_cost = pyo.Param(
+        mutable=True,
+        initialize=0.0786,
+        doc="Magnesium chloride cost",
+        units=pyo.units.USD_2020 / pyo.units.kg,
+    )
+    costing.add_defined_flow("magnesium chloride", blk.magnesium_chloride_cost)
+
+    blk.phosphorus_recovery_revenue = pyo.Param(
```
adam-a-a:
```suggestion
    blk.phosphorus_recovery_value = pyo.Param(
```

watertap/costing/units/electroNP.py
```diff
@@ -25,17 +25,36 @@ def build_electroNP_cost_param_block(blk):
...
+        mutable=True,
+        initialize=-0.07,
+        doc="Phosphorus recovery revenue",
```
adam-a-a:
```suggestion
        doc="Phosphorus recovery value",
```

watertap/costing/units/electroNP.py
```diff
@@ -25,17 +25,36 @@ def build_electroNP_cost_param_block(blk):
...
+    costing.add_defined_flow("phosphorus salt product", blk.phosphorus_recovery_revenue)
+
 
 @register_costing_parameter_block(
     build_rule=build_electroNP_cost_param_block,
     parameter_block_name="electroNP",
 )
-def cost_electroNP(blk, cost_electricity_flow=True, cost_MgCl2_flow=True):
+def cost_electroNP(
+    blk, cost_electricity_flow=True, cost_MgCl2_flow=True, cost_product_flow=True
```
adam-a-a:
```suggestion
    blk, cost_electricity_flow=True, cost_MgCl2_flow=True, cost_phosphorus_flow=True
```

watertap/costing/watertap_costing_package.py
```diff
@@ -176,13 +176,23 @@ def build_global_params(self):
...
+        #
+        # self.phosphorus_recovery_revenue = pyo.Param(
+        #     mutable=True,
+        #     initialize=-0.07,
+        #     doc="Phosphorus recovery revenue",
+        #     units=pyo.units.USD_2020 / pyo.units.kg,
+        # )
+        # self.add_defined_flow(
+        #     "phosphorus salt product", self.phosphorus_recovery_revenue
+        # )
```
adam-a-a: delete

watertap/examples/flowsheets/case_studies/electroNP/multi_sweep.py
```diff
@@ -0,0 +1,128 @@
...
+    LinearSample,
+    parameter_sweep,
+)
+import watertap.examples.flowsheets.case_studies.electroNP.electroNP_flowsheet as electroNP_flowsheet
+from pyomo.environ import units as pyunits
+
+
+def set_up_sensitivity(m):
+    outputs = {}
+    # optimize_kwargs = {"fail_flag": False}
```
adam-a-a: delete

watertap/examples/flowsheets/case_studies/electroNP/multi_sweep.py
```diff
@@ -0,0 +1,128 @@
...
+    opt_function = electroNP_flowsheet.solve
+
+    # create outputs
+    outputs["LCOW"] = m.fs.costing.LCOW
+    outputs["ElectroNP Capital Cost"] = m.fs.electroNP.costing.capital_cost
+    outputs["Electricity"] = pyunits.convert(
+        m.fs.costing.aggregate_flow_costs["electricity"] / m.fs.AD.inlet.flow_vol[0],
+        to_units=pyunits.USD_2018 / pyunits.m**3,
+    )
+    outputs["S_PO4 Concentration"] = m.fs.electroNP.treated.conc_mass_comp[0, "S_PO4"]
```
adam-a-a:
```suggestion
    outputs["S_PO4 Concentration, treated effluent stream"] = m.fs.electroNP.treated.conc_mass_comp[0, "S_PO4"]
```
I would add something to distinguish from treated stream concentration and byproduct.

watertap/unit_models/electroNP_ZO.py
```diff
@@ -462,8 +462,18 @@ def calculate_scaling_factors(self):
...
 
         iscale.set_scaling_factor(self.recovery_frac_mass_H2O, 1)
-        iscale.set_scaling_factor(self.energy_electric_flow_mass, 1e2)
-        iscale.set_scaling_factor(self.magnesium_chloride_dosage, 1e1)
+
+        if iscale.get_scaling_factor(self.energy_electric_flow_mass) is None:
+            sf = iscale.get_scaling_factor(
+                self.energy_electric_flow_mass, default=1e2, warning=True
+            )
+        iscale.set_scaling_factor(self.energy_electric_flow_mass, sf)
```
adam-a-a: Shouldn't this set_scaling_factor line be under the if conditional above?

watertap/unit_models/electroNP_ZO.py
```diff
@@ -462,8 +462,18 @@ def calculate_scaling_factors(self):
...
+
+        if iscale.get_scaling_factor(self.magnesium_chloride_dosage) is None:
+            sf = iscale.get_scaling_factor(
+                self.magnesium_chloride_dosage, default=1e1, warning=True
+            )
+        iscale.set_scaling_factor(self.magnesium_chloride_dosage, sf)
```
adam-a-a: Same here- this set_scaling_factor should be within the if conditional because we only want to set the scaling factor if the user has not provided one in their flowsheet. The way it is written now, if a user supplied a scaling factor for this variable and ran calculate_scaling_factors afterwards, the scaling factor they specified would be overwritten without them realizing.

adam-a-a: Also, just realized that you don't need both lines (get_scaling_factor _and_ set_scaling_factor), since the first will set a default scaling factor and show a warning.

adam-a-a (changes requested):
Just some suggested cleanup, but looks close to being merged

luohezhiming:
> Just some suggested cleanup, but looks close to being merged

Thank you for your careful review! I've addressed all revisions as you suggested.

watertap/unit_models/electroNP_ZO.py
luohezhiming: @adam-a-a I fixed the indent problem, but I am a little bit unsure why I don't need to `set_scaling_factor`, I checked other files like `nanofiltration_ZO.py`, they also do `get_scaling_factor` then `set_scaling_factor` even with a warning=True.

adam-a-a (approved):

watertap/examples/flowsheets/case_studies/electroNP/electroNP_flowsheet.py
```diff
@@ -177,125 +180,101 @@ def build_flowsheet():
...
-        "X_I",
-        "X_H",
-        "X_PP",
-        "X_S",
-    ]
-
-    for i in skip_constraint:
-        m.fs.electroNP.solute_removal_equation[0.0, "Liq", i].deactivate()
-        m.fs.electroNP.solute_treated_equation[0.0, "Liq", i].deactivate()
-
```
agarciadiego, line 242: Have you checked why we can skip this constraints and the model still works?

luohezhiming: The only reason I could imagine is the removal fraction is setting to a really small value 1e-7 and these components should just pass through, but I still want to keep these removal equations just in case if we want to update these removal fractions later.

agarciadiego (approved):

PR #1078, opened 2023-07-07
ASM2d compatibility with thickener and dewatering units

agarciadiego:
## Summary/Motivation:

- Created CONFIG list and gives the ability to use other property package other than ASM1

- deleted initialize lists for sets and TSS sums
- gives ability to user to give lists to initialize set and the TSS sum

### Legal Acknowledgement

### DIFF
--- a/watertap/property_models/activated_sludge/modified_asm2d_properties.py
+++ b/watertap/property_models/activated_sludge/modified_asm2d_properties.py
@@ -172,7 +172,7 @@ class ModifiedASM2dParameterData(PhysicalParameterBlock):
             doc="Reference temperature",
             units=pyo.units.K,
         )
-        
+
         # COD to VSS coefficients
         self.CODtoVSS_XI = pyo.Var(
             initialize=1.5686,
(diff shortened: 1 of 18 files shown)


luohezhiming (commented):
I think we need to set up a test file for ASM2D properties pakcage for both these two units

adam-a-a:
> I think we need to set up a test file for ASM2D properties pakcage for both these two units

@luohezhiming - I agree, and I am working on that now.

watertap/unit_models/thickener.py
```diff
@@ -47,16 +49,49 @@
 _log = idaeslog.getLogger(__name__)
 
 
+class ActivatedSludgeModelType(Enum):
+    """
+    ASM1: ASM1 model
+    ASM2D: ASM2D model
+    modified_ASM2D: modified ASM2D model for ADM1 compatibility
```
MarcusHolly, line 58: @adam-a-a Is this thickener only for use in BSM2 (as the file description says)? If so, I don't think it makes sense to include regular ASM2d as a config option since it is incompatible with BSM2. It might still be worth keeping the changes you made to `asm2d_properties.py` though.

adam-a-a: I should change that file description because a user could be interested in assembling a flowsheet with a thickener that could make use of ASM2D without the need to convert to ADM1.

adam-a-a:
> I think we need to set up a test file for ASM2D properties pakcage for both these two units

@luohezhiming I've added a full set of tests for the thickener and then selected what I thought were the more important tests for dewaterer (skipped all the solution tests for asm2d and modified asm2d, but did for thickener). I can add the full set of tests for dewaterer too, or we can go ahead and merge this so that you can make use of the dewaterer with modified asm2d

MarcusHolly (approved):
LGTM - not sure how @luohezhiming feels about Adam's last comment

adam-a-a:
@luohezhiming @MarcusHolly - I added one more modification: incorporated TSS, VSS, and ISS into modified ADM1 as well. I'll fix any failing tests and then should be ready to go.

watertap/property_models/activated_sludge/modified_asm2d_properties.py
```diff
@@ -132,6 +173,49 @@ def build(self):
             units=pyo.units.K,
         )
 
+        # COD to VSS coefficients
```
luohezhiming, line 176: Maybe we should mention the source for these parameters, it seems not from the original ASM2D properties model

luohezhiming (approved):
Really good work. Define different componnet sets for properties package is a very good method for compatiblity with other units, and the config added for the thickner and dewater works well. My only concern is probably we want to specify the references for those new added parameters for the modified ASM2D model

PR #1079, opened 2023-07-07
Add mapping keys between nominal and differential values for the DifferentialParameterSweep object

k1nshuk:
## Fixes/Resolves:

Adds additional keys in the output dictionary that contains indices that serve as a map between the nominal and differential `sweep_param` values

## Summary/Motivation:

This will enable easier integration with Alex's plotting and analysis tools.

- Additional mapping keys
- More robust differential parameter sweep tests

### Legal Acknowledgement


adam-a-a:
@k1nshuk do we have any documentation on the differential parameter sweep functionality? I'm not exactly clear on its usage yet and from a brief review of our readthedocs, I didn't notice any explicit documentation on this. If I am overlooking something, please point me in the right direction. Otherwise, if we are lacking the documentation, I think it would be good to add a technical reference and/or how-to to provide further clarity.

**EDIT:** at a minimum, I think just an additional section in one of the existing how-to docs would be a low-effort start.

k1nshuk:
@adam-a-a I was planning a documentation overhaul after the technical report gets done. But I will add an additional section to in the How-to docs in the meantime on this PR. The functions and the classes have the docstrings describing how to set it up, but I missed the updating the actual docs.

avdudchenko (commented):
Looks overall good tome, just need to fix the issue with par naming. I tested Diff outputs and the

avdudchenko (approved):

bknueven (approved):

PR #1082, opened 2023-07-11
Correct units in GAC costing `energy_consumption_constraint`

hunterbarber:
## Fixes/Resolves:
The costing results were validated for the GAC costing package but `assert_units_consistent` was never checked (could be a larger issue of no costing test conventions). The units for the regressed costing parameters were never embedded into the equation, resulting in a units error. This adds those units and the check into the test file.

## Summary/Motivation:
Correct units error in GAC costing.

- Adds `assert_units_consistent` for costing package, previously unchecked
- Adds embedded units for cost equation parameters

### Legal Acknowledgement


savannahsakhai (approved):
LGTM.

bknueven (approved):

PR #1085, opened 2023-07-16
Expose model configuration options

### LINKED ISSUE #1084: Enable configuration of model options in UI
### Description

Fixes https://github.com/watertap-org/watertap-ui/issues/85

### Motivation

_No response_

### Possible Implementation

_No response_

### Additional Context

_No response_

dangunter:
Fixes #1084

### Legal Acknowledgement

### DIFF
--- a/watertap/ui/fsapi.py
+++ b/watertap/ui/fsapi.py
@@ -19,7 +19,7 @@ __author__ = "Dan Gunter"
 import logging
 from collections import namedtuple
 from enum import Enum
-from typing import Any, Callable, Optional, Dict, Union, TypeVar
+from typing import Any, Callable, List, Optional, Dict, Union, TypeVar
 from types import ModuleType
 
 try:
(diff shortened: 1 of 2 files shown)


ksbeattie:
@dangunter, we put this (and the chain of 2 issues) on the Sept release board on the assumption that this will be ready in Sept.

dangunter:
@MichaelPesce marking ready for review to simplify any contributions/testing. As mentioned in https://github.com/watertap-org/watertap-ui/pull/90 , the main to-do here is to show how the option value(s) are used in building/solving e.g. the nanofiltration flowsheet, and documenting how to use when wrapping new models.

adam-a-a:
@dangunter - one thought about this functionality- selecting a specific configuration option could change whether certain variables should be available for the user to touch. One way to deal with this would be to be able to specify behind the scenese which vars would be "grayed out" and inaccessible with particular config options. It'd probably make sense to test this out starting simple with just an RO unit alone. Alternatively, we could look at this from the perspective of a given flowsheet, where you might want to toggle between predetermined variations of a given flowsheet. Personally, I think the former example would be better to start with.

adam-a-a:
@dangunter @MichaelPesce I am trying to get back up to speed with state of the UI. I am behind at this point. 

Is this PR still relevant, or is this one already covered by other UI-related PRs? Please let me know so we can decide whether to review or close this one.

adam-a-a:
This PR was superseded by #1152

PR #1086, opened 2023-07-17
Update electrodialysis costing 

lbibl:
## Summary/Motivation:

This PR makes minor modification to the rectifier costing util, adds rectifier cost onto electrodialysis (ED)'s costing as an option, updates ED's costing methods and their tests, and updates an ED flowsheet to reflect the changes. 

The updated ED costing is referenced to a widely-cited paper and quotes all materials and equipments of a ED stack as per unit of membrane area.  After all my investigation, I found this to be the mostly used approach in literature despite of minor variants. 

- Stated above. 

### Legal Acknowledgement


watertap/costing/units/electrodialysis.py
```diff
@@ -13,41 +13,23 @@
...
 
 
 def build_electrodialysis_cost_param_block(blk):
 
-    blk.cem_membrane_cost = pyo.Var(
-        initialize=43,
-        doc="Cost of CEM membrane used in Electrodialysis ($/CEM/area)",
+    blk.membrane_capital_cost = pyo.Var(
+        initialize=150,
+        doc="Membrane and capitcal costs in [US$/m^2-membrane-area], referenced to Desalination 142 (2002) 267-286.",
```
adam-a-a:
```suggestion
        doc="Stack capital cost in [US$/m^2-membrane area], which includes costs for membranes and other components within the stack",
```
I might also mention the reference (`"referenced to Desalination 142 (2002) 267-286."`) outside of this doc string

watertap/costing/units/electrodialysis.py
```diff
@@ -13,41 +13,23 @@
...
     make_fixed_operating_cost_var,
 )
 
 
 def build_electrodialysis_cost_param_block(blk):
 
-    blk.cem_membrane_cost = pyo.Var(
-        initialize=43,
-        doc="Cost of CEM membrane used in Electrodialysis ($/CEM/area)",
+    blk.membrane_capital_cost = pyo.Var(
```
adam-a-a, line 24: Does it make sense to just name this `stack_capital_cost`? Technically, the same can be said of RO, where we can name it something like `module_cost` although I think it is not as straightforward for ED as it is for RO.

watertap/costing/units/electrodialysis.py
```diff
@@ -13,41 +13,23 @@
...
-        units=pyo.units.USD_2018 / (pyo.units.meter**2),
-    )
-    blk.flowspacer_cost = pyo.Var(
-        initialize=3,
-        doc="Cost of the spacers used in Electrodialysis ($/spacer/area)",
-        units=pyo.units.USD_2018 / (pyo.units.meter**2),
-    )
-    blk.factor_membrane_housing_replacement = pyo.Var(
+
+    blk.factor_membrane_equipment_replacement = pyo.Var(
```
adam-a-a: Similarly, should this be `stack` replacement factor instead of membrane?

watertap/costing/util.py
```diff
@@ -100,7 +100,7 @@ def cost_membrane(blk, membrane_cost, factor_membrane_replacement):
     )
 
 
-def cost_rectifier(blk, ac_dc_conversion_efficiency=0.90):
+def cost_rectifier(blk, power=100 * pyo.units.kW, ac_dc_conversion_efficiency=0.90):
```
hunterbarber, line 103: I know that the power is set to the appropriate power variable when used, but I have some concern of assigning a default without a warning. Either `power=None` and error that no power variable was supplied or keep it as is and a warning that the rectifier is being costed with the default. Let me know your thoughts.

hunterbarber (commented):
Looks good, just had one comment.

watertap/costing/units/electrodialysis.py
avdudchenko: I agree, lest rename this to stack_capital_cost. This will later aid us if we want to split cost of stack into membranes/electrodes, as we can add it just as a constraint that calculates stack capital cost.

watertap/costing/units/electrodialysis.py
```diff
@@ -56,7 +38,7 @@ def build_electrodialysis_cost_param_block(blk):
     build_rule=build_electrodialysis_cost_param_block,
     parameter_block_name="electrodialysis",
 )
-def cost_electrodialysis(blk, cost_electricity_flow=True):
+def cost_electrodialysis(blk, cost_electricity_flow=True, has_rectifier=False):
```
avdudchenko, line 53: I think has_rectifier should default to true, as in general, ED requires an electrifier to operate.  @adam-a-a  @lbibl Thoughts?

watertap/costing/units/electrodialysis.py
```diff
@@ -65,96 +47,72 @@ def cost_electrodialysis(blk, cost_electricity_flow=True):
...
+            blk.costing_package.cost_flow(
+                pyo.units.convert(
+                    blk.unit_model.get_power_electrical(t0),
+                    to_units=pyo.units.kW,
+                ),
+                "electricity",
+            )
+        else:
+            power = blk.unit_model.get_power_electrical(blk.flowsheet().time.first())
+            cost_rectifier(blk, power=power, ac_dc_conversion_efficiency=0.9)
```
avdudchenko, line 77: So with the current rectifier implementation, it is not possible to update the value of ac_dc_conversion_efficiency (its an expression). So besides passing has_rectifier, the user should be able to pass in the ac_dc_conversion_efficiency as well.

watertap/costing/util.py
avdudchenko: do you mean adding a warning that user did not provide power variable to the unit? I think that should definetly be a warning (I assume it should throw an error, as it won't find a power value).

avdudchenko (approved):
Approving, my general comments are suggestions. (improvements for future PR's maybe?

watertap/costing/units/electrodialysis.py
lbibl: Methods changed with new names.

watertap/costing/units/electrodialysis.py
lbibl: new names made.

watertap/costing/units/electrodialysis.py
lbibl: I don't  really think this matters, as long as the user can read the word... :)

watertap/costing/units/electrodialysis.py
lbibl: I think the efficiency is in the argument too?

watertap/costing/util.py
lbibl: Honestly,  I think this is ok and we should assume the user has basic abilities to avoid mistakes from taking in default values.  We assign default values to variables all over the places. If we worry about this, we probably should not give default values to anything that is subjective to change.  Also at this point,  `util` always pairs with a unit having a `power` and the chance of mistake is even lower.    I would leave as-is  now.

lbibl:
ED costing method was updated again with new names.  Can reviews read again?

watertap/costing/util.py
hunterbarber: The difference is in other locations, using the default value for a variable is assuming that the variable equals that default value. Its like the conversion efficiency here. Not changing from the default is simply accepting 0.9 as the value. Using the 'default' for the power variable here is direct misuse of the function in most if not all applications of the function, which I why I think it warrants a warning.

hunterbarber (approved):
Changes since last review look fine, I did respond to a previous conversation that may need a subsequent issue to add one warning if deemed necessary if you would like to merge this through first.

PR #1088, opened 2023-07-20
Make pytest the only requirement for running test suite without errors

### LINKED ISSUE #1073: Ensure there are no test failures when only base dependencies + pytest are installed
Tests that need optional dependencies (e.g. `nbconvert`, `jsonschema`, `mongomock`, etc) should handle the absence of it gracefully using e.g. `pytest.importorskip()` or similar.

lbianchi-lbl:
- Add requirement for missing dependencies (`pyyaml`, `requests`)
- Add handling to skip tests if dependencies not available (`mongomock`, `json-schema-for-humans`, `requests`)
  - `requests` actually gets installed (for the moment) but I added the import guard anyways for futureproofing
- Remove redundant notebook tests (we're using `nbmake` now)
- Switch to using `find_packages()` instead of `find_namespace_packages()` since it's more explicit and we don't have namespace packages anyway

### Legal Acknowledgement


bknueven (commented):
Is there a reason we've elected not to use Pyomo's deferred import module here? It would remove some of the import logic here inside functions.

https://github.com/Pyomo/pyomo/blob/main/pyomo/common/dependencies.py

dangunter (approved):
Looks good

lbianchi-lbl:
> Is there a reason we've elected not to use Pyomo's deferred import module here? It would remove some of the import logic here inside functions.
> 
> https://github.com/Pyomo/pyomo/blob/main/pyomo/common/dependencies.py

Thanks for the suggestion, I think this is a fair point. My answer would be:

- For use in test files (i.e. `test_XYZ.py`) at the module level, I tend to use `pytest.importorskip()` directly since we don't have to worry about clients importing it as a module
- Conversely, for non-test modules, we could use Pyomo's `attempt_import()` to keep it more general and decouple the usage from the handling
- Most of the import guards are due to `mongomock` being used for `MockDB`, which is unfortunately re-defined several times in as many separate modules. Rather than trying to refactor this as part of this PR, I've opted for the "ugly and dumb" option of adding the import guard to each definition separately. In this case, my initial reaction was that using the Pyomo functionality wouldn't help much since you'd anyway have to do it for each file, but now I'm thinking that it might make the code a bit cleaner

**TL;DR** good point. I'll try changing some of the import guards to use `attempt_import()` instead.

bknueven (approved):

PR #1090, opened 2023-07-24
Remove lingering uses of MPI from parallel manager

shelman:
## Fixes/Resolves:

Gets rid of the remaining explicit uses of MPI from the base parameter sweep class. There's a bit of awkwardness in replacing MPI functions with ones on the parallel manager class, since a lot of the functions are similar to each other, but I wanted to keep as close as possible to how the code currently works.

### Legal Acknowledgement


bknueven (approved):

k1nshuk (approved):
Looks good to me, Once the unit tests are added, we can merge it.

Issue #1087, opened 2023-07-19
Enhance ElectroN-P model


luohezhiming:
### Description

- [x] Update model structure for elctroN-P unit model #1091 
- [x] Update documentation to reflect changes associated with inheriting from Separator

#1091

lbianchi-lbl:
The remaining open task is waiting for a PR, planned for September.

PR #1091, opened 2023-07-27
Changing ElectroN-P unit model to separator structure

agarciadiego:
## Summary/Motivation:

Changes electroN-P structure and utilizes an IDAES separator as a base model. Avoids unused constraints. #1087 

- Change unit model structure 
-

### Legal Acknowledgement


watertap/unit_models/electroNP_ZO.py
```diff
@@ -119,66 +71,18 @@ def _process_config(self):
...
 
-    def build(self):
-        # Call UnitModel.build to setup dynamics
-        super().build()
+        if "treated" and "byproduct" not in self.config.outlet_list:
+            raise ConfigurationError(
+                "{} encountered unrecognised "
+                "outlet_list. This should not "
+                "occur - please use overflow "
+                "and underflow as outlets.".format(self.name)
```
MarcusHolly: This ConfigurationError should say "treated" and "byproduct" instead of "overflow" and "underflow", right?

agarciadiego: yes

MarcusHolly (approved):
LGTM - just that one small typo

watertap/unit_models/electroNP_ZO.py
```diff
@@ -475,14 +220,15 @@ def calculate_scaling_factors(self):
...
-                sf = 1e6
-            iscale.set_scaling_factor(v, sf)
+        for (t, i, j), v in self.split_fraction.items():
+            for i in self.config.outlet_list:
+                if j == "S_PO4":
+                    sf = 1
+                elif j == "S_NH4":
+                    sf = 1
+                else:
+                    sf = 1
```
luohezhiming: sf should be `1e6` for `else`

watertap/costing/units/electroNP.py
```diff
@@ -104,7 +103,7 @@ def cost_electroNP_capital(blk, HRT, sizing_cost):
     blk.sizing_cost = pyo.Expression(expr=sizing_cost)
 
     flow_in = pyo.units.convert(
-        blk.unit_model.properties_in[0].flow_vol,
+        blk.unit_model.mixed_state[0].flow_vol,
```
luohezhiming, line 106: Just want to confirm with you, the `proerties_in` will be converted `mixed_state`, `properties_treated` will be converted to `treated_state`, and `properties_byproduct` will be converted to `byproduct_state`

watertap/unit_models/electroNP_ZO.py
luohezhiming:
```suggestion
                    sf = 1e6
```

luohezhiming (commented):
Looks good to me, some minor edits need to be addressed.

watertap/costing/units/electroNP.py
agarciadiego: yes in order to use the IDAES separator

adam-a-a: You could technically add object references to preserve the original names too, which would enable @luohezhiming to keep what he has without having to change things.

adam-a-a: We've done things like that in other models (e.g., RO).

adam-a-a:
One general comment I have concerns the documentation for the model. Didn't double-check readthedocs just yet, but we should make sure to add a technical ref for this (or modify existing) to reflect inheritance from IDAES separator. Personally, when I wasn't super familiar with the Separator naming conventions, I found it less straightforward to point to properties_in and either of the outlet stateblocks. Would be beneficial to make that clear in documentation. Not something for this PR if there is no doc for the model, but if there is a doc, then we should modify that doc in this PR too.

watertap/unit_models/electroNP_ZO.py
```diff
@@ -13,91 +13,43 @@
...
 __author__ = "Chenyu Wang"
 
 _log = idaeslog.getLogger(__name__)
 
 
 @declare_process_block_class("ElectroNPZO")
-class ElectroNPZOData(InitializationMixin, UnitModelBlockData):
+class ElectroNPZOdata(SeparatorData):
     """
     Zero order electrochemical nutrient removal (ElectroNP) model based on specified water flux and ion rejection.
```
adam-a-a: I know this doc string is old, but "specified water flux and ion rejection" seems inaccurate. Should be something like "based on specified removal efficiencies for nitrogen and phosphorus."

watertap/unit_models/electroNP_ZO.py
```diff
@@ -273,25 +111,23 @@ def solute_treated_equation(b, t, p, j):
...
+                return blk.split_fraction[t, "byproduct", i] == blk.N_removal
             else:
-                return b.removal_frac_mass_comp[t, j] == 1e-7
-
-        self._stream_table_dict = {
-            "Inlet": self.inlet,
-            "Treated": self.treated,
-            "Byproduct": self.byproduct,
-        }
+                return blk.split_fraction[t, "byproduct", i] == 1e-7
```
adam-a-a: Can this be 0 without causing issues? If not, should we add an epsilon parameter similar to what was done in previous PRs? 

Also, did you consider just fixing these values to the variables rather than writing as constraints?

agarciadiego: I did consider it, but I thought would be best to leave it as constraints in case we get expressions on how to calculate this through data. But I could change it

watertap/unit_models/electroNP_ZO.py
```diff
@@ -339,107 +175,16 @@ def MgCl2_demand(b, t):
...
-                f"{self.name} failed to initialize successfully. Please check "
-                f"the output logs for more information."
-            )
-
     def _get_performance_contents(self, time_point=0):
         var_dict = {}
         var_dict["Water Recovery"] = self.recovery_frac_mass_H2O[time_point]
         for j in self.config.property_package.solute_set:
-            var_dict[f"Solute Removal {j}"] = self.removal_frac_mass_comp[time_point, j]
+            var_dict[f"Solute Removal {j}"] = self.split_fraction[0, "byproduct", j]
```
adam-a-a: I just want to point out that `removal_frac_mass_comp` was a variable that was created for all the zero order models that we worked on previously (for WaterTAP3 migration as well as 1st year IEDO 2336 effort). 

Although we made this unit model to deviate away from the zero-order model framework, I do wonder if it would be worthwhile to preserve the same naming conventions (i.e., through object references so we don't lose the new names, such as split fraction). If we ever want to make use of an array of zero-order models down the line, say, to do that superstructure optimization with BSM2 and a set of zero-order model options, including this one, I could see variations in naming being an obstacle to a more streamlined analysis effort. 

Not that serious, but something to be aware of.

watertap/unit_models/electroNP_ZO.py
adam-a-a: is there a reason this was resolved?

watertap/unit_models/tests/test_electroNP_ZO.py
```diff
@@ -126,26 +123,22 @@ def test_calculate_scaling(self, ElectroNP_frame):
...
+        unscaled_var_list = list(iscale.unscaled_variables_generator(ElectroNP_frame))
+        assert len(unscaled_var_list) == 0
+
     @pytest.mark.solver
     @pytest.mark.skipif(solver is None, reason="Solver not available")
     @pytest.mark.component
     def test_solve(self, ElectroNP_frame):
         m = ElectroNP_frame
-        results = solver.solve(m)
+        results = solver.solve(m, tee=True)
```
adam-a-a: no need for tee=True for final merge

adam-a-a (commented):
LGTM- just a couple of comments regarding former naming conventions and whether we might still want to keep those around as references. I am curious on @luohezhiming 's thoughts on whether he'd want that or not.

watertap/costing/units/electroNP.py
luohezhiming: Yeah, I agree with Adam, if you could follow the conventions to rename the `mixed_state`, it would be really helpful.

luohezhiming (approved):

watertap/unit_models/electroNP_ZO.py
```diff
@@ -83,6 +84,10 @@ def build(self):
 
         units_meta = self.config.property_package.get_metadata().get_derived_units
 
+        add_object_reference(self, "proerties_in", self.mixed_state)
```
adam-a-a: We should probably have test for the refs. Note the typo here which wouldn't be caught since we don't test it.

adam-a-a (changes requested):
just need to fix typo for object ref and add tests

MarcusHolly:
> just need to fix typo for object ref and add tests

@adam-a-a This should be good to go now.

adam-a-a (approved):

PR #1092, opened 2023-08-02
Dye Desalination Sweep Visualization

MarcusHolly:
## Summary/Motivation:
This is essentially an extension of #1126 and adds visualization for the dye desalination sweep analysis. See issue #1127 for more details

### Legal Acknowledgement


watertap/data/techno_economic/nanofiltration.yaml
```diff
@@ -108,29 +108,29 @@ rHGO_dye_rejection:
...
     units: dimensionless
   removal_frac_mass_comp:
     dye:
-      value: 0.9839
+      value: 0.99
       units: dimensionless
       constituent_longform: Dye
     tds:
-      value: 0.2655
+      value: 0.05
```
adam-a-a: These are mass removal fractions--not exactly the same as rejection, which we're considering as 1- permeate concentration/feed concentration (and is stored in `rejection_comp`). You can check to see what the newly calculated rejection would be based on these changes and should find that the previous values would yield rejection rates that match these values. Thus, we should change these values back.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
```diff
@@ -148,7 +149,7 @@ def solve(blk, solver=None, checkpoint=None, tee=False, fail_flag=True):
     return results
 
 
-def add_costing(m):
+def add_costing(m, dye_revenue=False):
```
adam-a-a:
```suggestion
def add_costing(m, has_dye_revenue=False):
```

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
```diff
@@ -184,17 +186,32 @@ def add_costing(m):
...
+                * m.fs.zo_costing.dye_recovery_cost
+                * pyunits.convert(
+                    m.fs.dye_retentate.properties[0].flow_vol,
+                    to_units=pyunits.m**3 / m.fs.zo_costing.base_period,
+                )
+            ),
+            doc="Profit from recovering dye",
+        )
+    # Note: this is multiplied by -1 since the sign is handled implicitly
+    elif dye_revenue == False:
```
adam-a-a: Any reason why we can't just handle this with one expression for dye_value, eliminate the Boolean, and then toggle the value for `dye_value` between negative and positive values (representing either dye disposal cost or dye recovery cost)?

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
```diff
@@ -184,17 +186,32 @@ def add_costing(m):
...
+        m.fs.dye_value = Expression(
+            expr=(
+                m.fs.zo_costing.utilization_factor
+                * m.fs.zo_costing.dye_recovery_cost
+                * pyunits.convert(
+                    m.fs.dye_retentate.properties[0].flow_vol,
+                    to_units=pyunits.m**3 / m.fs.zo_costing.base_period,
+                )
+            ),
+            doc="Profit from recovering dye",
```
adam-a-a:
```suggestion
            doc="Revenue from recovering dye",
```

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
```diff
@@ -237,7 +259,10 @@ def set_operating_conditions(m):
     solve(m.fs.feed, checkpoint="solve feed block")
 
     # pretreatment
-    prtrt.wwtp.load_parameters_from_database(use_default_removal=True)
+    if include_pretreatment == True:
```
adam-a-a: Instead of having to pass `include_pretreatment` through each function, you could check if pretreatment exists on the model with `hasattr`.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
```diff
@@ -273,23 +299,29 @@ def set_operating_conditions(m):
...
-
-    # initialized nf
-    propagate_state(prtrt.s01)
+    # initialize pretreatment
+    if include_pretreatment == True:
+        propagate_state(m.fs.s_feed)
+        s = SequentialDecomposition()
+        s.options.tear_set = []
+        s.options.iterLim = 1
+        s.run(prtrt, lambda u: u.initialize())
```
```suggestion
        s.run(prtrt, lambda u: u.initialize())
```
This looks like the only thing that would need to be in the conditional--i.e., whether you pass `prtrt` or `dye_sep`.

watertap/data/techno_economic/nanofiltration.yaml
adam-a-a: Nevermind this comment. I realized after the fact that the collaborators specified this update in values, but we should verify that they truly intended these values for mass removal fractions (I think it is possible that they confused these with solute rejection.)

So, no need to revert these values.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
MarcusHolly: I could lump everything into one expression, but the two expressions aren't exactly the same. They have different signs (as you already mentioned), but if `dye_revenue == True`, it uses the dye recovery value and if `dye revenue == False`, it uses the dye disposal value. And the absolute value of the dye recovery value is not equal to the absolute value of the dye disposal value. So I felt like having one expression would be a bit too clunky, but if you still think otherwise, I could change it.

adam-a-a:
@MarcusHolly I am trying to get caught up and do some cleanup here. Can you refresh my memory on whether this PR is still relevant and should be revitalized/reviewed or closed? My feeling is that your most recent PR would include the needed changes, but could use your input.

MarcusHolly:
TLDR: If we want to add the visualization tools into WaterTAP, we can use this PR as a starting point. Otherwise, we can close it since the replica PR without visualization has already been merged.

adam-a-a:
Closing this draft PR since it's only purpose would be to bring in some visualization tools. However, @avdudchenko and others on the analysis team have a more robust set of visualization tools that we will ultimately bring in--either via a separate repo that we import from or directly into main (more likely the former).

That said, it'll take some time for those tools to be made available, and we thought having _some_ initial assistance via sweep visualization tools would be useful, but for some reason, the plotting script in this draft had some problems.

PR #1099, opened 2023-08-08
Allow passing build_outputs=None to parameter sweep

shelman:
## Fixes/Resolves:

Right now build_outputs=None breaks the concurrent futures parallel manager implementation because it tries to pickle a "return None" lambda on the fly. This is a bit confusing to the user, so build_outputs=None should probably still be allowed.

### Legal Acknowledgement

### DIFF
--- a/watertap/tools/parameter_sweep/parameter_sweep.py
+++ b/watertap/tools/parameter_sweep/parameter_sweep.py
@@ -814,6 +814,9 @@ class ParameterSweep(_ParameterSweepBase):
                 version="0.10.0",
             )
 
+        if build_outputs is None:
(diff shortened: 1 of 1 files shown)


k1nshuk (approved):

bknueven (approved):

avdudchenko (approved):
I will approve this as now. it fixes issues I faced before. It still does not address the issue, that the build_outputs, does not accept any kwargs, so it requires user to hardcode the outputs into the function being passed. Lets either update this in this PR or next, but user should be able to pass a build_output and build_output_kwargs options to specify which outputs to build dynamically.

PR #1038, opened 2023-05-22
BSM2 full flowsheet

agarciadiego:
## Summary/Motivation:
BSM2 full flowsheet #934 

## Changes proposed in this PR:
-
-

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #1100, opened 2023-08-09
Extended electroNP flowsheet with ASM and AD

luohezhiming:
## Fixes/Resolves:

-
-

### Legal Acknowledgement


adam-a-a:
@luohezhiming - Just for the record, this includes @Xinhong's revisions on modified ASM2d, correct? If so, that would allow us to remove those changes on the BSM2 flowsheet PR (#1038)

luohezhiming:
> @luohezhiming - Just for the record, this includes @XinHong's revisions on modified ASM2d, correct? If so, that would allow us to remove those changes on the BSM2 flowsheet PR (#1038)

Yeah, this included @XinHong's reformulation for ASM2D and ADM1 model.

luohezhiming:
@XinHong, this PR included your reformulations of the modified ASM2D and modified ADM1 models. A new `flowsheet modified_BSM2_flowsheet.py` is added in this PR, it is a sub-flowsheet of BSM2 without recycles but using modified ASM2D and modified ADM1 properties models instead of ASM1 and ADM1. Your reformulations really help to resolve scaling issues and the flowsheet can be solved to optimal, but I get an error `Cannot recompute multipliers for feasibility problem.  Error in eq_mult_calculator` when I check the solver log. I believe there are still some scaling issues or degeneracy, if you could help to check it, we will really appreciate you!

watertap/examples/flowsheets/case_studies/BSM2/BSM2_flowsheet2.py
```diff
@@ -0,0 +1,422 @@
+#################################################################################
```
adam-a-a, line 1: @luohezhiming I am not sure where this is coming from. I don't think it is from main because I reorganized this and renamed this BSM2 flowsheet, so you should start by making sure you're using the most up-to-date work

luohezhiming:
This PR is replaced by PR #1112

PR #1101, opened 2023-08-11
Modified ASM2D Refinement and Associated Changes

agarciadiego:
## Fixes/Resolves:

Reformulation of modified models

## Summary/Motivation:

Requires PR #1038 

- Changes and scaling for modified models

### Legal Acknowledgement


adam-a-a:
@agarciadiego -One issue I see is that the decay_switch config option was deleted, but this is used in the ASM2d/ADM1 translator from what I recall. One of the failed tests I observed seemed to indicate so as well. It's one thing to move said config option to the translator (if that ended up being the more logical choice), but I don't think we should delete that config option entirely. @MarcusHolly  and @luohezhiming had some dialogue about this config option in #1058.

MarcusHolly:
> @agarciadiego -One issue I see is that the decay_switch config option was deleted, but this is used in the ASM2d/ADM1 translator from what I recall. One of the failed tests I observed seemed to indicate so as well. It's one thing to move said config option to the translator (if that ended up being the more logical choice), but I don't think we should delete that config option entirely. @MarcusHolly and @luohezhiming had some dialogue about this config option in #1058.

@adam-a-a If I were to guess, the reason the decay_switch config option was deleted here is because I actually create the config option in two different places (ReactionParameterData & ReactionBlockData) in `modified_asm2d_reactions.py`. There's probably a way to get around creating it twice, but I couldn't get it working with just one instance because some tests would always fail, as is the case here. It's definitely not ideal to have it twice, so if there is some way around it, we should definitely implement it.

adam-a-a:
It looks like Xinhong's change of eps to 0 results in unitserror for some of the Expr_IFs. Basically, I think 0 *kg/m3 gets recognized as 0 and dimensionless, so tests fail because dimensionless is incompatible with kg/m3. I changed eps back to 1e-9 and edited some regression test tolerances after reverting back to @MarcusHolly's version.

adam-a-a (approved):
LGTM - I made some changes as I saw fit. Should be good to merge

MarcusHolly (approved):
LGTM - I believe the doc failures are unrelated to this PR

PR #1102, opened 2023-08-13
Add async multiprocessing, rayio, and update kernel in parameter sweep

avdudchenko:
## Fixes/Resolves:
(1) Adds multiprocessing with async, and ray io with async backends for multiprocessing
(2) Fixes the kernel for better stability when re-solving model after failure, fixed issue where re-initialization would not update sweep-parameters ( as parameters were not re-applied after re-initialization which would typically reset the vars in model) 
(3) Depreciated re-initialize function - and added initialize function to accurately mirror typical flow sheet setup 
(4) Fixes build_outputs function to accept kwargs so it can be dynamically generated. 
(5) Added build model, and build outputs to config. 
(6) Updated most test to work with new build_model, and build_outputs methods. 
(7) Minor updates to use pyomo object names instead of pyomo objects to update output_skeleton

## Summary/Motivation:
This update implements configuration updates that mirror typical flowsheet setups, and solving routines performed in testing. This implements better multiprocessing infrastructure. 

The addition of Async provides better resource utilization in situations where there is more "sweep_params" then actual logical cores, and model does not solve for every sweep_param, or does not have consistent solving performance (e.g. solving for one parameter results in longer solve time then another).

An example of this was benchmarked by running RO with ERD flowsheet example across water recoveries and feed salinities in an 8 by 8 grid (total 64 samples), using 8 cores on intel I7 12700H processor.  

The RO-ERD flowsheet is a good model with fast initiation and solving properties, providing most optimistic scenario for batch solving as currently implemented in MPI/ConcurrentFutures implementation. 

Table below shows that if all samples solve (no solve failures) there is no advantage for async implementation of MultiProcessing and RayIO over concurrent futures (and MPI, which not async). The rayio/multiprocessing involve extra overhead in starting up unique workers, and sending parmaters to them resulting in slightly worse performance in this case. 

The Async performs better under more relevant conditions of typical flowsheet simulations, where a number of samples can fail (an example is solving across feed salinities and water recoveries out side of feasible region, which comprises sometimes 50% of samples tested), if only 50% of samples solve (50% of sweep_params result in infeasible solves) the async implementation is 30% faster.  This advantage for async can be even greater when models exhibit poor solvability or being explored at edge of their stability.  
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/avdud/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/avdud/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">
# table updated with benchmark using latest implementation (8/14/2023)

Back-end | 0% of samples failed solves time   (s) | 50% of samples failed solves time   (s)
-- | -- | --
Serial | 28.43 | 360.89
ConcurrentFutures | 17.76 | 101.89
MultiProcessing | 17.59 | 59.17
RayIo | 19.5 | 65.14

</body>

</html>

Why RayIO- rayio provides an alternative to MPI for running simulations on network-connected clusters (where logical cores are located in separate multiple machines).  this alternative is needed as even on an MPI-supporting cluster, issues with running MPI can exist.  

-Adds multiprocessing and rayio parallel manager
-moves build and build output functions to config 
-removes old re-initialization routine, with build_model and initlize_function calls instead. 
-Further updates functions to decouple update of outputs and sweep_paramters using objects to model.find_component approach (pickle safe and model rebuild safe)

### Legal Acknowledgement


watertap/tools/parallel/parallel_manager_factory.py
```diff
@@ -38,7 +45,20 @@ def create_parallel_manager(parallel_manager_class=None, **kwargs):
...
+            return ConcurrentFuturesParallelManager(number_of_subprocesses)
+        elif parallel_backend == "MultiProcessing":
+            return MultiprocessingParallelManager(number_of_subprocesses)
+        elif parallel_backend == "RayIo":
+            try:
+                import ray
+
+                return RayIoParallelManager(number_of_subprocesses)
+            except:
+                print("Ray io not available, please install to use")
```
bknueven: This should be a proper `Exception`, or we should just let the `ImportError` be raised

watertap/tools/parallel/parallel_manager_factory.py
```diff
@@ -38,7 +45,20 @@ def create_parallel_manager(parallel_manager_class=None, **kwargs):
...
+        else:
+            raise NotImplementedError
```
bknueven: We should have a message here...e.g., 
```suggestion
            raise NotImplementedError(f"ParallelManager {parallel_backend} is not yet implemented")
```

watertap/tools/parallel/ray_io_parallel_manager.py
```diff
@@ -0,0 +1,196 @@
...
+        if ray.is_initialized() == False:
+            try:
+                # This will try to connect ot existing ray
+                # cluster, typical usage is for a cluster
+                # where ray needs to be started as head, with additional
+                # workers started on each node
+                # the parllel manaager will connect to this cluster
+                # useing ip_head addres of head node, and its password
+                # if these are not found it will reveret to local mode.
+                print(os.environ["ip_head"], os.environ["redis_password"])
```
bknueven: The `print` statements in this method should go to a logger instead.

watertap/tools/parallel/ray_io_parallel_manager.py
```diff
@@ -0,0 +1,196 @@
...
+        def excute_with_order(self, order_index, local_parameters):
+            local_parameters = local_parameters[order_index]
+            return (
+                order_index,
+                local_parameters,
+                self.execute(local_parameters),
+            )
+
+except:
+    pass
```
bknueven: We should instead have function which defines this class dynamically so we can do better error checking

watertap/tools/parallel/tests/test_parallel_manager.py
```diff
@@ -84,3 +84,30 @@ def test_multiple_subprocesses(self, number_of_subprocesses):
...
+        # verify that each subprocess did actual work
+        for i in range(len(execution_results)):
+            assert len(execution_results[i].parameters) > 0
+            assert len(execution_results[i].parameters) == len(
+                execution_results[i].results
+            )
+
+        # verify that the full list of results matches the expected
+        all_results = [r for result in execution_results for r in result.results]
+        assert all_results == [-101, -102, -103, -104, -105]
```
bknueven, line 122: No equivalent test for RayIO?

watertap/tools/parameter_sweep/parameter_sweep.py
```diff
@@ -20,9 +20,11 @@
 from idaes.core.solvers import get_solver
 
 from idaes.core.surrogate.pysmo import sampling
-from pyomo.common.collections import ComponentSet, ComponentMap
+
+# from pyomo.common.collections import ComponentSet, ComponentMap
```
bknueven:
```suggestion
```

watertap/tools/parameter_sweep/parameter_sweep.py
```diff
@@ -20,9 +20,11 @@
...
 from pyomo.common.deprecation import deprecation_warning
-from pyomo.common.tee import capture_output
+
+# from pyomo.common.tee import capture_output
```
bknueven:
```suggestion
```

watertap/tools/parameter_sweep/parameter_sweep.py
```diff
@@ -305,10 +400,21 @@ def _divide_combinations(self, global_combo_array):
...
         return local_combo_array
 
+    def _get_object(self, model, pyomo_object):
+        name = pyomo_object.name
+
+        # seems to be a bug, as indexed var with [None] exists
+        # but can't be found by find_component
+        if "[None]" in name:
+            name = name.replace("[None]", "")
+            return model.find_component(name)[None]
```
bknueven, line 422: This is because internally Pyomo uses `None` as an index for un-indexed objects.

Why do we have objects indexed by `None`?

watertap/tools/parameter_sweep/parameter_sweep.py
```diff
@@ -305,10 +400,21 @@ def _divide_combinations(self, global_combo_array):
...
+        else:
+            return model.find_component(name)
+
     def _update_model_values(self, m, param_dict, values):
         for k, item in enumerate(param_dict.values()):
-            param = item.pyomo_object
-
+            name = self._get_object(m, item.pyomo_object)
+            param = m.find_component(name)
```
bknueven, line 429: It seems like `_get_object` returns the component?
```suggestion
            param = self._get_object(m, item.pyomo_object)
```

watertap/tools/parameter_sweep/parameter_sweep.py
```diff
@@ -504,32 +613,48 @@ def _create_global_output(self, local_output_dict, req_num_samples=None):
...
+        # reinitialize_kwargs = self.config.reinitialize_kwargs
         run_successful = False  # until proven otherwise
-
         # Forced reinitialization of the flowsheet if enabled
-        if reinitialize_before_sweep:
-            if reinitialize_function is None:
+        if initialize_before_sweep or self.model_not_initialized:
+            if initialize_before_sweep and initialize_function is None:
                 raise ValueError(
                     "Reinitialization function was not specified. The model will not be reinitialized."
```
bknueven:
```suggestion
                    "Initialization function was not specified. The model will not be reinitialized."
```

watertap/tools/parameter_sweep/parameter_sweep.py
```diff
@@ -504,32 +613,48 @@ def _create_global_output(self, local_output_dict, req_num_samples=None):
...
-
+        build_model = self.config.build_model
+        build_model_kwargs = self.config.build_model_kwargs
+        initialize_function = self.config.initialize_function
+        initialize_kwargs = self.config.initialize_kwargs
+        initialize_before_sweep = self.config.initialize_before_sweep
+        update_sweep_params_before_init = self.config.update_sweep_params_before_init
+        # Depreciated
+        # reinitialize_function = self.config.reinitialize_function
+        # reinitialize_kwargs = self.config.reinitialize_kwargs
```
bknueven: The let's deprecate it -- remove these commented out lines and these options from the Config object.

watertap/tools/parameter_sweep/parameter_sweep.py
```diff
@@ -161,10 +254,12 @@ def __init__(
         parallel_manager_class = options.pop("parallel_manager_class", None)
 
         self.config = self.CONFIG(options)
-
+        self.model = None
+        self.model_not_initialized = True
```
bknueven: Can we switch the sense of this to not use a double-negative? I.e., call the variable `model_initialized`.

watertap/tools/parameter_sweep/tests/test_parameter_sweep.py
```diff
@@ -1030,11 +1024,14 @@ def test_parameter_sweep_recover(self, model, tmp_path):
                             ]
                         ),
                     },
+                    # removed upper bound tests, as in pracice
+                    # user should not be chanaing upper and lower bounds
+                    # during reinitilization and current frame work
+                    # this bounds get updated based on existing model in each worker
```
bknueven, line 1072: I agree that the bounds should not change during reinitialization.

Does this test still certify that the model *was* re-initialized? That's why this was written this way, so I could be sure that the re-initialize function was actually called.

watertap/tools/parameter_sweep/tests/test_parameter_sweep.py
```diff
@@ -1963,20 +1934,28 @@ def _bad_test_function(m):
...
+                        )
                         assert np.allclose(
                             test_dict[key][subkey]["value"],
                             subitem["value"],
                             equal_nan=True,
                         )
                     else:
+                        # print(
+                        #     key, subkey, subsubitem, test_dict[key][subkey][subsubkey]
+                        # )
```
bknueven:
```suggestion
```

watertap/tools/parameter_sweep/tests/test_parameter_sweep.py
bknueven: I think we should restore this aspect of the test to its previous behavior so we *know* that the initialization function was actually called.

bknueven (commented):
Lots of comments, most minor.

The two things I would like specifically addressed before merging are the creation of the RayIO `paramActor` class and the changes to `test_parameter_sweep.py` around initialization.

watertap/tools/parallel/tests/test_parallel_manager.py
avdudchenko: not until ray is part of WT.  (we should definetly have it once ray is part of WT reqs.)

watertap/tools/parallel/ray_io_parallel_manager.py
avdudchenko: I only put it into try /except block for now, as we don't have ray in wt, so the test fials on build due to the decorator. Once and if ray is part of reqs we can pull it out of try/except. (I will try adding a specific except error here).

watertap/tools/parameter_sweep/tests/test_parameter_sweep.py
avdudchenko: I plan to completely rework the way ps tool handles init states, and will also need to redo these tests. in its current form, the test don't really make sense as you would never try to solve an uninitialized model, and re-init just supposed to run a standard initialize sequence. So we should really just test that initi function worked, and pull out the build/init sequence into its own function to remove duplicity. I plan to update the logic a bit as well and will try adding tests that test the new implementation. 

and yes, currently it does test it, on a build i think slack is 1000 and on re-init its 10. so the test captures this change, if model failed to init, it would still be 1000.

watertap/tools/parallel/ray_io_parallel_manager.py
bknueven: We should never add ray as a hard dependency, IMHO.

I understand why these test fail.

```suggestion
# Spread ensures that load balancing makes all CPU's nodes work
# set it so it only uses 1 core per worekers as WT/IPOPT cant multiprocess
# might be not most efficienct on clusters with long data transfers...
def create_paramActor_class():
    @ray.remote(num_cpus=1)
    class paramActor(parallelActor):
        # this lets us track the order in execution
        def excute_with_order(self, order_index, local_parameters):
            local_parameters = local_parameters[order_index]
            return (
                order_index,
                local_parameters,
                self.execute(local_parameters),
            )
     return paramActor
```

Here is a potential work-around. This was `ray` only gets invoked when the function is called.

watertap/tools/parallel/tests/test_parallel_manager.py
bknueven: We should add it to the dev requirements.

watertap/tools/parallel/multiprocessing_parallel_manager.py
```diff
@@ -0,0 +1,144 @@
...
+    def gather(self):
+        results = []
+        # collect result from the actors
+        while len(results) < self.expected_samples:
+            if self.return_queue.empty() == False:
+                i, values, result = self.return_queue.get()
+
+                results.append(LocalResults(i, values, result))
+        # sort the results by the process number to keep a deterministic ordering
+        # results.sort(key=lambda result: result.process_number)
```
k1nshuk:
```suggestion
```

watertap/tools/parameter_sweep/parameter_sweep.py
```diff
@@ -511,124 +630,88 @@ def _create_global_output(self, local_output_dict, req_num_samples=None):
...
-                else:
-                    run_successful = True
-
-        else:
-            # If the simulation suceeds, report stats
-            run_successful = True
-
-        return run_successful
+        # or init if model was not initialized or prior solved failed (if solved failed, init state is false)
+        if initialize_before_sweep or self.model_manager.is_intilized == False:
```
k1nshuk:
```suggestion
        if initialize_before_sweep or self.model_manager.is_intialized == False:
```

watertap/tools/parameter_sweep/parameter_sweep.py
```diff
@@ -511,124 +630,88 @@ def _create_global_output(self, local_output_dict, req_num_samples=None):
...
+            self.model_manager.model,
             sweep_params,
             k,
             run_successful,
             local_output_dict,
         )
-
         return run_successful
 
     def _do_param_sweep(self, model, sweep_params, outputs, local_values):
```
k1nshuk:
```suggestion
    def _do_param_sweep(self, sweep_params, outputs, local_values):
```
I believe model is being no longer being used in the function.

watertap/tools/parameter_sweep/parameter_sweep_differential.py
```diff
@@ -330,6 +326,9 @@ def _run_differential_sweep(self, model, local_value):
             reinitialize_before_sweep=self.config.reinitialize_before_sweep,
             parallel_manager_class=SingleProcessParallelManager,
         )
+        # pass model_manager from refernce sweep, to diff sweep
+        # so we don't have to reijnit he model
+        diff_ps.model_manager = self.model_manager
 
         _, differential_sweep_output_dict = diff_ps.parameter_sweep(
             model,
```
k1nshuk: @bknueven @avdudchenko should we continue providing the model here given that we plan on deprecating passing a model in the future?

watertap/tools/parameter_sweep/parameter_sweep_differential.py
```diff
@@ -343,24 +342,19 @@ def _run_differential_sweep(self, model, local_value):
...
-            model,
-            reinitialize_values,
             local_value_k,
             k,
             sweep_params,
             local_output_dict,
         )
         self.differential_sweep_output_dict[k] = self._run_differential_sweep(
-            model, local_value_k
+            self.model_manager.model, local_value_k
```
k1nshuk: see previous comment

watertap/tools/parameter_sweep/parameter_sweep_differential.py
```diff
@@ -423,14 +469,15 @@ def parameter_sweep(
             local_results_dict = self._do_param_sweep(
                 model,
                 sweep_params,
-                outputs,
+                self.outputs,
```
k1nshuk, line 472: We could just get rid of this argument given that its contained within self.

watertap/tools/parameter_sweep/tests/test_differential_parameter_sweep.py
```diff
@@ -222,23 +223,22 @@ def test_bad_differential_sweep_specs(model, tmp_path):
         },
     }
 
-    A = m.fs.input["a"]
-    B = m.fs.input["b"]
-    sweep_params = {A.name: (A, 0.1, 0.9, 3), B.name: (B, 0.0, 0.5, 3)}
+    # A = m.fs.input["a"]
+    # B = m.fs.input["b"]
+    # sweep_params = {A.name: (A, 0.1, 0.9, 3), B.name: (B, 0.0, 0.5, 3)}
```
k1nshuk:
```suggestion
```

k1nshuk (approved):
Overall looks good. I have a few minor edits and suggestions. The differential parameter sweep test still explicitly uses MPI comm, but I think that can be replaced in a subsequent PR.

watertap/tools/parameter_sweep/parameter_sweep_differential.py
```diff
@@ -381,21 +375,73 @@ def _do_param_sweep(self, model, sweep_params, outputs, local_values):
...
-        sweep_params,
-        outputs=None,
+        build_model,
+        build_sweep_params,
+        build_outputs=None,
+        build_outputs_kwargs=dict(),
         num_samples=None,
         seed=None,
+        build_model_kwargs=dict(),
+        build_sweep_params_kwargs=dict(),
```
bknueven:
```suggestion
        build_outputs_kwargs=None,
        num_samples=None,
        seed=None,
        build_model_kwargs=None,
        build_sweep_params_kwargs=None,
```

We also need to add a conversion to `dict` for `build_outputs_kwargs` if it is `None`.

We never use mutable default arguments in Python, see, e.g., https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument

bknueven (changes requested):
We need to fix the mutable default arguments in differential parameter sweep's `parameter_sweep` method, and the `print` statement in the ray.io interface needs to be a logger so it can be silenced.

lbianchi-lbl:
It looks like the current CI failures are due to the default GHA 6 hours timeout being triggered:

image

Assuming the last test displayed before the time out cancellation notification is the one causing this, this should be it: https://github.com/avdudchenko/watertap/blob/276117ac5610b5128e2fdd63d8579ad4c294bbb1/watertap/tools/parallel/tests/test_multiprocessing_parallel_manager.py#L35-L42

bknueven:
> It looks like the current CI failures are due to the default GHA 6 hours timeout being triggered:
> 
> image
> 
> Assuming the last test displayed before the time out cancellation notification is the one causing this, this should be it: https://github.com/avdudchenko/watertap/blob/276117ac5610b5128e2fdd63d8579ad4c294bbb1/watertap/tools/parallel/tests/test_multiprocessing_parallel_manager.py#L35-L42

The thing that's a bit weird is it looks like the Windows tests got past that but hung at a different spot (but also in the `tools` tests). I've speculatively kicked off the Linux Python 3.10 test only to make sure the behavior is repeatable. If so, I'll try to reproduce it on my system.

bknueven (approved):

PR #1104, opened 2023-08-15
ED improvement

lbibl:
## Summary/Motivation:
Previously  the limiting current density in electrodialysis is calculated only when the Nernst diffusion layer (ndl) is simulated.   Although limiting current density does not directly control any ED performance when the ndl is not simulated, it can be beneficial to calculate the var in analyses.  This PR adds an config to control the calculation of limiting current density when ndl is not on.   No changes in actual computation but only model restructuring to get this improved feature. 
### Legal Acknowledgement


watertap/unit_models/electrodialysis_1D.py
```diff
@@ -195,6 +195,14 @@ class Electrodialysis1DData(InitializationMixin, UnitModelBlockData):
             description="The electrical operation mode. To be selected between Constant Current and Constant Voltage",
         ),
     )
+    CONFIG.declare(
+        "calculate_limiting_current_density",
```
adam-a-a, line 199: Wasn't involved in this discussion but any disadvantages to just have a "None" option for the limiting current density method (instead of a new config option)?

adam-a-a (commented):
Just two suggestions: 
1. Consider a none option for limiting current density method configuration instead of new Boolean config option.
2. Apply same changes to 0d as appropriate

watertap/unit_models/electrodialysis_1D.py
lbibl: Thanks for quick reviewing!   You mean combine this with another configuration?  I don't think that makes a technical difference.   But I would stay as-is because 1) this is clearer and 2) I don't think it's critical enough to forfeit the work I have done here.

hunterbarber (approved):
LGTM, the working implementation is fine, see Adam's comments for conventions

watertap/unit_models/electrodialysis_1D.py
adam-a-a: My thought was that it would leave the user with one less config option to be concerned with. Based on your tests, it looks like there are 22 config options, which seems over the top! It may work functionally, but the API/user experience should be considered as well.

 It's also a very simple change to make and would be clearer, IMHO.

adam-a-a: I'll add that it would be more analogous with other models as well (e.g., in RO, you can calculate concentration polarization effects or exclude; that is incorporated into a single config option, where `None` is used when we want to ignore consideration of concentration polarization.)

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -1758,6 +1672,104 @@ def eq_dl_thickness(self, mem, side, t, x):
...
                 )
 
+    def _get_limiting_current_density(self):
+        self.current_dens_lim_x = Var(
+            self.flowsheet().time,
+            self.diluate.length_domain,
+            initialize=500,
+            bounds=(0, 10000),
+            units=pyunits.amp * pyunits.meter**-2,
+            doc="Limiting Current Density accross the membrane as a function of the normalized length",
```
adam-a-a, line 1682:
```suggestion
            doc="Limiting current density across the membrane as a function of the normalized length",
```

adam-a-a (changes requested):
The deviation of the 1d model from 0d should be addressed. That is, the same changes made in 1D should be applied in the 0D model, unless there is some relevant reasoning behind excluding the additions from 0D.

I also would prefer if we didn't add this extra config option and consolidated into the existing config option for limiting current density.

watertap/unit_models/electrodialysis_1D.py
lbibl: If you insist 22 vs 21 config is critical,  I will close this and redo in the future.  This ED unit is large and linked with mutiple fs/analyses and such minor change needs careful inspection in multiple places so it may need longer time than you would think.  I don't know if that is the right/necesary analogy to RO.  I still think this is trivial and not worthy of more time that could be used to move forward much analysis work down the way.

lbibl:
> The deviation of the 1d model from 0d should be addressed. That is, the same changes made in 1D should be applied in the 0D model, unless there is some relevant reasoning behind excluding the additions from 0D.
> 
> I also would prefer if we didn't add this extra config option and consolidated into the existing config option for limiting current density.

0D is not excluded but also not prioritized, for it is not used in any places.

lbibl:
Attempted to combine the config.  Complications are encountered when limiting current density calculation is used internally by another function.  I am thus revoking this feature proposal and closing the PR.

PR #1115, opened 2023-08-25
Add concurrency group for GHA workflows to avoid redundant runs for PRs

lbianchi-lbl:
## Fixes/Resolves: PR congestion

This addresses a request/suggestion by @agarciadiego pointing out workflow runs "piling up" when commits are pushed to the same PR, which results in long queues, the need to manage runs manually, and/or wasted computation (for all runs except the one for the latest commit).

## Summary/Motivation:

- Once this is merged, when a workflow is triggered, in-progress runs of that workflow will be automatically canceled if the concurrency group is the same as the new workflow
- The concurrency group contains the workflow name, plus:
  - For PR triggers: the name of the PR's head branch (i.e. the "source" branch)
  - For non-PR triggers: the full commit SHA

## Caveats

- Canceled workflow runs will still trigger notifications; in other words, cancellations due to concurrency are not treated differently from other cancellations
- The new workflow run starts from the beginning, so the previous workflow run is still "wasted"
- With the current concurrency group settings, pushing to a non-PR branch will _not_ cancel in-progress runs for previous commits (since the commit SHA is different). This is on purpose so that every commit on `main` gets its own workflow run for Codecov

#### Immediately after pushing

- Existing runs in progress
- New (most recent) runs are queued (clock icon)

#### After a few seconds

- Previous runs are canceled automatically
- New runs start running (from the beginning)

### Legal Acknowledgement

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -28,6 +28,7 @@ env:
   # --color=yes needed for colorized output to be shown in GHA logs
   # --pyargs watertap is needed to be able to define CLI options in watertap/conftest.py
   PYTEST_ADDOPTS: "--color=yes"
+  PIP_PROGRESS_BAR: "off"
 
 jobs:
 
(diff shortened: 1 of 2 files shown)


bknueven (approved):
I'm impressed that this PR is self-referential.

PR #1120, opened 2023-09-06
Add NaCl recovery value to crystallizer model

Zhuoran29:
## Fixes/Resolves:

Added the recovery value of NaCl from the crystallization product

## Summary/Motivation:

- For economic analysis purpose, specifically in SETO project, it is necessary to account for the recovery value.
- In some cases the recovery value would result in a negative operating cost, which is not allowed in the current costing package. Associated change was made to it as well.

## Changes proposed in this PR:
- Changed the domain of `total_operating_cost` to `pyo.Reals`. 
- Update the solutions in three testing files that were impacted by the change above.
    - `watertap/examples/flowsheets/nf_dspmde/tests/test_nf.py`
    - `watertap/examples/flowsheets/full_treatment_train/analysis/tests/test_flowsheets.py`
    - `watertap/examples/flowsheets/full_treatment_train/flowsheet_components/chemistry/tests/test_posttreatment.py`
- Added a defined flow of `NaCl` in the crystallizer unit costing model 
- Added testing functions in the crystallizer tester file

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #1122, opened 2023-09-06
Unify costing APIs in WaterTAP

### LINKED ISSUE #676: WaterTAP costing improvements

### LINKED ISSUE #710: Add expression for total annualized cost to WT costing packages
Suggested by @bknueven in #709. The expression to be added is:

`costing.total_annualized_cost = costing.total_capital_cost * costing.capital_recovery_factor + m.fs.costing.total_operating_cost`

This is used in the default definition of LCOW and it is frequently used in other levelized cost metrics as well.

bknueven:
## Summary/Motivation:
See #676.

- [x] Creates a simple base class for costing, which extends the IDAES Flowsheet Costing Block
- [x] Moves duplicated functionality into the new base class
- [x] Adds documentation for this base class
- [x] Adds documentation for WaterTAPCosting Package and Utilities.
- [x] Both costing packages inherit from this base class
- [x] Removes some custom methods and tracking from the `WaterTAPCosting` package from when everything was constructed as a monolith.
- [x] Misc. updates to the code to keep up with these changes.
- [x] Add tests for using ZO models with WaterTAP Costing Package and high-fidelity models with the Zero-Order Costing Package

### Legal Acknowledgement


bknueven:
Let's wait for #1120 before merging this -- I would like to address this comment as well: https://github.com/watertap-org/watertap/pull/1120#issuecomment-1710603283

watertap/costing/watertap_costing_package.py
```diff
@@ -10,141 +10,31 @@
...
-
+class WaterTAPCostingData(WaterTAPCostingBlockData):
     def build_global_params(self):
         # Register currency and conversion rates based on CE Index
         register_idaes_currency_units()
 
+        self._build_common_global_params()
+
         # Set the base year for all costs
         self.base_currency = pyo.units.USD_2018
```
kurbansitterley, line 32: While we are making major changes to costing, can we change our default currency to be something more recent? 2021 is available as a unit, but probably not a good representative/default costing year since the pandemic made year/year inflation high, but 2020 or 2022 might be better. But maybe changing default parameter values is something you want to do in a separate PR.

watertap/costing/unit_models/ion_exchange.py
```diff
@@ -34,7 +34,7 @@ def build_hcl_cost_param_block(blk):
     )
 
     costing = blk.parent_block()
-    costing.add_defined_flow("HCl", blk.cost / blk.purity)
+    costing.register_flow_type("HCl", blk.cost / blk.purity)
```
kurbansitterley, line 37: This is unique to IX maybe, but is it possible to only register one of these material flows based on a CONFIG argument for the IX model? The model only requires one of these for costing, so having all just adds bloat

watertap/costing/watertap_costing_package.py
bknueven: Right now nothing in this PR changes the underlying models in any way, it's just a giant reshuffling and consolidation of existing code and functionality.

I would suggest putting off a change like this for when we merge the costing packages.

watertap/costing/unit_models/ion_exchange.py
bknueven: Not in a straightforward fashion, but of course anything can be done.

Right now the decorators ensure that the parameters are declared exactly once, even if two or more of the same unit is costed in a flowsheet.

The costing method doesn't need to use the decorators, however, to register the parameters. It could do it manually while ensuring they only get added once.

watertap/costing/watertap_costing_package.py
```diff
@@ -160,99 +50,24 @@ def build_global_params(self):
...
-        )
-        self.add_defined_flow("electricity", self.electricity_cost)
-
-        self.electrical_carbon_intensity = pyo.Param(
-            mutable=True,
-            initialize=0.475,
-            doc="Grid carbon intensity [kgCO2_eq/kWh]",
-            units=pyo.units.kg / pyo.units.kWh,
-        )
+        self.capital_recovery_factor.expr = self.factor_capital_annualization
```
kurbansitterley, line 53: This seems not ideal but I imagine will be rectified once we agree on how we want to handle which parameters to include and how they will be calculated/what their default values will be.

watertap/costing/costing_base.py
```diff
@@ -0,0 +1,239 @@
...
+        """
+        Allow the unit model to register its default costing method,
+        either through an attribute named "default_costing_method"
+        or by naming the default costing method "default_costing_method"
+        """
+        if hasattr(unit_model, "default_costing_method"):
+            return unit_model.default_costing_method
+        return super()._get_costing_method_for(unit_model)
+
+    def register_flow_type(self, flow_type, cost):
```
kurbansitterley, line 230: Can you explain what this is doing and why it is necessary? No issue with the code itself, just what it is doing in the larger framework here and why are units involved.

kurbansitterley (approved):
LGTM. I'd like to see further consolidation of costing packages, but that is for a future PR.

watertap/costing/watertap_costing_package.py
bknueven: Yeah, I agree it is a bit ugly. But it lets us put `add_LCOW` and similar aggregates in the base class.

watertap/costing/costing_base.py
bknueven: Ah, this goes back to something I think you found a while ago.

The IDAES costing framework will create a new `Var` in `register_flow_type`, unless there already is a component on the costing package named `flow_type + "_cost"`.

But this creates issues if you want to use mutable parameters or other variables to define the `cost`, then changes to those won't be reflected. There's a bit more discussion on the issue we're side-stepping here: https://github.com/IDAES/idaes-pse/pull/1014#issuecomment-1309527271.

bknueven: As for the units, it has to do with the fact that Pyomo treats units as expressions. So, e.g., we want `register_flow_type( "foo", 42.0 * pyo.units.USD_2020 / pyo.units.m )` to go ahead and create a new variable called `foo_cost` whose value is `42.0 * pyo.units.USD_2020 / pyo.units.m`.

But if I define a parameter or variable named `bar_base_cost` and another parameter called `bar_purity`, then `register_flow_type( "bar", bar_base_cost / bar_purity)`, we will *not* create a new `Var`, but instead create a new `Expression` called `bar_cost` with `bar_cost.expr = bar_base_cost / bar_purity`. It is assumed that if the user wants to do parameter sweeps over the bar_cost the user will modify `bar_base_cost`.


There is a proposed fix for some of this ugliness, but it hasn't been a priority for IDAES/Pyomo: https://github.com/IDAES/idaes-pse/pull/1014#issuecomment-1332709400

watertap/costing/unit_models/electrodialysis.py
```diff
@@ -55,7 +55,10 @@ def cost_electrodialysis(blk, cost_electricity_flow=True, has_rectifier=False):
     Function for costing the Electrodialysis unit
 
     Args:
-        cost_electricity_flow - Option for including the costing of electricity
+        cost_electricity_flow (:obj:`bool`, optional): Option for including the
+            costing of electricity. Defaults to True.
+        has_retifier (:obj:`bool`, optional): Option for including a rectifier.
```
adam-a-a:
```suggestion
        has_rectifier (:obj:`bool`, optional): Option for including a rectifier.
```

watertap/core/util/misc.py
```diff
@@ -0,0 +1,31 @@
...
+    Returns:
+        bool : True if the expr is constant besides units, False otherwise
+    """
+    for v in identify_variables(expr):
+        return False
+    for p in identify_mutable_parameters(expr):
+        if isinstance(p, _PyomoUnit):
+            continue
+        return False
+    return True
```
adam-a-a, line 31: Can you explain the logic behind this function? If variables are identified in an expression, or if any mutable params in the expression don't have units, we deem it as not being constant--that's what I'm interpreting but still not sure and also not absorbing the purpose.

watertap/core/util/tests/test_misc.py
```diff
@@ -0,0 +1,95 @@
...
+
+    # test non-mutable param with unit-only expression
+    assert is_constant_up_to_units(m.fixed_param * units.m**2 / units.s)
+
+    m.fixed_param_2 = Param(initialize=6.28)
+
+    # test expression of fixed params with units
+    assert is_constant_up_to_units(
+        (m.fixed_param**2) * m.fixed_param_2 * units.m**2
+    )
```
adam-a-a, line 46: OK so this returns True because it is an immutable Param, and if it were mutable, it would return False -- so it seems.

watertap/core/util/tests/test_misc.py
```diff
@@ -0,0 +1,95 @@
...
+
+    m.mutable_param = Param(initialize=42, mutable=True)
+
+    # test mutable param
+    assert not is_constant_up_to_units(m.mutable_param)
+
+    m.mutable_param_units = Param(initialize=42, units=units.m**2 / units.s)
+    # test mutable param with units specified
+    assert not is_constant_up_to_units(m.mutable_param_units)
```
adam-a-a, line 55: Does Pyomo automatically designate a Param as mutable when units are assigned? If not, then I don't get this.

watertap/costing/costing_base.py
```diff
@@ -35,6 +35,188 @@ class WaterTAPCostingBlockData(FlowsheetCostingBlockData):
...
+                pyo.units.convert(
+                    flow_rate, to_units=pyo.units.m**3 / self.base_period
+                )
+                * self.utilization_factor
+            ),
+            doc=f"Constraint for Levelized Cost of Water based on flow {flow_rate.name}",
+        )
+        self.add_component(name + "_constraint", LCOW_constraint)
+
+        self._registered_LCOWs[name] = (LCOW, LCOW_constraint)
```
adam-a-a, line 73: I don't think it would work with the existing formulation, but I think we eventually want to be able to get the LCOW breakdown, which would yield levelized cost components for capex and opex across a flowsheet. Just something to think about because it would be extremely useful to have at our fingertips. I manually did this in the LSRRO flowsheet when we working on that analysis.

watertap/costing/costing_base.py
```diff
@@ -35,6 +35,188 @@ class WaterTAPCostingBlockData(FlowsheetCostingBlockData):
...
+
+    def add_specific_energy_consumption(
```
adam-a-a, line 75: SImilarly, it'd be useful to get a full SEC breakdown by component in addition to this aggregate SEC for a model

watertap/costing/costing_base.py
```diff
@@ -35,6 +35,188 @@ class WaterTAPCostingBlockData(FlowsheetCostingBlockData):
...
+        )
+
+    def add_electricity_intensity(self, flow_rate, name="electricity_intensity"):
+        """
+        Add calculation of overall electricity intensity to costing block.
+        Args:
+            flow_rate - flow rate of water (volumetric) to be used in
+                        calculating electricity intensity
+            name (optional) - the name of the Expression for the specific
+                              electrical intensity (default: specific_electrical_carbon_intensity)
```
adam-a-a, line 127: I was never a big fan of the "electricity intensity" usage for the WT3 migration, and I think as we move towards consolidating into one package, we should reduce to specific energy consumption (which technically could be with respect to electricity or thermal energy, so we'd have to consider that as well).

watertap/costing/costing_base.py
```diff
@@ -35,6 +35,188 @@ class WaterTAPCostingBlockData(FlowsheetCostingBlockData):
...
+        """
+        self.add_specific_energy_consumption(flow_rate, name=name)
+
+    def add_specific_electrical_carbon_intensity(
+        self, flow_rate, name="specific_electrical_carbon_intensity"
+    ):
+        """
```
adam-a-a, line 134: As we consolidate packages, we may want to consider using wordage that aligns with the national decarbonization effort (e.g., CO2_equivalent_footprint or whatever is ubiquitously used nowadays).

watertap/costing/costing_base.py
```diff
@@ -35,6 +35,188 @@ class WaterTAPCostingBlockData(FlowsheetCostingBlockData):
...
+            doc="Total capital cost of the process",
+            units=self.base_currency,
+        )
+        self.total_operating_cost = pyo.Var(
+            initialize=0,
+            doc="Total operating cost of process per operating period",
+            units=self.base_currency / self.base_period,
+        )
+
+    def _build_common_global_params(self):
```
adam-a-a, line 174: So replaced `practical_investment_factor` with  `TIC`? I'd be good with that.

watertap/costing/watertap_costing_package.py
```diff
@@ -96,6 +96,14 @@ def build_process_costs(self):
                 + sum(self.aggregate_flow_costs.values()) * self.utilization_factor
             )
 
+        self.total_annualized_cost = pyo.Expression(
+            expr=(
+                self.total_capital_cost * self.capital_recovery_factor
```
adam-a-a, line 101: Similar to my earlier comments for future work, with regard to a breakdown of LCOW cost components, would be great to have an `annualized_cost` attribute for every capital costs. With that, all we'd need to do is then divide by the desired flowrate.

docs/technical_reference/costing/costing_base.rst
```diff
@@ -0,0 +1,175 @@
...
+Specific Energy Consumption
++++++++++++++++++++++++++++
+
+For a given volumetric flow `Q`, the specific energy consumption, :math:`E_{spec,Q}` is calculated by the `add_specific_energy_consumption` method as
+
+    .. math::
+    
+        E_{spec,Q} = \frac{C_{el,tot}}{Q}
+
+Specific Electrical Carbon Intesity
```
adam-a-a: "Intensity"

docs/technical_reference/costing/costing_base.rst
```diff
@@ -0,0 +1,175 @@
...
+=============================================  ====================  =====================================  ===============  ==============================================================================
+Plant capacity utilization factor                 :math:`f_{util}`    ``utilization_factor``                 90%                Percentage of year plant is operating
+Electricity price                                 :math:`P`           ``electricity_cost``                   $0.07/kWh          Electricity price in 2018 USD
+Electricity carbon intensity                      :math:`f_{eci}`     ``electrical_carbon_intensity``        0.475 kg/kWh       Carbon intensity of electricity
+Capital recovery factor                           :math:`f_{crf}`     ``capital_recovery_factor``            None               Calculated by derived class 
+=============================================  ====================  =====================================  ===============  ==============================================================================
+
+
+Aggregates calculation
```
adam-a-a: "Aggregate Metrics"

docs/technical_reference/costing/watertap_costing.rst
```diff
@@ -0,0 +1,168 @@
+.. _watertap_costing:
+
+WaterTAP Costing Package
+========================
+
+.. currentmodule:: watertap.costing.watertap_costing
+
+The WaterTAP costing module contains the costing package typically used for high-fidelity models, but it can also be utilized with zero-order models. Technoecomic data is set on the package or utilizing global parameters added by the unit model costing method.
```
adam-a-a: "Technoeconomic data"

docs/technical_reference/costing/watertap_costing.rst
```diff
@@ -0,0 +1,168 @@
...
+=============================================  ====================  ===============================================  ==============================================================================
+
+Calculations for each of these costs are presented below.
+
+Costing Index and Technoeconomic Factors
+----------------------------------------
+
+Default costing indices are provide with the WaterTAP Costing Package, but the user is free to modify these for their needs.
+
+WaterTAP uses the Consumer Price Index (CPI) to help account for the time-value of investments and are used in the capital
```
adam-a-a: Hmm.. I thought we used the Chemical Engineering Plant Cost Index (CEPCI)

watertap/costing/tests/test_costing_interop.py
```diff
@@ -0,0 +1,178 @@
...
+    m.fs.costing.LCOW_with_revenue = pyo.Expression(
+        expr=(
+            (
+                m.fs.costing.total_annualized_cost
+                + m.fs.costing.aggregate_flow_costs["ferric_chloride"]
+            )
+            / m.fs.costing.annual_water_production
+        ),
+        doc="Levelized Cost of Water With Revenue",
+    )
```
adam-a-a, line 80: For the future, I wonder if we want to build in revenue as a separate aggregate that can be switched on or off as desired and incorporated (or excluded) from LCOW calculations, without have to do so manually like so.

adam-a-a (commented):
This looks great. I have some minor suggestions (mostly typo fixes) and questions on parts that I just wasn't clearly understanding (even though I assume they work as intended). 

EDIT: I also had several comments on things we might want to consider in the future.

adam-a-a (approved):
@bknueven I am going to go ahead and approve this so that we can merge ASAP after typos addressed.

docs/technical_reference/costing/watertap_costing.rst
bknueven: You are correct. I copied this sentence from the ZO costing documentation and fixed it in both places.

watertap/core/util/misc.py
bknueven: Yes, more or less correct. This function checks to see if there are any non-constant terms (e.g., Vars, mutable Params) in the expression. Unfortunately, Pyomo's existing checks methods do not cover this use case.

As for the purpose, I will direct you this comment: https://github.com/watertap-org/watertap/pull/1122#discussion_r1320341507

watertap/core/util/tests/test_misc.py
bknueven: Yes.

watertap/core/util/tests/test_misc.py
bknueven: Actually, this is a good question. I have to admit that I wrote this test several months ago.

bknueven: Apparently so. In fact, you can't set a `Param` with units to be *not* mutable:
```
In [9]: m.param_units_2 = Param(initialize=42, units=units.m**2 / units.s, mutable=False )

In [10]: m.param_units_2.mutable
Out[10]: True
```

This seems like a bug. I've created an issue in Pyomo (https://github.com/Pyomo/pyomo/issues/3002).

andrewlee94: @bknueven No, this is not a bug. Due to how `Params` work, if they are given units then they must be mutable.

watertap/costing/costing_base.py
bknueven: They both still exist -- would need further refinement to eliminate one or the other.

bknueven:
Thanks for the detailed comments @adam-a-a. I have captured them in #1147 so we don't lose track of them.

PR #1125, opened 2023-09-08
BSM2 GUI

MarcusHolly:
## Summary/Motivation:
Creates the UI files for the Benchmark Simulation Model No.2 (BSM2). Costing is currently omitted since the BSM2 flowsheet has yet to incorporate it.

### Legal Acknowledgement


MarcusHolly:
Images of the GUI for reviewer convenience:

image

image

MarcusHolly:
@MichaelPesce and @dangunter I've tagged you guys because I recall there is something you have to do on the backend of the UI repo. Modifying the setup.py file I think? Or maybe the workflow has become more automatic since last time. Regardless, there will probably be a few more GUI PRs over the next couple of weeks, so it may be more convenient for you to wait and add them all at once, but I just wanted to give you guys a heads-up.

watertap/examples/flowsheets/case_studies/full_water_resource_recovery_facility/BSM2_ui.png
adam-a-a, line 1: This is a start, but we probably want to include all the complexity of the recycle streams, which will make it messier but more accurate. Just a note to consider for now- don't want you to spend time an revamping this just yet.

adam-a-a:
@agarciadiego I don't know if your Jupyter Notebook PR will include any changes that would apply to files in this PR. Let's make sure to coordinate both efforts to avoid difficulties

lbianchi-lbl:
@MarcusHolly As a reminder: make sure to add an entry (and/or commit the changes) to the `watertap.flowsheets` entry point in `setup.py`: https://github.com/watertap-org/watertap/blob/c49bf30c378f83f72331b6424025b9bd7fc52952/setup.py#L142-L157

MarcusHolly:
> @MarcusHolly As a reminder: make sure to add an entry (and/or commit the changes) to the `watertap.flowsheets` entry point in `setup.py`:
> 
> https://github.com/watertap-org/watertap/blob/c49bf30c378f83f72331b6424025b9bd7fc52952/setup.py#L142-L157

Oh, thanks for pointing this out @lbianchi-lbl. In the past, I think Dan or Mike had to make this change since the entry points in setup.py were handled in the watertap-ui repo instead of this one. Regardless, you are correct, so I'll make this change.

adam-a-a:
> > @MarcusHolly As a reminder: make sure to add an entry (and/or commit the changes) to the `watertap.flowsheets` entry point in `setup.py`:
> > https://github.com/watertap-org/watertap/blob/c49bf30c378f83f72331b6424025b9bd7fc52952/setup.py#L142-L157
> 
> Oh, thanks for pointing this out @lbianchi-lbl. In the past, I think Dan or Mike had to make this change since the entry points in setup.py were handled in the watertap-ui repo instead of this one. Regardless, you are correct, so I'll make this change.

@lbianchi-lbl I just had a thought about documenting this for any user wanting to add to the GUI (hypothetically). I wonder if we could streamline this process by having some function in our UI export code that adds that new addition to some master list, and have setup.py just load that list rather than list the entries manually. Does that make sense? If so, any thoughts on that?

lbianchi-lbl:
> @lbianchi-lbl I just had a thought about documenting this for any user wanting to add to the GUI (hypothetically). I wonder if we could streamline this process by having some function in our UI export code that adds that new addition to some master list, and have setup.py just load that list rather than list the entries manually. Does that make sense? If so, any thoughts on that?

In general I can see the point of not having to go through `setup.py`. However there are a few tradeoffs/subtleties to consider. I'm adding it to the agenda for discussion.

adam-a-a:
@MichaelPesce can you remind us how to add in parameter sweep capabilities?

MichaelPesce:
@adam-a-a you don't need to add anything extra on the export side for parameter sweep. it'll be available as a toggle on the UI as long as there's a solve function

adam-a-a:
@MichaelPesce Ah, I see- I was under the impression that we had to do something special since I recall "sweep" not showing up for the latest dye desal GUI example (#1126). Maybe we're missing something there though I thought a solve func was in place

MichaelPesce:
@adam-a-a yeah I'm pretty sure the solve function is required to make the flowsheet show up. looking at the screenshots from the dye desal example you referenced, that appears to be an older version of the UI

MarcusHolly:
@MichaelPesce Is it a complicated process to update my GUI to the latest version (i.e. how do I go about updating it)?

MichaelPesce:
@MarcusHolly it shouldn't be too complicated. 
first, you'll need to pull the latest code from the watertap-ui repo. 
then, install any new backend packages (this might not be necessary, but just in case). assuming you are in the watertap-ui directory: 
`cd backend && pip install -r requirements.txt`
last, install any new frontend packages: 
`cd electron/ui && npm install`

adam-a-a (approved):
LGTM - great job on this!

luohezhiming (approved):
Looks great!

PR #1126, opened 2023-09-08
Dye Desalination Parameter Update

MarcusHolly:
## Summary/Motivation:
This PR is essentially a clone of #1092, but it removes all of the visualization tools. The hope is that we can get this PR merged relatively quickly such that #1092 will become a PR to add in the visualization tools for WaterTAP. See issue #1127 for more details.

- Update various parameters related to the dye desal flowsheets
- Set up a config option to toggle pre-treatment on/off in the dye desal flowsheet with RO

### Legal Acknowledgement


watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
```diff
@@ -165,36 +166,52 @@ def add_costing(m):
...
+                -1
+                * m.fs.zo_costing.utilization_factor
+                * m.fs.zo_costing.dye_disposal_cost
+                * pyunits.convert(
+                    m.fs.dye_retentate.properties[0].flow_vol,
+                    to_units=pyunits.m**3 / m.fs.zo_costing.base_period,
+                )
+            ),
+            doc="Cost of disposing of dye waste",
+        )
```
adam-a-a: Technically, couldn't we just have a single paramreter for dye_value (or cost) and have it go negative when we want to consider recovery value of dye and set it positive when we want to consider it as a disposal cost? Maybe we talked about this already, but at the moment, I am thinking that this can be consolidated (i.e., maybe we don't even need the Boolean arg anymore)

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_ui.py
```diff
@@ -163,18 +163,18 @@ def export_variables(flowsheet=None, exports=None):
...
+    #     obj=fs.pretreatment.wwtp.energy_electric_flow_vol_inlet,
+    #     name="Specific energy consumption per inlet flow rate",
+    #     ui_units=pyunits.kWh / pyunits.m**3,
+    #     display_units="kWh/m3",
+    #     rounding=2,
+    #     description="Electrical energy consumption with respect to influent volumetric flow rate",
+    #     is_input=True,
+    #     input_category="Secondary Wastewater Treatment",
+    #     is_output=False,
+    # )
```
adam-a-a: Why not add a check like `if hasattr(fs, 'pretreatment')`, run this code that you currently have commented out? Else, skip it. It might not be useful right away, but maybe it can work out once the UI is updated (https://github.com/watertap-org/watertap-ui/issues/85)

adam-a-a: @dangunter @MichaelPesce curious about your thoughts on implementing something like this bit as an example when considering config options for the UI

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
MarcusHolly: @adam-a-a The reason I didn't use a single parameter here is because the dye disposal cost ($120/m3) is not equivalent to the value of recovered dye ($0.12/m3), so I think there's no way to get around having two separate expressions.

MarcusHolly:
(Updated 9/11/23 1:43pm EST) GUI Images for reviewer convenience:
image
image
image

MarcusHolly (commented):
Another potential improvement that would go along with https://github.com/watertap-org/watertap-ui/issues/85 is incorporating the config option for `has_dye_revenue=False` or `has_dye_revenue=True`. For now, the GUI just displays the dye revenue value and dye disposal cost (under "System Costs"), but it should really only display the cost parameter associated with the specified config option.

Edit: After further discussion, this won't be necessary anymore as this config option has been removed.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
adam-a-a: Right, but wouldn't we only consider one case or the other? In other words, would we consider a scenario with dye disposal cost AND dye recovery value?

adam-a-a:
> GUI Images for reviewer convenience: image image image

Thanks for showing this. I don't think the secondary wastewater treatment section should show up anymore since our default is to exclude secondary wastewater treatment (pretreatment). Similarly, we should update the figure so that it reflects the flowsheet (no pretreatment, either Pump+NF with streams going to disposal and or recovery to textile plant, or pump/nf/ro with disposal/value stream shown. The former is probably the priority to show if we want to start with one.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
MarcusHolly: No, we wouldn't ever consider a case with dye disposal cost and dye recovery. I'm still not understanding how we can consolidate these two costs into one parameter though. I actually left a GUI-related comment about this very issue...

MarcusHolly:
> > GUI Images for reviewer convenience: image image image
> 
> Thanks for showing this. I don't think the secondary wastewater treatment section should show up anymore since our default is to exclude secondary wastewater treatment (pretreatment). Similarly, we should update the figure so that it reflects the flowsheet (no pretreatment, either Pump+NF with streams going to disposal and or recovery to textile plant, or pump/nf/ro with disposal/value stream shown. The former is probably the priority to show if we want to start with one.

The secondary treatment only shows up because I was testing the GUI functionality to make sure it worked with `hasattr(fs, "pretreatment`). The default config option for this GUI is `include_pretreatment=False`, so the secondary treatment section won't show up unless the user goes in and changes this to True. But I will update the figure when I get the chance

adam-a-a:
> Another potential improvement that would go along with [watertap-org/watertap-ui#85](https://github.com/watertap-org/watertap-ui/issues/85) is incorporating the config option for `has_dye_revenue=False` or `has_dye_revenue=True`. For now, the GUI just displays the dye revenue value and dye disposal cost (under "System Costs"), but it should really only display the cost parameter associated with the specified config option.

@MarcusHolly I'm still not sure we need to have `has_dye_revenue` since the expression created in either case ( `has_dye_revenue` = True or False) is identical and the only thing that changes is the cost of dye. I am very likely missing something, but like I mentioned before, I figure we could have negative values for dye_cost represent value that would lead to revenue, and if we are considering dye disposal costs, then dye_cost would be represented by positive values. For example:

If dye revenue counted:
`dye_cost = -0.12`
If dye revenue not counted and disposal considered instead:
`dye_cost =120`

MarcusHolly:
Move dye cost from the revenue section to the operating costs

adam-a-a (approved):
LGTM

agarciadiego (approved):
LGTM

PR #1132, opened 2023-09-13
Skk model

zacharybinger:
## Fixes/Resolves:

Added the option for using the Spieler-Kedem-Katchalsky (SKK) model. The option is implemented in the config options for the membrane_channel_base.py so it can be used in any applicable unit models.

## Summary/Motivation:
The SKK model should more appropriately reflect the water and salt transport across "leaky" membranes

- Modified RO base class to include a config option for the SKK model transport model 
- Modified unit tests for RO-1D and RO-0D to account for the addition of the SKK model

### Legal Acknowledgement


watertap/core/membrane_channel_base.py
```diff
@@ -233,6 +233,20 @@ class FrictionFactor(Enum):
     ),
 )
 
+CONFIG_Template.declare(
+    "transport_model",
+    ConfigValue(
+        default="soultion-diffusion",
```
```suggestion
        default="solution-diffusion",
```
I just wondered how the tests would pass if this typo were here, but then I realized that you need to add a couple of new tests for 0DRO, 1DRO, and probably OARO 0D to check that the formulation gets you to expected results.

adam-a-a: Additionally, I think you need to add an enumerated class for transport_model and include the `solution_diffusion` and `SKK` options there. Then you would add that class in here as the domain and also describe SKK in docstring.

watertap/unit_models/reverse_osmosis_base.py
```diff
@@ -370,6 +370,64 @@ def eq_flux_mass(b, t, x, p, j):
                     - prop_perm.conc_mass_phase_comp[p, j]
                 )
 
+        if self.config.transport_model == "SKK":
+
+            self.eq_flux_mass.deactivate()
```
adam-a-a: I was thinking this implementation would be slightly different in that we would just use the existing `eq_flux_mass` constraint and add refl_coeff bits where needed. Then, you'd have a conditional to check whether SKK or sol-diff, and if sol-diff, fix refl_coeff to 1. To summarize, I was thinking we could use the one flux constraint for both transport models since refl_coeff = 1 in solution-diffusion (so long as it doesn't cause instability and numerous tests associated with RO to fail).

watertap/unit_models/reverse_osmosis_base.py
```diff
@@ -370,6 +370,64 @@ def eq_flux_mass(b, t, x, p, j):
...
+
+            self.reflect_coeff = Var(
```
adam-a-a: I figured this could be added in whether it's SKK or sol-diff. For SKK, leave as unfixed var. For sol-diff, fix to 1.

watertap/unit_models/reverse_osmosis_base.py
```diff
@@ -370,6 +370,64 @@ def eq_flux_mass(b, t, x, p, j):
...
+                initialize=1,
+                domain=NonNegativeReals,
+                units=pyunits.dimensionless,
+                doc="Reflection coefficient of the membrane",
+            )
+
+            self.alpha = Var(
```
adam-a-a: Maybe this one should only be created with SKK enabled (and not with sol-diff, since it would just be 0). Or maybe we do include for both transport models and alpha would just be 0.

watertap/unit_models/reverse_osmosis_base.py
```diff
@@ -370,6 +370,64 @@ def eq_flux_mass(b, t, x, p, j):
...
+                return b.alpha == (1 - b.reflect_coeff) / b.B_comp[t, j]
+
+            @self.Constraint(
+                self.flowsheet().config.time,
+                self.difference_elements,
+                self.config.property_package.phase_list,
+                self.config.property_package.component_list,
+                doc="Solvent and solute mass flux using SKK model",
+            )
+            def eq_flux_mass_skk(b, t, x, p, j):
```
adam-a-a: Just to echo earlier comments, I figured this constraint wouldn't need to be built anew. Rather, we just use this single constraint (rename it back to eq_flux_mass) for both transport models, so long as numerous tests don't fail due to unintuitive numerical issues.

watertap/unit_models/tests/test_osmotically_assisted_reverse_osmosis_0D.py
```diff
@@ -67,7 +67,7 @@ def test_config():
     m.fs.properties = props.NaClParameterBlock()
     m.fs.unit = OsmoticallyAssistedReverseOsmosis0D(property_package=m.fs.properties)
 
-    assert len(m.fs.unit.config) == 13
+    assert len(m.fs.unit.config) == 14
```
adam-a-a, line 70: I guess one question to consider is whether SKK is applicable to OARO. Never thought about it before myself...

adam-a-a: In any case, we should have tests to demonstrate successful model creation, scaling, initialization, and solve, following other test examples to make sure nothing breaks.

watertap/unit_models/tests/test_reverse_osmosis_0D.py
```diff
@@ -66,7 +66,7 @@ def test_config():
     m.fs.properties = props.NaClParameterBlock()
     m.fs.unit = ReverseOsmosis0D(property_package=m.fs.properties)
 
-    assert len(m.fs.unit.config) == 13
+    assert len(m.fs.unit.config) == 14
```
adam-a-a, line 72: Add tests to demo successful model creation, scaling, initialization, and solve, following other test examples for 0D and 1D.

adam-a-a (changes requested):
I have a few requested changes, the most important being to add tests that use the new config option for RO0D, 1D, and OARO 0D. I suspect there will be some issues to resolve.  Happy to discuss in more detail.

One more thing that should go in this PR is that the documentation file for RO and OARO should account for any additions/revisions made in the model(s). However, maybe we should make sure everything is good with this PR first, then follow-up with a subsequent PR for just the doc revisions, after we've finalized model changes.

adam-a-a:
Just an additional comment regarding B and omega: should we be dividing B by R*T when using SKK formulation (to represent inclusion of omega)? I might be mistaken in that we only need B but interested to hear your thoughts.

watertap/core/membrane_channel_base.py
zacharybinger: Addressed in commit c6bc666

watertap/unit_models/reverse_osmosis_base.py
zacharybinger: Addressed in commit d914a77

watertap/unit_models/reverse_osmosis_base.py
zacharybinger: Addressed in commit d914a77

watertap/core/membrane_channel_base.py
```diff
@@ -60,6 +60,15 @@ class MassTransferCoefficient(Enum):
     calculated = auto()
     # TODO: add option for users to define their own relationship?
 
+class TransportModel(Enum):
+    """
+    none: mass transfer coefficient not utilized for concentration polarization effect
+    fixed: mass transfer coefficient is a user specified value
+    calculated: mass transfer coefficient is calculated
```
adam-a-a: I think the docstring needs to be updated.

watertap/core/membrane_channel_base.py
```diff
@@ -233,6 +242,24 @@ class FrictionFactor(Enum):
...
+        default=TransportModel.SD,
+        domain=In(TransportModel),
+        description="Mass transfer model in RO feed channel",
+        doc="""
+        Options to account for mass transfer model.
+
+        **default** - ``TransportModel.SD``
+
+        "``TransportModel.SD``", "Solution-diffusion model for describing water and salt transport for most membrane types"
+        "``TransportModel.SKK``", "Speigler-Kedem-Katchalsky model for describing water and salt transport for 'leaky' membranes"
```
```suggestion
        "``TransportModel.SKK``", "Speigler-Kedem-Katchalsky model for describing water and salt transport"
```

watertap/unit_models/tests/test_reverse_osmosis_0D.py
zacharybinger: Should I create a new class for this or should I just add tests to the TestReverseOsmosis class?

watertap/unit_models/reverse_osmosis_base.py
```diff
@@ -343,33 +359,82 @@ def _add_flux_balance(self):
...
-                        interface.pressure_osm_phase[p]
-                        - prop_perm.pressure_osm_phase[p]
+        if self.config.transport_model == TransportModel.SD:
+
+            @self.Constraint(
+                self.flowsheet().config.time,
+                self.difference_elements,
+                self.config.property_package.phase_list,
+                self.config.property_package.component_list,
+                doc="Solvent and solute mass flux using SKK model",
```
adam-a-a:
```suggestion
                doc="Solvent and solute mass flux using SD model",
```

watertap/unit_models/reverse_osmosis_base.py
```diff
@@ -737,6 +802,13 @@ def calculate_scaling_factors(self):
         if iscale.get_scaling_factor(self.recovery_vol_phase) is None:
             iscale.set_scaling_factor(self.recovery_vol_phase, 1)
 
+        if self.config.transport_model == TransportModel.SKK:
```
adam-a-a, line 804: Could also handle alpha and reflect_coeff by checking they exist with `hasattr`, but this works too.

watertap/unit_models/reverse_osmosis_base.py
```diff
@@ -737,6 +802,13 @@ def calculate_scaling_factors(self):
...
+            if iscale.get_scaling_factor(self.alpha) is None:
+                iscale.set_scaling_factor(self.alpha, 1e-8)
```
adam-a-a, line 806: I am still getting accustomed to alpha. Is it typically on the order of 1e8? Do we have a decent idea of how much it could vary?

watertap/unit_models/tests/test_reverse_osmosis_0D.py
adam-a-a: Looks like you handled it!

watertap/unit_models/tests/test_osmotically_assisted_reverse_osmosis_0D.py
adam-a-a: I am still uncertain about this one... if we don't think SKK should be part of OARO, maybe we should disable it for OARO. Otherwise, I guess we can create tests for OARO with SKK,

adam-a-a (commented):
This is looking good. I think there are just a couple things left to consider before we merge.

watertap/unit_models/reverse_osmosis_base.py
zacharybinger: So alpha is approx 1/B and since B is initialized as 1e-8 I think initializing alpha as 1e8 is a good assumption. I don't think it's necessarily quite that simple since Menachem's students talked about characterizing membranes by alpha instead of B, but I think it is a good starting point.

watertap/unit_models/tests/test_osmotically_assisted_reverse_osmosis_0D.py
zacharybinger: Based on our discussion yesterday, it will be a part of OARO but tests will be added in a separate PR

luohezhiming (approved):
Looks good to me! We can verify SKK model's validity with OARO in future PRs.

adam-a-a (approved):
LGTM- we can add tests for SKK usage in OARO model in a subsequent PR. Also, we'll need to modify the technical reference documentation for RO in another PR as well.

adam-a-a:
I added two issues to track remaining items: #1189 #1190

PR #1133, opened 2023-09-13
Solve ASM2D Flowsheet without bound_push and autoscale function

adam-a-a:
## Summary/Motivation:
An existing flowsheet example for the conventional ASM2D model uses bound_push and an autoscale function. Neither should be used to solve the model in the main codebase. Moreover, the modified version of ASM2D, a trickier variant of conventional ASM2D, will subsequently be implemented in a similar flowsheet example, but we should make sure the original version works first.

**TODO:** 

- [ ] Adjust regression test values and review to make sure we accept the variations in results

- remove bound_push
- remove autoscaling

### Legal Acknowledgement

### DIFF
--- a/watertap/examples/flowsheets/case_studies/activated_sludge/ASM2D_flowsheet.py
+++ b/watertap/examples/flowsheets/case_studies/activated_sludge/ASM2D_flowsheet.py
@@ -49,6 +49,7 @@ from watertap.property_models.activated_sludge.asm2d_properties import (
 from watertap.property_models.activated_sludge.asm2d_reactions import (
     ASM2dReactionParameterBlock,
 )
+from idaes.models.unit_models.mixer import MomentumMixingType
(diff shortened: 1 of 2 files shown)


adam-a-a:
Note: initialization and solve was working last I checked. All I need to do is update regression test values accordingly since they were apparently sensitive to the change in scaling.

bknueven (commented):
I think my meta-comment is this could be okay -- though it would be better to run `calculate_scaling_factors` after the unit model is initialized instead. Sometimes it is difficult to know good scaling factors without a solution in hand.

If things still "work" at the flowsheet level having a few custom options which just need to be utilized for unit model initialization isn't a huge deal, IMHO, even if it's less than ideal.

adam-a-a:
This has sat around for a long time and is very low priority.

PR #1139, opened 2023-09-19
IonExchange0D Unit Model and Costing Improvements

kurbansitterley:
## Fixes/Resolves:

- Improved/updated costing relationships from most recent release of EPA-WBS model.
- Add ``single_use`` option for regeneration and associated costing
- Remove ``bed_capacity_param`` variable and associated constraints to improve model stability/solvability.

## Summary/Motivation:

Changes to model made as results of continuing PFAS analysis. 
- ``bed_capacity_param`` was found to explode to massive values (>1e+20) under many model conditions, resulting in unbounded/infeasible solves (or optimal solves that were non-sensical). This PR removes that variable and all associated constraints and expressions from the model. The only place this variable was found to be necessary was in the calculation of estimated breakthrough times for the trapezoid approach used for steady-state effluent concentration ``tb_traps``.  A suitable replacement was used (essentially the ``eq_clark_1`` constraint)
- The regressions used for calculating vessel/tank costs were updated by the EPA and are applied here. They provide better economy of scale.

- remove ``bed_capacity_param``, ``kinetic_param`` and associated constraints/expressions
- update costing regressions for tank/vessels based on recent EPA-WBS model
- add ``single_use`` CONFIG option that assumes entire bed is replaced rather than regenerated
     - when this is used, there is no regenerant is costed
- modify docs to reflect all these changes

### Legal Acknowledgement


hunterbarber (commented):
No quarrels with the costing changes. I do have a question regarding the `single_use` condition which a lot of the changes are surrounding the conditional with this port. Is the regen port containing solutes, resin, or both? See other comments regarding mass balance perspective.

docs/technical_reference/unit_models/ion_exchange_0D.rst
```diff
@@ -80,7 +83,7 @@ The model provides three ports (Pyomo notation in parenthesis):
 
 * Inlet port (inlet)
 * Outlet port (outlet)
-* Regeneration port (regen)
+* Regeneration port (regen, only if not using ``single_use`` resin configuration)
```
hunterbarber: Its slightly confusing from a mass balance perspective. If set to `single_use`, `inlet == outlet` so how is mass removed from a modeling perspective? There is no spent resin stream that would then be regenerated with x efficiency and then added as the regen stream, correct?

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -1051,10 +1026,16 @@ def eq_bed_flow(b):
...
+            if self.config.regenerant != RegenerantChem.single_use:
+                return (
+                    b.col_height
+                    == b.bed_depth
+                    + b.distributor_h
+                    + b.underdrain_h
+                    + b.bed_expansion_h
+                )
+            else:
+                return b.col_height == b.bed_depth + b.distributor_h + b.underdrain_h
```
hunterbarber: Does this make more sense than fixing bed_expansion_h to 0?

hunterbarber (commented):
No quarrels with the costing changes. I do have a question regarding the `single_use` condition which a lot of the changes are surrounding the conditional with this port. Is the regen port containing solutes, resin, or both? See other comments regarding mass balance perspective.

docs/technical_reference/unit_models/ion_exchange_0D.rst
kurbansitterley: Hm maybe a poor choice of words on my part. The mass balance is handled the same as before where effluent concentration is estimated via trapezoid rule, which is then used to set ``process_flow.mass_transfer_term`` (see e.g., ``eq_mass_transfer_target_fr`` - I think this is pretty much identical to how GAC does it) . Without single-use, that flows to the ``regeneration_stream`` but with single-use it goes nowhere. There was never any regeneration efficiency in the model, so anything that flowed to that port was assumed to be removed via regeneration (which is not handled by the model; a user could attach a separate unit process as a regen handling process if they wanted.

But maybe there is virtue in having the removed mass flow somewhere rather than a black hole.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: Well ``bed_expansion_h`` is an ``Expression`` that is a function of other Params so would require setting those also to zero. Would you prefer that rather than the conditionals?

docs/technical_reference/unit_models/ion_exchange_0D.rst
hunterbarber: I remember this 'into the void' method when originally developing the model. 

I think I'm still confused that if the contaminants are going 'into the void' and the resin is also not modeled with a pseudo-species method, what is the intention of the `regeneration_stream`? i.e., if someone wanted to customize a regen or waste handling unit, what is this stream/port accomplishing?

watertap/unit_models/ion_exchange_0D.py
hunterbarber: I'm just thinking of it from a user perspective. If the model was preconstructed and then toggled between `single_use` and not, is it more or less intuitive if:

1. Previous variables that did exist on the model are no longer there, calling them will result in an error
2. Previous variables that were nonzero are now set to zero (or 1 depending on the equation structure), users have to interpret this as inactive when `single_use`

I don't know the best answer, but this effects a few variables throughout the condition

watertap/costing/units/ion_exchange.py
```diff
@@ -282,7 +253,7 @@ def cost_ion_exchange(blk):
         units=blk.costing_package.base_currency / blk.costing_package.base_period,
         doc="Operating cost for hazardous waste disposal",
     )
-    blk.regen_soln_flow = pyo.Var(
+    blk.vol_flow_regen_soln = pyo.Var(
```
adam-a-a: Should this be something like `flow_mass_regen_soln`? The units are in terms of mass and `flow_mass` would coincide with how we refer to mass flow in IDAES and WaterTAP.

watertap/costing/units/ion_exchange.py
```diff
@@ -317,25 +288,64 @@ def cost_ion_exchange(blk):
...
-            ion_exchange_params.backwash_tank_intercept
-            + ion_exchange_params.backwash_tank_A_coeff * bw_tank_vol**3
-            + ion_exchange_params.backwash_tank_B_coeff * bw_tank_vol**2
-            + ion_exchange_params.backwash_tank_C_coeff * bw_tank_vol,
-            to_units=blk.costing_package.base_currency,
+    if blk.unit_model.config.regenerant == "single_use":
+        blk.capital_cost_backwash_tank.fix(0)
+        blk.capital_cost_regen_tank.fix(0)
+        blk.vol_flow_regen_soln.fix(0)
+        blk.vol_flow_resin = pyo.Var(
```
adam-a-a:
```suggestion
        blk.flow_vol_resin = pyo.Var(
```
Minor point, but `flow_vol` would be more consistent with naming conventions

watertap/costing/units/ion_exchange.py
```diff
@@ -317,25 +288,64 @@ def cost_ion_exchange(blk):
...
+            (
+                blk.unit_model.bw_flow * blk.unit_model.t_bw
+                + blk.unit_model.rinse_flow * blk.unit_model.t_rinse
+            ),
+            to_units=pyo.units.gal,
+        )
+        regen_tank_vol = pyo.units.convert(
+            blk.unit_model.regen_tank_vol,
+            to_units=pyo.units.gal,
+        )
```
adam-a-a: Should these be accessible Expressions for the user to get reported values on?

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -101,6 +101,7 @@ class RegenerantChem(StrEnum):
     H2SO4 = "H2SO4"
     NaCl = "NaCl"
     MeOH = "MeOH"
+    single_use = "single_use"
```
adam-a-a, line 104: Why do you need "single_use"? Can "none" just be used, or get rid of "none" and only have "single_use"? I figure one can be eliminated, but maybe I haven't seen enough of the code yet to tell.

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -808,97 +783,127 @@ def build(self):
...
-                b.bed_expansion_frac_A
-                + b.bed_expansion_frac_B * b.bw_rate
-                + b.bed_expansion_frac_C * b.bw_rate**2
-            )  # for 20C
-
-        @self.Expression(doc="Bed expansion from backwashing")
-        def bed_expansion_h(b):
-            return b.bed_expansion_frac * b.bed_depth
+                b.p_drop_A + b.p_drop_B * vel_bed + b.p_drop_C * vel_bed**2
+            ) * b.bed_depth  # for 20C;
```
adam-a-a, line 779: Does the documentation reflect the 20C dependence?

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -808,97 +783,127 @@ def build(self):
...
+                    + b.bed_expansion_frac_B * b.bw_rate
+                    + b.bed_expansion_frac_C * b.bw_rate**2
+                )  # for 20C
 
-        @self.Expression(doc="Volume per column")
-        def col_vol_per(b):
-            return b.col_height * (b.bed_vol / b.bed_depth)
+            @self.Expression(doc="Bed expansion from backwashing")
+            def bed_expansion_h(b):
+                return b.bed_expansion_frac * b.bed_depth
```
adam-a-a: I wonder if you can insert a conditional checking for single use or regeneration, and if not single use, return what you have now; if single use, return 0. Maybe this could address @hunterbarber 's comment below about fixing `bed_expansion_h` to 0 in line 1038.

watertap/unit_models/ion_exchange_0D.py
adam-a-a: I had a suggestion above regarding the Expression for `bed_expansion_h`. To reiterate, maybe we could just have `bed_expansion_h` return 0 when single use is selected. Then we could eliminate the conditional here and just have the single equation with all terms, including `bed_expansion_h`.

adam-a-a: I would prefer this if it didn't cause any numerical issues since it would provide some symmetry between config options (i.e., bed_expansion_h would exist for both single use and regen cases).

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -1192,20 +1171,8 @@ def eq_clark_1(b, j):
...
-            def eq_clark_2(b, j):
-                c0 = prop_in.conc_mass_phase_comp["Liq", j]
-                cb = b.c_breakthru[j]
-                denom = (
-                    1
-                    + b.bed_capacity_param
-                    * exp((-b.kinetic_param * b.bed_depth * b.bv) / b.vel_bed)
-                ) ** (1 / (b.freundlich_n - 1))
-                return c0 == denom * cb
+                # return c0 == denom * cb
```
adam-a-a: commented code

watertap/unit_models/tests/test_ion_exchange_0D.py
```diff
@@ -254,6 +252,7 @@ def test_solution(self, IX_lang):
             "resin_max_capacity": 3,
             "resin_eq_capacity": 1.5547810762853227,
             "resin_unused_capacity": 1.4452189237146773,
+            # "regen_dose": 300,
```
adam-a-a: commented code

watertap/unit_models/tests/test_ion_exchange_0D.py
```diff
@@ -307,17 +305,17 @@ def test_costing(self, IX_lang):
         results = solver.solve(m, tee=True)
         assert_optimal_termination(results)
 
-        assert pytest.approx(8894349.86900, rel=1e-3) == value(
+        assert pytest.approx(3294284.78759, rel=1e-3) == value(
```
adam-a-a: This seems like a large decrease in capex. Is this because we're omitting regeneration and associated capex?

watertap/unit_models/tests/test_ion_exchange_0D.py
```diff
@@ -533,7 +529,7 @@ def test_solution(self, IX_fr):
             "resin_diam": 0.0006749999999999999,
             "resin_bulk_dens": 0.72,
             "resin_surf_per_vol": 4444.444444444445,
-            "regen_dose": 300,
+            # "regen_dose": 300,
```
adam-a-a: commented code

watertap/unit_models/tests/test_ion_exchange_0D.py
```diff
@@ -582,7 +578,7 @@ def test_solution(self, IX_fr):
             "mass_transfer_coeff": 0.159346300525143,
             "bv": 18000,
             "bv_50": 20000,
-            "bed_capacity_param": 311.9325370632754,
+            # "bed_capacity_param": 311.9325370632754,
```
adam-a-a: commented code

watertap/unit_models/tests/test_ion_exchange_0D.py
```diff
@@ -611,21 +607,315 @@ def test_costing(self, IX_fr):
         results = solver.solve(m, tee=True)
         assert_optimal_termination(results)
 
-        assert pytest.approx(9701947.4187, rel=1e-3) == value(
+        assert pytest.approx(4220876.29698, rel=1e-3) == value(
```
adam-a-a: same comment as above

watertap/unit_models/tests/test_ion_exchange_0D.py
adam-a-a: Actually, I remember you changing a cost relationship from polynomial to power fit, so maybe the change is due to that. Not sure...

adam-a-a (commented):
Looks good! Just a couple of comments here and there

adam-a-a:
@kurbansitterley I know you probably realized already, but there is a merge conflict likely due to the reformulation/movement of costing files in #1122

watertap/costing/units/ion_exchange.py
kurbansitterley: Ah yes you are correct. Will address

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: Ok I am going with having the ``bed_expansion_h`` ``Expression`` return 0 for ``single_use`` and eliminate this conditional in the constraint.

docs/technical_reference/unit_models/ion_exchange_0D.rst
kurbansitterley: The ``regeneration_stream`` could be used to connect a brine/regen management unit to it, as you say... but it is a good point; I imagine the bigger challenge in managing IX regen streams is the high salinity/concentration of regenerant they typically contain rather than the flow of removed species, so one could argue the regeneration stream shoudl contain the regenerant rather than the target_ion (if it is to contain anything). But that would require some more development to accomplish and likely carries its own headaches... 

I'll have to think about this one.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: Sure does. But I would rely on the user to bring their own numbers (and thus temperature dependence) if they want to change these values. Manufacturers would typically have temperature-dependent curves in their spec sheet.

watertap/unit_models/ion_exchange_0D.py
kurbansitterley: Great catch- going with ``single_use``

watertap/costing/units/ion_exchange.py
kurbansitterley: Great idea!

watertap/unit_models/tests/test_ion_exchange_0D.py
kurbansitterley: You are correct- change is due to the power fit rather than the polynomial.

docs/technical_reference/unit_models/ion_exchange_0D.rst
kurbansitterley: Following up- I added the ``regeneration_stream`` into the model for all CONFIGs (i.e. undoing what I had initially implemented). I also added back backwashing/rinsing costs/params/expressions for the ``single_use`` option. Now only regenerant costs are excluded with the ``single_use`` configuration.

hunterbarber:
@kurbansitterley Is this waiting on anything other than review? I will review again if these failing checks are workflow incidental.

kurbansitterley:
> @kurbansitterley Is this waiting on anything other than review? I will review again if these failing checks are workflow incidental.

Yes I wanted to wait for #1175 to be merged since it also changes the IX costing.

bknueven:
@kurbansitterley let me know if you want assistance resolving the conflicts post-#1175.

watertap/examples/flowsheets/ion_exchange/tests/test_ion_exchange_demo.py
```diff
@@ -138,19 +139,97 @@ def test_specific_operating_conditions(self, ix_0D):
...
 
-        assert value(m.fs.ion_exchange.number_columns) == pytest.approx(4, rel=1e-3)
-        assert value(m.fs.ion_exchange.bed_depth) == pytest.approx(1.7, rel=1e-3)
-        assert value(m.fs.ion_exchange.t_breakthru) == pytest.approx(
-            56759.75759, rel=1e-3
-        )
-        assert value(m.fs.ion_exchange.partition_ratio) == pytest.approx(
-            235.99899, rel=1e-3
-        )
+        results_dict = {
```
hunterbarber, line 142: This is pretty unsightly and could get cleaned up. That is why we use approx after all, I think I include 4 significant figures and pytest to 1e-3. This applies to a lot of code across the test files.

hunterbarber (approved):
Reviewing changes since last review, it seemed good to go.  The single comment could be addressed but tedious and technically not incorrect.

adam-a-a (approved):

watertap/examples/flowsheets/ion_exchange/tests/test_ion_exchange_demo.py
kurbansitterley: Don't disagree, was trying to move fast. Ask and ye shall receive.

PR #1141, opened 2023-09-21
OLIApi Revision

veccp:
- Move some methods out of OLIApi class into new modules
- Add CredentialManager for OLIApi login
- Add module for WaterAnalysis call
- Add helper functions for state blocks and OLI Cloud
- Add tests for new modules

### Legal Acknowledgement


watertap/core/util/tests/test_credential_manager.py
```diff
@@ -0,0 +1,83 @@
...
+    assert credentials == credential_manager_encrypt.credentials
+
+    credential_manager_encrypt_copy.config_file = (
+        credential_manager_encrypt._config_file
+    )
+
+    try:
+        credential_manager_encrypt_copy._get_credentials()
+    except RuntimeError:
+        pass
```
adam-a-a:
```suggestion
    with pytest.raises(RuntimeError, match="Unable to upload DBS file."):
        credential_manager_encrypt_copy._get_credentials()
```
something like this will test that the exception is raised, whereas the current way it is written will pass either way, and hypothetically, there could be a case where the exception was unexpectedly not being raised.

watertap/core/util/oli_api.py
```diff
@@ -75,126 +77,34 @@ class OLIApi:
...
-    def __init__(self, username=None, password=None, root_url=None, auth_url=None):
+    def __init__(
+        self,
+        username=None,
+        password=None,
+        root_url=None,
+        auth_url=None,
+        config_file=None,
+        encryption_key=None,
+        credential_manager_class=CredentialManager,
```
lbianchi-lbl: In the current implementation, these arguments must be provided to the `OLIApi` constructor along with `credential_manager_class`, which is then used to instantiate a `CredentialManager` object. Would it be possible to instead pass a "pre-built" `CredentialManager` instance instead? This would make the `OLIApi` class even less coupled to the credential handling.

watertap/core/util/oli_api.py
```diff
@@ -307,26 +217,15 @@ def call(
         # formulate url
         endpoint = ""
         method = "POST"
-        if (
-            function_name == "chemistry-info"
-            or function_name == "corrosion-contact-surface"
+
+        if (function_name == "chemistry-info") or (
+            function_name == "corrosion-contact-surface"
```
lbianchi-lbl:
```suggestion
        if function_name in {
            "chemistry-info",
            "corrosion-contact-surface",
        }:
```

watertap/core/util/tests/test_credential_manager.py
```diff
@@ -0,0 +1,83 @@
...
+    "password": password,
+    "root_url": root_url,
+    "auth_url": auth_url,
+}
+ 
+config_file = None
+encryption_key = None
+    
+@pytest.mark.unit
+def test_encryption(credentials, config_file, encryption_key):
```
lbianchi-lbl: I'm not 100% sure how pytest manages this case when arguments of test functions are not defined as fixtures but match module-scope variables. In any case, it might be preferable to make it more explicit by either turning them into fixtures (which is also useful if clean-up logic is needed, e.g. deleting `config_file`, see https://docs.pytest.org/en/stable/how-to/fixtures.html#safe-fixture-structure), or passing them using `pytest.mark.parametrize` (https://docs.pytest.org/en/stable/how-to/parametrize.html).

lbianchi-lbl (commented):
This looks good, although I think it requires to double-check the syntax and then run Black again (Black doesn't work if there are syntax errors in the file).

One general question: would it be possible to change the package structure in the following way?

```
watertap/tools/oli_api
    __init__.py
    credentials.py  # with the contents of credential_manager.py
    client.py  # with the contents of the current oli_api.py
    tests/
        test_credentials.py  # with the contents of tests/test_credential_manager.py
        test_client.py  # with the contents of tests/test_oli_api.py
```

We could then use `watertap/tools/oli_api/__init__.py` to expose the user-facing objects, i.e.:

```py
from . import credentials
from . import client
from .client import OLIApi
from .credentials import CredentialManager
```

This way the code can be used by:

```py
from watertap.tools.oli_api import OLIApi, CredentialManager

client = OLIApi(...)
```

(The most important part of this would be moving the code to its own dedicated subpackage under `watertap.tools`; the alternate names are just a suggestion)

watertap/core/util/tests/test_credential_manager.py
veccp: Thanks for the references, I'm working on a few changes to improve the tests

watertap/core/util/oli_api.py
veccp: I think it'd be fine to restructure. In fact I created an OLI folder in that location for another PR

bknueven (approved):
I think we should merge this an revise as needed.

ksbeattie:
The RTD error is:

```/home/docs/checkouts/readthedocs.org/user_builds/watertap/checkouts/1141/watertap/tools/oli_api/client.py:docstring of watertap.tools.oli_api.client.OLIApi.get_flash_history:7: WARNING: Field list ends without a blank line; unexpected unindent.```

**Review by lbianchi-lbl — CHANGES_REQUESTED** [review]
Make sure that there's a `__init__.py` file in _every_ subdirectory containing Python files.  `__init__.py` files may be empty, in which case they don't need a copyright header.

**Review by lbianchi-lbl — CHANGES_REQUESTED** [review]
At the moment, the `OLIApi`-specific dependencies are declared under the `dev` target in `setup.py`. This works just fine, however, it means that if we want to test the OLI API code, we need to install dev dependencies, even if they aren't strictly needed.

Therefore, I propose the following:

1) Create a new `extras_require` target named e.g. `oli_api` containing the OLI API dependencies (`cryptography` and `requests`) and remove `cryptography` from the `dev` target:

  ```py
    extras_require={
        "testing": [
            "pytest",
            "json-schema-for-humans",
            "mongomock",
            "pandas",
            "nbmake",
            "nbconvert",
        ],
        "notebooks": [
            "jupyter",
            "ipykernel",
        ],
        "oli_api": [
             "requests",
             "cryptography",  # for encrypting OLI credentials
        ],
        "dev": [
            "nbsphinx",  # jupyter notebook support for sphinx
            "jinja2<3.1.0",  # see watertap-org/watertap#449
            "Sphinx==7.1.*",  # docs
            "sphinx_rtd_theme",  # docs
            "urllib3 < 2",  # see watertap-org/watertap#1021,
            # other requirements
            "linkify-it-py",
            "json-schema-for-humans",  # pretty JSON schema in HTML
            "black",  # code formatting
            # other requirements
            "pytest",  # test framework
            "pytest-cov",  # code coverage
            "mongomock",  # mongodb mocking for testing
            "nbmake",
        ],
    },
```
2) Add the `oli_api` target to the last line in `requirements-dev.txt` (so that the `oli_api` dependencies are installed by default when creating a development environment):

  ```txt
  -e .[dev,notebooks,oli_api]
  ```

bknueven:
@lbianchi-lbl I believe your dependency comment was address in [ae61ec8](https://github.com/watertap-org/watertap/pull/1141/commits/ae61ec845c780c59977a9ffa6039a0b53fe70b4e).

watertap/tools/oli_api/tests/test_credentials.py
```diff
@@ -0,0 +1,70 @@
...
+        == credential_manager.credentials
+    )
+    remove(credential_manager.config_file)
+
+
+# TODO: get access token from OLI to improve test coverage
+@pytest.mark.unit
+@pytest.mark.skipif(
+    not cryptography_available, reason="cryptography module not available"
+)
```
lbianchi-lbl: Since the availability of `cryptography` is tied to the credential manager, we can add the skip logic directly to the `credential_manager` fixture so that any tests that requires `credential_manager` will be skipped if not available without needing to add individual markers:

```suggestion
@pytest.fixture
def credential_manager():
    if not cryptography_available:
        pytest.skip(reason="cryptography module not available")
    credentials = {
        "username": "dummy_username@email.com",
        "password": "dummy_password",
        "root_url": "https://dummyrooturl.com",
        "auth_url": "https://dummyauthurl.com",
    }
    return CredentialManager(**credentials, test=True)


@pytest.mark.unit
def test_encryption(credential_manager):
    key = credential_manager.encryption_key
    assert (
        CredentialManager(encryption_key=key, test=True).credentials
        == credential_manager.credentials
    )
    remove(credential_manager.config_file)


# TODO: get access token from OLI to improve test coverage
@pytest.mark.unit
```

lbianchi-lbl (approved):

PR #1145, opened 2023-09-22
Update nf scaling.

avdudchenko:
## Fixes/Resolves:

This is meant to improve nf stability when adding new ions, with low concentrations, and simplifies some of the scaling calculations. 

## Summary/Motivation:
When adding Sr and Ba to NF flowsheet for external analysis, it was found that dual infesability was 3e-3 ,while for the current NF flowsheet the dual infeasibility was 1e-7. This change fixes the estimates for ion flux scailing, which improves dual infesaiblity, both in external flowsheet  (to 7e-6) and watertap flowsheet reducing dual infeasibility to 1e-8 and hopefully, improving the stability of the flowsheets. Further improvments will need to be made for low ion concentrations, where dual infeasibility seems to grow. 

Report from ipopt before updated scailing for psudo-box problem solve:

```
Number of Iterations....: 14
                                    (scaled)                 (unscaled)
Objective...............:   0.0000000000000000e+00    0.0000000000000000e+00
Dual infeasibility......:   2.9103830456733704e-11    1.4126505637889256e-07
Constraint violation....:   8.5265128291212022e-14    4.8316906031686813e-13
Complementarity.........:   2.5059035596800626e-09    2.5059035596800626e-09
Overall NLP error.......:   7.0141214751885410e-10    1.4126505637889256e-07


Number of objective function evaluations             = 18
Number of objective gradient evaluations             = 15
Number of equality constraint evaluations            = 18
Number of inequality constraint evaluations          = 18
Number of equality constraint Jacobian evaluations   = 15
Number of inequality constraint Jacobian evaluations = 15
Number of Lagrangian Hessian evaluations             = 14
Total CPU secs in IPOPT (w/o function evaluations)   =      0.029
Total CPU secs in NLP function evaluations           =      0.000

```

Output with updated scaling:

```
Number of Iterations....: 15
                                   (scaled)                 (unscaled)
Objective...............:   0.0000000000000000e+00    0.0000000000000000e+00
Dual infeasibility......:   2.9103830456733704e-11    2.1460611066004273e-08
Constraint violation....:   3.6948222259525210e-13    3.6948222259525210e-13
Complementarity.........:   2.5059035596800626e-09    2.5059035596800626e-09
Overall NLP error.......:   7.0141196129320849e-10    2.1460611066004273e-08


Number of objective function evaluations             = 21
Number of objective gradient evaluations             = 16
Number of equality constraint evaluations            = 21
Number of inequality constraint evaluations          = 21
Number of equality constraint Jacobian evaluations   = 16
Number of inequality constraint Jacobian evaluations = 16
Number of Lagrangian Hessian evaluations             = 15
Total CPU secs in IPOPT (w/o function evaluations)   =      0.030
Total CPU secs in NLP function evaluations           =      0.001
```

-flux scaling is calculated based on water flux, feed ion concentration, and rejection estimated based on ion valance, and ion size relative to membrane pore size - this can use improvement in the future.
-some small updated to other calculated scailing factors (they produce equivalent scaling factors as before, but simplify the code) 

### Legal Acknowledgement


bknueven (approved):

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -1920,8 +1919,19 @@ def calculate_scaling_factors(self):
...
                 if comp.is_solute():
                     # Todo: revisit later
+                    physical_rejection_factor = (
+                        self.permeate_side[t, x].radius_stokes_comp[j].value ** 2
+                        / self.radius_pore.value**2
+                    )
+                    # will be 9, 99,999 etc for 1,2,3 valance
+                    rejection_factor = (
+                        10 ** abs(self.permeate_side[t, x].charge_comp[j].value) - 1
+                    ) / 10 ** abs(self.permeate_side[t, x].charge_comp[j].value)
```
adam-a-a: I am trying to bring these into my PR and see what the result will be. However, is rejection_factor making much difference? The comment says will be 9, 99, 999 for 1,2,3 valence, but as written, it would be 0.9, 0.99, 0.999, leading me to believe that `rejection_factor might not be critical. Sanity check?

adam-a-a: Ah- it should actually make a difference since you are doing `(1-rejection_factor)` below.

avdudchenko: Yea. That was not the most clear note. should of added the %. We also need to think about how to deal with "physical hinderence".....wait. dont we have an expression for that in NF model already ....

adam-a-a: `lambda_comp ` is the ratio between stokes radius of solute and membrane pore radius, so we technically could've used that in place of your manual writing of it although it is an expression with a smooth_min function in it, so I didn't press for it. We do have convective and diffusive hindrance factors in the NF model if that's what you were hinting at.

avdudchenko: Yeah, I actually tested out using diffus_pore_comp, but it made a minor change to unscaled tolerances. So I did not update it. 

The issue still remains for me, that if I add low concentration of ion like Sr or Ba, my dual infeasibility is at about 7e-6, ideally, we want to push that to 1e-8, like in the current nf flowsheet. but there is still something missing in the puzzle.  I wounder if we are focusing on the wrong scaling. One other thought came to mind, that we don't properly scale concentrations at membrane water interface, which would be higher than bulk...maybe that is worth looking at next for me. 

I will keep this PR up so we can use it as a test bed, there must be a better way to do this scailing.

adam-a-a:
Trying to merge PR #1136 first and @avdudchenko wants to keep this up to try a few more mods after the fact.

watertap/unit_models/nanofiltration_DSPMDE_0D.py
adam-a-a: Sounds good. I will put this PR into draft mode to avoid confusion by other potential reviewers.

avdudchenko: Great. Go ahead and push PR#1136 first. I want to keep this up so I can test linux/windows impacts, and propose new scailing approaches if there is somethhing I figure out that's cleaner/better.

watertap/examples/flowsheets/nf_dspmde/nf.py
```diff
@@ -275,7 +275,6 @@ def unfix_opt_vars(m):
     m.fs.NF.nfUnit.velocity.unfix()
     m.fs.NF.nfUnit.velocity.setub(0.25)
     m.fs.product.max_hardness.fix(500)
-    m.fs.NF.nfUnit.recovery_vol_phase.fix(0.90)
```
adam-a-a, line 277: Does this still break if you fix to 90% after scale refinement (considering all the latest changes with all checks passing)?

watertap/examples/flowsheets/nf_dspmde/tests/test_nf.py
```diff
@@ -20,13 +20,13 @@
...
-        "lcow": [m.fs.costing.LCOW, 0.144583],
-        "pressure": [m.fs.NF.pump.outlet.pressure[0] * 1e-5, 5.372779],
-        "area": [m.fs.NF.nfUnit.area, 405.66782957],
+        "lcow": [m.fs.costing.LCOW, 0.141678],
+        "pressure": [m.fs.NF.pump.outlet.pressure[0] * 1e-5, 5.21856],
+        "area": [m.fs.NF.nfUnit.area, 444.423],
         "recovery": [
             m.fs.NF.nfUnit.recovery_vol_phase[0.0, "Liq"] * 100,
-            90,
+            94.64,
```
adam-a-a: My only concern about letting recovery be "optimized" is that it'll drive up to the upper bound and at least previously, the test solution will deviate by a little every now and then with new changes in future PRs. It may be the case that the latest changes to scaling have remedied the issue completely, but it might also be the case that we'll see future tests fail because the solution converged to another recovery value close to 95% upper bound but outside of the tolerance of the regression test.

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -1883,10 +1881,33 @@ def calculate_scaling_factors(self):
         for v in self.recovery_vol_phase.values():
             iscale.set_scaling_factor(v, 1)
         if iscale.get_scaling_factor(self.radius_pore) is None:
-            iscale.set_scaling_factor(self.radius_pore, 10 / value(self.radius_pore))
+            iscale.set_scaling_factor(self.radius_pore, 1 / value(self.radius_pore))
         if iscale.get_scaling_factor(self.membrane_thickness_effective) is None:
-            iscale.set_scaling_factor(self.membrane_thickness_effective, 1e7)
+            iscale.set_scaling_factor(
+                self.membrane_thickness_effective,
+                1 / self.membrane_thickness_effective.value,
```
adam-a-a: Generally, it's better to go with `value(object)` over `object.value` (after importing from pyomo.environ) since the current approach would work with variables but fail with expressions, whereas using `value(object)` should work across the board.

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -1922,16 +1944,22 @@ def calculate_scaling_factors(self):
...
-                    # will be 0.9, 0.99,0.999 etc for 1,2,3 valance
-                    rejection_factor = (
+                    # will be 0.1,0.01,0.001 etc for 2,3 valance
+                    valance_driven_rejection = (
                         10 ** abs(self.permeate_side[t, x].charge_comp[j].value) - 1
                     ) / 10 ** abs(self.permeate_side[t, x].charge_comp[j].value)
+                    conc_scale = iscale.get_scaling_factor(
+                        self.feed_side.properties_in[t].conc_mol_phase_comp["Liq", j]
+                    )
+                    # prevents over scaling of really small concetratons
```
adam-a-a:
```suggestion
                    # prevents over scaling of really small concentratons
```

watertap/unit_models/tests/test_nanofiltration_DSPMDE_0D.py
```diff
@@ -1770,6 +1770,7 @@ def test_pressure_recovery_step_2_ions():
     assert len(unscaled_var_list) == 0
 
     # Expect only flux_mol_phase_comp to be poorly scaled, as we have not
+
```
adam-a-a:
```suggestion
```

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -1923,52 +1946,35 @@ def calculate_scaling_factors(self):
...
-                        )
-                        * iscale.get_scaling_factor(
-                            self.feed_side.properties_in[t].mw_comp["H2O"]
-                        )
-                        * (1 - rejection_factor)
-                        * physical_rejection_factor
-                        * iscale.get_scaling_factor(
-                            self.permeate_side[t, x].conc_mol_phase_comp["Liq", j]
-                        )
+                    # simply based on feed flow mol scailing
```
adam-a-a:
```suggestion
                    # simply based on feed flow mol scaling
```

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -1923,52 +1946,35 @@ def calculate_scaling_factors(self):
...
+                    #
+                    sf = iscale.get_scaling_factor(
+                        self.feed_side.properties_in[t].flow_mol_phase_comp["Liq", j]
                     )
-                    # sf = 1e5
                     iscale.set_scaling_factor(v, sf)
 
+        for (t, p, j), v in self.flux_mol_phase_comp_avg.items():
```
adam-a-a, line 1959: This is an Expression and I never add scaling for them. Did this make a difference?

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -1904,6 +1926,7 @@ def calculate_scaling_factors(self):
 
         # these variables do not typically require user input,
         # will not override if the user does provide the scaling factor
+        # make sure we scale solvent first!
```
adam-a-a, line 1929: Good catch!

watertap/unit_models/nanofiltration_DSPMDE_0D.py
adam-a-a: I see- you use it later on...thought it looks like you are just using the scaling factor of `flow_mol_phase_comp`

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -2015,6 +2021,139 @@ def calculate_scaling_factors(self):
             sf = iscale.get_scaling_factor(self.flux_mol_phase_comp[t, x, p, j])
             iscale.constraint_scaling_transform(con, sf)
 
+        for (t, x, p, j), con in self.eq_interfacial_partitioning_feed.items():
+            hfd = value(self.hindrance_factor_diffusive_comp[t, j])
+            bs = value(self.partition_factor_born_solvation_comp[t, j])
+            fdc = value(self.partition_factor_donnan_comp_feed[t, x, j])
+            hinderence_factor = hfd * bs * fdc
+            # we expect that there will be less ion at membrane with same charge as it self
+            # eg. negative ions ar repulsed by negative membarnes, but positve ions are attracted
```
adam-a-a:
```suggestion
            # we expect that there will be less ions at the membrane with same charge
            # e.g., negative ions are repulsed by negative membranes, but positive ions are attracted
```

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -2015,6 +2021,139 @@ def calculate_scaling_factors(self):
             sf = iscale.get_scaling_factor(self.flux_mol_phase_comp[t, x, p, j])
             iscale.constraint_scaling_transform(con, sf)
 
+        for (t, x, p, j), con in self.eq_interfacial_partitioning_feed.items():
+            hfd = value(self.hindrance_factor_diffusive_comp[t, j])
+            bs = value(self.partition_factor_born_solvation_comp[t, j])
+            fdc = value(self.partition_factor_donnan_comp_feed[t, x, j])
+            hinderence_factor = hfd * bs * fdc
```
adam-a-a:
```suggestion
            hindrance_factor = hfd * bs * fdc
```

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -2015,6 +2021,139 @@ def calculate_scaling_factors(self):
             sf = iscale.get_scaling_factor(self.flux_mol_phase_comp[t, x, p, j])
             iscale.constraint_scaling_transform(con, sf)
 
+        for (t, x, p, j), con in self.eq_interfacial_partitioning_feed.items():
+            hfd = value(self.hindrance_factor_diffusive_comp[t, j])
```
adam-a-a: If I understand the logic here, why wouldn't this be `partition_factor_steric_comp` instead?

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -2015,6 +2021,139 @@ def calculate_scaling_factors(self):
...
+                membrane_charge < 0
+                and value(self.permeate_side[t, x].charge_comp[j]) < 0
+            ):
+                charge_reflection_multiplier = 10 ** -abs(
+                    value(self.permeate_side[t, x].charge_comp[j])
+                )
+            else:
+                charge_reflection_multiplier = 10 ** abs(
+                    value(self.permeate_side[t, x].charge_comp[j])
+                )
```
adam-a-a: Very interesting. I have a couple of comments:
1) Did you find that `hindrance_factor` was insufficient and that also dividing by this `charge_reflection_multiplier` was necessary? Following the equation for interfacial partitioning, one would think the `charge_reflection_multiplier` wouldn't be necessary.
2) In the case that the membrane charge density were unfixed, the user would need to know to set to some guess value I suppose.
3) Assuming the usage of `charge_reflection_multiplier ` makes for better scaling and we keep the logic as is, I wonder if this can be condensed. It seems that you check whether ion and membrane have same charge (either positive or negative) or not. Given that, could we do the following instead?
```suggestion
            membrane_charge = value(self.membrane_charge_density[0])
            # multiply the two numbers. if one is different charge, outcome will be negative; if same charge, whether negative or positive, outcome will be positive.
            if (
                membrane_charge* value(self.permeate_side[t, x].charge_comp[j])  > 0
               
            ):
                charge_reflection_multiplier = 10 ** -abs(
                    value(self.permeate_side[t, x].charge_comp[j])
                )
           else:
                charge_reflection_multiplier = 10 ** abs(
                    value(self.permeate_side[t, x].charge_comp[j])
                )
```

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -2015,6 +2021,139 @@ def calculate_scaling_factors(self):
...
+            iscale.constraint_scaling_transform(
+                con, 1 / (hinderence_factor * charge_reflection_multiplier)
+            )
```
adam-a-a: Same suggestion for condensing as above

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -2023,13 +2162,65 @@ def calculate_scaling_factors(self):
...
+                        )
+                        * iscale.get_scaling_factor(
+                            self.pore_entrance[t, x].charge_comp[j]
+                        )
+                    )
+                    ** -1
+                    for j in self.config.property_package.solute_set
+                )
+                ** -1
+            ) ** -1
```
adam-a-a, line 2160: Hmm... this is all well and good, following that logic of equivalent resistance in parallel circuits, but do you have an extra power of -1 that should be removed here?

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -2015,6 +2021,139 @@ def calculate_scaling_factors(self):
...
+            ):
+                charge_reflection_multiplier = 10 ** -abs(
+                    value(self.permeate_side[t, x].charge_comp[j])
+                )
+            else:
+                charge_reflection_multiplier = 10 ** abs(
+                    value(self.permeate_side[t, x].charge_comp[j])
+                )
+            iscale.constraint_scaling_transform(
+                con, 1 / (hinderence_factor * charge_reflection_multiplier)
```
adam-a-a:
```suggestion
                con, 1 / (hindrance_factor * charge_reflection_multiplier)
```

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -2015,6 +2021,139 @@ def calculate_scaling_factors(self):
...
+            )
+
+        for (t, x, p, j), con in self.eq_interfacial_partitioning_permeate.items():
+            hfd = value(self.hindrance_factor_diffusive_comp[t, j])
+            bs = value(self.partition_factor_born_solvation_comp[t, j])
+            fdp = value(self.partition_factor_donnan_comp_permeate[t, x, j])
+            hinderence_factor = hfd * bs * fdp
```
adam-a-a:
```suggestion
            hindrance_factor = hfd * bs * fdp
```

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -2015,6 +2021,139 @@ def calculate_scaling_factors(self):
...
+            ):
+                charge_reflection_multiplier = 10 ** -abs(
+                    value(self.permeate_side[t, x].charge_comp[j])
+                )
+            else:
+                charge_reflection_multiplier = 10 ** abs(
+                    value(self.permeate_side[t, x].charge_comp[j])
+                )
+            iscale.constraint_scaling_transform(
+                con, 1 / (hinderence_factor * charge_reflection_multiplier)
```
adam-a-a:
```suggestion
                con, 1 / (hindrance_factor * charge_reflection_multiplier)
```

adam-a-a (commented):
LGTM - just a couple of questions/comments before we merge this

watertap/unit_models/nanofiltration_DSPMDE_0D.py
avdudchenko: Good question...I guess I need more coffee -fixed in next pr.

watertap/unit_models/nanofiltration_DSPMDE_0D.py
avdudchenko: Yeah, I observed this empirically. You basically need to know apriori how ions will partition because we have not computed the ion concentration ratio between pore entrance and interface yet and thats what the charge ratio tries to establish. So I think we have to make assumptions of membrane charge direction, rather then magnitude here. I also really doubt a user would be able to optimize from negative to postive with out things exploding, what happens when charge cross zero? (I don't want to find out?) 

```
b.pore_entrance[t, x].act_coeff_phase_comp[p, j]
                * b.pore_entrance[t, x].conc_mol_phase_comp[p, j]
                / (
                    b.feed_side.properties_interface[t, x].act_coeff_phase_comp[p, j]
                    * b.feed_side.properties_interface[t, x].conc_mol_phase_comp[p, j]
                )
```

avdudchenko: Also, yes I will condese this.

watertap/unit_models/nanofiltration_DSPMDE_0D.py
avdudchenko: will do

watertap/unit_models/nanofiltration_DSPMDE_0D.py
avdudchenko: Its becouse we need to sum this with scailing factor for membrane charge which I do later down. 
 `(e.g ((sum(x^-1)^-1)^-1+mem_charge^-1)^-1`

watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -2015,6 +2021,119 @@ def calculate_scaling_factors(self):
...
+                con, 1 / (hindrance_factor * charge_reflection_multiplier)
+            )
+
+        for (t, x, p, j), con in self.eq_interfacial_partitioning_permeate.items():
+            fsc = value(self.partition_factor_steric_comp[t, j])
+            bs = value(self.partition_factor_born_solvation_comp[t, j])
+            fdp = value(self.partition_factor_donnan_comp_permeate[t, x, j])
+            hindrance_factor = fsc * bs * fdp
+            # we expect that there will be less ions at membrane with same charge as it self
+            # eg. negative ions ar repulsed by negative membarnes, but positve ions are attracted
```
adam-a-a:
```suggestion
            # we expect that there will be less ions at membrane with same charge as itself
            # e.g., negative ions are repulsed by negative membranes, but positive ions are attracted
```

watertap/unit_models/nanofiltration_DSPMDE_0D.py
adam-a-a: Makes sense- we could then cancel these powers out, but won't functionally change anything. I'll leave the choice to you.

watertap/examples/flowsheets/nf_dspmde/nf.py
adam-a-a: I still would prefer fixing recovery rate here simply for the sake of avoiding future test failures due to recov rate riding up to the upper bound and subtle variations that could occur from that, but going to approve this anyway and leave the choice to you

adam-a-a (approved):
LGTM!

PR #1146, opened 2023-09-22
Eliminate suffix warnings by hot patching Pyomo's NL Writer

bknueven:
## Summary/Motivation:
@hunterbarber reports that the warnings from Pyomo about unexported suffixes for scaling factors are dominating his log file. We could simply turn down the logger for the NLWriter, but it also can report other useful things. Therefore, this PR would eliminate these warnings by hot-patching the class that generates them.

Note that the hot-patch is undone after a call to the watertap-ipopt wrapper.

- Hot patch `_SuffixData` to eliminate spurious warnings from Pyomo

### Legal Acknowledgement

### DIFF
--- a/docs/how_to_guides/how_to_use_a_property_model.rst
+++ b/docs/how_to_guides/how_to_use_a_property_model.rst
@@ -70,12 +70,6 @@ A portion of the displayed output is shown below.
 
 .. testoutput::
 
-   WARNING: model contains export suffix 'fs.state_block[0].scaling_factor' that
-   contains 4 component keys that are not exported as part of the NL file.
-   Skipping.
-   WARNING: model contains export suffix 'fs.state_block[0].scaling_factor' that
-   contains 4 component keys that are not exported as part of the NL file.
-   Skipping.
    Block fs.state_block[0]
 
      Variables:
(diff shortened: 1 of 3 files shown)


hunterbarber:
For more context, I'm running ~10,000 iterations of resolving a model with 5 or so `_SuffixData` warning logs that it is almost the entirety of the log. I feel as though its hindering the computational time for terminal updating as well as it hits the history limit in the terminal that I cannot review what the relevant logs actually are. Thanks @bknueven.

hunterbarber (approved):
LGTM 

 I know IDAES readthedocs has some instances of 
```
WARNING: model contains export suffix 'fs.state_block[0].scaling_factor' that
   contains 4 component keys that are not exported as part of the NL file.
   Skipping.
```
in their documentation, but I only see this revised instance in WaterTAP's docs. It is good to see that `how_to_use_a_property_model.rst` is the only affected output throughout the code, but the whole output log isn't never really being tested with pytest (only error text assertions) so it should be expected I guess. More so it's good that it likely means lower effort on changes after the true revisions are made overwriting the need for this hot patch.

I hope to run with the revised watertap/core/plugins/solvers.py just to double check, but again I think it looks good to go.

adam-a-a (approved):
LGTM - I'm tired of seeing those warnings!

ksbeattie:
Might it be worth asking the Pyomo developers for an alternative way to suppress these warnings?

bknueven:
> Might it be worth asking the Pyomo developers for an alternative way to suppress these warnings?

I talked with @andrewlee94 about this, and given the new scaling framework the Pyomo and IDAES teams are working on it isn't a priority.

PR #1150, opened 2023-09-26
Minor updates to elecytrolyzer model

avdudchenko:
## Fixes/Resolves:

Fixes non-reals when should be reals
Fixes double creation of capital cost vars in costing package. 

-
-

### Legal Acknowledgement


watertap/unit_models/electrolyzer.py
```diff
@@ -301,8 +300,8 @@ def custom_method_cathode(t, j):
         )
         self.anode_electrochem_potential = Var(
             initialize=1,
-            bounds=(0, None),
-            domain=NonNegativeReals,
+            bounds=(None, None),
```
adam-a-a:
```suggestion
```
Minor note, but I think you can delete bounds all together since it should already have (None, None) bounds.

adam-a-a (approved):
LGTM- just a minor comment regarding bounds. Also, curious if you noticed any performance improvement with these changes, or rather, you made the changes for consistency and out of anticipation for performance improvement (i.e., with IPOPT).

watertap/unit_models/electrolyzer.py
avdudchenko: I will remove the bounds. Yes, in a full flowsheet, it would not solve as some of these values would be set to be negative even as they are fixed. Odd.

hunterbarber (approved):
Fixing known flaws, LGTM

PR #1151, opened 2023-09-26
Fixes issue for saving model files with missing keys

avdudchenko:
## Fixes/Resolves:
This will catch when a result is returned with missing or inactive variables. This can occur when a solve fails or the initialization process fails without re-activating the variable first. Or where some variables will be deactivated by user based on solved condition.  

## Summary/Motivation:
When working with NN's I build the NNs on model tree, but deactivate them. they only end up being reactivated if solve is successful, so if it fails, we will return a result with missing keys. This update will attempt to catch missing keys and rebuild them. 

Might not be the best implementation. 

needs tests

-
-

### Legal Acknowledgement


k1nshuk:
The changes look good to me so far. I can approve it once its out of draft.

adam-a-a:
Based on the codecov check, it looks like you just need to add some testing.

avdudchenko:
Looks like the tests now pass and are MPI safe. The test do need min 2 parallel workers to properly test the catches. Fails MacOS test as the objective results are not quiet right (1e-3 precision by looks of it...) 

Should be good to go/merge. This also includes some additional fixes that were caught during testing/development.

watertap/tools/parameter_sweep/tests/test_parameter_sweep.py
```diff
@@ -917,6 +917,110 @@ def test_parameter_sweep_optimize(self, model, tmp_path):
...
+                        "upper bound": 1,
+                        "value": np.array(
+                            [0.0, 0.25, 0.5, 0.0, 0.25, 0.5, 0.0, 0.25, 0.5]
+                        ),
+                    },
+                },
+            }
+
+            read_dict = _read_output_h5(h5_results_file_name)
+            # print(read_dict)
```
k1nshuk:
```suggestion
```

watertap/tools/parameter_sweep/tests/test_parameter_sweep.py
```diff
@@ -917,6 +917,110 @@ def test_parameter_sweep_optimize(self, model, tmp_path):
             _assert_dictionary_correctness(truth_dict, read_dict)
             _assert_h5_csv_agreement(csv_results_file_name, read_dict)
 
+    @pytest.mark.component
+    def test_parameter_sweep_optimize_with_added_var(self, model, tmp_path):
+        # this will run a solve function that adds a variable but only in some
+        # of the solves.
+        """THIS TEST IS DESIGNED FOR 2 Parallel workers!!!!!!"""
```
k1nshuk, line 972: What would be needed to make this test work with an arbitrary number of workers? This could be dealt with in a subsequent PR.

k1nshuk (approved):
Looks good to me so far. I am not sure about how to deal with the MacOS test, however.

watertap/tools/parameter_sweep/tests/test_parameter_sweep.py
avdudchenko: Its a minum number - if you have 1 worker then we don't test for missing keys. I think we should really just change how we store results from each solve (e.g instead of storing all results in arrays inside a result dictionory, we just store the value for each parameter in a dict and store an array of dicts--not sure how MPI safe that is but that would solve this issue. - this would be same as async multiprocessing or ray implementation, where we split parmans into size of 1...

ksbeattie:
@avdudchenko still for Dec?

avdudchenko:
> @avdudchenko still for Dec?

Would be nice, just needs one more review (@bknueven  plz) and otherwise its good to go?

ksbeattie:
> > @avdudchenko still for Dec?
> 
> Would be nice, just needs one more review (@bknueven plz) and otherwise its good to go?

There's still a failing test :(

bknueven (approved):

Issue #1123, opened 2023-09-06
Bug with calculate_state


adam-a-a:
### Description

using calculate_state with volumetric flow and mass concentration sometimes results in erroneous results for flow_mol_phase_comp

### Expected Behavior

using calculate_state with volumetric flow and mass concentration should not result in erroneous results for flow_mol_phase_comp

### Steps to Reproduce

@hunterbarber @kurbansitterley please point me to your sections of code where you implement this, as well as "problematic" flow_vol and conc_mass values where erroneous results for flow_mol were observed.

adam-a-a:
I have updates on this.
**TLDR:** `calculate_state` can yield erroneous values for state variables (e.g., `flow_mass_phase_comp` or `flow_mol_phase_comp`) **when providing very small values for, e.g., mass or molar concentrations.** The problem with this is that the user can easily miss that the solution is incorrect because `assert_optimal_termination` would pass. This issue is not just for MCAS but likely for all usages of `calculate_state` in property models.

**More details:**
We knew that we should calculate scaling factors before calling `calculate_state`, but what wasn't apparent was that if the user didn't provide scaling factors for very small values, or the scaling factors weren't sufficient enough, `calculate_state `could converge to a supposedly "optimal" solution because the small variable values provided by the user would lead to constraints being satisfied within tolerance. Technically, we knew this to be a potential caveat in general (i.e., very small values can "trick" the solver into "thinking" it converged to a correct solution)--we just didn't realize that this was slipping by calculate_state in particular.

**Remedy?**
Of course, one would think that scaling your property variables appropriately before running calculate_state would resolve the issue, and that's likely the case in most instances. The problematic scenario that I see here concerns the user who thinks they scaled appropriately but didn't, and now they are getting back their calculated state variables, which are incorrect, without realizing it. I am curious to hear thoughts on this, but one safeguard could be to add an optional argument to `calculate_state` that when true, would automatically scale the property variables specified in `var_args` (could be dangerous). Either that, or we can add default scaling factors for `var_args `in the case that they are missing scaling factors (e.g., scaling_factor= 10 /value) and log a warning. For example (omitting details, e.g., variable indices):

```
m.fs.properties_in.calculate_state(var_args = {'flow_vol': 1, 'conc_mass': 1e-12},
                                                                         hold_state=True}

#within calculate_state, scan the specified vars and check if they were given scaling factors. If not, set defaults.
for var in var_args.keys():
    model_var = getattr(self, var)
[...]

PR #1154, opened 2023-09-27
Automated MCAS state block creation for OLI cloud

veccp:
## Fixes/Resolves:

PR demonstrates initial attempt at automating MCAS state block creation for use with OLI Cloud.

## Summary/Motivation:

Followup from #1136.

- Uses periodic table to extract molecular weight, charge, and OLI name from IDAES names.
- Enables user to have OLI-compatible MCAS state block built from simple input dictionary
- Provides initial reverse lookup database to convert OLI -> WaterTAP names

### Legal Acknowledgement


watertap/tools/oli/tests/test_watertap_to_oli_helper_functions.py
```diff
@@ -0,0 +1,46 @@
...
+    test_conditions.append("K_+")
+    # anion, z = -1
+    test_conditions.append("Cl_-")
+    # cation, z > 1
+    test_conditions.append("Mg_2+")
+    # anion Z < -1
+    test_conditions.append("HCO3_2-")
+
+    for test_name in test_conditions:
+        watertap_to_oli(test_name)
```
lbianchi-lbl: A pattern I use often to test this kind of parsing/converting pure functions uses `@pytest.mark.parametrize` to declare inputs and expected outputs:

```py
@pytest.mark.parametrize(
     "input_name,expected_output", [
           ("Cl_-", "CHLORINE_MINUS"), # CHLORINE_MINUS is just a placeholder for the real thing
           ("Mg_2+", "MAGNESIUM_TWO_PLUS"),
    ]
)
def test_conversion(input_name: str, expected_output: str):
    assert watertap_to_oli(input_name) == expected_output
```

watertap/tools/oli/mcas_flowsheet.py
```diff
@@ -0,0 +1,123 @@
...
+    :param components: dictionary containing solute concentrations in mg/L.
+    """
+
+    mcas_input = {
+        "solute_list": [comp.oli_name for comp in components],
+        "mw_data": {comp.oli_name: comp.molar_mass * 1e-3 for comp in components},
+        "charge": {
+            comp.oli_name: comp.charge for comp in components if comp.charge != 0
+        },
+    }
```
adam-a-a: I definitely like the idea of populating `mw_data` and `charge` as another option besides manual entry, using the helper functions with the periodic table csv to whatever extent we can.

watertap/tools/oli/mcas_flowsheet.py
```diff
@@ -0,0 +1,123 @@
...
+from idaes.core.util.scaling import calculate_scaling_factors
+from idaes.core.solvers import get_solver
+import watertap.property_models.multicomp_aq_sol_prop_pack as props
+
+from watertap.tools.oli.watertap_to_oli_helper_functions import watertap_to_oli
+
+solver = get_solver()
+
+
+def mcas_flowsheet(source_water: dict):
```
adam-a-a: I think this seems more like an example that belongs in the examples directory somewhere.

watertap/tools/oli/mcas_flowsheet.py
```diff
@@ -0,0 +1,123 @@
...
+    Initializes scaling factors for flow_mol_phase_comp parameter block.
+
+    :param m: ConcreteModel object.
+    :param phase_comp_blk: object containing all solutes in state block.
+    """
+
+    comps = set(m.fs.properties.solvent_set | m.fs.properties.solute_set)
+    for comp in comps:
+        m.fs.properties.set_default_scaling(
+            "flow_mol_phase_comp", 1, index=("Liq", comp)
```
adam-a-a: Assigning a scaling factor of 1 for each component would be risky. You can explore this by adding a test that takes in very small concentrations (e.g., on the order of 10**-12). Afterwards, running calculate_state, try asserting that the resulting calculations for `flow_mol_phase_comp` actually hold up by manually calculating what it should be. You may find that IPOPT converged to a solution but that the solution is erroneous. See issue #1123.

watertap/tools/oli/mcas_flowsheet.py
veccp: I should be able to make this functionality stand alone - maybe as generate_mcas_input(), or more generally, generate_property_model_input() and make it compatible with other property models

watertap/tools/oli/mcas_flowsheet.py
veccp: That's fair, I did take this framework from your test_oli_api file

watertap/tools/oli/mcas_flowsheet.py
veccp: Thanks, perhaps we can assign 1/conc as a scaling factor instead?

(if conc is ~4-5 orders of magnitude from 1)

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -667,13 +667,15 @@ def initialize(
 
             # Vars indexed by phase and component_list
             for j in self[k].params.component_list:
-                if self.params.config.material_flow_basis == MaterialFlowBasis.molar:
+                if self[k].params.config.material_flow_basis == MaterialFlowBasis.molar:
```
lbibl, line 668: If this PR is taking longer time to be reviewed, can we open a tiny one to merge the correction of the two missing k's to smooth  some ongoing work?

watertap/tools/oli/mcas_flowsheet.py
adam-a-a: Right - so, I think we should move this and modify a little bit. One thought is that this can evolve into a how-to or tutorial for our documentation. Also, I think we should brainstorm alternative options to implement the loading of charge/molecular weight data into MCAS.

In summary though- we should drop this file and its test from where it is now and rethink the form it should take in the repo.

watertap/tools/oli/mcas_flowsheet.py
```diff
@@ -0,0 +1,123 @@
...
+import watertap.property_models.multicomp_aq_sol_prop_pack as props
+
+from watertap.tools.oli.watertap_to_oli_helper_functions import watertap_to_oli
+
+solver = get_solver()
+
+
+def mcas_flowsheet(source_water: dict):
+    """
+    Create a simple MCAS Flowsheet from a feedwater dictionary.
```
adam-a-a:
```suggestion
    Create a simple MCAS StateBlock from a feedwater dictionary.
```
Even though we technically build a FlowsheetBlock (as we do for any of our models), I think we should refer to this as creating an MCAS stateblock instead of an MCAS flowsheet.

watertap/tools/oli/watertap_to_oli_helper_functions.py
```diff
@@ -0,0 +1,195 @@
...
+
+        if charge_sign == "+":
+            charge = charge_magnitude
+        elif charge_sign == "-":
+            charge = -charge_magnitude
+        else:
+            raise RuntimeError("Unrecognized charge magnitude " + charge_magnitude)
+
+    else:
+        raise RuntimeError("Unrecognized name format " + watertap_name)
```
adam-a-a: I think this can be broken off into its own function so that we have helper functions explicitly for grabbing mw data and determining charge--though the latter is more strongly hinged to naming convention and we'd have to clearly document that discrepancy. On a side note, I think the general guidance for users should be--you don't _have_ to follow any particular naming conventions for solutes if you don't want to, but we have these helper functions that can make life easier (hopefully), and if you want to use them, then you do have to follow a particular naming convention.

Overall, the more we can decrease dependence on precise naming, the better.

watertap/tools/oli/periodic_table.csv
```diff
@@ -0,0 +1,119 @@
+AtomicNumber,Element,Symbol,AtomicMass,NumberOfNeutrons,NumberOfProtons,NumberOfElectrons,Period,Group,Phase,Radioactive,Natural,Metal,Nonmetal,Metalloid,Type,AtomicRadius,Electronegativity,ionizationEnergy,Density,MeltingPoint,BoilingPoint,stableIsotopes,Discoverer,Year,SpecificHeat,NumberOfShells,NumberOfValence
```
adam-a-a, line 1: I wonder if we want a test that ensures that state of this csv is preserved. That is, if any developer ever comes up with a need to modify this is any way, we should have a way to control such modifications to prevent unintended consequences (i.e., any changes to this csv would have to be reflected in the corresponding test file). For example, we can have a test that checks that all the headers currently in the file match the expected list.

watertap/tools/oli/watertap_to_oli_helper_functions.py
```diff
@@ -0,0 +1,195 @@
...
+# "https://github.com/watertap-org/watertap/"
+#
+#################################################################################
+
+"""
+This file contains methods to convert WaterTAP naming conventions to OLI
+and generate molecular weight and charge dictionaries from molecular formulae. 
+
+It calculates molecular weights using the periodic_table.csv from:
+https://gist.github.com/GoodmanSciences/c2dd862cd38f21b0ad36b8f96b4bf1ee.
```
adam-a-a, line 19: Does this satisfy any request from the source that might exist for citing?

EDIT: I realize it's the periodic table, but you never know...

watertap/tools/oli/watertap_to_oli_helper_functions.py
adam-a-a: Should we add a more user-friendly message here? Also, RuntimeError works but wondering if there is a more specific exception that might be preferable.

adam-a-a (changes requested):
I like where this is going. There are some requested changes that I think we need to think through.

watertap/property_models/multicomp_aq_sol_prop_pack.py
adam-a-a: @veccp - I can peel this bit off and throw up the separate pr for it since I was the one who introduced this oversight

veccp: @lbibl @adam-a-a that sounds good to me

watertap/tools/oli/mcas_flowsheet.py
adam-a-a: We could consider something like that--although we may want to let the user provide his/her own factors.

watertap/tools/oli/watertap_to_oli_helper_functions.py
veccp: I'm unable to find any citation requirements

watertap/tools/oli/watertap_to_oli_helper_functions.py
veccp: This method currently requires WaterTAP names to compute charge and molar mass, but I think conversion to OLI names could be made optional

watertap/tools/oli/periodic_table.csv
veccp: I put up a simple test in the latest commit

avdudchenko: I slightly wounder, if its better to convert this to json/yaml. so its important as a proper python dictionary from getgo? Might be easier to manage. Just a minor suggestion.

watertap/tools/oli/watertap_to_oli_helper_functions.py
```diff
@@ -0,0 +1,205 @@
...
+    """
+
+    file_path = Path(__file__).parents[0]
+    periodic_table = read_csv(join(file_path, "periodic_table.csv"))
+
+    # isolate single- and double- letter elements and add to element_counts dict
+    elements = findall("[A-Z][a-z]?[0-9]*", watertap_formula)
+
+    # TODO: strengthen element criteria to reduce potential for input bugs
+    element_counts = {}
```
avdudchenko, line 126: So I used a package before to do this, e.g. get moleular mass for a molecule called molmass https://pypi.org/project/molmass/  it might be a good option to use instead of manually doing these calculations. a suggestion. 

It might be also be possible to replace our own data base with this library instead.

watertap/tools/oli/periodic_table.csv
avdudchenko: On other note. I wounder if we should instead rely on a library designed for handling this sort of thing like molmass https://pypi.org/project/molmass/  ... it might remove all the headaches, but does add a dependency?

avdudchenko (approved):
This looks good, but brings up a question if we should be really handling ion mw, charges etc and calculation our selfs, or should use external package like molmass. Alternatively, should we build a generic utility to help with this in water tap that is not just meant for OLI? 

Other LGMT?

watertap/tools/oli/watertap_to_oli_helper_functions.py
adam-a-a: Whoa- `molmass `looks very interesting.  We can definitely get charge and molecular weight from that as well, and it seems rather flexible.

veccp: I've made a note to look into molmass, it appears to be a more robust solution, but would people have to start renaming everything in molmass convention?

veccp:
I made a lot of changes to these files for my code demo, and I'm pushing the revisions to #1141. So this PR could potentially be closed

ksbeattie:
> I made a lot of changes to these files for my code demo, and I'm pushing the revisions to #1141. So this PR could potentially be closed

"potentially be closed" or should be closed?  As in, is it waiting for something else?

veccp:
should be closed

ksbeattie:
Close in preference for #1141

PR #1162, opened 2023-10-04
Eliminate bounds relaxation in WaterTAP Ipopt

bknueven:
## Replaces #1140

@avdudchenko reports in consistencies in the way domain=pyo.NonNegativeReals vs bounds=(0, None). I tracked this down to an issue in the way "bounds relaxation" is handled in the WaterTAP wrapper for Ipopt.

This PR would resolve this issue by turning off bounds relaxation altogether for the WaterTAP wrapper for Ipopt.

## Summary/Motivation:
By default, Ipopt relaxes the bounds internally, e.g., `(0, 1)` becomes `(-1e-10, 1 + 1e+10)`. Because of WaterTAP's use of variable scaling, the `ipopt-watertap` wrapper was doing this operation itself. Unfortunately, the wrapper contained a bug wherein explicit bounds of `(0, None)` would get relaxed to `(-1e-10, None)` but `domain=pyo.NonNegativeReals` would *not* get relaxed, and would be represented as `(0, None)` within Ipopt.

Because these should be consistent, this PR would opt to turn off bounds relaxation entirely within the `ipopt-watertap` wrapper. The main affect of this is that *Ipopt will always return a solution **strictly** inside the specified bounds*. Therefore, it is generally *not* a good idea to specify implied bounds on variables -- a few commits (19a7dec, 3650135, 5e7d621) remove or loosen bounds on variables for model stability.

- Set `bounds_relax_factor == 0` to be the default behavior for `ipopt-watertap`
- Remove now redundant code and tests from `ipopt-watertap`
- Adjust models and test as needed for the changed default settings.

### Legal Acknowledgement

### DIFF
--- a/watertap/examples/chemistry/tests/test_pH_dependent_solubility.py
+++ b/watertap/examples/chemistry/tests/test_pH_dependent_solubility.py
@@ -1095,6 +1095,7 @@ def test_case_2_seawater_added_carbonates():
     )
 
 
+@pytest.mark.requires_idaes_solver
 @pytest.mark.component
 def test_case_2_low_pH_no_precip():
     model = run_case2(
--- a/watertap/examples/chemistry/tests/test_pH_dependent_solubility.py
+++ b/watertap/examples/chemistry/tests/test_pH_dependent_solubility.py
@@ -946,13 +946,16 @@ def run_case2(
     init_options = {**solver.options}
     init_options["tol"] = init_tol
     init_options["constr_viol_tol"] = init_tol
+    init_options["ma27_pivtol"] = 5e-1
     model.fs.unit.initialize(optarg=init_options, outlvl=idaeslog.DEBUG)
 
     assert degrees_of_freedom(model) == 0
 
     iscale.calculate_scaling_factors(model.fs.unit)
 
+    solver.options["ma27_pivtol"] = 5e-1
     results = solver.solve(model, tee=True)
+    del solver.options["ma27_pivtol"]
 
     assert results.solver.termination_condition == TerminationCondition.optimal
     assert results.solver.status == SolverStatus.ok
@@ -1028,6 +1031,7 @@ def run_case2(
 
 ## ================================= Case 1 Tests ===============================
 @pytest.mark.component
+@pytest.mark.xfail
 def test_case_2_do_nothing():
     model = run_case2(
         xOH=1e-7 / 55.2,
@@ -1105,7 +1109,7 @@ def test_case_2_low_pH_no_precip():
         rxn_config=case2_log_rxn_config,
         has_energy_balance=True,
         scaling_ref=1e-5,
-        init_tol=1e-8,
+        init_tol=1e-12,
     )
 
 
@@ -2470,8 +2474,10 @@ def run_case4(
 
     assert degrees_of_freedom(model) == 0
 
-    solver.options["tol"] = 1.0e-12
+    solver.options["tol"] = 1.0e-16
+    solver.options["ma27_pivtol"] = 5e-1
     results = solver.solve(model, tee=True)
+    solver.options["ma27_pivtol"] = 5e-1
     del solver.options["tol"]
 
     assert results.solver.termination_condition == TerminationCondition.optimal
--- a/watertap/examples/chemistry/tests/test_pure_water_pH.py
+++ b/watertap/examples/chemistry/tests/test_pure_water_pH.py
@@ -629,6 +629,7 @@ class TestPureWater:
         assert results.solver.status == SolverStatus.ok
 
     @pytest.mark.component
+    @pytest.mark.requires_idaes_solver
     def test_solution(self, model_solve):
         model, _ = model_solve
 
--- a/watertap/examples/chemistry/tests/test_pure_water_pH.py
+++ b/watertap/examples/chemistry/tests/test_pure_water_pH.py
@@ -654,8 +654,8 @@ class TestPureWater:
         pOH = -value(
             log10(model.fs.unit.outlet.mole_frac_comp[0, "OH_-"] * total_molar_density)
         )
-        assert pytest.approx(6.9997414, rel=1e-5) == pH
-        assert pytest.approx(6.9997414, rel=1e-5) == pOH
+        assert pytest.approx(7.000, rel=1e-3) == pH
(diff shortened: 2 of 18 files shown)


watertap/examples/chemistry/tests/test_pH_dependent_solubility.py
```diff
@@ -946,13 +946,16 @@ def run_case2(
     init_options = {**solver.options}
     init_options["tol"] = init_tol
     init_options["constr_viol_tol"] = init_tol
+    init_options["ma27_pivtol"] = 5e-1
```
avdudchenko, line 949: Have we tested this as a global default? Or does that break many flowsheets?

watertap/unit_models/zero_order/tests/test_blending_reservoir_zo.py
```diff
@@ -105,7 +105,7 @@ def test_solution(self, model):
                 value(model.fs.unit.inlet.flow_mass_comp[t, j]), rel=1e-5
             ) == value(model.fs.unit.outlet.flow_mass_comp[t, j])
 
-        assert pytest.approx(8.4491446e-11, rel=1e-5) == value(
+        assert pytest.approx(1.8449145e-10, rel=1e-5) == value(
```
avdudchenko, line 108: For all intents and purposes, this is in our eyes 0, should we not typically test these as value has to be <=1e-8 ? which is linked to tolerances of our constraints?

watertap/examples/chemistry/tests/test_pure_water_pH.py
```diff
@@ -654,8 +655,8 @@ def test_solution(self, model_solve):
         pOH = -value(
             log10(model.fs.unit.outlet.mole_frac_comp[0, "OH_-"] * total_molar_density)
         )
-        assert pytest.approx(6.9997414, rel=1e-5) == pH
-        assert pytest.approx(6.9997414, rel=1e-5) == pOH
+        assert pytest.approx(7.0000146, rel=1e-5) == pH
```
avdudchenko: It looks like our unscaled accuracy here is 1-e3 ? should we not simply relax the rel tolerance?  (For pH 7.001 ~= 6.999) 

I also presume, this change is due to some variables in this model being solved at bound, which now changed slightly as we removed that relaxation, and caused a small propagation into pH.

avdudchenko (approved):
I think this a positive change to WT  as in my tests it results in more consistent solving behavior on complex flow sheets  - e.g. If it solves, it will generally solve consistently, and changing scaling, bounds etc. produces immediate effects, while with current main implementation changes to bound in scaling/initialization might help solve one state, but model would be marginally stable after wards. This is anecdotal observation rather then a quantitative one.  My hypothesis on this is replacing 0 bound would allow small negative values, that could drive instability in some flow-sheets and this was primary impact on using Reals vs. NonNegativeReals -this behavior was really an issue if model was marginally scaled, or not well conditioned. I seen this directly when initializing NF-DSPM-DE model with wrong initial guess, as I would get warning of value being set below the lower bound of 0 to tiny negative value (-1e-15, or 1e-25) this some time would not impact the model and it would solve fine, some times it would break. 

P.S, Looks like updates to MCAS and NF scaling made those flowsheet solve with this change.

watertap/examples/chemistry/tests/test_pH_dependent_solubility.py
bknueven: I have not tested it extensively. I think I will explore it as a next step.

Ipopt will increase this automatically if it thinks it's "needed", so an in-between would be to set `ma27_pivtolmax` higher than the current default of `1e-04`.

watertap/unit_models/zero_order/tests/test_blending_reservoir_zo.py
bknueven: Yes, probably. But there are lots of small value comparisons in the ZO unit model tests, so I choose to preserve this for consistency.

watertap/examples/chemistry/tests/test_pure_water_pH.py
bknueven: Yeah, pH is logarithmic, so I think I will relax this tolerance.

carson-tucker (approved):

TimBartholomew (approved):
LGTM

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -658,7 +658,7 @@ def build(self):
             self.flowsheet().time,
             self.diluate.length_domain,
             initialize=0.9,
-            bounds=(0, 1),
+            bounds=(0, 1 + 1e-10),
```
adam-a-a, line 661: What happened without this?

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -658,7 +658,7 @@ def build(self):
...
             units=pyunits.dimensionless,
             doc="The overall current efficiency for deionizaiton",
```
```suggestion
            doc="The overall current efficiency for deionization",
```
totally unrelated to your changes

watertap/unit_models/mvc/tests/test_mvc.py
```diff
@@ -148,7 +148,7 @@ def test_mvc():
     solver = get_solver()
     initialize(m, solver=solver)
 
-    results = solver.solve(m, tee=False)
+    results = solver.solve(m, tee=True)
```
adam-a-a, line 151: Should we set back to False?

lbibl:
Changes on ED files look ok to me.

watertap/unit_models/mvc/tests/test_mvc.py
bknueven: No, because pytest suppresses the output by default, and if the test fails we always want to look at the solver logs.

Which is part of the reason I left some other debugging outputs in the tests as well. I don't see any point in *not* having them immediately when needed.

watertap/unit_models/electrodialysis_1D.py
bknueven: At least one test fails, and the solves look uglier..
https://github.com/bknueven/watertap/actions/runs/6423080058/job/17440867424?pr=11

bknueven: If it were possible to eliminate these bounds that would probably be better.

PR #1164, opened 2023-10-04
Updates to EC ZO model

kurbansitterley:
## Fixes/Resolves:

Improved model after validation with WaterTectonics data

## Summary/Motivation:

The model was initially to be developed with WaterTectonics data but there was a delay in getting that data. This model adds improvements based on that validation effort for both the unit model and costing model

- conductivity is a variable rather than determined from TDS.
- remove unneeded entries in yaml related to costing
- add CONFIG option for using regression to calculate overpotential
- rename some variables
- add floc basin to costing model
- eliminate unnecessary constraints and variables
- add calculate_scaling_factors to EC model
- update test to reflect the above changes

### Legal Acknowledgement


srikanthallu (commented):
Can you please update the documentation as well.

kurbansitterley:
> Can you please update the documentation as well.

Done

watertap/unit_models/zero_order/electrocoagulation_zo.py
```diff
@@ -333,64 +359,73 @@ def eq_cell_voltage(b):
...
-        def eq_electrode_volume_per(b):
+        def eq_electrode_volume(b):
             return (
-                b.electrode_volume_per
-                == b.electrode_width * b.electrode_height * b.electrode_thick
+                b.electrode_volume
+                == (b.anode_area + b.cathode_area) * b.electrode_thick
             )
 
+        @self.Constraint(doc="Cathode and anode areas are equal")
```
hunterbarber, line 369: Feels odd to give the option to explicitly define cathode and anode areas, only to be able to fix one or the other. Also, I don't think that the areas actually have to be equal? Maybe better to just omit the constraint and add a degree of freedom to specify both areas.

watertap/unit_models/zero_order/electrocoagulation_zo.py
```diff
@@ -288,43 +262,95 @@ def build(self):
...
+        self._fixed_perf_vars.append(self.conductivity)
         self._fixed_perf_vars.append(self.electrode_gap)
         self._fixed_perf_vars.append(self.current_efficiency)
-        self._fixed_perf_vars.append(self.overpotential)
+        self._fixed_perf_vars.append(self.floc_retention_time)
+
+        if self.config.overpotential_calculation is OverpotentialCalculation.fixed:
+            self._fixed_perf_vars.append(self.overpotential)
+
+        if self.config.overpotential_calculation == OverpotentialCalculation.calculated:
```
hunterbarber, line 292: Pretty easy to get this config option covered by tests? I'm not sure how particular the ZO unit models are to be.

hunterbarber (approved):
Pretty straightforward with a couple minor comments.

watertap/unit_models/zero_order/electrocoagulation_zo.py
kurbansitterley: True that the areas don't have to be equal, but given this is a ZO model I think it's best to keep them equal. Also I think we want to avoid leaving DOFs for a ZO model but could add additional complexity for the future detailed model. 

Defining both the cathode and anode areas is more for naming convention and to avoid the ambiguity of the previous model that just had ``electrode_area_per`` which got confusing when formulating other constraints.

srikanthallu (approved):

watertap/unit_models/zero_order/electrocoagulation_zo.py
kurbansitterley: Good catch, just added a test.

PR #1171, opened 2023-10-16
BSM2 GUI Variable Name Update

### LINKED ISSUE #1153: BSM2 GUI Flowsheet
### Description

https://github.com/watertap-org/watertap/pull/1125 creates an initial GUI for the BSM2 flowsheet, but we eventually want to display the flowrate and compositions of all the streams in the "Outlet" tab, rather than just the two outlet streams exiting the flowsheet.

### Motivation

It will be useful for the project team to see the composition of each stream in the flowsheet rather than just the outlets.

### Possible Implementation

_No response_

### Additional Context

Priority

- [x] Dewatering liquid outlet
- [x] Translator block inlets and outlets
- [x] Anaerobic digestor gas outlet
- [x] Every other stream (with the possible exceptions of the storage tank and individual reactors in the ASP)
- [x] Increase resolution of flowsheet image

Adding a task here regarding parameter sweep results in the GUI
- [ ] add labels to distinguish between flow rate/concentration results across different streams of the flowsheet

Currently, one cannot tell which outputs are associated with specific streams in the flowsheet.

MarcusHolly:
## Fixes/Resolves:

Remaining checkbox in #1153 

### Legal Acknowledgement

### DIFF
--- a/watertap/examples/flowsheets/case_studies/full_water_resource_recovery_facility/BSM2_ui.py
+++ b/watertap/examples/flowsheets/case_studies/full_water_resource_recovery_facility/BSM2_ui.py
@@ -634,7 +634,7 @@ def export_variables(flowsheet=None, exports=None):
     )
     exports.add(
         obj=fs.Treated.properties[0].conc_mass_comp["S_I"],
-        name="S_I concentration",
+        name="Treated S_I concentration",
         ui_units=pyunits.g / pyunits.m**3,
         display_units="g/m3",
         rounding=5,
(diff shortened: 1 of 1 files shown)


luohezhiming (approved):
LGTM

agarciadiego (approved):

adam-a-a (approved):
LGTM - @MichaelPesce @dangunter we realized that when looking at sweep results, if we weren't particular about naming the output, there could be confusion between multiple outputs with the same name. For example, a concentration at the outlet of one unit should be distinguished from the same component concentration at the outlet of a different unit in the flowsheet. 

TLDR: @MarcusHolly manually added category names in front of reported variable names to be able to distinguish between results. However, it might be better if this could be handled in the UI (i.e., for sweep results, display category and name for each entry).

Issue #1131, opened 2023-09-13
OARO improvements


luohezhiming:
### Description

This issue is made up for future improvements of multi-stage OARO flowsheet #897.

### Steps to Reproduce

- [x] Add recycle streams from product water to replace make-up stream;
- [x] Resolve initialization routine with number of stages larger than 5
- [x] Make box problem solvable for the flowsheet
- [x] Add 1d unit model
- [x] Add OARO1D documentation (#1261)
- [x] Implement 1d model in flowsheet

### Environment

- OS:
- Python:
- WaterTAP:

PR #848, opened 2022-11-17
OARO 1D [WIP]

bknueven:
## Fixes/Resolves:

(replace this with the issue # fixed or resolved, if no issue exists then a brief statement of what this PR does)

## Changes proposed in this PR:
-
-

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #1172, opened 2023-10-16
OARO 1D Model

luohezhiming:
## Fixes/Resolves:

This PR is to add OARO 1D model for issue #1131 
This PR inherits @bknueven previous revisions of initialization routine #848 and make all tests passed.

## Summary/Motivation:
Create OARO 1D model

- add OARO 1D model
- add tests

### Legal Acknowledgement


watertap/unit_models/osmotically_assisted_reverse_osmosis_1D.py
```diff
@@ -0,0 +1,240 @@
...
+
+        units_meta = self.config.property_package.get_metadata().get_derived_units
+        mem_side = getattr(self, side)
+        setattr(
+            mem_side,
+            "deltaP_stage",
+            Var(
+                self.flowsheet().config.time,
+                initialize=-1e5,
+                bounds=(-1e6, 0),
```
adam-a-a, line 124: I remember the bounds of pressure drop causing some issues in the multi-stage 0D OARO flowsheet. If my recollection is correct, should we remove these bounds? Even if it isn't correct, can we remove these bounds and still ensure (or even improve) model stability?

watertap/unit_models/tests/test_osmotically_assisted_reverse_osmosis_1D.py
```diff
@@ -0,0 +1,1213 @@
...
+            has_pressure_change=True,
+            concentration_polarization_type=ConcentrationPolarizationType.calculated,
+            mass_transfer_coefficient=MassTransferCoefficient.calculated,
+            pressure_change_type=PressureChangeType.calculated,
+        )
+
+        # fully specify system
+        feed_flow_mass = 5 / 18
+        feed_mass_frac_NaCl = 0.075
+        feed_pressure = 65e5
```
adam-a-a, line 1143: This pressure could be one reason you'd have trouble reaching lower recoveries.

watertap/unit_models/tests/test_osmotically_assisted_reverse_osmosis_1D.py
```diff
@@ -0,0 +1,1213 @@
...
+        # test initialization
+        initialization_tester(m)
+
+        m.fs.unit.permeate_inlet.pressure[0].unfix()
+        m.fs.unit.area.unfix()
+        m.fs.unit.recovery_mass_phase_comp[0, "Liq", "H2O"].fix(water_recovery)
+        m.fs.unit.permeate_outlet.pressure[0].fix(1e5)
+
+        # # test variable scaling
+        # badly_scaled_var_lst = list(badly_scaled_var_generator(m))
```
adam-a-a, line 1195: uncomment or delete

watertap/unit_models/tests/test_osmotically_assisted_reverse_osmosis_1D.py
```diff
@@ -0,0 +1,1213 @@
...
+        # assert badly_scaled_var_lst == []
```
adam-a-a: uncomment or delete

adam-a-a (commented):
Looks good! Can you update documentation as well?

watertap/unit_models/osmotically_assisted_reverse_osmosis_1D.py
luohezhiming: For some of unknown reasons, once I relax the lower bound to None, the Linux tests all failed, Windows still pass so it is hard to debug.

zacharybinger: Instead of relaxing the bounds to None, did you try 0? I guess it doesn't matter since you changed the bounds back to a negative value.

luohezhiming: Yeah, my upper bound is set as 0

zacharybinger: Ahh yes, I had the bounds backwards in my head since the deltaP values are negative, my mistake

zacharybinger (approved):
LGTM

adam-a-a (approved):

adam-a-a:
@luohezhiming just realized my last comment still would need to be addressed, but we can do that in a subsequent PR. I'll create an issue to track adding/updating OARO documentation

adam-a-a:
@luohezhiming - since we merged the SKK PR in first, that is breaking tests here. Hopefully, it's as easy as increasing the number of config options expected and nothing more than that.

luohezhiming:
> @luohezhiming just realized my last comment still would need to be addressed, but we can do that in a subsequent PR. I'll create an issue to track adding/updating OARO documentation

Hi, Adam. I've already updated the issue for documentation in issue #1131

luohezhiming:
> @luohezhiming - since we merged the SKK PR in first, that is breaking tests here. Hopefully, it's as easy as increasing the number of config options expected and nothing more than that.

Yeah, it seems just config number issue.

adam-a-a:
> > @luohezhiming just realized my last comment still would need to be addressed, but we can do that in a subsequent PR. I'll create an issue to track adding/updating OARO documentation
> 
> Hi, Adam. I've already updated the issue for documentation in issue #1131

Thanks for the reminder! I forgot about that. Deleting the redundant issue that I created then.

PR #1174, opened 2023-10-17
Costing Unification: parameter consistency across costing packages

bknueven:
## Summary/Motivation:
This PR would make a few minor changes to both the WaterTAPCosting package and the ZeroOrderCosting package, working towards their merge.

- In WaterTAPCosting, redefine the parameter `factor_maintenance_labor_chemical` to represent (fraction of equipment cost / year). It currently is (fraction of investment cost / year), which has no exact analogue in the ZOCosting package. 
- Create Expression for the high-level parameter `factor_maintenance_labor_chemical` and `factor_total_investment` in the ZOCosting package
- Redefine how aggregates are calculated in both costing packages to utilize the same variables and expressions.
- Misc. updates to tests to keep up with the above.

### Legal Acknowledgement


bknueven:
I want to resolve the `TIC` / `factor_total_investment` issue before merging. I do so in #1175.

docs/technical_reference/costing/watertap_costing.rst
```diff
@@ -128,7 +128,7 @@ Technoeconomic factors used to calculate various system metrics, capital, and op
 =============================================  ====================  =======================================  ===============  ==============================================================================
 Plant capacity utilization factor                 :math:`f_{util}`    ``utilization_factor``                   90%             Percentage of year plant is operating
 Total investment factor                           :math:`f_{toti}`    ``factor_total_investment``              2.0             Total investment factor (investment cost / equipment cost)
-Maintenance-labor-chemical factor                 :math:`f_{mlc}`     ``factor_maintenance_labor_chemical``    0.03            Maintenance, labor, and chemical factor (fraction of investment cost / year)
+Maintenance-labor-chemical factor                 :math:`f_{mlc}`     ``factor_maintenance_labor_chemical``    0.03            Maintenance, labor, and chemical factor (fraction of equipment cost / year)
```
TimBartholomew, line 131: This change effectively reduces our base maintenance-labor-chemical costs by 2, maybe we should double our base value to 0.06?

This change would also make the costing method not exactly the same as our LSRRO paper and our other papers under review (MVC and our OLI-WaterTAP), although the change is minor and easily explainable. We could do something where we have an option where maintenance-labor-chemical costs are based on either investment costs or equipment costs, but maybe that option unnecessarily complicates things.

watertap/costing/watertap_costing_package.py
```diff
@@ -59,42 +59,38 @@ def build_process_costs(self):
...
-                == self.maintenance_labor_chemical_operating_cost
-                + self.aggregate_fixed_operating_cost
-                + self.aggregate_variable_operating_cost
-                + sum(self.aggregate_flow_costs.values())
-                * self.base_currency
-                / self.base_period
+        self.total_variable_operating_cost = pyo.Expression(
+            expr=(
+                self.aggregate_variable_operating_cost
+                + sum(self.aggregate_flow_costs[f] for f in self.used_flows)
```
TimBartholomew, line 81: This is not part of this PR, but I'm wondering if we want aggregate_flow_costs to actually already incorporate the utilization factor. To me it makes sense that we apply the utilization factor wherever we calculate costs because we are estimating the annual costs. Are the units of aggregate_flow_costs $/year?

I think its fine for the flows to not include the utilization factor because they are the steady state flows. But when we estimate costs, we are estimating annual costs, not instantaneous (steady state) costs.

watertap/unit_models/zero_order/tests/test_anaerobic_digestion_reactive_zo.py
```diff
@@ -250,7 +250,6 @@ def test_costing():
 
     assert m.fs.unit.electricity[0] in m.fs.costing._registered_flows["electricity"]
     assert isinstance(m.fs.costing.total_capital_cost, Var)
-    assert isinstance(m.fs.costing.total_fixed_operating_cost, Var)
```
TimBartholomew, line 253: Also not part of this PR, but I think we really should implement test harnesses and drop tests like this. If the variable doesn't exist, there will be an error message that it doesn't exist either on build if the model needs it or in a regression test.

TimBartholomew (approved):
LGTM, I have a few comments, but I think its mergeable. I'm surprised we don't have to update the LCOW regression tests for all of our examples because I think this cuts our MLC costs in half.

watertap/costing/watertap_costing_package.py
kurbansitterley: I am not sure but I think ``aggregate_flow_costs`` is calculated in the base IDAES class and not WaterTAP's costing package, so applying ``utilization_factor`` then might not be an option. I might be wrong.

watertap/costing/zero_order_costing.py
```diff
@@ -141,97 +146,71 @@ def build_process_costs(self):
...
         self.total_variable_operating_cost = pyo.Expression(
-            expr=self.aggregate_variable_operating_cost
-            + sum(self.aggregate_flow_costs[f] for f in self.used_flows)
-            * self.utilization_factor,
+            expr=(
+                self.aggregate_variable_operating_cost
+                + sum(self.aggregate_flow_costs[f] for f in self.used_flows)
+                * self.utilization_factor
+            )
+            if self.used_flows
```
kurbansitterley, line 212: Curious why this if/else statement is needed here.

kurbansitterley (approved):
LGTM - fully support the use of Expressions in place of Var+Constraints where appropriate.

docs/technical_reference/costing/watertap_costing.rst
bknueven: I am surprised too now that I think about it more closely -- but in any case `total_capital_cost` and `aggregate_capital_cost` are now the same as the "investment factor" is now applied to each unit model's capital cost as part of #1175, and therefore included in "equipment" cost.

There can still be other indirect factors, e.g., land, which are applied at the process-level.

In any case this PR will restore the original pre-#1175 calculation (which apparently doesn't matter with the significant figures we usually check).

watertap/costing/watertap_costing_package.py
bknueven: Yes, it is done here: https://github.com/IDAES/idaes-pse/blob/87ee4c506d8b90a582270e5bb9b8943822a29b7f/idaes/core/base/costing_base.py#L470.

watertap/costing/zero_order_costing.py
bknueven: if `self.used_flows` is empty, the the `sum` on line 209 will be 0, but its units will be incorrect.

There is similar logic in the `WaterTAPCosting` package, but it wasn't here. I am guessing it hasn't come up in a ZO flowsheet, but it conceivably could. So this was added primarily to have similar code in both packages which is easier to integrate.

watertap/unit_models/zero_order/tests/test_anaerobic_digestion_reactive_zo.py
bknueven: I agree -- these kinds of tests create "make work" when we refactor the code. In any case the existence of these `Var`s should be tested by the costing package.

PR #1175, opened 2023-10-19
Costing Unification: move from `factor_total_investment` to `TIC`

bknueven:
## Summary/Motivation:
The two costing packages have different ways of accounting for indirect capital costs. The ZO Costing Package allows the indirect capital cost to be specified per unit model. The WT Costing Package has a global-level indirect capital cost.

This PR would resolve the conflict in favor of the ZO costing package. This requires modifying WaterTAP's existing detailed model's costing methods, and a few related changes to our detailed flowsheets.

- Move `TIC` / `TPEC` factor calculation to the costing base class
- Add `TIC` factor to all unit models which do not already have a factor
- Specify a default o `2` for `TIC` when used in the WT Costing Package
- Make `factor_total_investment` have a default value of `1` in the WT Costing Package
- Various changes to detailed flowsheets to reproduce existing results.
- Set automatic scaling factors on unit model costing aggregates to resolve some resulting detailed flowsheet instability (in particular, NF and LSRRO).

### Legal Acknowledgement

### DIFF
--- a/watertap/costing/unit_models/ion_exchange.py
+++ b/watertap/costing/unit_models/ion_exchange.py
@@ -336,9 +336,10 @@ def cost_ion_exchange(blk):
             to_units=blk.costing_package.base_currency,
         )
     )
+    blk.costing_package.add_cost_factor(blk, "TIC")
     blk.capital_cost_constraint = pyo.Constraint(
         expr=blk.capital_cost
-        == blk.costing_package.TIC
+        == blk.cost_factor
         * pyo.units.convert(
             (
                 ((blk.capital_cost_vessel + blk.capital_cost_resin) * tot_num_col)
--- a/watertap/costing/unit_models/ion_exchange.py
+++ b/watertap/costing/unit_models/ion_exchange.py
@@ -197,11 +197,6 @@ def build_ion_exhange_cost_param_block(blk):
         units=pyo.units.dimensionless,
         doc="Number of cycles the regenerant can be reused before disposal",
     )
-    blk.total_installed_cost_factor = pyo.Var(
-        initialize=1.65,
-        units=pyo.units.dimensionless,
-        doc="Costing factor to account for total installed cost of equipment",
-    )
 
 
 @register_costing_parameter_block(
@@ -345,8 +340,7 @@ def cost_ion_exchange(blk):
                 ((blk.capital_cost_vessel + blk.capital_cost_resin) * tot_num_col)
                 + blk.capital_cost_backwash_tank
                 + blk.capital_cost_regen_tank
-            )
-            * ion_exchange_params.total_installed_cost_factor,
+            ),
             to_units=blk.costing_package.base_currency,
         )
     )
--- a/watertap/costing/unit_models/ion_exchange.py
+++ b/watertap/costing/unit_models/ion_exchange.py
@@ -338,7 +338,8 @@ def cost_ion_exchange(blk):
     )
     blk.capital_cost_constraint = pyo.Constraint(
         expr=blk.capital_cost
-        == pyo.units.convert(
+        == blk.costing_package.TIC
+        * pyo.units.convert(
             (
                 ((blk.capital_cost_vessel + blk.capital_cost_resin) * tot_num_col)
                 + blk.capital_cost_backwash_tank
--- a/watertap/examples/flowsheets/lsrro/tests/test_lssro_paper_analysis.py
+++ b/watertap/examples/flowsheets/lsrro/tests/test_lssro_paper_analysis.py
@@ -86,18 +86,21 @@ def test_against_paper_analysis(csv_file, row_index):
         for property_name, (converter, argument) in _input_headers.items()
     }
     number_of_stages = input_arguments["number_of_stages"]
-    model, results = run_lsrro_case(
-        **input_arguments,
-        has_NaCl_solubility_limit=True,
-        permeate_quality_limit=1000e-6,
-        has_calculated_concentration_polarization=True,
-        has_calculated_ro_pressure_drop=True,
-        A_value=5 / 3.6e11,
-        B_max=None,
-        number_of_RO_finite_elements=10
-    )
+    try:
+        model, results = run_lsrro_case(
+            **input_arguments,
+            has_NaCl_solubility_limit=True,
+            permeate_quality_limit=1000e-6,
+            has_calculated_concentration_polarization=True,
+            has_calculated_ro_pressure_drop=True,
+            A_value=5 / 3.6e11,
+            B_max=None,
+            number_of_RO_finite_elements=10
+        )
+    except ValueError:
+        results = None
 
-    if check_optimal_termination(results):
+    if results is not None and check_optimal_termination(results):
         for property_name, flowsheet_attribute in _results_headers.items():
             assert value(model.find_component(flowsheet_attribute)) == pytest.approx(
                 float(row[property_name]),
--- a/watertap/unit_models/tests/test_anaerobic_digestor.py
+++ b/watertap/unit_models/tests/test_anaerobic_digestor.py
@@ -410,7 +410,7 @@ class TestAdm(object):
         assert_optimal_termination(results)
 
         # Check solutions
-        assert pytest.approx(1083290.8, rel=1e-5) == value(
+        assert pytest.approx(2.0 * 1083290.8, rel=1e-5) == value(
             m.fs.unit.costing.capital_cost
         )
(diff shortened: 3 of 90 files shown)


avdudchenko (approved):
This looks good to me.

watertap/costing/unit_models/ion_exchange.py
```diff
@@ -336,9 +336,11 @@ def cost_ion_exchange(blk):
             to_units=blk.costing_package.base_currency,
         )
     )
+    blk.costing_package.add_cost_factor(blk, "TIC")
     blk.capital_cost_constraint = pyo.Constraint(
         expr=blk.capital_cost
-        == pyo.units.convert(
+        == blk.cost_factor
```
kurbansitterley, line 337: There is currently a ``total_installed_cost_factor`` with value of 1.65 used in capital costs for IX that I had included before there was a global ``TIC``. That should be removed entirely and use the global variable to not be double counting.

watertap/costing/unit_models/ion_exchange.py
```diff
@@ -336,9 +336,11 @@ def cost_ion_exchange(blk):
             to_units=blk.costing_package.base_currency,
         )
     )
+    blk.costing_package.add_cost_factor(blk, "TIC")
```
kurbansitterley, line 334:
```suggestion
    blk.costing_package.add_cost_factor(blk, None)
```
I suppose we could also just use the local IX ``total_installed_cost_factor`` doing something like this? I am fine with the system TIC also and would let you make the decision.

watertap/costing/unit_models/ion_exchange.py
kurbansitterley: Of course whatever you decide to do it will cause the IX test to fail.

watertap/unit_models/tests/test_anaerobic_digestor.py
```diff
@@ -407,7 +407,7 @@ def test_costing(self, adm):
         assert_optimal_termination(results)
 
         # Check solutions
-        assert pytest.approx(1083290.8, rel=1e-5) == value(
+        assert pytest.approx(2.0 * 1083290.8, rel=1e-5) == value(
```
kurbansitterley, line 413: How does doubling the capital costs via this factor not affect LCOW? I don't see any LCOW tests changed.

kurbansitterley (changes requested):
There is a change needed for the ion_exchange costing model to not double count indirect costs with the use of TIC. As far as I know, that is the only unit model that had a local factor to account or indirect costs.

watertap/costing/unit_models/ion_exchange.py
bknueven: Okay, thanks for catching this. I will use the system TIC and modify IX costing model appropriately.

watertap/unit_models/tests/test_anaerobic_digestor.py
bknueven: The factor is now part of the unit model's `capital_cost` variable, as opposed to being applied at the global level.

kurbansitterley: Ok I see. 

Explanation for myself and anyone else: ``factor_capital_cost`` was set to 1 so mooted the influence of having ``TIC`` applied at the unit model level. So the value of ``aggregate_capital_cost``, which is the sum of unit model ``capital_cost``,  didn't change; ``aggregate_capital_cost``  is used in ``total_capital_cost``, which is used in ``LCOW``.

bknueven: For others, this change is being made to create uniformity around "where" the indirect capital cost is applied.

Right now the ZOCostingPackage does things at the unit-level, and the WaterTAPCosting package at the global-level. The PR modifies all the existing detailed models to put it at the unit-level.

bknueven:
@kurbansitterley I believe I have addressed your concerns around the ion exchange model in 8264aae.

kurbansitterley (approved):
LGTM

bknueven:
There are two costing PRs open, #1186 and #1196. Should this wait on those?

adam-a-a:
> There are two costing PRs open, #1186 and #1196. Should this wait on those?

Not necessarily.

watertap/examples/flowsheets/lsrro/tests/test_lssro_paper_analysis.py
```diff
@@ -86,18 +86,21 @@ def test_against_paper_analysis(csv_file, row_index):
...
+            **input_arguments,
+            has_NaCl_solubility_limit=True,
+            permeate_quality_limit=1000e-6,
+            has_calculated_concentration_polarization=True,
+            has_calculated_ro_pressure_drop=True,
+            A_value=5 / 3.6e11,
+            B_max=None,
+            number_of_RO_finite_elements=10
+        )
+    except ValueError:
```
adam-a-a, line 100: Why was this needed? To sidestep the occasional fail related to this test or something else?

bknueven: If Ipopt terminates with a weird termination condition, then Pyomo will raise a `ValueError` trying to load the results back into the model.

This simply catches this mode of failure. Note that if the solve fails (or succeeds) and we didn't expect it to the test will still fail.

bknueven:
@kurbansitterley could you have a look at 55e1754? I left a few things out on the previous commit...

kurbansitterley:
> @kurbansitterley could you have a look at [55e1754](https://github.com/watertap-org/watertap/commit/55e1754e258e7f66b935d4a834f53c31e10f3c61)? I left a few things out on the previous commit...

Yes, see that. Most important change was addressed (not double counting indirect costs) and is now reflected in the tests, it appears. So my approval remains!

bknueven:
This is currently held up by the failing notebook test `tutorials/nawi_spring_meeting2023.ipynb`.

@adam-a-a @avdudchenko any thoughts?

watertap/unit_models/tests/test_cstr.py
```diff
@@ -456,16 +457,18 @@ def test_costing(self):
...
+        results = solver.solve(m, tee=True)
 
         assert_optimal_termination(results)
 
         # Check solutions
         assert pytest.approx(526.45 * 2, rel=1e-5) == value(
             m.fs.unit.costing.capital_cost
         )
-        assert pytest.approx(1.07948e-5, rel=1e-5) == value(m.fs.costing.LCOW)
+        assert pytest.approx(8.94365e-07, rel=1e-5) == value(m.fs.costing.LCOW)
```
bknueven: @MarcusHolly can you have a look at this? It seems like this flowsheet had degrees of freedom but no objective, and therefore the LCOW could converge to an arbitrary value.

I've fixed it here to what I got out of the optimization.

MarcusHolly: Sure, I can take a look at it soon, just not right now...

bknueven:
@adam-a-a should this wait on #1213? 7d33365 might be of help with that, so we might want to merge this anyways.

adam-a-a:
> This is currently held up by the failing notebook test `tutorials/nawi_spring_meeting2023.ipynb`.
> 
> @adam-a-a @avdudchenko any thoughts?

Is this still an issue or have your latest commits resolved this?

adam-a-a:
> @adam-a-a should this wait on #1213? [7d33365](https://github.com/watertap-org/watertap/commit/7d33365e4285183cab6fa0c312f6942e49ab9efc) might be of help with that, so we might want to merge this anyways.

I suppose this doesn't need to wait on #1213 , assuming it won't break the BSM2 flowsheet.

bknueven:
> > This is currently held up by the failing notebook test `tutorials/nawi_spring_meeting2023.ipynb`.
> > @adam-a-a @avdudchenko any thoughts?
> 
> Is this still an issue or have your latest commits resolved this?

No, that was fixed by 7d33365 (making LCOW an Expression)

bknueven:
> > @adam-a-a should this wait on #1213? [7d33365](https://github.com/watertap-org/watertap/commit/7d33365e4285183cab6fa0c312f6942e49ab9efc) might be of help with that, so we might want to merge this anyways.
> 
> I suppose this doesn't need to wait on #1213 , assuming it won't break the BSM2 flowsheet.

Sounds like we should wait ... I mean, it shouldn't, but I am not extraordinarily confident either.

bknueven:
Going to wait on #1213.

PR #1176, opened 2023-10-19
MCAS Usage Example and Raising Appropriate Exceptions

### LINKED ISSUE #1071: add MCAS usage example
### Description

Thanks for the excellent work on WaterTAP. The multicomponent `MCASParameterBlock` property model model looks particularly compelling and powerful. I am requesting that you add a How-To along the lines of [How to Use a Property Model](https://watertap.readthedocs.io/en/stable/how_to_guides/how_to_use_a_property_model.html) for this model. Thanks!

### Motivation

The MCAS property model appears to be the only one in WaterTAP that can report activity coefficients of species, making it the only feasible choice for certain types of analysis (e.g., Ion Exchange). However, the example code shown in [How to Use a Property Model](https://watertap.readthedocs.io/en/stable/how_to_guides/how_to_use_a_property_model.html) is apparently not quite adequate for using this particular property model, because if I replace `NaCl_prop_pack` and `NaClParameterBlock` with `multicomp_aq_sol_prop_pack` and `MCASParameterBlock`, respectively, the example fails with an error (details in "Additional Context" section). As such, the MCAS model is effectively inaccessible to users until an example can be added to the documentation.

### Possible Implementation

_No response_

### Additional Context

Copying the example code from [How to Use a Property Model](https://watertap.readthedocs.io/en/stable/how_to_guides/how_to_use_a_property_model.html) and then replacing `NaCl_prop_pack` and `NaClParameterBlock` with `multicomp_aq_sol_prop_pack` and `MCASParameterBlock`, respectively, fails with

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
[/var/folders/pz/fdnb8v9529zftp5v9d8s9gb40000gn/T/ipykernel_72009/1390382182.py](https://file+.vscode-resource.vscode-cdn.net/var/folders/pz/fdnb8v9529zftp5v9d8s9gb40000gn/T/ipykernel_72009/1390382182.py) in ()
     13 m = ConcreteModel()
     14 m.fs = FlowsheetBlock(dynamic=False)
---> 15 m.fs.properties = MCASParameterBlock()
     16 
     17 # Build the state block and specify a time (0 = steady state).

[~/miniconda3/envs/watertap/lib/python3.8/site-packages/pyomo/core/base/block.py](https://file+.vscode-resource.vscode-cdn.net/Users/ryan/Library/CloudStorage/OneDrive-PrincetonUniversity/1_Research/4_execution/WaterTAP/~/miniconda3/envs/watertap/lib/python3.8/site-packages/pyomo/core/base/block.py) in __setattr__(self, name, val)
    578                 # Pyomo components are added with the add_component method.
    579                 #
--> 580                 self.add_component(name, val)
    581             else:
    582                 #

[~/miniconda3/envs/watertap/lib/python3.8/site-packages/pyomo/core/base/block.py](https://file+.vscode-resource.vscode-cdn.net/Users/ryan/Library/CloudStorage/OneDrive-PrincetonUniversity/1_Research/4_execution/WaterTAP/~/miniconda3/envs/watertap/lib/python3.8/site-packages/pyomo/core/base/block.py) in add_component(self, name, val)
   1183                 )
   1184             try:
-> 1185                 val.construct(data)
   1186             except:
   1187                 err = sys.exc_info()[1]

[~/miniconda3/envs/watertap/lib/python3.8/site-packages/pyomo/core/base/block.py](https://file+.vscode-resource.vscode-cdn.net/Users/ryan/Library/CloudStorage/OneDrive-PrincetonUniversity/1_Research/4_execution/WaterTAP/~/miniconda3/envs/watertap/lib/python3.8/site-packages/pyomo/core/base/block.py) in construct(self, data)
   2189                             obj.construct(data.get(name, None))
   2190                 # Trigger the (normal) initialization of the block
-> 2191                 self._getitem_when_not_present(_idx)
   2192         finally:
   2193             # We must allow that id(self) may no longer be in

[~/miniconda3/envs/watertap/lib/python3.8/site-packages/pyomo/core/base/block.py](https://file+.vscode-resource.vscode-cdn.net/Users/ryan/Library/CloudStorage/OneDrive-PrincetonUniversity/1_Research/4_execution/WaterTAP/~/miniconda3/envs/watertap/lib/python3.8/site-packages/pyomo/core/base/block.py) in _getitem_when_not_present(self, idx)
   2109 
   2110         try:
-> 2111             obj = self._rule(_block, idx)
   2112             # If the user returns a block, transfer over everything
   2113             # they defined into the empty one we created.  We do

[~/miniconda3/envs/watertap/lib/python3.8/site-packages/pyomo/core/base/initializer.py](https://file+.vscode-resource.vscode-cdn.net/Users/ryan/Library/CloudStorage/OneDrive-PrincetonUniversity/1_Research/4_execution/WaterTAP/~/miniconda3/envs/watertap/lib/python3.8/site-packages/pyomo/core/base/initializer.py) in __call__(self, parent, idx)
    311             return self._fcn(parent, *idx)
    312         else:
--> 313             return self._fcn(parent, idx)
    314 
    315 

[~/miniconda3/envs/watertap/lib/python3.8/site-packages/idaes/core/base/process_block.py](https://file+.vscode-resource.vscode-cdn.net/Users/ryan/Library/CloudStorage/OneDrive-PrincetonUniversity/1_Research/4_execution/WaterTAP/~/miniconda3/envs/watertap/lib/python3.8/site-packages/idaes/core/base/process_block.py) in _rule_default(b, *args)
     37     """
     38     try:
---> 39         b.build()
     40     except Exception:
     41         logging.getLogger(__name__).exception("Failure in build: {}".format(b))

[~/miniconda3/envs/watertap/lib/python3.8/site-packages/watertap/property_models/multicomp_aq_sol_prop_pack.py](https://file+.vscode-resource.vscode-cdn.net/Users/ryan/Library/CloudStorage/OneDrive-PrincetonUniversity/1_Research/4_execution/WaterTAP/~/miniconda3/envs/watertap/lib/python3.8/site-packages/watertap/property_models/multicomp_aq_sol_prop_pack.py) in build(self)
    314 
    315         # Group components into different sets
--> 316         for j in self.config.solute_list:
    317             if j == "H2O":
    318                 raise ConfigurationError(

TypeError: 'NoneType' object is not iterable
```

I gather from the error (and some code-diving) that I need to somehow specify either the species property database to draw from (perhaps the EDB [component.json](https://github.com/watertap-org/watertap/blob/main/watertap/edb/data/component.json)? ) or the composition of the solution when I instantiate the property block, but I can't figure out how to do so.

Interestingly, if I just directly instantiate the block (outside of a flowsheet), I don't get the error. I can also retrieve `m.fs.properties` and get back an instance of the MCAS property block

```
>>> m.fs.properties
<idaes.core.base.process_block._ScalarMCASParameterBlock at 0x7fbe39ffccc0>
```

```
>>> MCASParameterBlock()
<idaes.core.base.process_block._ScalarMCASParameterBlock at 0x7fbe39f9a130>
```

Note that I am using the `pip` version of water tap, v0.8.0, though I may try the git version soon.

adam-a-a:
## Summary/Motivation:
The MCAS property model needs a usage example (provided in the form of a how-to) for further clarity to users. Furthermore, certain exceptions should be raised were particular arguments are not provided, such as `solute_list`.

- add how-to for MCAS usage
- raise exception if no solute_list provided
- raise exception if no molecular weight data provided
- log warning if no charge data provided

### Legal Acknowledgement


watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -382,16 +382,24 @@ def build(self):
             else:
                 self.neutral_set.add(j)
 
+        if len(self.neutral_set)>0 and not len(self.ion_set):
+            _log.warning("Since the charge argument was not provided for any solutes while instantiating the MCAS property model, all solutes are assumed to be neutral and have no charge. If any ions exist in solute_list, make sure to provide charge data to avoid erroneous property calculations.")
```
adam-a-a: This warning will always show up if no charge data is provided, which isn't ideal, but in case this was unintentionally done by the user, I wanted to notify them in some way. I am open to suggestions here.

Overall, if the user mistakenly excludes charge data while instantiating MCAS, they can end up with erroneous calculations. Simultaneously, it'd be hard to know if exclusion of charge was intentional or not.

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -382,16 +382,24 @@ def build(self):
...
         # reference
         # Todo: enter any relevant references
 
+        # Check for molecular weight data
+        if not len(self.config.mw_data):
+            raise ConfigurationError("The mw_data argument was not provided while instantiating the MCAS property model. Provide a dictionary with solute names and associated molecular weights as keys and values, respectively.")
```
adam-a-a: Note to self: probably should improve this conditional to check that each solute has its own MW value provided. This would only catch the case where _no_ MW data were provided at all but would not catch a partial set of MW data.

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -382,16 +382,24 @@ def build(self):
...
+      
         # TODO: consider turning parameters into variables for future param estimation
+        mw_temp = {"H2O": 18e-3}
+        mw_temp.update(self.config.mw_data)
```
adam-a-a, line 412: Adding H2O here so that the user doesn't have to anymore. However, our older code where we had to add MW data explicitly for H2O would still work (even though it'd be unnecessary).

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -665,13 +674,13 @@ def initialize(
...
             for j in self[k].params.component_list:
-                if self.params.config.material_flow_basis == MaterialFlowBasis.molar:
+                if self[k].params.config.material_flow_basis == MaterialFlowBasis.molar:
                     if self[k].is_property_constructed("flow_mass_phase_comp"):
                         self[k].flow_mass_phase_comp["Liq", j].set_value(
                             self[k].flow_mol_phase_comp["Liq", j]
                             * self[k].params.mw_comp[j]
                         )
-                elif self.params.config.material_flow_basis == MaterialFlowBasis.mass:
+                elif self[k].params.config.material_flow_basis == MaterialFlowBasis.mass:
```
adam-a-a: Incorporated changes that @veccp identified in his PR (#1154)

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -1752,23 +1761,29 @@ def _total_hardness(self):
...
-                    if value(b.charge_comp[j]) > 1
+                    for j in self.params.cation_set
+                    if value(self.charge_comp[j]) > 1
                 ),
                 to_units=pyunits.mg / pyunits.L,
             )
-
+        except InconsistentUnitsError:
+            self.total_hardness.fix(0)
+            return
```
adam-a-a: This is a hack to handle the case where no multivalent cations are provided and `total_hardness `is called. Without this, we'd get an `InconsistentUnitsError `having to do with a zero value being rendered as `dimensionless`.

adam-a-a: Note to self: double-check that there is a test for this

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
```diff
@@ -238,7 +238,9 @@ def model2():
 
     m.fs = FlowsheetBlock(dynamic=False)
     m.fs.properties = MCASParameterBlock(
-        solute_list=["A", "B", "C", "D"], charge={"A": 1, "B": -2, "C": 2, "D": -1}
+        solute_list=["A", "B", "C", "D"], 
+        charge={"A": 1, "B": -2, "C": 2, "D": -1},
+        mw_data={"A": 10e-3, "B": 25e-3, "C": 100e-3, "D": 25e-3}
```
adam-a-a: `solute_list`: if omitted, a `ConfigurationError` is raised.
`charge`: if omitted, a warning will be logged.
`mw_data`: if omitted, a `ConfigurationError` is raised.

watertap/property_models/multicomp_aq_sol_prop_pack.py
avdudchenko: only critique here, is this will replace a user-provided value. I wounder if instead, we should check if the user provided a value for H2O in first place, and only update if the have not.

watertap/property_models/multicomp_aq_sol_prop_pack.py
avdudchenko: would it not be better to simply raise a warning/error to the user if they construct hardness vars when there are no divalent ions?

avdudchenko (approved):
LGMT

watertap/property_models/multicomp_aq_sol_prop_pack.py
adam-a-a: Addressed this

watertap/property_models/multicomp_aq_sol_prop_pack.py
adam-a-a: I kept the fix to 0 but took your suggestion and added a warning. I also added a test for this case and fixed some subtle errors around this. Also added a test to check that the logger warning shows up as expected,

watertap/property_models/multicomp_aq_sol_prop_pack.py
adam-a-a: Ah- that was also my concern, so I figured writing it this way (putting our default for H2O first, then updating the dictionary with whatever the user provided) would prevent overwriting what the user entered for "H2O".

As a sanity check, I just added one last test to verify that the user's provided value for H2O mw data wouldn't be replaced.

adam-a-a: https://github.com/watertap-org/watertap/blob/be1aeacb6cfc6b400500691d42d4cfbf4ab50c3e/watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py#L2229-L2243

avdudchenko: oh, oops. I misread the code. I thought you were updating the user-provided config with a default value. my bad. all good then.

kurbansitterley (approved):

PR #1177, opened 2023-10-20
RO w/ERD Sweep Analysis

MarcusHolly:
## Summary/Motivation:
TEA on RO w/ERD

### Legal Acknowledgement


watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
```diff
@@ -500,6 +516,12 @@ def display_system(m):
     )
     print("Feed: %.2f kg/s, %.0f ppm" % (feed_flow_mass, feed_mass_frac_NaCl * 1e6))
 
+    salt_concentration = value(
+        m.fs.feed.properties[0].flow_mass_phase_comp["Liq", "NaCl"]
+        / m.fs.feed.properties[0].flow_vol_phase["Liq"]
+    )
+    print(f"Feed salt concentration: {salt_concentration} g/L")
+
```
adam-a-a:
```suggestion
```

adam-a-a:
@MarcusHolly - I have all the cases running successfully now. Just waiting on checks to pass. So far, MPI tests fail but I am not sure why yet. I don't think it is related to this PR since it has to do with the looptool.

MarcusHolly:
> @MarcusHolly - I have all the cases running successfully now. Just waiting on checks to pass. So far, MPI tests fail but I am not sure why yet. I don't think it is related to this PR since it has to do with the looptool.

I've tried (unsuccessfully) to resolve the MPI failures with the loop tool. It does appear to be at least loosely related to some of these RO files, but I'm not sure why the failure only popped up after your recent commits because they were all fairly simple changes...

watertap/examples/flowsheets/RO_with_energy_recovery/multi_sweep.py
```diff
@@ -0,0 +1,112 @@
...
+            m.fs.feed.properties[0].conc_mass_phase_comp["Liq", "NaCl"],
+            [0.963, 1.927, 4.816],
+        )
+        sweep_params["volumetric_recovery"] = PredeterminedFixedSample(
+            m.fs.RO.recovery_vol_phase[0, "Liq"], [0.7, 0.8, 0.9]
+        )
+    else:
+        raise ValueError(f"{case_num} is not yet implemented")
+
+    global_results = parameter_sweep(
```
k1nshuk, line 97: I would recommend updating the function call here to look more like the one here. Passing a model directly to parameter sweep will be removed in the future.


https://github.com/watertap-org/watertap/blob/2c69710913ee3f61ac77fa3b9ab4abfb526b4eb6/watertap/examples/flowsheets/RO_with_energy_recovery/monte_carlo_sampling_RO_ERD.py#L122-L143

adam-a-a: @MarcusHolly we are technically using the "old" way of doing things with parameter sweep. It'd be good for us to update to the latest formulation, following @k1nshuk's example.

MarcusHolly: I've tried updating the multisweep file, but I can't figure out why it's failing to converge with these changes. I've also tried a file locally that is much closer to the `monte_carlo_sampling_RO_ERD.py`, but it runs into the same infeasible solution issue.

bknueven: We need to go through and clean up all the examples of the old API anyways.

I wouldn't worry about it if it's a bottleneck for this PR.

MarcusHolly: I've reverted this change so that we can just get this PR merged in. Perhaps there should be an issue mentioning that the old API will be deprecated. Not sure how urgent it is to update all of these files...

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
```diff
@@ -47,6 +47,7 @@
 class ERDtype(StrEnum):
     pressure_exchanger = "pressure_exchanger"
     pump_as_turbine = "pump_as_turbine"
+    no_ERD = "no_ERD"
```
adam-a-a, line 50: This option should also be listed in the exception message for `erd_type_not_found` function, line 52

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
```diff
@@ -212,11 +215,38 @@ def set_operating_conditions(
...
+
+    iscale.set_scaling_factor(
+        m.fs.P1.control_volume.properties_out[0].flow_vol_phase["Liq"], 1
+    )
+    iscale.set_scaling_factor(m.fs.P1.work_fluid[0], 1)
+    iscale.set_scaling_factor(m.fs.RO.mass_transfer_phase_comp[0, "Liq", "NaCl"], 1e4)
+    iscale.set_scaling_factor(
+        m.fs.RO.feed_side.mass_transfer_term[0, "Liq", "NaCl"], 1e4
+    )
+
```
adam-a-a, line 241: any reason why these scale factors are being set **after** calling `calculate_scaling_factors` as opposed to **before** calling it?

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
```diff
@@ -350,7 +382,7 @@ def initialize_system(m, solver=None):
     optarg = solver.options
 
     # ---initialize RO---
-    m.fs.RO.initialize(optarg=optarg)
+    m.fs.RO.initialize(optarg=optarg, outlvl=idaeslog.DEBUG)
```
adam-a-a: We can probably take away this DEBUG setting to minimize logger output by default

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
MarcusHolly: Nope... will update this

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
```diff
@@ -573,18 +616,26 @@ def print_state(s, b):
...
+        print_state("Mix out   ", m.fs.M1.outlet)
+        print_state("RO perm   ", m.fs.RO.permeate)
+        print_state("RO reten  ", m.fs.RO.retentate)
+        print_state("PXR HP out", m.fs.PXR.high_pressure_outlet)
 
 
 if __name__ == "__main__":
     # m = main(erd_type=ERDtype.pressure_exchanger)
     m = main(erd_type=ERDtype.pump_as_turbine)
+    # m = main(erd_type=ERDtype.no_ERD)
```
adam-a-a: remove commented code

watertap/examples/flowsheets/RO_with_energy_recovery/multi_sweep.py
```diff
@@ -0,0 +1,112 @@
...
+
+    m = RO.build(erd_type=ERDtype.pump_as_turbine)
+    RO.set_operating_conditions(m)
+    RO.initialize_system(m)
+    RO.solve(m)
+    m.fs.feed.properties[0].flow_mass_phase_comp.unfix()
+    m.fs.feed.properties[0].flow_vol_phase["Liq"].fix()
+    m.fs.feed.properties[0].conc_mass_phase_comp["Liq", "NaCl"].fix()
+    RO.optimize_set_up(m)
+    m.fs.eq_minimum_water_flux.deactivate()
```
adam-a-a, line 30: What happens if we remove this deactivate line? I wonder if this was something that we thought would help solution convergence across different conditions but ended up not being necessary. Can you quickly double-check whether tests fail or solutions no longer converge for cases (which I assume would be part of your testing anyhow)?

If it is indeed necessary, let's keep it; if not, let's remove this line.

watertap/examples/flowsheets/RO_with_energy_recovery/multi_sweep.py
adam-a-a: OK, we'll save this update for a separate PR that addresses most, if not all, sweep cases.

watertap/examples/flowsheets/RO_with_energy_recovery/multi_sweep.py
```diff
@@ -0,0 +1,112 @@
...
+    else:
+        raise ValueError(f"{case_num} is not yet implemented")
+
+    global_results = parameter_sweep(
+        m,
+        sweep_params,
+        outputs,
+        csv_results_file_name=output_filename,
+        optimize_function=RO.solve,
+        # optimize_kwargs=optimize_kwargs,
```
adam-a-a: delete commented code

watertap/examples/flowsheets/RO_with_energy_recovery/multi_sweep.py
MarcusHolly: Get this warning when that line is removed:
`WARNING: Loading a SolverResults object with a warning status into
model.name="unknown";
    - termination condition: maxIterations
    - message from solver: Ipopt 3.13.2\x3a Maximum Number of Iterations
      Exceeded.`

So I'll keep it in.

adam-a-a (approved):
LGTM- this PR is ready to go- you just need to finish up addressing a couple of minor comments, so I am approving in anticipation of you handling those changes. Great job!

watertap/examples/flowsheets/RO_with_energy_recovery/multi_sweep.py
```diff
@@ -0,0 +1,111 @@
...
+    # outputs["RO Permeate Salt Mass Flow"] = m.fs.RO.permeate.flow_mass_phase_comp[
+    #     0, "Liq", "NaCl"
+    # ]
+    # outputs["RO Retentate H2O Mass Flow"] = m.fs.RO.retentate.flow_mass_phase_comp[
+    #     0, "Liq", "H2O"
+    # ]
+    # outputs["RO Retentate Salt Mass Flow"] = m.fs.RO.retentate.flow_mass_phase_comp[
+    #     0, "Liq", "NaCl"
+    # ]
+
```
luohezhiming, line 65: Do we want to keep these commented outputs? I know it's used for different simulations, but might make more sense in a local version.

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
```diff
@@ -459,8 +500,8 @@ def optimize_set_up(m):
     # RO
     m.fs.RO.area.unfix()
     m.fs.RO.area.setlb(1)
-    m.fs.RO.area.setub(150)
-
+    m.fs.RO.area.setub(1500)
+    m.fs.RO.feed_side.friction_factor_darcy[...].setub(None)
```
luohezhiming, line 504: It seems necessary for some of the sweep analysis to get a solution, I am curious what is the original upper bound and what if we keep it, does most sweep runs fail?

luohezhiming (approved):
It looks great, I only have some questions about some minor details, which I don't think must be addressed in this PR.

watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
MarcusHolly: The original upper bound appears to be 5, and you're right that if we remove this upper bound, the flowsheet is still able to solve, but the sweeps will fail (specifically those at 90% recovery)

watertap/examples/flowsheets/RO_with_energy_recovery/multi_sweep.py
MarcusHolly: I'd like to keep these here just so that I _don't_ have to keep a local version (and neither does anyone else who wants to run the TEA analysis we did for the dye desal team). But if we think it really should be removed, I can do that.

Issue #1173, opened 2023-10-16
Merge in Stakeholder Code Demos


adam-a-a:
This is a reminder to put up PRs for and merge code demos presented at the IDAES stakeholder meeting. I thought the demos were all great, and we should make them accessible to all.

Please add any that I may be omitting. Note: some may require more revision than others (e.g., if new code was shown that hasn't been reviewed/merged yet), but we should aim to get everything merged in after some iteration:

- [x] adding custom constraints to models- @avdudchenko #1163 
- [x] parameter sweep demo - @k1nshuk #1181
- [x] parameter estimation demo - @OOAmusat @savannahsakhai #1182
- [x] OLI demo - @veccp #1141 
- [x] connecting flowsheets to UI - @MichaelPesce #1152
- [x] multi-period modeling - @zacharybinger (will reference on watertap-reflo issue too since it might belong in that repo since we use pysam)

PR #1182, opened 2023-10-26
Parameter Estimation (parmest) Demo

savannahsakhai:
## Fixes/Resolves:
This is one item of the issue: Merge in Stakeholder Code Demos #1173.

## Summary/Motivation:
This adds the parameter estimation using parmest code demonstration given at the 2023 PSE+ stakeholder summit. 

- adds a jupyter notebook for the code demo
- adds a csv for the necessary data from phreeqc

### Legal Acknowledgement


OOAmusat (approved):
My minor comment here would be maybe changing the format/structure of the input data, but that's not a reason to delay this.

Otherwise, LGTM.

OOAmusat:
For a bit more context on why I commented on fixing the structure of the input data, this is what the data file looks like when I open it locally with Excel:

image

Cell A1 contains the number "55" which is not really part of the dataset. However, deleting the value and leaving the cell empty leads to failures in the notebook (at least on my end).

In the notebook, the data is pre-processed and the cell is ignored, so it is not a problem per-se. However, it is something we could (and probably should) easily fix.

savannahsakhai:
> For a bit more context on why I commented on fixing the structure of the input data, this is what the data file looks like when I open it locally with Excel:
> 
> image
> Cell A1 contains the number "55" which is not really part of the dataset. However, deleting the value and leaving the cell empty leads to failures in the notebook (at least on my end).
> 
> In the notebook, the data is pre-processed and the cell is ignored, so it is not a problem per-se. However, it is something we could (and probably should) easily fix.

Thank you for adding more context to your original comment! I understand your concern now. If you look at the data provided in this PR, it should be corrected. I have updated the data set since the stakeholder summit. This new data was generated using phreeqc and the format should exclude that random placeholder number.

TimBartholomew (approved):

adam-a-a (approved):
Great job, @savannahsakhai! Will set to automerge

PR #1183, opened 2023-10-28
multiple choice costing

bknueven:
## Summary/Motivation:
Via @adam-a-a, @hunterbarber needs an easy way to "swap out" costing methods on the same flowsheet. This PR would implement wrapper class around `UnitModelCostingBlock` to enable this.

The test demonstrates the basic usage. If we think this is valuable I could add documentation, more robust tests, etc.

-
-

### Legal Acknowledgement


avdudchenko (approved):
Looks good to me overall. Only suggestion is, maybe instead of multiple choice costing block it should be indexed costing block? This would be consistent with other naming methodology, but also with relative implementation if I am reading it right (e.g. costing_block['a'])

bknueven:
> Looks good to me overall. Only suggestion is, maybe instead of multiple choice costing block it should be indexed costing block? This would be consistent with other naming methodology, but also with relative implementation if I am reading it right (e.g. costing_block['a'])

I think "indexed" would potentially cause more confusion, as by default all indexed blocks are active at once. In this case, exactly one of the costing blocks should be active, and all the others should be "inactive".

After thinking about it some more though I think I should change the API slightly to be less cute and more explicit.

In general, it seems like this could be potentially useful so I will put some additional effort into testing and documentation.

hunterbarber:
> > Looks good to me overall. Only suggestion is, maybe instead of multiple choice costing block it should be indexed costing block? This would be consistent with other naming methodology, but also with relative implementation if I am reading it right (e.g. costing_block['a'])
> 
> I think "indexed" would potentially cause more confusion, as by default all indexed blocks are active at once. In this case, exactly one of the costing blocks should be active, and all the others should be "inactive".
> 
> After thinking about it some more though I think I should change the API slightly to be less cute and more explicit.
> 
> In general, it seems like this could be potentially useful so I will put some additional effort into testing and documentation.

I'll tell you a bit of the details as they may help with that. So without getting into mixed integer optimization, I was looking to cost a unit model via two different methods controlled by a config argument, and select the cheaper option. I wanted both of them to exist on the same `m` block so they would be callable by a single `solve(m)`. Currently, a unit model may only have one `UnitModelCostingData` assoicated with it so two `UnitModelCostingBlock`s wasn't an option. I also wanted to compare the LCOW which have to also exist on different `WaterTAPCostingData`. By this I can manually compare the results with some additional code instead of optimizing the config argument.

I'm doing it via cloning the `UnitModelBlock` and creating a second `WaterTAPCostingData`, (having `m.fs.costing_pressure`, `m.fs.gac.costing_pressure`, `m.fs.costing_gravity`, and `m.fs.gac_clone.costing_gravity`) but this solves the model twice which eats up unnecessary computational time. So the ideal method would be to have both config options applied on the same model, select the cheaper option, and then apply it to the flowsheet costing for LCOW. Maybe this would've been doable through a dummy continuous variable treated as boolean for the config option and switching from a simulation to an optimization, but not sure and it also would've been some effort to implement.

I think this makes it toggle-able but not technically what I would've needed without some more handling code at the flowsheet level to select the better option. Is that correct?

bknueven:
> I'll tell you a bit of the details as they may help with that. So without getting into mixed integer optimization...
> I think this makes it toggle-able but not technically what I would've needed without some more handling code at the flowsheet level to select the better option. Is that correct?

These two things are in contraction. It's a little bit strange to me from a math programming perspective that the solver would get to "choose" the costing method. Generally I would think that would be determined by the modeler.

This framework could be fairly readily changed over to use disjunctions, etc., and solve with GDPOpt, or to a regular MINLP formulation and solved with one of the MINLP solvers we have available. In general this type of formulation isn't possible to solve with pure Ipopt unless the flowsheet (or maybe just all the costing methods) meet some technical assumptions. Otherwise you could be in a situation where the "optimal" solution is to take 30% from costing method "A" and 70% from costing method "B".

I hesitate to add such things to our codebase because they are likely to cause confusion more than anything.

hunterbarber:
> > I'll tell you a bit of the details as they may help with that. So without getting into mixed integer optimization...
> > I think this makes it toggle-able but not technically what I would've needed without some more handling code at the flowsheet level to select the better option. Is that correct?
> 
> These two things are in contraction. It's a little bit strange to me from a math programming perspective that the solver would get to "choose" the costing method. Generally I would think that would be determined by the modeler.
> 
> This framework could be fairly readily changed over to use disjunctions, etc., and solve with GDPOpt, or to a regular MINLP formulation and solved with one of the MINLP solvers we have available. In general this type of formulation isn't possible to solve with pure Ipopt unless the flowsheet (or maybe just all the costing methods) meet some technical assumptions. Otherwise you could be in a situation where the "optimal" solution is to take 30% from costing method "A" and 70% from costing method "B".
> 
> I hesitate to add such things to our codebase because they are likely to cause confusion more than anything.

Just to be clear it's not a config option for a different costing equation, its a config option to select whether the construction is an open concrete basin or closed steel pressurized vessel which inherently have different costing expressions. So it's not optimizing the costing equation, it is optimizing the design based on costing.

And yea I'm aware of the challenges with MINLP in the framework. That's why I wanted a method that just calculated both on the solve and I could select between the two just based on if option_1 < option_2: select option_1 after the solution is available.

watertap/costing/tests/test_multiple_choice_costing_block.py
```diff
@@ -0,0 +1,102 @@
...
+    m.fs.costing = WaterTAPCosting()
+
+    m.fs.RO = ReverseOsmosis0D(
+        property_package=m.fs.properties,
+        has_pressure_change=True,
+        pressure_change_type=PressureChangeType.calculated,
+        mass_transfer_coefficient=MassTransferCoefficient.calculated,
+        concentration_polarization_type=ConcentrationPolarizationType.calculated,
+    )
+    m.fs.RO.costing = MultipleChoiceCostingBlock(
```
adam-a-a: maybe something shorter like `MultiCostingBlock`

adam-a-a: or `CostingBlockChoices` even though I'd prefer something as concise as possible

watertap/costing/multiple_choice_costing_block.py
```diff
@@ -0,0 +1,209 @@
...
+                            f"{unit_model.name} {v} component must be a Var. "
+                            "Please check the costing package you are using to "
+                            "ensure that all costing components are declared as "
+                            "variables."
+                        )
+                    elif cvar.lb is None or cvar.lb < 0:
+                        _log.warning(
+                            f"{unit_model.name} {v} component has a lower bound "
+                            "less than zero. Be aware that this may result in "
+                            "negative costs during optimization."
```
adam-a-a, line 165: We could have negative components to account for revenue. Is that included in `DefaultCostingComponents` ?

bknueven: Yes -- this line is actually from IDAES such that we replicate the existing behavior, warts and all. I don't think the user should notice a difference between using a `UnitModelCostingBlock` or a (tentatively named) `MultipleChoiceCostingBlock` with a single method and get different behavior.

bknueven:
Modulo a name change I think this is ready to go.

The base name we inherit from IDAES is `UnitModelCostingBlock`. So I think whatever we go with should reflect that route. E.g.,

- `UnitModelCostingBlockChoices`
- `MultiChoiceUnitModelCostingBlock`
- `MultiChoiceCostingBlock` ( unit model is implicit )
- `CostingBlockChoices` ( unit model is implicit )
- `CostSelectorBlock`
- others...?

adam-a-a:
@bknueven I made a poll to choose the name we want since we had no action on this (and thought it would be fun to try out) #1197

hunterbarber:
So both costing options are added to the flowsheet costing options with this implementation, i.e., costing option 1 and costing option 2 for the same unit op both contribute to LCOW. The multi choice is working for me, but now I'm thinking about how to handle this.
I think maybe having a config option on multi choice costing block to only have one option contribute to flowsheet cost metrics may be necessary. Either having a primary and secondary where the primary is the one to contribute to LCOW and the second is viewable but inactive in terms of flowsheet costing; or config option to smooth min the two LCOW contributions between the two costing options.

hunterbarber:
Another nuance I want to identify. The way that the current GAC costing is set up, this actually doesn't work because variables are built on `m.fs.costing.gac`, not on `m.fs.costing.gac.costing_block[option_n]`. So even though in the `cost_gac` function the costing coefficients depend on the configuration argument, they are not being overwritten properly.
```
    # with ContactorType not assigned in build_gac_cost_param_block and blk.costing_package.gac variables
    # fixed when register_costing_parameter_block, refix parameters based on contactor type here
    print(contactor_type == ContactorType.pressure)
    # costing data parameters based on contactor type
    adsorbent_unit_cost_coeff_data = {0: 4.58342, 1: -1.25311e-5}
    if contactor_type == ContactorType.pressure:
        contactor_cost_coeff_data = {0: 10010.9, 1: 2204.95, 2: -15.9378, 3: 0.110592}
        other_cost_param_data = {0: 16660.7, 1: 0.552207}
        energy_consumption_coeff_data = {0: 8.09926e-4, 1: 8.70577e-4, 2: 0}
    elif contactor_type == ContactorType.gravity:
        contactor_cost_coeff_data = {0: 75131.3, 1: 735.550, 2: -1.01827, 3: 0.000000}
        other_cost_param_data = {0: 38846.9, 1: 0.490571}
        energy_consumption_coeff_data = {0: 0.123782, 1: 0.132403, 2: -1.41512e-5}
    else:
        raise ConfigurationError(
            f"{blk.unit_model.name} received invalid argument for contactor_type:"
            f" {contactor_type}. Argument must be a member of the ContactorType Enum."
        )

    # iterable matching coeff_data to vars
    gac_cost = blk.costing_package.gac
    cost_params = (
        (adsorbent_unit_cost_coeff_data, gac_cost.adsorbent_unit_cost_coeff),
        (contactor_cost_coeff_data, gac_cost.contactor_cost_coeff),
        (other_cost_param_data, gac_cost.other_cost_param),
        (energy_consumption_coeff_data, gac_cost.energy_consumption_coeff),
    )
```
My short term solution is to calculate LCOW for each isolated costing option using a custom function because there's only one unit op on my flowsheet, so I can just look at the CAPEX and OPEX for each individual costing option on the unit op. Then I had to move the constructed coefficient parameters onto the costing option blocks instead of the main costing block. It's working, but may be helpful to decide what we want to do on this.

hunterbarber:
Parameter sweep also looks for these variables when solve is unsuccessful, and they are not constructed on the multiple choice costing block
```
"fs.gac.costing.<DefaultCostingComponents.capital: 'capital_cost'>": {'value': array([0.]), 'full_name': "fs.gac.costing<DefaultCostingComponents.capital: 'capital_cost'>"},
"fs.gac.costing.<DefaultCostingComponents.fixed: 'fixed_operating_cost'>": {'value': array([0.]), 'full_name': "fs.gac.costing.<DefaultCostingComponents.fixed: 'fixed_operating_cost'>"},```

bknueven:
> Another nuance I want to identify. The way that the current GAC costing is set up, this actually doesn't work because variables are built on `m.fs.costing.gac`, not on `m.fs.costing.gac.costing_block[option_n]`. So even though in the `cost_gac` function the costing coefficients depend on the configuration argument, they are not being overwritten properly.
> 
> ```
>     # with ContactorType not assigned in build_gac_cost_param_block and blk.costing_package.gac variables
>     # fixed when register_costing_parameter_block, refix parameters based on contactor type here
>     print(contactor_type == ContactorType.pressure)
>     # costing data parameters based on contactor type
>     adsorbent_unit_cost_coeff_data = {0: 4.58342, 1: -1.25311e-5}
>     if contactor_type == ContactorType.pressure:
>         contactor_cost_coeff_data = {0: 10010.9, 1: 2204.95, 2: -15.9378, 3: 0.110592}
>         other_cost_param_data = {0: 16660.7, 1: 0.552207}
>         energy_consumption_coeff_data = {0: 8.09926e-4, 1: 8.70577e-4, 2: 0}
>     elif contactor_type == ContactorType.gravity:
>         contactor_cost_coeff_data = {0: 75131.3, 1: 735.550, 2: -1.01827, 3: 0.000000}
>         other_cost_param_data = {0: 38846.9, 1: 0.490571}
>         energy_consumption_coeff_data = {0: 0.123782, 1: 0.132403, 2: -1.41512e-5}
>     else:
>         raise ConfigurationError(
>             f"{blk.unit_model.name} received invalid argument for contactor_type:"
>             f" {contactor_type}. Argument must be a member of the ContactorType Enum."
>         )
> 
>     # iterable matching coeff_data to vars
>     gac_cost = blk.costing_package.gac
>     cost_params = (
>         (adsorbent_unit_cost_coeff_data, gac_cost.adsorbent_unit_cost_coeff),
>         (contactor_cost_coeff_data, gac_cost.contactor_cost_coeff),
>         (other_cost_param_data, gac_cost.other_cost_param),
>         (energy_consumption_coeff_data, gac_cost.energy_consumption_coeff),
>     )
> ```

This logic is very bittle, IMHO. If we have two GAC models of different types, whichever one was costed *last* will have its parameters in effect.

Instead we should create two different parameter blocks and utilize whichever one is "right" for this particular unit model. It will have the side effect of eliminating this issue. I will open a PR which does this.

hunterbarber:
> > Another nuance I want to identify. The way that the current GAC costing is set up, this actually doesn't work because variables are built on `m.fs.costing.gac`, not on `m.fs.costing.gac.costing_block[option_n]`. So even though in the `cost_gac` function the costing coefficients depend on the configuration argument, they are not being overwritten properly.
> > ```
> >     # with ContactorType not assigned in build_gac_cost_param_block and blk.costing_package.gac variables
> >     # fixed when register_costing_parameter_block, refix parameters based on contactor type here
> >     print(contactor_type == ContactorType.pressure)
> >     # costing data parameters based on contactor type
> >     adsorbent_unit_cost_coeff_data = {0: 4.58342, 1: -1.25311e-5}
> >     if contactor_type == ContactorType.pressure:
> >         contactor_cost_coeff_data = {0: 10010.9, 1: 2204.95, 2: -15.9378, 3: 0.110592}
> >         other_cost_param_data = {0: 16660.7, 1: 0.552207}
> >         energy_consumption_coeff_data = {0: 8.09926e-4, 1: 8.70577e-4, 2: 0}
> >     elif contactor_type == ContactorType.gravity:
> >         contactor_cost_coeff_data = {0: 75131.3, 1: 735.550, 2: -1.01827, 3: 0.000000}
> >         other_cost_param_data = {0: 38846.9, 1: 0.490571}
> >         energy_consumption_coeff_data = {0: 0.123782, 1: 0.132403, 2: -1.41512e-5}
> >     else:
> >         raise ConfigurationError(
> >             f"{blk.unit_model.name} received invalid argument for contactor_type:"
> >             f" {contactor_type}. Argument must be a member of the ContactorType Enum."
> >         )
> > 
> >     # iterable matching coeff_data to vars
> >     gac_cost = blk.costing_package.gac
> >     cost_params = (
> >         (adsorbent_unit_cost_coeff_data, gac_cost.adsorbent_unit_cost_coeff),
> >         (contactor_cost_coeff_data, gac_cost.contactor_cost_coeff),
> >         (other_cost_param_data, gac_cost.other_cost_param),
> >         (energy_consumption_coeff_data, gac_cost.energy_consumption_coeff),
> >     )
> > ```
> 
> This logic is very bittle, IMHO. If we have two GAC models of different types, whichever one was costed _last_ will have its parameters in effect.
> 
> Instead we should create two different parameter blocks and utilize whichever one is "right" for this particular unit model. It will have the side effect of eliminating this issue. I will open a PR which does this.

I think GAC was done in this way because parameters such as `membrane_replacement_factor` for RO make sense to specify globally onto the flowsheet (and to minimize the amount of times the same parameters get constructed on different blocks), so these parameters were implemented by the same convention. I already have the file which moves these variables over to the costing block that is constructed in `cost_gac` as opposed to the variables that initially get built if that is what you are suggesting.

adam-a-a (approved):
BDFL's name it is then!

Issue #1117, opened 2023-08-31
Models to add to the UI


dangunter:
### Description

List of flowsheets to "wrap" so they can be accessed through the UI

### Motivation

Need POC and flowsheet locations. Will need help choosing list of variables, etc.

### Additional Context

- Models that are currently UI-enabled are not as interesting/complex as others we have
- Therefore, we want to have more of the complex models (FIXME) to become UI-enabled to showcase the full potential of the GUI

## Wishlist
In order of priority

- [x] RO with energy recovery #1159 
- [x] IEDO full WWT (in addition to the Jupyter Notebook tutorial) (conventional BSM2 - as opposed to the extended version that's not fully working yet) #1125 
- [x] IEDO dye desal #1126 
- [ ] LSSRO (@zacharybinger, @adam-a-a) #1383 
- [ ] ED (@lbibl) #1439 
- [x] MVC (@MarcusHolly, @TimBartholomew) #1185 
- [x] GAC (@hunterbarber)  #1238
- [x] OARO (@luohezhiming) #1290

## UI questions

- Once we have these in, is there a way to group/identify/hide the simpler (current) models?
  - @MichaelPesce: working on categories
  - @adam-a-a what about "desalination & wastewater treatment"? or "detailed" and "simple"?

hunterbarber:
@adam-a-a @TimBartholomew So I have a surrogate that reduces the parameter input for the CPHSDM GAC model. In all it adds two surrogate constraints to replace 7 parameters obtained from a lookup table. I think the plan is that once the corresponding manuscript is submitted/published, to add those surrogates as an config option to the model. With the GUI models, I'm assuming lower parameter input is preferred. Should the GAC hold off until the surrogates have been implemented in main or should the GAC model for the GUI include these parameters that are readily available, but need looked up in literature as a function of other values in the model?

ksbeattie:
@dangunter  do you want to split apart the issues that will be in the March release into a separate issue and close this one for the Dec release?

adam-a-a:
@hunterbarber I think the UI should have the 7 parameters (the As and Bs I am assuming) with defaults in there. Later we could think about adding the option to switch to a custom surrogate, with yours surrogates being the default forms.

hunterbarber:
> @hunterbarber I think the UI should have the 7 parameters (the As and Bs I am assuming) with defaults in there. Later we could think about adding the option to switch to a custom surrogate, with yours surrogates being the default forms.

[...]

PR #1185, opened 2023-10-30
MVC GUI

MarcusHolly:
## Summary/Motivation:

Issue #1117 

### Legal Acknowledgement


MarcusHolly:
@TimBartholomew Is there a figure you have to depict this flowsheet?

GUI Screenshots:
image
image
image

TimBartholomew:
@carson-tucker can you look at this and see if all the key inputs and outputs are there, or if you suggest anything else should or should not be included.

carson-tucker:
Mostly looks good to me. The split fraction for the feed separator is free, but it is fixed to the recovery ratio in the model. I think maybe it should be included as an output instead because a user may change the split fraction to be different from the recovery and get an infeasible result.

MarcusHolly:
> Mostly looks good to me. The split fraction for the feed separator is free, but it is fixed to the recovery ratio in the model. I think maybe it should be included as an output instead because a user may change the split fraction to be different from the recovery and get an infeasible result.

Sounds good. I'll try to have this change made today.

dangunter:
You should add `(...,**kwargs)` to `build_flowsheet` and `export_variables` function signatures, to handle `build_options=` parameter and any future parameters being passed as keyword args.

lbianchi-lbl:
@MarcusHolly to work around the failing fsapi test on macOS, you can add `or mvc.mvc_single_stage_ui` to the list of deselected pytest keywords in the last step of the `macos` CI job:

https://github.com/watertap-org/watertap/blob/094da03c162d87c470615db65a69b5bf0af403ef/.github/workflows/checks.yml#L283

watertap/examples/flowsheets/mvc/mvc_single_stage_ui.py
```diff
@@ -0,0 +1,649 @@
...
+
+    m.fs.Q_ext[0].fix(0)  # no longer want external heating in evaporator
+    del m.fs.objective
+    set_up_optimization(m)
+    results = solve(m, solver=solver, tee=False)
+
+    return m
+
+
+def solve_flowsheet(flowsheet=None):
```
bknueven, line 651:
```suggestion
def solve_flowsheet(flowsheet):
```
Not sure why this would be an optional argument.

watertap/examples/flowsheets/mvc/mvc_single_stage_ui.py
```diff
@@ -0,0 +1,649 @@
...
+        display_units="$(2001)",
+        rounding=2,
+        description="Compressor unit cost",
+        is_input=True,
+        input_category="System costing",
+        is_output=False,
+    )
+
+    # Outlets
+    feed_salinity = 1e3 * fs.feed.properties[0].mass_frac_phase_comp["Liq", "TDS"]
```
bknueven: Not sure why we're manually converting units here.

watertap/examples/flowsheets/mvc/mvc_single_stage_ui.py
```diff
@@ -0,0 +1,649 @@
...
+        display_units="g/kg",
+        rounding=2,
+        description="Feed salinity",
+        is_input=False,
+        is_output=True,
+        output_category="Outlets",
+    )
+    brine_salinity = (
+        1e3 * fs.evaporator.properties_brine[0].mass_frac_phase_comp["Liq", "TDS"]
+    )
```
bknueven: Same here regarding units.

Does the `ui_units` argument not do that automatically?

bknueven (approved):

watertap/examples/flowsheets/mvc/mvc_single_stage_ui.py
MarcusHolly: I'm not sure of the logic behind it either, but it seems like all of the ui files do this. @dangunter or @MichaelPesce Would there be any problems with making this change?

watertap/examples/flowsheets/mvc/mvc_single_stage_ui.py
MarcusHolly: I think the issue is that we're going from dimensionless (`mass_frac_phase_comp`) to g/kg, so I think this manual conversion is necessary? This is also how it's done in the `def display_metrics` in the MVC flowsheet itself

watertap/examples/flowsheets/mvc/mvc_single_stage_ui.py
MichaelPesce: I see no reason why this would be a problem

watertap/examples/flowsheets/mvc/mvc_single_stage_ui.py
bknueven: Seems okay on my system...

Screenshot 2023-12-14 at 1 37 38 PM

MarcusHolly: Alright, I've procrastinated long enough. I automatically convert the units and change `flowsheet=None` to just `flowsheet` now.

watertap/examples/flowsheets/mvc/mvc_single_stage_ui.py
MarcusHolly: I've tested this and ui_units does not do this conversion automatically.

TimBartholomew (approved):

PR #1186, opened 2023-11-07
BSM2 Anoxic CSTR Costing

MarcusHolly:
## Summary/Motivation:
Adds costing method for anoxic CSTRs

### Legal Acknowledgement


watertap/costing/unit_models/cstr.py
```diff
@@ -0,0 +1,80 @@
...
+
+# class CSTRType(StrEnum):
+#     anoxic = "anoxic"
+#     anaerobic = "anaerobic"
+#     aerobic = "aerobic"
+
+
+def build_cstr_cost_param_block(blk):
+    # Hovers in the 3-5 range in the literature, but the exact value should be determined w/optimization
+    blk.HRT = pyo.Var(
```
adam-a-a: THis should be on the unit model.

luohezhiming (commented):
Probably we want to add a testing file for the costing, rather than direclty add to the current BSM2 cause other unit models' costings are still pending.

MarcusHolly:
> Probably we want to add a testing file for the costing, rather than directly add to the current BSM2 cause other unit models' costings are still pending.

I could be wrong, but I don't think we have tests for any of the other costing models? I don't see an obvious problem with just adding costing to the BSM2 flowsheet piece-by-piece

luohezhiming:
> > Probably we want to add a testing file for the costing, rather than directly add to the current BSM2 cause other unit models' costings are still pending.
> 
> I could be wrong, but I don't think we have tests for any of the other costing models? I don't see an obvious problem with just adding costing to the BSM2 flowsheet piece-by-piece

The other costing tests are within their specific test files, like if you check `test_electroNP_ZO.py`, there is a test for costing

MarcusHolly:
> > > Probably we want to add a testing file for the costing, rather than directly add to the current BSM2 cause other unit models' costings are still pending.
> > 
> > 
> > I could be wrong, but I don't think we have tests for any of the other costing models? I don't see an obvious problem with just adding costing to the BSM2 flowsheet piece-by-piece
> 
> The other costing tests are within their specific test files, like if you check `test_electroNP_ZO.py`, there is a test for costing

Would that belong in WaterTAP or IDAES tho? For example, the HEX and mixer are IDAES models that we added WaterTAP costing for, and neither of these models have tests associated with them from what I can see. Although maybe we should add tests for these...

watertap/examples/flowsheets/case_studies/full_water_resource_recovery_facility/BSM2.py
```diff
@@ -442,7 +441,27 @@ def function(unit):
 
 def add_costing(m):
     # TODO: implement unit model and flowsheet level costing
-    pass
+    m.fs.costing = WaterTAPCosting()
+    m.fs.costing.base_currency = pyo.units.USD_2020
+
+    # Costing Blocks
+    m.fs.MX1.costing = UnitModelCostingBlock(flowsheet_costing_block=m.fs.costing)
```
adam-a-a: Not sure we should include all the mixers in costing right away.

bknueven:
> Would that belong in WaterTAP or IDAES tho? For example, the HEX and mixer are IDAES models that we added WaterTAP costing for, and neither of these models have tests associated with them from what I can see. Although maybe we should add tests for these...

The test should be in WaterTAP, probably in the costing directory.

There's not a test for the Mixer because we never wrote one. There's an open issue (#1187) to make sure all the costing methods have a unit test. Most do, but a few do not.

watertap/costing/unit_models/cstr.py
```diff
@@ -0,0 +1,69 @@
...
+    )
+
+    print(f"base_currency: {blk.costing_package.base_currency}")
+    blk.capital_cost_constraint = pyo.Constraint(
+        expr=blk.capital_cost
+        == pyo.units.convert(
+            HRT * flow_in * blk.sizing_cost,
+            to_units=blk.costing_package.base_currency,
+        )
+    )
```
bknueven: You could probably replace this with `cost_by_flow_volume(blk, HRT * blk.sizing_cost, flow_in)`.
https://github.com/watertap-org/watertap/blob/main/watertap/costing/util.py#L222

MarcusHolly:
@bknueven I'll make your `cost_by_flow_volume` change, but do you know why the UI tests are failing in the linux builds? Maybe this is more of a question for Dan...

watertap/unit_models/cstr.py
```diff
@@ -0,0 +1,84 @@
...
+# Technology Laboratory (subject to receipt of any required approvals from
+# the U.S. Dept. of Energy). All rights reserved.
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and license
+# information, respectively. These files are also available online at the URL
+# "https://github.com/watertap-org/watertap/"
+#
+###############################################################################
+"""
+Anoxic CSTR unit model for BSM2 and plant-wide wastewater treatment modeling.
```
adam-a-a: Rename filename to `anoxic_cstr` instead of `cstr`. Either that, or rename the class which is currently named `AnoxicCSTR`. The only dilemma I see with this sort of naming is that anoxic is not anaerobic, technically. So maybe it makes more sense to rename the class as CSTR, which would just build on top of IDAES' CSTR and be the WaterTAP version,

watertap/unit_models/cstr.py
```diff
@@ -0,0 +1,84 @@
...
+        Args:
+            None
+        Returns:
+            None
+        """
+
+        # Call UnitModel.build to set up dynamics
+        super(AnoxicCSTRData, self).build()
+
+        self.HRT = Var(
```
adam-a-a: I know that we used the acronym (HRT) in ZO models previously. I spelled it out in my recent PR and I think @agarciadiego did too. We should be consistent and spell it out here too.

adam-a-a (changes requested):
Looks good- just some quick changes needed

watertap/unit_models/cstr.py
```diff
@@ -0,0 +1,84 @@
...
+from watertap.costing.unit_models.cstr import cost_cstr
+
+__author__ = "Marcus Holly"
+
+
+# Set up logger
+_log = idaeslog.getLogger(__name__)
+
+
+@declare_process_block_class("CSTR")
```
MarcusHolly, line 42: Renaming this from AnoxicCSTR to CSTR is causing a pylint issue because it's overwriting the name that IDAES uses. Is there a way to get around this and, if not, what should the new name be?

lbianchi-lbl: I'm not 100% sure, but is it at all possible that the thing Pylint is complaining about is that a derived class is being defined with the same name as its base class? Since `CSTRData` was imported, defining a new `CSTRData` class would result in a different object (i.e. the derived `CSTRData`) reusing the same name as another object (i.e. the `CSTRData` imported from `idaes.models.unit_models.cstr`).

One way to check this could be to change the import alias, e.g. `from idaes.models.unit_models.cstr import CSTRData as CSTRDataFromIDAES` (or something).

MarcusHolly: @lbianchi-lbl I believe you're right - I just didn't explain it well enough in my initial comment. I'll incorporate your change and see if it resolves the issue. Thanks!

Edit: It resolved the issue

watertap/unit_models/tests/test_cstr.py
```diff
@@ -0,0 +1,446 @@
...
+        m.fs.unit.inlet.temperature[0].set_value(308.15)
+        m.fs.unit.inlet.pressure[0].set_value(84790.0)
+
+        m.fs.unit.volume[0].fix(1000)
+
+        m.fs.costing = WaterTAPCosting()
+
+        m.fs.unit.costing = UnitModelCostingBlock(flowsheet_costing_block=m.fs.costing)
+        m.fs.costing.cost_process()
+        m.fs.costing.add_LCOW(m.fs.unit.control_volume.properties_out[0].flow_vol)
```
adam-a-a, line 458: Can you also `assert_units_consistent`? I am skeptical about the units in the cost equation.

watertap/costing/unit_models/cstr.py
```diff
@@ -0,0 +1,45 @@
...
+from ..util import (
+    register_costing_parameter_block,
+    cost_by_flow_volume,
+)
+
+
+def build_cstr_cost_param_block(blk):
+    # Source: https://www.fwrj.com/articles/9812.pdf
+    blk.sizing_cost = pyo.Var(
+        initialize=0.34,
```
adam-a-a, line 23: This is the unit cost for the largest size shown in the range from the reference. Perhaps we should choose a value like the mean or median.

watertap/unit_models/tests/test_cstr.py
```diff
@@ -0,0 +1,446 @@
...
+        results = solver.solve(m)
+
+        assert_optimal_termination(results)
+
+        # Check solutions
+        assert pytest.approx(526.45, rel=1e-5) == value(m.fs.unit.costing.capital_cost)
+        assert pytest.approx(9.0847e-8, rel=1e-5) == value(m.fs.costing.LCOW)
```
adam-a-a: THis is the only check failing (on Mac) which leads me to believe that a tweak in scaling might resolve. However, this LCOW is approximately 0, and I am not totally confident that the result should/would be that low.

watertap/costing/unit_models/cstr.py
MarcusHolly: The reason I chose the largest size is because the flowrate in BSM2 actually goes way beyond the ranges listed in this reference. The largest flow in the reference is 100,000 GPD (378.5 m3/day) whereas the BSM2 feed flowrate is 20648 m3/day. So I didn't think it made sense to choose a mean or median value. Although if we intend on using this CSTR for non-BSM2 purposes, then it would make more sense to use an average value.

watertap/unit_models/tests/test_cstr.py
MarcusHolly: Perhaps I'll try using the median value for the sizing cost above and see if this resolves the issue (although I'm not convinced it will).

MarcusHolly: It was actually just a unit/scaling problem. Once I specified that the inlet flowrate was in m3/s and corrected the LCOW value, the Mac failure resolved itself. I didn't change the unit cost value though... still not sure if using the average/median makes sense.

watertap/costing/unit_models/cstr.py
adam-a-a: Fair enough!

watertap/unit_models/tests/test_cstr.py
adam-a-a: What was the unit/scaling problem?

adam-a-a (approved):
LGTM!

watertap/unit_models/tests/test_cstr.py
MarcusHolly: I needed to specify the units of volumetric flow when I set the operating conditions in the test file

agarciadiego (approved):
LGTM

PR #1198, opened 2023-11-13
Add clarifier model and costing

luohezhiming:
## Summary/Motivation:
Add a simple clarifier model based on IDAES separator and costing method

-  add a clarifier model
- add costing method for circular, rectangular and primary types
- add relevant tests

### Legal Acknowledgement


watertap/costing/unit_models/clarifier.py
```diff
@@ -0,0 +1,316 @@
...
+        units=pyo.units.USD_2011 / pyo.units.ft**4,
+    )
+
+    blk.concstruction_b_parameter = pyo.Param(
+        initialize=98.952,
+        doc="B parameter for construction cost",
+        units=pyo.units.USD_2011 / pyo.units.ft**2,
+    )
+
+    blk.concstruction_c_parameter = pyo.Param(
```
adam-a-a: typo - should be `blk.construction_...`

watertap/costing/unit_models/clarifier.py
```diff
@@ -0,0 +1,316 @@
...
+    elif clarifier_type == clarifier_type.primary:
+        cost_primary_clarifier(blk, **kwargs)
+    else:
+        raise ConfigurationError(
+            f"{blk.unit_model.name} received invalid argument for clarifier_type:"
+            f" {clarifier_type}. Argument must be a member of the ClarifierType Enum."
+        )
+
+
+def build_circular_clarifier_cost_param_block(blk):
```
adam-a-a, line 57: I think it'd be good to note the refs too at this point.

watertap/costing/unit_models/clarifier.py
```diff
@@ -0,0 +1,316 @@
...
+    build_rule=build_primary_clarifier_cost_param_block,
+    parameter_block_name="primary",
+)
+def cost_primary_clarifier(blk, cost_electricity_flow=True):
+    """
+    Primary clarifier costing method
+    """
+    make_capital_cost_var(blk)
+
+    t0 = blk.flowsheet().time.first()
```
adam-a-a, line 194: Should `t0` and `cost_electricity_flow` also be included for the alternative costing methods? Not clear on whether the O&M costs include that already, but maybe you can clarify.

watertap/costing/unit_models/clarifier.py
```diff
@@ -0,0 +1,316 @@
...
+    """
+    Clarifier costing method
+
+    Args:
+        clarifier_type: ClarifierType Enum indicating clarifier type,
+            default = ClarifierType.circular
+
+        # `**kwargs`: Additional keywords for the ClarifierType, e.g., NaOCl
+        #     and CaOH2 mixers expect the `dosing_rate` keyword
+        #     argument.
```
adam-a-a: is this relevant?

watertap/costing/unit_models/clarifier.py
luohezhiming: Primary clarifier costing method is based on the `flow_vol` which has time index, but the other costing methods curently only depends on `surface_area` and does not have time index. And for circular and rectangular clarifier, the electricity has already been included in O&M cost so I don't register the electricity flow.

watertap/costing/unit_models/clarifier.py
luohezhiming: Deleted

watertap/costing/unit_models/clarifier.py
luohezhiming: Good catch, I've revised them

watertap/costing/unit_models/clarifier.py
luohezhiming: Yeah, good suggestion, I add reference at the begining and note them with respect to specific costing method.

watertap/costing/unit_models/clarifier.py
adam-a-a: Do we know what fraction of O&M is attributed to electricity?

luohezhiming: > Do we know what fraction of O&M is attributed to electricity? Yeah, from the paper it's around 3%

Yeah, from the paper it's around 3%

watertap/unit_models/tests/test_clarifier.py
```diff
@@ -0,0 +1,187 @@
...
+        m = clarifier_frame
+        results = solver.solve(m)
+
+        # Check for optimal solution
+        assert_optimal_termination(results)
+
+    @pytest.mark.solver
+    @pytest.mark.skipif(solver is None, reason="Solver not available")
+    @pytest.mark.component
+    def test_circular_clarifier_costing(self, clarifier_frame):
```
adam-a-a, line 125: should we add a test without specifying the costing method to make sure the default costing method works?

luohezhiming: The default is set as circular, I test on my on workstation and it works find, and we can also directly delete the costing_method here

watertap/costing/unit_models/clarifier.py
luohezhiming: I've revised this part for O&M costing, the electricity is registered separately, and the fixed O&M cost is adjusted based on the proportion.

MarcusHolly:
@luohezhiming This PR looks good to me, but Alejandro and I also added our costing into the BSM2 flowsheet, so maybe you should also do that in this PR.

MarcusHolly (approved):
LGTM

agarciadiego (approved):
LGTM

adam-a-a (approved):
LGTM- any other changes can be done in a later PR, as needed.

watertap/costing/unit_models/clarifier.py
```diff
@@ -0,0 +1,361 @@
...
+            + blk.costing_package.circular.construction_c_parameter,
+            to_units=blk.costing_package.base_currency,
+        )
+    )
+
+    max_surface_area_limit = 200 * pyo.units.ft**2
+    if pyo.value(surface_area) > pyo.value(max_surface_area_limit):
+        num_O_and_M = math.ceil(
+            pyo.value(surface_area) / pyo.value(max_surface_area_limit)
+        )
```
adam-a-a: Oh, wait a second- this can be problematic. Surface area is a variable, so when unfixed, this value for surface area will be changing with solver iterations.

adam-a-a (changes requested):
@luohezhiming - sorry- I just noticed a potential issue that would need at least a discussion to make sure I am not misinterpreting the code. At most, we'd need to change said code.

watertap/costing/unit_models/clarifier.py
```diff
@@ -0,0 +1,361 @@
...
+        units=pyo.units.USD_2011 / pyo.units.ft**4 / pyo.units.year,
+    )
+
+    blk.O_and_M_c_parameter = pyo.Param(
+        initialize=1.6945,
+        doc="C parameter for operation and maintenance cost",
+        units=pyo.units.USD_2011 / pyo.units.ft**2 / pyo.units.year,
+    )
+
+    blk.O_and_M_d_parameter = pyo.Param(
```
adam-a-a: THese should be `Vars`, not `Params`. The underlying function (`register_costing_parameter_block`) will fix all `Vars`, if I recall correctly.

watertap/costing/unit_models/clarifier.py
luohezhiming: So, how can I revise this? I think the surface area should be fixed for solving the costing problem, and once it's value is larger than the limit, we have to change to another option for costing this unit

watertap/unit_models/clarifier.py
```diff
@@ -0,0 +1,148 @@
...
+                "outlet_list. This should not "
+                "occur - please use underflow "
+                "and effluent as outlets.".format(self.name)
+            )
+
+        self.surface_area = Var(
+            initialize=1500,
+            doc="Cross section surface area of the clarifier",
+            units=pyunits.m**2,
+            bounds=(0, 3000),
```
adam-a-a, line 78: Where does this upper bound come from? For example, the cost curve that I found for gravity sludge thickeners provides cost related to the diameter of the thickener, and the stated valid range for diameter is 20 - 50 feet. Did your upper bound originate from a similar valid range stated in the literature or did you just assume this?

luohezhiming: This upper bound is also from the same source literature

adam-a-a: Shouldn't this be 200 sq ft, based on what I see in the costing file?

watertap/costing/unit_models/clarifier.py
adam-a-a: Actually, the underlying issue here is that once you go beyond the valid range for surface area, as stated in the literature, the assumption is that you would need to add on another clarifier. Thus, there can be _n_ clarifiers, and one might think that you'd need integer programming to deal with this. 

I am facing a similar issue with the thickener now.

However, I think we have some options. First, if you wanted to have a conditional like this, it'd have to be something like and `Expr_If `or reworked as a `smooth_max `or `smooth_min ` approach, since surface area can't also be guaranteed to be fixed, and more than that, I would argue that surface area shouldn't be fixed. Rather it should be a decision variable.

At this point, I might lean towards holding off on these O&M costs (or leave them with the intention of changing in next PR) since we want to be careful about how we implement usage of _n_ number of units.

The other thing to consider is that we can calculate the effective volume (i.e., the total volume of all clarifier units) based on (1) HRT* Qin and (2) surface_area * effective height, where effective height would account for _n_ number of units implicitly. (Imagine stacking all of our clarifiers on top of each other, each with the same diameter; this would give us the total volume of all reactors and sidestep using _n_.) This is the approach I am considering now. 

As your capital cost calculation stands now, it isn't clear that we'd be computing the total cost for all clarifiers needed, if we limit surface area to that of what one clarifier would typically have.

adam-a-a: On the other hand, there would be no penalty for the optimizer in terms of allowing effective height to go as high as possible to minimize surface area and thus clarifier capital cost. Perhaps this is where O&M costs can come in and serve as that penalty. That is, minimizing surface are would lead to large effective height and thus more individual units, and if O&M costs are driven up by the number of individual units, then that impose some sort of limit on minimizing surface area. Still thinking this through. . .

luohezhiming: Yeah, it's a good point. I think I can take off this fixed O&M cost first for this PR. For circular clarifier, the upper limit is very low (200 ft^2) and only account for O&M cost, but the construction cost relationship is valid for (0 to 32000 ft^2). So I guess the limit for O&M cost does not mean we have multiple clarifiers for this case, just the O&M cost relationship is valid for membrane exchange within 200 ft^2 and we have to do multiple operations.

watertap/unit_models/clarifier.py
luohezhiming: the 200 sq ft is only for O&M cost, the construction cost relationship is valid up to 32000 sq ft

luohezhiming:
@adam-a-a I currently delete all fixed O&M costing, and I create a issue #1215 for refine the clarifier model.

watertap/unit_models/clarifier.py
```diff
@@ -0,0 +1,148 @@
...
+class ActivatedSludgeModelType(Enum):
+    """
+    ASM1: ASM1 model
+    ASM2D: ASM2D model
+    modified_ASM2D: modified ASM2D model for ADM1 compatibility
+    """
+
+    ASM1 = auto()
+    ASM2D = auto()
+    modified_ASM2D = auto()
```
adam-a-a: It's not clear that you need this class here.

adam-a-a (approved):
LGTM - further refinement can come in a subsequent PR.

watertap/costing/unit_models/clarifier.py
```diff
@@ -0,0 +1,219 @@
...
+"""
+[1] Sharma, Jwala R., Mohammad Najafi, and Syed R. Qasim.
+"Preliminary cost estimation models for construction, operation, and maintenance of water treatment plants."
+Journal of Infrastructure Systems 19.4 (2013): 451-464.
+
+[2] Byun, Jaewon, Maravelias, Christos.
+Benchmark Model for Wastewater Treatment Using an Activated Sludge Process.
+United States: N.p., 21 Jan, 2022. Web. doi: 10.7481/1844539.
+"""
+import math
```
bknueven:
```suggestion
```

PR #1207, opened 2023-11-17
OLI API login via access key

veccp:
## Fixes/Resolves:

Simplifies logins to OLI Cloud API

## Summary/Motivation:

Required to improve OLI Cloud API test coverage

- Create/save access keys
- Delete access keys

### Legal Acknowledgement


veccp:
@lbianchi-lbl I think we're ready to look at adding the access key to tests

watertap/tools/oli_api/tests/test_credentials.py
```diff
@@ -8,59 +8,95 @@
...
+    test_contents = listdir(test_dir)
+
+    credential_manager.login()
+    with OLIApi(credential_manager, test=True) as oliapi:
+        local_dbs_file = join(test_dir, "test.dbs")
+        oliapi.get_dbs_file_id(local_dbs_file)
+        yield oliapi
+
+    _cleanup(root_contents, root_dir)
+    _cleanup(test_contents, test_dir)
```
lbianchi-lbl: Since it's not being used in this test file and we have the fixture defined in `conftest.py`, can this be deleted?

bknueven:
It looks like some tests are being skipped still I would not expect:
Screenshot 2023-12-11 at 12 44 39 PM

lbianchi-lbl:
> It looks like some tests are being skipped still I would not expect: Screenshot 2023-12-11 at 12 44 39 PM

The tests that are being skipped are the ones that use the `auth_credentials` fixture, including those tests that use it indirectly through the derived `oliapi_instance` fixture. The fixture is set up to look for credentials in environment variables; if the variables are not found, `pytest.skip()` is called, which will skip any tests that require that fixture without having to implement the skip logic in the test itself. If all of the `OLI_API_XYZ` variables are found in the environment, the tests are not skipped:

https://github.com/watertap-org/watertap/blob/e4d53d8e5d3ca68adfb6b87474dc634fc2c38949/watertap/tools/oli_api/conftest.py#L69-L74

In summary: I'm not sure whether the above fully addresses your comment, but the fact that those tests are being skipped here (i.e. in the `watertap-org/watertap` CI) is by design and expected.

watertap/tools/oli_api/tests/test_credentials.py
veccp: Yes, will push my changes to the test.credentials file shortly

bknueven:
> In summary: I'm not sure whether the above fully addresses your comment, but the fact that those tests are being skipped here (i.e. in the `watertap-org/watertap` CI) is by design and expected.

Right, of course.

lbianchi-lbl (approved):
The missing coverage is expected, with progress to address this being tracked in #1249.

bknueven (approved):

PR #1209, opened 2023-11-20
Separating GAC costing parameters based on type

bknueven:
## Summary/Motivation:
@hunterbarber points out in #1183 some trouble with the way parameters are set up for the GAC costing parameter block. This PR would rectify this issue by separating out the flowsheet-level costing attribute by GAC costing type.

As-is, the unit model costing block builder *updates* the global GAC parameters. This creates a situation where whatever `ContractorType` was specified *last* will be the one in effect for all GAC models on the flowsheet. This could lead to hard-to-debug modeling errors

- Create two separate global-level GAC costing parameter blocks: `gac_pressure` and `gac_gravity`
- Update the GAC tests to reflect this change.

### Legal Acknowledgement


hunterbarber:
This is a better implementation than my temporary fix IMO, I'll check further once testing is done.

adam-a-a (approved):

hunterbarber (approved):

bknueven:
@adam-a-a I like your design on #1212 better so I am going to copy it here.

bknueven:
055c5d2 adds another function to the call stack, but with the benefit of (1) you only get the costing parameter block you need (not both) and (2) you can call `cost_gac_gravity` or `cost_gac_pressure` directly instead of using the `ContactorType`.

PR #1214, opened 2023-11-21
OARO SKK config option and tests

zacharybinger:
## Fixes/Resolves:

Added the option for using the Spieler-Kedem-Katchalsky (SKK) model in OARO models. 

## Summary/Motivation:

The SKK model should more appropriately reflect the water and salt transport across "leaky" membranes which should also be applicable to OARO operations.

- Modified OARO base class to include a config option for the SKK model transport model
- Modified unit tests for OARO-1D and OARO-0D to account for the addition of the SKK model

### Legal Acknowledgement


adam-a-a (approved):
LGTM - can you make an issue to note the addition of SKK in documentation for OARO? Don't remember if we have that already

watertap/unit_models/tests/test_osmotically_assisted_reverse_osmosis_0D.py
```diff
@@ -465,6 +484,237 @@ def test_solution(self, RO_frame):
...
+
+        m.fs.unit = OsmoticallyAssistedReverseOsmosis0D(
+            property_package=m.fs.properties,
+            has_pressure_change=True,
+            concentration_polarization_type=ConcentrationPolarizationType.fixed,
+            mass_transfer_coefficient=MassTransferCoefficient.none,
+            transport_model=TransportModel.SKK,
+            has_full_reporting=True,
+        )
+
```
luohezhiming, line 502: I think it will make more sense if we have the tets for solving the system with `concentration_polarization_type` and `mass_transfer_coefficient` set as `calculated` because it is the most general case.

watertap/unit_models/tests/test_osmotically_assisted_reverse_osmosis_1D.py
```diff
@@ -474,7 +489,258 @@ def test_solution(self, RO_frame):
             m.fs.unit.permeate_side.deltaP_stage[0]
         )
 
-    @pytest.mark.requires_idaes_solver
+    # NOTE Begin SKK tests
```
luohezhiming: Similarly as 0D model, I think maybe we can add a test for more general case with `concentration_polarization_type` and  `mass_transfer_coefficient` set as `cacluated`

luohezhiming (commented):
Great work. My only comment is to consider adding a test for more general case with the config set as `calcuated`, just to verify we can solve these cases.

luohezhiming (approved):
LGTM

PR #1224, opened 2023-11-30
Fixes for Pyomo 6.7

bknueven:
## Fixes/Resolves:
Make WaterTAP compatible with Pyomo 6.7.0.

## Summary/Motivation:
Keeping up with upstream dependencies.

- Revert requirement `pyomo<6.7`. 832b2eb
- Revert #1146 (NL interface changed in Pyomo/pyomo#3037). 8e1084b
- Change bounds on B_comp for NF0D. This variable was sometimes fixed out-of-bounds which is not compatible with Pyomo 6.7. 14f38d8
- Remove scaling of Expression in cryst_prop_pack (not compatible with Pyomo 6.7). 9496411
- Update small regression values in full treatment train chemical post-treatment and loosen test requirement. Remove IDAES solver requirement from these tests. f04426d
- Update regression values in ASM2D_noPHA and remove custom `bound_push`. Remove IDAES solver requirement for this test. c9d4c98
- Update regression value in OARO1D; remove IDAES solver requirement for these tests. 7b66e15
- Remove `mutable=False` from Params with units. These were always mutable and now Pyomo raises a warning (Pyomo/pyomo#3004). 3d83b60
- New implementation of `scaling_factor` suffix warning suppression from Pyomo's NL writer. 2c73bfa

### Legal Acknowledgement

### DIFF
--- a/watertap/core/plugins/solvers.py
+++ b/watertap/core/plugins/solvers.py
@@ -14,7 +14,6 @@ import pyomo.environ as pyo
 from pyomo.core.base.block import _BlockData
 from pyomo.core.kernel.block import IBlock
 from pyomo.solvers.plugins.solvers.IPOPT import IPOPT
-import pyomo.repn.plugins.nl_writer as _nl_writer
 
 import idaes.core.util.scaling as iscale
 from idaes.core.util.scaling import (
@@ -25,17 +24,6 @@ from idaes.core.util.scaling import (
 from idaes.logger import getLogger
 
 _log = getLogger("watertap.core")
-_SuffixData = _nl_writer._SuffixData
-
-
-class _WTSuffixData(_SuffixData):
-    def update(self, suffix):
-        self.datatype.add(suffix.datatype)
-        for obj, val in suffix.items():
-            self._store(obj, val)
-
-    def store(self, obj, val):
-        self._store(obj, val)
 
 
 @pyo.SolverFactory.register(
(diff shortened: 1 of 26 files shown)


bknueven:
@agarciadiego @adam-a-a could you look at the changes to the ASM2D noPHA flowsheet in c9d4c98?

I don't particularly care for how the tests are implemented. I think we should be checking small values with an absolute tolerance against 0.

bknueven:
@lbianchi-lbl could you have a look at how I suppress a log message we don't want to see in 2c73bfa?

I've not used this feature of `logging` before but it seems to do what we want, and more robustly than #1146.

watertap/core/plugins/solvers.py
```diff
@@ -25,17 +26,15 @@
...
-        for obj, val in suffix.items():
-            self._store(obj, val)
 
-    def store(self, obj, val):
-        self._store(obj, val)
+def _pyomo_nl_writer_logger_filter(record):
+    msg = record.getMessage()
+    if "scaling_factor" in msg and "model contains export suffix" in msg:
+        return False
+    return True
```
lbianchi-lbl, line 37: I think using `Logger.addFilter()`/`removeFilter()` is a great way to get rid of these warnings.

lbianchi-lbl (approved):
This looks great! I like the solution to get rid of the `scaling_factors` messages in the logs.

At some point I'd like to try to see if we can find a way to make all the global solver configuration a bit more explicit and less side-effect-y, but that's definitely not for this PR.

agarciadiego (approved):
LGTM

PR #1235, opened 2023-12-05
Resolve neutral charge on MCAS

### LINKED ISSUE #1200: Come to resolution for 0 charge on MCAS
https://github.com/watertap-org/watertap/pull/1176/files#diff-4b5fe003ad0128ad3cc082b8d7c4d64296d8d634b04dcc00538592a59628866eR389-R391

Currently a warning is always logged if no charge data are provided, as a safeguard. However, if 0 charge is provided, an error is raised. A change should be made so that we (1) safeguard the novice user and also (2) don't annoy advanced users with warnings that are unnecessary for their purposes.

EDIT: deleting the logger warning is trivial. The true underlying issue at hand is that we need a way to tell the user that charge data need to be supplied for ions if they didn't provide the data. Seems obvious, but users acclimating to WaterTAP might not be aware of how to do so or that they have to do so. A user may also forget to do so.

adam-a-a:
## Summary/Motivation:
This PR enforces the provision of charge values for all solutes, including charge = 0 for neutral solutes. However, setting config option `ignore_neutral_charge=True` allows user to bypass providing charge=0 for neutral solutes. 

In doing so, the new user is protected from erroneous results due to not providing appropriate inputs, while the advanced user can bypass this necessity when desired with minimal additional effort.

- raise configerror if charge not provided for all solutes by default
- add config option to ignore neutral solutes in terms of providing charge data upfront
-

### Legal Acknowledgement


MarcusHolly (approved):
LGTM - just wanted to delay my approval until I saw that all the tests passed.

hunterbarber (approved):
Implementation, testing, and docs seem good to me

PR #1237, opened 2023-12-05
Revision of OLI Cloud Flash Calculations

veccp:
## Fixes/Resolves: 

Provides a framework for adding additional OLI Cloud flash calculations as the need emerges. Currently supporting Water Analysis and Isothermal Analysis.

## Summary/Motivation:

Prior analyses used bespoke scripts to make flash calculations and extract data; this approach simplifies the process for the user

- Standardized setup for flash calculation calls
- Extracting output data is simplified
- Added flexibility to parametric sweeps/surveys

### Legal Acknowledgement


bknueven:
@lbianchi-lbl this PR includes a Jupyter Notebook demonstrating the OLI workflow. But as this requires logging in to OLI it currently fails.

I suspect we should set up the notebook test harness to skip this notebook for now. Is there an easy way to achieve that?

watertap/tools/oli_api/client.py
```diff
@@ -267,7 +266,8 @@ def get_flash_history(self, dbs_file_id, job_id=False):
         :param dbs_file_id: string ID of DBS file
         :param job_id: bool flag to return job ID for troubleshooting
 
-        :return flash_history: dictionary containing array of submitted jobs OR just job ID if specified
+        :return flash_history: dictionary containing array of submitted jobs OR
+        just job ID if specified
```
bknueven:
```suggestion
        :return flash_history: dictionary containing array of submitted jobs OR
            just job ID if specified
```
I think this change will resolve the RTD error, per https://stackoverflow.com/a/74798788

setup.py
```diff
@@ -126,6 +126,7 @@
             "pytest-cov",  # code coverage
             "mongomock",  # mongodb mocking for testing
             "nbmake",
+            "cryptography",  # for encrypting OLI credentials
```
bknueven:
```suggestion
```

watertap/tools/oli_api/client.py
```diff
@@ -327,49 +328,44 @@ def _delete_dbs_file(self, dbs_file_id):
 
     def call(
         self,
-        function_name,
-        dbs_file_id,
-        json_input=dict(),
+        flash_method="",
+        dbs_file_id="",
+        input_params=dict(),
```
bknueven: These should all default to `None`.

watertap/tools/oli_api/flash.py
```diff
@@ -0,0 +1,425 @@
...
+        :param get_oli_names: boolean switch to convert name into OLI form
+        :param tee: boolean switch to show printable output
+
+        :return survey: DataFrame containing surveys (or empty)
+        """
+
+        exclude_items = ["Temperature", "Pressure"]
+        convert_name = (
+            lambda k: get_oli_name(k) if get_oli_names and k not in exclude_items else k
+        )
```
bknueven:
```suggestion
        if get_oli_names:
            convert_name = lambda k: k if k in exclude_items else get_oli_name(k)
        else:
            convert_name = lambda k: k
```

We should consider if `exclude_items` should be part of the `get_oli_name` function instead.

watertap/tools/oli_api/flash.py
```diff
@@ -0,0 +1,425 @@
...
+        survey_vars = {convert_name(k): v for k, v in survey_arrays.items()}
+        survey_vars_product = list(product(*(survey_vars[key] for key in survey_vars)))
```
bknueven:
```suggestion
        survey_vars_product = product(*survey_vars.values())
```

watertap/tools/oli_api/flash.py
```diff
@@ -0,0 +1,425 @@
...
+        survey = DataFrame(data=survey_vars_product, columns=survey_vars.keys())
+        if tee:
+            print(f"Survey contains {len(survey)} items.")
```
bknueven: We should just using `logging`.

watertap/tools/oli_api/flash.py
```diff
@@ -0,0 +1,425 @@
...
+                    "value": value,
+                    "charge": charge,
+                }
+            )
+
+        # build entries for other specified properties
+        for k, v in self.water_analysis_properties.items():
+            if isinstance(v["value"], list):
+                self.water_analysis_properties[k]["value"] = v["value"][0]
+            if k not in [i["name"] for i in self.water_analysis_input_list]:
```
bknueven:
```suggestion
            if all(k != i["name"] for i in self.water_analysis_input_list):
```

But maybe we should rethink this a bit to avoid quadratic behavior.

watertap/tools/oli_api/flash.py
```diff
@@ -0,0 +1,425 @@
...
+                return
+        else:
+            if not self.optional_properties["prescalingTendenciesRigorous"] == True:
+                new_values = {
+                    k: not v
+                    for k, v in self.optional_properties.items()
+                    if "prescaling" in k
+                }
+            else:
+                return
```
bknueven:
```suggestion
        # if both bools agree, do nothing
        if bool(use_scaling_rigorous) == bool(self.optional_properties["prescalingTendenciesRigorous"]):
            return
       # if both bools disagree, update
       new_values = {
            k: not v
            for k, v in self.optional_properties.items()
            if "prescaling" in k
       }
```

Should we be flipping the `prescalingTendenciesRigorous` flag if it disagrees with `use_scaling_rigorous`?

watertap/tools/oli_api/flash.py
```diff
@@ -0,0 +1,425 @@
...
+            )
+            result = {
+                k: oliapi_instance.call(flash_method, dbs_file_id, v)
+                for k, v in clones.items()
+            }
+            suffix = "composition_survey"
+
+        else:
+            result = {0: oliapi_instance.call(flash_method, dbs_file_id, initial_input)}
+            suffix = "single_point"
```
bknueven:
```suggestion
        if survey is None:
            result = {0: oliapi_instance.call(flash_method, dbs_file_id, initial_input)}
            suffix = "single_point"
            
        else:
            clones = self.modify_inputs(
                initial_flash_input=initial_input,
                survey=survey,
                flash_method=flash_method,
            )
            result = {
                k: oliapi_instance.call(flash_method, dbs_file_id, v)
                for k, v in clones.items()
            }
            suffix = "composition_survey"
```

Nit (optional)

watertap/tools/oli_api/flash.py
```diff
@@ -0,0 +1,425 @@
...
+            file_name = f"{flash_method}_{suffix}"
+            self.write_output_to_yaml(result, file_name)
+        return result
+
+    def modify_inputs(self, initial_flash_input, survey, flash_method):
+        """
+        Iterates over a survey to create modified clones of an initial flash analysis output.
+
+        :param initial_flash_input: flash analysis input to copy
+        :param survey: DataFrame containing modifications for each test
```
bknueven: Maybe `survey` should be a `dict`?

watertap/tools/oli_api/flash.py
```diff
@@ -0,0 +1,425 @@
...
+            Path(__file__).parents[0],
+            f"{t.day:02}{t.month:02}{t.year:04}_{file_name}.yaml",
+        )
+
+        if tee:
+            print(f"Saving file...")
+        with open(file_path, "w") as yamlfile:
+            yaml.dump(flash_output, yamlfile)
+        if tee:
+            print(f"{file_path} saved to working directory.")
```
bknueven: logging

watertap/tools/oli_api/flash.py
```diff
@@ -0,0 +1,425 @@
...
+
+        :param flash_output: dictionary output from OLI API call
+        :param filename: string name of file to write
+
+        :return file_path: string name of file written
+        """
+
+        t = datetime.utcnow()
+        file_path = join(
+            Path(__file__).parents[0],
```
bknueven:
```suggestion
            Path(__file__).parent,
```

watertap/tools/oli_api/flash.py
```diff
@@ -0,0 +1,425 @@
...
+                )
+            clones[clone_index] = modified_clone
+        return clones
+
+    def write_output_to_yaml(self, flash_output, file_name, tee=True):
+        """
+        Writes OLI API flash output to .yaml file.
+
+        :param flash_output: dictionary output from OLI API call
+        :param filename: string name of file to write
```
bknueven, line 397: Should document that we will write a different file, or just trust them

watertap/tools/oli_api/flash.py
```diff
@@ -0,0 +1,425 @@
...
+
+        t = datetime.utcnow()
+        file_path = join(
+            Path(__file__).parents[0],
+            f"{t.day:02}{t.month:02}{t.year:04}_{file_name}.yaml",
+        )
+
+        if tee:
+            print(f"Saving file...")
+        with open(file_path, "w") as yamlfile:
```
bknueven: Alternatively, we could raise an error if the file exists -- and have a flag to ignore that error. This would be my preference.

watertap/tools/oli_api/flash.py
```diff
@@ -0,0 +1,425 @@
...
+        :param write: boolean switch to write extracted properties to yaml
+
+        :return extracted_properties: dictionary containing values for specified properties
+        """
+
+        def _filter_zeroes(data, filter_zero):
+            if "values" in data:
+                if filter_zero:
+                    data["values"] = {k: v for k, v in data["values"].items() if v != 0}
+            return data
```
bknueven:
```suggestion
        if filter_zero:
            def _filter_zeroes(data):
                if "values" in data:
                    data["values"] = {k: v for k, v in data["values"].items() if v != 0}
                return data
        else:
            _filter_zeroes = lambda data: data
```

watertap/tools/oli_api/flash.py
```diff
@@ -0,0 +1,425 @@
...
+                    phases.append("total")
+                options = self.stream_output_options
+                prop_groups = [k for k in options if prop in options[k]]
+                for group in prop_groups:
+                    if group in ["result", "properties"]:
+                        for phase in phases:
+                            deep_path = deepcopy(root_path)
+                            deep_path.extend(_get_nested_phase_keys(phase, group))
+                            result = _filter_zeroes(
+                                _get_nested_data(raw_result, deep_path), filter_zero
```
bknueven:
```suggestion
                                _get_nested_data(raw_result, deep_path)
```

watertap/tools/oli_api/flash.py
```diff
@@ -0,0 +1,425 @@
...
+                            )
+                            extracted_properties[i][f"{prop}_{phase}"] = result
+                    if group in ["additionalProperties", "waterAnalysisOutput"]:
+                        deep_path = deepcopy(root_path)
+                        deep_path.extend([group, prop])
+                        result = _filter_zeroes(
+                            _get_nested_data(raw_result, deep_path), filter_zero
```
bknueven:
```suggestion
                            _get_nested_data(raw_result, deep_path)
```

bknueven:
```suggestion
        extracted_properties = {}
        property_groups = { p : [] for p in properties }
        groups_with_phase = set(("result", "properties"))
        groups_additional = set(("additionalProperties", "waterAnalysisOutput"))
        # TODO: make exception
        assert set(self.stream_output_options.keys()) == groups_with_phase + groups_additional
        for group, props in self.stream_output_options.items():
            for p in props:
                if p in properties:
                    property_groups[p].append(group)
        for i in raw_result:
            extracted_properties[i] = {}
            root_path = [i, "result"]
            base_result = _get_nested_data(raw_result, root_path)
            phases = {p : [] for p in properties}
            for phase, prop in base_result["phases"].items():
                if prop in properties:
                    phases[prop].append(phase)
            for prop in properties:
                if prop in base_result["total"]:
                    phases[prop].append("total")
                for group in property_groups[prop]:
                    if group in groups_with_phase:
                        for phase in phases[prop]:
                            deep_path = deepcopy(root_path)
                            deep_path.extend(_get_nested_phase_keys(phase, group))
                            result = _filter_zeroes(
                                _get_nested_data(raw_result, deep_path)
                            )
                            extracted_properties[i][f"{prop}_{phase}"] = result
                    if group in groups_additional:
                        deep_path = deepcopy(root_path)
                        deep_path.extend([group, prop])
                        result = _filter_zeroes(
                            _get_nested_data(raw_result, deep_path)
```

Don't need to take all of this --

bknueven (commented):
Some suggestions -- looks pretty good.

lbianchi-lbl:
> @lbianchi-lbl this PR includes a Jupyter Notebook demonstrating the OLI workflow. But as this requires logging in to OLI it currently fails.
> 
> I suspect we should set up the notebook test harness to skip this notebook for now. Is there an easy way to achieve that?

At the moment, unfortunately, the notebook testing infrastructure doesn't offer a good way to do that beyond deselecting the notebook from the test run at the pytest invocation level. I think adding the flag `-k "not incorporating_oli_calculations.ipynb"` in the last step of the `notebooks` CI job should work: https://github.com/watertap-org/watertap/blob/a88cf2f4e7edb9c5e048ce9d096da5ebb2140937/.github/workflows/checks.yml#L242

watertap/tools/oli_api/flash.py
veccp: I think that makes sense

watertap/tools/oli_api/flash.py
veccp: Yes, there's 4 keys with prescaling (2 for rigorous and 2 for estimated). If booleans disagree all four will be flipped

watertap/tools/oli_api/flash.py
veccp: Will trust the user

watertap/tools/oli_api/flash.py
```diff
@@ -0,0 +1,436 @@
...
+        :param flash_output: dictionary output from OLI API call
+        :param filename: string name of file to write
+        :param logging: boolean switch to show printable output
+
+        :return f: string name of file written
+        """
+
+        f = f"{file_name}.yaml"
+        if logging:
+            print(f"Saving file...")
```
bknueven: We should use the `logging` module here. See, e.g., https://github.com/watertap-org/watertap/blob/dcb5c28e707980e03783db336f4efdf341c949f0/watertap/tools/parallel/multiprocessing_parallel_manager.py#L13, https://github.com/watertap-org/watertap/blob/dcb5c28e707980e03783db336f4efdf341c949f0/watertap/tools/parallel/multiprocessing_parallel_manager.py#L26, and https://github.com/watertap-org/watertap/blob/dcb5c28e707980e03783db336f4efdf341c949f0/watertap/tools/parallel/multiprocessing_parallel_manager.py#L141

bknueven:
```suggestion
        _logger.debug(f"Saving to file {f}")
```

watertap/tools/oli_api/flash.py
```diff
@@ -0,0 +1,436 @@
...
+                        path = survey[i][param]
+
+            else:
+                raise IOError(
+                    " Only 'wateranalysis' and 'isothermal' currently supported."
+                )
+            clones[i] = modified_clone
+        return clones
+
+    def write_output_to_yaml(self, flash_output, file_name, logging=True):
```
bknueven:
```suggestion
    def write_output_to_yaml(self, flash_output, file_name):
```

bknueven:
We need to make sure the flash calculations tests work post-PR #1207.

watertap/tools/oli_api/util/fixed_keys_dict.py
```diff
@@ -44,46 +46,193 @@ def pprint(self):
...
+            "pyomo_unit": pyunits.Pa * pyunits.second,
+        },
+        "alkalinity": {"oli_unit": "mg HCO3/L", "pyomo_unit": pyunits.mg / pyunits.L},
+        "TIC": {"oli_unit": "mol C/L", "pyomo_unit": pyunits.mol / pyunits.L},
+        "CO2GasFraction": {
+            "oli_unit": "mol/mol",
+            "pyomo_unit": pyunits.mol / pyunits.mol,
+        },
+    }
+)
```
OOAmusat, line 92: Are we hard-coding the units?  If we are, then we need to rethink whether we want to do this. If we're not, is there an example where we show hoe to use the functionality with a different set of units?

watertap/tools/oli_api/core/water_analysis.py
OOAmusat, line 1: What happened to the water analysis function? Did we move it elsewhere?

bknueven:
The idea is this defines the conversion between OLI units and the
equivalent Pyomo/pint unit.

Pyomo’s existing units conversion framework can be used to convert to/from
the Pyomo unit listed here.

watertap/tools/oli_api/util/fixed_keys_dict.py
bknueven: The idea is this defines the conversion between OLI units and the equivalent Pyomo/pint unit. Pyomo’s existing units conversion framework can be used to convert to/from the Pyomo unit listed here.

watertap/tools/oli_api/core/water_analysis.py
bknueven: It is subsumed into the new `Flash` class.

watertap/tools/oli_api/util/fixed_keys_dict.py
OOAmusat: So what happens if the user is not using Pyomo? For example, if I want to run a flash calculation by passing my ranges and info directly to OLI? How are units handled then? Or are we not handling this use case for now?

OOAmusat (commented):
This looks good to me, and I'd be happy to approve it once the Jupyer tests pass. 

Two things:

1. We're still missing the parallelization bit. That is critical, especially given the fact that we're going to be testing OLI's burst capabilities soon. Maybe we address that in a different PR though.
2. This framework seems more geared to starting from a WT (or Pyomo) flowsheet. I'm a bit unclear as to how easy it would be to use as a standalone utility. That is something we definitely want to allow for.

watertap/tools/oli_api/core/water_analysis.py
OOAmusat: Yeah, I figured as much.

watertap/tools/oli_api/util/fixed_keys_dict.py
bknueven: Well, if we really are writing a generic Python wrapper for the OLI API (I wasn't aware that was our task), then we should probably replace the Pyomo units here with their `pint` counterparts.

bknueven:
> This looks good to me, and I'd be happy to approve it once the Jupyer tests pass. 
> 
> 
> 
> Two things:
> 
> 
> 
> 1. We're still missing the parallelization bit. That is critical, especially given the fact that we're going to be testing OLI's burst capabilities soon. Maybe we address that in a different PR though.
> 
> 2. This framework seems more geared to starting from a WT (or Pyomo) flowsheet. I'm a bit unclear as to how easy it would be to use as a standalone utility. That is something we definitely want to allow for.

On (1), I was thinking we could do that once this is merged and everyone is happy with the serial API. It's relatively straightforward, and this PR is sufficiently complicated as it is.

watertap/tools/oli_api/util/fixed_keys_dict.py
OOAmusat: This is fine as-is. Just wanted to clarify.

I ask because for our previous work with OLI that required a lot of simulations, we did everything with generic python, and just brought the surrogates or data back to WT. We just found that more efficient. That is not everyone's use case though, and anyone that wants to do that (READ: Mayo) can write their own custom scripts for that.

Let's just get this in.

watertap/tools/oli_api/credentials.py
```diff
@@ -237,19 +237,22 @@ def set_access_key(self):
             return self.credentials["access_keys"][r]
 
     # TODO: improve header management for class
-    def generate_oliapi_access_key(self):
+    # TODO: possibly save api keys as objects with expiry and other attributes
+    def generate_oliapi_access_key(self, expiry_timecode):
```
bknueven: Either here or further up in the API we should consider automatically converting `datetime.datetime` objects to the appropriate `expiry_timecode`, with a sane default base on the current time (or maybe leaving it off and defaulting this to `None`).

watertap/tools/oli_api/credentials.py
```diff
@@ -238,21 +239,39 @@ def set_access_key(self):
...
+        def _set_expiry_timestamp(key_lifetime):
+            """
+            Set expiry date for OLI Cloud access key.
+
+            :param key_lifetime: integer number of days key will be valid (up to 365)
+
+            :return unix_timestamp_ms: unix timestamp (in ms) for when key will expire
+            """
+
+            if key_lifetime < 1 or key_lifetime > 365:
```
adam-a-a, line 298: Should you also check `if key_lifetime is None`? I am not sure the check for <1 or >365 will work if the arg is None

veccp: I think I'll change the default to `365` instead of `None`

watertap/tools/oli_api/client.py
```diff
@@ -44,31 +44,50 @@
...
+        self,
+        credential_manager,
+        interactive_mode=True,
+    ):
         """
         Constructs all necessary attributes for OLIApi class
 
         :param credential_manager_class: class used to manage credentials
-        :param test: bool switch for automation during tests
+        :param interactive_mode: bool to switch level of logging display from info to debug only
```
adam-a-a, line 82: `outlvl` would be more consistent with what we use to set logging in WaterTAP/IDAES, but not a big deal to have this deviation.

watertap/tools/oli_api/client.py
```diff
@@ -123,29 +146,28 @@ def _upload_dbs_file(self, dbs_file_path):
...
     def _create_dbs_dict(self, chemistry_source, phases, model_name):
-
         """
         Creates dict for chemistry-builder to later generate a DBS file id.
 
-        :param chemistry_source: OLI-compatible ion names as a Pyomo Set, list, or dict,
-            where the keys are ion names
-        :param model_name: Name for model as a string
-        :param phases: OLI-compatible phases
+        :param chemistry_source: path (str), dict, or state block containing chemistry info
```
adam-a-a, line 168: Are you only updating doc strings here but preserving the functionality? That is, can the input still be the same as before (i.e., provide OLI ion names as set, list, or dict), and is the only change here in the doc string? That's what it looks like. If that is the case, then IMHO, the new doc string is less clear about that than before.

watertap/tools/oli_api/client.py
```diff
@@ -321,122 +357,114 @@ def _delete_dbs_file(self, dbs_file_id):
...
+            if "data" in result:
+                if "status" in result["data"]:
+                    if "resultsLink" in result["data"]:
+                        _logger.debug(
+                            f"Call status is {result['data']['status']},"
+                            + f" result link is {result['data']['resultsLink']}"
+                        )
+                        result_link = result["data"]["resultsLink"]
+        else:
+            _logger.debug("No result link found")
```
adam-a-a, line 469: So, this message will only be displayed in DEBUG mode, but an exception will always be raised nevertheless, correct (e.g., line 420)?

watertap/tools/oli_api/conftest.py
```diff
@@ -88,9 +91,37 @@ def oliapi_instance(
         "config_file": cred_file_path,
     }
     credential_manager = CredentialManager(**credentials, test=True)
-    credential_manager.login()
```
adam-a-a, line 91: Can you clarify why this isn't needed anymore (or why it was there before and is now being removed)?

watertap/tools/oli_api/conftest.py
```diff
@@ -88,9 +91,37 @@ def oliapi_instance(
...
+            "K_+": 9,
+            "HCO3_-": 385,
+            "SiO2": 30,
+        },
+        "units": {
+            "temperature": pyunits.K,
+            "pressure": pyunits.Pa,
+            "components": pyunits.mg / pyunits.L,
+        },
+    }
```
adam-a-a, line 127: Not sure what this is being used for.

watertap/tools/oli_api/client.py
veccp: `interactive_mode` also affects a few other lines (such as write permission) so I just left it as is for now

watertap/tools/oli_api/client.py
veccp: Here, `chemistry_source` can still be an iterable with oli names as keys. However, it can also be a string pointing to a local dbs file to upload. Once the PR is resolved I'll prioritize updating the doc strings for clarity/consistency

watertap/tools/oli_api/credentials.py
```diff
@@ -191,164 +267,138 @@ def set_access_key(self):
...
-            return self.credentials["access_keys"][int(input(" "))]
-
-    # TODO: improve header management for class
-    def generate_oliapi_access_key(self):
+                _logger.info(f"{i}\t{self.credentials['access_keys'][i]}")
+                if self.test:
+                    r = 0
+                else:
+                    r = int(input(" "))
+            return self.credentials["access_keys"][r]
```
adam-a-a, line 277: Can you explain this bit? Not sure what `self.test` is (or rather why it's called `test`), but my guess is this bit of code just allows the user to select the ith entry from a list of access keys.

watertap/tools/oli_api/conftest.py
veccp: Essentially, `login` is now executed in `__init__` rather than by the user. It's needed to set `refresh_token` and `jwt_token` which are required when using username/password to login

watertap/tools/oli_api/conftest.py
veccp: This input should be used to generate a test dbs file and provide state variable info for flash calculations. It's supposed to represent what would be extracted from a stateblock

watertap/tools/oli_api/credentials.py
```diff
@@ -191,164 +267,138 @@ def set_access_key(self):
...
         return response.text
 
-    def login(self):
+    def login(self, refresh=False):
         """
-        Login into user credentials for the OLI Cloud.
+        Log in to OLI Cloud using access key or credentials.
 
-        :return boolean: True on success, False on failure
+        :return status: bool indicating success or failure
```
adam-a-a, line 342: Should "refresh" be in the doc string (even though it is self-explanatory)?

watertap/tools/oli_api/credentials.py
veccp: True - `self.test` is there to bypass `input()` lines so that tests are autonomous. It'll be replaced with `interactive_mode`, i.e., where `False` skips all the input lines and shows debug messages as well

watertap/tools/oli_api/flash.py
```diff
@@ -0,0 +1,494 @@
...
+        :param water_analysis_output: dictionary to extract inflows from (required if not wateranalysis flash)
+        :param inflows_phase: string indicating desired phase of inflows
+        :param use_scaling_rigorous: boolean switch to use estimated or rigorous solving for prescaling metrics
+
+        :return inputs: dictionary containing inputs for specified OLI flash analysis
+        """
+
+        inputs = {"params": {}}
+
+        if flash_method == "wateranalysis":
```
adam-a-a, line 250: OLI presents "wateranalysis" as a separate function from "isothermal flash", but it seems that you're combining the two into a single "Flash" class. Am I interpreting that correctly? Are there reasons for doing it that way rather than maintaining two separate classes to emulate how OLI presents these capabilities in their API documentation?

watertap/tools/oli_api/credentials.py
veccp: Yes, I just missed it

watertap/tools/oli_api/flash.py
```diff
@@ -0,0 +1,494 @@
...
+                            modified_clone["params"]["inflows"]["values"][
+                                param
+                            ] += survey[i][param]
+                        else:
+                            modified_clone["params"]["inflows"]["values"][
+                                param
+                            ] = survey[i][param]
+                    else:
+                        raise ValueError(
+                            "Only composition and temperature/pressure surveys are currently supported."
```
adam-a-a, line 383: Any reason why a pH survey isn't supported here? We can do this betting setting reconciliation option on for pH calculation, from what I recall.

lbianchi-lbl:
The low reported patch coverage is expected: see https://github.com/watertap-org/watertap/issues/1249 for details.

watertap/tools/oli_api/flash.py
veccp: Wateranalysis and Isothermal flash (and other flash calcs, including corrosion analysis) use the same url structure and headers. They both use the `call` function from the OLIApi class. The only difference is the input structure. The corrosion functions would also go in the Flash class

watertap/tools/oli_api/flash.py
```diff
@@ -0,0 +1,494 @@
...
+handler.setFormatter(formatter)
+_logger.addHandler(handler)
+_logger.setLevel(logging.DEBUG)
+
+# TODO: consider config for file_name for each writing method
+
+
+class Flash:
+    """
+    A class to execute OLI Cloud flash calculations, replacing and augmenting WaterAnalysis class functionality.
```
adam-a-a, line 83: Should we revise this to clarify that this class is meant to represent "Main Methods" from OLI Engine API? Probably don't need to mention "replacing and augmenting WaterAnalysis class functionality" here either. In one of my other comments, I asked why you packed water analysis inside of Flash, thinking that "Flash" represented the isothermal flash function from OLI, but now I realize you are using "Flash" to contain any of the "main methods" available via OLI (right?).

watertap/tools/oli_api/flash.py
veccp: I figured a pH sweep could be done by creating a survey for H+ or OH- for now. The alternative reconciliation options are in the code but they haven't been tested, so I didn't include advanced functionality for them

watertap/tools/oli_api/flash.py
veccp: Yes, the Flash class will include all the main methods from the API. Also - I have reworked each of these 3 classes (OLIApi, CredentialManager, and Flash) to remove a lot of the inconsistencies you noted. Those are relatively small changes compared to the major restructuring involved in this PR

watertap/tools/oli_api/flash.py
```diff
@@ -0,0 +1,494 @@
...
+
+    :param water_analysis_properties: dictionary containing pre-built water analysis input template blocks
+    :param optional_properties: dictionary containing pre-configured optional properties to attach to OLI calls (defaults to True for all properties)
```
adam-a-a, line 86: Interesting - is there a quick way to toggle all to False as well?

veccp: I don't think I built that in, but a dict comprehension would probably work

watertap/tools/oli_api/flash.py
```diff
@@ -0,0 +1,494 @@
...
+    :param input_unit_set: dictionary containing conversions between OLI and Pyomo unit names
+    :param output_unit_set: dictionary containing preferred units for output expression
+    :param stream_output_options: dictionary pointing to properties that can be extracted from flash stream outputs
+    :param relative_inflows: bool to switch between additive and absolute substitutions for solutes when using surveys
```
adam-a-a, line 90: Not sure about this one- can you clarify what you mean?

veccp: @OOAmusat @avdudchenko and I discussed whether values in surveys should be absolute "inflow concentration = X", or additive, "inflow concentration + X", so this lets the user select their preference for solutes. As written, state variables will always use absolute values

watertap/tools/oli_api/flash.py
```diff
@@ -0,0 +1,494 @@
...
+        """
+
+        inputs = {"params": {}}
+
+        if flash_method == "wateranalysis":
+            inputs["params"] = {
+                "waterAnalysisInputs": self._build_input_list(state_vars)
+            }
+        else:
+            if not water_analysis_output:
```
adam-a-a, line 255: Hmm-- so this is basically saying-- if you want to you use isothermal flash, the only way to do that is to run water analysis first, which takes true species as input, and in the output from that, you can grab apparent species, which is the required input for isothermal flash.

However, what if I just want to provide apparent species from the get-go? Although in our team's usage of OLI thus far, we've used water analysis followed by isothermal flash, it's possible that the general user might want to just start off with isothermal flash. So, I wonder if you considered including something in `_build_input_list` that allow you to build inputs based on true species (water analysis) OR apparent species (isothermal flash). 

In that sense, the user wouldn't necessarily be restricted to having to run a water analysis first if they simply wanted to begin with apparent species, i.e., isothermal flash, when initiating a model. Does that make sense?

adam-a-a: BTW- I don't think this point should hold up this PR necessarily, unless you see a quick way to do this and have a burning desire to do so. I mostly just want to get your thoughts on this, hear from you on whether there are some subtle complexities with taking that approach, and then decide if we should create an issue for this to address at a later point.

veccp: My impression was that Isothermal Flash required true species as input, so I forced that part of the workflow. If that's not the case, the workflow could certainly be made more flexible to allow for different inputs

watertap/tools/oli_api/flash.py
```diff
@@ -0,0 +1,494 @@
...
+        """
+        Extract molecular concentrations from OLI flash output.
+
+        :param water_analysis_output: stream output from OLI flash calculation
+        :param inflows_phase: string indicating desired phase of inflows
+
+        :return inflows: dictionary containing molecular concentrations
+        """
+
+        if inflows_phase in ["liquid1", "vapor"]:
```
adam-a-a, line 287: Another thing to contemplate for the future---I think liquid2 is another possible phase in OLI. I can see this _**potentially**_ being useful for our upcoming work on solvent extraction

adam-a-a: (which would ideally make use of two liquid phases in a more detailed representation of the required property model)

veccp: That's true, probably along with unlocking the other flash calculations

watertap/tools/oli_api/flash.py
adam-a-a: Again- not something necessary for this PR of course. Rather, we can consider adding this to an issue that tracks all OLI-related enhancements that we want to incorporate in subsequent PRs.

veccp: That's a great idea

watertap/tools/oli_api/flash.py
adam-a-a: Ah, so Isothermal Flash takes in apparent species, while we can provide true species to Water Analysis. Water Analysis gives apparent species in its ouput though, and that's why we used that output as input to Isothermal Flash. The main reason we wanted to do that was because OLI makes certain optional properties accessible through Isothermal Flash results that are NOT produced by Water Analysis (or weren't at the time at least).

So to clarify with a simple example, using water analysis, we would feed in Na_+ and Cl_-. Using isothermal flash, we would use NaCl as input.

veccp: Ok, I'll look into that

watertap/tools/oli_api/credentials.py
adam-a-a: @veccp Can you add it? This PR is pretty much ready to go but this is a super minor revision that should be handled in this PR.

adam-a-a (approved):
LGTM - I think the only outstanding comment was a minor docstring issue, and even though I asked @veccp to add it in this PR, I don't mind just leaving it for a subsequent PR so that we don't have to wait for all tests to rerun in this PR.

adam-a-a:
> LGTM - I think the only outstanding comment was a minor docstring issue, and even though I asked @veccp to add it in this PR, I don't mind just leaving it for a subsequent PR so that we don't have to wait for all tests to rerun in this PR.

Oh- we'll have to wait for tests to rerun anyway since we have to merge main. I'll leave it open.

OOAmusat (approved):
LGTM

watertap/tools/oli_api/credentials.py
veccp: Yes, I just pushed the change

bknueven:
@adam-a-a I was waiting for us to test this against OLI before merging.

bknueven (approved):
Now that we can run the tests and they pass, I am happy.

PR #1240, opened 2023-12-06
Fixes and updates for differential parmater sweep tool 

avdudchenko:
## Fixes/Resolves:
This PR fixes differential parameter tool and updates it to comply with current ParamterSweep class usage

(1) Fixes issue with model re-initialization when solving differential sample, resulting in erroneous results 
(2) Makes Diff tool compatiable with parallel manager useing MultiProcessing, ConcurrentFutures, and RayIO, ensures stable usage with MPI
(3) Refactors parallel functions, and moves them into parameter_sweep_parallel_utils.py to provide generalized functions for use with parameter sweep and differential sweep class

## Summary/Motivation:
Diff tool was essentially broken, this provides critical fixes and ensures compatibility with current parallel infrastructure. 

-
-

### Legal Acknowledgement


watertap/tools/parameter_sweep/parameter_sweep.py
```diff
@@ -692,167 +701,9 @@ def parameter_sweep(self, *args, **kwargs):
         pass
 
 
-class ParameterSweep(_ParameterSweepBase):
+class ParameterSweep(_ParameterSweepBase, _ParameterSweepParallelUtils):
```
bknueven, line 725: What is going on that we need multiple inheritance here?

Is this just to separate the methods or does some other class need to just inherit from `_ParameterSweepBase`?

avdudchenko: its so both ParameterSweep and DifferentialParamterSweep class can get access to same underlying parallel functions. 

I was thinking of just putting them into _ParamterSweepBase as well, but figured it would be clenaer to store all relevant parallel function in a single place. 

We can also have _ParamterSweepBase inhert _ParameterSweepParallelUtils as ParatemterSweep and DifferntialParamterSweep inherit _ParamterSweepBase.  

Ultimately, the underlying functions would need to move out of paramtersweep class and either into their own class or _ParmaterSweepBase.  We can do what you think is better/cleaner.

bknueven (approved):

watertap/tools/parameter_sweep/tests/test_differential_parameter_sweep.py
```diff
@@ -740,14 +742,14 @@ def test_differential_parameter_sweep_selective(model, tmp_path):
             "pyomo_object": m.fs.input["b"],
         },
     }
-
+    build_spec = lambda model: differential_sweep_specs
     ps = DifferentialParameterSweep(
         csv_results_file_name=csv_results_file_name,
         h5_results_file_name=h5_results_file_name,
-        debugging_data_dir=tmp_path,
+        # debugging_data_dir=tmp_path, # Does not work at the moment
```
k1nshuk: Should we add a warning for this while it gets fixed?

k1nshuk (changes requested):
Looks good, except adding a warning for the debugging data dir.

k1nshuk (approved):

PR #1242, opened 2023-12-08
Add limiting current density methods to electrodialysis_0D

kejia6002:
## Fixes/Resolves:

## Summary/Motivation:
1. add empirical and theoretical limiting current density methods in electrodialysis_0D.py
2. add tests in test_electrodialysis_0D.py
3. update the 0D documentation

-
-

### Legal Acknowledgement

4. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.


lbibl (approved):

watertap/unit_models/electrodialysis_0D.py
```diff
@@ -2294,6 +2348,47 @@ def calculate_scaling_factors(self):
...
+                    iscale.set_scaling_factor(self.current_dens_lim_ioa, sf)
+                elif (
+                    self.config.limiting_current_density_method
+                    == LimitingCurrentDensityMethod.Theoretical
+                ):
+                    sf = (
+                        iscale.get_scaling_factor(self.N_Sh)
+                        * iscale.get_scaling_factor(self.diffus_mass)
+                        * iscale.get_scaling_factor(self.hydraulic_diameter) ** -1
+                        * 96485**-1
```
hunterbarber, line 2379: Just Constants.faraday_constant?

watertap/unit_models/electrodialysis_0D.py
```diff
@@ -2294,6 +2348,47 @@ def calculate_scaling_factors(self):
...
+                    self.config.limiting_current_density_method
+                    == LimitingCurrentDensityMethod.InitialValue
+                ):
+                    sf = self.config.limiting_current_density_data**-1
+                    iscale.set_scaling_factor(self.current_dens_lim_ioa, sf)
+                elif (
+                    self.config.limiting_current_density_method
+                    == LimitingCurrentDensityMethod.Empirical
+                ):
+                    sf = 25**-1 * sum(
```
hunterbarber, line 2363: What is the reasoning for 1/25 here? It's pretty mysterious reviewing it (without looking back at the equation and seeing what combination of variables is being approximated at 25), but a lot of scaling routines are.

watertap/unit_models/tests/test_electrodialysis_0D.py
```diff
@@ -1133,7 +1134,7 @@ def test_initialization_scaling(self, EDcell):
         iscale.set_scaling_factor(m.fs.unit.cell_length, 10)
         iscale.set_scaling_factor(m.fs.unit.cell_pair_num, 0.1)
         iscale.calculate_scaling_factors(m.fs)
-        initialization_tester(m)
+        initialization_tester(m, outlvl=idaeslog.DEBUG)
```
hunterbarber: Is this change just lingering from actually debugging the test file? Probably shouldn't remain.

watertap/unit_models/tests/test_electrodialysis_0D.py
```diff
@@ -1761,3 +1762,159 @@ def test_deltaP_configerr(self):
...
+        # Test ed_m2
+        iscale.calculate_scaling_factors(ed_m[2])
+        ed_m[2].fs.unit.diffus_mass.fix(1.6e-9)
+        assert degrees_of_freedom(ed_m[2]) == 0
+        initialization_tester(ed_m[2], outlvl=idaeslog.DEBUG)
+        results = solver.solve(ed_m[2])
+        assert_optimal_termination(results)
+        assert value(ed_m[2].fs.unit.current_dens_lim_ioa[0]) == pytest.approx(
+            281.100, rel=1e-3
+        )
```
hunterbarber, line 1920: I figured about this change in limiting current density between the different methods, and it puts into question the need for 'exact' scaling methods between the various methods. Idealistically I think that when running the simulation one could toggle between the configuration options (here limiting current density method) and resolve without rescaling each time because the methods should not result in orders of magnitude changes. Having unique scaling methods for each configuration option somewhat implies that it should be rescaled each time, but I would bet isn't the case and it would actually resolve without rescaling. I would potentially opt for a generalizable scaling method for all method options for conciseness, but would be curious to hear others opinions.

hunterbarber (commented):
Probably ready to approve but I had a couple comments.

watertap/unit_models/tests/test_electrodialysis_0D.py
lbibl: I honestly don't fully know the implication of different `outlvl` setting.  We can change it back if the change is not proper.

watertap/unit_models/tests/test_electrodialysis_0D.py
lbibl: Changing the method means changing one or more constraints in the model, which should in general point to re-running the scaling method I think. Calling the scaling method runs one method only.  Herein the scaling method has to use different sets of variables for different methods so I don't know how to generalize.

watertap/unit_models/electrodialysis_0D.py
lbibl: I don't know all scaling method needs to be based on a physical meaning.  Here the number was inspired by an empirical coefficient from a literature.

watertap/unit_models/electrodialysis_0D.py
lbibl: No difference and save the computational effort to strip the unit.

watertap/unit_models/tests/test_electrodialysis_0D.py
hunterbarber:
```suggestion
        initialization_tester(m)
```

hunterbarber (approved):

PR #1243, opened 2023-12-08
Add influent/effluent quality metrics for ASM1

adam-a-a:
## Summary/Motivation:
This PR adds quality metrics to ASM1 which can be used to evaluate WWTP performance and should be useful for optimization cases. Said metrics/properties can now be calculated for any ASM1 stateblock.

- add TSS
- add BOD5- note- this is the only one indexed by "raw" or "effluent", each of which incorporate a different conversion factor
- add COD
- add Kjeldahl nitrogen
- add total nitrogen
- add associated params as needed and move a couple params from ASM1 rxn to prop model

### Legal Acknowledgement

### DIFF
--- a/watertap/property_models/activated_sludge/asm1_properties.py
+++ b/watertap/property_models/activated_sludge/asm1_properties.py
@@ -125,6 +125,14 @@ class ASM1ParameterData(PhysicalParameterBlock):
             doc="Reference temperature",
             units=pyo.units.K,
         )
+        #TODO: f_p exists on the rxn parameter block and duplicating here temporarily; consolidate
+        self.f_p = pyo.Var(
+            initialize=0.08,
+            units=pyo.units.dimensionless,
+            domain=pyo.PositiveReals,
+            doc="Fraction of biomass yielding particulate products, f_p",
+        )
+        self.f_p.fix()
 
     @classmethod
     def define_metadata(cls, obj):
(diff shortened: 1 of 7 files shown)


watertap/property_models/activated_sludge/asm1_properties.py
```diff
@@ -356,6 +395,54 @@ def energy_density_expression(self):
...
+        )
+
+        def _BOD5(self, i):
+            bod5 = self.conc_mass_comp["X_S"] + self.conc_mass_comp["X_S"] + (1-self.params.f_p)*(self.conc_mass_comp["X_BH"] + self.conc_mass_comp["X_BA"])
+            #TODO: 0.25 should be a parameter instead as it changes by influent/effluent
+            return self.params.BOD5_factor[i]*bod5
+
+        self.BOD5 = pyo.Expression(
+            ["raw", "effluent"],
+            rule=_BOD5,
```
MarcusHolly, line 425: Maybe I'm just not understanding, but is there a reason these two are bundled together as opposed to having a separate expression for each?

adam-a-a: The only difference between the two is that for treated effluent, the whole expression should be multiplied by one value (0.25), while for another stateblock in the flowsheet (e.g., influent, stream bypassing activated sludge process, etc.), the expression is multiplied by another value (0.65). I could've made them separate but chose to do it this way. Note the `BOD5_factor` that I added to the parameter block is also indexed by "raw" or "effluent". (This is the factor that is 0.65 for "raw" and 0.25 for "effluent".)

MarcusHolly (approved):
LGTM - I'm assuming this PR is done

agarciadiego (approved):

luohezhiming (approved):
LGTM. Once it get merged, we can move forward to do some anlysis on these key metrics.

PR #1246, opened 2023-12-08
Automatically load MW and charge in MCAS

adam-a-a:
## Summary/Motivation:
This PR allows the user to now skip the step of entering MW and charge data for solutes---assuming they follow proper naming convention. If a name doesn't line up with what is expected, the user could manually enter data for those solutes but skip for more typical solute names (e,g,, `"Na_+"` and similar ions can be handled without a problem, but `"target_ion"` wouldn't turn up search results).

- load mw data by default if missing
- load charge data by default if missing
- bolster some of the watertap_to_oli helper functions with exceptions
- add testing for helper functions
- add get_molar_mass_quantity function to assign pint units to MW

### Legal Acknowledgement


watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
```diff
@@ -1398,23 +1398,6 @@ def model6():
...
-            mw_data={"H2O": 0.018, "Na_+": 0.023, "Cl_-": 0.0355, "N": 0.01},
-            charge={"Na_+": 1, "Cl_-": -1},
-        )
-    with pytest.raises(
-        ConfigurationError,
-        match="The charge argument was not provided while instantiating the MCAS property model. Provide a dictionary with solute names and associated charge as keys and values, respectively",
-    ):
-        m[0].fs.properties = MCASParameterBlock(
-            solute_list=["Na_+", "Cl_-", "N"],
-            mw_data={"H2O": 0.018, "Na_+": 0.023, "Cl_-": 0.0355, "N": 0.01},
```
veccp, line 1416: minor observation - mw for N is 0.014

veccp (approved):
looks good to me

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
adam-a-a: True - although I wasn't too careful when doing this since the tests are really intended to test functionality rather than true values (in these particular instances).

kurbansitterley (approved):
LGTM

PR #1252, opened 2023-12-14
Add flowsheet example of a 3-effect crystallizer

Zhuoran29:
## Fixes/Resolves:

Built an example of creating a flowsheet that describe a 3-effect crystallizer using WaterTAP crystallizer unit model.

## Summary/Motivation:

- Multi-effect crystallization has higher thermal efficiency compared to a single-effect crystallizer and is applied in industry.
- The example flowsheet depicts the following configuration with default setup.
3_effect_crys

- This PR doesn't include the HX component for the optimization of the operating conditions in each effect.

- Modified the crystallizer unit model to consider the available energy from the vapor
- Added a flowsheet in the `examples` folder

### Legal Acknowledgement

### DIFF
--- a/watertap/examples/flowsheets/crystallization/multi_effect_crystallizer.py
+++ b/watertap/examples/flowsheets/crystallization/multi_effect_crystallizer.py
@@ -179,7 +179,7 @@ def initialize_and_unfix_dof(m):
 
     # Solve
     results = solver.solve(m)
-    assert results.solver.termination_condition == TerminationCondition.optimal
+    assert_optimal_termination(results)
 
 
 def get_model_performance(m):
--- a/watertap/examples/flowsheets/crystallization/multi_effect_crystallizer.py
+++ b/watertap/examples/flowsheets/crystallization/multi_effect_crystallizer.py
@@ -1,0 +1,277 @@
+import pandas as pd
+import numpy as np
+import pytest
+from pyomo.environ import (
+    ConcreteModel,
+    TerminationCondition,
+    SolverStatus,
+    value,
+    Var,
+)
+from pyomo.network import Port
+from idaes.core import FlowsheetBlock
+from pyomo.util.check_units import assert_units_consistent
+from watertap.unit_models.crystallizer import Crystallization
+import watertap.property_models.cryst_prop_pack as props
+
+from idaes.core.solvers import get_solver
+from idaes.core.util.model_statistics import (
+    degrees_of_freedom,
+    number_variables,
+    number_total_constraints,
+    number_unused_variables,
+)
+from idaes.core.util.testing import initialization_tester
+from idaes.core.util.scaling import (
+    calculate_scaling_factors,
+    unscaled_variables_generator,
+    badly_scaled_var_generator,
+)
+from idaes.core import UnitModelCostingBlock
+
+from watertap.costing import WaterTAPCosting, CrystallizerCostType
+
+solver = get_solver()
+
+def build_multi_effect_crystallizer_fs(
+        m = None,
+        operating_pressure_eff1 = 0.35, # bar
+        operating_pressure_eff2 = 0.208, # bar
+        operating_pressure_eff3 = 0.095, # bar 
+        feed_mass_frac_NaCl = 0.2126, 
+        feed_pressure = 101325, # Pa
+        feed_temperature = 273.15 + 20, # K
+        crystallizer_yield = 0.5,
+        ):
+    '''
+    This flowsheet depicts a 3-effect crystallizer, with brine fed in parallel
+    to each effect, and the operating pressure is specfied individually.
+    '''
+
+    if m is None:
+        m = ConcreteModel()
+    m.fs = FlowsheetBlock(dynamic=False)
+    m.fs.props = props.NaClParameterBlock()
+
+    eff_1 = m.fs.effect_1 = Crystallization(property_package=m.fs.props)
+    eff_2 = m.fs.effect_2 = Crystallization(property_package=m.fs.props)
+    eff_3 = m.fs.effect_3 = Crystallization(property_package=m.fs.props)
+
+    # Specify the feed properties
+    feed_mass_frac_NaCl = 0.2126
+    feed_mass_frac_H2O = 1 - feed_mass_frac_NaCl
+    eps = 1e-6
+
+    eff_1.inlet.pressure[0].fix(feed_pressure)
+    eff_1.inlet.temperature[0].fix(feed_temperature)
+    eff_1.inlet.flow_mass_phase_comp[0, "Sol", "NaCl"].fix(eps)
+    eff_1.inlet.flow_mass_phase_comp[0, "Vap", "H2O"].fix(eps)
+
+    eff_2.inlet.pressure[0].fix(feed_pressure)
+    eff_2.inlet.temperature[0].fix(feed_temperature)
+    eff_2.inlet.flow_mass_phase_comp[0, "Sol", "NaCl"].fix(eps)
+    eff_2.inlet.flow_mass_phase_comp[0, "Vap", "H2O"].fix(eps)
+
+    eff_3.inlet.pressure[0].fix(feed_pressure)
+    eff_3.inlet.temperature[0].fix(feed_temperature)
+    eff_3.inlet.flow_mass_phase_comp[0, "Sol", "NaCl"].fix(eps)
+    eff_3.inlet.flow_mass_phase_comp[0, "Vap", "H2O"].fix(eps)
+
+    # Set up for effect 1
+    # Operating pressure: 0.35 bar
+    c1_pressure =  101325 * operating_pressure_eff1
+    feed_flow_mass = 1
+
+    eff_1.inlet.flow_mass_phase_comp[0, "Liq", "NaCl"].fix(
+        feed_flow_mass * feed_mass_frac_NaCl
+        )
+    eff_1.inlet.flow_mass_phase_comp[0, "Liq", "H2O"].fix(
+        feed_flow_mass * feed_mass_frac_H2O
+        )
+
+    eff_1.pressure_operating.fix(c1_pressure) 
+    eff_1.crystallization_yield["NaCl"].fix(crystallizer_yield)
+
+    eff_1.crystal_growth_rate.fix()
+    eff_1.souders_brown_constant.fix()
+    eff_1.crystal_median_length.fix()
+
+    # Set up for effect 2
+    # Same crystallizer_yield
+    # Operating pressure: 0.208 bar
+    c2_pressure =  101325 * operating_pressure_eff2
+
+    eff_2.pressure_operating.fix(c2_pressure) 
+    eff_2.crystallization_yield["NaCl"].fix(crystallizer_yield)
+
+    eff_2.inlet.flow_mass_phase_comp[0, "Liq", "NaCl"].fix(
+        feed_flow_mass * feed_mass_frac_NaCl
+    )
+    eff_2.inlet.flow_mass_phase_comp[0, "Liq", "H2O"].fix(
+        feed_flow_mass * feed_mass_frac_H2O
+    )
+
+    eff_2.crystal_growth_rate.fix()
+    eff_2.souders_brown_constant.fix()
+    eff_2.crystal_median_length.fix()
+
+    # Set up for effect 2
+    # Same crystallizer_yield
+    # Operating pressure: 0.095 bar
+    c3_pressure =  101325 * operating_pressure_eff3
+
+    eff_3.pressure_operating.fix(c3_pressure) 
+    eff_3.crystallization_yield["NaCl"].fix(crystallizer_yield)
+
+    eff_3.inlet.flow_mass_phase_comp[0, "Liq", "NaCl"].fix(
+        feed_flow_mass * feed_mass_frac_NaCl
+    )
+    eff_3.inlet.flow_mass_phase_comp[0, "Liq", "H2O"].fix(
+        feed_flow_mass * feed_mass_frac_H2O
+    )
+
+    eff_3.crystal_growth_rate.fix()
+    eff_3.souders_brown_constant.fix()
+    eff_3.crystal_median_length.fix()
+
+    assert degrees_of_freedom(m) == 0
+
+    # Scale
+    m.fs.props.set_default_scaling(
+        "flow_mass_phase_comp", 1e-1, index=("Liq", "H2O")
+    )
+    m.fs.props.set_default_scaling(
+        "flow_mass_phase_comp", 1e-1, index=("Liq", "NaCl")
+    )
+    m.fs.props.set_default_scaling(
+        "flow_mass_phase_comp", 1e-1, index=("Vap", "H2O")
+    )
+    m.fs.props.set_default_scaling(
+        "flow_mass_phase_comp", 1e-1, index=("Sol", "NaCl")
+    )
+    calculate_scaling_factors(m)
+
+    
+    return m
+
+def initialize_and_unfix_dof(m):
+    eff_1 = m.fs.effect_1
+    eff_2 = m.fs.effect_2
+    eff_3 = m.fs.effect_3 
+
+    # Initialize 
+    eff_1.initialize()
+    eff_2.initialize()
+    eff_3.initialize()
+        
+    # Unfix dof
+    brine_salinity = eff_1.properties_in[0].conc_mass_phase_comp["Liq", "NaCl"].value
+
+    eff_2.inlet.flow_mass_phase_comp[0, "Liq", "NaCl"].unfix()
+    eff_2.inlet.flow_mass_phase_comp[0, "Liq", "H2O"].unfix()
+    eff_2.properties_in[0].conc_mass_phase_comp["Liq", "NaCl"].fix(brine_salinity)
+
+    eff_3.inlet.flow_mass_phase_comp[0, "Liq", "NaCl"].unfix()
+    eff_3.inlet.flow_mass_phase_comp[0, "Liq", "H2O"].unfix()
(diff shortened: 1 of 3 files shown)


watertap/unit_models/crystallizer.py
```diff
@@ -558,6 +593,20 @@ def eq_crystallizer_height_constraint(b):
                 a + b + ((a - b) ** 2 + eps**2) ** 0.5
             )
 
+        # 13. Energy available from the vapor
+        @self.Constraint(doc="Thermal energy in the vapor")
+        def eq_vapor_energy_constraint(b):
```
OOAmusat, line 598: I'm not sure I understand this. It looks like the pure water and the vapor streams are exactly the same, or am I missing something? If they are not, how are they different?
- The flow rates are the same
- The pressures are the same

watertap/unit_models/crystallizer.py
```diff
@@ -558,6 +593,20 @@ def eq_crystallizer_height_constraint(b):
...
+            return b.energy_from_vapor == (
+                b.properties_vapor[0].flow_mass_phase_comp["Vap", "H2O"]
+                * (
+                    b.properties_pure_water[
+                        0
+                    ].dh_vap_mass_solvent  # Latent heat from the vapor
+                    + b.properties_vapor[0].enth_mass_solvent["Vap"]
```
OOAmusat, line 605: If the two streams are the same, then these cancel out, right?

watertap/unit_models/crystallizer.py
```diff
@@ -446,6 +473,14 @@ def eq_p_con2(b):
         def eq_p_con3(b):
             return b.properties_vapor[0].pressure == self.pressure_operating
 
+        @self.Constraint()
+        def eq_p_con4(b):
+            return b.properties_pure_water[0].pressure == self.pressure_operating
+
+        @self.Constraint()
+        def eq_p_con5(b):
+            return b.properties_pure_water[0].pressure_sat == self.pressure_operating
```
OOAmusat, line 482: I'm not sure I understand this. It looks like the pure water and the vapor streams are exactly the same, or am I missing something? If they are not, how are they different?

- The flow rates are the same
- The pressures are the same

watertap/unit_models/crystallizer.py
```diff
@@ -558,6 +593,20 @@ def eq_crystallizer_height_constraint(b):
...
 
+        # 13. Energy available from the vapor
+        @self.Constraint(doc="Thermal energy in the vapor")
+        def eq_vapor_energy_constraint(b):
+            return b.energy_from_vapor == (
+                b.properties_vapor[0].flow_mass_phase_comp["Vap", "H2O"]
+                * (
+                    b.properties_pure_water[
+                        0
+                    ].dh_vap_mass_solvent  # Latent heat from the vapor
```
OOAmusat, line 604: If the two streams are the same, then these cancel out, right?

watertap/unit_models/crystallizer.py
Zhuoran29: Yes thanks for asking as it's a bit confusing here and I'll describe it in the document. So you're right that the flow rates and pressure are the same, but the corresponding temperatures are not. 

The temperature of `properties_vapor` is in equilibrium with the solution from which it is generated, and the boiling point of the saturated brine is higher than that of pure water, leading to the superheated vapor going into the next effect (heat exchanger). The temperature of `properties_pure_water` is in equilibrium of pure water at the same pressure as calculated from `eq_p_con4` and `eq_p_con5`, while the other one is specified the same as the operating temperature.

For example in first effect of the sample diagram, the vapor temperature is 80.3 C, but it won't condense until it's cooled down to 73 C in the next effect.

Hope it helps clarify!

watertap/unit_models/crystallizer.py
Zhuoran29: As explained above, the actual available heat from the vapor includes the latent heat from condensation and the sensible heat from the superheated steam, although the latter is much smaller than the former.

It's a good idea to report those in the model results.

watertap/unit_models/crystallizer.py
OOAmusat: So you're bringing the steam from the superheated state to its saturation point and condensing. So, in effect, what you have at the end is the saturated liquid, correct?

watertap/unit_models/crystallizer.py
OOAmusat: Got it. Makes sense with the above clarification.

watertap/unit_models/crystallizer.py
Zhuoran29: Yes, I suppose this model naturally generates saturated liquid at the end, correspondingly related to the operating pressure/temperature, so that the salts can precipitate.

watertap/unit_models/crystallizer.py
```diff
@@ -316,6 +329,20 @@ def eq_mass_balance_constraints(b, j):
                 if (p, j) in b.properties_solids[0].phase_component_set
             )
 
+        @self.Constraint()
+        def eq_pure_vapor_flow_rate(b):
+            return (
+                b.properties_pure_water[0].flow_mass_phase_comp["Vap", "H2O"]
+                == b.properties_vapor[0].flow_mass_phase_comp["Vap", "H2O"]
+            )
```
adam-a-a, line 337: It's not immediately clear why we need this additional stateblock for water vapor. Furthermore, we are creating a constraint that sets the the new stateblock's mass flowrate to that of `properties_vapor`, which I worry could potentially lead to degeneracy in the model. 

If you are simply trying to use an alias for water in vapor phase, you could alternatively try adding a reference without creating a whole new stateblock. I might be missing something important here that would justify doing it this way but curious to hear your thoughts.

watertap/unit_models/crystallizer.py
```diff
@@ -316,6 +329,20 @@ def eq_mass_balance_constraints(b, j):
...
+
+        self.properties_pure_water[0].flow_mass_phase_comp["Liq", "H2O"].fix(1e-8)
+        self.properties_pure_water[0].flow_mass_phase_comp["Liq", "NaCl"].fix(0)
+        self.properties_pure_water[0].flow_mass_phase_comp["Sol", "NaCl"].fix(0)
+        self.properties_pure_water[0].mass_frac_phase_comp["Liq", "NaCl"]
+        self.properties_in[0].conc_mass_phase_comp["Liq", "NaCl"]
+        self.properties_in[0].flow_vol_phase["Liq"]
```
adam-a-a, line 344: Why touch these properties here and now within the unit model? My guess is that you want them to be calculated when reporting results after a solve (so a matter of convenience), but that could technically be done at the flowsheet level without introducing within the unit.

watertap/unit_models/crystallizer.py
adam-a-a, line 604: It is still not clear why we need this new stateblock. The only other reason that comes to mind is that it somehow helps with the multi-effect configuration.

watertap/unit_models/crystallizer.py
```diff
@@ -558,6 +593,20 @@ def eq_crystallizer_height_constraint(b):
                 a + b + ((a - b) ** 2 + eps**2) ** 0.5
             )
 
+        # 13. Energy available from the vapor
+        @self.Constraint(doc="Thermal energy in the vapor")
+        def eq_vapor_energy_constraint(b):
+            return b.energy_from_vapor == (
```
adam-a-a, line 599: I'm thinking that there might be an IDAES standard name that can be used in place of `energy_from_vapor` or something closer to an IDAES standard if an exact one does not exist: https://idaes-pse.readthedocs.io/en/stable/explanations/conventions.html

adam-a-a: Also, depending on your response for the new stateblock, I wonder if `energy_from_vapor` should actually be on the property model (it seems like it should be).

watertap/examples/flowsheets/crystallization/multi_effect_crystallizer.py
```diff
@@ -0,0 +1,306 @@
...
+    eff_1.crystal_growth_rate.fix()
+    eff_1.souders_brown_constant.fix()
+    eff_1.crystal_median_length.fix()
+
+    # Set up for effect 2
+    # Same crystallizer_yield
+    # Operating pressure: 0.208 bar
+    c2_pressure = 101325 * operating_pressure_eff2
+
+    eff_2.pressure_operating.fix(c2_pressure)
```
adam-a-a, line 105: you could just do 
`eff_2.pressure_operating.fix(0.28 * pyunits.bar)`

(assuming you imported `from pyomo.environ import units as pyunits`)

Also, it looks like you multiply each operating pressure by 101325 Pa, but the conversion is technically 1e5 Pa/ bar so wouldn't your effective operating pressures be slightly higher than intended?

watertap/examples/flowsheets/crystallization/multi_effect_crystallizer.py
```diff
@@ -0,0 +1,306 @@
...
+    def eqn_energy_from_eff1(b):
+        return b.fs.effect_2.work_mechanical[0] == b.fs.effect_1.energy_from_vapor
+
+    @m.Constraint(doc="Energy supplied to the 3rd effect")
+    def eqn_energy_from_eff2(b):
+        return b.fs.effect_3.work_mechanical[0] == b.fs.effect_2.energy_from_vapor
+
+    # Solve
+    results = solver.solve(m)
+    assert results.solver.termination_condition == TerminationCondition.optimal
```
adam-a-a:
```suggestion
    assert_optimal_termination(results)
```

adam-a-a: I think this also checks if solver status is OK.

watertap/unit_models/crystallizer.py
adam-a-a: > It's a good idea to report those in the model results.

Looking at the reporting done at the flowsheet level, I didn't see any usage of the unit model's `report` method, and given your comment here about reporting in model results, I wonder if you should add this to the unit model's `_get_performance_contents` method which is essentially part of what gets called when invoking `my_unit.report()`.

adam-a-a (changes requested):
This looks good and will definitely be of interest for the WaterTAP community to make use of this new flowsheet. There are some potential changes to consider however before we push this through.

watertap/examples/flowsheets/crystallization/multi_effect_crystallizer.py
```diff
@@ -0,0 +1,306 @@
+import pandas as pd
```
adam-a-a, line 1: You also would need to add a test file for this file.

watertap/examples/flowsheets/crystallization/multi_effect_crystallizer.py
```diff
@@ -0,0 +1,306 @@
...
+    """
+
+    if m is None:
+        m = ConcreteModel()
+    m.fs = FlowsheetBlock(dynamic=False)
+    m.fs.props = props.NaClParameterBlock()
+
+    eff_1 = m.fs.effect_1 = Crystallization(property_package=m.fs.props)
+    eff_2 = m.fs.effect_2 = Crystallization(property_package=m.fs.props)
+    eff_3 = m.fs.effect_3 = Crystallization(property_package=m.fs.props)
```
adam-a-a, line 59: So there is no explicit addition of IDAES heat exchangers in the flowsheet. Are you considering any efficiency losses, and is there a reason not to explicitly add HXs?

watertap/unit_models/crystallizer.py
Zhuoran29: Ah I think you saw my response to Mayo's similar question and hope it helps clarify. Please let me know if you have further questions regarding this new property package.

watertap/unit_models/crystallizer.py
Zhuoran29: Yea I realized that after I pushed the PR. Thanks! Will clean it up in the next commit.

watertap/unit_models/crystallizer.py
Zhuoran29: Same as above.

watertap/unit_models/crystallizer.py
Zhuoran29: Thanks for the reference. I think renaming it to `energy_flow_superheated_vapor` or `energy_superheated_vapor` could be better.

The difference between the formula I implemented and `enth_mass_solvent["Vap"]` from the property model (I think you were referring to this one probably) is that I'm trying to get the energy retrieved from the condensation of a superheated steam:

E (kJ/kg) = $\lambda$(T<sub>sat</sub>) + H<sub>vap</sub>(T<sub>super</sub>) - H<sub>vap</sub>(T<sub>sat</sub>) , 
where $\lambda$ is the latent heat, T<sub>super</sub> and T<sub>sat</sub> are the temperature of superheated vapor and saturation point.

And in the property model, the `enth_mass_solvent["Vap"]` of the existing`properties_vapor` is:
H<sub>vap</sub>(T<sub>sat</sub>) = $\lambda$(T<sub>sat</sub>) + H<sub>liq</sub>(T<sub>sat</sub>)

watertap/examples/flowsheets/crystallization/multi_effect_crystallizer.py
Zhuoran29: Yes you're right. I'll correct it, thanks!

watertap/examples/flowsheets/crystallization/multi_effect_crystallizer.py
Zhuoran29: Will add a test file for this, thanks!

watertap/examples/flowsheets/crystallization/multi_effect_crystallizer.py
Zhuoran29: Initially I also wanted to include the IDAES heat exchanger (0D) model to consider the heat transfer loss, but I didn't see it's included in the model - could be wrong about that. This PR focuses on determining how much waste water could be  disposed in each effect, so knowing the energy flow is sufficient, but of course the feed flow will decrease a bit when considering the losses.

However, I do plan to include the IDAES HX model (and a self-implemented costing function) for the cost optimization, as its capital cost (heat transfer area) is dependent on the approaching temperature between effects. Hope it makes sense.

watertap/unit_models/crystallizer.py
adam-a-a: I did, but it didn't justify the need to make a new stateblock, which is still foggy to me. I saw your explanation for the energy from vapor calculation, but I don't think you need a new stateblock for that either.

watertap/unit_models/crystallizer.py
adam-a-a: Ah, so based on this, you are essentially trying to get H<sub>vap</sub>-H<sub>liq</sub>, which sounds like it could be added to the property model perhaps as `enth_XXX` (XXX to fill in blank). In any case, if nothing from the unit model directly impacts the constraint and it is entirely based on properties, we usually try to put that relationship in the property model. I think we should aim to do that here before we merge in.

watertap/unit_models/crystallizer.py
Zhuoran29: Please let me put it in this way, the existing state `properties_vapor` describes a superheated state of the vapor (for example 0.35 bar, 80 C in effect 1 ), and I need a state describing the condensation state of vapor at the same pressure(0.35 bar, 72 C), in order to consider the sensible heat there. It's a virtual state because it's not a real state from the crystallizer but happens in a condenser or heat exchanger. 

So 80 C is in equilibrium with the brine feed `properties_out` as calculated by the model, while 72 C should be calculated from the equilibrium with pure water, which is not described in the current model.

Perhaps there's a way to derive the actual condensation temperature? I couldn't find a way with the existing states so I created a new one.  This new property should probably be renamed to `properties_vapor_equil` to indicate the equilibrium state.

watertap/unit_models/crystallizer.py
adam-a-a: EDIT: maybe not Hvap-Hliq exactly, but my point being that this is something that should be resolved in the property package, not in the unit model. Is there a reason not to do so, other than "this approach works too"?

Zhuoran29: Ohh, I got your point. I agree.

Although I think we should add `temperature_sat` as we have `pressure_sat`, and then calculate H<sub>vap</sub>(Tsat) as `enth_mass_vap_sat` or something.

watertap/unit_models/crystallizer.py
adam-a-a: OK, that makes a lot more sense now. Thanks for the detailed response!

watertap/unit_models/crystallizer.py
```diff
@@ -446,6 +473,14 @@ def eq_p_con2(b):
...
+        @self.Constraint()
+        def eq_p_con4(b):
+            return b.properties_pure_water[0].pressure == self.pressure_operating
+
+        @self.Constraint()
+        def eq_p_con5(b):
+            return b.properties_pure_water[0].pressure_sat == self.pressure_operating
+
         @self.Constraint()
         def eq_T_con1(b):
```
adam-a-a, line 485: Though these constraints are nearly self-explanatory, they should have doc strings describing what they are (e.g., "Isobaric pressure constraint for pressure of vapor at crystallizer outlet"). I know a couple of these were already missing doc strings before this PR, so cleaning those up too would be nice.

watertap/unit_models/crystallizer.py
adam-a-a: > Although I think we should add `temperature_sat` as we have `pressure_sat`, and then calculate Hvap(Tsat) as `enth_mass_vap_sat` or something.

Exactly what I was thinking too. That seems like the best way to incorporate with property model.

Zhuoran29:
This model is expected to be further modified in the next two months and is closed for now.

Issue #1219, opened 2023-11-28
Add documentation for flowsheet examples


adam-a-a:
None of our flowsheet examples have documentation. This is a considerable gap, and we should add docs for these. Moreover, as we add flowsheets to the GUI (see issue #1117), the need for documentation will increase. I will add an asterisk for those we should prioritize as high priority and a tilde (~) for those we may want to deprioritize. In terms of prioritization, all are still up for discussion though.

EDIT: Tagging people who I think should at the very least should be key points of contact for each:

# For March 2024:
- [x] BSM2 flowsheet* @adam-a-a @agarciadiego @MarcusHolly #1263 and #1354

# For June 2024:
- [x] RO w/ERD* @kurbansitterley #1446 
- [x] MD* @ElmiraShamlou #1444 
- [x] GAC* @hunterbarber #1396 
- [x] NaCl Crystallizer * @OOAmusat #1403 
- [x] ED * @lbibl #1385 #1386 
- [x] IX #1418 
- [x] LSRRO* @adam-a-a @ElmiraShamlou #1450
- [x] MVC* @MarcusHolly @ElmiraShamlou #1430 
- [x] OARO* @luohezhiming #1351 
- [x] Activate Sludge Model 1 (ASM1) flowsheets @agarciadiego #1376 
- [x] ASM2d flowsheets @agarciadiego @adam-a-a @luohezhiming #1376 
- [x] ADM1 flowsheet @agarciadiego #1376 
- [x] ~~electroN-P flowsheet @luohezhiming~~
- [x] ~~Municipal Treatment case study (originally from WT3) @kurbansitterley @TimBartholomew~~
- [x] Seawater RO case study (originally from WT3) @kurbansitterley @TimBartholomew #1425
- [x] ~~14 zero-order WRRF cases associated with IEDO 2336 @adam-a-a @MarcusHolly @luohezhiming @agarciadiego~~

**---Unit Model Documentation---**
- [x] CSTR (mostly a copy of the IDAES CSTR) #1379 
- [x] CSTR with injection (Aeration Tank) #1379 
- [x] Primary Clarifier #1379 
- [x] Secondary Clarifier #1379

adam-a-a:
@dangunter highlighted the fact that this effort should also work in coordination with the UI (e.g., being able to point to or pull up model descriptions via the UI)

PR #1263, opened 2024-01-03
BSM2 Flowsheet Documentation

MarcusHolly:
## Fixes/Resolves:

Issue #1219 

- Adds documentation for BSM2 and Extended BSM2 flowsheets
- Corrects all instances of "digestor" with "digester"

### Legal Acknowledgement


MarcusHolly:
I drafted some initial documentation for the BSM2 flowsheet - let me know your thoughts. Notably, there is documentation missing for some of the BSM2 unit models (clarifiers and CSTRs), so this might be something we want to add in this PR or a subsequent one.

docs/technical_reference/flowsheets/BSM2.rst
```diff
@@ -0,0 +1,44 @@
+Benchmark Simulation Model No.2
+===============================
+
+Introduction
+------------
+
+The Benchmark Simulation Model No.2 (BSM2) is an industry benchmark for modeling a full biological wastewater
+treatment plant that includes a primary clarifier, the activated sludge process, and an anaerobic digester. These
```
luohezhiming: We can also include thickener and dewatering unit for this description.

docs/technical_reference/flowsheets/extended_BSM2.rst
```diff
@@ -0,0 +1,57 @@
+Extended Benchmark Simulation Model No.2
+========================================
+
+Introduction
+------------
+
+Like the conventional Benchmark Simulation Model No.2 (BSM2), Extended BSM2 is an industry benchmark for modeling a full biological wastewater
```
luohezhiming: Probably we can add a link the the conventional BSM2 page

luohezhiming:
```suggestion
Like the conventional Benchmark Simulation Model No.2 (BSM2), extended BSM2 is an industry benchmark for modeling a full biological wastewater
```

docs/technical_reference/flowsheets/extended_BSM2.rst
```diff
@@ -0,0 +1,58 @@
...
+Documentation for each of the unit models can be found here:
+    * `Thickener <https://watertap.readthedocs.io/en/latest/technical_reference/unit_models/thickener.html>`_
+    * `Anaerobic digester <https://watertap.readthedocs.io/en/latest/technical_reference/unit_models/anaerobic_digester.html>`_
+    * `Dewatering unit <https://watertap.readthedocs.io/en/latest/technical_reference/unit_models/dewatering_unit.html>`_
+
+Documentation for each of the property models can be found here:
+    * `Modified ASM2d <https://watertap.readthedocs.io/en/latest/technical_reference/property_models/modified_ASM2D.html>`_
+    * `Modified ADM1 <https://watertap.readthedocs.io/en/latest/technical_reference/property_models/modified_ADM1.html>`_
+
+Future Refinements
```
agarciadiego, line 59: I think this should be on the BSM2 and not the extended version

agarciadiego (commented):
Just one comment the rest looks good to me

docs/technical_reference/flowsheets/extended_BSM2.rst
MarcusHolly: You bring up a good point. I think we should include a list on the BSM2 file, but I think we should also keep a list on the extended BSM2 file. While there is significant overlap between the two lists, I _believe_ some modifications only make sense for extended BSM2. For example, the sulfur and iron extensions (which I need to add to the list) would just be for extended BSM2. Same with precipitation and ion speciation - we've already identified literature (Flores-Alsina) that describes how to incorporate these into extended BSM2, but I forget if we've found similar sources for regular BSM2.

agarciadiego: Yes I agree two lists would be better

docs/technical_reference/flowsheets/BSM2.rst
```diff
@@ -0,0 +1,52 @@
...
+.. figure:: ../../_static/flowsheets/BSM2.png
+    :width: 800
+    :align: center
+
+    Figure 1. BSM2 flowsheet
+
+Documentation for each of the unit models can be found here:
+    * `Thickener <https://watertap.readthedocs.io/en/latest/technical_reference/unit_models/thickener.html>`_
+    * `Anaerobic digester <https://watertap.readthedocs.io/en/latest/technical_reference/unit_models/anaerobic_digester.html>`_
+    * `Dewatering unit <https://watertap.readthedocs.io/en/latest/technical_reference/unit_models/dewatering_unit.html>`_
```
adam-a-a, line 43: Should include links to (1) CSTR with some explanation saying that CSTR is used for anoxic reactors in activated sludge process (2) CSTR with injection to account for aeration tanks, (3) primary clarifier,  and (4) secondary clarifier. Additionally, we should state where any mixers, separators, or pumps are, outside of the aforementioned units. Ideally those would also be included in our flowsheet diagram (ATT: @luohezhiming). 

If we don't have technical ref for a particular unit yet, and IDAES doesn't either, we can skip the links for those in this PR and raise and issue to get to them eventually--but let's at least include them in the list.

docs/technical_reference/flowsheets/BSM2.rst
```diff
@@ -0,0 +1,52 @@
...
+
+Documentation for each of the property models can be found here:
+    * `ASM1 <https://watertap.readthedocs.io/en/latest/technical_reference/property_models/ASM1.html>`_
+    * `ADM1 <https://watertap.readthedocs.io/en/latest/technical_reference/property_models/ADM1.html>`_
+
```
adam-a-a, line 54: We might want to include a section on `Costing`, pointing to cost relationships we use in some way (whether that be manually writing relationships in table or simply pointing to other docs where we already provide said relationships).

Then we can add to Future Refinements that we plan to improve costing relationships in terms of detail, completeness, and reasonable validity.

adam-a-a: Alternatively, we can just do away with the Future Refinements section and just track these ourselves in Issues, as I think we may already do for some if not all of these

docs/technical_reference/flowsheets/extended_BSM2.rst
```diff
@@ -0,0 +1,60 @@
...
+.. figure:: ../../_static/flowsheets/BSM2.png
+    :width: 800
+    :align: center
+
+    Figure 1. BSM2 flowsheet
+
+Documentation for each of the unit models can be found here:
+    * `Thickener <https://watertap.readthedocs.io/en/latest/technical_reference/unit_models/thickener.html>`_
+    * `Anaerobic digester <https://watertap.readthedocs.io/en/latest/technical_reference/unit_models/anaerobic_digester.html>`_
+    * `Dewatering unit <https://watertap.readthedocs.io/en/latest/technical_reference/unit_models/dewatering_unit.html>`_
```
adam-a-a, line 47: same comment applies as for bsm2, but perhaps you can just point to conventional BSM2 unit list from here instead of rewriting here (or just copy+paste)

adam-a-a (changes requested):
A couple suggestions for the flowsheet docs. We can consider pushing some of this off to a future PR so long as we track the unaddressed items in issues. Let me know what you think @MarcusHolly

docs/technical_reference/flowsheets/BSM2.rst
MarcusHolly: I've added a "Unit Model Documentation" list in the current documentation issue #1219 in case other people come across unit models that don't have documentation. I just like having everything in one place, but we can easily split up the issue later, if necessary.

docs/technical_reference/flowsheets/BSM2.rst
MarcusHolly: Added a quick blurb in the implementation section pointing to this page: https://watertap.readthedocs.io/en/latest/technical_reference/costing/watertap_costing.html.

I feel like if we wanted to expand more on costing, the unit model documentation would be the place to do that, not in the flowsheet documentation.

adam-a-a: > I feel like if we wanted to expand more on costing, the unit model documentation would be the place to do that, not in the flowsheet documentation.

This was what I was implying when I said "pointing to other docs where we already provide said relationships." Since we don't have them in unit model docs yet, looks like we'll need to add those costing relationships in (subsequent PR).  Then from here, we could just include a sentence about implementation of technical performance and costing of each unit can be found in their technical references.

docs/technical_reference/flowsheets/BSM2.rst
adam-a-a: I like it!

adam-a-a (approved):
LGTM - we can refine this as other unit model documentation is updated too.

adam-a-a:
@MarcusHolly looks like a test failing because of `digestor` remaining in the file name and/or import for BSM2 extension

docs/technical_reference/flowsheets/BSM2.rst
```diff
@@ -0,0 +1,60 @@
...
+    Figure 1. BSM2 flowsheet
+
+Documentation for each of the unit models can be found here:
+    * `Thickener <https://watertap.readthedocs.io/en/latest/technical_reference/unit_models/thickener.html>`_
+    * `Anaerobic digester <https://watertap.readthedocs.io/en/latest/technical_reference/unit_models/anaerobic_digester.html>`_
+    * `Dewatering unit <https://watertap.readthedocs.io/en/latest/technical_reference/unit_models/dewatering_unit.html>`_
+    * `CSTR <https://idaes-pse.readthedocs.io/en/latest/reference_guides/model_libraries/generic/unit_models/cstr.html>`_
+    * CSTR with injection
+    * Primary clarifier
+    * Secondary clarifier
```
adam-a-a, line 49: Two more items that I forgot to mention- the translators!

docs/technical_reference/flowsheets/extended_BSM2.rst
```diff
@@ -0,0 +1,68 @@
...
+    * `CSTR <https://idaes-pse.readthedocs.io/en/latest/reference_guides/model_libraries/generic/unit_models/cstr.html>`_
+    * CSTR with injection
+    * Primary clarifier
+    * Secondary clarifier
```
adam-a-a, line 53: Same here- we should add the translators with links

adam-a-a (changes requested):
Just need to add the translators in the documentation and then I think we're set for now.

docs/technical_reference/flowsheets/BSM2.rst
MarcusHolly: Oh yea, don't know how I forgot those....

adam-a-a (approved):

luohezhiming (approved):

Issue #1262, opened 2024-01-02
More detailed pressure exchanger


TimBartholomew:
### Description

Allow users to model a pressure exchanger that accounts for leakage, mixing, and the pressure differential within a pressure exchanger.

### Motivation

Including mixing and leakage will make the model more realistic. Additionally, the primary manufacturer of pressure exchangers (Energy Recovery) provides mixing and leakage rates on their specification sheets.

### Possible Implementation

This implementation should be relatively straight forward.

1. Enable mass transfer for each control volume: has_mass_transfer = True
   - Adding a constraint to connect the mass transfer terms: mass_transfer_term_HPS = - mass_transfer_term_LPS
3. Incorporate leakage by:
   - Ensuring the equal volumetric flowrate constraint is only for the inlets: Q_LPS_in = Q_HPS_in
   - Adding a new fixed variable for the volumetric leakage called leakage_vol (default value = 0.01)
   - Adding a constraint to account for the leakage: Q_HPS_in * (1 - leakage_vol) = Q_HPS_out
4. Incorporate mixing by:
   - Adding a new fixed variable for the volumetric mixing called mixing_vol (default value = 0.035)
   - Adding a constraint to account for the mixing: C_LPS_out = C_LPS_in * (1 - mixing_vol) + C_HPS_in * mixing_vol
5. Incorporate pressure differential by:
   - Ensuring the efficiency_pressure_exchanger is unfixed. Note: that we can keep it and the pressure transfer constraint because now the unfixed variable will just be calculated.
   - Adding a new fixed variable for the pressure differential called high_pressure_difference (default value = 0.8 bar)
   - Adding a constraint to account for the pressure differential: P_LPS_out + high_pressure_difference = P_HPS_in
   - Note: we can keep our previous assumption that P_HPS_out = P_LPS_in

Since we would like users to have access to our previous simpler model and this newer more detailed one, we should pair these changes with a configuration argument(s). I can see two options:
1. A single configuration argument for switching between the two models. The configuration argument could be level_of_detail, with the options being simple and detailed. (Although I'm not sure about those names, both are relatively simple assumed performance models)
2. Multiple configuration arguments for the modeled phenomena/relationships.
   - A configuration argument for leakage - has_leakage and the options can be True or False
   - A configuration argument for mixing - has_mixing and the options can be True or False
   - A configuration argument for the pressure exchange calculation - pressure_exchange_calculation and the options can be "efficiency" or "high pressure difference"
[...]

PR #1264, opened 2024-01-05
Add more detailed pressure exchanger

luohezhiming:
## Fixes/Resolves:

replace this with the issue #1262  fixed or resolved, 

## Summary/Motivation:
Add more details for the pressure exchanger that accounts for leakage, mixing, and the pressure differential

- revise current pressure exchanger
- add corresponding tests

### Legal Acknowledgement


watertap/unit_models/pressure_exchanger.py
```diff
@@ -201,24 +232,46 @@ def build(self):
...
+            self.config.pressure_exchange_calculation
+            is PressureExchangeType.high_pressure_difference
+        ):
+            self.high_pressure_difference = Var(
+                self.flowsheet().config.time,
+                initialize=0.8e5,
+                bounds=(1e-6, None),
+                domain=NonNegativeReals,
+                units=pyunits.Pa,
+                doc="High pressure difference",
```
adam-a-a: Remind me- is this meant to be the pressure drop on the high-pressure side of the ERD? Assuming that's the case, why not use `deltaP` which should exist on the `ControlVolume `representing the high-pressure side?

watertap/unit_models/pressure_exchanger.py
```diff
@@ -201,24 +232,46 @@ def build(self):
...
-        self.efficiency_pressure_exchanger = Var(
-            self.flowsheet().config.time,
-            initialize=0.95,
-            bounds=(1e-6, 1),
-            domain=NonNegativeReals,
-            units=pyunits.dimensionless,
-            doc="Pressure exchanger efficiency",
-        )
+        if self.config.pressure_exchange_calculation is PressureExchangeType.efficiency:
+            self.efficiency_pressure_exchanger = Var(
```
adam-a-a: I'm not sure that efficiency should be excluded in the new configuration. I would think we should keep it in and just calculate instead of fixing it.

watertap/unit_models/pressure_exchanger.py
adam-a-a: OK- seems like this is not just pressure drop on the high-pressure control volume.

docs/technical_reference/unit_models/pressure_exchanger.rst
```diff
@@ -5,8 +5,8 @@ This pressure exchanger unit model:
     * is isothermal
     * supports a single liquid phase only
     * supports steady-state only
-    * assumes no mixing or leakage between the low and high pressure side
-    * assumes equal flowrates on both sides
+    * supports mixing or leakage between the low and high pressure side
+    * assumes equal flowrates on both side
```
TimBartholomew: this should be changed to: enforces a constraint that the inlet flowrates to each side must be equal

docs/technical_reference/unit_models/pressure_exchanger.rst
```diff
@@ -32,6 +32,13 @@ Where the system is also subject to following constraints:
     
     Figure 1. Schematic representation of an energy recovery system using a pressure exchanger.
 
+When setting the ``has_leakage`` configuration option to ``True``, there is 1 additional variables ``leakage_vol`` that must be fixed.
```
TimBartholomew: typo: "1 additional variables" should drop the plural s -> "1 additional variable"

docs/technical_reference/unit_models/pressure_exchanger.rst
```diff
@@ -32,6 +32,13 @@ Where the system is also subject to following constraints:
...
+
+When setting the ``has_mixing`` configuration option to ``True``, there is 1 additional variables ``mixing_vol`` that must be fixed.
```
TimBartholomew: same as above

watertap/unit_models/pressure_exchanger.py
TimBartholomew: I agree

docs/technical_reference/unit_models/pressure_exchanger.rst
```diff
@@ -32,6 +32,13 @@ Where the system is also subject to following constraints:
...
+
+When setting the ``pressure_exchange_calculation`` configuration option to ``PressureExchangeType.high_pressure_difference``,
+then there is no ``efficiency_pressure_exchanger`` variable, users need to fix ``high_pressure_difference`` instead.
```
TimBartholomew: Don't forget to update this, once you allow pressure exchanger efficiency to be created even with high_pressure_difference option. See Adam's comment

watertap/unit_models/pressure_exchanger.py
```diff
@@ -320,26 +370,41 @@ def eq_equal_low_pressure(b, t):
...
+                )
+
+        elif (
+            self.config.pressure_exchange_calculation
+            is PressureExchangeType.high_pressure_difference
+        ):
+
+            @self.Constraint(self.flowsheet().config.time, doc="Pressure transfer")
+            def eq_pressure_transfer(b, t):
+                return b.high_pressure_side.properties_in[
```
TimBartholomew: if you put parentheses around the statement, I think black won't make it look crazy:

`return (pressure_HPS_in = pressure_LPS_out + HPD)`

watertap/unit_models/pressure_exchanger.py
```diff
@@ -352,6 +417,93 @@ def eq_mass_transfer_term(b, t, p, j):
...
+                    self.config.property_package.solute_set,
+                    doc="Mixing effect of the unit",
+                )
+                def eq_mixing(b, t, p, j):
+                    return (
+                        b.low_pressure_side.properties_out[t].conc_mass_phase_comp[p, j]
+                        == b.low_pressure_side.properties_in[t].conc_mass_phase_comp[
+                            p, j
+                        ]
+                    )
```
TimBartholomew: I don't think this is correct. If there is some leakage, it will impact the concentration. This implementation essentially would allow leakage which is basically dictated by the water mass transfer term, but then the concentration would be held constant by adjusting the salt mass transfer term.

Maybe the real fix is that we don't allow there to be no mixing if there is leakage (i.e. we return an error if leakage = True and mixing = False)

watertap/unit_models/pressure_exchanger.py
```diff
@@ -448,16 +601,26 @@ def propogate_state(sb1, sb2):
...
+            self.low_pressure_side.properties_out[
+                0
+            ].pressure = self.low_pressure_side.properties_in[
+                0
+            ].pressure.value + self.efficiency_pressure_exchanger[
+                0
+            ].value * (
+                self.high_pressure_side.properties_in[0].pressure.value
+                - self.low_pressure_side.properties_in[0].pressure.value
+            )
```
TimBartholomew: I think if you put parentheses around this, black won't make the formatting look so bad

watertap/unit_models/pressure_exchanger.py
```diff
@@ -487,12 +650,20 @@ def calculate_scaling_factors(self):
...
-        if hasattr(self, "mass_transfer_fraction_comp"):
-            if iscale.get_scaling_factor(self.mass_transfer_fraction_comp) is None:
-                iscale.set_scaling_factor(self.mass_transfer_fraction_comp, 1)
+        if hasattr(self, "efficiency_pressure_exchanger"):
+            if iscale.get_scaling_factor(self.efficiency_pressure_exchanger) is None:
+                # efficiency should always be between 0.1-1
+                iscale.set_scaling_factor(self.efficiency_pressure_exchanger, 1)
+        if hasattr(self, "high_pressure_difference"):
+            if iscale.get_scaling_factor(self.high_pressure_difference) is None:
+                # efficiency should always be between 0.1-1
```
TimBartholomew: Should be deleted - "# efficiency should always be between 0.1-1"

watertap/unit_models/pressure_exchanger.py
avdudchenko: Agreed! We should add an constraint to calc efficiency, which also means ,this constraint can be used to calculate one of he parameters in the new model (either pressure loss, or leakage amount I would guess).

watertap/unit_models/pressure_exchanger.py
```diff
@@ -561,7 +737,7 @@ def _get_performance_contents(self, time_point=0):
         t = time_point
         return {
             "vars": {
-                "Efficiency": self.efficiency_pressure_exchanger[t],
+                # "Efficiency": self.efficiency_pressure_exchanger[t],
```
TimBartholomew: Let's keep this and just have it always be active.

TimBartholomew (commented):
Some minor comments, a couple proposed changes

watertap/unit_models/pressure_exchanger.py
luohezhiming: I tried, but Black will directly reformat back to this formula and deleted the parentheses I added.

watertap/unit_models/pressure_exchanger.py
luohezhiming: Same as above, added paranthesses will be deleted and change back to this formula by Black

watertap/unit_models/pressure_exchanger.py
luohezhiming: I revise this to directly raise an error if only set `has_leakage=True`.

luohezhiming:
@TimBartholomew @adam-a-a I've revised the model based on your suggestion. `efficiency_pressure_exchanger` is changed to a universal variable, and the model will raise an error as `Setting 'has_leakage=True' requires 'has_mixing=True'` if only set `has_leakage=True`. Relevant documentation also updated.

ElmiraShamlou (changes requested):
It's a great effort to enhance the accuracy of the pressure exchanger model by addressing inefficiencies. However, essential modifications are necessary to accurately reflect the inefficiencies within the pressure exchanger. A primary problem is the current model's treatment of leakage and mixing as if they were interconnected, whereas they are distinct processes with different impacts on the system. It's critical to note that leakage does not result in mass transfer from concentrated brine to the feed stream, and the volumetric flow rates of both the feed stream and concentrated brine remain unchanged within the PX system. The only mass transfer occurs through mixing at the interface of the two streams during the pressure exchange process.

To elaborate, leakage is typically quantified as a percentage of the high-pressure concentrated brine stream that bypasses the seals and circulates around the rotor within the PX device assembly, acting as a lubrication flow. This lubrication flow exits the system with the depressurized concentrated brine, not the feed stream, without transferring its useful energy to the feed stream, which should be considered a loss.

Furthermore, in the context of the pressure exchanger, minimizing mixing is achieved by operating the feed and concentrated brine streams at the same flow rate during pressure exchange. Given that a portion of the concentrated brine leaks and does not participate in pressure exchange, the flow rate of the feed stream should be adjusted to match the flow rate of the concentrated brine, minus the leakage, to maintain equal flow.

The current model's approach to defining high-pressure and low-pressure sides may also contributing to this confusion. In pressure exchanger terminology, the high-pressure side includes both the feed and concentrated reject as the high-pressure outlet and inlet, respectively, and the input feed and depressurized reject as the low-pressure inlet and outlet. Therefore, leakage from the high-pressure side to the low-pressure side indicates that the high-pressure reject leaks through the seals to the low-pressure side reject, exiting the system. This contradicts the current model's terminology, where the high-pressure and low-pressure sides are defined as the feed and concentrated reject, respectively.

docs/technical_reference/unit_models/pressure_exchanger.rst
```diff
@@ -32,6 +31,13 @@ Where the system is also subject to following constraints:
     
     Figure 1. Schematic representation of an energy recovery system using a pressure exchanger.
 
+When setting the ``has_mixing`` configuration option to ``True``, there is 1 additional variable ``mixing_vol`` that must be fixed.
```
adam-a-a, line 34:
```suggestion
When setting the ``has_mixing`` configuration option to ``True``, there is 1 additional variable ``mixing_vol_frac`` that must be fixed.
```
Adding `_frac` to the variable name would be more appropriate.

docs/technical_reference/unit_models/pressure_exchanger.rst
```diff
@@ -32,6 +31,13 @@ Where the system is also subject to following constraints:
...
+
+When setting the ``has_leakage`` configuration option to ``True``, there is 1 additional variable ``leakage_vol`` that must be fixed.
```
adam-a-a, line 36:
```suggestion
When setting the ``has_leakage`` configuration option to ``True``, there is 1 additional variable ``leakage_vol_frac`` that must be fixed.
```

docs/technical_reference/unit_models/pressure_exchanger.rst
```diff
@@ -81,18 +93,56 @@ Each property block on both control volumes (i.e. `properties_in` and `propertie
...
 
    "Mass balance for each side", ":math:`M_{out, j} = M_{in, j}`"
    "Momentum balance for each side", ":math:`P_{out} = P_{in} + ΔP`"
    "Isothermal assumption for each side", ":math:`T_{out} = T_{in}`"
-   "Equal volumetric flowrate*", ":math:`Q_{LPS} = Q_{HPS}`"
+   "Equal volumetric flowrate*", ":math:`Q_{out, LPS} = Q_{in, HPS}`"
+   "Equal pressure*", ":math:`P_{out, HPS} = P_{in, LPS}`"
    "Pressure transfer*", ":math:`ΔP_{LPS} = - \eta ΔP_{HPS}`"
 
 \* LPS stands for low pressure side, HPS stands for high pressure side
```
adam-a-a: I wonder if we should add a diagram of the ERD to clearly illustrate the "LPS" and "HPS" as well as label the inlets and outlets on the diagram along with where leakage would occur and where mixing would occur (to make it visually apparent to readers/users).

docs/technical_reference/unit_models/pressure_exchanger.rst
```diff
@@ -81,18 +93,56 @@ Each property block on both control volumes (i.e. `properties_in` and `propertie
...
+and there are 3 additional constraints:
+
+.. csv-table::
+   :header: "Description", "Equation"
+
+   "Mixing effect of solute*", ":math:`C_{out, LPS} = C_{in, LPS} (1-\chi) + C_{in, HPS} \chi`"
+   "Linking mass transfer terms", ":math:`MTT_{j, LPS} = -MTT_{j, HPS}`"
+   "Equal low pressure side volumetric flowrate", ":math:`Q_{out, LPS} = Q_{in, LPS}`"
+
+\* C only represents solute concentration, not solvent
```
adam-a-a:
```suggestion
\* C represents solute concentration
```

docs/technical_reference/unit_models/pressure_exchanger.rst
```diff
@@ -56,6 +62,10 @@ The pressure exchanger unit model includes the following variables:
    :header: "Description", "Symbol", "Variable Name", "Index", "Units", "Pyomo Type" 
 
    "Efficiency", ":math:`\eta`", "efficiency_pressure_exchanger", "[t]", ":math:`\text{dimensionless}`", "Var"
+   "Volumetric leakage fraction", ":math:`\delta`", "leakage_vol", "[t]", ":math:`\text{dimensionless}`", "Var"
+   "Volumetric mixing fraction", ":math:`\chi`", "mixing_vol", "[t]", ":math:`\text{dimensionless}`", "Var"
+   "High pressure difference", ":math:`HPD`", "high_pressure_difference", "[t]", ":math:`\text{Pa}`", "Var"
+   "Low pressure difference", ":math:`LPD`", "low_pressure_difference", "[t]", ":math:`\text{Pa}`", "Var"
```
adam-a-a: Clarify the sign convention for HPD and LPD (i.e., fix negative or positive values). I believe they are meant to be positive.

docs/technical_reference/unit_models/pressure_exchanger.rst
```diff
@@ -81,18 +93,56 @@ Each property block on both control volumes (i.e. `properties_in` and `propertie
...
+
+\* C only represents solute concentration, not solvent
+
+if ``pressure_change_calculation`` is set to ``PressureExchangeType.high_pressure_difference``,
+then there is 1 additional constraint and the equal pressure equation is replaced:
+
+.. csv-table::
+   :header: "Description", "Equation"
+
+   "High pressure difference", ":math:`P_{out, LPS} + HPD = P_{in, HPS}`"
```
adam-a-a:
```suggestion
   "Pressure drop across HPS,in and LPS, out", ":math:`P_{out, LPS} + HPD = P_{in, HPS}`"
```

docs/technical_reference/unit_models/pressure_exchanger.rst
```diff
@@ -81,18 +93,56 @@ Each property block on both control volumes (i.e. `properties_in` and `propertie
...
+   "Equal pressure", ":math:`P_{out, HPS} = P_{in, LPS} + LPD`"
```
adam-a-a:
```suggestion
   "Pressure drop across HPS, out and LPS, in", ":math:`P_{out, HPS} = P_{in, LPS} + LPD`"
```

watertap/unit_models/pressure_exchanger.py
```diff
@@ -231,10 +292,16 @@ def build(self):
...
-        )
+        if self.config.has_mixing:
+            self.high_pressure_side.add_material_balances(
+                balance_type=self.config.material_balance_type,
+                has_mass_transfer=True,
+            )
+        else:
+            self.high_pressure_side.add_material_balances(
+                balance_type=self.config.material_balance_type,
+                has_mass_transfer=False,
```
adam-a-a:
```suggestion
        self.high_pressure_side.add_material_balances(
                balance_type=self.config.material_balance_type,
                has_mass_transfer=self.config.has_mixing,
            )
```

watertap/unit_models/pressure_exchanger.py
```diff
@@ -265,10 +332,16 @@ def work(b, t):
...
+        if self.config.has_mixing:
+            self.low_pressure_side.add_material_balances(
+                balance_type=self.config.material_balance_type,
+                has_mass_transfer=True,
+            )
+        else:
+            self.low_pressure_side.add_material_balances(
+                balance_type=self.config.material_balance_type,
+                has_mass_transfer=False,
+            )
```
adam-a-a:
```suggestion
        self.low_pressure_side.add_material_balances(
                balance_type=self.config.material_balance_type,
                has_mass_transfer=self.config.has_mixing,
            )
```

adam-a-a (changes requested):
LGTM - just a couple of suggestions to consider.

docs/technical_reference/unit_models/pressure_exchanger.rst
luohezhiming: Yeah, they should be non-negative values, I've added a comment to clarify.

docs/technical_reference/unit_models/pressure_exchanger.rst
luohezhiming: I would keep the variable as `mixing_vol` for consistency, since they are defined vairables in the unit model.

docs/technical_reference/unit_models/pressure_exchanger.rst
luohezhiming: Same as above.

docs/technical_reference/unit_models/pressure_exchanger.rst
luohezhiming: The original doc has a diagram with ERD and RO, I think I can try to modify it to correctly reflect the LPS and HPS. Actually, I review the literatuer and find that LPS and HPS are also convention terminology for PX. However, it is different than our definition, it is only determined by the pressure of seawater side. In other words, their LPS_in flows to HPS_out. I think probably we'd better rename the control volume.   I will ask if Tim has the original PPT file with figure and I can modify it.

watertap/unit_models/pressure_exchanger.py
```diff
@@ -210,134 +241,203 @@ def build(self):
...
+        self.add_inlet_port(name="feed_inlet", block=self.feed_side)
+        self.add_outlet_port(name="feed_outlet", block=self.feed_side)
 
         # Performance equations
         @self.Constraint(self.flowsheet().config.time, doc="Pressure transfer")
         def eq_pressure_transfer(b, t):
             return (
-                b.low_pressure_side.deltaP[t]
-                == b.efficiency_pressure_exchanger[t] * -b.high_pressure_side.deltaP[t]
+                b.feed_side.deltaP[t]
```
ElmiraShamlou, line 363: I suggest implementing a more precise efficiency calculation method for PressureExchangeType.high_pressure_difference option, defining efficiency as the ratio of energy output to energy input. This approach effectively incorporates the effects of HPD and leakage into efficiency

luohezhiming: @ElmiraShamlou the current definition of  `efficiency` in `eq_pressure_transfer` is exactly same as ratio of energy output to energy input. I've tried to directly define the `efficiency` based on the energy output/energy input and get exact same results. Though, it is not so evident, ΔP_FS=-ηΔP_BS is same as energy output/ energy input

ElmiraShamlou: Thanks @luohezhiming . The current definition of efficiency is equivalent to energy ratios only when there's no leakage and the flow rates on both the feed and brine sides are the same. You got the same results from both methods because the leakage was either very small or zero. But if leakage increases, it will have a bigger effect on efficiency. We can include the leakage inefficiency in a complete energy equation that incorporates both pressure and flow rates.

luohezhiming: I redefine the efficiency as the ratio of energy_in over energy_out, but I find the solution results will change for the `no_mass_transfer` case. Since a lot of flowsheet use this basic pressure exchanger without mass transfer, @TimBartholomew , I want to confirm with you should we change this efficiency equation? If we do, then all flowsheets using the pressure exchanger will have different results.

watertap/unit_models/pressure_exchanger.py
```diff
@@ -403,31 +514,31 @@ def initialize_build(
...
-                value(self.low_pressure_side.properties_in[0].flow_vol)
-                - value(self.high_pressure_side.properties_in[0].flow_vol)
+                value(self.feed_side.properties_in[0].flow_vol)
+                - value(self.brine_side.properties_in[0].flow_vol)
             )
-            / value(self.high_pressure_side.properties_in[0].flow_vol)
+            / value(self.brine_side.properties_in[0].flow_vol)
             > 1e-4
-            and not self.config.has_mass_transfer
+            and not self.config.has_mixing
```
ElmiraShamlou, line 534: I'm not sure if this error message actually helps with initialization. If we need to keep it, maybe we should change self.config.has_mixing to self.config.has_leakage.

ElmiraShamlou (approved):
Approved. Great work! 
Based on the team's decision, some of the requested changes can wait until future PRs.

watertap/unit_models/pressure_exchanger.py
luohezhiming: @ElmiraShamlou I agree we also have to add `has_leakage ` for this check, but we also need to include `has_mixing` because we introduce `mass_transfer_term` into the control volume when add mixing even though it might not affect the volumetric flowrate if there is no leakage, but the introduced `mass_transfer_term` might cause some minor errors. Note that in most cases without transfer, after initialization, we should have Q_FS_in=Q_BS_out and deactivate the `equal_flow_vol` constraint to make initialization work.

adam-a-a: @luohezhiming please tag this in the issue for tracking any potential updates we may (or may not) want to make in in a future PR. For now, I think we can merge what you have for the March release and refine later, if needed.

adam-a-a (approved):
LGTM - @luohezhiming please make an issue to track any potential improvements we might want to consider later for the detailed presssure exchanger. I have on or two that come to mind, so I will add after you start the issue.

PR #1266, opened 2024-01-09
Add model to solve function in UI

avdudchenko:
## Fixes/Resolves:
Currently, UI passes the flowsheet (m.fs) into the solve function used by the UI, but to be consistent with rest of watertap usage/IDAES usage, we should be passing in the model it self (m) into the solve function. This updates the fast api to pass the model in, and rest of uis to handle the new behavior. 

## Summary/Motivation:
- include model for users to solve in ui solve function. 

-
-

### Legal Acknowledgement

### DIFF
--- a/watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery_ui.py
+++ b/watertap/examples/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery_ui.py
@@ -30,12 +30,11 @@ def export_to_ui():
     )
 
 
-def export_variables(flowsheet=None, exports=None, build_options=None, **kwargs):
-    fs = flowsheet.fs
+def export_variables(model=None, exports=None, build_options=None, **kwargs):
     # --- Input data ---
     # Feed conditions
     exports.add(
-        obj=fs.feed.properties[0].flow_mass_phase_comp["Liq", "H2O"],
+        obj=model.fs.feed.properties[0].flow_mass_phase_comp["Liq", "H2O"],
         name="Water mass flowrate",
         ui_units=pyunits.kg / pyunits.s,
         display_units="kg/s",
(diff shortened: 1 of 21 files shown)


MichaelPesce (approved):

adam-a-a (commented):
My only issue with this is a petty one--we are maintaining the `flowsheet` arg when it should now be `model`, based on your changes, technically.

avdudchenko:
> My only issue with this is a petty one--we are maintaining the `flowsheet` arg when it should now be `model`, based on your changes, technically.

I left it as that for now to make it clear that the change is happening, but over time we should switch it over to "model", not sure it makes sense @MichaelPesce and @dangunter  and if we should just update flowsheet to be model through out...

avdudchenko:
@adam-a-a @MichaelPesce @dangunter I updated the PR so it now uses model through out.

adam-a-a:
@dangunter from discussions during last dev call, I recall that you wanted to weigh in here. @avdudchenko has gone ahead and revised the `flowsheet` arg as `model`, matching up with his new changes. However, you may want to see how this channels through under the hood.

ksbeattie:
@avdudchenko, @adam-a-a I believe we decided to hold on this for the moment.

avdudchenko:
We should complete merging this - in effect the way we use UI right now with flowsheet vs model is broken. can we merge this in? Before more UI's are added into WT?) @dangunter @MichaelPesce @adam-a-a

lbianchi-lbl:
> We should complete merging this - in effect the way we use UI right now with flowsheet vs model is broken. can we merge this in? Before more UI's are added into WT?) @dangunter @MichaelPesce @adam-a-a

@avdudchenko Apologies for not providing more context. After a bit of a back and forth in the dev call, the consensus seems to be:

- Conceptually:
  - The general question of "should `watertap.ui.fsapi` be based around the `ConcreteModel` instance or the `FlowsheetBlock` instance?" might have some merit
  - However, both have their pros and cons, so the opinion seems to be that `ConcreteModel` is not the obviously "better" choice
- Concretely:
  - This would be a disruptive change that would require a considerable amount of effort, and the maintainers of this part of the WaterTAP codebase don't have the bandwidth to deal with this for the foreseeable future
  - The need for being able to access the `ConcreteModel` instance has not been a requirement for most UI modules until now
  - There is a straightforward workaround that only relies on documented/public Pyomo APIs and is therefore a valid choice when this need does exist

The workaround is calling `fs.model()`, which will return the `ConcreteModel` instance from any `Block` no matter its position in the block tree.

Given the above, our decision was to close this PR as not planned. Does it make sense? If you get a chance of trying out the `fs.model()` workaround and you think it's a reasonable solution, we could think of adding some documentation (perhaps a short "How to" guide) for future reference.

PR #1268, opened 2024-01-15
Add method for ParamterSweepTool to change lower/upper bounds, and fix/unfix variables. 

avdudchenko:
## Summary/Motivation:

This adds new optional arguments to parameter sweep options,enabling user to sweep across lower bounds, upper bound, or fix and unfix variables. Primary objective is to enable complete depreciation of prior method of passing in a model directly into a PS tool and instead requiring user to pass in build_model function. 

This future change will mean a user would need to explicitly specify lower/upper bound or fixed/unfixed states in model init or optimize function.

This PR enables user to now include the changed lower/upper bound or fixed/unfixed states as part of sweep_paramters, ensuring these values are updated after a model is built or initialized.  

This update is also meant for the WaterTAP UI which needs to update the mode with user supplied variables before running a sweep, currently UI relies on old ParamterSweep interface, this will enable us to migrate to new interface and enable local parallel utilization. 

-add SetMode class, the defines if PS tool should fix variable, set upper or lower bound, or fix unfix variables
-Includes updates to ParameterSweepReader and tests through out.

### Legal Acknowledgement


watertap/tools/parameter_sweep/parameter_sweep.py
```diff
@@ -722,7 +775,7 @@ def parameter_sweep(self, *args, **kwargs):
         pass
 
 
-class ParameterSweep(_ParameterSweepBase, _ParameterSweepParallelUtils):
+class ParameterSweep(_ParameterSweepBase, _ParameterSweepParallelUtils):  #
```
k1nshuk:
```suggestion
class ParameterSweep(_ParameterSweepBase, _ParameterSweepParallelUtils):
```

bknueven (approved):

k1nshuk (approved):
Can be merged in after adding the docs.

k1nshuk (approved):
Looks good to me. Thanks for adding the documentation.

avdudchenko:
This is ready for merge.

Issue #1239, opened 2023-12-06
Input based `build_options` in UI validated by `class`


hunterbarber:
### Description

The UI supports build options that are validated from selecting a value from a drop down list. This is good where the config options are only a small fixed set. For the GAC model though, the flowsheet is generalized for any water and solute system. The name and properties (molecular weight, etc.) of the "solute" should be constructed during the build for the most standard solve and ability to integrate the specific name into the input/output results. I think the possible implementation provided is one such idea to make this possible.

### Motivation

Advancing robustness and options of UI.

### Possible Implementation

So I started testing the implementation of a `var_type_allowed` in a addition to the existing `values_allowed` option. In `var_type_allowed` the `class` could be specified ('str' or 'float') that is validated before the `values_allowed`. Then either or both options are to be specified, and raise an error if neither is specified in the `build_options`. However this also requires the drop down box to change to a input box when `values_allowed` is empty (input your own string) in the UI which was out of my expertise.

hunterbarber:
Note, is unnecessary with other changes proposed in #1238. It may still be a nice future feature so I will change to low priority but not close this for now.

hunterbarber:
@dangunter @MichaelPesce Is this closed as of #1270, or still pending and recorded in UI [#98](https://github.com/watertap-org/watertap-ui/issues/98) that this is duplicate? Just trying to clean up and properly link PRs to issues I'm assigned/authored.

MichaelPesce:
@hunterbarber yup this is closed in #1270 (and UI PR [#104](https://github.com/watertap-org/watertap-ui/pull/104)) as well as recorded in UI [#98](https://github.com/watertap-org/watertap-ui/issues/98).

hunterbarber:
Closed in [#1270](https://github.com/watertap-org/watertap/pull/1270) (and UI PR [#104](https://github.com/watertap-org/watertap-ui/pull/104))

PR #1270, opened 2024-01-16
Add functionality for input based model options

MichaelPesce:
## Fixes/Resolves:

Addresses issue #1239 and UI issue https://github.com/watertap-org/watertap-ui/issues/98

## Summary/Motivation:

Allow for flowsheets to have model options that take string, integer, or float types as the value.

- Add additional options for ModelOption class in fsapi

### Legal Acknowledgement


watertap/ui/fsapi.py
```diff
@@ -611,6 +631,14 @@ def __init__(
             self.add_action("diagram", None)
 
         self._actions["custom_do_param_sweep_kwargs"] = custom_do_param_sweep_kwargs
+        self.fs_exp.add_option(
+            name="NumParallelWorkers",
+            display_name="Number of multi-processing workers",
```
avdudchenko: It might be more clear to say "Number of multi-processing workers for sensitivity analysis" 

right now when there are build options, we don't have a way to do any groupings. Might be nice if we had a groups for build options like we do for flowsheet inputs (e.g. flowsheet model build options, analysis options etc..)

watertap/ui/fsapi.py
```diff
@@ -218,11 +220,29 @@ def validate_description(cls, v, values):
     @classmethod
     def validate_value(cls, v, values):
```
avdudchenko, line 224: If I read this right, we can allow users select an option from a list, or enter a string, int or float?

avdudchenko (approved):
LGMT - my comments can be addressed in future PRs.

watertap/ui/fsapi.py
adam-a-a: That's true according to @MichaelPesce

adam-a-a (approved):
LGTM

watertap/ui/fsapi.py
adam-a-a: I think @MichaelPesce mentioned that he would exclude this bit from the PR

PR #1271, opened 2024-01-17
Fix DOF in IX costing function for single use resin and scale pumping power

kurbansitterley:
## Summary/Motivation:

There was a remaining degree of freedom in the IX costing model when using the ``single_use`` regenerant configuration. This was due to the exclusion of costing the backwashing/rinsing tank (i.e., ``capital_cost_backwash_tank`` was unconstrained for that configuration). Also added backwash/rinsing pumping power to the ``single_use`` regenerant configuration (if we are costing the tank, we need to include the pumping costs also). 

This led to the realization that the model was assuming the main pump, backwash pump, rinsing pump, and regeneration pump were all pumping their respective flows for the entirety of the duration of the cycle time ``t_cycle``. In reality, they would only be pumping during the time allotted. So, this PR scales all the pumping power Expressions by the fraction of time they are running. For example, ``bw_pump_power`` is scaled by ``t_bw / t_cycle``. While this changed the amount of power used, it resulted in pretty minor changes to overall cost.

- add backwashing tank costs if regenerant == single_use
- add backwashing and rinsing pumping costs if regenerant == single_use
- scale pumping power required for each stage of IX cycle to the fraction of time for that stage 
- make necessary changes to test file

### Legal Acknowledgement


hunterbarber (approved):
Pretty straightforward, LGTM

docs/technical_reference/unit_models/ion_exchange_0D.rst
```diff
@@ -455,10 +455,11 @@ disposal costs are calculated as a function of the annual mass of resin replaced
 Where :math:`M_{res}` is the resin mass for a single bed and :math:`\dot{v}_{regen}` is the volumetric flow of regenerant solution. If ``hazardous_waste`` is set to ``False``,
 :math:`C_{op,haz} = 0`
 
-The total energy consumed by the unit is the summation of the power required for each of the booster pump, backwashing pump, regeneration pump, and rinsing pump:
+The total energy consumed by the unit is the summation of the power required for each of the booster pump, backwashing pump, regeneration pump, and rinsing pump. Each is scaled
```
adam-a-a, line 458: One extremely subtle thing I noticed was in your pump power equations, you essentially are multiplying `rho*g*h` with `Q` divided by efficiency. However, instead of using head directly (`h`), you have some constant (0.7ish) multiplied by pressure drop in psi, leading me to believe your constant (assuming it's a conversion factor) multiplied by pressure drop equates to head added by the pump. 

Not necessarily anything wrong here--I just wanted to make sure I am interpreting correctly and looking for a little clarification.

adam-a-a (approved):
LGTM - I just had one question about your power equations but nothing that should hold up the PR.

docs/technical_reference/unit_models/ion_exchange_0D.rst
kurbansitterley: You are correct it is a conversion factor, was used before it was pointed out to me that you can convert the units directly to kW...

The conversion is not used anymore in the model equations (instead just directly using ``pyunits.convert``), but are still in the docs. I will fix- thanks for catching that.

PR #1272, opened 2024-01-17
Adding spiral-wound vs flat-sheet config options to RO and OARO

zacharybinger:
## Fixes/Resolves:

This PR adds the option to define the ModuleType for RO and OARO unit models to more accurately represent the geometry of the membranes. Spiral-wound membranes are folded, creating an envelope, which results in the membrane channel width being 1/2 the size compared to a flat-sheet membrane of equivalent area.

- Add ModuleType config option
- Changes constraint calculation for membrane area based on ModuleType

### Legal Acknowledgement


watertap/unit_models/tests/test_reverse_osmosis_0D.py
```diff
@@ -33,7 +33,7 @@
     PressureChangeType,
 )
 
-from watertap.unit_models.reverse_osmosis_base import TransportModel
+from watertap.unit_models.reverse_osmosis_base import TransportModel, ModuleType
```
adam-a-a, line 27: I think you should add some logic relating the ModuleType config to FrictionFactor config. You can start this by adding a test that intentionally sets the two to different values (when we know they should be the same) and see what you get. After that experiment, you could consider either replacing the friction factor config (partially or completely) or ensuring the user couldn't specify an incorrect combo of the two config options.

watertap/unit_models/reverse_osmosis_base.py
```diff
@@ -287,10 +288,24 @@ def _add_area(self, include_constraint=True):
...
+                @self.Constraint(doc="Total Membrane area")
+                def eq_area(b):
+                    return b.area == b.length * 2 * b.width
+
+            else:
+                raise ConfigurationError(
+                    "Unsupported membrane module type: {}".format(
+                        self.config.membrane_module_type
+                    )
+                )
```
adam-a-a: SHould add this modification to the technical reference

watertap/unit_models/tests/test_reverse_osmosis_0D.py
adam-a-a: @zacharybinger alternatively, since you're strictly trying to give an alternative for the effective area calculation, maybe instead of what I suggest above, you can just rename this new config option to something more specific like `ModuleAreaCalculation`. That way, it doesn't seem strange if I want to specify `spiral_wound` for my area but for whatever reason want to maintain the `flat_sheet` option for friction factor.

Does that make sense?

adam-a-a (changes requested):
I'm glad you're bringing this change in. I think we just need a couple of simple changes to merge:
1) Rename the config option to something specific to membrane/module **area** to circumvent the need to sort out logic between seemingly contradictory config options.
2) Add the new option to relevant technical references.

watertap/unit_models/tests/test_reverse_osmosis_0D.py
zacharybinger: @adam-a-a Personally I lean towards just having the one config option for `ModuleType` and having it apply to both the area calculation and the friction factor correlations. What do you think?

adam-a-a: I was originally leaning in that direction as well, but open to hear alternative ideas from others (e.g., @ElmiraShamlou @luohezhiming @TimBartholomew @avdudchenko )

adam-a-a:
@zacharybinger I may have missed this, but do you check length and width solutions in RO and OARO tests? If not, I would be interested seeing just a couple tests checking width and length solutions comparatively between `flat_sheet` and `spiral_wound`.

Additionally, I just thought of one drawback to consolidating the config options to just module_type. Say I want to add a custom relationship for friction_factor, which is perfectly reasonable. Under the consolidated approach, I would need to define a new module_type option, rewrite the membrane area calculation, which in most cases will be the same, then provide my new friction factor relationship. Does that make sense? In other words, it might introduce unnecessary redundancy for future/customized enhancement efforts. I don't want to make you go back and forth on changes, so maybe we should talk it through a little more to arrive at the best approach.

watertap/unit_models/tests/test_reverse_osmosis_1D.py
```diff
@@ -1191,8 +1190,8 @@ def test_friction_factor_spiral_wound(self):
         )
 
         assert value(flow_mass_inlet) == pytest.approx(1.0, rel=1e-3)
-        assert value(flow_mass_retentate) == pytest.approx(0.8981, rel=1e-3)
-        assert value(flow_mass_permeate) == pytest.approx(0.1011, rel=1e-3)
+        assert value(flow_mass_retentate) == pytest.approx(0.9488, rel=1e-3)
+        assert value(flow_mass_permeate) == pytest.approx(5.122e-2, rel=1e-3)
```
ElmiraShamlou: It appears that the permeate production is halved as a result of the current changes, which seems incorrect. In the revised code, the spacer width is equated to the effective width for permeate production calculations. It's crucial to differentiate between the physical dimensions of the spacer and the effective width used in mass transfer term computations. Although the membrane is folded, there is flux to the permeate spacer from both sides of the leaves.
The 0D models, where mass transfer is calculated using variable area, is not affected by this change. In contrast, the 1D models calculate mass transfer per unit length based on width variables, which overlooks the dual-side flux in folded membranes

ElmiraShamlou (changes requested):

watertap/unit_models/reverse_osmosis_base.py
```diff
@@ -287,10 +287,24 @@ def _add_area(self, include_constraint=True):
         )
 
         if include_constraint:
-            # Membrane area equation
-            @self.Constraint(doc="Total Membrane area")
-            def eq_area(b):
-                return b.area == b.length * b.width
+            if self.config.module_type == ModuleType.flat_sheet:
```
ElmiraShamlou, line 290: Following up on my earlier comment, I'm not sure it's a good idea to group total membrane area based on membrane type and formulate it based on spacer width (overall leaf length), for these reasons:

1) Error in mass transfer term Calculations: As previously mentioned, this approach could lead to inaccuracies in mass transfer term (permeate production) calculations. To ensure precision, mass transfer term calculations would need to be categorized by module type and adjusted to consider 2*width for spiral-wound modules, which adds complexity.

2) Incomplete Picture of Membrane Dimensions: Reporting the width as the total leaf's length doesn’t provide a complete overview of the membrane dimensions.

3) Complexity in Future Custom Friction Factor Calculations: As @adam-a-a  pointed out, this approach also complicates future calculations for custom friction factors.

ElmiraShamlou (commented):
Great work on this PR, it's an important step towards better cross flow area calculations. Along with my previous comments, here are some additional notes for consideration, particularly regarding pressure drop and module area calculation:

1) Active (effective) membrane area in Spiral Wound Membranes: 
In spiral wound membranes, not all the membrane area is available for permeation due to glue lines and dead zones, especially in each leaf's outer membrane areas (further from the central permeate collection tube). The extent of this inactive area varies with the number of leaves, winding geometry, and central tube diameter. Consequently, for the same total membrane area, different winding geometry would yield different effective areas available for permeation, influencing both pressure drop and permeate production. 
Since our current models don't calculate these details, a possible solution could be introducing an 'effective membrane area coefficient' in the mass transfer term calculations for both 0D and 1D models. This coefficient, defaulted at 1, could be adjusted based on manufacturer data to better represent actual spiral wound modules.

2) Permeate Side Pressure Drop Calculations: 
Our current pressure drop calculations might not fully capture the increased pressure drop in the permeate channel with spiral and cross-flow paths, especially for OARO models.

watertap/unit_models/tests/test_reverse_osmosis_0D.py
ElmiraShamlou: Building on my previous comments, I'm inclined to suggest that we should not categorize total membrane area and available membrane area for flux based on module type. A more effective approach might be to focus on the differences attributed to feed spacer width in our cross-sectional area calculations. This could potentially offer a more straightforward and universal method, applicable across various module types without the complexity of categorizing by each type.

adam-a-a:
@zacharybinger just checking in to see where you are on this and what your thoughts are in response to the comments

watertap/unit_models/tests/test_reverse_osmosis_1D.py
zacharybinger: So I found that the flux remained the approx the same, but the permeate flow rate almost halved. It turned out to be the constraint used in the mass transfer calculations multiplied the flux times width then by length, instead of just the differential area itself. This has been corrected in [bd06f48](https://github.com/watertap-org/watertap/pull/1272/commits/bd06f485db28fa6d1fe060a8275207eca862fac0)

adam-a-a: @zacharybinger want to point out something that isn't immediately obvious regarding the the 1d controlvolume in IDAES-- the mass transfer term has units of mass/unit length (e.g., kg/m), which explains why the related equations n RO1d were formulated the way they were. It looks like you figured that bit out, but just thought I'd clarify for anyone who might be reading this through.

zacharybinger:
So one reason I held off in including this PR in the March release was I wanted to double check the flux values in the tests. The previous implementation didn't multiply the flux by the differential area to calculate mass flow across a differential element. Instead did the multiplication in two steps:
```
def eq_connect_mass_transfer(b, t, x, p, j):
            return (
                b.permeate_side[t, x].get_material_flow_terms(p, j)
                == -b.feed_side.mass_transfer_term[t, x, p, j] * b.length / b.nfe
            )
```

```
def eq_mass_flux_equal_mass_transfer(b, t, x, p, j):
            return (
                b.flux_mass_phase_comp[t, x, p, j] * b.width
                == -b.feed_side.mass_transfer_term[t, x, p, j]
            )
```

So the length was accounted for in `eq_connect_mass_transfer` and the width was accounted for in `eq_mass_flux_equal_mass_transfer`. This caused an issue due to the new width implementation resulting in the flux being reduced by half since the width is halved, even though the total area remained the same. 

`b.area == b.length * 2 * b.width` therfore `flux * b.length * b.width != flux * b.area`

So instead I updated the equations to use the area instead:
```
def eq_connect_mass_transfer(b, t, x, p, j):
            return (
                b.permeate_side[t, x].get_material_flow_terms(p, j)
                == -b.feed_side.mass_transfer_term[t, x, p, j] * b.length / b.nfe
            )
```

```
def eq_mass_flux_equal_mass_transfer(b, t, x, p, j):
            return (
                b.flux_mass_phase_comp[t, x, p, j] * (b.area / b.length)
                == -b.feed_side.mass_transfer_term[t, x, p, j]
            )
```

membrane_width_fix

So you can see that now the water flux is correctly calculated and is slightly higher than the previous flat-plate value. This is explained by 1/2 channel width -> higher fluid velocity -> lower concentration polarization -> lower osmotic pressure at the membrane interface.

watertap/unit_models/tests/test_osmotically_assisted_reverse_osmosis_1D.py
```diff
@@ -87,7 +86,7 @@ def test_config():
     assert m.fs.unit.config.pressure_change_type == PressureChangeType.fixed_per_stage
     assert m.fs.unit.feed_side._flow_direction == FlowDirection.forward
     assert m.fs.unit.permeate_side._flow_direction == FlowDirection.backward
-    assert m.fs.unit.config.friction_factor == FrictionFactor.flat_sheet
+    assert m.fs.unit.config.module_type == ModuleType.flat_sheet
```
ElmiraShamlou: @zacharybinger I suspect the same issue with flux calculation might also affect OARO 1D, though it's not currently evident from the OARO 1D tests since they don't cover Spiral wound scenarios. I suggest verifying that OARO Spiral wound operates as intended and implementing any required adjustments

zacharybinger: Fixed in the meeting, validated the results in the following figure, and added a OARO spiral-wound test
membrane_width_fix_oaro

zacharybinger: Fixed in meeting, validated with the following figure, and added an OARO spiral-wound test
membrane_width_fix_oaro

watertap/unit_models/reverse_osmosis_base.py
zacharybinger: added

watertap/unit_models/tests/test_osmotically_assisted_reverse_osmosis_1D.py
adam-a-a: Is the figure above for OARO? Did you also check the permeate channel pressure drop and osmotic pressure? I don't expect any dramatic findings but just thought I'd ask.

adam-a-a:
@zacharybinger so it looks like that one test that was failing for dye_desalination (and maybe 1dRO) was resolved by bringing the denominator to the other side of the equation. Is that correct?

docs/technical_reference/unit_models/reverse_osmosis_1D.rst
```diff
@@ -4,6 +4,7 @@ This reverse osmosis (RO) unit model
    * is 1-dimensional
    * supports a single liquid phase only
    * supports steady-state only
+   * supports flat-sheet and spiral-wound module designs
```
adam-a-a, line 7: I am guessing the 1d documentation references the 0d documentation for covering equations, hence the exclusion of the alternative equations for effective membrane area. Correct?

watertap/core/membrane_channel_base.py
```diff
@@ -83,16 +93,6 @@ class PressureChangeType(Enum):
     calculated = auto()
 
 
-class FrictionFactor(Enum):
```
adam-a-a, line 86: My only hesitation about complete removal of a friction_factor config option is that, although it appears cleaner now with your implementation, what happens in the future if someone wants to add a custom friction_factor relationship? 

On the one hand, we could consider adding a new ModuleType option--which would require provision of a new membrane area and friction factor calculation. 

On the other hand, if we wanted to just maintain a spiral wound type for the membrane area determination and only wanted to provide a custom friction factor relationship, we couldn't do that with the new changes. 

There might be other options that I didn't cover, but figured this might be the last issue to reflect on before removing separate friction_factor options completely.

Another thing we could consider is keeping friction_factor with modified options. FOr example, we could have a "default" or "default_by_module_type") option for friction_fractor configuration that would simply use the relationship associated with the module_type selected; i.e., the user would not need to specify "flat_sheet" or "spiral_wound" explicitly for setting up the friction_factor config. Additional config options for the friction_factor config would consider alternative relationships that might be added later on, or custom relationships provided on-the-fly.

adam-a-a: Another case where I think allowing for flexibility in adding in friction factor relationships is for counter or co-current modules like OARO for example. We may want to consider different friction factor equations for each side of the membrane (e.g., higher pressure drop on permeate side embodied by some equation that differs from the one used on the feed side). Keeping the friction factor configuration option would allow the user to make more detailed decisions. That said, a flaw in the existing implementation (before your changes) is that I think we only had a single friction factor config option intended for the feed side and didn't consider alternative relationships for the permeate side (when relevant, i.e., in the case of OARO).

docs/technical_reference/unit_models/reverse_osmosis_1D.rst
zacharybinger: Yes, since the membrane area equations weren't originally in the 1D docs I kept it that way

watertap/unit_models/tests/test_osmotically_assisted_reverse_osmosis_1D.py
zacharybinger: Yeah permeate side checked out

watertap/core/membrane_channel_base.py
zacharybinger: Well we could keep the friction_factor option purely as an option for the constraint used for pressure loss. The module_type option isn't meant to have anything to do with the pressure loss equations used, it's just to assure correct channel dimensions are used.

We can set the friction_factor option to a default, but then we're asking the user to implement their custom constraints on the unit model level and not their flowsheet level

adam-a-a: > The module_type option isn't meant to have anything to do with the pressure loss equations used, it's just to assure correct channel dimensions are used.

True, but currently, we happen to have two friction factor relationships that correspond to a specified flat sheet correlation and another corresponding to a classic spiral wound correlation. This is the only reason why I mentioned possibly having "default" depend on module_type.

ElmiraShamlou: I support setting default_by_module_type as the default option for the friction_factor config

ElmiraShamlou: > On the one hand, we could consider adding a new ModuleType option--which would require provision of a new membrane area and friction factor calculation.

I believe it's appropriate to define new area calculations for additional module types. Considering other module types like tubular, capillary, and hollow fiber, all require area calculations based on the diameter and number of the tubes or fibers, it makes sense to introduce a new area calculation method tailored to these specifications.

ElmiraShamlou (approved):
Approved. Great work!

watertap/core/__init__.py
```diff
@@ -16,7 +16,6 @@
     ConcentrationPolarizationType,
     MassTransferCoefficient,
     PressureChangeType,
-    FrictionFactor,
```
adam-a-a: Should you bring this back?

watertap/core/membrane_channel_base.py
```diff
@@ -71,6 +71,16 @@ class TransportModel(Enum):
     SKK = auto()
 
 
+class ModuleType(Enum):
+    """
+    flat_sheet:
+    spiral_wound:
+    """
```
```suggestion
    flat_sheet: flat-sheet module configuration
    spiral_wound: spiral-wound module configuration
    """
```

watertap/core/membrane_channel_base.py
```diff
@@ -89,8 +99,7 @@ class FrictionFactor(Enum):
     spiral_wound: Darcy's friction factor correlation by Schock & Miquel, 1987
```
adam-a-a: Can replace this doc string to explain how default_by_module_type will revert to either flat_sheet correlation by G&H or spiral_wound correlation by S&M

watertap/core/membrane_channel_base.py
```diff
@@ -303,19 +330,19 @@ class FrictionFactor(Enum):
...
-        **default** - ``FrictionFactor.flat_sheet`` 
+        **default** - ``FrictionFactor.default_by_module_type`` 
 
     .. csv-table::
         :header: "Configuration Options", "Description"
 
-        "``FrictionFactor.flat_sheet``", "Friction factor correlation for flat-sheet membrane modules"
-        "``FrictionFactor.spiral_wound``", "Friction factor correlation for spiral-wound membranes"
+        "``FrictionFactor.default_by_module_type``", "Friction factor correlation that is specific to the supported membrane modules type"
+        "``FrictionFactor.custom``", "Custom user-defined friction factor correlations"
```
```suggestion
        "``FrictionFactor.custom``", "Custom user-defined friction factor correlations"
```
I suppose we should delete for the time being.

watertap/core/membrane_channel_base.py
zacharybinger: Done!

watertap/core/membrane_channel_base.py
zacharybinger: Done!

watertap/core/__init__.py
zacharybinger: Done!

watertap/unit_models/osmotically_assisted_reverse_osmosis_base.py
```diff
@@ -340,15 +343,23 @@ def _add_area(self, include_constraint=True):
             )
 
         if include_constraint:
-            if not hasattr(self, "eq_area"):
```
adam-a-a, line 343: Why delete this check to make sure `eq_area` isn't already defined on the model?

adam-a-a: I assume we originally had this here because there was potential for a user to erroneously replicate this constraint without realizing, which would cause issues with the Pyomo model.

watertap/unit_models/tests/test_osmotically_assisted_reverse_osmosis_0D.py
```diff
@@ -1395,7 +1395,7 @@ def test_friction_factor_spiral_wound(self):
             concentration_polarization_type=ConcentrationPolarizationType.calculated,
             mass_transfer_coefficient=MassTransferCoefficient.calculated,
             pressure_change_type=PressureChangeType.calculated,
-            friction_factor=FrictionFactor.spiral_wound,
```
adam-a-a: I assume with the new config option, friction factor is still being tested since it will change accordingly with module type.

adam-a-a (commented):
Just a couple more comments and then I think this is good to go.

watertap/unit_models/osmotically_assisted_reverse_osmosis_base.py
zacharybinger: That might be an error caused by copy and paste. I had no reason to remove that but it's not in the RO_base, so likely I copied the structure over and erroneously removed that check

watertap/unit_models/tests/test_osmotically_assisted_reverse_osmosis_0D.py
zacharybinger: Correct. The spiral-wound module type is selected in this test

watertap/unit_models/osmotically_assisted_reverse_osmosis_base.py
zacharybinger: Placed it back now

adam-a-a (approved):
LGTM! Thanks for persevering!

PR #1232, opened 2023-12-02
CSV-driven variable exports

dangunter:
## Fixes/Resolves:

(replace this with the issue # fixed or resolved, if no issue exists then a brief statement of what this PR does)

## Summary/Motivation:

Allow variable exports to be listed in a CSV file instead of (or in addition to) Python API calls.
This is a more compact way to export a lot of variables.
Does not alter or invalidate the existing Python API export method.

## Changes proposed in this PR:
- Changes in watertap.ui.fsapi
    - FlowsheetExports: new methods `from_csv()` and `to_csv()`
- Changes in docs
   - Added the `from_csv()` to the how-to on the API. Streamlined the page too.
- Changes in configuration
   - in _setup.py_, added `importlib-resources` for Python < 3.11 support for the features that "find" the CSV datafiles in the package directory
   
### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #1277, opened 2024-01-23
Unit Model Test Harness

MarcusHolly:
## Fixes/Resolves:

Issue #66

- Re-create unit model test harness from PR #306 
- No longer uses the report feature to test the model solution

### Legal Acknowledgement


watertap/unit_models/tests/test_unit_test_harness.py
```diff
@@ -0,0 +1,111 @@
...
+            "number_unused_variables": 0,
+        }
+        self.unit_solution = {
+            (
+                "feed_side",
+                "properties_in",
+                "flow_mass_phase_comp",
+                ("Liq", "H2O"),
+            ): 0.965,
+            (
```
MarcusHolly: @TimBartholomew I believe we wanted it to look like this, but I could never get it to work.
` self.unit_solution = {(m.fs.unit.feed_side_properties_in.flow_mass_phase_comp["Liq", "H2O"]): 0.965, ... }`
            

My current workaround is functional, but far from ideal imo...

lbianchi-lbl: @MarcusHolly do you remember what was the error you were getting when trying to use the Pyomo object as a dict key directly? I've tried with `m.fs.unit.inlet.flow_mass_phase_comp[0, "Liq", "H2O"]` and I'm getting `TypeError: unhashable type: '_GeneralVarData'`.

I've quickly tried to check if there's a way to "drill down" a Pyomo object from a string (sort of like a "deep `getattr`) e.g. in this case

```py
obj_1 = m.fs.unit.inlet.flow_mass_phase_comp[0, "Liq", "H2O"]
obj_2 = m["fs.unit.inlet.flow_mass_phase_comp[0, 'Liq', 'H2O']"]
assert obj_1 is obj_2
```

I wasn't able to find much but it's entirely possible that this is due to my very superficial knowledge of Pyomo, so I'll defer to @bknueven for a proper answer.

If this functionality is not natively available, we should still be able to find an alternative so reduce the complexity of the current solution based on `if len(key) ...`. For example:

1. In `self.configure()`, define `self.unit_solution` using string keys:
  ```py
  self.unit_solutions = {
      "m.fs.unit.inlet.flow_mass_phase_comp[0, 'Liq', 'H2O']": 0.965,
      ...
  }
  ```
2. When testing, use `eval()` to "extract" the Pyomo object back from the `ConcreteModel` instance:
  ```py
  for key, expected_val in self.unit_solutions.items():
      pyo_obj = eval(key, {"m": self.model})
      assert pytest.approx(val(pyo_obj), expected_val, ...)
  ```

MarcusHolly: Well part of the complexity is that the unit model test harness is intended to be applied to all unit models, so we want to avoid hard-coding in any `m.fs.unit...` inside of `unit_test_harness.py` because the list would be almost infinitely long. I'm just using this RO example as the first implementation, but, ideally, this should work for all unit models.

lbianchi-lbl: I see what you mean, and I think this is compatible with what I was proposing. The model-specific keys would be defined only in the derived class, leaving `unit_test_harness.py` completely model-agnostic.

In other words, in the model-specific file `test_my_model.py` you'd have

```py
from watertap.unit_models.tests.unit_test_harness import UnitTestHarness

class TestForMyModel(UnitTestHarness):
    def configure(self):
        ...
        self.unit_solutions = {
             "m.fs.unit.inlet.flow_mass_phase_comp[0, 'Liq', 'H2O']": 0.965,
             # rest of the model-specific keys
         ...
     }
```

In the model-agnostic `watertap/unit_models/tests/unit_test_harness.py`

```py
class UnitTestHarness:

    def test_solution(self):
    ...
    for key, expected_val in self.unit_solutions.items():
        pyo_obj = eval(key, {"m": self.model})
        assert pytest.approx(val(pyo_obj), expected_val, ...)
```

As you can see, the base `UnitTestHarness` class doesn't need to know anything about the specific model being tested (apart from assuming that `self.model` refers to the `ConcreteModel` instance).

MarcusHolly: I've tried my hand at this method, but I definitely missed something because I'm seeing ` AttributeError: 'TestReverseOsmosis0D_default' object has no attribute 'model'`

watertap/unit_models/tests/test_unit_test_harness.py
```diff
@@ -0,0 +1,97 @@
...
+# Get default solver for testing
+solver = get_solver(options={"bound_push": 1e-8})
+
+
+# -----------------------------------------------------------------------------
+class TestReverseOsmosis0D_default(UnitTestHarness):
+    def configure(self):
+        self.model = ConcreteModel()
+        # build unit
+        m = ConcreteModel()
```
lbianchi-lbl: I believe that this would cause `self.model` and `m` to refer to two completely separate instances.

Python actually lets you assign multiple names (left-hand-side) to the same object (right-hand-side) in one statement, which I frequently use for these cases:

```suggestion
        m = self.model = ConcreteModel()
```

watertap/unit_models/tests/test_unit_test_harness.py
```diff
@@ -0,0 +1,97 @@
...
+
+        self.unit_model_block = m.fs.unit
+        self.unit_statistics = {
+            "number_variables": 121,
+            "number_total_constraints": 109,
+            "number_unused_variables": 0,
+        }
+        self.unit_solutions = {
+            "m.fs.feed_side.properties_in[0].flow_mass_phase_comp[Liq, H2O]": 0.965,
+            "m.fs.feed_side.properties_in[0].flow_mass_phase_comp[Liq, NaCl]": 0.035,
```
lbianchi-lbl: If we're using `eval()`, then the indices in these strings need to have their own quotes as well:

```suggestion
            "m.fs.feed_side.properties_in[0].flow_mass_phase_comp['Liq', 'H2O']": 0.965,
            "m.fs.feed_side.properties_in[0].flow_mass_phase_comp['Liq', 'NaCl']": 0.035,
```

watertap/unit_models/tests/unit_test_harness.py
```diff
@@ -167,79 +167,82 @@ def test_unit_solution(self, frame_unit):
...
-                elif not pytest.approx(val, rel=1e-3) == value(var):
-                    raise UnitValueError(
-                        "Variable {v_name} is expected to have a value of {val} +/- 0.1%, but it "
-                        "has a value of {val_t}. \nUpdate unit_solution in the configure function "
-                        "that sets up the UnitRegressionTest".format(
-                            v_name=key[0], ind=key[1], val=val, val_t=value(var)
-                        )
-                    )
+        for key, val in solutions.items():
+            pyo_obj = eval(key, {"m": self.model})
```
lbianchi-lbl: I'm not sure what's the relationship between the `UnitTestHarness` instance (i.e. `self` in this scope) and `blk._test_objs`.

If attributes set in the body of the model-specific subclasses are supposed to be accessed via `blk._test_objs`, then you might need to use `{"m": blk._test_objs.model}` in the second argument of the `eval()` call.

lbianchi-lbl: If this still doesn't work, I might have to take a more in-depth look into this - there are a number of subtleties involved when using pytest `Test` classes (essentially, pytest only uses them as namespaces for test functions and most of the usual class instance functionality doesn't work as one would expect. See https://docs.pytest.org/en/stable/getting-started.html#group-multiple-tests-in-a-class for more info).

watertap/unit_models/tests/test_unit_test_harness.py
bknueven: I would recommend instead that we just store the string location on the model, e.g.
```suggestion
            "fs.feed_side.properties_in[0].flow_mass_phase_comp[Liq, H2O]": 0.965,
            "fs.feed_side.properties_in[0].flow_mass_phase_comp[Liq, NaCl]": 0.035,
```

And then use the `find_component` method on the model object below in place of `eval` (detailed below).

watertap/unit_models/tests/unit_test_harness.py
```diff
@@ -0,0 +1,268 @@
...
+            )
+        results = opt.solve(blk)
+
+        # check solve
+        assert len(list(iscale.badly_scaled_var_generator(blk, zero=1e-8))) == 0
+        assert_optimal_termination(results)
+
+        # check results
+        for key, val in solutions.items():
+            pyo_obj = eval(key, {"m": self.model})
```
bknueven:
```suggestion
            pyo_obj = self.model.find_component(key)
```
Per my suggestion above.

bknueven (commented):
Let's not use `eval()`, ever. `Block.find_component` is the Pyomo-thonic way to reference a pyomo object via string.

watertap/unit_models/tests/unit_test_harness.py
MarcusHolly: Thanks for the help @lbianchi-lbl! I'll take a look at these at some point tomorrow.

lbianchi-lbl:
> Let's not use `eval()`, ever. `Block.find_component` is the Pyomo-thonic way to reference a pyomo object via string.

Thanks @bknueven, I knew there was a better way of doing this ("friends don't let friends use `eval()`" :sweat_smile: ). 

@MarcusHolly, I'd suggest to go with what Ben is suggesting since it seems to have the same functionality as my proposal, but a much more "proper" implementation.

Unrelated, but @dangunter might also find `Block.find_component()` useful (if he isn't using it already) for the `fsapi` text-based definitions (e.g. #1232).

watertap/unit_models/tests/unit_test_harness.py
```diff
@@ -0,0 +1,173 @@
...
+            )
+        results = opt.solve(blk)
+
+        # check solve
+        assert len(list(iscale.badly_scaled_var_generator(blk, zero=1e-8))) == 0
+        assert_optimal_termination(results)
+
+        # check results
+        for key, val in solutions.items():
+            pyo_obj = self.model.find_component(key)
```
MarcusHolly: @bknueven or @lbianchi-lbl This is currently returning `None` for each key... any thoughts on why?

watertap/unit_models/tests/unit_test_harness.py
```diff
@@ -168,101 +168,6 @@ def test_unit_solutions(self, frame_unit):
...
-        #     elif not pytest.approx(val, rel=1e-3) == value(var):
-        #         raise UnitValueError(
-        #             "Variable {v_name} is expected to have a value of {val} +/- 0.1%, but it "
-        #             "has a value of {val_t}. \nUpdate unit_solution in the configure function "
-        #             "that sets up the UnitRegressionTest".format(
-        #                 v_name=v_name, ind=ind, val=val, val_t=value(var)
-        #             )
-        #         )
+            pyo_obj = self.model.find_component(key)
+            assert pytest.approx(value(pyo_obj), rel=1e-5) == val
```
bknueven: I think we should think through a slick way to supply relative and absolute tolerances which are dependent on the value, while not overly burdening the test harness implementer if it doesn't matter for their model.

In particular, we have a lot of values close to `0` where a relative tolerance of `1e-5` is too tight.

lbianchi-lbl: Off the top of my head:

- `self.unit_solutions` values can be either:
  - A single number, in which case a default (per-class?) tolerance is used
  - A tuple `(val, tol)` where the value is specified together with a custom tolerance

The added complication for the `UnitTestHarness` code is... tolerable (pun clearly intended), e.g.:

```py
for key, val in solutions.items():
    expected, tol = val if isinstance(val, tuple) else val, self.default_tol
    assert pytest.approx(value(pyo_obj), tol=tol) == expected
```

lbianchi-lbl:
I'm not sure of the reasons and/or possible implications, but something that caught my eye: it looks like the rich assertion context usually generated by pytest is missing. For instance, this looks like a legitimate assertion failure (i.e. the code that generates the values to compare works correctly, but the value itself is off-spec):

image

My concern is that this could be a fundamental limitation of the current implementation: the fact that the code that actually raise the exception is not in the test class somehow throws off the ~~forbidden magic~~ mechanism used by pytest to rewrite assertions. However, this is just a guess. I'll try to look into it.

watertap/unit_models/tests/unit_test_harness.py
MarcusHolly: is `pyo_obj` still `self.model.find_component(key)` in this case? Assuming so, I'm still seeing the same AssertionError where `pyo_obj` is `None`.

Also, the first tol in `assert pytest.approx(value(pyo_obj), tol=tol) == expected` has to be either `rel` or `abs` I think.

watertap/unit_models/tests/unit_test_harness.py
bknueven: This happens if the look-up fails ... so maybe the name was misspelled, or we have the wrong Pyomo model here.

bknueven: In particular, shouldn't we be looking for the component on `blk`? But then what is `self.model`?

MarcusHolly:
Added an example of what the full unit model test files would look like using the unit model test harness (`test_anaerobic_digestor.py`). Feedback is welcome on what other aspects of the model should be tested here as well as feedback on how `unit_test_harness.py` is currently set up.

Edit: One thing I definitely need to add to the test is costing...

watertap/unit_models/tests/test_anaerobic_digestor.py
```diff
@@ -163,269 +130,53 @@ def adm(self):
...
+        self.unit_model_block = m.fs.unit
+        self.unit_statistics = {
+            "number_variables": 202,
+            "number_total_constraints": 170,
+            "number_unused_variables": 0,
+        }
+        self.unit_solutions = {
+            "liquid_outlet.pressure[0]": 101325,
+            "liquid_outlet.temperature[0]": 308.15,
+            "liquid_outlet.conc_mass_comp[0, S_I]": 0.328772,
```
TimBartholomew: @MarcusHolly @lbianchi-lbl @bknueven, why don't we make this more similar to the interface of the parameter sweep or GUI where we just provide the variable itself? In other words, what's the benefit of providing the string and then using `find_component`?

For example, we could change the input dictionary to:
```
 self.unit_solutions = {
            m.fs.unit.liquid_outlet.pressure[0]: 101325,
            m.fs.unit.liquid_outlet.temperature[0]: 308.15,
}
```

lbianchi-lbl: AFAIK not all of Pyomo's objects are hashable, which is a requirement for Python objects to be used as dictionary keys (see my comment above: https://github.com/watertap-org/watertap/pull/1277#discussion_r1464145901). I suppose a workaround would be to use a list of tuples instead:

```py
self.unit_solutions = [
    (m.fs.unit.liquid_outlet.pressure[0], 101325),
    (m.fs.unit.liquid_outlet.temperature[0], 308.15),
]
```

The only disadvantage I can see is that the list-of-tuples syntax is somewhat busier visually.

watertap/unit_models/tests/unit_test_harness.py
```diff
@@ -0,0 +1,183 @@
...
+
+        # check results
+        for key, val in solutions.items():
+            pyo_obj = blk.find_component(key)
+            pyo_obj_value = pyo_obj()
+            try:
+                if val == 0:
+                    assert pytest.approx(pyo_obj_value, abs=1e-08) == val
+                else:
+                    assert pytest.approx(pyo_obj_value, rel=1e-03) == val
```
lbianchi-lbl: Two minor suggestions:

- We can move the logic of setting the tolerance to set a kwargs dict instead of having to call `pytest.approx()` twice
- Since we won't get the pytest rich display for failed assertions anyway, we can check the result of the `pytest.approx` comparison as a bool directly instead of having to go through `assert pytest.approx(...) == ` first

```suggestion
            tol = {"abs": 1e-08} if val == 0 else {"rel": 1e-03}
            if not pytest.approx(pyo_obj_value, **tol) == val:
                raise AssertionError(
                    f"{pyo_obj}: Expected {val}, got {pyo_obj_value} instead"
                )
```

watertap/unit_models/tests/test_anaerobic_digestor.py
bknueven: You can have a "dictionary with pyomo objects as keys" by using the built in `ComponentMap`:
```python
from pyomo.environ import ComponentMap

...

self.unit_solutions = ComponentMap()
self.unit_solutions[m.fs.liquid_outlet.pressure[0]] = 100325
```

watertap/unit_models/tests/unit_test_harness.py
andrewlee94: You can also pass both the `abs` and `rel` argument to `pytest.approx` and it will pass as long as one of them is satisfied which is often all you need (and also handles cases where the value is very small but non-zero).

bknueven:
Is the need to have tolerances optional a good use of `dataclass`?

```python
from dataclasses import dataclass
from pyomo.environ import PyomoObject

@dataclass
class TestValue:
    pyo_obj: PyomoObject
    val: float
    abs: float = 0.0
    rel: float = 1e-5
    
...

self.unit_solutions = [
    TestValue(m.fs.liquid_outlet.pressure[0], 100325),
    TestValue(m.fs.unit.liquid_outlet.temperature[0], 308.15, abs=1e-4)
    ...
]
```

or if we use tuples / `ComponentMap`, maybe it can be optionally used and in the check we cast it for ease?

```python
from dataclasses import dataclass

@dataclass
class TestValue:
    val: float
    abs: float = 0.0
    rel: float = 1e-5
    
...

self.unit_solutions = [
   (m.fs.liquid_outlet.pressure[0], 100325),
   (m.fs.unit.liquid_outlet.temperature[0], TestValue(308.15, abs=1e-4),
   ...
]
```

If the user wrote their own class (and specified it somewhere) we could also have the default easily changed.



On a meta note, I do worry a bit that this is trending towards re-investing functionality in `pytest`..

lbianchi-lbl:
> Is the need to have tolerances optional a good use of `dataclass`?
> 
> ```python
> from dataclasses import dataclass
> from pyomo.environ import PyomoObject
> 
> @dataclass
> class TestValue:
>     pyo_obj: PyomoObject
>     val: float
>     abs: float = 0.0
>     rel: float = 1e-5
>     
> ...
> 
> self.unit_solutions = [
>     TestValue(m.fs.liquid_outlet.pressure[0], 100325),
>     TestValue(m.fs.unit.liquid_outlet.temperature[0], 308.15, abs=1e-4)
>     ...
> ]
> ```
> 
> or if we use tuples / `ComponentMap`, maybe it can be optionally used and in the check we cast it for ease?
> 
> ```python
> from dataclasses import dataclass
> 
> @dataclass
> class TestValue:
>     val: float
>     abs: float = 0.0
>     rel: float = 1e-5
>     
> ...
> 
> self.unit_solutions = [
>    (m.fs.liquid_outlet.pressure[0], 100325),
>    (m.fs.unit.liquid_outlet.temperature[0], TestValue(308.15, abs=1e-4),
>    ...
> ]
> ```
> 
> If the user wrote their own class (and specified it somewhere) we could also have the default easily changed.
> 
> On a meta note, I do worry a bit that this is trending towards re-investing functionality in `pytest`..

I like this. I think it should be possible to use `pytest.approx` directly instead of the `TestValue` class (i.e. `pytest.approx` is a regular class constructor, and AFAIK returns normal instances that encode the approx logic in their `__eq__()` method.

So the logic could become 

```py
val = val if isinstance(val, pytest.approx) else pytest.approx(val, rel=1e-03)
```

watertap/unit_models/tests/test_unit_test_harness.py
TimBartholomew: Oh I didn't realize my comment below was already discussed here

watertap/unit_models/tests/unit_test_harness.py
```diff
@@ -0,0 +1,189 @@
...
+
+        # check results
+
+        for key, val in solutions.items():
+            pyo_obj = blk.find_component(key)
+            pyo_obj_value = pyo_obj()
+            if not pytest.approx(pyo_obj_value, abs=1e-08, rel=1e-03) == val:
+                raise AssertionError(
+                    f"{pyo_obj}: Expected {val}, got {pyo_obj_value} instead"
+                )
```
MarcusHolly: Not quite sure how this should be modified if 
```
self.unit_solutions = [
    (m.fs.unit.liquid_outlet.pressure[0], 101325),
    (m.fs.unit.liquid_outlet.temperature[0], 308.15),
]
```
or
```
self.unit_solutions = ComponentMap()
self.unit_solutions[m.fs.unit.liquid_outlet.pressure[0]] = 100325
```
because `blk.find_component` doesn't appear to work in these scenarios. So I'm not sure how we'd get the new `pyo_obj_value`

watertap/unit_models/tests/unit_test_harness.py
```diff
@@ -0,0 +1,186 @@
...
+        assert_optimal_termination(results)
+
+        # check results
+
+        for key, val in solutions.items():
+            try:
+                var_name = str(key)
+                truncated_var_name = var_name[8:]
+                pyo_obj = blk.find_component(truncated_var_name)
+                pyo_obj_value = pyo_obj()
```
MarcusHolly: Not sure how to get around hard-coding `truncated_var_name` because `find_component` won't work if the variable starts with "m.fs.unit"

TimBartholomew: I think this has to do with blk being m.fs.unit, as opposed to just m. I think this is probably fine, it just requires users to create the unit model with the name m.fs.unit. They'd run into troubles if they used any other name.

@bknueven do you think this is fine too? Or do you have any other ideas?

watertap/unit_models/tests/test_anaerobic_digestor.py
```diff
@@ -73,44 +55,9 @@
...
-    assert not m.fs.unit.config.has_holdup
-    assert m.fs.unit.config.material_balance_type == MaterialBalanceType.useDefault
-    assert m.fs.unit.config.energy_balance_type == EnergyBalanceType.useDefault
-    assert m.fs.unit.config.momentum_balance_type == MomentumBalanceType.pressureTotal
-    assert not m.fs.unit.config.has_heat_transfer
-    assert not m.fs.unit.config.has_pressure_change
-    assert not m.fs.unit.config.has_equilibrium_reactions
-    assert not m.fs.unit.config.has_phase_equilibrium
-    assert not m.fs.unit.config.has_heat_of_reaction
-    assert not m.fs.unit.config.has_pressure_change
```
TimBartholomew, line 104: The test harness no longer includes these types of tests, which is testing the default configuration arguments. I'm wondering if that is fine.

If a developer changed the default configuration arguments, then what would happen?
- the solution regression test will likely fail since now it is evaluating a different model
- the developer would then update the solution regression test
- the developer might not create a new test that covers the previous default options (I think this is the main issue). 

However, this could always happen, the developer could fail to test all configuration argument options. The only way we could prevent this test omission is if we somehow ensure all configuration arguments are tested. I'm not sure how we would do this, and its definitely out of scope for this PR. So I think this is fine.

TimBartholomew (approved):
LGTM

watertap/unit_models/tests/unit_test_harness.py
TimBartholomew: Never mind, the key is actually the variable. So you don't need these lines at all, just get its value with `key.value`.

MarcusHolly:
Subsequent PRs will update the remaining unit model tests.

watertap/unit_models/tests/test_anaerobic_digestor.py
bknueven: > The test harness no longer includes these types of tests, which is testing the default configuration arguments. I'm wondering if that is fine.
> 
> If a developer changed the default configuration arguments, then what would happen?
> 
>     * the solution regression test will likely fail since now it is evaluating a different model
> 
>     * the developer would then update the solution regression test
> 
>     * the developer might not create a new test that covers the previous default options (I think this is the main issue).
> 
> 
> However, this could always happen, the developer could fail to test all configuration argument options. The only way we could prevent this test omission is if we somehow ensure all configuration arguments are tested. I'm not sure how we would do this, and its definitely out of scope for this PR. So I think this is fine.

I think the answer here is we just need to look at the coverage reports to make sure all the configuration options are covered.

watertap/unit_models/tests/unit_test_harness.py
```diff
@@ -0,0 +1,175 @@
...
+                )
+            raise AssertionError(
+                "Check the output logs for variables that need to be re-scaled"
+            )
+        assert_optimal_termination(results)
+
+        # check results
+
+        for var, val in solutions.items():
+            if not pytest.approx(value(var), abs=1e-08, rel=1e-03) == val:
```
bknueven: I think we should do something different here, along the lines @lbianchi-lbl suggested.

A relative value of `1e-03` is very loose, different than we use elsewhere. I have two suggestions:
1. We should register these defaults in `configure_class` and let the developer overwrite the values in `configure`
2. Let `val` be an instance of `_pytest.python_api.ApproxScalar` or anything else which implements `__eq__()` but cannot be cast to `float`. See my code examples.

bknueven:
```suggestion
        for var, val in solutions.items():
            comp_obj = None
            try:
                val = float(val)
            except:
                # expect the same API as pytest.approx
                comp_obj = val
                val = comp_obj.expected
            if comp_obj is None:
                comp_obj = pytest.approx(val, abs=self.default_absolute_tolerance, rel=self.default_relative_tolerance)
            if not comp_obj == value(var):
```

watertap/unit_models/tests/unit_test_harness.py
```diff
@@ -0,0 +1,175 @@
...
+    """
+
+
+class UnitTestHarness:
+    def configure_class(self):
+        self.solver = None  # string for solver, if None use WaterTAP default
+        self.optarg = (
+            None  # dictionary for solver options, if None use WaterTAP default
+        )
+
```
bknueven, line 60:
```suggestion

    # same as pytest.approx: can be overwritten by `configure()`
    self.default_absolute_tolerance = 1e-12
    self.default_relative_tolerance = 1e-06
    
```

watertap/unit_models/tests/unit_test_harness.py
```diff
@@ -0,0 +1,175 @@
...
+        self.configure()
+        blk = self.unit_model_block
+
+        # attaching objects to model to carry through in pytest frame
+        assert not hasattr(blk, "_test_objs")
+        blk._test_objs = Block()
+        blk._test_objs.solver = self.solver
+        blk._test_objs.optarg = self.optarg
+        blk._test_objs.stateblock_statistics = self.unit_statistics
+        blk._test_objs.unit_solutions = self.unit_solutions
```
bknueven: We always have access to `self`, so I don't understand the need to attach these objects to the block.

watertap/unit_models/tests/test_anaerobic_digestor.py
```diff
@@ -163,269 +111,99 @@ def adm(self):
...
+        self.unit_statistics = {
+            "number_variables": 203,
+            "number_total_constraints": 171,
+            "number_unused_variables": 0,
         }
-
-    @pytest.mark.unit
-    def test_report(self, adm):
-        adm.fs.unit.report()
+        self.unit_solutions = ComponentMap()
```
bknueven: If we're going to require this to be a `ComponentMap`, I would suggest setting it (and the unit_statistics) up in `configure_class` before the call to `configure`.

watertap/unit_models/tests/unit_test_harness.py
```diff
@@ -0,0 +1,175 @@
...
+            opt = get_solver()
+        else:
+            opt = get_solver(
+                solver=blk._test_objs.solver, options=blk._test_objs.optarg
+            )
+        results = opt.solve(blk)
+
+        # check solve
+        try:
+            assert len(list(iscale.badly_scaled_var_generator(blk, zero=1e-8))) == 0
```
bknueven: I also think we should optionally let the user optionally define the defaults for the various options for `badly_scaled_var_generator` per unit model.

bknueven (commented):
This is looking good. I have several comments to make this more versatile such that we can more easily roll it out to more unit models.

watertap/unit_models/tests/test_anaerobic_digestor.py
MarcusHolly: How would `self.unit_statistics` be defined here? `self.unit_statistics = ComponentMap()?`. This ties in with your earlier comment about the need to attach those objects to the block.

watertap/unit_models/tests/unit_test_harness.py
MarcusHolly: Would this just look like defining the arguments in `configure_class` similar to how the tolerances are now handled?

watertap/unit_models/tests/test_anaerobic_digestor.py
```diff
@@ -163,269 +111,60 @@ def adm(self):
...
+        self.unit_solutions[m.fs.unit.liquid_outlet.conc_mass_comp[0, "X_pr"]] = 0.10
+        self.unit_solutions[m.fs.unit.liquid_outlet.conc_mass_comp[0, "X_pro"]] = 0.14
+        self.unit_solutions[m.fs.unit.liquid_outlet.conc_mass_comp[0, "X_su"]] = 0.42
+        self.unit_solutions[m.fs.unit.liquid_outlet.conc_mass_comp[0, "S_IC"]] = 1.8
+        self.unit_solutions[m.fs.unit.liquid_outlet.conc_mass_comp[0, "S_IN"]] = 1.82
+        self.unit_solutions[m.fs.unit.liquid_outlet.anions[0]] = 0.02
+        self.unit_solutions[m.fs.unit.liquid_outlet.cations[0]] = 0.04
+        self.unit_solutions[m.fs.unit.vapor_outlet.pressure[0]] = 101325
+        self.unit_solutions[m.fs.unit.vapor_outlet.temperature[0]] = 298.15
+        self.unit_solutions[m.fs.unit.vapor_outlet.flow_vol[0]] = 1
```
MarcusHolly: @bknueven I haven't been able to figure out why, but something about the latest commit changed the expected values for a lot of these variables, and I don't think they're accurate anymore. For example, the `m.fs.unit.vapor_outlet.temperature` should be 308.15, not 298.15. It looks like pressure, temperature and volumetric flow are getting set to default values...

watertap/unit_models/tests/test_anaerobic_digestor.py
bknueven: No, it could just be a dictionary, or we could define a small class to hold the model statistics.

I don't really like checking these anyways. I don't know what it's suppose to catch, and it seems like something that has to be updated any time the model changes slightly -- even if the revised model is mathematically equivalent. (Thoughts @TimBartholomew?)

watertap/unit_models/tests/test_anaerobic_digestor.py
bknueven: Maybe try un-doing this suggestion first? https://github.com/watertap-org/watertap/pull/1277#discussion_r1469970321

If that works let me know...

watertap/unit_models/tests/test_anaerobic_digestor.py
MarcusHolly: Tim and I had discussed just getting rid of these tests entirely. Removing them would probably make everyones' lives a little easier with virtually no downside since the tests aren't very helpful in the first place.

watertap/unit_models/tests/test_anaerobic_digestor.py
MarcusHolly: If I use `solutions = blk._test_objs.unit_solutions` in `test_unit_solutions` and add `blk._test_objs.unit_solutions = self.unit_solutions,` back into `configure_class` the expected values revert back to what they should be. I'm still not sure why this is happening tho

bknueven: Okay. Ignore my earlier comment and leave a note about it where you create that block.

There's surely a good Pyomo explanation but I cannot think of the cause off the top of my head. We should probably investigate further, but that's not needed for this PR.

watertap/unit_models/tests/test_anaerobic_digestor.py
bknueven: Yeah, please get rid of them. Checking the number of unused variables is good, but should probably just be done at the model development step and not in the tests.

watertap/unit_models/tests/unit_test_harness.py
bknueven: Needs to be done for some reason related to the variable references getting out of date. Let's just leave this as-is but make a note.

bknueven (commented):
Looks good!

@lbianchi-lbl can you have a final look at this before it's merged?

watertap/unit_models/tests/test_anaerobic_digestor.py
```diff
@@ -163,269 +110,104 @@ def adm(self):
...
-        }
-
-    @pytest.mark.unit
-    def test_report(self, adm):
-        adm.fs.unit.report()
+        # Set scaling factors for badly scaled variables
+        iscale.set_scaling_factor(
+            m.fs.unit.liquid_phase.mass_transfer_term[0, "Liq", "S_h2"], 1e7
+        )
+        iscale.set_scaling_factor(m.fs.unit.costing.capital_cost, 1e-6)
```
adam-a-a, line 121: Didn't verify, but do we apply these same scaling factors on the BSM2 flowsheet? Just curious

adam-a-a: I also wonder whether we should add this scaling to the unit itself under its `calculate_scaling_factors ` method. I know that isn't the focus of the PR, but if you can probe into this and address as needed in a separate issue/follow-on PR, that would be great. For example, would adding the scale factor for `mass_transfer_term` to the unit break other tests (i.e., BSM2)?

watertap/unit_models/tests/unit_test_harness.py
adam-a-a: I would also include the comment that Ben suggested: `# same as pytest.approx: can be overwritten by `configure()``

adam-a-a:
I would also suggest adding documentation on the unit test harness (and I suppose the property test harness) for developers, before we go replacing all tests with test harnesses. Can create an issue to track this and save for another PR though.

MarcusHolly:
> I would also suggest adding documentation on the unit test harness (and I suppose the property test harness) for developers, before we go replacing all tests with test harnesses. Can create an issue to track this and save for another PR though.

Good point - I'll create documentation for both of the test harnesses before replacing additional tests. I assume a How-To would be most appropriate?

watertap/unit_models/tests/test_anaerobic_digestor.py
MarcusHolly: These scaling factors aren't applied in the BSM2 flowsheet.

I'm not sure that we should set this scaling factor on the unit model since it's only applying the scaling to a single component.

adam-a-a:
I think this is good to go, but @lbianchi-lbl - your call.

watertap/unit_models/tests/test_anaerobic_digestor.py
```diff
@@ -73,44 +54,9 @@
...
-    assert m.fs.unit.config.reaction_package is m.fs.rxn_props
-
-
-# -----------------------------------------------------------------------------
-class TestAdm(object):
-    @pytest.fixture(scope="class")
-    def adm(self):
+class TestUnitDefault(UnitTestHarness):
+    @pytest.mark.unit
+    def configure(self):
```
lbianchi-lbl:
```suggestion

    def configure(self):
```

I'm not sure the marker here on a non-`test_` function would have any effect.

watertap/unit_models/tests/unit_test_harness.py
```diff
@@ -0,0 +1,175 @@
...
+                small=self.default_small,
+                zero=self.default_zero,
+            )
+            for x in badly_scaled_var_list:
+                print(
+                    f"{x[0].name}\t{x[0].value}\tsf: {iscale.get_scaling_factor(x[0])}"
+                )
+            raise AssertionError(
+                "Check the output logs for variables that need to be re-scaled"
+            )
```
lbianchi-lbl:
```suggestion
        badly_scaled_vars = list(
            iscale.badly_scaled_var_generator(
                blk,
                large=self.default_large,
                small=self.default_small,
                zero=self.default_zero,
            )
        )
        if badly_scaled_vars:
            lines = [
                f"{x[0].name}\t{x[0].value}\tsf: {iscale.get_scaling_factor(x[0])}"
                for x in badly_scaled_vars
            ]
            msg = "One or more badly scaled variables found:\n" 
            msg += "\n".join(lines)
            raise AssertionError(msg)
```

lbianchi-lbl (changes requested):
I think it'd be worth to streamline the `badly_scaled_vars` part  before merging (see suggestion for a possible way to do so) but other than that looks good!

watertap/unit_models/tests/unit_test_harness.py
MarcusHolly: LGTM - thanks @lbianchi-lbl.

Just as an FYI, the "short test summary info" doesn't display the full error message, but maybe that's why it's called "short"
image

bknueven (approved):

lbianchi-lbl (approved):

PR #1278, opened 2024-01-24
Dye Desalination TEA Update

MarcusHolly:
## Summary/Motivation:
This PR adds a separator that will function as a dewatering unit to reduce the cost of sludge disposal by reducing the volume of sludge that has to be incinerated in the dye desalination w/RO flowsheet.

### Legal Acknowledgement


watertap/costing/unit_models/dewatering.py
```diff
@@ -104,8 +104,26 @@ def cost_centrifuge(
...
     cost_blk = blk.costing_package.centrifuge
     t0 = blk.flowsheet().time.first()
+    # Assume inlet state block is `mixed_state` (IDAES Separator or WaterTAP Dewatering unit)
+    if hasattr(blk.unit_model.config, 'mixed_state_block'):
+        if blk.unit_model.config.mixed_state_block is None:
+            sb = blk.unit_model.mixed_state[t0]
+        else:
+            sb = blk.unit_model.config.mixed_state_block[t0]
+    else:
+        raise BurntToast("Costing of the dewatering unit is only compatible with an IDAES Separator or WaterTAP Dewatering Unit.")
```
adam-a-a: Probably not the best choice of exception to raise, but I tried to be quick about it.

watertap/costing/unit_models/dewatering.py
```diff
@@ -104,8 +104,26 @@ def cost_centrifuge(
...
+    
+    if hasattr(sb, "flow_vol"):
+        sb_flow_vol = getattr(sb, "flow_vol")
+    elif hasattr(sb, "flow_vol_phase"):
+        sb_flow_vol = getattr(sb, "flow_vol_phase['Liq']")
+    else:
+        raise AttributeError(
+            "Expected 'flow_vol' or 'flow_vol_phase['Liq'] as volumetric flowrate property name for costing the dewatering unit."
+        )
```
adam-a-a, line 125: same as above, though not that important

watertap/costing/unit_models/dewatering.py
andrewlee94: That would be better as a `TypeError` - `BurntToast` is meant as a safety for things the really should never happen.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
```diff
@@ -113,6 +120,8 @@ def build(include_pretreatment=False):
     # define flowsheet inlets and outlets
     m.fs.feed = FeedZO(property_package=m.fs.prop_nf)
     m.fs.dye_retentate = Product(property_package=m.fs.prop_nf)
+    # TODO: Consider adding a config option to recycle the centrate product back to the inlet
+    m.fs.centrate = Product(property_package=m.fs.prop_nf)
```
adam-a-a: @MarcusHolly after going through this a bit more, I think we should aim for the same approach we took on the RO w/ERD revisions: include options to `add_dewatering_unit` (in this PR) and perhaps later on we can do the same for RO and get rid of the other NF only flowsheet (latter shouldn't happen in this PR).

adam-a-a: In doing so, we can, for example, easily simulate our old results without the dewatering unit and rerun against inclusion of the dewatering unit. We might also want to run cost options for each of the 3 cost methods (centrifuge, belt press filter, plate press filter).

adam-a-a: We can do the latter by using that new `MultiUnitModelCostingBlock `feature.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
```diff
@@ -191,9 +216,22 @@ def eq_flow_mass_comp(blk, j):
...
     )
-    dye_sep.s02 = Arc(
-        source=dye_sep.nanofiltration.byproduct, destination=m.fs.dye_retentate.inlet
-    )
+    if hasattr(m.fs, "dewater"):
+        dewater.s01 = Arc(
+            source=dye_sep.nanofiltration.byproduct, destination=dewater.dewaterer.inlet
+        )
+        # TODO: Recycle centrate stream back to the feed via a mixer
+        dewater.s02 = Arc(
```
MarcusHolly: @adam-a-a I've tried adding this recycle loop and RO was failing to initialize even after playing around with tear guesses. So I figured I'd just leave this centrate as a product for now so that I can generate some numbers for tomorrow. One thing I did notice is that the RO permeate isn't being recycled either - is there a reason for this? If so, maybe it doesn't make sense to recycle this centrate stream either.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
```diff
@@ -428,6 +496,35 @@ def add_costing(m):
...
+        #             "costing_method": cost_filter_belt_press,
+        #             "costing_method_arguments": {"cost_electricity_flow": False},
+        #         },
+        #         "filter_plate_press": {
+        #             "costing_method": cost_filter_plate_press,
+        #             "costing_method_arguments": {"cost_electricity_flow": False},
+        #         },
+        #     },
+        #     initial_costing_block="centrifuge",
+        # )
```
MarcusHolly: This MultiUnitCostingBlock was running into the following error and I haven't had the time to dig into it further:
image

I'll likely revisit this after I've generated some results/figures.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
adam-a-a: @MarcusHolly Recycling the centrate stream back to the "headworks" of the plant is something that seems practical and closer to reality in my view. The RO permeate on the other hand would be recycled back to the textile plant for reuse, and this can just be represented by connecting the RO permeate stream with a Product block.

adam-a-a: We might have an issue similar to what @bknueven identified for on BSM2, where there could be some degeneracy introduced by a redundant constraint on the Mixer. Mind you, I haven't revisited the code yet to confirm, but it's worth mentioning. In any case, we can leave this as a to-do item for later.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
adam-a-a: Why add `dewaterer` after `dewater`? can you try  `m.fs.dewater.costing = MultiUnitModelCostingBlock `instead and see if you get the same error?

MarcusHolly: `dewater` is the name of the `Block()` and `dewaterer` is the name of the actual unit model. It's identical to how costing was handled for the pretreatment section. If I just use `m.fs.dewater.costing`, I get the following error: `TypeError: fs.dewater.costing - parent object (fs.dewater) is not an instance of a UnitModelBlockData object. UnitModelCostingBlocks can only be added to UnitModelBlocks.`

adam-a-a: I see... `dewater` is just a `Block`. In any case, because we are compartmentalizing in this fashion (not a bad thing), it looks like cost functions that use `blk.flowsheet()` would break since `flowsheet` attribute is higher up on the parent block. @bknueven - what do you think?

bknueven: Use an IDAES `FlowsheetBlock` instead of a Pyomo `Block`.

MarcusHolly: Replacing `dewater = m.fs.dewater = Block()` with `dewater = m.dewater = FlowsheetBlock(dynamic=False)` or `dewater = m.fs.dewater = FlowsheetBlock(dynamic=False)` still results in the same failure.

watertap/costing/unit_models/dewatering.py
```diff
@@ -139,8 +159,28 @@ def cost_filter_belt_press(
...
     cost_blk = blk.costing_package.filter_belt_press
     t0 = blk.flowsheet().time.first()
+    # Assume inlet state block is `mixed_state` (IDAES Separator or WaterTAP Dewatering unit)
+    if hasattr(blk.unit_model.config, "mixed_state_block"):
+        if blk.unit_model.config.mixed_state_block is None:
+            sb = blk.unit_model.mixed_state[t0]
+        else:
+            sb = blk.unit_model.config.mixed_state_block[t0]
+    else:
+        raise BurntToast(
```
adam-a-a: I introduced this, but can you replace BurntToast with TypeError?

watertap/costing/unit_models/dewatering.py
```diff
@@ -177,8 +217,29 @@ def cost_filter_plate_press(
...
     x_units = pyo.units.gallon / pyo.units.hr
-    x = flow_in = pyo.units.convert(blk.unit_model.inlet.flow_vol[t0], to_units=x_units)
+    # Assume inlet state block is `mixed_state` (IDAES Separator or WaterTAP Dewatering unit)
+    if hasattr(blk.unit_model.config, "mixed_state_block"):
+        if blk.unit_model.config.mixed_state_block is None:
+            sb = blk.unit_model.mixed_state[t0]
+        else:
+            sb = blk.unit_model.config.mixed_state_block[t0]
+    else:
+        raise BurntToast(
```
adam-a-a: Can you replace with TypeError?

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
adam-a-a: @MarcusHolly just to clarify- is it true that you can get cost results without this error when you do not use the MultiUnitModelCostingBlock, but when you try using MultiUnitModelCostingBlock to obtain cost results, you see this error?

MarcusHolly: Correct - the regular UnitModelCostingBlock is able to generate results

adam-a-a (changes requested):
Just a couple of cleanup comments and I think we can merge. Remaining issues can be resolved in future PRs. Let's create any issues, as needed. I created one related to use of MultiUnitModelCostingBlock already.

adam-a-a (approved):

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
```diff
@@ -112,7 +120,14 @@ def build(include_pretreatment=False):
 
     # define flowsheet inlets and outlets
     m.fs.feed = FeedZO(property_package=m.fs.prop_nf)
-    m.fs.dye_retentate = Product(property_package=m.fs.prop_nf)
+
+    if include_dewatering == True:
+        dewater = m.fs.dewater = Block()
```
luohezhiming: Can you clarify why you introduced this dewater Block rather than directly add `m.fs.dewater=Separator(...)`. Then for the following arc, you can directly say` if include_dewatering == True:` and define another flowsheet

MarcusHolly: Ironically, I had the same question, but I was just following the convention set by how pretreatment was handled in this flowsheet (which is also a config option that can be toggled on and off). I assume there was a reason why the original flowsheet was set up this way... any insight on this @adam-a-a ?

MarcusHolly: @luohezhiming @adam-a-a Directly using `m.fs.dewaterer = Separator(...)` does not make the MultiUnitModelCostingBlock work - it still runs into the same failure (see https://github.com/watertap-org/watertap/pull/1278#discussion_r1471618168).

luohezhiming (approved):

PR #1281, opened 2024-01-26
update LCOW print in display_metrics of mvc flowsheet example

savannahsakhai:
## Fixes/Resolves:

in line 692, in display_metrics
    "Levelized cost of water:                  %.2f $/m3" % m.fs.costing.LCOW.value
AttributeError: 'ScalarExpression' object has no attribute 'value'. Did you mean: 'values'?

LCOW is now an expression so this method of obtaining its value is no longer applicable. 

## Summary/Motivation:

Update the display metrics print statement to account for the change in LCOW. 

- 
-

### Legal Acknowledgement


watertap/examples/flowsheets/mvc/mvc_single_stage.py
```diff
@@ -688,9 +688,7 @@ def display_metrics(m):
         "Specific energy consumption:              %.2f kWh/m3"
         % value(m.fs.costing.specific_energy_consumption)
     )
-    print(
-        "Levelized cost of water:                  %.2f $/m3" % m.fs.costing.LCOW.value
-    )
+    print("Levelized cost of water:                  %.2f $/m3" % m.fs.costing.LCOW())
```
adam-a-a:
```suggestion
    print("Levelized cost of water:                  %.2f $/m3" % value(m.fs.costing.LCOW)
```
We generally wrap the expression within value() (from pyomo.environ), since this way works with Vars and Expressions.

adam-a-a (changes requested):
Ready to merge after applying suggestion.

adam-a-a:
@savannahsakhai while we're here, I wonder if we want to add one test for MVC that simply runs the `main `function and ensures that it runs without issues.

I noticed this only because the codecov patch check failed. When I inspected, I saw this (https://app.codecov.io/gh/watertap-org/watertap/pull/1281/blob/watertap/examples/flowsheets/mvc/mvc_single_stage.py), indicating that the main function itself wasn't tested, though parts of it were tested independently:
image

savannahsakhai:
> @savannahsakhai while we're here, I wonder if we want to add one test for MVC that simply runs the `main `function and ensures that it runs without issues.
> 
> I noticed this only because the codecov patch check failed. When I inspected, I saw this (https://app.codecov.io/gh/watertap-org/watertap/pull/1281/blob/watertap/examples/flowsheets/mvc/mvc_single_stage.py), indicating that the main function itself wasn't tested, though parts of it were tested independently: image

I agree. That was the issue from the start, that this portion of the file was not tested. The test is added now. Let me know if you think it needs any changes.

hunterbarber (approved):
LGTM

watertap/examples/flowsheets/mvc/tests/test_mvc_single_stage.py
```diff
@@ -335,3 +336,7 @@ def test_optimization(self, mvc_single_stage):
             22.36, rel=1e-2
         )
         assert value(m.fs.costing.LCOW) == pytest.approx(4.52, rel=1e-2)
+
```
hunterbarber, line 339:
```suggestion

    @pytest.mark.requires_idaes_solver
```

adam-a-a (approved):

PR #1287, opened 2024-02-01
Debugging solver wrapper

bknueven:
## Fixes/Resolves: N/A
This is a capability I half developed while working on a model diagnostics tool. I think it would be broadly useful, however.

I would like some feedback before working on tests and documentation.

## Summary/Motivation:
Debugging a failed initialization can be cumbersome. Even using python's debugging tools, the ideal breakpoint made need to be set outside WaterTAP. Further, it is often useful to have the model state (i.e., initial variable values) from before the solve failed.

This PR would create a solver wrapper to enable this type of debugging. The solver wrapper can
1. Be called like a regular Pyomo solver.
2. Stores initialization information before calling the underlying solver
3. Uses `check_optimal_termination` to return the underlying solver's result if it succeeded normally.
4. If the solve did not succeed, dumps the user out into an IPython notebook with the model state restored to begin debugging.

The PR also contains a module which when imported will replace the default WaterTAP solver, `ipopt-watertap`, with the wrapped solver, for ease-of-use.

Example:

Current behavior:
```
2024-02-01 13:06:07 [DEBUG] idaes.solve.fs.bed_stack: EXIT: Converged to a point of local infeasibility. Problem may be infeasible.
2024-02-01 13:06:07 [DEBUG] idaes.solve.fs.bed_stack: WARNING: Loading a SolverResults object with a warning status into
2024-02-01 13:06:07 [DEBUG] idaes.solve.fs.bed_stack: model.name="fs.bed_stack";
2024-02-01 13:06:07 [DEBUG] idaes.solve.fs.bed_stack:     - termination condition: infeasible
2024-02-01 13:06:07 [DEBUG] idaes.solve.fs.bed_stack:     - message from solver: Ipopt 3.14.11\x3a Converged to a locally infeasible
2024-02-01 13:06:07 [DEBUG] idaes.solve.fs.bed_stack:       point. Problem may be infeasible.
2024-02-01 13:06:07 [INFO] idaes.init.fs.bed_stack: Initialization Step 3 infeasible - Converged to a locally infeasible point. Problem may be infeasible..
2024-02-01 13:06:07 [WARNING] idaes.init.fs.bed_stack:  The solver at the Initialization Step 3 step failed to converge to an optimal solution.This suggests that the user provided infeasible inputs or that the model is poorly scaled, poorly initialized, or degenerate.
2024-02-01 13:06:07 [INFO] idaes.init.fs.bed_stack: Initialization Complete: infeasible - Converged to a locally infeasible point. Problem may be infeasible.
  File "/Users/bknueven/Software/ed-pilots/ed-pilots/brine_electrodialysis/bed_simulation.py", line 439, in <module>
    m, res = main()
  File "/Users/bknueven/Software/ed-pilots/ed-pilots/brine_electrodialysis/bed_simulation.py", line 53, in main
    m, res = run_simulation(case, parameter_estimates)
  File "/Users/bknueven/Software/ed-pilots/ed-pilots/brine_electrodialysis/bed_simulation.py", line 106, in run_simulation
    model_initialize(m, case)
  File "/Users/bknueven/Software/ed-pilots/ed-pilots/brine_electrodialysis/bed_simulation.py", line 313, in model_initialize
    model.fs.bed_stack.initialize(outlvl=idaeslog.DEBUG, ignore_dof=True)
  File "/Users/bknueven/Software/watertap/watertap/core/initialization_mixin.py", line 23, in initialize
    return super().initialize(*args, **kwargs)
  File "/Users/bknueven/opt/anaconda3/envs/watertap/lib/python3.10/site-packages/idaes/core/base/unit_model.py", line 540, in initialize
    flags = blk.initialize_build(*args, **kwargs)
  File "/Users/bknueven/Software/watertap/watertap/unit_models/electrodialysis_1D.py", line 2146, in initialize_build
    raise InitializationError(f"Unit model {self.name} failed to initialize")
idaes.core.util.exceptions.InitializationError: Unit model fs.bed_stack failed to initialize
```
Now the user knows that there's some issue with the unit model `fs.bed_stack`, but extracting the exact subproblem that failed my require setting breakpoints in WaterTAP and/or IDAES, and slows down the development process.

In the user's python module, they would make a simple import:
```python
import watertap.core.util.model_debug_mode import activate
activate()
```
and now when the solve fails, the user is dumped into a IPython notebook with the problematic model ready for using any diagnostic tool the developer wishes to deploy.
```
EXIT: Converged to a point of local infeasibility. Problem may be infeasible.
WARNING: Loading a SolverResults object with a warning status into
model.name="fs.bed_stack";
    - termination condition: infeasible
    - message from solver: Ipopt 3.14.11\x3a Converged to a locally infeasible
      point. Problem may be infeasible.
Solver debugging mode: the block fs.bed_stack failed to solve.
fs.bed_stack is called `blk` in this context.
The solver ipopt-watertap is available in the variable `solver`.
The Initial values have be restored into the block.
You can restore them anytime by calling `debug.restore_initial_values(blk)`.
The model has been loaded into an IDAES DiagnosticsToolbox instance called `dt`.
Python 3.10.9 (main, Jan 11 2023, 09:18:20) [Clang 14.0.6 ]
Type 'copyright', 'credits' or 'license' for more information
IPython 7.34.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: blk.name
Out[1]: 'fs.bed_stack'

In [2]: dt.report_structural_issues()
====================================================================================
Model Statistics

        Activated Blocks: 15 (Deactivated: 0)
        Free Variables in Activated Constraints: 566 (External: 0)
            Free Variables with only lower bounds: 136
            Free Variables with only upper bounds: 0
            Free Variables with upper and lower bounds: 240
        Fixed Variables in Activated Constraints: 42 (External: 7)
        Activated Equality Constraints: 566 (Deactivated: 0)
        Activated Inequality Constraints: 0 (Deactivated: 0)
        Activated Objectives: 0 (Deactivated: 0)

------------------------------------------------------------------------------------
1 WARNINGS

    WARNING: Found 354 potential evaluation errors.

------------------------------------------------------------------------------------
2 Cautions

    Caution: 3 variables fixed to 0
    Caution: 11 unused variables (0 fixed)

------------------------------------------------------------------------------------
Suggested next steps:

    display_potential_evaluation_errors()

====================================================================================

In [3]: 
```

### Legal Acknowledgement


watertap/core/plugins/solvers.py
```diff
@@ -202,6 +206,84 @@ def _is_user_scaling(self):
...
+
+    assert pyo.SolverFactory(solver_name).available()
+
+    debug_solver_name = f"debug-solver-wrapper-{solver_name}"
+
+    @pyo.SolverFactory.register(
+        debug_solver_name,
+        doc=f"Debug solver wrapper for {solver_name}",
+    )
+    class DebugSolverWrapper:
```
lbianchi-lbl: I don't know if this is compatible with the way `@pyo.SolverFactory.register()` works, but would it be possible to move the class definition for `DebugSolverWrapper` in the outer (module) scope, and only do the registration (as a callable rather than as a decorator) inside `create_debug_solver_wrapper()`? e.g.

```py
class DebugSolverWrapper:
    ...


def create_debug_solver_wrapper(solver_name):
    DebugSolverWrapper._base_solver = solver_name
    debug_solver_name = ...
    pyo.SolverFactory.register(debug_solver_name, doc=...)(DebugSolverWrapper)
```

I don't think there'd be any significant change in functionality, but it'd be a bit more readable (and I think at least the class definition lines would be covered since they'd be in the module scope).

lbianchi-lbl:
Ugh, I was almost done writing a long comment, and GitHub seems to have lost it :upside_down_face: 

Anyway, the main points were:

- I like this solution and it looks like it could be very useful for debugging
- The import-based syntax is nice and concise but I'm a bit concerned of the module-level code being executed unexpectedly by e.g. any tools that could import modules dynamically (Pylint? pytest? Sphinx plugins?)
- As a way to avoid this without making the syntax significantly more verbose, what about wrapping the code in a function, e.g.:

```py

import idaes
from idaes.logger import solver_capture_off
from watertap.core.plugins.solvers import create_debug_solver_wrapper


def activate():
    # create a debug solver around ipopt-watertap
    debug_solver_name = create_debug_solver_wrapper("ipopt-watertap")

    # reconfigure the default IDAES solver to use the debug wrapper
    _default_solver_config_value = idaes.cfg.get("default_solver")
    _default_solver_config_value.set_default_value(debug_solver_name)
    if not _default_solver_config_value._userSet:
        _default_solver_config_value.reset()

    # disable solver log capturing so the resulting notebook
    # can use the whole terminal screen
    solver_capture_off()
```

And then the client code would be:

```py
from watertap.core.util import model_debug_mode; mode_debug_model.activate()
```

The code being in a function would still allow us to provide a bit more "magic" if we wanted to, e.g.

```py
import os

if os.environ.get("WATERTAP_MODEL_DEBUG_MODE", None):
    from watertap.core.util.model_debug_mode import activate; activate()
```

(it could be tricky to find a place to put the above snippet so that it's reliably imported without running into circular import problems, but *handwaves*)

watertap/core/plugins/solvers.py
bknueven: Maybe a case of me trying to be too general, but I coded it this way so that you could have a `DebugSolverWrapper` for different solvers at the same time.

Because the `_base_solver` is a class attribute, you need to redefine the class to allow it to take different values. But maybe it would be worth figuring out if it can be an instance attribute without breaking things.

lbianchi-lbl: Oh, I see. So maybe this would work?

```py
class _BaseDebugSolverWrapper:
    ...

def create_debug_solver_wrapper(solver_name):
    debug_solver_name = ...
    @pyo.SolverFactory.register(debug_solver_name, doc=...)
    class DebugSolverWrapper(_BaseDebugSolverWrapper):
        _base_solver = solver_name
```

... but I fully admit this might be stretching my "a bit more readable" claim too thin compared to your initial implementation (that, to be clear, I have no concrete issue with).

bknueven: I agree this is more readable, for two reasons:
1. removes a level of indentation
2. Makes it easier to reason about this generator function.

So I implemented it in 57e510f.

bknueven:
@lbianchi-lbl any thoughts on testing? Seems difficult given we'd need to supply input to break out of the embedded IPython environment.

lbianchi-lbl:
> @lbianchi-lbl any thoughts on testing? Seems difficult given we'd need to supply input to break out of the embedded IPython environment.

That's an excellent question... Do you have a sample broken (or easily breakable) model that I can use to test this PR manually first?

bknueven:
> > @lbianchi-lbl any thoughts on testing? Seems difficult given we'd need to supply input to break out of the embedded IPython environment.
> 
> That's an excellent question... Do you have a sample broken (or easily breakable) model that I can use to test this PR manually first?

This model is structurally infeasible, and will cause Ipopt to fail every time:
```python
import pyomo.environ as pyo
m = pyo.ConcreteModel()
m.x = pyo.Var([1,2], bounds=(0,1))
m.c = pyo.Constraint(expr=m.x[1] * m.x[2] == -1)
```

lbianchi-lbl:
@bknueven 0fb11ff adds a (somewhat involved) test file:

- The code standing in for the non-solving model is a string executed in a subprocess
- The IPython environment is tested using `Popen.communicate`
- A utility class is used to wrap a list of statements to run within the IPython session via `sys.stdin` and collect any error raised through an intermediate text file
  - I resorted to this crude "inter-process communication at home" method after trying and failing to get IPython to exit with a nonzero return code :man_shrugging: 
- I left an error-causing statement in 0fb11ff on purpose to verify that the `assert ipy.error_text == ""` line actually works (the `RuntimeError` however is there by mistake :sweat_smile: )

PS: after tying the functionality in an interactive session before writing the test: it's super cool!

lbianchi-lbl:
Ugh, the tests are "failing to fail" on Windows (see below). Let me try to see why that might be the case.

image

lbianchi-lbl:
@bknueven OK, I think Windows is now failing the same (expected) way as Linux (the failures are caused by the `pyo.Blockkk` misspell in one of the statements).

bknueven:
Thanks @lbianchi-lbl, I think it would have taken me a lot longer to figure out how to test this.

Maybe besides some documentation, is this ready then? Should it be in IDAES instead?

bknueven:
@lbianchi-lbl I might need your help updating the test. I am still seeing this locally, and I am not sure what it means..

```
        out, err = proc.communicate(input=ipy.for_stdin, timeout=30)
>       assert ipy.error_text == ipy.message_when_no_errors
E       AssertionError: assert '' == 'NO ERRORS WHATSOEVER'
E         
E         - NO ERRORS WHATSOEVER

tests/test_model_debug_mode.py:104: AssertionError
====================================================== slowest 1 durations ======================================================
1.41s call     watertap/core/util/tests/test_model_debug_mode.py::test_debug_mode
```

bknueven:
> @lbianchi-lbl I might need your help updating the test. I am still seeing this locally, and I am not sure what it means..
> 
> ```
>         out, err = proc.communicate(input=ipy.for_stdin, timeout=30)
> >       assert ipy.error_text == ipy.message_when_no_errors
> E       AssertionError: assert '' == 'NO ERRORS WHATSOEVER'
> E         
> E         - NO ERRORS WHATSOEVER
> 
> tests/test_model_debug_mode.py:104: AssertionError
> ====================================================== slowest 1 durations ======================================================
> 1.41s call     watertap/core/util/tests/test_model_debug_mode.py::test_debug_mode
> ```

Does [df8d011](https://github.com/watertap-org/watertap/pull/1287/commits/df8d0116fc550338401fff2752bc13e1865ca3ad) fix this?

lbianchi-lbl:
> > @lbianchi-lbl I might need your help updating the test. I am still seeing this locally, and I am not sure what it means..
> > ```
> >         out, err = proc.communicate(input=ipy.for_stdin, timeout=30)
> > >       assert ipy.error_text == ipy.message_when_no_errors
> > E       AssertionError: assert '' == 'NO ERRORS WHATSOEVER'
> > E         
> > E         - NO ERRORS WHATSOEVER
> > 
> > tests/test_model_debug_mode.py:104: AssertionError
> > ====================================================== slowest 1 durations ======================================================
> > 1.41s call     watertap/core/util/tests/test_model_debug_mode.py::test_debug_mode
> > ```
> 
> Does [df8d011](https://github.com/watertap-org/watertap/pull/1287/commits/df8d0116fc550338401fff2752bc13e1865ca3ad) fix this?

It does, but I'm not sure this is what we want:

- The purpose of `message_when_no_errors` to be a non-empty sentinel string is to make sure the functionality that collects and stores the exceptions and/or output emitted when running the subprocess is working correctly
- If `message_when_no_errors` is set to an empty string, we can't distinguishing between A) exceptions were not raised, and the capture is working correctly, or B) exceptions were raised, but the capture didn't work

lbianchi-lbl:
@bknueven after adding some `print(out)` and running it locally:
- It seems that for whatever reason the `else: to_print = {self.message_if_no_errors}` wasn't getting executed correctly by IPython
- The changes in ae93ceb (which in principle should be essentially the same in terms of execution :man_shrugging: ) seem to work to restore the functionality as intended

If you're OK with accepting the solution without digging deeper into why this started happening (perhaps a change in IPython?), then this is good to go (once I revert the `-k test_model_debug_mode`).

bknueven:
> @bknueven after adding some `print(out)` and running it locally:
> 
>     * It seems that for whatever reason the `else: to_print = {self.message_if_no_errors}` wasn't getting executed correctly by IPython
> 
>     * The changes in [ae93ceb](https://github.com/watertap-org/watertap/commit/ae93ceb458a927b6c068e09e404b8cfddd0bb470) (which in principle should be essentially the same in terms of execution 🤷‍♂️ ) seem to work to restore the functionality as intended
> 
> 
> If you're OK with accepting the solution without digging deeper into why this started happening (perhaps a change in IPython?), then this is good to go (once I revert the `-k test_model_debug_mode`).

I am okay without digging deeper. Some mysteries aren't interesting enough to warrant solving.

lbianchi-lbl (approved):

adam-a-a (approved):
There might be some polish to put on this, but I really want this. LGTM

bknueven:
I am indisposed; however, it would be a good idea to turn my above comment into a small rst file as some basic documentation on this capability.

adam-a-a:
> I am indisposed; however, it would be a good idea to turn my above comment into a small rst file as some basic documentation on this capability.

I put up PR #1470 to address this. See you on the other side!

PR #1290, opened 2024-02-02
Add GUI for OARO

luohezhiming:
## Fixes/Resolves:

-
-

### Legal Acknowledgement


watertap/examples/flowsheets/oaro/oaro_multi_ui.py
```diff
@@ -0,0 +1,640 @@
...
+        do_build=build_flowsheet,
+        do_solve=solve_flowsheet,
+        build_options={
+            "NumberOfStages": {
+                "name": "NumberOfStages",
+                "display_name": "Number of stages",
+                "values_allowed": "int",
+                "value": 3,  # default value
+                "max_val": 8,  # optional
+                "min_val": 0,  # optional
```
adam-a-a: Should 0 be an option? I wouldn't think so.

adam-a-a:
@luohezhiming I wonder if the `NoneType` error we're seeing is attributed to included 0 number of stages as a build option. If the tests actually check all build options from min to max values provided, that would explain the error.

luohezhiming:
> @luohezhiming I wonder if the `NoneType` error we're seeing is attributed to included 0 number of stages as a build option. If the tests actually check all build options from min to max values provided, that would explain the error.

I revised the lower bound to 1, but it still has the NoneType error. I guess it is something related to the setting of `build_options`. @MichaelPesce The GUI interface works but the `test_flowsheet_interfaces.py` fails with a NoneType error, I guess it is something related to the new functions with build_option that has not been covered by the test file.

andrewlee94:
@luohezhiming A 'NoneType' error in Pyomo often indicates that you have a variable with no value (i.e. `var.value == None`), which is mostly likely due to the variable not being initialized (yet). Can you show the full traceback for the error?

adam-a-a:
@luohezhiming compare with this example and see if you can get around the error: https://github.com/watertap-org/watertap/blob/main/watertap/examples/flowsheets/nf_dspmde/nf_ui.py#L405-L435

watertap/examples/flowsheets/oaro/oaro_multi_ui.py
```diff
@@ -0,0 +1,667 @@
...
+                "max_val": 8,  # optional
+                "min_val": 1,  # optional
+            },
+            "SystemRecovery": {
+                "name": "SystemRecovery",
+                "display_name": "System Mass Recovery",
+                "values_allowed": "float",
+                "value": 0.5,  # default value
+                "max_val": 1,  # optional
+                "min_val": 0,  # optional
```
MarcusHolly, line 46: I think the min and max value for system mass recovery should be adjusted since the system is not able to solve for me at the extremes. In fact, if I change the value from 0.5 to 0.6, it still fails to solve. So I think the min and max values should be set such that everything in between them solves. Alternatively, maybe it's just better to remove this model option from the GUI altogether since it seems finicky.

adam-a-a (approved):
LGTM! FYI- I did not run this in the UI.

watertap/examples/flowsheets/oaro/oaro_multi_ui.py
MarcusHolly: @luohezhiming I think this should be addressed before merging.

MarcusHolly (changes requested):
I think the min and max value for system mass recovery should be adjusted since the system is not able to solve for me at the extremes. In fact, if I change the value from 0.5 to 0.6, it still fails to solve. So I think the min and max values should be set such that everything in between them solves. Alternatively, maybe it's just better to remove this model option from the GUI altogether since it seems finicky.

watertap/examples/flowsheets/oaro/oaro_multi_ui.py
MarcusHolly: @luohezhiming Just realized that my comments had "Pending" next to them, so I re-sent them by requesting changes. Sorry for the late notice...

luohezhiming: @MarcusHolly The system recovery depends on OARO stages, for the default case study, the number of stage is set to 3, and the system cannot achieve a higher recovery, but if we increase the number of stages, we can achieve a higher system recovery. The theoretical upper limit for system recovery is 1 if we have infinite stages but we can never achieve. So if the system fail to solve with higher recovery, it is normal.

MarcusHolly (approved):

watertap/examples/flowsheets/oaro/oaro_multi_ui.py
adam-a-a: Well, on our LSRRO flowsheet, we imposed an upper limit on the final brine concentration corresponding to the solubility of NaCl. Thus, you wouldn't be able to approach a recovery of 100% with infinite stages but rather the max recovery corresponding to a final brine concentration where NaCl is saturated in solution (beyond this point, halite would form). 

We should probably impose the same limit for the OARO flowsheet in a future PR.

I'll also note that considering NaCl solubility as our upper limit is still quite optimistic. In reality, we'd have to consider other mineral scalants that'd probably dictate the "true" max recovery, but that's a different story.

watertap/examples/flowsheets/oaro/oaro_multi_ui.py
```diff
@@ -0,0 +1,667 @@
...
+        ui_units=pyunits.dimensionless,
+        display_units="fraction",
+        rounding=1,
+        description="Total Purchased Equipment Cost (TPEC)",
+        is_input=True,
+        input_category="System costing",
+        is_output=False,
+    )
+    exports.add(
+        obj=fs.costing.factor_total_investment,
```
MarcusHolly: @luohezhiming This should be `total_investment_factor`

adam-a-a:
```suggestion
        obj=fs.costing.total_investment_factor,
```

luohezhiming: I think it should be factor_total_investment, please check `watertap_costing_package.py`

adam-a-a: @luohezhiming - that was true previously, but see PR #1299 which updated some of the naming for costing.

luohezhiming: I push the modification, hope there is no other conflicts

PR #1292, opened 2024-02-06
In the UI export, allow output variables (which may not exist pre-solve) to be marked as 'deferred'

dangunter:
## Summary/Motivation:

Without this, the attempt to export variables that don't exist yet will prevent some flowsheets from being used with the UI.

In the `watertap.ui.fsapi` module:
- Add a `deferred_obj` attr to the ModelExport class that holds the (string) description of the object
- Add an `undefer()` method to the FlowsheetExport class
- Call `undefer()` automatically once the flowsheet has been solved to resolve the deferred values

### Legal Acknowledgement

### DIFF
--- a/watertap/ui/fsapi.py
+++ b/watertap/ui/fsapi.py
@@ -40,7 +40,7 @@ except ImportError:
 # third-party
 import idaes.logger as idaeslog
 from idaes.core.util.model_statistics import degrees_of_freedom
-from pydantic import BaseModel, validator, Field
+from pydantic import BaseModel, validator, Field, PrivateAttr
 import pyomo.environ as pyo
 
 #: Forward-reference to a FlowsheetInterface type, used in
@@ -71,7 +71,7 @@ class ModelExport(BaseModel):
     _SupportedObjType = Union[
         pyo.Var,
         pyo.Expression,
-        pyo.Param,
+        pyo.Param
     ]
     "Used for type hints and as a shorthand in error messages (i.e. not for runtime checks)"
 
@@ -81,6 +81,7 @@ class ModelExport(BaseModel):
     # so we're using Optional[object] unless we find a way to tell pydantic to skip this check
     # inputs
     obj: Optional[object] = Field(default=None, exclude=True)
+    deferred_obj: Optional[str] = Field(default=None, exclude=True)
     name: str = ""
     value: float = 0.0
     ui_units: object = Field(default=None, exclude=True)
@@ -106,13 +107,14 @@ class ModelExport(BaseModel):
 
     @validator("obj", always=True, pre=True)
     @classmethod
-    def ensure_obj_is_supported(cls, v):
+    def validate_obj(cls, v):
         if v is not None:
             cls._ensure_supported_type(v)
         return v
 
     @classmethod
     def _ensure_supported_type(cls, obj: object):
+        _log.debug(f"ensure supported type obj={obj}")
         is_valid = (
             obj.is_variable_type()
             or obj.is_expression_type()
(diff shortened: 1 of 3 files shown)


dangunter:
@lbianchi-lbl and @adam-a-a I think the rough consensus is that this should not really be needed if the model is properly formulated, and potentially encourages bad practices. Despite this, I do need it right now for the Prommis UKy model, which is a deliverable. Also, there is a lot of miscellaneous cleanup code I would like to keep. Could you review this (when it passes tests) and I will do my best to figure out a more appropriate solution for this particular case in the future, at which point the small delta of code for "undefer"ing the values can be taken back out.

watertap/ui/fsapi.py
```diff
@@ -438,41 +470,97 @@ def from_csv(self, file: Union[str, Path], flowsheet):
...
+        Raises:
+            ValueError, if bad units are provided
+        """
+        # evaluate the units
+        norm_units = text.strip()
+        if norm_units in ("", "none", "-"):
+            parsed_units = pyo.units.dimensionless
+        else:
+            try:
+                parsed_units = eval(norm_units, {"units": pyo.units})
```
lbianchi-lbl: I think we could get rid of both the `eval()` and the need for a `units.` prefix is we used e.g. `pyo.units.get_units()`, which takes a string in input (the underlying Pint library is pretty good at parsing units from strings):

```suggestion
                parsed_units = pyo.units.get_units(norm_units)
```

watertap/ui/fsapi.py
```diff
@@ -438,41 +470,97 @@ def from_csv(self, file: Union[str, Path], flowsheet):
...
+
+        Returns:
+            Pyomo block/var/etc. within the flowsheet for 'text'
+
+        Raises:
+            ValueError, if the object named by 'text' could not be found
+        """
+        _log.debug(f"eval text={text}")
+        try:
+            obj = eval(text, {"fs": flowsheet})
```
lbianchi-lbl: Following @bknueven's suggestion in #1277, Pyomo's `find_component()` would be a safer option than `eval()`, with the important added benefit that we don't require the flowsheet to be called `fs` (as long as `text` contains the correct attribute name, of course).

I believe that using it on the underlying `ConcreteModel` instance (which IIRC you can get from the flowsheet object with `flowsheet.parent_block()`) would not even require changes to existing string-based flowsheet exports (i.e. the same string starting with `fs.` would work in both cases).

watertap/ui/fsapi.py
```diff
@@ -66,21 +66,39 @@ def __init__(
...
+            rounding=4,
+            ui_units=pyo.units.kg/pyo.units.hour,
+            display_units="kg/hr",
+            description="solid flow mass",
+            is_input=False,
+            is_output=True,
+            output_category="solids"
+        )
+
+    Note that it makes no sense to use both attributes at the same time.
```
lbianchi-lbl: This is more of a high-level comment, but I'm wondering if making the `obj` field more versatile (i.e. `obj: Union[_SupportedObjType, str]`) could be an alternative to the current solution of having separate fields and having to enforce (or trust the clients to comply with) their mutual-exclusiveness.

However, I don't know enough about the concrete implementation details so this suggestion might not even make sense (e.g. making it impossible to undefer `str` objects in the very scenarios that require deferred objects in the first place).

lbianchi-lbl (commented):
I've left a few comments with suggestions. 

- Generally I don't think any of these issues are blocking (i.e. I won't object if the PR needs to be merged as is), but I would recommend that these changes are tested extensively (i.e. beyond the automated `fsapi` tests, e.g. going through the UI workflows) to mitigate the risk of introducing regressions
- The suggestions can be ordered in increasing complexity (and likelihood of not working because of edge cases that I haven't thought of) as:
  - Use `pyo.units.get_units()` instead of `eval()`
  - Use `flowsheet.parent_block().find_component()` instead of `eval()`
  - Use a single, more flexible `obj` field that supports `str` instead of a separate `deferred_obj`

watertap/ui/fsapi.py
dangunter: I don't think safe is a real issue , but sure whatever

watertap/ui/fsapi.py
dangunter: I think the current solution is more explicit

dangunter: on second thought, sure..

watertap/ui/fsapi.py
dangunter: I dropped the fs. actually

lbianchi-lbl (approved):

TimBartholomew:
All variables should be built in the build function and therefore should already exist. We want there to be a failure if a variable is being exported that doesn't exist.

dangunter:
Won't fix

PR #1294, opened 2024-02-07
Adding GAC into Dye Desalination w/RO

MarcusHolly:
## Summary/Motivation:
Adds a configuration option to include GAC in the `dye_desal_withRO.py` flowsheet and `dye_desalination.py` flowsheet.

### Legal Acknowledgement


watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
```diff
@@ -131,6 +226,76 @@ def set_operating_conditions(m):
...
+    m.fs.gac.a0.fix(3.68421)
+    m.fs.gac.a1.fix(13.1579)
+    m.fs.gac.b0.fix(0.784576)
+    m.fs.gac.b1.fix(0.239663)
+    m.fs.gac.b2.fix(0.484422)
+    m.fs.gac.b3.fix(0.003206)
+    m.fs.gac.b4.fix(0.134987)
+
+    m.fs.gac.gac_removed[0].flow_mass_phase_comp["Liq", "H2O"] = 1e-10
+    m.fs.gac.gac_removed[0].flow_mass_phase_comp["Liq", "tds"] = 1e-10
```
MarcusHolly: @hunterbarber We are trying to get this `dye_desalination.py` flowsheet to solve before the `dye_desalination_withRO.py`, but the solver always seem to enter restoration. So far we've tried the following:

- L138: Tweaking the dye diffusivity value (which we don't currently have an actual number for)
- L256-257: Setting the problematic zero-flow terms in the `gac_removed` streams to a small, non-zero value. We had also initially set `ele_operational_time` and `ele_conc_ratio_replace` to zero, but this didn't improve the solver stability at all

Have you encountered such issues with the GAC model before and/or do you have any advice for how we should go about solving this (Variables to update, bounds to relac/tighten, etc.)?

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
```diff
@@ -131,6 +202,39 @@ def set_operating_conditions(m):
     dye_sep.P1.eta_pump.fix(0.75)  # pump efficiency [-]
     dye_sep.P1.lift_height.unfix()
 
+    if hasattr(m.fs, "gac"):
+        m.fs.tb_nf_gac.properties_out[0].temperature.fix(298.15)
```
hunterbarber, line 206: Shouldn't this occur somewhere in the translator block? Fixing the outlet T and P implies that the inlet T and P are not fixed, which I would imagine should happen when the stream is translated at the GAC inlet.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
```diff
@@ -131,6 +202,39 @@ def set_operating_conditions(m):
...
+        m.fs.gac.particle_dens_app.fix(750)
+        m.fs.gac.particle_dia.fix(0.001)
+        # adsorber bed specifications
+        m.fs.gac.ebct.fix(600)
+        m.fs.gac.bed_voidage.fix(0.4)
+        m.fs.gac.bed_length.fix(6)
+        # design spec
+        m.fs.gac.conc_ratio_replace.fix(0.50)
+        # parameters
+        m.fs.gac.a0.fix(3.68421)
```
hunterbarber, line 223: I'm assuming these are just the correct parameters at the conditions for the solve in this example? Probably needs some sort of documenting but the examples documentation is it's own thing.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
```diff
@@ -131,6 +202,39 @@ def set_operating_conditions(m):
...
+        m.fs.gac.a1.fix(13.1579)
+        m.fs.gac.b0.fix(0.784576)
+        m.fs.gac.b1.fix(0.239663)
+        m.fs.gac.b2.fix(0.484422)
+        m.fs.gac.b3.fix(0.003206)
+        m.fs.gac.b4.fix(0.134987)
+
+        m.fs.gac.gac_removed[0].flow_mass_phase_comp["Liq", "H2O"] = 1e-10
```
hunterbarber: I think this may be handled by the GAC initialization now, but curious if this is necessary.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
```diff
@@ -144,7 +248,7 @@ def initialize_system(m):
 def solve(blk, solver=None, checkpoint=None, tee=False, fail_flag=True):
     if solver is None:
         solver = get_solver()
-    results = solver.solve(blk, tee=tee)
+    results = solver.solve(blk, tee=True)
```
hunterbarber: Need to revert to `tee=tee` and use the input argument if desired in function call.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
```diff
@@ -94,21 +113,26 @@ def main():
...
 
     return m, results
 
 
-def build(include_pretreatment=False, include_dewatering=False):
+def build(
+    include_pretreatment=False,
+    include_dewatering=False,
+    include_gac=False,
+):
```
hunterbarber, line 126: Is it not possible to merge `dye_desalination` and `dye_desalination_withRO` based on config arguments?

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
```diff
@@ -134,6 +158,52 @@ def build(include_pretreatment=False, include_dewatering=False):
         )
         m.fs.centrate = Product(property_package=m.fs.prop_nf)
         m.fs.precipitant = Product(property_package=m.fs.prop_nf)
+    elif include_gac == True:
```
hunterbarber: Same comments carry over here.

hunterbarber (commented):
Comments on the flowsheets without looking at the tests yet

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
MarcusHolly: These probably can and should be merged into one flowsheet. To provide some additional context, all of these configuration options in `dye_desalination_withRO` are fairly new, and ig the original developer decided to create two separate flowsheets instead of having a `include_RO` config option.

@adam-a-a Is there a reason I'm missing as to why these two flowsheets _shouldn't_ be combined into a single flowsheet.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
MarcusHolly: This is no longer necessary.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
MarcusHolly: Most of the `m.fs.gac` parameters are unchanged from the values set in https://github.com/watertap-org/watertap/blob/main/watertap/examples/flowsheets/gac/gac.py since we haven't identified alternative values in the dye separation literature.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
MarcusHolly: @hunterbarber The inlet T and P to the translator block are not fixed since the ZO property package (the inlet property package) does not support T or P. So we fix the temperature and pressure outlet of this translator, which is then fed to the GAC inlet. Let me know if that doesn't address your question or if I'm misunderstanding something.

hunterbarber: Oh it is the outlet to the translator, I saw `...gac.properties_out...` and mistaken it for the outlet of the GAC block, my fault.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
hunterbarber: Yea this is probably still relevant for later PR improvements.

hunterbarber (approved):
Changes seem good. I would say let's convert the merging of the dye desal flowsheets into an issue for later.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
```diff
@@ -185,18 +286,30 @@ def add_costing(m):
...
+        m.fs.dye_disposal_cost = Expression(
+            expr=(
+                m.fs.zo_costing.utilization_factor
+                * m.fs.zo_costing.dye_disposal_cost
+                * pyunits.convert(
+                    m.fs.dye_retentate.properties[0].flow_vol,
+                    to_units=pyunits.m**3 / m.fs.zo_costing.base_period,
+                )
+            ),
+            doc="Cost of disposing of dye waste",
```
adam-a-a, line 311: In subsequent PR, we can probably handle this with a Product block for the disposal stream. Then, we wouldn't need to worry about the logic of whether gac is included.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
```diff
@@ -248,6 +361,28 @@ def display_results(m):
     for u in unit_list:
         m.fs.dye_separation.component(u).report()
 
+    if hasattr(m.fs, "gac"):
+        m.fs.costing = WaterTAPCosting()
+        m.fs.costing.base_currency = pyunits.USD_2021
```
adam-a-a, line 366: Only skimming changed code and not the whole body of code here, but I would imagine that creating a costing block and setting a base currency would be something you'd do with or without GAC. I suppose this may be due to the rest of the flowsheet only having ZO models and costing and this is the first instance of using a WaterTAPCosting block for a "detailed" model.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
```diff
@@ -76,10 +95,10 @@
 
 
 def main():
-    m = build(include_pretreatment=False, include_dewatering=True)
+    m = build(include_pretreatment=False, include_dewatering=False, include_gac=False)
     set_operating_conditions(m)
 
-    assert_degrees_of_freedom(m, 0)
+    # assert_degrees_of_freedom(m, 0)
```
adam-a-a: delete
```suggestion
```

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
adam-a-a: I definitely agree that these should ultimately be merged into one flowsheet, where the inclusion of RO would become another Boolean option. As @MarcusHolly pointed out, the original developer decided to break them down initially.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
```suggestion
    elif include_gac:
```

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
```diff
@@ -737,16 +924,41 @@ def LCOW(b):
...
+        else:
+            return (
+                b.total_capital_cost * b.zo_costing.capital_recovery_factor
+                + b.total_operating_cost
+            ) / (
+                pyunits.convert(
+                    b.permeate.properties[0].flow_vol,
+                    to_units=pyunits.m**3 / pyunits.year,
+                )
+                * b.zo_costing.utilization_factor
```
adam-a-a, line 959: I think this logic can be eliminated with a proper Product block that captures all relevant product water streams in any flowsheet configuration. Another item for a subsequent PR.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py
```diff
@@ -819,6 +1037,42 @@ def display_results(m):
...
+                m.fs.adsorbed_dye.properties[0].flow_vol,
+                to_units=pyunits.m**3 / pyunits.hr,
+            )
+        )
+        adsorbed_dye_mass_flow = m.fs.adsorbed_dye.flow_mass_phase_comp[
+            0, "Liq", "dye"
+        ].value
+        adsorbed_tds_mass_flow = m.fs.adsorbed_dye.flow_mass_phase_comp[
+            0, "Liq", "tds"
+        ].value
```
adam-a-a, line 1051: Generally better to wrap in `value()` rather than check for `.value`.

adam-a-a (approved):
Although I had a few suggestions that we should consider in subsequent PRs and one or two minor things that could be addressed now (e.g., delete commented code), I approve.

Issue #1279, opened 2024-01-24
Feature suggestions for OLI Cloud API


veccp:
### Description

This issue tracks tasks to help with OLI Cloud API development. 

**Fixes for March 2024 release:**
- [x] Enable all flash calculations (isenthalpic, etc.) (#1296)
- [x] Enable any input parameters to be surveyed (pH, etc.) (#1296)
- [x] Enable reconciliation options (pH, alkalinity, total inorganic carbon, etc.) (#1296)
- [x] Enable other thermodynamic frameworks as required for corrosion analysis (#1296)
- [x] Exception handling for empty results (#1296)
- [x] Apply absolute/relative status to all survey variables (not just solutes) (#1296)
- [x] Address workflow bottlenecks for true/apparent species (#1296)
- [x] Revision of extracting/handling data from JSON output (#1296)

_Tasks last updated: March 14, 2024_

PR #1296, opened 2024-02-08
OLI API Flash update

veccp:
## Fixes/Resolves:

This addresses some issues of #1279, dealing mainly with the Flash and OLIApi classes and adding test coverage for new functionality.

E.g.: move more general functions out of Flash class, rework Client and Flash frontend, improve compatibility with async #1303...

## Summary/Motivation:

To allow for other flash calculations and frameworks to be used with OLI Cloud, some streamlining of functions was necessary.

- Streamline data extraction functions
- Enable `Aqueous (H+ ion)` thermo framework (required for corrosion analyzer)
- Enable `XSC` private databank (required for surface complexation calculations, silica scaling)
- Modify OLI stream output data storage: storing samples as array (no repetition of JSON keys for each sample)
- Apply same modification method to all survey variables (either additive - "relative" or substitutive - "absolute")
- Add JSON writing function
- Update OLI API tutorial `incorporating_oli_calculations`
- Clarify that Water Analysis is not required if you already have apparent species
- Consistency and clarity improvements for some doc strings

### Legal Acknowledgement


watertap/tools/oli_api/client.py
```diff
@@ -104,100 +104,108 @@ def __exit__(self, exc_type=None, exc_value=None, traceback=None):
...
+        if isinstance(chemistry_source, (str, Path)):
+            chemistry_source = Path(chemistry_source).resolve()
+            if not chemistry_source.is_file():
                 raise OSError(
-                    "Could not find requested path to file. Please "
-                    "check that this path to file exists."
+                    "Could not find requested path to file. "
+                    + "Check that this path to file exists."
                 )
+            e = "upload"
```
bknueven: Not sure where `e` is used. If it is used please use a less terse variable name.

watertap/tools/oli_api/client.py
```diff
@@ -104,100 +104,108 @@ def __exit__(self, exc_type=None, exc_value=None, traceback=None):
...
+        thermo_framework,
+        private_databanks,
+        phases,
+        model_name,
+    ):
         """
-        Creates dict for chemistry-builder to later generate a DBS file id.
+        Creates dict for chemistry-builder to later generate a DBS file ID.
 
         :param chemistry_source: path (str), dict, or state block containing chemistry info
```
bknueven: This docstring should be updated.

watertap/tools/oli_api/client.py
```diff
@@ -104,100 +104,108 @@ def __exit__(self, exc_type=None, exc_value=None, traceback=None):
...
-            raise IOError(
-                f" Unable to create DBS dict from empty {type(chemistry_source)}."
-            )
-        else:
-            try:
+        if isinstance(chemistry_source, (list, dict)):
+            if len(chemistry_source) != 0:
                 solute_list = [
                     {"name": get_oli_name(solute)} for solute in chemistry_source
                 ]
```
bknueven:
```suggestion
        solute_list = [
            {"name": get_oli_name(solute)} for solute in chemistry_source
        ]
```

watertap/tools/oli_api/client.py
```diff
@@ -104,100 +104,108 @@ def __exit__(self, exc_type=None, exc_value=None, traceback=None):
...
-            except AttributeError:
-                solute_list = [{"name": solute.oli_name} for solute in chemistry_source]
-
+            if not solute_list:
+                raise IOError(
+                    "Chemistry input must contain OLI-compatible solute names."
+                )
```
bknueven:
```suggestion
    if not solute_list:
        raise RuntimeError(
            "Chemistry input must contain OLI-compatible solute names."
        )
```
Is the message here sensible?

watertap/tools/oli_api/client.py
```diff
@@ -368,103 +326,116 @@ def _delete_dbs_file(self, dbs_file_id):
...
+            url = f"{base_url}/file/{dbs_file_id}/{flash_method}"
+        elif flash_method in valid_post_flashes:
+            mode = "POST"
+            url = f"{base_url}/flash/{dbs_file_id}/{flash_method}"
+            headers = self.credential_manager.update_headers(
+                {"Content-Type": "application/json"},
+            )
+        else:
+            valid_flashes = [*valid_get_flashes, *valid_post_flashes]
             raise IOError(
```
bknueven: Not sure this is an `IOError` -- I would just use `RuntimeError` here.

watertap/tools/oli_api/client.py
```diff
@@ -368,103 +326,116 @@ def _delete_dbs_file(self, dbs_file_id):
...
+                    result_link,
+                    headers=headers,
+                ).json()
+                last_poll_time = time.time()
+                request_iter += 1
+                poll_status = self.check_result(req_result)
+                if poll_status in ["PROCESSED", "FAILED"]:
+                    if req_result["data"]:
+                        return req_result["data"]
                     else:
```
bknueven:
```suggestion
        for _ in range(max_request):
            # wait a bit before trying the first time
            time.sleep(poll_time)
            req_result = requests.get(
                result_link,
                headers=headers,
            ).json()
            poll_status = self.check_result(req_result)
            if poll_status in ["PROCESSED", "FAILED"]:
                if req_result["data"]:
                    return req_result["data"]
                else:
```

watertap/tools/oli_api/client.py
```diff
@@ -368,103 +326,116 @@ def _delete_dbs_file(self, dbs_file_id):
...
-                        poll_timer = time.time() - last_poll_time
-                else:
-                    raise RuntimeError("Poll limit exceeded.")
+                        raise IOError(" Poll returned empty data.")
+                elif poll_status == "ERROR":
+                    raise RuntimeError(f"Call failed with status {req_result['code']}")
+            poll_timer = time.time() - last_poll_time
+        raise RuntimeError("Poll limit exceeded.")
```
bknueven:
```suggestion
                    raise IOError(" Poll returned empty data.")
            elif poll_status == "ERROR":
                raise RuntimeError(f"Call failed with status {req_result['code']}")
        raise RuntimeError("Poll limit exceeded.")
```

watertap/tools/oli_api/flash.py
```diff
@@ -116,64 +114,72 @@ def __init__(
...
+        values = list(product(*(survey_arrays.values())))
         _name = lambda k: get_oli_name(k) if get_oli_names else k
-
-        i = 0
-        survey = {}
-        for v in values:
-            survey[i] = {_name(keys[j]): v[j] for j in range(len(v)) for j in num_keys}
-            i = i + 1
-        _logger.info(f"Survey contains {len(survey)} items.")
+        survey = {_name(keys[i]): [val[i] for val in values] for i in range(len(keys))}
```
bknueven:
```suggestion
        survey = {_name(key): [val[idx] for val in values] for idx, key in enumerate(survey_arrays)}
```

watertap/tools/oli_api/flash.py
```diff
@@ -116,64 +114,72 @@ def __init__(
...
 
-        :return survey: dictionary containing each point in survey
+        :return survey: dictionary for product of survey variables and values
         """
 
-        keys = list(survey_arrays.keys())
-        num_keys = range(len(keys))
-        values = product(*(survey_arrays.values()))
-
+        keys = [get_oli_name(k) if get_oli_names else k for k in survey_arrays]
```
bknueven:
```suggestion
```

watertap/tools/oli_api/flash.py
```diff
@@ -216,279 +204,346 @@ def _set_prescaling_calculation_mode(self, use_scaling_rigorous):
...
+                data = data[key]
+            nested_data = data
+            unit = nested_data["unit"]
+            # sample nested data
+            if "value" in data:
+                sampled_data = [nested_data["value"][s] for s in samples]
+            elif "values" in nested_data:
+                sampled_data = {
+                    k: [v[s] for s in samples] for k, v in nested_data["values"].items()
+                }
```
bknueven: No `else` -- `sampled_data` won't exist below.

Should catch the `else`.
```suggestion
                }
            else:
                raise RuntimeError("a meaningful message")
```

watertap/tools/oli_api/flash.py
```diff
@@ -216,279 +204,346 @@ def _set_prescaling_calculation_mode(self, use_scaling_rigorous):
...
+            # filter sampled data
+            filtered_values = deepcopy(sampled_data)
+            if filter_zero:
+                if isinstance(sampled_data, dict):
+                    for k, v in sampled_data.items():
+                        if not any(val for val in v):
```
bknueven:
```suggestion
                        if all(val == 0 for val in v):
```

watertap/tools/oli_api/flash.py
```diff
@@ -216,279 +204,346 @@ def _set_prescaling_calculation_mode(self, use_scaling_rigorous):
...
-                        extracted_properties[i][prop] = result
-                        _logger.debug(result)
+                prop_label = deepcopy(prop)
+                for path in paths[prop]:
+                    phase = None
+                    label = deepcopy(prop)
+                    if "total" in path:
+                        phase = "total"
+                    elif "phases" in path:
+                        phase = path[path.index("phases") + 1]
```
bknueven:
```suggestion
                    label = deepcopy(prop)
                    if "total" in path:
                        phase = "total"
                    elif "phases" in path:
                        phase = path[path.index("phases") + 1]
                    else:
                        phase = None
```

bknueven (commented):
Some initial comments.

watertap/tools/oli_api/flash.py
```diff
@@ -216,279 +204,346 @@ def _set_prescaling_calculation_mode(self, use_scaling_rigorous):
...
+        :return json_file: string for full path of write file
         """
 
-        _logger.debug("Saving file...")
-        yaml_file = f"{file_name}.yaml"
-        with open(yaml_file, "w", encoding="utf-8") as f:
-            yaml.dump(flash_output, f, indent=4)
-        _logger.info(f"{yaml_file} saved to working directory.")
-        return yaml_file
+        json_file = Path(f"./{file_name}.json").resolve()
```
bknueven:
```suggestion
        json_file = Path(file_name).with_suffix(".json")
```
Per @lbianchi-lbl

watertap/tools/oli_api/flash.py
```diff
@@ -216,279 +204,346 @@ def _set_prescaling_calculation_mode(self, use_scaling_rigorous):
...
+        _logger.info(f"Saving content to {json_file}")
+        with open(json_file, "w", encoding="utf-8") as f:
```
bknueven:
```suggestion
        _logger.info(f"Saving content to {file_name}")
        with open(file_name, "w", encoding="utf-8") as f:
```
Or maybe we should just trust the user, even if they didn't have a `*.json` suffix. I prefer the control...

watertap/tools/oli_api/flash.py
```diff
@@ -75,420 +73,1163 @@
...
+
         if interactive_mode:
             _logger.setLevel(logging.INFO)
         else:
             _logger.setLevel(logging.DEBUG)
 
-    def build_survey(self, survey_arrays, get_oli_names=False, file_name=None):
+    def configure_water_analysis(
+        self,
+        inflows,
```
adam-a-a:
```suggestion
        inflows=None,
```
As is, if I call this method and don't provide inflows, I will get a different exception related to `configure_water_analysis` missing a positional argument. That is, the intended `RuntimeError: Inflows must be defined for Water Analysis.`  exception, currently at line 160, would never raise.

adam-a-a: In fact, one question I have is why did you make this a method, and why is `inflows` required at this stage?

veccp: I made this method to make it more user friendly to configure and save the input JSON for water analysis based on the arguments provided - `inflows` is required because it is essential to that JSON input

tutorials/incorporating_oli_calculations.ipynb
```diff
@@ -234,44 +299,86 @@
...
    "source": [
     "if credential_manager:\n",
+    "    # OLIApi is used as a context manager\n",
     "    with OLIApi(credential_manager) as oliapi:\n",
     "        # create a new DBS file\n",
+    "        # alternative thermo_frameworks and databanks are available.\n",
+    "        \n",
+    "        # use keep_file = True to save a DBS file ID on the Cloud\n",
+    "        # for use in more than one session\n",
     "        dbs_file_id = oliapi.get_dbs_file_id(\n",
```
adam-a-a:
```suggestion
    "        dbs_file_id = oliapi.generate_dbs_file(\n",
```

adam-a-a: This will result in an error as is given the api changes in this PR.

tutorials/incorporating_oli_calculations.ipynb
```diff
@@ -234,44 +299,86 @@
...
-    "            chemistry_source=source_water[\"components\"],\n",
+    "            inflows=inflows,\n",
+    "            thermo_framework=\"MSE (H3O+ ion)\",\n",
     "            phases=[\"liquid1\", \"solid\"],\n",
+    "            model_name=\"test\",\n",
+    "            keep_file=true,\n",
```
adam-a-a:
```suggestion
    "            keep_file=True,\n",
```

adam-a-a: As is, this also produces an error. @lbianchi-lbl I definitely think it is important to incorporate testing for this notebook in our GH checks at some point, since things can break easily if API changes are not also reflected here.

tutorials/incorporating_oli_calculations.ipynb
```diff
@@ -234,43 +296,85 @@
...
+    "            reconciliation=\"ReconcilePh\",\n",
+    "            ph=5,\n",
+    "            allow_solids=True,\n",
+    "            file_name=\"water_analysis_inputs\",\n",
     "        )\n",
-    "    \n",
-    "        # create water analysis case\n",
-    "        water_analysis_base_case = f.run_flash(\n",
+    "        # run Water Analysis flash calculation survey as specified        \n",
+    "        stream_output = flash.run_flash(\n",
```
adam-a-a: Once this cell finishes running, I get a prompt for entering y/n but I have no information about what I am saying yes or no to. We should change this so that what the user is being prompted for is clear. The same goes in the following cell, which has another separate issue that I'll comment on.

tutorials/incorporating_oli_calculations.ipynb
```diff
@@ -234,43 +296,85 @@
...
+    "        isothermal_input = flash.configure_flash_analysis(\n",
+    "            apparent_species, \"isothermal\", file_name=\"isothermal_inputs\"\n",
+    "        )\n",
+    "        # specify a new survey to compare apparent species\n",
+    "        isothermal_survey = build_survey(\n",
+    "            {\"NaCl\": linspace(0, 1e6, 10)},\n",
+    "            get_oli_names=True,\n",
+    "            file_name=\"isothermal_survey\",\n",
+    "        )\n",
+    "        flash.run_flash(\n",
```
adam-a-a: Running this `run_flash` call will result in an error. I also came across this error in my own script when trying to run this functionality. Here's the traceback:
```
TypeError                                 Traceback (most recent call last)
Cell In[11], [line 20](vscode-notebook-cell:?execution_count=11&line=20)
     [14](vscode-notebook-cell:?execution_count=11&line=14) # specify a new survey to compare apparent species
     [15](vscode-notebook-cell:?execution_count=11&line=15) isothermal_survey = build_survey(
     [16](vscode-notebook-cell:?execution_count=11&line=16)     {"NaCl": linspace(0, 1e6, 10)},
     [17](vscode-notebook-cell:?execution_count=11&line=17)     get_oli_names=True,
     [18](vscode-notebook-cell:?execution_count=11&line=18)     file_name="isothermal_survey",
     [19](vscode-notebook-cell:?execution_count=11&line=19) )
---> [20](vscode-notebook-cell:?execution_count=11&line=20) flash.run_flash(
     [21](vscode-notebook-cell:?execution_count=11&line=21)     "isothermal",
     [22](vscode-notebook-cell:?execution_count=11&line=22)     oliapi,
     [23](vscode-notebook-cell:?execution_count=11&line=23)     dbs_file_id,
     [24](vscode-notebook-cell:?execution_count=11&line=24)     isothermal_input,
     [25](vscode-notebook-cell:?execution_count=11&line=25)     isothermal_survey,
     [26](vscode-notebook-cell:?execution_count=11&line=26)     file_name="isothermal_outputs",
     [27](vscode-notebook-cell:?execution_count=11&line=27) )

File [c:\users\adam\my-nawi-hub\watertap\watertap\tools\oli_api\flash.py:889](file:///C:/users/adam/my-nawi-hub/watertap/watertap/tools/oli_api/flash.py:889), in Flash.run_flash(self, flash_method, oliapi_instance, dbs_file_id, json_input, survey, file_name)
    [883](file:///C:/users/adam/my-nawi-hub/watertap/watertap/tools/oli_api/flash.py:883) for idx in range(num_samples):
    [884](file:///C:/users/adam/my-nawi-hub/watertap/watertap/tools/oli_api/flash.py:884)     _logger.info(f"Flash sample #{idx+1} of {num_samples}")
    [885](file:///C:/users/adam/my-nawi-hub/watertap/watertap/tools/oli_api/flash.py:885)     requests_to_process.append(
    [886](file:///C:/users/adam/my-nawi-hub/watertap/watertap/tools/oli_api/flash.py:886)         {
    [887](file:///C:/users/adam/my-nawi-hub/watertap/watertap/tools/oli_api/flash.py:887)             "flash_method": flash_method,
    [888](file:///C:/users/adam/my-nawi-hub/watertap/watertap/tools/oli_api/flash.py:888)             "dbs_file_id": dbs_file_id,
--> [889](file:///C:/users/adam/my-nawi-hub/watertap/watertap/tools/oli_api/flash.py:889)             "input_params": self.get_clone(
    [890](file:///C:/users/adam/my-nawi-hub/watertap/watertap/tools/oli_api/flash.py:890)                 flash_method, json_input, idx, survey
    [891](file:///C:/users/adam/my-nawi-hub/watertap/watertap/tools/oli_api/flash.py:891)             ),
    [892](file:///C:/users/adam/my-nawi-hub/watertap/watertap/tools/oli_api/flash.py:892)         }
    [893](file:///C:/users/adam/my-nawi-hub/watertap/watertap/tools/oli_api/flash.py:893)     )
    [894](file:///C:/users/adam/my-nawi-hub/watertap/watertap/tools/oli_api/flash.py:894) processed_requests = oliapi_instance.process_request_list(
    [895](file:///C:/users/adam/my-nawi-hub/watertap/watertap/tools/oli_api/flash.py:895)     requests_to_process,
    [896](file:///C:/users/adam/my-nawi-hub/watertap/watertap/tools/oli_api/flash.py:896)     # burst_job_tag=burst_job_tag,
    [897](file:///C:/users/adam/my-nawi-hub/watertap/watertap/tools/oli_api/flash.py:897)     # max_concurrent_processes=max_concurrent_processes,
    [898](file:///C:/users/adam/my-nawi-hub/watertap/watertap/tools/oli_api/flash.py:898)     # batch_size=batch_size,
    [899](file:///C:/users/adam/my-nawi-hub/watertap/watertap/tools/oli_api/flash.py:899) )
    [900](file:///C:/users/adam/my-nawi-hub/watertap/watertap/tools/oli_api/flash.py:900) _logger.info("Completed running flash calculations")

File [c:\users\adam\my-nawi-hub\watertap\watertap\tools\oli_api\flash.py:963](file:///C:/users/adam/my-nawi-hub/watertap/watertap/tools/oli_api/flash.py:963), in Flash.get_clone(self, flash_method, json_input, index, survey)
    [961](file:///C:/users/adam/my-nawi-hub/watertap/watertap/tools/oli_api/flash.py:961)     _logger.warning(f"Survey key {k} not found in JSON input.")
    [962](file:///C:/users/adam/my-nawi-hub/watertap/watertap/tools/oli_api/flash.py:962) if self.relative_inflows:
--> [963](file:///C:/users/adam/my-nawi-hub/watertap/watertap/tools/oli_api/flash.py:963)     if "value" in d[k]:
    [964](file:///C:/users/adam/my-nawi-hub/watertap/watertap/tools/oli_api/flash.py:964)         d[k]["value"] += v[index]
    [965](file:///C:/users/adam/my-nawi-hub/watertap/watertap/tools/oli_api/flash.py:965)     else:

TypeError: argument of type 'float' is not iterable

```

tutorials/incorporating_oli_calculations.ipynb
```diff
@@ -284,7 +388,7 @@
    "source": [
     "# 7. Extract Filtered Output\n",
```
adam-a-a: Previous step was step 5, and now we skip to step 7. Change to step 6, or add step 6 in before this.

tutorials/incorporating_oli_calculations.ipynb
```diff
@@ -297,22 +401,23 @@
...
+    "import json\n",
+    "\n",
     "if credential_manager:\n",
-    "    extracted_properties = f.extract_properties(\n",
-    "        isothermal_survey_result,\n",
-    "        properties,\n",
-    "        filter_zero=True,\n",
-    "        file_name=\"properties\",\n",
-    "    )"
+    "    with open(\"isothermal_outputs\", \"rb\") as json_output:\n",
```
adam-a-a:
```suggestion
    "    with open(\"isothermal_outputs.json\", \"rb\") as json_output:\n",
```

tutorials/incorporating_oli_calculations.ipynb
```diff
@@ -297,22 +401,23 @@
...
+    "        results = json.load(json_output)[\"result\"]\n",
+    "\n",
+    "    print(\"Available keys:\")\n",
+    "    for k in result:\n",
```
adam-a-a:
```suggestion
    "    for k in results:\n",
```

tutorials/incorporating_oli_calculations.ipynb
```diff
@@ -297,22 +401,23 @@
...
+    "        print(k)\n",
+    "\n",
+    "    result = data[\"prescalingTendencies\"]\n",
```
adam-a-a:
```suggestion
    "    result = results[\"prescalingTendencies\"]\n",
```

watertap/tools/oli_api/flash.py
```diff
@@ -75,420 +73,1163 @@
...
+                        "value": get_oli_name(base_titrant),
+                    },
+                ]
+            )
+        if "Alk" in reconciliation:
+            additional_req_args.append([alkalinity, alkalinity_ph, alkalinity_titrant])
+            if not alkalinity_titrant:
+                alkalinity_titrant = "H2SO4"
+            if not alkalinity_ph:
+                alkalinity_ph = 4.5
```
adam-a-a, line 245: Any reason why this is hard-coded to 4.5?

adam-a-a: I am guessing this is just a ballpark default if the user doesn't provide any value for this. Should we let the user know that this is the case with logger output?

watertap/tools/oli_api/flash.py
```diff
@@ -75,420 +73,1163 @@
...
 
-        :return input_list: deepcopy of self.water_analysis_input_list
+        """
+        Configure Flash Analysis JSON input.
+
+        :param inflows: dictionary of solutes, of the form {"unit": unit, "values": {solute: concentration}}
+        :param flash_method: string for flash calculation name
+        :param temperature: float for temperature in Kelvins
+        :param pressure: float for pressure in Pascals
+        :param calculated_variable: string for variable to calculate, such as temperature or pressure, used in 'bubblepoint', 'dewpoint', 'vapor-amount', 'vapor-fraction', and 'isochoric' flashes
```
adam-a-a, line 393: Haven't looked that far ahead yet in the PR, but we should make sure to include logic to check this arg if it is not None and ensure that only valid options can be specified. Otherwise, raise an exception stating the erroneous specification by the user and then display the valid options.

watertap/tools/oli_api/flash.py
```diff
@@ -75,420 +73,1163 @@
...
+
+        pres_input = {
+            "unit": self.input_unit_set["pressure"]["oli_unit"],
+            "value": 101325,
+        }
+        if pressure:
+            if float(pressure):
+                pres_input.update({"value": float(pressure)})
+            else:
+                raise RuntimeError(f"Invalid pressure: {pressure}. Expected number")
```
adam-a-a: Should this be `ValueError`?

watertap/tools/oli_api/flash.py
```diff
@@ -75,420 +73,1163 @@
...
+
+        temp_input = {
+            "unit": self.input_unit_set["temperature"]["oli_unit"],
+            "value": 273.15,
+        }
+        if temperature:
+            if float(temperature):
+                temp_input.update({"value": float(temperature)})
+            else:
+                raise RuntimeError(
```
adam-a-a: Should this be `ValueError`?

watertap/tools/oli_api/flash.py
```diff
@@ -75,420 +73,1163 @@
...
+        if flash_method in [
+            "bubblepoint",
+            "dewpoint",
+            "vapor-amount",
+            "vapor-fraction",
+            "isochoric",
+        ]:
+            if calculated_variable:
+                _logger.info(
+                    f"{flash_method} will calculate {calculated_variable} as its variable"
```
adam-a-a, line 479: As mentioned in previous comment, I think this is where you want to add some check(s) to ensure that the specified calculated variable is valid. Also, what if I want to enter more than one calculated variable? A test to verify this functionality would be good.

watertap/tools/oli_api/flash.py
```diff
@@ -75,420 +73,1163 @@
...
+
+        if flash_method == "isenthalpic":
+            enth_input = {
+                "unit": self.input_unit_set["enthalpy"]["oli_unit"],
+                "value": None,
+            }
+            if float(enthalpy):
+                enth_input.update({"value": float(enthalpy)})
+            else:
+                raise RuntimeError(f"Invalid enthalpy: {enthalpy}. Expected number")
```
adam-a-a: ValueError? Same applies for all exceptions related to case where number was expected but not provided

watertap/tools/oli_api/flash.py
```diff
@@ -75,420 +73,1163 @@
...
-    def _oli_units(self, conc, unit):
-        """
-        Converts concentrations between specified units.
+        input_dict["inflows"] = inflows
 
-        :param conc: concentration value for a solute
-        :param unit: concentration unit for a solute
+        if flash_method == "corrosion-rates":
+            _logger.info(
+                f"Ensure DBS file uses 'AQ' thermodynamic framework to use Corrosion Analyzer"
```
adam-a-a, line 578: Can we raise exception instead here if thermodynamic framework is not AQ? I see you do that in `client` for calling corrosion-contact-surfaces function, but it is unclear whether that will _always_ precede this flash method call (I don't think it necessarily would for the unknowing user.)

Furthermore, we should have at least one test demonstrating a simple use-case that includes using `get_corrosion_contact_surfaces` and this `corrosion_rates` flash method to make sure we observe the expected behavior from this functionality.

watertap/tools/oli_api/flash.py
```diff
@@ -75,420 +73,1163 @@
...
+                if k in d:
+                    pass
+                elif k in d["inflows"]["values"]:
+                    d = d["inflows"]["values"]
+                elif k in d["corrosionParameters"]:
+                    d = d["corrosionParameters"]
+                else:
+                    _logger.warning(f"Survey key {k} not found in JSON input.")
+                if self.relative_inflows:
+                    if "value" in d[k]:
```
adam-a-a: The error message that I pasted in an earlier comment is attributed to this `if` logic, since `d[k]` in the cases I was testing was a float. 
```suggestion
                    if isinstance(d[k], float):
                        d[k] += v[index]
                    elif "value" in d[k]:
```

watertap/tools/oli_api/flash.py
```diff
@@ -75,420 +73,1163 @@
...
+                        d[k]["value"] += v[index]
+                    else:
+                        d[k] += v[index]
+                else:
+                    if "value" in d[k]:
```
adam-a-a:
```suggestion
                    if isinstance(d[k], float):
                        d[k] = v[index]
                    elif "value" in d[k]:
```

adam-a-a (changes requested):
There are a few bugs that I spotted after trying this on my local machine. I added suggestions based on local fixes that seemed to work on my end. Additionally, I dropped several comments that most importantly point out specific areas where I think testing should be included.

tutorials/incorporating_oli_calculations.ipynb
lbianchi-lbl: @adam-a-a I've added notebook testing to the OLI API GHA check suite, see https://github.com/watertap-org/watertap/issues/1298#issuecomment-1969208395.

watertap/tools/oli_api/flash.py
veccp: This is what OLI specifies as a default in their API documentation: https://devdocs.olisystems.com/group1/api-functions/wateranalysis#reconcile-ph-and-alkalinity

watertap/tools/oli_api/flash.py
veccp: The logic could be added to run_flash since a DBS file is already required for that function

watertap/tools/oli_api/flash.py
```diff
@@ -75,420 +73,1163 @@
...
+                    d = d["inflows"]["values"]
+                elif k in d["corrosionParameters"]:
+                    d = d["corrosionParameters"]
+                else:
+                    _logger.warning(f"Survey key {k} not found in JSON input.")
+                if self.relative_inflows:
+                    if "value" in d[k]:
+                        d[k]["value"] += v[index]
+                    else:
+                        d[k] += v[index]
```
bknueven:
```suggestion
                    if isinstance(d[k], dict):
                        d[k]["value"] += v[index]
                    else:
                        d[k] += v[index]
```

watertap/tools/oli_api/flash.py
```diff
@@ -75,420 +73,1163 @@
...
+                else:
+                    if "value" in d[k]:
+                        d[k]["value"] = v[index]
+                    else:
+                        d[k] = v[index]
```
bknueven:
```suggestion
                    if isinstance(d[k], dict):
                        d[k]["value"] = v[index]
                    else:
                        d[k] = v[index]
```

bknueven (commented):
Slightly better version of @adam-a-a's fix.

watertap/tools/oli_api/flash.py
bknueven: fixed in [0900a69](https://github.com/watertap-org/watertap/pull/1296/commits/0900a69dcfa09994e9b77982c579e68b9a1d04be)

watertap/tools/oli_api/flash.py
bknueven: fixed in [0900a69](https://github.com/watertap-org/watertap/pull/1296/commits/0900a69dcfa09994e9b77982c579e68b9a1d04be)

tutorials/incorporating_oli_calculations.ipynb
veccp: The message (WaterTAP will delete 0 DBS files) is in the logger, but if it's more clear to have a message in line with the prompt then that can be done

tutorials/incorporating_oli_calculations.ipynb
veccp: Step 5 and 6 are in the same box, but I will make this more clear

watertap/tools/oli_api/client.py
```diff
@@ -276,7 +276,11 @@ def dbs_file_cleanup(self, dbs_file_ids=None):
 
         if dbs_file_ids is None:
             dbs_file_ids = self.get_user_dbs_file_ids()
-        r = input(f"WaterTAP will delete {len(dbs_file_ids)} DBS files: [y]/n: ") if self.interactive_mode else ""
+        r = (
+            input(f"WaterTAP will delete {len(dbs_file_ids)} DBS files: [y]/n: ")
+            if self.interactive_mode
+            else ""
+        )
```
lbianchi-lbl: I'm really not sure, but I think the logic here might not behave the way we want to, i.e. `input()` is run even when `self.interactive_mode` is false-y.

Maybe we can try the following:

```suggestion
        if self.interactive_mode:
            r = input(...)
        else:
            r = ""
```

lbianchi-lbl: I guess a better solution would be to abstract away the references to `input()` with a dedicated method:

```py
    def _prompt(self, msg: str, default: str) -> str:
        if self.interactive_mode:
            return input(msg)
        else:
            _logger.info(msg)
            return default
```

And then in the rest of the class, call `self._prompt()` without any further logic needed

adam-a-a:
> Slightly better version of @adam-a-a's fix.

@bknueven I realized this (check for dict and just have an if-else) as I was writing the suggestion and literally thought to myself "ah, Ben will probably suggest the better way if we need it."  =)

watertap/tools/oli_api/flash.py
```diff
@@ -75,420 +73,1175 @@
...
+                    else:
+                        d[k] += v[index]
+                        val = d[k]
+                else:
+                    if isinstance(d[k], dict):
+                        d[k]["value"] += v[index]
+                        val = d[k]["value"]
+                    else:
+                        d[k]["value"] += v[index]
+                        val = d[k]["value"]
```
adam-a-a: Two things: (1) `if not self.relative_inflows`, shouldn't the value be equal instead of additive? And (2) it looks like you accidentally kept in reference to a "value" key for types that are not dicts. Here's the suggestion to reflect these items:
```suggestion
                    if isinstance(d[k], dict):
                        d[k]["value"] = v[index]
                        val = d[k]["value"]
                    else:
                        d[k] = v[index]
                        val = d[k]
```

veccp: Yes, thanks for catching those!

watertap/tools/oli_api/client.py
veccp: I added this to `OLIApi` for now at least. Will add it to #1300 for `CredentialManager` as well

tutorials/incorporating_oli_calculations.ipynb
```diff
@@ -487,9 +484,9 @@
  "metadata": {
   "celltoolbar": "Slideshow",
   "kernelspec": {
-   "display_name": "Python 3 (ipykernel)",
+   "display_name": "watertap-dev",
    "language": "python",
-   "name": "python3"
+   "name": "watertap-dev"
```
lbianchi-lbl: @veccp I think this is causing failures in the oli-api-ci workflows. Sorry about the confusion. I'll try to open a PR next week so that the CI jobs for notebooks are consistent with the `watertap-dev` name we have in the documentation (and will cause an error otherwise).

lbianchi-lbl: This would be #1320

veccp: @lbianchi-lbl Sounds good, are you saying the notebook kernel should be `watertap-dev` and not `Python 3 (ipykernel)`?

lbianchi-lbl: Yes, the current `watertap-dev` name is OK, and will be the supported one going forward once #1320 is merged.

tutorials/incorporating_oli_calculations.ipynb
```diff
@@ -233,45 +299,90 @@
...
+    "    )\n",
+    "    # NOTE: liquid2 (non-aqueous) and vapor phases are available but not yet fully supported.\n",
+    "\n",
+    "    # if a DBS file has been retained from a previous session,\n",
+    "    # its flash history and chemistry information can be summarized.\n",
+    "    file_summary = oliapi.get_dbs_file_summary(dbs_file_id)\n",
+    "\n",
+    "    # save chemistry information\n",
+    "    chemistry_info = file_summary[\"chemistry_info\"]\n",
+    "    write_output(chemistry_info[\"result\"], \"chemistry_info\")\n",
```
adam-a-a: @veccp Getting an error here:
image

tutorials/incorporating_oli_calculations.ipynb
```diff
@@ -233,45 +299,90 @@
...
-    "            oliapi,\n",
-    "            dbs_file_id,\n",
-    "            water_analysis_apparent_composition,\n",
-    "            survey,\n",
-    "            \"isothermal_composition_survey\",\n",
-    "        )"
+    "# Flash calculation example #1: Water Analysis survey with pH reconciliation.\n",
+    "\n",
+    "# OLIApi is used as a context manager\n",
+    "with OLIApi(credential_manager, interactive_mode=interactive_mode) as oliapi:\n",
```
adam-a-a:
```suggestion
    "with OLIApi(credential_manager, interactive_mode=False) as oliapi:\n",
```
as is, this will produce an error since interactive_mode no longer exists.

tutorials/incorporating_oli_calculations.ipynb
adam-a-a: This bit on getting the summary/flash history should be shown later, after walking through the main essentials.

tutorials/incorporating_oli_calculations.ipynb
```diff
@@ -233,45 +299,90 @@
...
+   "outputs": [],
+   "source": [
+    "# Flash calculation example #2: isothermal analysis survey\n",
+    "with OLIApi(credential_manager, interactive_mode=False) as oliapi:\n",
+    "\n",
+    "    # Other flash calculations use apparent species for inputs.\n",
+    "\n",
+    "    # These can be obtained in one step with get_apparent_species_from_true(),\n",
+    "    # and saved to a file and reloaded as needed.\n",
+    "    apparent_species = flash.get_apparent_species_from_true(\n",
```
adam-a-a, line 376: @veccp If I skip the problem above with dbs file summary and go straight to these calcs, I get the following error (I also got this error in my own script, which is what made me look back to this tutorial in its latest state to see if it was my doing or not): 
image

adam-a-a: Actually- this could still be my doing! I forgot to mention that I merged this PR with #1303 on a local branch and quickly tried to resolve merge conflicts while maintaining the key changes from both PRs. So I probably should just check this tutorial based on this PR's changes only....

veccp: @adam-a-a Thanks for the comments. There were a few bugs in the tutorial that I just fixed. It should work as is

lbianchi-lbl:
@lbianchi-lbl Fix (expected) failures in notebooks tests.

lbianchi-lbl:
The OLI API CI notebooks job is failing seemingly because of `interactive_mode=True); maybe some change has been inadvertently reverted in the notebook file? **EDIT** It seems so, judging from https://github.com/watertap-org/watertap/commit/b982cbd9f9beaa598d7a0c9d955e4f5ef5a90de4

image

lbianchi-lbl:
@adam-a-a For reference, this is how the OLI API CI checks look like:

image

lbianchi-lbl:
> @lbianchi-lbl Fix (expected) failures in notebooks tests.

This is done in bfcb306, which adds functionality to skip/xfail OLI notebooks if the credentials are not found in the environment:

image

lbianchi-lbl:
OK, I think I managed to get the Codecov upload to work:

image
image
image

lbianchi-lbl:
The OLI API CI jobs now also create a "pending" status check when the workflow is started: 
image

tutorials/incorporating_oli_calculations.ipynb
```diff
@@ -163,16 +153,60 @@
...
+    "    get_oli_names=True,\n",
+    "    mesh_grid=True,\n",
+    "    file_name=\"oli_survey.json\",\n",
+    ")\n",
+    "\n",
+    "# uncomment the following lines to view results:\n",
+    "#print(\"Survey using grid_mesh\")\n",
+    "#print(survey)\n",
+    "\n",
+    "# setting grid_mesh=False allows custom surveys to be defined, i.e., no Cartesian product is computed.\n",
```
OOAmusat: this should be ``mesh_grid=False``

watertap/tools/oli_api/tests/test_flash.py
OOAmusat, line 1: Is my understanding correct that this just goes off and checks that things run, without actually validating the results?

Just want to make sure I'm not missing anything

watertap/tools/oli_api/flash.py
```diff
@@ -75,420 +73,1175 @@
...
-        interactive_mode=True,
+        debug_level="INFO",
     ):
-        # set values based on inputs
-        self.water_analysis_properties = water_analysis_properties
         self.optional_properties = optional_properties
         self.input_unit_set = input_unit_set
         self.output_unit_set = output_unit_set
-        self.stream_output_options = stream_output_options
         self.relative_inflows = relative_inflows
```
OOAmusat, line 101: is there any case where the user may want to do different types relative vs absolute) for different survey variables? If that is the case, should we be thinking about making this a multi-dimensional vector where each survey has its own Boolean?

Just a thought for the future, not this PR.

watertap/tools/oli_api/flash.py
```diff
@@ -75,420 +73,1175 @@
...
+                        "value": get_oli_name(base_titrant),
+                    },
+                ]
+            )
+        if "Alk" in reconciliation:
+            additional_req_args.append([alkalinity, alkalinity_ph, alkalinity_titrant])
+            if not alkalinity_titrant:
+                alkalinity_titrant = "H2SO4"
+            if not alkalinity_ph:
+                alkalinity_ph = 4.5
```
OOAmusat, line 245: I think this should just thrown an error if the user does not supply a value. Naive users may assume that we have chosen that value in some scientific way and go with it.

Again, maybe for another PR.

watertap/tools/oli_api/flash.py
```diff
@@ -75,420 +73,1175 @@
...
-                    "unit": str(state_vars["units"]["pressure"]),
-                    "value": float(state_vars["pressure"]),
-                },
-                "inflows": self._extract_inflows(water_analysis_output, inflows_phase),
-            }
+        additional_params = {
+            "optionalProperties": dict(self.optional_properties),
+            "unitSetInfo": dict(self.output_unit_set),
+        }
+        if included_solids and excluded_solids:
```
OOAmusat, line 813: Is there a reason why both cannot be specified at the same time?

watertap/tools/oli_api/flash.py
```diff
@@ -75,420 +73,1175 @@
...
-                    if param["name"] in survey[i].keys():
-                        param.update({"value": survey[i][param["name"]]})
-            elif flash_method == "isothermal":
-                for param in survey[i]:
-                    if param in ["Temperature", "Pressure"]:
-                        modified_clone["params"][param.lower()]["value"] = survey[i][
-                            param
-                        ]
-                    elif param in modified_clone["params"]["inflows"]["values"]:
+                d = d["waterAnalysisInputs"]
```
OOAmusat, line 949: why did you remove the temperature/pressure catch here? If it is removed, then the relative/absolute value boolean for the survey definitely needs to be multi-dimensional.

If a user wants to sweep across temperature (more likely to be absolute) and concentration (more likely to be relative), how would this work? If T/P are handled explicitly as previously, we could maybe avoid that issue. As it is now, we definitely need to deal with it.

watertap/tools/oli_api/flash.py
veccp: Yes, probably for pH, some state variables that are usually within certain ranges, etc. I'll add it to the issue #1328

watertap/tools/oli_api/flash.py
veccp: Ok, this sounds like another good thing to add to #1328: config functions could be cleaned up.

watertap/tools/oli_api/flash.py
veccp: As per OLI: 

> params.excludedSolids and params.includedSolids cannot be specified at the same time.

watertap/tools/oli_api/flash.py
veccp: Makes sense - I'll make note to reimplement the catch for T/P and also pH

watertap/tools/oli_api/tests/test_flash.py
veccp: Yes

tutorials/incorporating_oli_calculations.ipynb
adam-a-a:
```suggestion
    "# setting mesh_grid=False allows custom surveys to be defined, i.e., no Cartesian product is computed.\n",
```

bknueven (approved):

adam-a-a (approved):
LGTM - we will resolve any outstanding issues in subsequent PRs. 

I want to reiterate that before we continue adding any new or altered functionality (or API changes in general), we should prioritize adding tests to cover the wide range of functionality introduced in this and previous PRs.

PR #1304, opened 2024-02-19
Add validity ranges and clean up water_prop_pack, seawater_prop_pack, and NaCl_T_dep_prop_pack

savannahsakhai:
## Summary/Motivation:
This PR cleans up three of the property models by reformatting the in-code citations and adding validity ranges for each of the properties. 

- Update the references for the NaCl_T_dep_prop_pack in the documentation
- Add the reference list to NaCl_T_dep_prop_pack
- Reformat citations and add validity ranges to water_prop_pack and seawater_prop_pack
- Remove commented out or unnecessary code from property functions in the water_prop_pack (that had been adapted from the seawater_prop_pack)

### Legal Acknowledgement


docs/technical_reference/property_models/NaCl_T_dep.rst
```diff
@@ -1,7 +1,7 @@
 NaCl Property Package with Temperature Dependence
 =================================================
 
-This package implements property relationships for an NaCl solution.
+This package implements property relationships for a NaCl solution.
```
adam-a-a:
```suggestion
This package implements property relationships for an NaCl solution.
```
I've always gone with reading this as "an N-a-c-l solution".

watertap/property_models/seawater_prop_pack.py
```diff
@@ -107,7 +107,8 @@ def build(self):
         mw_comp_data = {
             "H2O": 18.01528e-3,
             "TDS": 31.4038218e-3,
-        }  # average atomic weight of sea salt, based on
+        }
+        # average atomic weight of sea salt, based on
```
adam-a-a:
```suggestion
        # molecular weight of TDS is taken as the average atomic weight of sea salt, based on
```

watertap/property_models/seawater_prop_pack.py
```diff
@@ -1500,10 +1496,8 @@ def _therm_cond_phase(self):
             units=pyunits.W / pyunits.m / pyunits.K,
             doc="Thermal conductivity",
         )
-
-        def rule_therm_cond_phase(
-            b, p
-        ):  # thermal conductivity, eq. 13 in Sharqawy  et al. (2010)
+        # Sharqawy  et al. (2010), eq. 13, 0-180 C, 0-160 g/kg
```
adam-a-a, line 1499: It would be nice if we could show logger output, say, after a model was solved, to display warnings for any variables out of range. Just a thought--not something that I think should be in this PR if we did have some form of an implementation in mind.

docs/technical_reference/property_models/NaCl_T_dep.rst
savannahsakhai: So, it depends on whether we want to consider this a formal or informal use case because technically, according to ACS, the full element name should be used when choosing whether to use a or an. 
(Informal: "an N-a-c-l solution" or Formal:  "a sodium chloride solution")

watertap/property_models/seawater_prop_pack.py
```diff
@@ -246,7 +249,8 @@ def build(self):
             doc="Dynamic viscosity parameter 3 for term B",
         )
 
-        # diffusivity parameters, eq 6 in Bartholomew
+        # diffusivity parameters, 25 C
```
ElmiraShamlou, line 252: Could we consider incorporating temperature-dependent diffusivity into the seawater property package, similar to what's done in the NaCl_T_dep_prop package? It looks like we're using NaCl diffusivity approaches for seawater by substituting TDS for NaCl, but we're not adjusting for temperature changes. The impact of temperature on diffusivity is pronounced in systems with large temperature gradients like MD.

ElmiraShamlou (changes requested):

watertap/property_models/water_prop_pack.py
```diff
@@ -906,7 +908,7 @@ def _pressure_sat(self):
             doc="Saturation vapor pressure",
         )
 
-        # vapor pressure, eq. 5 and 6 in Nayar et al.(2016)
+        # Nayar et al.(2016), eq. 5 and 6, 0-180 C
         def rule_pressure_sat(b):
```
ElmiraShamlou, line 912: I suggest adding saturation temperature and saturation enthalpy to the water property package, alongside the existing saturation pressure (not necessarily in this PR). These additions would be beneficial for modeling heating systems

docs/technical_reference/property_models/NaCl_T_dep.rst
adam-a-a: I'm good with either "an NaCl solution" or "a sodium chloride solution", so long as it is _not_ changed to "a NaCl solution".

watertap/property_models/seawater_prop_pack.py
adam-a-a: I don't recall if the MIT thermodynamic library, the main source of the seawater prop pack, has temperature-dependent diffusivity. Also, our solute is "TDS" in this prop pack, so deciding on what the effective diffusivity should be, especially as a function of temperature, might take some thought. One approach could be to consider the composition from Millero (that was used to estimate molecular weight value) and derive some relationship via something like PHREEQC. An easier approach could be just using the relationship from NaCl with temp dependence here too as an optional calculation for temp-dependent diffusivity. That said, I haven't looked at the lit recently to see if there is a more obvious approach. 

Any ideas?

watertap/property_models/water_prop_pack.py
adam-a-a: I'd like to add that IDAES does have the IAPWS prop pack (https://idaes-pse.readthedocs.io/en/latest/reference_guides/model_libraries/generic/property_models/iapws95.html?highlight=iapws) which might have additional properties that would be useful.

watertap/property_models/seawater_prop_pack.py
savannahsakhai: The MIT thermodynamic library does not have this property. I agree that one approach could be to consider the composition from Millero and derive a relationship using PHREEQC. I will look into the literature to see if this has been done or if there is a better approach. 

I think as a band-aid fix for now we could replace the current calculation with the relationship from the NaCl with temperature dependence property package. This would at least represent the impact of temperature.

adam-a-a:
@savannahsakhai I think we can reserve the inclusion of saturation temperature and temperature-dependent diffusivity in a separate PR since we are prepping for a release and can get these straightforward changes in in the meantime.

hunterbarber (approved):
Changes to comment clean up are good. Convert any feature requests to corresponding issues.

ElmiraShamlou (approved):
LGTM

PR #1313, opened 2024-02-26
Correcting error checks for MCAS properties config combinations

### LINKED ISSUE #1305: MCAS `ElectricalMobilityCalculation.EinsteinRelation` does not allow for `DiffusivityCalculation.HaydukLaudie`
### Description

`ElectricalMobilityCalculation.EinsteinRelation` does not allow for `DiffusivityCalculation.HaydukLaudie` due to the check in line 1526 of MCAS

### Anything Else?

Need to change the check in line 1526 from looking for `self.params.config.diffusivity_data.keys()` for `DiffusivityCalculation`s that are functions of other parameters and do not require `self.params.config.diffusivity_data.keys()`. This is an easy fix to change the check through a couple different implementations, but not sure of which is the most future-proof, and not sure if this needs tested somewhere.

hunterbarber:
## Fixes/Resolves:
#1305

## Summary/Motivation:
Need to change the check in line 1526 from looking for `self.params.config.diffusivity_data.keys()` for `DiffusivityCalculation`s that are functions of other parameters and do not require `self.params.config.diffusivity_data.keys()`.

- Adjust error check

### Legal Acknowledgement

### DIFF
--- a/watertap/property_models/multicomp_aq_sol_prop_pack.py
+++ b/watertap/property_models/multicomp_aq_sol_prop_pack.py
@@ -1523,7 +1523,10 @@ class MCASStateBlockData(StateBlockData):
                         * pyunits.second**-1
                     )
             else:
-                if (p, j) not in self.params.config.diffusivity_data.keys():
+                if (
+                    self.params.config.diffus_calculation == DiffusivityCalculation.none
+                    and (p, j) not in self.params.config.diffusivity_data.keys()
+                ):
                     raise ConfigurationError(
                         """
                         Missing a valid diffusivity_data configuration to use EinsteinRelation 


hunterbarber:
So this isn't as straightforward as I initially thought.

So `diffus_phase_comp` can be calculated via `DiffusivityCalculation.none` which requires `diffusivity_data` or via `DiffusivityCalculation.HaydukLaudie` which calculates as a function of `molar_volume`. The `diffus_phase_comp` is used in the `ElectricalMobilityCalculation.EinsteinRelation` to calculate `elec_mobility_phase_comp`. This is where the past check (see changed files) wouldn't let you use `DiffusivityCalculation.HaydukLaudie` with `ElectricalMobilityCalculation.EinsteinRelation` because it required `diffusivity_data`, and `DiffusivityCalculation.HaydukLaudie` calculates  `diffus_phase_comp` without `diffusivity_data`.

So the challenge is that the check for `diffus_phase_comp` only logs warnings because some of the unit models only need this variable for select species which a hard `ConfigurationError` would block. However other properties like `equiv_conductivity_phase` use the sum of `elec_mobility_phase_comp` for all ions. As such `elec_mobility_phase_comp` is a hard `ConfigurationError` when it is not supplied for all of the ions in the ion set.

See my single edit. The current correction doesn't work for the edge case where `DiffusivityCalculation.HaydukLaudie` but `diffus_phase_comp` isn't calculated for all ions in the ion set, and I thought of a couple of other changes but they all have edge cases. Say converting the diffusivity to `ConfigurationError`s fixes the checks but no longer allows for that partial specification of diffusivity data. @adam-a-a @lbibl Any immediate ideas?

hunterbarber:
See #1305

PR #1316, opened 2024-02-28
Work around for issue #1312

### LINKED ISSUE #1312: Jupyter notebook for Parameter Sweep tutorial occasional cell timeout failures on Linux/Python 3.8
image

bknueven:
## Summary/Motivation:
We observed in #1314 that the notebook only seems to be failing on Python 3.8, but not Python 3.9+. We suspect the issue is with Python 3.8 itself. However, we don't see this issue when using the `multiprocessing` parallel back end for Python 3.8.

Since we still support Python 3.8, this PR would change the notebook to use the multiprocessing parallel back end and makes that the default for parameter sweep.

- Use `multiprocessing` parallel manager in example notebook
- Change default back end from `concurrent.futures` to `multiprocessing`
- Drive by: make the parallel factory more user friendly by not caring about case and catch a few other potential spellings, etc.

### Legal Acknowledgement

### DIFF
--- a/tutorials/parameter_sweep_demo.ipynb
+++ b/tutorials/parameter_sweep_demo.ipynb
@@ -221,7 +221,7 @@
     "        \"optimize_kwargs\": {\"solver\": solver, \"check_termination\": False},\n",
     "        \"initialize_function\": None,\n",
     "        \"initialize_kwargs\": {},\n",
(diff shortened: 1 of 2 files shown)


lbianchi-lbl (approved):

lbianchi-lbl:
I've tested these changes with the dedicated CI harness from #1314 (2f64b97) and it looks all green. Since the notebook tests pass, I'll go ahead and merge this in so that the CI checks can start working reliably again for the other PRs.

PR #1323, opened 2024-03-07
add MD1D

ElmiraShamlou:
Add Membrane distillation 1D
## Fixes/Resolves:

-
-

### Legal Acknowledgement


watertap/unit_models/MD/membrane_distillation_0D.py
```diff
@@ -406,24 +409,30 @@ def _add_mass_transfer(self):
...
-            )
+
+            if p == "Liq":
+                return (
+                    b.cold_ch.mass_transfer_term[t, "Liq", j]
+                    == -b.hot_ch.mass_transfer_term[t, "Liq", j]
+                )
+
+            else:
+                b.cold_ch.mass_transfer_term[t, p, j].fix(0)
```
adam-a-a, line 420: How is `hot_ch.mass_transfer_term` handled for vapor?

adam-a-a: Is it essentially just a stale variable?

watertap/unit_models/MD/membrane_distillation_0D.py
```diff
@@ -406,24 +409,30 @@ def _add_mass_transfer(self):
...
         )
         def eq_permeate_production(b, t, p, j):
             if j == "H2O":
                 return (
-                    b.cold_ch.mass_transfer_term[t, p, j] == b.area * b.flux_mass_avg[t]
+                    b.hot_ch.mass_transfer_term[t, p, j] == -b.area * b.flux_mass_avg[t]
                 )
             else:
-                b.cold_ch.mass_transfer_term[t, p, j].fix(0)
+                b.hot_ch.mass_transfer_term[t, p, j].fix(0)
```
adam-a-a, line 435: Ah, I guess this is where you handle `hot_ch.mass_transfer_term`.

watertap/unit_models/MD/membrane_distillation_base.py
```diff
@@ -126,14 +129,37 @@ def build(self):
 
         # Concentration polarization constraint is not accounted for in the below method; it is
         # written later in the base model (eq_concentration_polarization)
-        self.cold_ch.add_concentration_polarization(
-            concentration_polarization_type=ConcentrationPolarizationType.none,
-            mass_transfer_coefficient=MassTransferCoefficient.none,
-        )
+        if hasattr(self.cold_ch.config.property_package, "solute_set"):
+            # If solute_set is defined, add concentration polarization configuration
```
adam-a-a, line 133: Is this necessary because of the vapor state block?

watertap/unit_models/MD/membrane_distillation_1D.py
```diff
@@ -0,0 +1,548 @@
...
+            default=useDefault,
+            description="Discretization scheme to use for DAE transformation",
+            doc="""Discretization scheme to use when transformating domain. See
+        Pyomo documentation for supported schemes.""",
+        ),
+    )
+    CONFIG.declare(
+        "finite_elements",
+        ConfigValue(
+            default=20,
```
adam-a-a, line 399: Any reason for bumping this up to 20?

watertap/unit_models/tests/test_membrane_distillation_1D.py
```diff
@@ -0,0 +1,843 @@
...
+    m.fs.unit.permeability_coef.fix(1e-10)
+    m.fs.unit.membrane_thickness.fix(1e-4)
+    m.fs.unit.membrane_thermal_conductivity.fix(0.2)
+
+    m.fs.unit.cold_ch_inlet.flow_mass_phase_comp[0, "Liq", "H2O"].fix(hot_ch_flow_mass)
+    m.fs.unit.cold_ch_inlet.flow_mass_phase_comp[0, "Liq", "TDS"].fix(0)
+    m.fs.unit.cold_ch_inlet.pressure[0].fix(7e5)
+    m.fs.unit.cold_ch_inlet.temperature[0].fix(273.15 + 25)
+
+    lenght = 8
```
adam-a-a:
```suggestion
    length = 8
```

adam-a-a:
Should the documentation for 0D be updated to match subtle changes here?

watertap/unit_models/MD/membrane_distillation_base.py
ElmiraShamlou: @adam-a-a Some of the changes in this PR, including this one, lay the groundwork for supporting various MD configurations like VMD (upcoming PRs), ensuring channels can adapt to various property packages including pure water. This specific code line addresses scenarios with pure water, where concentration and concentration polarization do not apply.

watertap/unit_models/MD/membrane_distillation_1D.py
ElmiraShamlou: @adam-a-a Based on previous MD model experiences, changes in results are negligible beyond this point.

watertap/unit_models/MD/membrane_distillation_0D.py
ElmiraShamlou: @adam-a-a The hot channel does not include a mass_transfer_term for the vapor phase because it uses solution property packages that exclude vapor phase properties. Instead, the vapor state block is managed independently using a separate vapor property package, and it does not affect the main state blocks of the channel control volume

watertap/unit_models/MD/membrane_distillation_0D.py
ElmiraShamlou: @adam-a-a This condition ensures that solutes do not cross the membrane (assumes there is no membrane wetting).

ElmiraShamlou:
> Should the documentation for 0D be updated to match subtle changes here?

@adam-a-a I believe the core process principles remain unchanged in this PR, so there's no need to update the documentation for 0D. The changes primarily address how the model interacts with different property packages, setting the stage for future MD configurations. However, I need to include documentation for MD 1D, which will be very similar to MD 0D.

watertap/unit_models/MD/membrane_distillation_1D.py
adam-a-a: That makes sense--although I still question whether the default needs to be that value as opposed to a lower value that would solve quicker. I would say if the time to solve isn't too lengthy (for exploratory purposes) at 20 elements, then this is fine. I could see the case where we might want to reduce this value for some tests that will constantly rerun with every PR. 

Whatever the case may be, we can save this for a future PR if we think it would be best to make any adjustments (either in tests or directly to default).

adam-a-a (approved):
LGTM - great job on this! I think this is one of those PRs where we'll just have to play with the model a bit to shake out any issues. As it looks now, I don't see any issues.

adam-a-a:
@ElmiraShamlou could you also create an issue to track any potential items for the future? For example, 1D documentation, future configs, etc. ?

MarcusHolly (approved):
LGTM from a high-level

Issue #1302, opened 2024-02-16
Applying Unit Test Harness


MarcusHolly:
### Description

To be completed in the December release:
- [ ] Ultraviolet Advanced Oxidation Process @luohezhiming 
- [ ] Dewatering Unit @adam-a-a 
- [ ] Electrodialysis 0D & 1D @lbibl 
- [ ] ElectroN-P @luohezhiming 
- [ ] Nanofiltration (0D, DSPMDE, & ZO) @adam-a-a 
- [ ] Pressure Exchanger @luohezhiming 
- [ ] Stoichiometric Reactor @avdudchenko 

Completed in the June release:
- [x] Boron Removal @savannahsakhai #1415
- [x] Coag & Floc @MarcusHolly #1363
- [x] CSTR @agarciadiego #1371 
- [x] CSTR Injection @agarciadiego #1371 
- [x] Ion Exchange @kurbansitterley 
- [x] 0D Membrane Distillation @ElmiraShamlou 
- [x] Osmotically Assisted Reverse Osmosis (0D & 1D) @luohezhiming #1367 
- [x] Pressure Changer @MarcusHolly #1363
- [x] Thickener @agarciadiego  #1371 

Completed in the March Release:
- [x] Reverse Osmosis (0D & 1D) @MarcusHolly #1301 
- [x] Clairifer @MarcusHolly #1301 
- [x] Crystallizer @MarcusHolly #1301 
- [x] Granular Activated Carbon @hunterbarber #1311 
- [x] Electrolyzer @hunterbarber #1321
- [x] 1D Membrane Distillation @ElmiraShamlou #1323  

### Motivation

This will make unit model testing more standardized across the repository.

### Possible Implementation

See [Anaerobic Digester Test.](https://github.com/watertap-org/watertap/blob/main/watertap/unit_models/tests/test_anaerobic_digester.py)

hunterbarber:
Can I make some changes to `UnitTestHarness` to add a `self.costing_model_block` to also assert values on `m.fs.costing` via `self.costing_solutions[m.fs.costing.variable] = value`? @TimBartholomew @MarcusHolly Thoughts?

MarcusHolly:
[...]

PR #1324, opened 2024-03-07
Costing Model Test Harness

hunterbarber:
## Fixes/Resolves:
Similarly implement a costing model test harness, adapted from the unit model test harness

## Summary/Motivation:
Systematically test unit model costing

- Add costing test harness file
- Test with minimally one unit costing model
- Potentially add a test harness for unit model error and warning logs

### Legal Acknowledgement

### DIFF
--- a/watertap/costing/unit_models/tests/unit_costing_test_harness.py
+++ b/watertap/costing/unit_models/tests/unit_costing_test_harness.py
@@ -1,0 +1,188 @@
+#################################################################################
+# WaterTAP Copyright (c) 2020-2023, The Regents of the University of California,
+# through Lawrence Berkeley National Laboratory, Oak Ridge National Laboratory,
+# National Renewable Energy Laboratory, and National Energy Technology
+# Laboratory (subject to receipt of any required approvals from the U.S. Dept.
+# of Energy). All rights reserved.
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and license
+# information, respectively. These files are also available online at the URL
+# "https://github.com/watertap-org/watertap/"
+#################################################################################
+
+import pytest
+import abc
+
+from pyomo.environ import Block, assert_optimal_termination, ComponentMap, value
+from pyomo.util.check_units import assert_units_consistent
+from idaes.core.util.model_statistics import (
+    degrees_of_freedom,
+)
+from idaes.core.solvers import get_solver
+from idaes.core.util.testing import initialization_tester
+import idaes.core.util.scaling as iscale
+
+
+# -----------------------------------------------------------------------------
+
+
+# -----------------------------------------------------------------------------
+class UnitAttributeError(AttributeError):
+    """
+    WaterTAP exception for generic attribute errors arising from unit model testing.
+    """
+
+
+class UnitValueError(ValueError):
+    """
+    WaterTAP exception for generic value errors arising from unit model testing.
+    """
+
+
+class UnitRuntimeError(RuntimeError):
+    """
+    WaterTAP exception for generic runtime errors arising from unit model testing.
+    """
+
+
+class UnitCostingTestHarness(abc.ABC):
+    def configure_class(self):
+        # string for solver, if None use WaterTAP default
+        self.solver = None
+        # dictionary for solver options, if None use WaterTAP default
+        self.optarg = None
+
+        # solution map from var to value
+        self.unit_solutions = ComponentMap()
+
+        # arguments for badly scaled variables
+        self.default_large = 1e4
+        self.default_small = 1e-3
+        self.default_zero = 1e-10
+
+        # arguments for solver tolerance
+        self.default_absolute_tolerance = 1e-12
+        self.default_relative_tolerance = 1e-6
+
+        model = self.configure()
+        if not hasattr(self, "unit_model_block"):
+            self.unit_model_block = model.find_component("fs.unit")
+            if self.unit_model_block is None:
+                raise RuntimeError(
+                    f"The {self.__class__.__name__}.configure method should either "
+                    "set the attribute `unit_model_block` or name it `fs.unit`."
+                )
+        # keep the model so it does not get garbage collected
+        self._model = model
+        blk = self.unit_model_block
+
+        # attaching objects to model to carry through in pytest frame
+        # TODO: Consider removing these objects and directly calling self
+        assert not hasattr(blk, "_test_objs")
+        blk._test_objs = Block()
+        blk._test_objs.solver = self.solver
+        blk._test_objs.optarg = self.optarg
+        blk._test_objs.unit_solutions = self.unit_solutions
+
+    @abc.abstractmethod
+    def configure(self):
+        """
+        Placeholder method to allow user to setup test harness.
+
+        The configure method must set the attributes:
+        unit_solutions: ComponentMap of values for the specified variables
+
+        The unit model tested should be named `fs.unit`, or this method
+        should set the attribute `unit_model_block`.
+
+        Returns:
+            model: the top-level Pyomo model
+        """
+
+    @pytest.fixture(scope="class")
+    def frame(self):
+        self.configure_class()
+        return self._model, self.unit_model_block
+
+    @pytest.mark.unit
+    def test_units_consistent(self, frame):
+        m, unit_model = frame
+        assert_units_consistent(unit_model)
+
+    @pytest.mark.unit
+    def test_dof(self, frame):
+        m, unit_model = frame
+        if degrees_of_freedom(unit_model) != 0:
+            raise UnitAttributeError(
+                "The unit has {dof} degrees of freedom when 0 is required."
+                "".format(dof=degrees_of_freedom(unit_model))
+            )
+
+    @pytest.mark.component
+    def test_initialization(self, frame):
+        m, blk = frame
+        initialization_tester(
+            m,
+            unit=blk,
+            solver=blk._test_objs.solver,
+            optarg=blk._test_objs.optarg,
+        )
+
+    @pytest.mark.component
+    def test_unit_solutions(self, frame):
+        self.configure_class()
+        m, blk = frame
+        solutions = blk._test_objs.unit_solutions
+
+        # solve unit
+        if blk._test_objs.solver is None:
+            opt = get_solver()
+        else:
+            opt = get_solver(
+                solver=blk._test_objs.solver, options=blk._test_objs.optarg
+            )
+        results = opt.solve(blk)
+
+        # check solve
+        badly_scaled_vars = list(
+            iscale.badly_scaled_var_generator(
+                blk,
+                large=self.default_large,
+                small=self.default_small,
+                zero=self.default_zero,
+            )
+        )
+        if badly_scaled_vars:
+            lines = [
+                f"{x[0].name}\t{x[0].value}\tsf: {iscale.get_scaling_factor(x[0])}"
+                for x in badly_scaled_vars
+            ]
+            msg = "One or more badly scaled variables found:\n"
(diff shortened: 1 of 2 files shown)


hunterbarber:
I think this should have some discussion through reviews, a couple comments
- Costing blocks don't have their scaling factors tested
- Not sure what other common tests should be applied in the costing harness
- This decouples unit and cost model tests, unit and cost model documentation are still shared

watertap/costing/unit_models/tests/unit_costing_test_harness.py
```diff
@@ -0,0 +1,171 @@
...
+    """
+
+
+class UnitRuntimeError(RuntimeError):
+    """
+    WaterTAP exception for generic runtime errors arising from unit costing testing.
+    """
+
+
+class UnitCostingTestHarness(abc.ABC):
```
bknueven, line 54: This shares a lot of code with the `UnitTestHarness`. Can it inherit from it instead?

@lbianchi-lbl do you foresee any issues with using inheritance here given how `pytest` classes work?

bknueven:
> I think this should have some discussion through reviews, a couple comments
> 
>     * Costing blocks don't have their scaling factors tested
> 
>     * Not sure what other common tests should be applied in the costing harness
> 
>     * This decouples unit and cost model tests, unit and cost model documentation are still shared

I think not testing the scaling factors is okay.

The documentation should still be shared. I am generally not sure about separating the costing and unit model tests (or functions themselves). The current location of the unit model costing methods is mostly historical -- they were inside the costing package originally and got moved into separate modules (#835). I think we should consider if the costing modules for the unit models *in* WaterTAP should be closer to the unit models themselves.

hunterbarber:
> > I think this should have some discussion through reviews, a couple comments
> > ```
> > * Costing blocks don't have their scaling factors tested
> > 
> > * Not sure what other common tests should be applied in the costing harness
> > 
> > * This decouples unit and cost model tests, unit and cost model documentation are still shared
> > ```
> 
> I think not testing the scaling factors is okay.
> 
> The documentation should still be shared. I am generally not sure about separating the costing and unit model tests (or functions themselves). The current location of the unit model costing methods is mostly historical -- they were inside the costing package originally and got moved into separate modules (#835). I think we should consider if the costing modules for the unit models _in_ WaterTAP should be closer to the unit models themselves.

There's nothing stopping calling the separate `UnitCostingTestHarness` within the same unit model testing to consolidate files. Similarly both the `UnitCostingTestHarness` and `UnitModelTestHarness` can exist in the same file, it is just a matter of going through and deciding what the best structure is.

hunterbarber:
Note, should be behind #1302

hunterbarber:
Accidentally deleted branch when cleaning up, so just restoring

ksbeattie:
@hunterbarber is this still active?  Currently it is not on any release board.

hunterbarber:
> @hunterbarber is this still active? Currently it is not on any release board.

Yes, it was waiting for some 1.0 changes to avoid conflicts before becoming active again.

watertap/costing/unit_models/tests/unit_costing_test_harness.py
bknueven: My comment above about sharing huge amounts of code with `UnitTestHarness` still stands. Every method is largely similar -- I would suggest factoring out the commonalities into a base class.

hunterbarber: Yea I agree, just waiting to get back around to this in order to implement it.

MarcusHolly:
@hunterbarber @bknueven Is there really a need to create a CostingModelTestHarness in the first place? It seems like the current version of the UnitModelTestHarness is already capable of testing costing. See the following example: https://github.com/watertap-org/watertap/blob/b27f92994182535a6832edc02b313cfc25a788fc/watertap/unit_models/tests/test_crystallizer.py#L196.
In other words, what additional functionality would we want the costing test harness to have that the UnitModelTestHarness doesn't have already?

hunterbarber:
> @hunterbarber @bknueven Is there really a need to create a CostingModelTestHarness in the first place? It seems like the current version of the UnitModelTestHarness is already capable of testing costing. See the following example:
> 
> https://github.com/watertap-org/watertap/blob/b27f92994182535a6832edc02b313cfc25a788fc/watertap/unit_models/tests/test_crystallizer.py#L196
> 
> .
> In other words, what additional functionality would we want the costing test harness to have that the UnitModelTestHarness doesn't have already?

I actually hadn't seen this, but good to know that at least the values can be checked with UnitModelTestHarness. So let's assume the solved variable values and the supporting model consistency information like units, dof, and etc. can be checked with the UnitTestHarness if m.fs.unit.costing is built on m.fs.unit. 

Is it worth separating costing tests to have clarity whether a failing test is a consequence of the unit or costing model? They in theory exist as separate products although are frequently used together. Also it becomes more relevant for costing models that can be used for multiple unit models like RO. Is it better to reduce test workflow by using an isolated 1 test attaching to the costing code or have to implement it multiple times within the unit model tests of each applicable unit model.

I guess in this case the unit model test harness could still be the only one used.

Just my initial thoughts.

MarcusHolly:
> > @hunterbarber @bknueven Is there really a need to create a CostingModelTestHarness in the first place? It seems like the current version of the UnitModelTestHarness is already capable of testing costing. See the following example:
> > https://github.com/watertap-org/watertap/blob/b27f92994182535a6832edc02b313cfc25a788fc/watertap/unit_models/tests/test_crystallizer.py#L196
> > 
> > .
> > In other words, what additional functionality would we want the costing test harness to have that the UnitModelTestHarness doesn't have already?
> 
> I actually hadn't seen this, but good to know that at least the values can be checked with UnitModelTestHarness. So let's assume the solved variable values and the supporting model consistency information like units, dof, and etc. can be checked with the UnitTestHarness if m.fs.unit.costing is built on m.fs.unit.
> 
> Is it worth separating costing tests to have clarity whether a failing test is a consequence of the unit or costing model? They in theory exist as separate products although are frequently used together. Also it becomes more relevant for costing models that can be used for multiple unit models like RO. Is it better to reduce test workflow by using an isolated 1 test attaching to the costing code or have to implement it multiple times within the unit model tests of each applicable unit model.
> 
> I guess in this case the unit model test harness could still be the only one used.
> 
> Just my initial thoughts.

When you say "Also it becomes more relevant for costing models that can be used for multiple unit models like RO", are you referring to unit models that have multiple costing methods? If so, you can see how its handled in the crystallizer file, which leads to your question. Is it better to have separate test(s) for the costing methods that aren't lumped together with the unit model tests or is it better to just lump everything together (what is currently done). Personally, I don't mind having everything lumped together, and I think it makes writing the tests a bit simpler

hunterbarber:
> > > @hunterbarber @bknueven Is there really a need to create a CostingModelTestHarness in the first place? It seems like the current version of the UnitModelTestHarness is already capable of testing costing. See the following example:
> > > https://github.com/watertap-org/watertap/blob/b27f92994182535a6832edc02b313cfc25a788fc/watertap/unit_models/tests/test_crystallizer.py#L196
> > > 
> > > .
> > > In other words, what additional functionality would we want the costing test harness to have that the UnitModelTestHarness doesn't have already?
> > 
> > 
> > I actually hadn't seen this, but good to know that at least the values can be checked with UnitModelTestHarness. So let's assume the solved variable values and the supporting model consistency information like units, dof, and etc. can be checked with the UnitTestHarness if m.fs.unit.costing is built on m.fs.unit.
> > Is it worth separating costing tests to have clarity whether a failing test is a consequence of the unit or costing model? They in theory exist as separate products although are frequently used together. Also it becomes more relevant for costing models that can be used for multiple unit models like RO. Is it better to reduce test workflow by using an isolated 1 test attaching to the costing code or have to implement it multiple times within the unit model tests of each applicable unit model.
> > I guess in this case the unit model test harness could still be the only one used.
> > Just my initial thoughts.
> 
> When you say "Also it becomes more relevant for costing models that can be used for multiple unit models like RO", are you referring to unit models that have multiple costing methods? If so, you can see how its handled in the crystallizer file, which leads to your question. Is it better to have separate test(s) for the costing methods that aren't lumped together with the unit model tests or is it better to just lump everything together (what is currently done). Personally, I don't mind having everything lumped together, and I think it makes writing the tests a bit simpler

I was actually referring to the opposite, 1 costing model that is used across 2+ unit models. _If the costing models are tested only at the unit model level_, then the costing model testing would be redundant to some extent (e.g. costing/reverse_osmosis being tested multiple times across units/reverse_osmosis_0D, units/reverse_osmosis_1D, units/oaro_0D, and units/oaro_1D).

Disclaimer that all of these are just my thoughts and I very well could be wrong: I think having independent testing in units/tests and costing/tests can more quickly isolate problems (on the unit model or costing model) when tests fail, and reduce the length of the testing workflow in general. With that being said, I think both could use this `UnitModelTestHarness` based on what you brought up with the crystallizer, but maybe could be refactored into util/ModelTestHarness.py and generalized from needing a m.fs.unit.

Also I was trying to unify costing model overall just because I don't think there's any real consistency at the moment.

MarcusHolly:
@bknueven Any thoughts on the conversation above? I agree with Hunter that we should have independent tests in unit_models/tests and costing/unit_models/tests, but is it fine to use the UnitTestHarness for both purposes? If so, should the UnitTestHarness just stay as it is (see crystallization implementation above) or is there additional functionality we want to add for testing unit costing?

bknueven:
> @bknueven Any thoughts on the conversation above? I agree with Hunter that we should have independent tests in unit_models/tests and costing/unit_models/tests, but is it fine to use the UnitTestHarness for both purposes? If so, should the UnitTestHarness just stay as it is (see crystallization implementation above) or is there additional functionality we want to add for testing unit costing?

I think testing the costing *with* each unit model is okay, even if redundant. Testing the costing methods without the unit models certainly seems possible, but also a lot of effort. And in any case, you would want to test the costing methods with each unit model anyway to make sure there's compatibility in the different versions.

I am also not sure if it needs a separate harness. Maybe it should just be optional functionality in the unit test hardness, which already has all the information you need.

MarcusHolly:
> > @bknueven Any thoughts on the conversation above? I agree with Hunter that we should have independent tests in unit_models/tests and costing/unit_models/tests, but is it fine to use the UnitTestHarness for both purposes? If so, should the UnitTestHarness just stay as it is (see crystallization implementation above) or is there additional functionality we want to add for testing unit costing?
> 
> I think testing the costing _with_ each unit model is okay, even if redundant. Testing the costing methods without the unit models certainly seems possible, but also a lot of effort. And in any case, you would want to test the costing methods with each unit model anyway to make sure there's compatibility in the different versions.
> 
> I am also not sure if it needs a separate harness. Maybe it should just be optional functionality in the unit test hardness, which already has all the information you need.

@bknueven @hunterbarber Alright, so it seems to me like costing/unit_model/tests would test costing methods w/o unit models and unit_models/tests would test costing methods with the corresponding unit model. Both cases could just use the current iteration of this UnitTestHarness, so I think we can close this PR.

The only issue now is actually adding the costing/unit_model/tests testing and making sure the unit model tests actually test costing. As Ben mentioned, testing the costing methods w/o the unit models would require a decent bit of effort/time, and it doesn't strike me as a high-priority task, so I'm not sure when this would get done. Ensuring the unit_models/tests costing actually test the costing method(s) also doesn't seem to be high-priority, although it would take much less work than the former.

MarcusHolly:
Closing this PR since the UnitModelTestHarness can already handle adding unit model costing tests. For an example, see https://github.com/watertap-org/watertap/blob/b27f92994182535a6832edc02b313cfc25a788fc/watertap/unit_models/tests/test_crystallizer.py#L196

PR #1244, opened 2023-12-08
Tinkering with BSM2 Optimization

adam-a-a:
## Summary/Motivation:
Started as a draft PR for the team to look into optimization of the BSM2 flowsheet. Problems could stem from poor scaling, poor initialization, problematic cost relationships (ran into ampl errors a couple times already for aeration basin capital cost constrint), or all of the above (or some other unknown reason).  

## Changes proposed in this PR:
- add setup_optimization function
- revise costing
- refactor cstr_injection and create AerationTank unit model class which inherits from cstr_injection 
- alter the cost method for cstr (without injection) to one that appears more practical

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #1333, opened 2024-03-21
add missed test for aeration tank

adam-a-a:
## Fixes/Resolves:
Adds test file that should have been included in #1244 
## Summary/Motivation:
This test was written for PR #1244 but (unintentionally) was never staged. 

- add test file for aeration tank, which mostly inherits from cstr_injection and specifies some config option limitations

### Legal Acknowledgement


agarciadiego (approved):
LGTM

MarcusHolly (approved):
I imagine this will eventually be replaced with the unit model test harness, but in the interest of time, this transition can happen in the June releasee.

adam-a-a:
> I imagine this will eventually be replaced with the unit model test harness, but in the interest of time, this transition can happen in the June releasee.

Right--I had written this test before the unit test harness was in place. Furthermore, I am not sure that the unit test harness would cover everything that these tests cover, but I suppose it could replace a portion of the more fundamental tests.

PR #1112, opened 2023-08-24
Extended ElectroNP flowsheet with ASM and anaerobic digestor

luohezhiming:
## Fixes/Resolves:
This PR is the same flowsheet and can replace PR #1100 

## Changes proposed in this PR:
- electroNP flowsheet with ASP
- add test files

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #1341, opened 2024-03-26
BSM2 flowsheet with electroNP

luohezhiming:
## Fixes/Resolves:

Replaces PR #1112 

## Summary/Motivation:
BSM2 flowsheet with electroNP, no biological removal for ASM2d-ADM1 translator,
ElectroNP ZO adjusted to account for water recovered in byproduct solids

- BSM2 flowsheet with electroNP
- modified electroNP ZO (scaling and water recovery)

### Legal Acknowledgement


watertap/examples/flowsheets/case_studies/electroNP/BSM2_electroNP_no_bioP.py
```diff
@@ -0,0 +1,666 @@
...
+    )
+
+    return m, results
+
+
+def build_flowsheet(has_electroNP=False):
+    m = pyo.ConcreteModel()
+
+    m.fs = FlowsheetBlock(dynamic=False)
+    m.fs.has_electroNP = has_electroNP
```
MarcusHolly, line 119: Is it necessary to have this? I assume if someone is running this flowsheet, they'd want ElectroN-P to be included. If they didn't want ElectroN-P, they could just run the Extended BSM2 flowsheet. Was this just leftover from when you were diagnosing issues with the flowsheet?

MarcusHolly (approved):
LGTM - just two things to consider:

1. Do we want documentation for this flowsheet? It'd be very similar to the existing documentation for the Extended BSM2 flowsheet, so maybe we don't need it or could just mention this flowsheet in the existing documentation rather than create new documentation.
2. Do we want to AND is it feasible to just add the `has_electroNP` config option into the existing BSM2 flowsheet? It may be too much of a hassle because I haven't cross-checked how many of the default values changed in this flowsheet compared to the existing Extended BSM2 flowsheet. But in terms of future developments, I'm not sure if we want to make a distinct flowsheet each time we want to add a unit model to BSM2 when it might just be easier to add a config option (see https://github.com/watertap-org/watertap/blob/main/watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination_withRO.py for how config options could be handled in the future)

watertap/examples/flowsheets/case_studies/electroNP/BSM2_electroNP_no_bioP.py
```diff
@@ -0,0 +1,666 @@
+#################################################################################
+# WaterTAP Copyright (c) 2020-2023, The Regents of the University of California,
```
adam-a-a, line 2: I would remove "no_bioP" from file names, so this file and the test file.

agarciadiego (approved):
LGTM

lbianchi-lbl:
It looks like 4fb7036ae is the cause of the current initialization failure. Looking at the diff, the most relevant change seems to be #1353. Is it possible that this PR needs to be updated by adding `solver="ipopt-watertap"` similarly to how other flowsheets had to be updated for #1353?

watertap/examples/flowsheets/case_studies/electroNP/BSM2_electroNP_no_bioP.py
```diff
@@ -0,0 +1,666 @@
...
+from idaes.models.unit_models import (
+    CSTR,
+    Feed,
+    Separator,
+    Product,
+    Mixer,
+    PressureChanger,
+)
+from idaes.models.unit_models.separator import SplittingType
+from idaes.core.solvers import get_solver
```
MarcusHolly:
```suggestion
from watertap.core.solvers import get_solver
```

watertap/examples/flowsheets/case_studies/electroNP/BSM2_electroNP_no_bioP.py
```diff
@@ -0,0 +1,666 @@
...
+        "temperature": {0: 308.15},
+        "pressure": {0: 101325},
+    }
+
+    # Pass the tear_guess to the SD tool
+    seq.set_guesses_for(m.fs.R3.inlet, tear_guesses)
+    seq.set_guesses_for(m.fs.translator_asm2d_adm1.inlet, tear_guesses2)
+
+    def function(unit):
+        unit.initialize(outlvl=idaeslog.INFO, optarg={"bound_push": 1e-2})
```
lbianchi-lbl, line 609: @MarcusHolly @bknueven does `solver="ipopt-watertap"` need to be added to the `unit.initialize()` call here after #1353? (asking because of the current initialization failures)

**EDIT** fixed the solver name to `"ipopt-watertap"`

bknueven: Certainly a good idea if a solve is failing.

lbianchi-lbl: I've tested this locally and it seems to work!

```suggestion
        unit.initialize(outlvl=idaeslog.INFO, optarg={"bound_push": 1e-2}, solver="ipopt-watertap")
```

luohezhiming: All tests pass after import `get_solver` from watertap, no need to change the initialization settings. @lbianchi-lbl @MarcusHolly

lbianchi-lbl: Yes, I believe you're right @luohezhiming. I think I have mixed things up and what resolved the failure was c5e1b6f rather than the `initialize(..., solver="ipopt-watertap")` change.

PR #1343, opened 2024-03-28
MCAS compatibility with ZO models

adam-a-a:
## Summary/Motivation:

To use MCAS in combination with ZO and detailed models, I made MCAS compatible with  ZO models.

- Add isothermal and isobaric config options to ZO base as boolean options
- Add temperature and pressure to ZO SIDO and SISO models
- Add isothermal/isobaric constraints to SIDO and SISO by default (when isothermal/isobaric config options set to true)
- Add total_dissolved_solids property to MCAS
- Add flow_mass_comp property to MCAS for compatibility with ZO model naming which omits `phase`
- ...

### Legal Acknowledgement


adam-a-a:
This PR could probably use more testing, which is why I am keeping this in draft mode.

watertap/core/zero_order_sido.py
```diff
@@ -158,6 +158,37 @@ def solute_treated_equation(b, t, j):
...
     self._get_Q = MethodType(_get_Q_sido, self)
 
+    if ("temperature" in self.properties_in[0].define_state_vars()) and (
+        self.config.isothermal
+    ):
+        _add_isothermal_constraints(self)
+    if ("pressure" in self.properties_in[0].define_state_vars()) and (
+        self.config.isothermal
+    ):
+        _add_isobaric_constraints(self)
```
MarcusHolly:
```suggestion
    if ("pressure" in self.properties_in[0].define_state_vars()) and (
        self.config.isobaric
    ):
        _add_isobaric_constraints(self)
```

watertap/core/zero_order_sido_reactive.py
```diff
@@ -328,6 +328,37 @@ def solute_treated_equation(b, t, j):
...
     self._get_Q = MethodType(_get_Q_sidor, self)
 
+    if ("temperature" in self.properties_in[0].define_state_vars()) and (
+        self.config.isothermal
+    ):
+        _add_isothermal_constraints(self)
+    if ("pressure" in self.properties_in[0].define_state_vars()) and (
+        self.config.isothermal
+    ):
+        _add_isobaric_constraints(self)
```
MarcusHolly:
```suggestion
    if ("pressure" in self.properties_in[0].define_state_vars()) and (
        self.config.isobaric
    ):
        _add_isobaric_constraints(self)
```

watertap/core/zero_order_siso.py
```diff
@@ -125,6 +125,33 @@ def solute_treated_equation(b, t, j):
...
     self._get_Q = MethodType(_get_Q_siso, self)
 
+    if ("temperature" in self.properties_in[0].define_state_vars()) and (
+        self.config.isothermal
+    ):
+        _add_isothermal_constraints(self)
+    if ("pressure" in self.properties_in[0].define_state_vars()) and (
+        self.config.isothermal
+    ):
+        _add_isobaric_constraints(self)
```
MarcusHolly:
```suggestion
    if ("pressure" in self.properties_in[0].define_state_vars()) and (
        self.config.isobaric
    ):
        _add_isobaric_constraints(self)
```

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -444,6 +439,20 @@ def build(self):
...
+                # TODO: setting to 1 should effectively ignore mw in the model for now, but conversions between mass and mole won't be valid. Need a long-term solution.
+                mw_comp[i] = 1
+            if not isinstance(value(mw_comp[i]), (int, float)):
+                track_bad_mw_input.update({i: mw_comp[i]})
+            else:
+                pass
+
+        if len(track_bad_mw_input) > 0:
+            raise ConfigurationError(
+                f"'mw_data' values must either by numeric or None when molecular weight is not applicable. The following inputs should be revised:\n {track_bad_mw_input}"
```
MarcusHolly:
```suggestion
                f"'mw_data' values must either be numeric or None when molecular weight is not applicable. The following inputs should be revised:\n {track_bad_mw_input}"
```

MarcusHolly (changes requested):

MarcusHolly (approved):

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -1849,6 +1881,36 @@ def rule_total_hardness(b):
...
+            initialize=1000,
+            domain=NonNegativeReals,
+            bounds=(0, None),
+            units=pyunits.mg / pyunits.L,
+            doc="total dissolved solids",
+        )
+        # add try/except to handle case without ions,
+        # which would return 0 and result in Inconsitentunits error due to conversion of dimensionless to mg/L
+        try:
+            total_dissolved_solids_temp = pyunits.convert(
```
kurbansitterley, line 1895: I am not wild about this calculation for TDS because a proper TDS measurement would include non ionic species and organic species (I would argue this is "total dissolved salts" rather than "solids"). However, I think that this is the best we can do at the moment...

Why not first check for a "TDS" component in the solute list (i.e., allow the user to set the TDS concentration directly) and if it isn't found then default to this calculation?

kurbansitterley (approved):
I approve. But I am torn on this approval for a very pedantic reason. 

On one hand, I am not wild about the constraint for ``total_dissolved_solids`` because it is more of a ``total_dissolved_salts`` equation. Essentially, there _can_ be other things in a TDS measurement that are not encompassed in the ``ion_set`` (e.g., organic molecules, carbs, proteins, etc). This is why you can estimate TDS from a conductivity measurement. Ideally, we would let the user set the TDS concentration via the solute list, but then this would present problems when calculating ``flow_mass_comp`` because you end up double counting components if you still want individual ions in your solute list. So not really an ideal solution.

On the other hand, this calculation for ``total_dissolved_solids`` would probably be fine for _most_ use cases and the ZO models that just want to see TDS in the component stream would probably be minimally effected by this nuanced difference. Also, I don't think there is any easy/realistic way we could use a multi-component stream and calculate TDS from it, so this is essentially as good as we can do at the moment. I am also a very big supporter of making all of the ZO models compatible with MCAS. So I also approve as a step in that direction.

So approved, but I give @adam-a-a something to ponder for future consideration.

adam-a-a:
Thanks for the thoughtful review, @kurbansitterley 

I did think about this and thought of a couple ways to improve this in the future. For example, we can have the current implementation as default but provide the option for the user to specify components to include/exclude in the tds calc. We can chat more about this.

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
```diff
@@ -2255,6 +2321,11 @@ def test_no_mw_data_provided():
         m.fs.properties = MCASParameterBlock(
             solute_list=["foo", "Na_+", "Cl_-"], ignore_neutral_charge=True
         )
+    # Choose capitalized solute names that won't be found in periodic table and omit mw_data
+    # Because of helper function logic, an example like following would result in an empty result
+    msg = "The symbol 'T' from the component name 'TDS' could not be found in the periodic table."
+    with pytest.raises(ConfigurationError, match=re.escape(msg)):
+        m.fs.mcas_props = MCASParameterBlock(solute_list=["TDS"])
```
lbianchi-lbl, line 2328: @adam-a-a It looks like this is currently causing test failures (https://github.com/watertap-org/watertap/actions/runs/9764214692/job/26952037278#step:6:302): 
image

I tried searching the code for that error message (i.e. the code that builds the string used in the error message), but I couldn't find anything. Do you have more information about what might be happening?

PR #1349, opened 2024-03-29
Minor Dye Desalination Flowsheet Updates

MarcusHolly:
## Fixes/Resolves:
Resolves various comments pointed out in #1294; namely:

- [x] Replacing instances of .value with value()
- [x] Adding a product block to reduce the number of if-statements
- [x] Add config option for toggling between 1DRO and 0DRO

### Legal Acknowledgement


MarcusHolly:
I have not combined the `dye_desalination_withRO.py` and `dye_desalination.py` flowsheets in this PR (i.e. merging into one flowsheet that has a configuration option to include the RO treatment train). I've started implementing this change locally, but it's a bigger hassle than I initially thought - mainly because modifying the costing is a pain. So perhaps this smaller PR can be merged first and a subsequent PR will focus on combining the two flowsheets (it should also be easier to review this way).

hunterbarber (approved):
Changes are all very minor so LGTM. I would maybe still bullet the naming conventions, conversion to pyo.value() and other changes in the PR describtion for clarity.

agarciadiego (approved):

luohezhiming (approved):

adam-a-a:
LGTM - I think we can rethink little bits here and there, but we can aim to polish on subsequent PR that combines both flowsheets (with and without RO).

PR #1352, opened 2024-04-03
Combine Dye Desal Flowsheets

### LINKED ISSUE #1337: Combine Dye Desal Flowsheets
### Description

There are currently two flowsheets for dye desalination: `dye_desalination.py` and `dye_desal_withRO.py`. These two flowsheets should be combined such that `include_RO` is a configuration option in `dye_desal_withRO`, similar to how other configuration options are set up in this flowsheet. Then we can delete the redundant `dye_desalination.py` flowsheet (which just excludes RO) and rename `dye_desal_withRO.py` to something like `dye_desalination.py`.

### Motivation

We have two flowsheets that are nearly identical, so it would make sense to address the differences by adding a configuration option rather than having an entirely separate flowsheet. See [#1294 ](https://github.com/watertap-org/watertap/pull/1294#discussion_r1530740175) for additional context.

### Possible Implementation

Add configuration option to the existing `dye_desal_withRO.py` flowsheet.

### Additional Context

_No response_

MarcusHolly:
## Fixes/Resolves:

#1337 

-  Deletes original `dye_desalination.py` flowsheet and renames `dye_desalination_withRO.py` to `dye_desalination.py`
- Adds include_RO config option into flowsheet
- Modifies the relevant files to account for the first two changes

### Legal Acknowledgement


watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
```diff
@@ -9,102 +9,172 @@
...
+        include_gac=False,
+    )
     set_operating_conditions(m)
 
-    assert_degrees_of_freedom(m, 0)
     assert_units_consistent(m)
 
+    dt = DiagnosticsToolbox(m)
+    print("---Structural Issues---")
+    dt.report_structural_issues()
```
adam-a-a: Do you get any structural issues reported back?

MarcusHolly: Nope, it was just leftover from when I was diagnosing issues. I'll delete this.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
```diff
@@ -9,102 +9,172 @@
...
+    initialize_costing(m)
+    assert_degrees_of_freedom(m, 0)  # ensures problem is square
+
+    if hasattr(m.fs, "desalination"):
+        optimize_operation(m)  # unfixes specific variables for cost optimization
+    else:
+        pass
 
-    results = solve(m, checkpoint="solve flowsheet after costing")
+    solve(m, checkpoint="solve flowsheet after costing")
```
adam-a-a: So, I recall that `solve` uses `check_solve`, which would throw an InitializationError if `check_optimal_termination` returns False. However, an InitializationError wouldn't be appropriate here, so maybe we should just use assert_optimal_termination here. 

The original point of using the check_solve function was to help with troubleshooting.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
```diff
@@ -148,71 +215,216 @@ def eq_flow_mass_comp(blk, j):
                     == blk.properties_out[0].flow_mass_phase_comp["Liq", "H2O"]
                 )
 
+    elif include_dewatering and include_gac:
+        raise TypeError("This system cannot have both dewatering and GAC units.")
```
hunterbarber: Is this conceptually it should not have both, or that it's just not implemented in the model?

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
```diff
@@ -148,71 +215,216 @@ def eq_flow_mass_comp(blk, j):
...
+                )
+            else:
+                return (
+                    blk.properties_in[0].flow_mass_comp["H2O"]
+                    == blk.properties_out[0].flow_mass_phase_comp["Liq", "H2O"]
+                )
+
     else:
-        m.fs.dye_retentate = Product(property_package=m.fs.prop)
+        pass
```
hunterbarber, line 301: Is this unnecessary?

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
```diff
@@ -148,71 +215,216 @@ def eq_flow_mass_comp(blk, j):
...
+    if hasattr(m.fs, "pretreatment"):
+        m.fs.s_feed = Arc(source=m.fs.feed.outlet, destination=prtrt.wwtp.inlet)
+        prtrt.s01 = Arc(source=prtrt.wwtp.treated, destination=dye_sep.P1.inlet)
+        prtrt.s02 = Arc(
+            source=prtrt.wwtp.byproduct, destination=m.fs.wwt_retentate.inlet
+        )
+    else:
+        m.fs.s_feed = Arc(source=m.fs.feed.outlet, destination=dye_sep.P1.inlet)
+
+    dye_sep.s01 = Arc(
```
hunterbarber, line 313: Have you ensured s01is never overwritten? Unclear to me based on the restriction of config combinations.

hunterbarber: I see some exist on fs and some exist on treatment section blocks though

hunterbarber (approved):
Didn't see any major flags, but only reviewed in brief because this is only a consolidation and this may move to a separate repo anyway.

If subsequent changes don't move this, I can review in more detail.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
MarcusHolly: I don't think it would make sense conceptually to have both since the purpose behind including these unit models was to remove water from/further concentrate the dye waste stream. If we had these two units in series, the second unit would be redundant and not do very much.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
MarcusHolly: The else statement isn't necessary, but I just figured it was good practice. I can get rid of this and similar "pass" else statements if we'd like.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
MarcusHolly: `m.fs.s01` is only created now if there is a GAC or dewatering unit. I believe arcs like `prtrt.s01` are fundamentally different than `m.fs.s01`. I'm not sure if I've addressed your question...

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
adam-a-a:
```suggestion
        raise TypeError("This system is not designed to have both dewatering and GAC units.")
```

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
adam-a-a: I think this can stay in. The reason being that it helps with readability.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/tests/test_dye_desalination.py
```diff
@@ -49,53 +50,82 @@ def test_set_operating_conditions(self, system_frame):
...
+        optimize_operation(m)
+
         results = solve(m)
         assert_optimal_termination(results)
 
-        # check values
-        assert pytest.approx(19.626110, rel=1e-3) == value(m.fs.LCOT)
+        # check costing
+        assert pytest.approx(26.933615, rel=1e-3) == value(m.fs.LCOW)
+        assert pytest.approx(19.781864, rel=1e-3) == value(m.fs.LCOT)
```
adam-a-a, line 128: Any known reasons why the LCOT changed?

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
```diff
@@ -148,71 +215,218 @@ def eq_flow_mass_comp(blk, j):
...
+            desal.RO = ReverseOsmosis0D(
+                property_package=m.fs.prop_ro,
+                has_pressure_change=True,
+                pressure_change_type=PressureChangeType.calculated,
+                mass_transfer_coefficient=MassTransferCoefficient.calculated,
+                concentration_polarization_type=ConcentrationPolarizationType.calculated,
+            )
+
+        desal.RO.width.setub(2000)
+        desal.RO.area.setub(20000)
```
adam-a-a, line 266: Do we need this?

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
```diff
@@ -228,71 +442,341 @@ def set_operating_conditions(m):
...
+
+    # RO inlet velocity
+    desal.RO.feed_side.velocity[0, 0].unfix()
+    desal.RO.feed_side.velocity[0, 0].setub(0.3)
+    desal.RO.feed_side.velocity[0, 0].setlb(0.1)
+
+    # RO membrane area
+    desal.RO.area.unfix()
+    desal.RO.area.setub(5000)
+    desal.RO.area.setlb(50)
```
adam-a-a, line 593: I wonder if we want to check these values in tests since they could play a role in convergence later on if we make changes.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
```diff
@@ -351,89 +935,540 @@ def LCOT(b):
...
+                b.feed.properties[0].flow_vol,
+                to_units=pyunits.m**3 / pyunits.year,
+            )
+            * b.zo_costing.utilization_factor
+        )
+
+    @m.fs.Expression(
+        doc="Levelized cost of water with respect to volumetric permeate flow"
+    )
+    def LCOW(b):
```
adam-a-a, line 956: This is likely something we can revisit since this was put in place before our costing framework was refined. I would think this could be trimmed down using functionality from our costing framework.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/tests/test_dye_desalination.py
MarcusHolly: This is a product of the naming being changed. So this test file was originally called `test_dye_desalination_withRO.py`, but since we combined the two flowsheets, I decided to drop the `withRO`, which just so happens to be the name of the test file that I deleted (the test file for the `dye_desalination.py` flowsheet, which did not include RO).

TLDR: these values didn't actually change. See https://github.com/watertap-org/watertap/blob/main/watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/tests/test_dye_desalination_withRO.py#L127

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
MarcusHolly: Yea, I can address this in a subsequent PR.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/tests/test_dye_desalination.py
```diff
@@ -21,26 +21,27 @@
...
     display_results,
+    display_costing,
 )
 
-
 solver = get_solver()
 
 
-class TestDyeFlowsheet:
+class TestDyewithROFlowsheetwithPretreatment:
```
adam-a-a, line 35: The only major item that I think we should make sure of is whether RO (0D and 1D) is being tested. We should check a couple of basic results to be safe.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
MarcusHolly: This appears to be necessary (i.e. deleting it results in the solver failing to return an optimal solution)

adam-a-a (changes requested):
I might have missed it, but I didn't notice testing to check RO results, when included in the flowsheet. I think we should do some minimal checks when RO 0D and 1D are included in the flowsheet. For example, can verify solution for membrane area, operating pressure, and flux (`flux_mass_phase_comp`).

MarcusHolly:
The `class TestDyewithROFlowsheetDefault:` includes 1DRO in the flowsheet, but I don't currently have a test for 0D RO, so I will add that in soon.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/dye_desalination/dye_desalination.py
MarcusHolly: I will be sure to include checks for these in cases where RO is in the flowsheet

adam-a-a (approved):
LGTM

PR #1358, opened 2024-04-12
add steam heater and condenser

ElmiraShamlou:
## Fixes/Resolves:

Adds a steam heater and condenser unit model, assuming full condensation at the outlet.
Estimate cooling water flow for the condenser option assuming specific design delta T.
Compatible with the WaterTap Property Packages.
Evaporative based desalinations like MD flowsheet, and crystallizer flowsheet will benefit from this added steam heater and condenser unit models. 

-
-

### Legal Acknowledgement


watertap/unit_models/steam_heater.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+@declare_process_block_class(
+    "FWHCondensing0D",
+    doc="""Feedwater Heater Condensing Section
+The feedwater heater condensing section model is a normal 0D heat exchanger
+model with an added constraint to calculate the steam flow such that the outlet
+of shell is a saturated liquid.""",
+)
+class FWHCondensing0DData(HeatExchangerData):
```
adam-a-a: I think it would be preferable for the class name to match the filename--e.g., `SteamHeater0D `with filename updated to `steam_heater_0D`.

andrewlee94:
@ElmiraShamlou @adam-a-a I am guessing you aren't aware of this, but IDAES already has a model for a heat exchanger with a condensing section. In order to avoid duplication of code and effort, it would probably be  good idea to check this out to see if it might meet your needs: https://github.com/IDAES/idaes-pse/blob/main/idaes/models_extra/power_generation/unit_models/feedwater_heater_0D.py

ElmiraShamlou:
> @ElmiraShamlou @adam-a-a I am guessing you aren't aware of this, but IDAES already has a model for a heat exchanger with a condensing section. In order to avoid duplication of code and effort, it would probably be good idea to check this out to see if it might meet your needs: https://github.com/IDAES/idaes-pse/blob/main/idaes/models_extra/power_generation/unit_models/feedwater_heater_0D.py

@andrewlee94 Thanks for the comment. Yes, I'm aware of the IDAES feed water heater. The current PR aims to enhance the steam heater's compatibility with the WaterTAP property packages. This is because the existing IDAES steam heater requires properties that are not available in WaterTAP property packages. We have previously discussed the possibility of integrating these properties to simplify future development efforts with the team.

watertap/unit_models/tests/test_steam_heater.py
```diff
@@ -0,0 +1,75 @@
...
+    m.fs.unit.cold_side_outlet.flow_mass_phase_comp[0, "Vap", "H2O"].fix(0)
+    m.fs.unit.cold_side_inlet.flow_mass_phase_comp[0, "Liq", "H2O"].fix(6)
+    m.fs.unit.area.fix(50)
+    m.fs.unit.overall_heat_transfer_coefficient.fix(2e3)
+
+    assert degrees_of_freedom(m) == 0
+
+    m.fs.unit.initialize()
+
+    # assert degrees_of_freedom(m) == 0
```
adam-a-a: delete commented code

watertap/unit_models/steam_heater.py
```diff
@@ -0,0 +1,116 @@
...
+                b.hot_side.properties_in[t].flow_mass_phase_comp["Vap", j]
+                + b.hot_side.properties_in[t].flow_mass_phase_comp["Liq", j]
+                == b.hot_side.properties_out[t].flow_mass_phase_comp["Liq", j]
+            )
+
+        units_meta = (
+            self.config.hot_side.property_package.get_metadata().get_derived_units
+        )
+
+        self.pressure_diff = Var(
```
adam-a-a: Is this terminal pressure difference across hot and cold side? Or pressure drop across hot side? If the latter, we can use `deltaP` from ControlVolume and remove this. If the former, should add a doc string describing, and I'd prefer naming this `terminal_pressure_diff`.

watertap/unit_models/steam_heater.py
```diff
@@ -0,0 +1,116 @@
...
+            self.flowsheet().time, initialize=0, units=units_meta("pressure")
+        )
+
+        self.pressure_diff.fix()
```
adam-a-a: I would also opt not to fix this to 0 like this. I think we can take a couple alternative approaches:
1) Don't unfix here, and require the user to fix this to 0 or some other value. On that note, do we want to enforce a domain for the variable to be NonNegativeReals? This would be in case the user tries entering a negative value (unsure whether this would be caught by something else).
2)  Add a config option such as `has_terminal_pressure_difference` with default set to False, which would in turn fix `pressure_diff` to 0 as you do here. Then, if user sets to True, they would then be responsible for fixing a value to `pressure_diff` to set up the square problem.

watertap/unit_models/steam_heater.py
```diff
@@ -0,0 +1,116 @@
...
+        def outlet_pressure_sat(b, t):
+            return (
+                b.hot_side.properties_out[t].pressure + b.pressure_diff[t]
+                == b.hot_side.properties_out[t].pressure_sat
+            )
+
+    def initialize_build(self, *args, **kwargs):
+        """
+        Use the regular heat exchanger initialization, with the extraction rate
+        constraint deactivated; then it activates the constraint and calculates
```
adam-a-a: replace "extraction rate constraint" with appropriate constraints in description

watertap/unit_models/tests/test_steam_heater.py
```diff
@@ -0,0 +1,75 @@
...
+        flow_pattern=HeatExchangerFlowPattern.countercurrent,
+    )
+
+    m.fs.unit.hot_side_inlet.flow_mass_phase_comp[0, "Vap", "H2O"].set_value(6)
+    m.fs.unit.hot_side_inlet.flow_mass_phase_comp[0, "Liq", "H2O"].fix(0)
+    m.fs.unit.hot_side_inlet.temperature.fix(273.15 + 140)
+    m.fs.unit.cold_side_inlet.pressure.fix(101325)
+    m.fs.unit.cold_side_inlet.temperature.fix(273.15 + 70)
+    m.fs.unit.cold_side_inlet.flow_mass_phase_comp[0, "Vap", "H2O"].fix(0)
+    m.fs.unit.cold_side_outlet.flow_mass_phase_comp[0, "Vap", "H2O"].fix(0)
```
adam-a-a: Should _these_ vapor mass flowrates on the cold side just be fixed to 0 within the unit model?

watertap/unit_models/steam_heater.py
```diff
@@ -0,0 +1,116 @@
...
+from idaes.core.solvers import get_solver
+from idaes.core.util.model_statistics import degrees_of_freedom
+import idaes.logger as idaeslog
+
+
+_log = idaeslog.getLogger(__name__)
+
+
+__author__ = "Elmira Shamlou"
+
```
adam-a-a: I might add a doc string up here describing the model, its relation to the IDAES version that it is based off of, and describe why this separate implementation is being used instead.

adam-a-a: Oh, and you can also add a TODO note along the lines of "Consider incorporating as a modification to IDAES' FeedWaterHeater model directly in the IDAES repo."

watertap/unit_models/steam_heater.py
ElmiraShamlou: @adam-a-a This difference is not related to any of the previously mentioned options; it actually represents the pressure difference between the outlet on the hot side and the saturation pressure. We want to revise the IDAES heat exchanger model to ensure that the outlet pressure on the hot side matches the saturation pressure, thus guaranteeing full condensation of the steam. Based on this, we will also calculate the steam extraction rate necessary to achieve this

watertap/unit_models/steam_heater.py
ElmiraShamlou: @adam-a-a I'm concerned that this configuration option might unnecessarily complicate the model. The primary purpose of this steam heater is to adjust the IDAES heat exchanger's outlet so that full steam condensation occurs, which requires the outlet pressure to equal the saturation pressure.

watertap/unit_models/steam_heater.py
ElmiraShamlou: @adam-a-a The name was adopted from the IDAES feed water heater. I'll update it to your suggestion.

watertap/unit_models/tests/test_steam_heater.py
ElmiraShamlou: The selection depends on the application and whether the property package for the cold side includes a vapor phase. So I believe it is better to not fix it within the unit model.

watertap/unit_models/steam_heater.py
ElmiraShamlou: addressed

watertap/unit_models/steam_heater.py
ElmiraShamlou: addressed

watertap/unit_models/steam_heater.py
adam-a-a: Right, that makes sense. I suggest adding a doc string to clarify what the variable represents.

watertap/unit_models/steam_heater.py
adam-a-a: I get that, but I do not think we need to hard-code the outlet pressure to exactly equal saturation pressure. So, if adding a config option would be overcomplicating this, then option 1 sounds like the way to go. In reality, I wouldn't expect to target the outlet pressure to be exactly the same as the saturation pressure; there would be some overshoot, and I figure that this variable should be more transparent to the user so that he/she could specify the overpressure that they want to simulate.

watertap/unit_models/tests/test_steam_heater.py
adam-a-a: I see. I was assuming the unit was specific to heating a water stream--is it not?

ElmiraShamlou: This unit can heat both pure water and brine streams. Initially, I tested it with pure water; however, later I switched to seawater brine to align with common WaterTAP applications. This is reflected in revised tests.

watertap/unit_models/steam_heater.py
ElmiraShamlou: addressed

watertap/unit_models/steam_heater.py
ElmiraShamlou: addressed

watertap/unit_models/steam_heater.py
ElmiraShamlou: @adam-a-a I've implemented the second suggestion by incorporating a configuration option in the model. By default, this configuration sets the outlet pressure to match the saturation pressure, unless otherwise specified by the user.

watertap/unit_models/tests/test_steam_heater.py
adam-a-a: Overall, I take this to mean that if the property package does not have `Vap`, an error would occur if attempting to fix these vapor flowrates to 0 within the unit model. 

If that is the case, you could get around this by checking first that vapor exists as a phase in the prop pack, and if so, then fix the mass flowrates to 0 on the cold side.

adam-a-a:
I am unsure whether the Mac failure is something we want to just dodge with `requires_idaes_solver` or if this is something that we deem fixable. It appears that something is going wrong with the `report` method with a `NoneType` where a path is expected. @bknueven I am guessing you'd recall

ElmiraShamlou:
> I am unsure whether the Mac failure is something we want to just dodge with `requires_idaes_solver` or if this is something that we deem fixable. It appears that something is going wrong with the `report` method with a `NoneType` where a path is expected. @bknueven I am guessing you'd recall

@adam-a-a I believe it's because of Chen's approximation

watertap/unit_models/tests/test_steam_heater.py
ElmiraShamlou: The cold side can indeed include vapor as well. In the initial test, I had fixed it to only include liquid as a specific case example.

bknueven:
> I am unsure whether the Mac failure is something we want to just dodge with `requires_idaes_solver` or if this is something that we deem fixable. It appears that something is going wrong with the `report` method with a `NoneType` where a path is expected. @bknueven I am guessing you'd recall

The issue here is we call some external function not available without IDAES extensions. We should just use `requires_idaes_solver`.

@lbianchi-lbl we should probably consider updating the macOS tests given these external functions are now available for macOS, although that would require re-working the `requires_idaes_solver` handle to be a bit more subtle (really should be `ipopt_has_HSL_solvers`...

watertap/unit_models/steam_heater_0D.py
```diff
@@ -0,0 +1,151 @@
...
+            self.hot_side.config.property_package.component_list,
+            doc="Mass balance",
+        )
+        def outlet_liquid_mass_balance(b, t, j):
+            b.hot_side.properties_out[t].flow_mass_phase_comp["Vap", j].fix(0)
+            return (
+                b.hot_side.properties_in[t].flow_mass_phase_comp["Vap", j]
+                + b.hot_side.properties_in[t].flow_mass_phase_comp["Liq", j]
+                == b.hot_side.properties_out[t].flow_mass_phase_comp["Liq", j]
+            )
```
Morgan88888888, line 88: I don't quite understand why you need to have an extra mass balance. It seems to duplicate the pressure saturation constraints later. Additionally, it's not very good practice to fix any variable as 0, especially at the unit model level.

watertap/unit_models/steam_heater_0D.py
```diff
@@ -0,0 +1,151 @@
...
+            self.config.hot_side.property_package.get_metadata().get_derived_units
+        )
+
+        self.saturation_pressure_deviation = Var(
+            self.flowsheet().time,
+            initialize=0,
+            units=units_meta("pressure"),
+            doc="Difference between the outlet pressure and the saturation pressure of the condensed steam",
+        )
+
```
Morgan88888888: I'm not sure if it's important to have it as a variable. It doesn't seem like a very important variable to manipulate in later optimizations.

watertap/unit_models/steam_heater_0D.py
```diff
@@ -0,0 +1,151 @@
...
+        else:
+            self.saturation_pressure_deviation.fix()
+
+        @self.Constraint(self.flowsheet().time, doc="Saturation pressure constraint")
+        def outlet_pressure_sat(b, t):
+            return (
+                b.hot_side.properties_out[t].pressure
+                + b.saturation_pressure_deviation[t]
+                == b.hot_side.properties_out[t].pressure_sat
+            )
```
Morgan88888888: Do you have any specific reasons for NOT using an inequality constraint here?

watertap/unit_models/tests/test_steam_heater_0D.py
```diff
@@ -0,0 +1,80 @@
...
+            "property_package": m.fs.cold_side_properties,
+        },
+        delta_temperature_callback=delta_temperature_chen_callback,
+        flow_pattern=HeatExchangerFlowPattern.countercurrent,
+    )
+
+    m.fs.unit.hot_side_inlet.flow_mass_phase_comp[0, "Vap", "H2O"].set_value(6)
+    m.fs.unit.hot_side_inlet.flow_mass_phase_comp[0, "Liq", "H2O"].fix(0)
+    m.fs.unit.hot_side_inlet.temperature.fix(273.15 + 140)
+    m.fs.unit.hot_side_inlet.pressure[0].set_value(201325)
```
Morgan88888888: Can you explain the application where you need to vary only vapor fraction and pressure? One potential reason I could think of is that you introduced one variable and two equality constraints. Again, this setting is not very common to me.

watertap/unit_models/tests/test_steam_heater_0D.py
```diff
@@ -0,0 +1,80 @@
...
+    return m
+
+
+class TestSteamHeater0D(UnitTestHarness):
+    def configure(self):
+        m = build()
+
+        self.unit_solutions[
+            m.fs.unit.hot_side_inlet.flow_mass_phase_comp[0, "Vap", "H2O"]
+        ] = 2.92784431885
```
Morgan88888888: I'm not very familiar with the guidance in WaterTAP to create tests. I'm going to follow the rules in IDAES temporarily. It seems too simple to me; at least, testing is needed for building blocks. Also, the most important saturation condition is not tested.

Morgan88888888 (changes requested):
My main comment is that I'm not convinced about introducing a new unit model to replicate the 0D heat exchanger with additional saturation constraints. In my previous experience with power plant modeling, you could simply build a heat exchanger and impose additional performance constraints. As long as you have good initial values, the optimization should work fine. Please correct me if the steam heater model includes more than that.

watertap/unit_models/tests/test_steam_heater_0D.py
adam-a-a: Hi @Morgan88888888 - I believe @ElmiraShamlou is using a unit test harness, which actually runs a whole sequence of standard tests (e.g., testing the build). The nice part for developers familiar with it is that you don't have to write all tests from scratch, and you are less prone to forgetting one of the "must-have" tests. On the other hand, for those unfamiliar with it, it can be a little unclear what is happening. Here is a how-to on unit test harness: https://watertap.readthedocs.io/en/latest/how_to_guides/how_to_use_unit_test_harness.html

To see what tests run under the hood, you can also check out `watertap\unit_models\tests\unit_test_harness.py`

ElmiraShamlou:
> My main comment is that I'm not convinced about introducing a new unit model to replicate the 0D heat exchanger with additional saturation constraints. In my previous experience with power plant modeling, you could simply build a heat exchanger and impose additional performance constraints. As long as you have good initial values, the optimization should work fine. Please correct me if the steam heater model includes more than that.

@Morgan88888888 The primary reason for developing this unit model is to calculate the steam consumption rate (given complete condensation of the steam at the outlet) across various flowsheets and configurations in treatment systems like membrane distillation where the steam requirement significantly varies with design. Given that the hot inlet steam flow rate is variable, additional steps are added to the main heat exchanger's initialization method to improve initialization for this specific case. From my experience, initializing a steam heater in the treatment flowsheet without these additional steps was challenging. If you have experience using heat exchangers in similar applications without the need for modifications in initialization, sharing your experience could be greatly helpful.  Another consideration is that we plan to add flowsheets containing multiple steam heaters; having this unit prevents the need to replicate extra constraints for each one.

bknueven:
>  Another consideration is that we plan to add flowsheets containing multiple steam heaters; having this unit prevents the need to replicate extra constraints for each one.

I strongly agree with this -- as soon as we have the same code in two different places a refactor with tests like this makes the most sense for maintainability and reliability.

watertap/unit_models/steam_heater_0D.py
ElmiraShamlou: @Morgan88888888 WaterTAP property packages require modeling the complete condensation of the steam at saturation condition differently than IDAES, where IAPWS is used for steam properties. In the case of the WaterTAP property package, it's necessary to ensure a liquid state at the outlet through mass balance. However, maintaining this mass balance alone doesn't guarantee the saturation condition. Therefore, we also need to impose the saturation condition using saturation pressure. Similarly, enforcing the saturation pressure condition alone does not guarantee that the outlet remains in a liquid state, necessitating the inclusion of both saturation pressure and mass balance to meet these conditions.

watertap/unit_models/tests/test_steam_heater_0D.py
ElmiraShamlou: @Morgan88888888 In this specific test example, it is assumed that there is no liquid condensate in the incoming heating steam. For the WaterTAP property package, we address this by specifying the fraction of each phase in the flow:
https://github.com/watertap-org/watertap/blob/main/watertap/property_models/water_prop_pack.py

watertap/unit_models/tests/test_steam_heater_0D.py
ElmiraShamlou: @Morgan88888888 I echo Adam's points and would like to add that I am also integrating few more unit_solutions into this. I will include these updates in my next commit to this PR which includes the flowsheet integration.

watertap/unit_models/steam_heater_0D.py
ElmiraShamlou: @Morgan88888888 Thanks for the comment. I definitely agree with you that using an inequality constraint is a much better approach in this case. Specifically, in the MD flowsheet, I specify the steam pressure based on the assumed source—whether it's low, medium, or high pressure, along with the pressure value. Initially, I had not fixed the inlet pressure, opting instead to fix the saturation at the outlet to aid in the analysis. I've switched to using an inequality constraint and fixed the inlet pressure in my previous commit.

watertap/costing/unit_models/heat_exchanger.py
```diff
@@ -28,12 +28,20 @@ def build_heat_exchanger_cost_param_block(blk):
         units=pyo.units.dimensionless,
     )
 
+    blk.steam_cost = pyo.Var(
+        initialize=0.008,
+        units=pyo.units.USD_2018 / (pyo.units.kg),
+        doc="steam cost per kg",
+    )
```
MarcusHolly, line 35: It would be nice to have this update reflected in the documentation, but if it would hold up this PR, we can just update the documentation later (make an issue?)

MarcusHolly (approved):
LGTM - just one comment that may or may not hold up this PR

watertap/costing/unit_models/heat_exchanger.py
ElmiraShamlou: updated the hx doc

Morgan88888888 (approved):
Looks good to me! 👍 Thanks for your work on this.

PR #1359, opened 2024-04-15
spiral-wound vs flat-sheet config options to RO and OARO

zacharybinger:
## Fixes/Resolves:

Using this draft to help debug  #1272

How the flux is calculated

### Legal Acknowledgement

### DIFF
--- a/watertap/unit_models/reverse_osmosis_1D.py
+++ b/watertap/unit_models/reverse_osmosis_1D.py
@@ -188,7 +188,7 @@ class ReverseOsmosis1DData(ReverseOsmosisBaseData):
         )
         def eq_mass_flux_equal_mass_transfer(b, t, x, p, j):
             return (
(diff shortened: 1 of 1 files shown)


watertap/unit_models/reverse_osmosis_1D.py
```diff
@@ -188,7 +188,7 @@ def eq_connect_mass_transfer(b, t, x, p, j):
         )
         def eq_mass_flux_equal_mass_transfer(b, t, x, p, j):
             return (
-                b.flux_mass_phase_comp[t, x, p, j] * b.width
+                b.flux_mass_phase_comp[t, x, p, j] * (b.area / b.length)
```
zacharybinger: This change breaks the test for:
case_studies/wastewater_resource_recovery/dye_desalination/tests/test_dye_desalination_withRO.py

but only for that one test and only on the linux checks. This change is critical to #1272

Issue #1360, opened 2024-04-18
Adding cost documentation


adam-a-a:
**Running to do list:**

- [x] Add cost landing page #1361 

**Costing technical references for** 

- [x] Mixer (can be minimal but should clean up the template at least) @MarcusHolly #1456 
- [x] RO @adam-a-a #1395 
- [x] GAC @hunterbarber #1375
- [x] aeration tank @MarcusHolly #1373
- [x] anaerobic digester @agarciadiego #1433 
- [x] clarifier @MarcusHolly #1373
- [x] compressor @ElmiraShamlou #1442  
- [x] crystallizer @OOAmusat #1391 
- [x] cstr injection @MarcusHolly #1373
- [x] cstr @MarcusHolly #1373
- [x] dewatering @MarcusHolly #1373
- [x] electrodialysis @lbibl @hunterbarber #1454 
- [x] electrolyzer @hunterbarber #1392
- [x] electroN-P @luohezhiming #1377 
- [x] ERD @adam-a-a #1395 
- [x] evaporator @ElmiraShamlou #1442 
- [x] heat exchanger @ElmiraShamlou #1442 
- [x] heater_chiller @ElmiraShamlou #1442 
- [x] IX @kurbansitterley #1402 
- [x] MD @ElmiraShamlou #1442 
- [x] NF @adam-a-a #1395
- [x] OARO @luohezhiming #1377 
- [x] Pressure exchanger  @adam-a-a #1395
- [x] Pump @adam-a-a #1395
- [x] stoichiometric reactor @avdudchenko #1449
- [x] thickener @MarcusHolly #1373
- [x] UV AOP @luohezhiming #1377

adam-a-a:
Woohoo!

PR #1361, opened 2024-04-18
Costing docs- initial template rollout

adam-a-a:
## Fixes/Resolves:

Partially addresses #1360 

## Summary/Motivation:
Initiates our effort to document costing of unit models by creating a landing page and template for all unit models currently in the `costing/unit_models` directory. 

After this PR, assignees should manually fill out their unit model costing rst files.

- landing page for unit model costing docs
- template for unit model costing doc
- automation script that writes out template and index file for all units listed in `costing.unit_models`

### Legal Acknowledgement


docs/technical_reference/costing/automate_detailed_unit_costing_rst_file.py
```diff
@@ -0,0 +1,105 @@
...
+        #     f"   pair: watertap.unit_models.zero_order.{unit_name_list[i]};{unit_name_list[i]}",
+        #     f".. currentmodule:: watertap.unit_models.zero_order.{unit_name_list[i]}",
+        #     f".. automodule:: watertap.unit_models.zero_order.{unit_name_list[i]}",
+        # ]
+
+        # append unit doc to index
+        with open("detailed_unit_model_costing.rst", "a") as f:
+            f.write(f"   {unit_name_list[i]}\n")
+
+        with open(f"{unit_name_list[i]}.rst", "w", encoding="utf-8") as f:
```
bknueven: So we can reuse this when more models are added, I think this should check for the existence of f"{unit_name_list[i]}.rst" before opening this file.

adam-a-a: As in- "do not overwrite the file if it exists already"?

adam-a-a: Or simply "only open and write to the file if it has been created already"?

bknueven: > As in- "do not overwrite the file if it exists already"?

This one -- I assume these will be filled-out in subsequent PRs, and we don't want to overwrite files which already exist.

adam-a-a: I was originally contemplating whether to automatically fill the tables with model params/vars as was done previously for the ZO model costing documentation. However, I am thinking we might as well just let each assignee manually fill out the relevant sections without needing to run this script (at least for the time being.

I can add a boolean option to enable/disable overwriting here, but unsure how necessary it would be if we turn the guidance into--don't run the script--just manually edit your rst file as needed.

adam-a-a:
@hunterbarber - I used the same structure that you used for GAC costing documentation, except I added a separate section for Costing Parameters (and Costing Variables). Check it out and let me know if you have thoughts on why it should stay as one section, as you had it.

docs/technical_reference/costing/automate_detailed_unit_costing_rst_file.py
adam-a-a: Actually, we'll want you're suggestion for later on, when new unit model methods get added. So I will try to add it in this PR.

docs/technical_reference/costing/WT_unit_costing_for_docs.csv
```diff
@@ -0,0 +1,27 @@
+﻿name,unit
```
adam-a-a, line 1: I could probably get rid of this csv and just read filenames from costing/unit_models, but isn't _necessary_ to make that change yet.

docs/technical_reference/costing/automate_detailed_unit_costing_rst_file.py
```diff
@@ -0,0 +1,191 @@
...
+                "Default costing methods are provided for the unit models listed below. However, users should supply their own cost relationship, if possible, instead of relying completely on the defaults.\n\n"
+            )
+            f.write(".. toctree::\n")
+            f.write("   :maxdepth: 1\n\n")
+        with open("detailed_unit_model_costing.rst", "r") as f:
+            lines = f.readlines()
+
+    for i, unit_title in enumerate(unit_title_list):
+        for l, line in enumerate(lines):
+            # set on_list flag to False in unit name not found in landing page list
```
MarcusHolly:
```suggestion
            # set on_list flag to False if unit name not found in landing page list
```

MarcusHolly (approved):
LGTM. One minor comment: should we include a link to the unit model documentation in the unit model costing method page? I think we had intended to include a link to the unit model costing method page in the unit model documentation, so doing the inverse on top of that might be redundant. Although doing so couldn't hurt...

docs/technical_reference/costing/automate_detailed_unit_costing_rst_file.py
```diff
@@ -0,0 +1,191 @@
...
+                    else:
+                        f.write(f"{unit_title} Costing Method")
+                        f.write("\n")
+                        f.write("=" * len(f"{unit_title} Costing Method\n"))
+
+                    # TODO: add parameter tables
+                    f.write("\n\nCosting Method Parameters\n")
+                    f.write("+" * len("Costing Method Parameters"))
+                    f.write(
+                        f"\n\nThe following parameters are constructed when applying the `cost_{unit_name_list[i]}` costing method in the ``watertap_costing_package``:\n\n"
```
hunterbarber: Is it important to clarify in this section that the parameters exist on m.fs.costing.unit (and are used across the flowsheet) while the variables on m.fs.unit.costing (and unique to each unit op)?

docs/technical_reference/costing/automate_detailed_unit_costing_rst_file.py
```diff
@@ -0,0 +1,191 @@
...
+                    f.write("-" * len("Code Documentation"))
+                    f.write(f"\n\n* :mod:`watertap.unit_models.{unit_name_list[i]}`")
+                    f.write(
+                        f"\n* :mod:`watertap.costing.unit_models.{unit_name_list[i]}`"
+                    )
+
+                    f.write("\n\nReferences\n")
+                    f.write("-" * len("References"))
+                    f.write(
+                        "\nAim to include at least one reference in most cases, but delete this section if no references used for cost relationships/default values"
```
hunterbarber, line 189: A disclaimer that the default values of parameters (say unit cost) correspond to the dollar value of the year of the reference (e.g. USD2020). Not necessarily saying this should go in the references, but if supplying units of default parameters should $ be listed as USDYYYY in all of these docs?

docs/technical_reference/costing/automate_detailed_unit_costing_rst_file.py
```diff
@@ -0,0 +1,191 @@
...
+                    f.write(
+                        "        C_{op,tot} = C_{op,example1}+C_{op,example2}+C_{op,other}"
+                    )
+                    f.write("\n\n    .. math::\n\n")
+                    f.write(
+                        "        C_{op,example1} = fill in equation for each component in total opex equation\n\n "
+                    )
+
+                    # TODO: add module directives to unit and cost method
+                    f.write("\nCode Documentation\n")
```
hunterbarber, line 180: I'm assuming the author can should just include both the 0D and 1D models here, or other instances where unit_model:cost_model aren't 1:1. Say the some units having more dependencies on rectifier, heat exchanger, pumps, etc.

hunterbarber: Because of that, I don't know if it makes sense to have once section as the model _directly_ using the method/class and an optional section for other costing dependencies. Just a thought.

docs/technical_reference/costing/automate_detailed_unit_costing_rst_file.py
```diff
@@ -0,0 +1,191 @@
...
+    "cstr": "Completely Stirred Tank Reactor",
+    "dewatering": "Dewatering Unit",
+    "electroNP": "ElectroN-P",
+    "gac": "Granular Activated Carbon",
+    "heater_chiller": "Heater/Chiller",
+    "uv_aop": "UV with Advanced Oxidation Processes",
+}
+
+
+if __name__ == "__main__":
```
hunterbarber, line 35: Nice!

hunterbarber (commented):
Just a couple comments. After seeing responses I'll come back and approve.

docs/technical_reference/costing/automate_detailed_unit_costing_rst_file.py
adam-a-a:
```suggestion
            # set on_list flag to False if unit name not found in landing page list
```

docs/technical_reference/costing/automate_detailed_unit_costing_rst_file.py
adam-a-a: Perhaps a link to other docs on costing, wherever currency units are explained, with a brief description on the fact that we can resolve variations in defaults (or provided values) with `base_currency`

hunterbarber: Everything would be converted to base flowsheet currency when solving, so that link probably should exist at a higher level documentation like the `watertap_costing_package` just so it is not unnecessarily duplicated in every unit model. For these docs then, are you agreeing that the parameter tables should be completed using USD_YYYY instead of $ for the currency units?

docs/technical_reference/costing/automate_detailed_unit_costing_rst_file.py
adam-a-a: So instead of including any link to unit models here, I can just include the cost method, and later, within unit model tech refs, we can point to the relevant default costing method for the unit at hand. For example, in the tech ref for GAC, under costing section, We would just have a sentence pointing to the default costing method used and its tech ref (or something like that).

docs/technical_reference/costing/automate_detailed_unit_costing_rst_file.py
adam-a-a: @hunterbarber do you think this should be duplicated on every costing tech ref, or should it just be in one place in the higher level costing documentation (if not there already)? I could go either way but would lean towards duplicating across all templates so that users don't miss the message.

docs/technical_reference/costing/automate_detailed_unit_costing_rst_file.py
adam-a-a: Yes, if reporting default parameters, makes sense to report the currency year.

hunterbarber (approved):

PR #1362, opened 2024-04-19
Replacing Expr_if & Changing Bio_P Default

MarcusHolly:
## Fixes/Resolves:

Issue #934 

Replaces the Expr_ifs in the ASM2d/ADM1 translator block (with the exception of 1 set of Expr_ifs) and changes the `bio_P` default to `False` such that the default setting does not set P components to 0 in the translator block.

### Legal Acknowledgement


watertap/unit_models/translators/translator_asm2d_adm1.py
```diff
@@ -622,12 +629,13 @@ def XN_org(blk, t):
...
-                    eps * pyunits.kg / pyunits.m**3,
+                return smooth_max(
+                    0 * pyunits.kg / pyunits.m**3,
                     (blk.properties_in[t].conc_mass_comp["X_S"] - blk.XN_org[t]) * 0.4,
+                    blk.eps,
                 )
 
+            # TODO: Can this be replaced with smooth_max or smooth_min?
             @self.Expression(self.flowsheet().time, doc="Protein mapping")
             def Xpr_mapping(blk, t):
```
MarcusHolly, line 640: I'm not sure if this (and line 1039) can be expressed as a `smooth_max` or `smooth_min`

watertap/unit_models/translators/translator_asm2d_adm1.py
```diff
@@ -395,18 +402,18 @@ def SN_org(blk, t):
...
         @self.Expression(self.flowsheet().time, doc="Monosaccharides mapping")
         def Ssu_mapping(blk, t):
-            return Expr_if(
-                blk.SN_org[t] >= blk.properties_in[t].conc_mass_comp["S_F"],
-                eps * pyunits.kg / pyunits.m**3,
+            return smooth_max(
+                0 * pyunits.kg / pyunits.m**3,
                 blk.properties_in[t].conc_mass_comp["S_F"] - blk.SN_org[t],
+                blk.eps,
             )
```
luohezhiming, line 409: I am a little confused here, I think the original equation, if A>=B, then we have `eps`, but in the the smooth_max function, you directly change this `eps` to 0 if I understand correctly, and the `eps` in the smooth_max function is smoothing parameter set at `1e-10`. My first question is: does change previous `eps` to `0` have no impacts to all of the examples? And why do you set `1e-10` for the smoothing parameter

andrewlee94: Setting `eps` to 0 means it is no longer smooth, and can thus trigger numerical issues.

MarcusHolly: So there are two instances of `eps` in this file. `eps` is 0 and `blk.eps` is 1e-10. `blk.eps` is the one used in the smooth_max and smooth_min functions. I think I should just directly replace instances of `eps` with zero to make this less confusing

MarcusHolly: @luohezhiming I changed the epsilon used for smooth_max and smooth_min to `eps_smooth` to make things a bit more transparent. Also, the 1e-10 is the smoothing factor used in the ASM1/ADM1 translator, so I just carried that value over. It can be changed if necessary.

andrewlee94: @MarcusHolly You should not be using `eps=0` in the `smoothmax` and `smoothmin` functions - that defeats the entire purpose if those functions. If you really want a non-smooth max or min then you should refactor these to use `Expr_if`.

MarcusHolly: eps=1e-10 in the smooth_max. Are you saying it shouldn't be that close to 0 @andrewlee94?

andrewlee94: Probably not - the value you use for `eps` represents how smooth the operation is. Smaller values mean less smoothness and thus more chance for numerical issues (but less numerical error). Ultimately, `eps` needs to be tuned for the application, and should be a few orders of magnitude less than the inputs to the `max` operation.

MarcusHolly: Changed the `eps_smooth` value to 1e-4 since this seems to give the best solver performance in the `BSM2_P_extension.py` flowsheet.

MarcusHolly:
When `eps_smooth` is 1e-10, the BSM2_P_extension test fails locally but passes the check here; however, when `eps_smooth` is 1e-4, the BSM2_P_extension test passes locally but fails the check here. Any thoughts @andrewlee94 or @adam-a-a ?

andrewlee94:
@MarcusHolly That is indicative of poor scaling. What do the diagnostics tools tell you?

watertap/unit_models/translators/translator_asm2d_adm1.py
luohezhiming: So the `eps` now only represents the smoothing parameter, we can try to vary the `smooth_eps` if all tests get passed.

MarcusHolly:
> @MarcusHolly That is indicative of poor scaling. What do the diagnostics tools tell you?

The scaling is definitely poor and/or non-existent in some places - this is a known issue. The DiagnosticToolbox identified 155 potential evaluation errors and variables/constraints with extreme jacobians, but none of these are new issues and are also present in the version currently merged in main.

watertap/examples/flowsheets/case_studies/full_water_resource_recovery_facility/test/test_BSM2_P_extension.py
```diff
@@ -57,56 +57,72 @@ def test_solve(self, system_frame):
...
         )
+
+
+class TestFullFlowsheetBioPTrue:
+    @pytest.mark.requires_idaes_solver
+    @pytest.fixture(scope="class")
+    def system_frame(self):
+        m, res = main()
+        m.results = res
+        return m
```
MarcusHolly: Looks like I never finished this test... I'll work on it tomorrow

MarcusHolly:
@luohezhiming I've added a config option into the flowsheet so that the user can specify whether to use `bio_P=True` or `bio_P=False`.

luohezhiming:
> @luohezhiming I've added a config option into the flowsheet so that the user can specify whether to use `bio_P=True` or `bio_P=False`.

Looks good, Let's see if tests will get pass

luohezhiming (approved):
LGTM! Does all tests will pass with (using the new tear_guesses) `bioP=True`?

agarciadiego (approved):
LGTM

watertap/examples/flowsheets/case_studies/full_water_resource_recovery_facility/BSM2_P_extension.py
```diff
@@ -530,58 +531,113 @@ def initialize_system(m):
...
-            (0, "X_PHA"): 0.0044,
-            (0, "X_PP"): 2.76,
-            (0, "X_S"): 3.83,
-        },
-        "temperature": {0: 308.15},
-        "pressure": {0: 101325},
-    }
+    if bio_P:
+        # Initial guesses for flow into first reactor
+        tear_guesses = {
```
adam-a-a, line 536: How did you arrive at these new guesses?

adam-a-a (commented):
My general question is how do the changes affect flowsheet convergence? I'd be interested to know, e.g., the number of iterations to solve before and after the changes, notable differences in IPOPT output, or anything of that nature. It would just be informative for all involved I think.

MarcusHolly:
> My general question is how do the changes affect flowsheet convergence? I'd be interested to know, e.g., the number of iterations to solve before and after the changes, notable differences in IPOPT output, or anything of that nature. It would just be informative for all involved I think.

Before changes (Bio_P = True):
image


After changes (Bio_P = True):
image


After changes (Bio_P = False):
image

watertap/examples/flowsheets/case_studies/full_water_resource_recovery_facility/BSM2_P_extension.py
MarcusHolly: I got the initial guesses from Chenyu's ElectroN-P flowsheet, and then I made some very slight modifications to them.

andrewlee94:
That is a lot of iterations for solving a square problem - if IPOPT needs more than 50 iterations (and in practice often far fewer) on a square problem it often indicates the model has issues that should be resolved.

adam-a-a:
> That is a lot of iterations for solving a square problem - if IPOPT needs more than 50 iterations (and in practice often far fewer) on a square problem it often indicates the model has issues that should be resolved.

While we know that this model is problematic, we are continuously refining it as we go. We also suspect that poor scaling is significantly hampering the model. This flowsheet would actually be a great (and likely challenging) test case for the new scaling tools underway. In any case, we're making incremental improvements with every PR (or at least trying to).

adam-a-a (approved):

PR #1363, opened 2024-04-22
Pressure Changer and Coag & Floc Test Harness

MarcusHolly:
## Fixes/Resolves:

#1302 

### Legal Acknowledgement


watertap/unit_models/tests/test_coag_floc_model.py
```diff
@@ -9,585 +9,167 @@
...
+        ] = 0.009112749
+        self.unit_solutions[m.fs.unit.outlet.flow_mass_phase_comp[0, "Liq", "TDS"]] = (
+            0.01
         )
-        with pytest.raises(ConfigurationError, match=re.escape(error_msg)):
-            model.fs.properties = NaClParameterBlock()
-            model.fs.unit = CoagulationFlocculation(
-                property_package=model.fs.properties
-            )
-
```
adam-a-a: Where are we maintaining the checks on ConfigurationError and other expected exception handling? Is that part of the test harness?

adam-a-a:
@lbianchi-lbl Maybe I am missing it, but I am not seeing the codecov checks here. Do codecov checks not run if the only changed files are test files? I wanted to check to see if test coverage decreased or not with this PR.

lbianchi-lbl:
> @lbianchi-lbl Maybe I am missing it, but I am not seeing the codecov checks here. Do codecov checks not run if the only changed files are test files? I wanted to check to see if test coverage decreased or not with this PR.

@adam-a-a thanks for bringing this up.

- It looks like the Codecov step in the CI jobs is failing (which is consistent with the coverage reports not showing up as comments on this PR), but without causing the entire job to fail because the `fail_ci_if_error` option in `codecov/codecov-action` is not set to `true`
- I've restarted the two Python 3.8 jobs where the coverage report is generated and uploaded, which might fix the issue for this specific instance
- A more permanent fix requires a change to the CI configuration. I'm currently working on the same issue for IDAES so I might have something ready to use here as well

**EDIT** it looks like restarting the job didn't help. I'll try to have a PR up first thing tomorrow.

hunterbarber:
I'd prefer to review this with the codecov for context (since largely this is minor change assuming the coverage is equivalent). Maybe @lbianchi-lbl can you just update the team when that gets resolved?

lbianchi-lbl:
> I'd prefer to review this with the codecov for context (since largely this is minor change assuming the coverage is equivalent). Maybe @lbianchi-lbl can you just update the team when that gets resolved?

#1365 should address this, so feel free to leave a review to get it merged.

lbianchi-lbl:
@adam-a-a @hunterbarber FYI: I realized I've made a mistake in the Codecov configuration in #1365, which causes the Codecov comment to erroneously report a 50% drop in coverage because the comment is created before all uploading jobs have run. This should be fixed by #1366.

**EDIT** the Codecov comment now seems to be working properly:

image

watertap/unit_models/tests/test_coag_floc_model.py
hunterbarber: Yea, my comment was also going to be to re-add this in. [I just appended mine at the end of GAC after the harness](https://github.com/watertap-org/watertap/blob/main/watertap/unit_models/tests/test_gac.py#L312). Thoughts if that is the best implementation?

MarcusHolly: Seems like a good implementation to me. I've addressed this in the latest commit.

hunterbarber (approved):
LGTM after re-adding the logging coverage. Maybe we want to add a way to test error and warning logs in the test harness, but lets save that for when I also add the costing checks into the harness (not 1.0).

luohezhiming (approved):
LGTM

adam-a-a (approved):

PR #1371, opened 2024-04-28
test harness for unit models of waste treatment 

agarciadiego:
## Fixes/Resolves:
Addresses for 3 unit models #1302 

## Summary/Motivation:
Adds unit test harness for:

- [x] CSTR
- [x] CSTR injection
- [x] Thickener

- changes in the test for CSTR, CSTR injection, Thickener
-

### Legal Acknowledgement


watertap/unit_models/tests/test_cstr.py
```diff
@@ -71,214 +57,195 @@
...
-                "Pressure Change": sapon.fs.unit.deltaP[0],
-            }
-        }
+        return m
 
 
 class TestInitializers:
     @pytest.fixture
     def model(self):
+        m = build()
```
MarcusHolly, line 248: This class seems mostly unchanged. Are you able to make use of `self.unit_solutions` here instead of the assert statements?

watertap/unit_models/tests/test_cstr_injection.py
```diff
@@ -82,237 +69,249 @@
...
+
+
+class TestSaponification(object):
+    @pytest.mark.solver
+    @pytest.mark.skipif(solver is None, reason="Solver not available")
+    @pytest.mark.component
+    def test_conservation(self):
+
+        sapon = build()
+
```
MarcusHolly: Is this actually testing conservation? It seems unfinished

watertap/unit_models/tests/test_cstr.py
```diff
@@ -392,85 +359,30 @@ def test_block_triangularization(self, model):
...
-        )
-        m.fs.unit.inlet.conc_mass_comp[0, "X_ND"].set_value(
-            0.0053166 * units.kg / units.m**3
-        )
 
-        m.fs.unit.inlet.temperature[0].set_value(308.15 * units.K)
-        m.fs.unit.inlet.pressure[0].set_value(84790.0 * units.Pa)
+class TestCosting(UnitTestHarness):
+    def configure(self):
+        m = build_ASM1()
```
MarcusHolly, line 365: In the cstr with injection test, you also test the solutions for `build_ASM1`. Is there a reason to not do the same here?

watertap/unit_models/tests/test_cstr_injection.py
```diff
@@ -338,184 +337,71 @@ def test_get_performance_contents(self, sapon):
...
-        )
         assert pytest.approx(0.0132455, rel=1e-5) == value(m.fs.costing.LCOW)
 
-    @pytest.mark.unit
-    def test_report(self, model):
-        model.fs.unit.report()
+        return m
 
 
 @pytest.mark.build
```
MarcusHolly, line 400: Seems like the ASM2d and modified ASM2d tests below are just making sure the build works. Do we also want to test a few of the solutions for these builds?

watertap/unit_models/tests/test_cstr.py
agarciadiego: In the injection CSTR, the saponification test is testing the CSTR part but requires a test with injection, which is why we test ASM1. ASM1 reaction is being tested in the reaction parameter block test and the flowsheets

watertap/unit_models/tests/test_cstr_injection.py
agarciadiego: Same as above, we are really only testing the builds here as the reaction blocks are being tested in their own tests

watertap/unit_models/tests/test_cstr.py
agarciadiego: we are testing the initialization methods here not the solutions

watertap/unit_models/tests/test_cstr_injection.py
agarciadiego: deleted

watertap/unit_models/tests/test_cstr.py
MarcusHolly: If I were to channel my "inner Tim", I'd say is testing the initialization methods necessary? If not, we should delete it. If so, we should consider doing something similar for other unit models or adding it to the harness (which might not be feasible/a good idea)

andrewlee94: To add my 2 cents: the `Initializers` are not part of WaterTAP and are tested as part of IDAES, so the question becomes twofold:

1. do you trust IDAES not to break things on you, and
2. what do you do if these tests start failing?

From a software engineering point of view, the "best" answer to this would be to add this case study to the IDAES backward compatibility tests. That way IDAES will see immediately if they break the tests, rather than having to wait for WaterTAP to discover and report it.

MarcusHolly (approved):
LGTM (other than the unrelated test failure). I'll just note here that the team internally discussed the inclusion of the initializer tests in the CSTR test file, and we ultimately decided that it's fine to add additional tests outside of the TestHarness framework.

savannahsakhai (approved):
LGTM

PR #1378, opened 2024-05-07
Passing through the base solver attributes for solver wrapper

bknueven:
## Summary/Motivation:
@hunterbarber found that `ipopt-watertap` does not have an `available` attribute, which is used in some Pyomo workflows. This PR would update the solver wrapper for `ipopt-watertap` to pass-through the attributes for the base solver.

- Generalize `executable` implementation in the `ipopt-watertap` wrapper to `__getattr__` to retrieve missing attributes from the `_base_solver`.

### Legal Acknowledgement

### DIFF
--- a/watertap/core/plugins/tests/test_solvers.py
+++ b/watertap/core/plugins/tests/test_solvers.py
@@ -75,6 +75,11 @@ class TestIpoptWaterTAP:
         assert s.available()
 
     @pytest.mark.unit
+    def test_attribute_nonpassthrough(self, s):
+        with pytest.raises(AttributeError, match=".*this_attribute_should_not_exist.*"):
+            s.this_attribute_should_not_exist
+
+    @pytest.mark.unit
     def test_presolve_scales_constraints_and_relaxes_bounds(self, m, s):
         s._scale_constraints(m)
         for c, sf in s._scaling_cache:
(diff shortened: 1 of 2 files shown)


lbianchi-lbl (approved):

watertap/core/plugins/tests/test_solvers.py
```diff
@@ -69,6 +69,11 @@ def test_pyomo_registration(self, s):
     def test_idaes_registration(self):
         assert get_solver().__class__ is IpoptWaterTAP
 
+    @pytest.mark.unit
+    @pytest.mark.requires_idaes_solver
+    def test_attribute_passthrough(self, s):
+        assert s.available()
+
```
lbianchi-lbl, line 76: It might make sense to also test the non-existing-attribute code path:

```suggestion
    @pytest.mark.unit
    @pytest.mark.requires_idaes_solver
    def test_attribute_passthrough(self, s):
        assert s.available()
        
    @pytest.mark.unit
    def test_attribute_nonpassthrough(self, s):
        with pytest.raises(AttributeError, match=".*this_attribute_should_not_exist.*"):
            s.this_attribute_should_not_exist
```

hunterbarber (approved):

PR #1383, opened 2024-05-10
Add LSRRO GUI

MarcusHolly:
## Fixes/Resolves:

Issue #1117 

## Summary/Motivation:
Adds LSRRO GUI

### Legal Acknowledgement


MarcusHolly:
@adam-a-a @zacharybinger @TimBartholomew A few things I'd like feedback on:

1. Are there additional outputs we'd want to display? 
image

2. Should some of the config options be removed (i.e. we don't want users to touch them) or config options that should be added to the interface?
image

3. Are the default values and bounds on the config options reasonable?

adam-a-a:
@MarcusHolly Starting with the outputs, you can check out the multisweep file for the LSRRO flowsheet, which will contain all the outputs that I included for supplementary data in our [paper](https://www.sciencedirect.com/science/article/abs/pii/S0011916423000395). I should note that this might be one of the more difficult cases to export, if we are to try to include all details that could be of interest to the user.

For options, I would say the "A_case" and "B_case" arguments should be included. Each one would allow the user to either fix values, optimize each value for each stage, or arrive at single optimum for all stages for the water and salt permeability coefficients of the membrane (i.e., A and B values, respectively). "AB_tradeoff" could also be included, and this allows the user to assume an equality constraint relating A and B values to each other, an inequality constraint, or no constraint at all.

The defaults should be reasonable but not sure which bounds you're referring to.

MarcusHolly:
> @MarcusHolly Starting with the outputs, you can check out the multisweep file for the LSRRO flowsheet, which will contain all the outputs that I included for supplementary data in our [paper](https://www.sciencedirect.com/science/article/abs/pii/S0011916423000395). I should note that this might be one of the more difficult cases to export, if we are to try to include all details that could be of interest to the user.
> 
> For options, I would say the "A_case" and "B_case" arguments should be included. Each one would allow the user to either fix values, optimize each value for each stage, or arrive at single optimum for all stages for the water and salt permeability coefficients of the membrane (i.e., A and B values, respectively). "AB_tradeoff" could also be included, and this allows the user to assume an equality constraint relating A and B values to each other, an inequality constraint, or no constraint at all.
> 
> The defaults should be reasonable but not sure which bounds you're referring to.

@adam-a-a The bounds are only visible in the UI file (begins at L33), not the GUI itself. I've set the default values based on what was given in the flowsheet, but I set the bounds somewhat arbitrarily.

watertap/examples/flowsheets/lsrro/lsrro_ui.py
```diff
@@ -0,0 +1,617 @@
...
+                "max_val": 1,  # optional
+                "min_val": 0,  # optional
+            },
+            "FeedFlowRate": {
+                "name": "FeedFlowRate",
+                "display_name": "Feed Flow Rate (m3/s)",
+                "values_allowed": "float",
+                "value": 1e-3,  # default value
+                "max_val": 1,  # optional
+                "min_val": 1e-5,  # optional
```
adam-a-a: can `None` be supplied for bounds?

MarcusHolly: Yes, bounds do not have to be provided.

MarcusHolly:
Updated Outputs
image
image
image

adam-a-a:
How will the AB constraint input be displayed, and what does the required input look like?

MarcusHolly:
@adam-a-a 
Configuration Options (Maximum NaCl Permeability hidden by dropdown):
image

Input:
image
image
image
image

adam-a-a:
One issue I am noticing from the snapshot is that the RO water permeability coefficient, salt permeability coefficient, and channel height are showing up as 0 in your snapshot. Since all of these are small values (especially the first 2), you probably need to adjust `rounding`.  

Additionally, in this flowsheet, stage 1 is the conventional RO stage, whereas all subsequent stages are low-salt-rejection (LSR) RO stages, but each stage is currently labled just as "RO". It'd be nice to have differentiation to make it clearer to users who aren't familiar with the configuration.

MarcusHolly:
@adam-a-a I've addressed your comments, as shown here:
image

TimBartholomew:
A few comments:
- I think we should change some of the display units and rounding as follows (hmm interestingly it seems the units are good for the outputs but not the inputs):
   - All pressures should be in bar (round to 0.1)
   - The channel height should be mm (round to 0.1)
   - The water permeability coefficient should be LMH/bar (round to 0.01)
   - The salt permeability coefficient should be LMH (round to 0.01)
   - Water flux should be LMH (round to 0.01)
   - Salt flux should be gMH (round to 0.1)
- I think the build configuration arguments should be just the things that are necessary for the build, this means removing:
   - feed volumetric flowrate
   - feed salt concentration
   - volumetric water recovery
- The default build configurations should be the default used in the paper, I think the only change necessary is that water permeability coefficient was not optimized and that the water & salt permeability coefficient constraint was none
- I think the membrane inputs should be grouped by stage (i.e. Stage 1, Stage 2, Stage 3), then we should label the variables consistently (right now the input variables have names like "RO membrane area - stage 1" and "RO 1 channel height"). I think we can drop the RO, and just do "Stage 1 membrane area", "Stage 1 channel height". I wanted to drop "Stage 1" since its under the Stage 1 category, but I realized that if we do sensitivity analyses it only shows the variable name and not its grouping, so we still need the stage in the variable name.
- RO X outlet pressure should be relabeled Stage X permeate pressure
- Feed inputs should be volumetric flowrate in m3/h and salinity in g/L, not NaCl mass flowrate. I also think we should drop the feed temperature (this model is using the NaCl property package that doesn't have temperature dependence) and pressure as inputs.
- We should include the key system constraints as inputs (possibly under that grouping), the ones missing I think are water recovery and maximum allowable pressure for RO and LSR

MarcusHolly:
@TimBartholomew I believe I've addressed all your comments. See the screenshots below for the current version of this GUI
image
image
image
image
image

adam-a-a:
A question I have after skimming @TimBartholomew 's comment - if we remove feed concentration and recovery from inputs, how do we allow the user to try different feed conc/recovery cases? Are we allowing the user to enter feed conc or recovery anywhere?

MarcusHolly:
@TimBartholomew Because of how the flowsheet is set up, flow rate and salinity must be config options for this flowsheet, not variables in the input tab.

adam-a-a (approved):

lbianchi-lbl (changes requested):
Just as a reminder before merging - I think this should also be changed so that the files (and the corresponding entry point in `setup.py`) are placed in `watertap.flowsheets` instead of `watertap.examples.flowsheets` following #1438.

lbianchi-lbl (approved):

PR #1388, opened 2024-05-14
Fix bubblepoint function and unit inconsistency in OLI API

### LINKED ISSUE #1369: Updating units in OLI Flash
### Description

While attempting to alter units for properties to be used for calculation or to be returned as output, the user needs to set the `oli_unit` desired, which is part of a FixedKeysDict. Additionally, there is a `pyomo_unit` value that also needs to be updated to match the new `oli_unit` assigned. When changing units, the user shouldn't need to ensure that they update pyomo units in addition to setting the oli units.

For example,
```
input_unit_set = FixedKeysDict({
        "molecularConcentration": {
            "oli_unit": "mg/L",
            "pyomo_unit": pyunits.mg / pyunits.L,
        }, ...})
```
If I attempt the following change,
`input_unit_set['molecularConcentration']['oli_unit'] = "mol/L"`,
I should then be able to `print(input_unit_set['molecularConcentration']['pyomo_unit'])` and see that "mol/L" is returned. However, as it stands now, I would need to set `input_unit_set['molecularConcentration']['pyomo_unit'] = pyunits.mol/pyunits/L` afterwards to avoid an erroneous mismatch between stated oli units and pyomo units.

Another issue is that for a given property, there is a set of acceptable units that OLI will be able to work with. Selecting an appropriate set of units for properties can result in erroneous results (or no results returned at all, without exceptions being raised). Thus, we should have predefined sets of valid entries for units per property, and if an invalid unit is supplied to a particular property, an exception should be raised.

### Expected Behavior

What I was trying to implement (unsuccessfully so far) was the ability to automatically update new pyomo units after manually providing new oli_units. For example, I start with these defaults:
```
input_unit_set = FixedKeysDict({
        "molecularConcentration": {
            "oli_unit": "mg/L",
            "pyomo_unit": pyunits.mg / pyunits.L,
        }, ...})
```
Later, if I attempt the following change,
`input_unit_set['molecularConcentration']['oli_unit'] = "mol/L"`,
I should then be able to `print(input_unit_set['molecularConcentration']['pyomo_unit'])` and see that "mol/L" is returned. 

### Steps to Reproduce

```
from pyomo.environ import units as pyunits
# used to execute OLI Cloud functions
from watertap.tools.oli_api.flash import Flash
from watertap.tools.oli_api.credentials import CredentialManager
from watertap.tools.oli_api.client import OLIApi
import os

if __name__ == "__main__":
    credential_manager = CredentialManager(
        encryption_key=os.environ["OLI_secret"],
        config_file=path_to_credentials_file,
    )
   case= {
            "Na_+": 130680,
            "Cl_-": 351140,
            "Mg_2+": 65940,
            "Ca_2+": 21100,
            "K_+": 20240,
            "SO4_2-": 137250,
        }
      
    source_water = {
        "temperature": 298.15,
        "pressure": 101325,
        "components": case,
        "units": {
            "temperature": pyunits.K,
            "pressure": pyunits.Pa,
            "molecularConcentration": pyunits.mg/pyunits.L,
        },
    }

    f = Flash( )
    print(f.input_unit_set["molecularConcentration"]["oli_unit"])
    print(f.input_unit_set["molecularConcentration"]["pyomo_unit"])
    
    # try setting new unit to oli_unit
    f.input_unit_set["molecularConcentration"]["oli_unit"] = "mg"
    # note that pyomo_unit will not match oli_unit, and user would have to manually correct
    print(f.input_unit_set["molecularConcentration"]["pyomo_unit"])
```

### Environment

- OS: win64
- Python: v3.8
- WaterTAP: v0.12

### Anything Else?

The relative path to where the FixedKeysDict is defined, which is what I was tinkering with to address this issue: `watertap\tools\oli_api\util\fixed_keys_dict.py `

adam-a-a:
## Fixes/Resolves:

- Fixes problem that prevents successfully calling bubblepoint function to get saturation pressure or temperature
- also fixes #1369 

## Summary/Motivation:
In an attempt to call the bubblepoint function, I discovered that it doesn't work as intended and the user would not be able to run the function. Along the way, I also noticed that input and/or output units provided for OLI properties have `oli_unit` and `pyomo_unit` attributes, and when one sets one or the other, there would be inconsistencies between the two (see #1369).

- add vapor phase as requirement for bubblepoint func to run
- add bubblepoint to list of post flash methods that are available
- deal with flatten_results for unanticipated vaporDiffusivityMatrix, which breaks the logic of flatten_results
- refine FixedKeysDict so that when the user sets either `pyomo_unit` or `oli_unit`, the other will automatically update via `__setitem__` and using nested FixedKeysDicts
-  add tests for new FixedKeysDict functionality
- add test for bubblepoint
- add test for valid_phases

### Legal Acknowledgement


watertap/tools/oli_api/flash.py
```diff
@@ -1089,14 +1094,29 @@ def _extract_values(data, keys):
                         "units": unit,
                         "values": values["value"],
                     }
-                else:
+                # elif "values" not in values.keys()
+                #     extracted_values =
```
adam-a-a:
```suggestion
```

watertap/tools/oli_api/util/fixed_keys_dict.py
```diff
@@ -25,7 +27,48 @@ def __setitem__(self, k, v):
             raise RuntimeError(f" Key {k} not in dictionary.")
         # also check for valid value if a list of values is given in the default dictionary
         else:
-            self.data[k] = v
+            # if user setting value in pyomo units
```
bknueven, line 30: Maybe we should have a sub-class of `FixedKeysDict` just for the units?

watertap/tools/oli_api/util/fixed_keys_dict.py
```diff
@@ -45,50 +88,55 @@ def pprint(self):
...
+    "liquidFlowInPipe": {
+        "oli_unit": "L/s",
+        "pyomo_unit": pyunits.L / pyunits.second,
+    },
+    "gasFlowInPipe": {"oli_unit": "L/s", "pyomo_unit": pyunits.L / pyunits.second},
+    "viscAbs2ndLiq": {
+        "oli_unit": "Pa-s",
+        "pyomo_unit": pyunits.Pa * pyunits.second,
+    },
+    "alkalinity": {"oli_unit": "mg HCO3/L", "pyomo_unit": pyunits.mg / pyunits.L},
```
bknueven, line 127: This example, among others, is why Paul and I gave up on an automatic approach. It seemed like you could be working forever on every edge case and still miss some.

bknueven (commented):
I am not sure about the automatic units conversion piece of this.

In the original conception, I imaged using the "pyomo_unit" attribute so the user could convert whatever input they had to units which will work with OLI by using the existing units conversion in Pyomo, and then back to those units after OLI returns the data.

It could be the answer is to restructure this so that the "pyomo_unit" and "oli_unit" are read-only attributes so as not to confuse the user.

adam-a-a:
So, I am not completely sure about the utility of pyomo_unit since we technically haven't applied it yet. Nevertheless, assuming that it might come in handy later and aiming to maintain what you guys put together, I decided to implement in this fashion, primarily because the immediate priority is to allow for the user to be able change units for inputs/outputs readily. I suppose this wasn't the original intent for FixedKeysDict, but that is what is most important to start with. These changes should make that possible. I think we can decide on totally reworking this later, but for now, i want to use it for a subsequent PR that will involve the need to toggle units of inputs and outputs.

bknueven:
Maybe "pyomo unit" should be replaced with "pint unit".

The idea of the existing mapping is if the user wants to change the units on the input / output they can do that in a standard way through Python, instead of needing to deal with the units in OLI at all.

adam-a-a:
> Maybe "pyomo unit" should be replaced with "pint unit".
> 
> The idea of the existing mapping is if the user wants to change the units on the input / output they can do that in a standard way through Python, instead of needing to deal with the units in OLI at all.

Ah, so say I want to alter the units for "molecularConcentration" to mol/L, which happens to be set to mg/L by default. How would you imagine that would go? Would it be something like the following? Or do you have a different example in mind?

```
f = Flash()
f.input_unit_set['molecularConcentration']['pint_unit'] = pyunits.mol/pyunits.L 
```

bknueven:
> Ah, so say I want to alter the units for "molecularConcentration" to mol/L, which happens to be set to mg/L by default. How would you imagine that would go? Would it be something like the following? Or do you have a different example in mind?

So, for this particular issue, I would suggest we mirror what we already do in other parts of IDAES / WaterTAP: force the user to tell us up-front if they want an molar or mass basis and select the appropriate mapping for them.

adam-a-a:
> > Ah, so say I want to alter the units for "molecularConcentration" to mol/L, which happens to be set to mg/L by default. How would you imagine that would go? Would it be something like the following? Or do you have a different example in mind?
> 
> So, for this particular issue, I would suggest we mirror what we already do in other parts of IDAES / WaterTAP: force the user to tell us up-front if they want an molar or mass basis and select the appropriate mapping for them.

That could be problematic given the way properties are passed/returned in OLI. We can think about changing this entirely. In any case, the current code resolved the issue that I identified. Now, if you overwrite either oli_unit or pyomo_unit with string or pint unit (all options are compatible), pyomo_unit would automatically match the newly set oli_unit; conversely, setting the pyomo_unit to a new value (string or pint units), would automatically update the corresponding oli_unit. 

I think we can decide to accept as an incremental change and revisit in subsequent PRs, or I can strip that segment from this PR. If I do the latter, I will likely reintroduce those mods in a subsequent PR which would illustrate an example of why the user might need to change units for certain workflows in a non-obvious way. In the longer-term, we can think about a total refactor of the related code to better match practical user needs. Most importantly, I want to have _some_ treatment of this before it ends up in 1.0.

bknueven:
> Now, if you overwrite either oli_unit or pyomo_unit with string or pint unit (all options are compatible), pyomo_unit would automatically match the newly set oli_unit; conversely, setting the pyomo_unit to a new value (string or pint units), would automatically update the corresponding oli_unit.

My issue is that this does not work for all cases because the OLI units are bespoke, e.g., https://github.com/watertap-org/watertap/pull/1388#discussion_r1602254182, and that issue aside, according to the documentation OLI will only take a fixed set of units for each parameter.

Because of this, I would suggest if the user wants different units to go to OLI we make them specify them as OLI unit / pint unit pair.

watertap/tools/oli_api/util/fixed_keys_dict.py
adam-a-a: Right- we'd need to set valid units accepted for each property, which is documented. We may catch bugs along the way (i.e., what's documented may not match reality in some cases).

adam-a-a:
> > Now, if you overwrite either oli_unit or pyomo_unit with string or pint unit (all options are compatible), pyomo_unit would automatically match the newly set oli_unit; conversely, setting the pyomo_unit to a new value (string or pint units), would automatically update the corresponding oli_unit.
> 
> My issue is that this does not work for all cases because the OLI units are bespoke, e.g., [#1388 (comment)](https://github.com/watertap-org/watertap/pull/1388#discussion_r1602254182), and that issue aside, according to the documentation OLI will only take a fixed set of units for each parameter.
> 
> Because of this, I would suggest if the user wants different units to go to OLI we make them specify them as OLI unit / pint unit pair.

Right- so, I think I have a solution for this, but that would come as a later PR if we decided to handle that way. Essentially, the same way we have `valid_phases` that OLI accepts, for each OLI property, we'd define `valid_units`. That would take more time/effort but can be handled separately. I also am not too crazy about needing to specify a pair. In any case, maybe the first thing to do is think through the use-case for `pyomo_unit` with a mini working example and see whether we want to keep/change. From that, if it becomes clear that we _don't_ need to have `pyomo_unit` for whatever reason, then we wouldn't have wasted time on it. 

But on this PR- it sounds like you'd opt to strip the changes to FixedKeysDict from here. I would reinject it in a subsequent PR for a common use case that WaterTAP users (myself included) might want to run often, and we can take it from there. How does that sound?

adam-a-a:
@OOAmusat @avdudchenko I think you two are the only ones who could add reviews to this PR at this point. Please take a look if you have a chance. If this is something you don't think you'll have time for this quarter, let me know, and I will merge without reviews (because we really should have the major fix included in this PR as part of 1.0).

watertap/tools/oli_api/client.py
```diff
@@ -188,17 +188,17 @@ def generate_dbs_file(
...
+            invalid_phases = [p for p in phases if p not in self.valid_phases]
             if invalid_phases:
                 raise RuntimeError(
                     "Failed DBS file generation. "
                     + f"Unexpected phase(s): {invalid_phases}"
                 )
             dbs_file_inputs["phases"] = phases
         else:
-            dbs_file_inputs["phases"] = ["liquid1", "solid"]
+            dbs_file_inputs["phases"] = ["liquid1", "solid", "vapor"]
```
OOAmusat, line 201: @adam-a-a Is there a reason why we do not have "liquid2" here?

adam-a-a: For now, because we haven't experimented with it enough and I didn't want to open up more cans of worms until we get the more fundamental bits of code working, I think it is safer to add liquid2 as part of the default in a later PR, which would need to include some thorough testing to see how various OLI functions may or may not be impacted.

adam-a-a:
I have tests passing for OLI. @TimBartholomew @OOAmusat @avdudchenko - just a reminder that I would like to get this merged for the June release since it fixes some issues that needed repair. Hoping to get your reviews in time; otherwise, we may just end up merging without the full 2 reviews.

OOAmusat (approved):
LGTM.

Would have liked to test the units switching myself to check for edge cases, but in the absence of OLI access, I guess this is the best we can do.

avdudchenko (approved):
This looks good to me - nothing really popped out that would cause issues.  LTMG

PR #1396, opened 2024-05-20
Flowsheet Documentation for GAC

hunterbarber:
## Fixes/Resolves:
GAC flowsheet documentation for #1219 

## Summary/Motivation:
Adding GAC flowsheet documentation

- Only includes adding .rst for the GAC flowsheet

### Legal Acknowledgement


docs/technical_reference/flowsheets/gac.rst
```diff
@@ -0,0 +1,91 @@
...
+------------
+
+The simple granular activated carbon (GAC) flowsheet can be simulated to predict the performance of a GAC system to treat residual organics. This flowsheet can
+be useful to expedite the set-up, usage, and costing of of a GAC system for its conventional water treatment applications using the constant pattern homogeneous
+surface diffusion model (CPHSDM) model.
+
+Implementation
+--------------
+
+As a single unit operation, the assumptions for the flowsheet are aligned with those detailed in the :doc:`GAC unit model documentation <../unit_models/gac>`.
```
hunterbarber: @MarcusHolly I'm going to point you to this because I see you were hard linking some internal documentation hyperlinks in the flowsheet documentation. I think this method is better because it should link to the version of the documentation the reader is using instead of having the hard link to `latest`. But honestly someone more familiar than me with .rst compiling should give a recommendation.

MarcusHolly: Thanks for the suggestion Hunter. I definitely think this is better than hard-coding. This change should probably be applied across the board tho (not just flowsheet documentation) since I believe most links are hard-coded. This probably warrants it's own PR.

hunterbarber: So I was looking through the docs to see how it is done to get this one in, and the answer is that it depends. I think all the core files like getting started and how-to's use .rst utilities instead of hard linking, but they rely on establishing a name at the top of certain documentation files (which the unit models don't have but the how-to's and others do) and then referencing the name inline, instead of here which references the file path tree to the desired link.

MarcusHolly: From a quick skim, it looks like some flowsheets, unit models, and one property model documentation are hard-coding links

MarcusHolly: Maybe the best solution it to just change the hard-coded link from `latest `to `stable `because a user might have an outdated version of the documentation, and we would still want the links to re-direct them to more up-to-date documentation. I see that some documentation does this already (`getting_started.rst`, `how_to_setup_simple_chemistrt.rst`, etc.)

hunterbarber: IMO you would always want the version to stay the same before and after clicking the link which you can't get through hard-coding, but yea this would be a tedious review of the current documentation.

docs/technical_reference/flowsheets/gac.rst
```diff
@@ -0,0 +1,91 @@
...
+Introduction
+------------
+
+The simple granular activated carbon (GAC) flowsheet can be simulated to predict the performance of a GAC system to treat residual organics. This flowsheet can
+be useful to expedite the set-up, usage, and costing of a GAC system for conventional water treatment applications using the constant pattern homogeneous
+surface diffusion model (CPHSDM) model.
+
+Implementation
+--------------
+
```
MarcusHolly, line 13: Expand this section to include a link to the property model documentation and WaterTAP costing package. Also add a brief sentence (feel free to copy from other flowsheet examples) on the levelized cost of water and the equation for it.

TLDR: Make this section more consistent with the other flowsheet documentation.

MarcusHolly: Actually, it looks like not all of the examples link to the costing package and/or describe the LCOW equation, so you don't necessarily need to add these (although it'd still be nice to have). But you should definitely link to the MCAS property package

docs/conf.py
```diff
@@ -96,6 +96,10 @@
 #
 html_favicon = "_static/favicon.ico"
 
+# intersphinx mapping to idaes
+intersphinx_mapping = {
```
hunterbarber, line 100: @TimBartholomew @ksbeattie @lbianchi-lbl Are there any downsides to adding this? See [https://docs.readthedocs.io/en/stable/guides/intersphinx.html](https://docs.readthedocs.io/en/stable/guides/intersphinx.html)

docs/technical_reference/flowsheets/gac.rst
```diff
@@ -0,0 +1,103 @@
...
+Property models:
+    * :doc:`/technical_reference/property_models/mc_aq_sol`
+Unit models:
+    * :doc:`/technical_reference/unit_models/gac`
+    * :doc:`idaes:reference_guides/model_libraries/generic/unit_models/feed`
+    * :doc:`idaes:reference_guides/model_libraries/generic/unit_models/product`
+Costing models:
+    * :doc:`/technical_reference/costing/watertap_costing`
+    * :doc:`/technical_reference/costing/gac`
```
MarcusHolly: For consistency, I would use the same verbiage as other flowsheet documentation: "Documentation for each of the unit models can be found below"

MarcusHolly (commented):
One more comment

MarcusHolly (approved):
LGTM

docs/conf.py
hunterbarber: Compare [the ability to link with this change shown in this PR](https://github.com/watertap-org/watertap/pull/1396/files#diff-1009f2493fb3cb501cbd2ac5af69180ca4ba15b279a71519ed0fbe636de9dd14R25) to [these linking methods currently in main](https://github.com/watertap-org/watertap/blob/main/docs/technical_reference/flowsheets/electrodialysis_1stack.rst?plain=1#L50)

adam-a-a: From the looks of it, intersphinx mapping seems like the better way to go--even though we have other ways of dealing with these links without hard-coding to a version from my vague recollection. Can't speak to the downsides though. @lbianchi-lbl @dangunter  - any thoughts?

savannahsakhai (approved):
LGTM

docs/conf.py
lbianchi-lbl: Sorry for the late reply, I missed the notification(s). I think InterSphinx mappings is a good approach. I've tested that the GAC links to the IDAES docs work as expected on this PR's ReadTheDocs build.

PR #1399, opened 2024-05-20
Use nbviewer for statically browsing tutorials

lbianchi-lbl:
## Fixes/Resolves:

- Notebooks not rendering using GitHub
- This might or might not be a temporary problem on the side of GitHub's infrastructure

image

## Summary/Motivation:

- nbviewer should provide a more reliable and dedicated interface for users for browsing the tutorials: https://nbviewer.org/github/watertap-org/watertap/blob/main/tutorials/00-index.ipynb

- Use nbviewer instead of GitHub for browsing tutorials
- Add some subsections to improve clarity

### Legal Acknowledgement

### DIFF
--- a/docs/tutorials/index.rst
+++ b/docs/tutorials/index.rst
@@ -1,8 +1,16 @@
 Tutorials
 =========
 
+Browsing the tutorials
+----------------------
+
 The tutorials included with WaterTAP are available in the "tutorials" directory
-on the WaterTAP GitHub Homepage: `<https://github.com/watertap-org/watertap/tree/main/tutorials/00-index.ipynb>`_.
+of the WaterTAP GitHub repository: `<https://github.com/watertap-org/watertap/tree/main/tutorials/>`_.
+
+To view a static version of the notebooks online, you can use the following link: `<https://nbviewer.org/github/watertap-org/watertap/blob/main/tutorials/00-index.ipynb>`_.
+
+Running the notebooks
+---------------------
 
 .. note::
     For important details on running tutorial notebooks, refer to :ref:`this guide<notebooks>`.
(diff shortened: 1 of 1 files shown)


MarcusHolly (approved):
LGTM

docs/tutorials/index.rst
```diff
@@ -1,8 +1,16 @@
 Tutorials
 =========
 
+Browsing the tutorials
+----------------------
+
 The tutorials included with WaterTAP are available in the "tutorials" directory
-on the WaterTAP GitHub Homepage: `<https://github.com/watertap-org/watertap/tree/main/tutorials/00-index.ipynb>`_.
+of the WaterTAP GitHub repository: `<https://github.com/watertap-org/watertap/tree/main/tutorials/>`_.
```
adam-a-a, line 8: Can this link to the 00-index notebook, or will that perpetuate the issue? I thought it'd be nice to be taken to the landing page which describes all available tutorials with links.

adam-a-a (commented):
A quick q on whether we can link directly to the landing page

adam-a-a (approved):
LGTM - users can just use the nbviewer link to get to the landing page. Thanks for the quick turnover!

docs/tutorials/index.rst
lbianchi-lbl: That's a good question, I don't know if I have a definite opinion on it. I guess that linking to the directory should more or less always work (since it just uses the default GitHub file browser viewer), but at the same time you'd have to rely on the filenames instead of the the extended description in the intro/index. :man_shrugging: One thing to consider is also that the relative links in the intro/index won't work in the GitHub view (at least I think that's the case), while they work just fine in nbviewer. So maybe one possibility could be to just point to nbviewer as the default static browser, and leave the GitHub browser as a fallback.

PR #1372, opened 2024-05-01
Add Conservation Checks to Test Harness Examples

MarcusHolly:
## Fixes/Resolves:

Some of the test harness examples that have previously been merged into the repo did not account for the conservation tests. This PR adds those missing tests.

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #1401, opened 2024-05-20
Apply UnitTestHarness to IonExchange0D

kurbansitterley:
## Fixes/Resolves:

part of #1302 

- apply unit test harness to IX model

### Legal Acknowledgement


watertap/unit_models/tests/test_ion_exchange_0D.py
```diff
@@ -10,1070 +10,287 @@
...
-        m = IX_lang
-        results = solver.solve(m, tee=True)
-        assert_units_consistent(m)
-        # Check for optimal solution
-        assert_optimal_termination(results)
-
-    @pytest.mark.requires_idaes_solver
-    @pytest.mark.component
-    def test_mass_balance(self, IX_lang):
-        m = IX_lang
```
MarcusHolly, line 250: We should try to keep at least one of these conservation checks in the testing. See #1372 for examples

kurbansitterley: Yes agreed, but wasn't sure how to do so with the test harness without just keeping the existing test, thanks for the direction

MarcusHolly: Eventually conservation checks will work their way into the test harness, but it's not a high priority atm.

hunterbarber (approved):
Coverage and implementation seem fine. I do see you're removing asserting the resulting costing variables because of how the unit model test harness works. The PR for the costing test harness is planned for after the 1.0 overhaul, so you may want to keep the `_cost_results` assertions in a test after the normal harness tests just so they're easy to grab in the future.

MarcusHolly (approved):
LGTM

kurbansitterley:
> Coverage and implementation seem fine. I do see you're removing asserting the resulting costing variables because of how the unit model test harness works. The PR for the costing test harness is planned for after the 1.0 overhaul, so you may want to keep the `_cost_results` assertions in a test after the normal harness tests just so they're easy to grab in the future.

Great idea, just added the costing tests back.

PR #1394, opened 2024-05-20
Update Costing Doc Automation

MarcusHolly:
## Summary/Motivation:

In the unit model costing documentation there was a column named "Default Value" under the Costing Method Variables section. Since this column does not make sense for many of these variables, this column has been changed to "Index". This PR also changes the variable and parameter names to the monospacing font to distinguish them from regular text.

## Changes proposed in this PR:
- "Default Value" column changed to "Index" in unit model costing documentation template
- Variable and Parameter names changed to monospacing font in unit model costing documentation template
- Updates existing unit model costing documentation to use monospacing font for Vars and Params

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #1402, opened 2024-05-20
IonExchange0D costing documentation into template dedicated file

kurbansitterley:
## Fixes/Resolves:

Move existing IX costing documentation into template dedicated file for #1360 

- Move IX costing documentation from unit model to costing 

### Legal Acknowledgement


docs/technical_reference/costing/ion_exchange.rst
```diff
@@ -6,10 +6,29 @@ Costing Method Parameters
 
 The following parameters are constructed for the unit on the FlowsheetCostingBlock (e.g., `m.fs.costing.ion_exchange`) when applying the `cost_ion_exchange` costing method in the ``watertap_costing_package``:
 
-.. csv-table::
-   :header: "Description", "Symbol", "Parameter Name", "Default Value", "Units"
 
-   "description", ":math:`Symbol_{example}`", "parameter_name", "1", ":math:`\text{dimensionless}`"
+.. csv-table::
+   :header: "Description", "Symbol", "Parameter Name", "Default Value", "Units", "Notes"
```
MarcusHolly, line 11: If possible, we should replace instances of $ with USD_YYYY

docs/technical_reference/costing/ion_exchange.rst
```diff
@@ -19,41 +38,134 @@ The following variables are constructed on the unit block (e.g., m.fs.unit.costi
 .. csv-table::
    :header: "Description", "Symbol", "Variable Name", "Default Value", "Units"
 
-   "description", ":math:`Symbol_{example}`", "variable_name", "1", ":math:`\text{dimensionless}`"
+   "Density of regenerant solution", ":math:`\rho_{regen}`", "``regen_soln_dens``", "1000", ":math:`\text{kg}\text{m}^{3}`"
```
MarcusHolly: In #1394, we are replacing the "Default Value" column with "Index", so you can probably just adjust this now rather than waiting for the conflict to emerge.

docs/technical_reference/costing/ion_exchange.rst
```diff
@@ -19,41 +38,134 @@ The following variables are constructed on the unit block (e.g., m.fs.unit.costi
...
+   "Regenerant dose required for regeneration per volume of resin [kg regenerant/m3 resin]", ":math:`D_{regen}`", "``regen_dose``", "300", ":math:`\text{kg}\text{m}^{3}`"
+   "Capital cost for one vessel", ":math:`C_{col}`", "``capital_cost_vessel``", "10000", ":math:`\text{\$/}`"
+   "Capital cost for resin for one vessel", ":math:`C_{resin}`", "``capital_cost_resin``", "10000", ":math:`\text{\$/}`"
+   "Capital cost for regeneration solution tank", ":math:`C_{regen}`", "``capital_cost_regen_tank``", "10000", ":math:`\text{\$/}`"
+   "Capital cost for backwash + rinse solution tank", ":math:`C_{bw}`", "``capital_cost_backwash_tank``", "10000", ":math:`\text{\$/}`"
```
MarcusHolly: Looks like the units here have an extra "/" because they're appearing as "$/" rather than "$"

docs/technical_reference/costing/ion_exchange.rst
```diff
@@ -19,41 +38,134 @@ The following variables are constructed on the unit block (e.g., m.fs.unit.costi
...
+   "Operating cost for hazardous waste disposal", ":math:`D_{regen}`", "``operating_cost_hazardous``", "10000", ":math:`\text{\$/}\text{year}`"
+   "Regeneration solution flow", ":math:`\dot{v}_{regen}`", "``flow_mass_regen_soln``", "1", ":math:`\text{kg}\text{year}`"
```
MarcusHolly:
```suggestion
   "Regeneration solution flow", ":math:`\dot{v}_{regen}`", "``flow_mass_regen_soln``", "1", ":math:`\text{kg}\text{/year}`"
```
Should be kg/yr rather than kg-yr

docs/technical_reference/costing/ion_exchange.rst
```diff
@@ -6,10 +6,29 @@ Costing Method Parameters
...
 The following parameters are constructed for the unit on the FlowsheetCostingBlock (e.g., `m.fs.costing.ion_exchange`) when applying the `cost_ion_exchange` costing method in the ``watertap_costing_package``:
 
-.. csv-table::
-   :header: "Description", "Symbol", "Parameter Name", "Default Value", "Units"
 
-   "description", ":math:`Symbol_{example}`", "parameter_name", "1", ":math:`\text{dimensionless}`"
+.. csv-table::
+   :header: "Description", "Symbol", "Parameter Name", "Default Value", "Units", "Notes"
+
+   "Anion exchange resin cost", ":math:`c_{res}`", "``anion_exchange_resin_cost``", "205", ":math:`\text{\$/}\text{ft}^{3}`", "Assumes strong base polystyrenic gel-type Type II. From EPA-WBS cost model."
```
MarcusHolly: I like how you formatted the parameter names in this table. Perhaps I should update the automation script so that the templates automatically use this format. Thoughts @adam-a-a @hunterbarber or others? To be consistent, we would need to go back and update all the existing costing documentation tho...

docs/technical_reference/costing/ion_exchange.rst
kurbansitterley: That's a good idea, I thought providing the default value was odd for some of these.

docs/technical_reference/costing/ion_exchange.rst
kurbansitterley: I also just noticed I have a "Notes" column here since I had it on the previous doc. Should I remove or consolidate with the Description column? It provides some more detail on the source of the costing data and/or how it can be used in the model

kurbansitterley: > If possible, we should replace instances of $ with USD_YYYY

with the underscore?

docs/technical_reference/costing/ion_exchange.rst
kurbansitterley: Yes I like using the monospaced font for anything that I could find in a Python file/object, just hits the eyes better imo. I am happy to change it if there is consensus either way.

docs/technical_reference/costing/ion_exchange.rst
adam-a-a: Preferably with the underscore

MarcusHolly: I think it's fine to keep the Notes section as it is.

docs/technical_reference/costing/ion_exchange.rst
```diff
@@ -6,54 +6,172 @@ Costing Method Parameters
...
 The following parameters are constructed for the unit on the FlowsheetCostingBlock (e.g., `m.fs.costing.ion_exchange`) when applying the `cost_ion_exchange` costing method in the ``watertap_costing_package``:
 
-.. csv-table::
-   :header: "Description", "Symbol", "Parameter Name", "Default Value", "Units"
 
-   "description", ":math:`Symbol_{example}`", "parameter_name", "1", ":math:`\text{dimensionless}`"
+.. csv-table::
+   :header: "Description", "Symbol", "Parameter Name", "Default Value", "Units", "Notes"
+
+   "Anion exchange resin cost", ":math:`c_{res}`", "``anion_exchange_resin_cost``", "205", ":math:`\text{USD2020 }\text{ft}^{-3}`", "Assumes strong base polystyrenic gel-type Type II. From EPA-WBS cost model."
```
MarcusHolly: Sorry to be so nitpicky about the units, but they should look like `\text{USD}_{2012}\text{/ft}^3`. This makes it a bit easier to read because the year is in the subscript and a fraction is used rather than the negative exponent.

docs/technical_reference/costing/ion_exchange.rst
MarcusHolly: I don't think anyone would be opposed to this idea, so I can address it in #1394

docs/technical_reference/costing/ion_exchange.rst
kurbansitterley: Ah was lost on me that this was to make it a subscript so that makes more sense.

MarcusHolly: Not sure if you're handling this piecewise or not, but the negative exponents should also be changed to fractions. And there are also still a few "$" in the "Costing Method Variables" that should be changed to USD_YYYY

kurbansitterley: The "$" in the "Costing Method Variables" would be converted to the ``base_currency`` that is assigned at the costing package, so there isn't really a year to assign... I see other unit model docs do have USD_XXXX, so I guess they calculate in USD_XXXX and then convert to ``base_currency`` elsewhere, but that isn't strictly true for the IX implementation (e.g., displaying the value for ``capital_cost_resin`` would be in units of ``base_currency``, not USD_2020).  I suppose I could do something like USD_base? If we want the USD units in the docs to reflect what would be the displayed result for a variable, I would have to change the IX costing model. I am not opposed to this but just want to be clear.

I had converted the fractions to negative exponents to be consistent with other docs... but looking around now I guess there isn't real consistency in that regard for our docs so I can convert it back.

MarcusHolly: Ah, the "$" is probably fine then. Or maybe we should just put USD instead. As for the fractions vs negative exponents, I _think_ all of the costing documentation uses fractions, so we can at least keep that consistent, if nothing else.

MarcusHolly (approved):
LGTM

docs/technical_reference/costing/ion_exchange.rst
```diff
@@ -6,54 +6,172 @@ Costing Method Parameters
...
+   "Unit cost for hazardous waste resin disposal", ":math:`f_{haz,res}`", "``hazardous_resin_disposal``", "347.10", ":math:`\text{USD}_{2020}\text{/ton}`", "From EPA-WBS cost model."
+   "Unit cost for hazardous waste regeneration solution disposal", ":math:`f_{haz,regen}`", "``hazardous_regen_disposal``", "3.64", ":math:`\text{USD}_{2020}\text{/gal}`", "From EPA-WBS cost model."
+   "Number of cycles the regenerant can be reused before disposal", ":math:`f_{recycle}`", "``regen_recycle``", "1", ":math:`\text{dimensionless}`", "Can optionally be set by the user to investigate more efficient regen regimes."
+   "Costing factor to account for total installed cost installation of equipment", ":math:`f_{TIC}`", "``total_installed_cost_factor``", "1.65", ":math:`\text{dimensionless}`", "Costing factor to account for total installed cost of equipment"
+   "Unit cost of NaCl", ":math:`c_{regen}`", "``costing.nacl``", "0.09", ":math:`\text{USD}_{2020}\text{/kg}`", "Assumes solid NaCl. From CatCost v 1.0.4"
```
hunterbarber: Need to correct the symbols in these couple of lines

hunterbarber: Also, aren't all of these variables constructed on `costing.variable_name`? I don't think you would need to be explicit for these few.

hunterbarber (changes requested):
Only one thing to double-check

docs/technical_reference/costing/ion_exchange.rst
kurbansitterley: > Also, aren't all of these variables constructed on `costing.variable_name`? I don't think you would need to be explicit for these few.

Actually these are chemical costs so would be on ``m.fs.costing`` whereas the rest of the parameters in this table would be on ``m.fs.costing.ion_exchange``. So, when applying the ``cost_ion_exchange`` method, these are created, but the end up somewhere else but are part of the costing process for IX. I think that is why I included them here and made that distinction when I first created this table.

@MarcusHolly I am wondering if you have a preference for how to handle this. It is a bit nuanced and for a reader who is interested in default chemical costs, they might want to know this info but I am happy to remove them. Quickly looking at the docs for the costing package, there isn't anything about costing chemical flows, but that might be a bit too IDAES-y for WaterTAP docs.

MarcusHolly: Generally, I think it's better to overshare than undershare in the documentation, so I would keep these in.

hunterbarber: I think you probably should append a table above it and just revise the introduction for it to
```
The following parameters are constructed for the unit on the **WaterTAPCostingBlock
(e.g., m.fs.costing)** when applying the cost_ion_exchange costing method in the 
watertap_costing_package:
```
Then move those parameters to the appropriate table

hunterbarber (approved):
Updates since last review look good.

Issue #1407, opened 2024-05-28
Add membrane area to inputs page of OARO GUI


adam-a-a:
Membrane area should be available to specify as a fixed or free variable in the OARO GUI. Even though we typically like to optimize membrane area per stage, we should still allow the user the option to fix membrane area as desired. 

The default option for membrane area would be to have it set as a free variable. @luohezhiming does this sound reasonable to you?

luohezhiming:
@adam-a-a Since the OARO flowhseet is an optimization flowsheet and the membrane area is a decision vairbale, it is always free. And we fixed `recovery_mass_phase_comp` for each OARO units for ease of initialization instead of  membrane_area. So, the membrane area is always a free varible, even export as an input variable, it is still free.
If we choose to fix membrane area at the begining, then the system might be hard to pass initialization. Anyhow, the major problem is this is an optimization problem, and we choose membrane area as a decision variable, so user cannot fix it and it is meaningless to fix it at the begining

adam-a-a:
@luohezhiming By default, we have rendered it a free variable--yes. But that is not a requirement--just a preferred approach to better ensure convergence to a solution. We have other "difficult" models in the GUI that may fail to solve when the user fiddles with the inputs in a certain way, but we still allow the flexibility to do so. 

Thus, even though the user could get an error back instead of a successful solve when changing this, we should include area in the inputs section, labeled as a free var be default, which lines up with what we both were saying, I think. Then the user could optionally change to a fixed variable after the initial solve or even do a sweep on area, even though it would be risky.

luohezhiming:
@adam-a-a The GUI fully depends on the flowsheet, even though we export the membrane area as an input variable, I believe it will only be shown as a free variable (with default valuess), we cannot make it as fixed or do sweep analysis according to my previous experience. Since the GUI is just a UI interface, it cannot implement some analysis beyond the capability of original flowsheet. So unless we choose to solve another optimization problem where membrane_area is fixed, otherwise it will only be shown as a free variable I guess, but I'll confirm by another attempt

adam-a-a:
I understand how the GUI works and have implemented my suggestion for another case, which is why I am 99% sure that what I am suggesting is feasible and should be done. The first step is setting `is_input=True` for stage area. In fact, all the decision vars should be exposed in this way--at least the inlet pressures and areas. @MarcusHolly - any thoughts on this since you have worked with exporting to GUI more than most of us?

adam-a-a:
[...]

PR #1411, opened 2024-05-28
Add more input options for OARO GUI

luohezhiming:
## Fixes/Resolves:

addresses #1407 

## Summary/Motivation:
Add membrane area and inlet operating pressure as two free inlet variables to GUI

### Legal Acknowledgement


MarcusHolly (approved):
LGTM

adam-a-a (commented):
LGTM although I wonder if we should expose the pump discharge pressure rather than OARO/RO feed inlet pressure (unsure if it makes a difference with regard to model stability based on the way the flowsheet was set up). Can you (or did you) try running a simulation in the GUI where you fix the pressures an areas?

luohezhiming:
> LGTM although I wonder if we should expose the pump discharge pressure rather than OARO/RO feed inlet pressure (unsure if it makes a difference with regard to model stability based on the way the flowsheet was set up). Can you (or did you) try running a simulation in the GUI where you fix the pressures an areas?

I make some tests, the flowsheet can run with pressure and area fixed. There is no difference between fixing pump discharge pressure and OARO inlet pressure.

adam-a-a (approved):

PR #1412, opened 2024-05-28
improved error message.

k1nshuk:
## Fixes/Resolves:

Provides better instructions for debugging scaling problems, e.g., 0/0 calculations to identify which function the error is coming from.

## Summary/Motivation:
Making the debugging process easier

- Better error message. See file for changes

### Legal Acknowledgement


bknueven (approved):

watertap/core/plugins/solvers.py
```diff
@@ -158,7 +158,9 @@ def _scale_constraints(self, blk):
                     self._cleanup()
                     raise RuntimeError(
                         "Error in AMPL evaluation.\n"
-                        "Run ipopt with halt_on_ampl_error=yes and symbolic_solver_labels=True to see the affected function."
+                        "Re-run ipopt with:\n"
+                        '1. solver options = {"halt_on_ampl_error" : "yes", "nlp_scaling_method" : "gradient-based"\n'
```
adam-a-a, line 162: I am forgetting- should `nlp_scaling_method` be what you have here or `user-scaling`? What happens if you don't specify gradient -based scaling?

bknueven: This option ensures watertap's own scaling gets skipped and the model gets handed directly to Ipopt. This is what we want to see the problematic constraint / expression.

adam-a-a: I just thought I remembered this working regardless of that option so maybe I'm not recalling correctly.

bknueven: It certainly does in IDAES or if you just use `SolverFactory("ipopt")`.

adam-a-a (approved):

PR #1416, opened 2024-05-30
Adding Conservation Tests to the Test Harness

### LINKED ISSUE #1387: Add Conservation Check to Test Harness
### Description

Have the user specify an input and output expression and the test harness will check that the two expressions are equal. If the user does not specify these conservation expressions, an error should be raised. An option could be added to ignore this check in cases where it may not be applicable. Alternatively, the user could just set an expression equal to itself to bypass this test.

### Motivation

Makes the test harness more inclusive and robust.

### Possible Implementation

{"check_1": {"input expression": input_expression,
             "output expression": output_expression},
 "check_2": {...}
}

### Additional Context

_No response_

MarcusHolly:
## Fixes/Resolves:

Resolves Issue #1387; replaces #1372

- Embeds conservation check into the test harness
- Updates existing tests that use the test harness
- Updates How-TO documentation for test harness

### Legal Acknowledgement


MarcusHolly:
I'm curious what people think about this implementation of adding a conservation check into the unit model test harness and if there are any ways to make it more robust. I've updated the anaerobic digester test to showcase what the current implementation looks like, but this PR will eventually have to update all of the existing test harness use cases so that the tests pass.

watertap/unit_models/tests/test_anaerobic_digester.py
```diff
@@ -206,4 +206,27 @@ def configure(self):
...
+                * (m.fs.unit.inlet.temperature[0] - m.fs.props.temperature_ref)
+            )
+            - (
+                m.fs.unit.liquid_outlet.flow_vol[0]
+                * m.fs.props.dens_mass
+                * m.fs.props.cp_mass
+                * (m.fs.unit.liquid_outlet.temperature[0] - m.fs.props.temperature_ref)
+            ),
+            "Outlet expression 2": -1 * m.fs.unit.liquid_phase.enthalpy_transfer[0],
+        }
```
TimBartholomew: I think it would be better to nest the related expressions, like the following:
```
conservation_check = {
   "check 1": [expression_in, expression_out],
   "check 2": [expression_in, expression_out],
}
```
Or if we think it would be good to be explicit with a key for "in" and "out":
```
conservation_check = {
   "check 1": {"in": expression_in, "out": expression_out},
   "check 2": {"in": expression_in, "out": expression_out},
}
```

This way we don't have to require the user to provide the exact string like "Outlet expression X". The check can just iterate through the highest key (in my example its "check 1" and "check 2", but those names could be anything), and the first example will just make sure that the dictionary results in a 2 item list and know those things should be checked to be approximately equal, or the second example requires the "in" and "out" hard coded but that's not too bad.

MarcusHolly: My one concern with this is that it might come across looking a bit too clunky. The current implementation is also a bit clunky imo, but I think it looks better than the alternatives. Maybe others disagree...

Current implementation:
image

Tim's 1st suggestion:
image


Tim's 2nd suggestion:
image

bknueven: Tim's 2nd suggestion seems the least error prone to me and is the most explicit.

Just to throw in another curveball, I don't see any reason we couldn't just use Pyomo's expression system itself. e.g.
```
"Check 1": m.fs.unit.inlet.flow_vol[0] == m.fs.unit.outlet.flow_vol[0],
```

This is a bit more complicated on the back end, but not overly so.

MarcusHolly: I've implemented Tim's 2nd suggestion. I will address the failures by updating the remaining uses of TestHarness later this week.

TimBartholomew (approved):
LGTM!

agarciadiego (approved):
LGTM

adam-a-a (approved):

PR #1417, opened 2024-05-30
Delete Simple IEDO GUI Flowsheets

MarcusHolly:
- Deletes UI flowsheets from the https://github.com/watertap-org/watertap/tree/main/watertap/examples/flowsheets/case_studies/wastewater_resource_recovery directory except for dye desal
- Remove flowsheets from UI flowsheet list in `setup.py`
### Legal Acknowledgement


MarcusHolly:
Looks much cleaner now:
image

TimBartholomew (approved):

TimBartholomew:
@lbianchi-lbl codecov isn't making sense here either. Patch passes but project decreases by ~3%, but the PR only removes lines of code and doesn't touch any tests.

lbianchi-lbl:
> @lbianchi-lbl codecov isn't making sense here either. Patch passes but project decreases by ~3%, but the PR only removes lines of code and doesn't touch any tests.

Yeah, it looks like it might have been a transitory thing since the problem seems to have gone away on the other affected PRs after updating from `main`.

watertap/examples/flowsheets/case_studies/wastewater_resource_recovery/metab/metab.py
```diff
@@ -480,42 +480,6 @@ def rule_LCOCR_comp(b, c):
     m.fs.costing.LCOCR_comp = Expression(m.fs.costing.LC_comp, rule=rule_LCOCR_comp)
 
 
-def adjust_default_parameters(m):
-    m.fs.metab_hydrogen.hydraulic_retention_time.fix(6)  # default - 12 hours, 0.5x
```
MarcusHolly, line 484: @TimBartholomew Is it fine to delete this function. Codecov was failing because of this, and it was apparently only used in the `metab_ui.py` file, which has been deleted.

MarcusHolly:
@dangunter @MichaelPesce Tagging the two of you here since I think you're the most familiar with the `fsapi.py`. Do you know why code coverage is decreasing in this PR? It just deletes some GUI flowsheets and I had to touch one check in `test_fsapi.py` since it was referencing a now-deleted flowsheet. I'm not sure if this is one of the scenarios where we should override the loss in coverage or if we need to add more checks in `test_fsapi.py`

MichaelPesce:
@MarcusHolly As far as I can tell, the change made in `test_fsapi.py` shouldn't decrease coverage, as it shouldn't matter which flowsheet export is used for that test. I remember there was discussion of some weird behavior with codecov last week - not sure if that's been resolved or not

lbianchi-lbl:
> @dangunter @MichaelPesce Tagging the two of you here since I think you're the most familiar with the `fsapi.py`. Do you know why code coverage is decreasing in this PR? It just deletes some GUI flowsheets and I had to touch one check in `test_fsapi.py` since it was referencing a now-deleted flowsheet. I'm not sure if this is one of the scenarios where we should override the loss in coverage or if we need to add more checks in `test_fsapi.py`

@MarcusHolly I agree with @MichaelPesce, removing code generally shouldn't result in a significant change in coverage (assuming the code being removed had a similar coverage to the project as a whole). The detailed summary on the Codecov webpage seems to show no significant changes, so the failing check is likely to be a fluke. Let's see if updating the head branch as you've just done resolves this.

MarcusHolly:
@lbianchi-lbl It's still failing the codecov check. Is it acceptable to just bypass it (i.e. ignore the failure)?

lbianchi-lbl:
> @lbianchi-lbl It's still failing the codecov check. Is it acceptable to just bypass it (i.e. ignore the failure)?

@MarcusHolly I'm looking into this more carefully, and it seems that one of the flowsheets, namely `metab`, shows up as modified rather than deleted: 

image

Since the corresponding entry point in `setup.py` has been deleted, this might explain the discrepancy in coverage (there are still executable code lines left, but they are not run by any test since the UI interface has been removed).

Is there any obstacle in deleting that file altogether? If not, could you try to do that and see if that resolves the coverage check failure?

MarcusHolly:
> > @lbianchi-lbl It's still failing the codecov check. Is it acceptable to just bypass it (i.e. ignore the failure)?
> 
> @MarcusHolly I'm looking into this more carefully, and it seems that one of the flowsheets, namely `metab`, shows up as modified rather than deleted:
> 
> image
> 
> Since the corresponding entry point in `setup.py` has been deleted, this might explain the discrepancy in coverage (there are still executable code lines left, but they are not run by any test since the UI interface has been removed).
> 
> Is there any obstacle in deleting that file altogether? If not, could you try to do that and see if that resolves the coverage check failure?

So I've deleted the METAB UI files, but I don't think we've reached a verdict yet on whether or not we should be deleting these flowsheets just yet (we discussed in the dev call last week and it seemed like we were leaning towards keeping these flowsheets here rather than archiving them elsewhere). The only reason this METAB flowsheet was modified is because that one function was only being used in the METAB ui flowsheet (and METAB ui flowsheet test file). None of the other flowsheets have a function like this.

EDIT: I could temporarily delete this file "for science" and see if the test passes, but we'd have to undo it afterwards.

lbianchi-lbl:
@MarcusHolly I think I understand now, thanks for the clarification. I think I mistakenly assumed that both the flowsheet and the UI code were being removed by this PR, but as you point out, that's not actually the case. In that case, I think we can ignore the Codecov check failure.

adam-a-a (approved):

luohezhiming (approved):
Good to go if Tim agree to delete the functions for METAB analysis

MarcusHolly:
@bknueven @lbianchi-lbl (I'm assuming one of you two will merge this) This PR can be merged, but we'll have to bypass the codecov check as mentioned here: https://github.com/watertap-org/watertap/pull/1417#issuecomment-2150783659

PR #1420, opened 2024-05-30
IX model small fix for anions in demo

kurbansitterley:
## Fixes/Resolves:

While writing the docs for the IX flowsheet, the anion demos would not solve easily (we were not testing them). This is a quick fix for easier solvability of the anions in the IX demo flowsheet.

## Summary/Motivation:

Ensure the IX demo will work for the anions.

- use ``abs(value(charge))`` to calculate ``mass_removed`` for IX langmuir model
- add test to demo test file for SO4_2- to cover bases

### Legal Acknowledgement


kurbansitterley:
This issue was likely making the Langmuir IX model unsolvable for anions since it was calculating a negative value for ``mass_removed`` due to the negative charge. This was not caught because we weren't testing for that configuration. Lesson learned.

watertap/unit_models/ion_exchange_0D.py
```diff
@@ -1053,7 +1054,7 @@ def eq_partition_ratio(b):
                 self.target_ion_set, doc="Removed total mass of ion in equivalents"
             )
             def eq_mass_removed(b, j):
-                charge = prop_in.charge_comp[j]
+                charge = abs(value(prop_in.charge_comp[j]))
```
bknueven: Is `prop_in.charge_comp[j]` a parameter which should never change?

adam-a-a: don't know if this would be overkill, but would it be safer to use `smooth_abs`? One potential scenario could be that the charge for a particular component is unknown, and we set it as a var to estimate. I think in that case, you'd want something like `smooth_abs`. I don't know how likely that scenario is, but just thinking out loud on whether we think it's fine to just use `abs` here for the long run.

bknueven: Maybe the compromise is to use `pyo.environ.abs`.

If this value is fixed but the user updates it for some reason you'll still get the correct value put to the solver.

andrewlee94: `smooth_abs` would have no effect in this case as written as you are using `value` inside the `abs`; i.e. it is effectively a constant and not a variable. However, that does lead me to ask if this is the intended behaviour, as with the `value` in there this is now just a constant and thus will not change if the value of the variable is changed later.

kurbansitterley: > don't know if this would be overkill, but would it be safer to use `smooth_abs`? One potential scenario could be that the charge for a particular component is unknown, and we set it as a var to estimate. I think in that case, you'd want something like `smooth_abs`. I don't know how likely that scenario is, but just thinking out loud on whether we think it's fine to just use `abs` here for the long run.

A ``target_ion`` with an unknown charge _should_ raise a ``ConfigurationError`` for this model. I also think if we were in a situation where we were trying to predict the charge of an unknown component, we would likely need a different model.

kurbansitterley: > Maybe the compromise is to use `pyo.environ.abs`.
> 
> If this value is fixed but the user updates it for some reason you'll still get the correct value put to the solver.

I tried this initially (i.e., ``from pyomo.environ import abs``) and it resulted in an ``ImportError``. Could it exist elsewhere in pyomo?

kurbansitterley: > `smooth_abs` would have no effect in this case as written as you are using `value` inside the `abs`; i.e. it is effectively a constant and not a variable. However, that does lead me to ask if this is the intended behaviour, as with the `value` in there this is now just a constant and thus will not change if the value of the variable is changed later.

Yes this is the intended behavior - the value of ``charge_comp`` (a ``Param``) must be either positive or negative to use the model, and would be set as part of the configuration of the property package (it is checked at the beginning of the build for this model) . It is assumed it is a fundamental property of the solute and wouldn't change later.

kurbansitterley: Correct, should never change and is set as part of the property package configuration.

bknueven:
```suggestion
                charge = abs(prop_in.charge_comp[j])
```
This should work just fine as it is a parameter. Pyomo will create an expression for it such that even if the parameter value changes you will get the correct result.

If `charge_comp` has units you might need to remove and re-apply them after outside the `abs` function.

bknueven (approved):

watertap/unit_models/ion_exchange_0D.py
adam-a-a: True, it is currently a Param. I guess I was just thinking of a use case of my own, where I might need to set charge_comp as a Var for parameter estimation since, in that use case, the charge is unknown (just know it should be negative). But- right now charge_comp is just a Param so shouldn’t matter at this point.

adam-a-a (approved):

Issue #1248, opened 2023-12-12
Clean up WaterTAP


TimBartholomew:
WaterTAP is past due for some general cleaning of unused/obsolete code and reorganizing some of the repo. For example, recently we decided to remove the chemistry tests in #1229. We should use this issue to identify other clean up opportunities.

Prioritized clean up:
- [x] Remove watertap/unit_models/nanofiltration_0D, this was one of the first models added to WaterTAP but there are many known issues. The biggest issue is that it is only applicable for binary mixtures (one salt and water) but the primary reason to use nanofiltration is for selective removal of divalents and monovalents. Additionally, we have two other NF models that are more applicable. (#1334)
- [x] Remove watertap/flowsheets/full_treatment_train, this was an important flowsheet for a ProteusLib (previous name for WaterTAP) milestone in Sept 2021 but it is finicky (oftentimes tests need to be changed) and it is convoluted. Additionally, many of the assumptions for the water chemistry are troubling, we have much better non-public work for this same type of analysis. (#1335)
- [x] Move the electrolyte database to a new repository, we've had various discussions around this, currently I believe the consensus is that its maintenance is not worth it when it is used so little @lbianchi-lbl
- [x] Remove watertap/unit_models/selective_oil_permeation and watertap/property_models/selective_oil_permeation_prop_pack, these are simple models that were made for a collaboration that ended without results (#1336)
- [x] Rework watertap/examples/custom_model_demo, this should be converted into a jupyter notebook and a tutorial #1390 @MarcusHolly 
- [x] Move property models that are specific to unit models (e.g. cryst_prop_pack, coagulation_prop_pack) into a new folder for these limited property packages (maybe watertap/property_models/unit_specific) @savannahsakhai #1429 
- [x] Remove seawater_ion_generic and seawater_ion_prop_pack, these are likely obsolete to MCAS (#1347)
- [x] Remove IEDO simple flowsheets from GUI @MarcusHolly #1417
- [x] Reorganize and purge watertap/examples/flowsheets/case_studies. This structure was made to isolate/bury the non-predictive models. Instead, I believe we should rename this folder something like "zero_order". All flowsheets that are not zero order like BSM1, BSM2, etc. should just be moved up to flowsheets (we want these to be visible anyways). In my opinion, all FOA 2336 zero order flowsheets that do not have an active collaboration should be removed (we can always dig them back up if we begin working with the teams in the future). @MarcusHolly #1438 

Potential clean up:

[...]

PR #1422, opened 2024-05-31
Remove legacy WT3 data files

kurbansitterley:
## Fixes/Resolves:

Removes old WT3 data files (.csvs) that were buried in the case_studies directory.
Possibly part of #1248 ??

## Summary/Motivation:

As far as I can tell, these .csvs were not used in any unit model or example/demonstration. I assume we kept them around just as a reference as we were porting the WT3 models into WaterTAP, but they are not needed anymore so could be confusing to someone poking around the examples folder. Also, based on some of the entries, they appear to be copies of versions I had on my local machine at the time of the merger... I am not sure how that happened. Their purpose for being in the watertap repo is not documented anywhere. Their origin is not documented anywhere. Just a couple of random .csvs floating in the ether...

If ever they are needed again, they are available here: https://github.com/NREL/WaterTAP3

- Removes from WaterTAP old legacy WT3 .csvs that are not used for anything.

### Legal Acknowledgement


MarcusHolly (approved):
LGTM - I was not even aware these files existed

bknueven (approved):

Issue #1423, opened 2024-06-03
RO .report() Units


zacharybinger:
### Description

The RO model report units have been bugging me. In the report, the solvent flux is reported as L/m^2/hr (LMH), which is convention, but the actual variable being passed is the mass flux not volumetric flux. This is also an issue with the other variables printed in the Expression section or the .report()

Report Key | Units Indicated | Actual Units Printed
Average Solvent Flux (LMH) | L/m2/hr | kg/m2/s
NaCl Average Mass Transfer Coefficient (mm/h) | mm/hr | m/s
NaCl Average Solute Flux (GMH) | g/m2/hr | kg/m2/s

https://github.com/watertap-org/watertap/blob/f2ae7c7f1b849d1489b135dea01983c1e41d7f9a/watertap/unit_models/reverse_osmosis_base.py#L773-L788

When configuring the report, the values to be printed are converted to the indicated units (LMH, GMH, etc) but the base units are printed (kg/m2/s, etc), so the values and the units do not match

![Screenshot 2024-06-03 at 11 49 12 AM](https://github.com/watertap-org/watertap/assets/23547346/379bd6a3-9750-494c-9b91-a6d9aaab5f68)

It's a VERY small detail, but probably an important correction. I tried creating a volumetric flux variable that would print here instead, but the report logic convert all the values and variables to their "base_units" anyways, so printing a report with volumetric flux would print the units as m/s anyways, which would probably be even more confusing.

### Steps to Reproduce

Any environment
Build, Initialize, and Solve any RO Unit Model with the config option: has_full_reporting = True
print RO unit report

### Environment

- OS: MacOS
- Python: 3.8.12
- WaterTAP: 1.0.dev0

adam-a-a:
To add to that, if you check reporting on other models, e.g., wherever we intended for kWh/m3 to be displayed, the conversion to base units behind the scenes converts SEC (or electricity intensity, as named in some models) to Pa, which is definitely not desirable. I have an initial idea on how to fix this, but it would require changes in IDAES that would not go through quickly and probably undergo some back-and-forth with reviewers, so putting a pin in this for now.

PR #1424, opened 2024-06-03
Fix RO Report Units

zacharybinger:
## Fixes/Resolves:
#1423 

## Summary/Motivation:
The RO model report units have been bugging me. In the report, the solvent flux is reported as L/m^2/hr (LMH), which is convention, but the actual variable being passed is the mass flux not volumetric flux. This is also an issue with the other variables printed in the Expression section or the .report()

Further details in liked issue

- Remove value conversions to match printed units
- Remove units is Key that indicate the different units than printed by report

### Legal Acknowledgement


adam-a-a (approved):
LGTM- I actually spent some time trying to maintain the originally intended units and attempted to alter the reported units, but it ends up that such a fix is not straightforward and would require changes in IDAES.

kurbansitterley (approved):
LGTM - I have gone down this road before with the same determination as @adam-a-a

PR #1428, opened 2024-06-06
Remove Units from Report Strings

MarcusHolly:
## Summary/Motivation:
Gets rid of units that are explicitly stated in the strings of reports

### Legal Acknowledgement


watertap/unit_models/nanofiltration_DSPMDE_0D.py
```diff
@@ -1579,7 +1579,7 @@ def _get_performance_contents(self, time_point=0):
             time_point
         ].flow_vol_phase["Liq"]
 
-        expr_dict["Average Volumetric Flux (LMH)"] = (
+        expr_dict["Average Volumetric Flux"] = (
             self.flux_vol_water_avg[time_point] * 3.6e6
```
kurbansitterley: I believe this needs the same treatment from #1424 (remove the hardcoded *3.6e6)

kurbansitterley (changes requested):
just the change to NF and should be good

watertap/unit_models/nanofiltration_DSPMDE_0D.py
MarcusHolly: Thanks for catching this!

kurbansitterley (approved):
easy enough, LGTM

watertap/unit_models/MD/membrane_distillation_base.py
```diff
@@ -696,7 +696,7 @@ def _get_performance_contents(self, time_point=0):
         if hasattr(self, "width"):
             var_dict["Membrane Width"] = self.width
 
-        expr_dict["Average Solute Flux (LMH)"] = self.flux_mass_avg[time_point] * 1000
+        expr_dict["Average Solute Flux"] = self.flux_mass_avg[time_point] * 1000
```
hunterbarber: @zacharybinger Was the conclusion that the multiplication of the pyomo variables with units by a float actually doesn't do anything in the printing, or these still correctly print the value in the appropriate units?

TimBartholomew (approved):
LGTM

PR #1434, opened 2024-06-12
Enable GreyBoxBlocks with cyipopt-watertap

bknueven:
## Summary/Motivation:
We have an emerging need to use GreyBox models with WaterTAP. This PR would modify our cyipopt wrapper so that GreyBoxBlocks work correctly. This necessitates making a few other changes to the solver wrapper, including bring into WaterTAP the IDAES jacobian scaling so that we can utilize a custom-built PyomoNLP object when we have a model with GreyBoxBlocks.

- Add _get_pyomo_nlp abstract method and implementations for both classes
- Bring in the Jacobian constraint autoscaling capability from IDAES to run on custom-built `PyomoNLP` objects
- Add tests to make sure we deal with the dummy objective correctly
- Add test for Pyomo model with GreyBoxBlocks and `cyipopt-watertap`.

### Legal Acknowledgement


avdudchenko (approved):
LGTM

TimBartholomew (approved):

PR #1435, opened 2024-06-13
ADM1/ASM2d Translator Update

MarcusHolly:
## Summary/Motivation:
Correctly builds the translator block based on c-code from Flores-Alsina

### Legal Acknowledgement


docs/technical_reference/unit_models/translators/translator_adm1_asm2d.rst
```diff
@@ -69,15 +76,35 @@ Equations and Relationships
...
-   "Acetic acid conversion", ":math:`S_{A, out} = Σ_{r2} C_{r2, in}`"
-   "Unchanged component conversions", ":math:`C_{u, out} = C_{u, in}`"
-   "Ammonium conversion", ":math:`S_{NH4, out} = S_{IN, in}`"
-   "Phosphate conversion", ":math:`S_{PO4, out} = S_{IP, in}`"
-   "Biodegradable particulate organics conversion", ":math:`X_{S, out} = Σ_{s} C_{s, in}`"
    "Zero-flow component conversions", ":math:`C_{z, out} = 0`"
-
-
+   "Biomass concentration", ":math:`bio = X_{su, in} + X_{aa, in} + X_{fa, in} + X_{c4, in} + X_{pro, in} + X_{ac, in} + X_{h2, in} + X_{PAO, in}`"
+   "S_ac concentration", ":math:`S_{ac, 1} = S_{ac, in} + X_{PHA, in}`"
```
adam-a-a, line 81: What are the 1 and 2 subscripts meant to represent (e.g., S_{ac, 1} or S_{ac, 2})?

watertap/examples/flowsheets/case_studies/full_water_resource_recovery_facility/BSM2_P_extension.py
```diff
@@ -91,19 +93,57 @@
...
+    #     iscale.set_scaling_factor(x[0], sf)
+    #
+    #     print(
+    #         f" The scaling factor for {var_name} ({x[0].value}) was {old_sf}, but is now {sf}"
+    #     )
+
+    dt = DiagnosticsToolbox(m)
+    print("---Structural Issues---")
+    dt.report_structural_issues()
+
```
adam-a-a: You can probably clean this code up (delete this bit) once you're satisfied with debugging.

watertap/examples/flowsheets/case_studies/full_water_resource_recovery_facility/BSM2_P_extension.py
```diff
@@ -91,19 +93,57 @@
...
     m.fs.MX3.pressure_equality_constraints[0.0, 3].deactivate()
     print(f"DOF after initialization: {degrees_of_freedom(m)}")
 
+    badly_scaled_var_list = iscale.badly_scaled_var_generator(m, large=1e2, small=1e-2)
+    print(
+        "----------------   badly_scaled_var_list after initialization  ----------------"
+    )
+    for x in badly_scaled_var_list:
+        print(f"{x[0].name}\t{x[0].value}\tsf: {iscale.get_scaling_factor(x[0])}")
+
```
adam-a-a: Delete when done troubleshooting

watertap/examples/flowsheets/case_studies/full_water_resource_recovery_facility/BSM2_P_extension.py
```diff
@@ -124,9 +164,22 @@ def main(bio_P=True):
         fail_flag=True,
     )
 
+    print("---Numerical Issues---")
+    dt.report_numerical_issues()
+
+    badly_scaled_var_list = iscale.badly_scaled_var_generator(m, large=1e2, small=1e-2)
```
adam-a-a: Probably should just delete all the checks for badly scaled vars across the code

watertap/examples/flowsheets/case_studies/full_water_resource_recovery_facility/BSM2_P_extension.py
```diff
@@ -665,7 +764,45 @@ def initialize_system(m, bio_P=False):
     seq.set_guesses_for(m.fs.translator_asm2d_adm1.inlet, tear_guesses2)
 
     def function(unit):
-        unit.initialize(outlvl=idaeslog.INFO)
+        # unit.initialize(outlvl=idaeslog.INFO)
+        if unit == m.fs.translator_asm2d_adm1:
+            try:
+                print("Trying to initialize ASM2d-ADM1 translator")
```
adam-a-a: Might be better to use a logger

watertap/examples/flowsheets/case_studies/full_water_resource_recovery_facility/BSM2_P_extension.py
```diff
@@ -665,7 +764,45 @@ def initialize_system(m, bio_P=False):
...
+            except:
+                print("Entering exception clause")
+                m.fs.translator_asm2d_adm1.inlet.flow_vol.fix()
+                m.fs.translator_asm2d_adm1.inlet.conc_mass_comp.fix()
+                m.fs.translator_asm2d_adm1.inlet.temperature.fix()
+                m.fs.translator_asm2d_adm1.inlet.pressure.fix()
+
+                solver = pyo.SolverFactory("ipopt")
+                solver.solve(m.fs.translator_asm2d_adm1, tee=True)
+
```
adam-a-a: Do you add this because the solve doesn't work without this? Or is this due to transcribing improvements from Soraya? Just want to point out that the default solver we use is ipopt-watertap which is basically a tweaked version of ipopt that ends up being more suitable for problems in watertap. However, by calling ipopt here, you are switching to the IDAES version of ipopt, without the applied tweaks meant for solving watertap models. If you simply delete line 779, you'll see if this solves with the default ipopt-watertap

watertap/unit_models/translators/tests/test_translator_adm1_asm2d.py
```diff
@@ -237,7 +250,7 @@ def test_solution(self, asmadm):
         assert pytest.approx(1e-10, abs=1e-6) == value(
             asmadm.fs.unit.outlet.conc_mass_comp[0, "S_O2"]
         )
-        assert pytest.approx(0.22677, rel=1e-3) == value(
+        assert pytest.approx(20.814923, abs=1e-6) == value(
```
adam-a-a: Interesting shift in order of magnitude

watertap/unit_models/translators/tests/test_translator_adm1_asm2d.py
```diff
@@ -283,31 +296,5 @@ def test_conservation(self, asmadm):
...
-                        * asmadm.fs.props_ASM2D.dens_mass
-                        * asmadm.fs.props_ASM2D.cp_mass
-                        * (
-                            asmadm.fs.unit.outlet.temperature[0]
-                            - asmadm.fs.props_ASM2D.temperature_ref
-                        )
-                    )
-                )
-            )
-            <= 1e-6
```
adam-a-a: Is this no longer converging with <=1e-6? And is the second check just not working at all anymore? Curious whether you removed it due to test failures or because you thought it was no longer valid.

watertap/unit_models/translators/translator_adm1_asm2d.py
```diff
@@ -96,8 +122,49 @@ def build(self):
         # Call UnitModel.build to setup dynamics
         super(TranslatorDataADM1ASM2D, self).build()
 
+        eps = 0
+        mw_p = 31 * pyunits.kg / pyunits.kmol
+        mw_n = 14 * pyunits.kg / pyunits.kmol
         mw_c = 12 * pyunits.kg / pyunits.kmol
 
+        self.f_sI_xc = Param(
```
adam-a-a: These factors look very familiar... are you sure these don't already live on ASM2D or ADM1 reaction or prop models?

adam-a-a (commented):
I have a few comments/questions on this.

docs/technical_reference/unit_models/translators/translator_adm1_asm2d.rst
MarcusHolly: Yes - the 1 and 2 subscripts correspond to the concentration of the given component at steps 1 and 2 (named like `Sac_AD1` in the code). This translator only has 2 steps (whereas the other translator has 5).

watertap/examples/flowsheets/case_studies/full_water_resource_recovery_facility/BSM2_P_extension.py
MarcusHolly: Yea, I think all the flowsheets are cleaned up now

watertap/examples/flowsheets/case_studies/full_water_resource_recovery_facility/BSM2_P_extension.py
MarcusHolly: I was mainly using this to try to get an initial solution that would help with providing better tear guesses. This is no longer necessary, so it's been deleted. Thanks for the ipopt clarification too!

watertap/unit_models/translators/tests/test_translator_adm1_asm2d.py
MarcusHolly: I think a significant change in this value is to be expected. The equation for the outlet S_PO4 concentration changed from `S_{PO4, out} = S_{IP, in}` to `S_{PO4, out} = S_{IP, in} + S_{IP, 1} + S_{IP, 2}

watertap/unit_models/translators/tests/test_translator_adm1_asm2d.py
MarcusHolly: I removed it due to test failures, but I've since added it back in. Turns out the issue was that I had accidentally deleted `iscale.set_scaling_factor(self.properties_out[0].flow_vol, 1e5)` in the translator block, although I'm not exactly sure why this change made such a difference.

watertap/unit_models/translators/translator_adm1_asm2d.py
MarcusHolly: They live in the ADM1 reaction models, but I'm not sure of the correct syntax for implementing it here. I assumed it would be `blk.config.inlet_reaction_package.f_sI_xc`, but that's not correct. If we do update this, we should also update it in the ASM2d/ADM1 translator as well.

watertap/examples/flowsheets/case_studies/full_water_resource_recovery_facility/test/test_BSM2_P_extension.py
```diff
@@ -148,58 +148,58 @@ def test_solve(self, system_frame):
...
-        ) == pytest.approx(0.058799, rel=1e-3)
+        ) == pytest.approx(0.024905, rel=1e-3)
         assert value(
             m.fs.Treated.properties[0].conc_mass_comp["S_NH4"]
-        ) == pytest.approx(0.00014743, rel=1e-3)
+        ) == pytest.approx(0.1816027, rel=1e-3)
         assert value(
             m.fs.Treated.properties[0].conc_mass_comp["S_NO3"]
-        ) == pytest.approx(0.0077378, rel=1e-3)
+        ) == pytest.approx(5.11984e-11, rel=1e-3)
```
adam-a-a:
```suggestion
        ) == pytest.approx(5.11984e-11, abs=1e-6)
```

MarcusHolly:
@adam-a-a @luohezhiming Tests are still failing here (initialization issues), but I'm not sure how to go about fixing it since everything still passes locally

watertap/flowsheets/full_water_resource_recovery_facility/BSM2_P_extension.py
```diff
@@ -138,7 +140,7 @@ def main(bio_P=True):
     return m, results
 
 
-def build(bio_P=False):
+def build(bio_P=True):
```
Morgan88888888, line 146: I'm wondering if there's a reason for changing the default values (False to True). It would be helpful to document this, even with a simple comment.

docs/technical_reference/unit_models/translators/translator_adm1_asm2d.rst
```diff
@@ -69,15 +76,35 @@ Equations and Relationships
...
-   "Fermentable substrate conversion", ":math:`S_{F, out} = Σ_{r1} C_{r1, in}`"
-   "Acetic acid conversion", ":math:`S_{A, out} = Σ_{r2} C_{r2, in}`"
-   "Unchanged component conversions", ":math:`C_{u, out} = C_{u, in}`"
-   "Ammonium conversion", ":math:`S_{NH4, out} = S_{IN, in}`"
-   "Phosphate conversion", ":math:`S_{PO4, out} = S_{IP, in}`"
-   "Biodegradable particulate organics conversion", ":math:`X_{S, out} = Σ_{s} C_{s, in}`"
    "Zero-flow component conversions", ":math:`C_{z, out} = 0`"
-
-
+   "Biomass concentration", ":math:`bio = X_{su, in} + X_{aa, in} + X_{fa, in} + X_{c4, in} + X_{pro, in} + X_{ac, in} + X_{h2, in} + X_{PAO, in}`"
```
luohezhiming: I would suggest to rename `bio` as `X_{bio}` for consistency

docs/technical_reference/unit_models/translators/translator_adm1_asm2d.rst
luohezhiming: A little bit confused here, e.g., I see you set `X_{PP,1} = 0` and then there are some conversion equations like `S_IP` will add `X_{PP,1}`, but it is set as 0, so what is the purpose for this conversion? Add 0 does not seem to have any impacts of the outcome. and where does `X_PP` convert to?

watertap/flowsheets/electroNP/BSM2_electroNP_no_bioP.py
```diff
@@ -627,6 +627,8 @@ def solve(m, solver=None):
     if m.fs.has_electroNP is False:
         stream_table = create_stream_table_dataframe(
             {
+                "R3 inlet": m.fs.R3.inlet,
+                "ASM-ADM translator inlet": m.fs.translator_asm2d_adm1.inlet,
                 "Feed": m.fs.FeedWater.outlet,
```
luohezhiming: You can delete these two outputs for R3 inlet and ASM-ADM translator inlet, I used them for getting tear_guess but don't need to be shown in the report.

watertap/flowsheets/electroNP/BSM2_electroNP_no_bioP.py
```diff
@@ -646,6 +648,8 @@ def solve(m, solver=None):
     else:
         stream_table = create_stream_table_dataframe(
             {
+                "R3 inlet": m.fs.R3.inlet,
+                "ASM-ADM translator inlet": m.fs.translator_asm2d_adm1.inlet,
                 "Feed": m.fs.FeedWater.outlet,
```
luohezhiming: Same as above

docs/technical_reference/unit_models/translators/translator_adm1_asm2d.rst
MarcusHolly: I have done that just to be explicit, and the c-code also leaves in such terms even when they are set equal to 0. I wanted to avoid somebody cross-checking our code with the c-code and wondering why a term is missing from an expression

watertap/flowsheets/full_water_resource_recovery_facility/test/test_BSM2_P_extension.py
```diff
@@ -206,10 +114,10 @@ def test_costing(self, system_frame):
         m = system_frame
 
         # check costing
-        assert value(m.fs.costing.LCOW) == pytest.approx(0.46904, rel=1e-3)
+        assert value(m.fs.costing.LCOW) == pytest.approx(0.46814, rel=1e-3)
         assert value(m.fs.costing.total_capital_cost) == pytest.approx(
-            23985398.4909, rel=1e-3
+            23939185.51, rel=1e-3
```
Morgan88888888: I think it would be better to use a smaller tolerance for this large value. A tolerance of 0.1% can still represent a significant absolute difference.

watertap/flowsheets/full_water_resource_recovery_facility/BSM2_P_extension.py
```diff
@@ -510,16 +513,17 @@ def set_operating_conditions(m):
...
+def set_scaling(m):
+    for var in m.fs.component_data_objects(pyo.Var, descend_into=True):
+        if "flow_vol" in var.name:
+            iscale.set_scaling_factor(var, 1e1)
+        if "temperature" in var.name:
+            iscale.set_scaling_factor(var, 1e-2)
+        if "pressure" in var.name:
+            iscale.set_scaling_factor(var, 1e-4)
+        if "conc_mass_comp" in var.name:
+            iscale.set_scaling_factor(var, 1e2)
```
luohezhiming: Why do you change the indent here? the `scale_variables` should be some internal function for `build`

docs/technical_reference/unit_models/translators/translator_adm1_asm2d.rst
MarcusHolly: X_PP is converted into S_Mg, S_K, and S_IP

Morgan88888888 (changes requested):
The models look good, and the tests are passing. Just a few questions to address.

watertap/flowsheets/full_water_resource_recovery_facility/BSM2_P_extension.py
```diff
@@ -532,12 +536,56 @@ def scale_variables(m):
...
+    iscale.set_scaling_factor(
+        m.fs.translator_asm2d_adm1.properties_out[0].conc_mass_comp["X_ac"], 1e9
+    )
+    iscale.set_scaling_factor(
+        m.fs.translator_asm2d_adm1.properties_out[0].conc_mass_comp["X_I"], 1e-1
+    )
+    iscale.set_scaling_factor(
+        m.fs.translator_asm2d_adm1.properties_out[0].conc_mass_comp["X_PHA"], 1
+    )
+
```
luohezhiming: Not so sure how you get these scaling factors but the `scale_variable` function is not used in this code anymore

luohezhiming (changes requested):
Great work, couple comments and I'll take a look at scaling

watertap/flowsheets/full_water_resource_recovery_facility/BSM2_P_extension.py
MarcusHolly: I was playing around with when scaling was applied and just never reverted this back. Currently this function is being called in `main` after setting the operating conditions. Is there a meaningful difference between applying scaling in `build` and applying it after `set_operating_conditions`?

watertap/flowsheets/full_water_resource_recovery_facility/BSM2_P_extension.py
MarcusHolly: False is what this flowsheet should be using ideally, but it's currently not as stable - hence why I've had to remove testing the `bio_P=False` config option from the flowsheet test file. This is also why `bio_P=True` is set as the default in the GUI.

docs/technical_reference/unit_models/translators/translator_adm1_asm2d.rst
luohezhiming: From the equation I can read, it seems that you set X_PP=0 then S_Mg, S_K ans S_IP just add 0, the initial value of X_PP does not have any impact of the outcome

MarcusHolly: Yes, that is correct. It has no effect since it's zero. But is it possible that in the future we want to change X_PP from 0 to something else? If so, I think it's worth keeping this zero term here in the documentation and in the code itself. It's also just the most accurate representation of these equations.

watertap/flowsheets/full_water_resource_recovery_facility/BSM2_P_extension.py
MarcusHolly: Well I've changed the name from `scale_variable` to `set_scaling` to be a bit more consistent with other function names, but it is being used in the code. `main` calls it after setting the operating conditions

docs/technical_reference/unit_models/translators/translator_adm1_asm2d.rst
luohezhiming: @adam-a-a hope you could also join the talk, I am confused because I believe X_PP should be important components of the system, right now, it seems no matter the initial value of X_PP is, it will directly set to 0 and no actual conversion happens to the system. Same conditions to X_PHA and X_PAO

MarcusHolly: Oh... yea there's a mistake here. S_{IP, 1} should be X_{PP, in) + ... and not S_{IP, in) + ...

MarcusHolly: Great catch @luohezhiming. This may or may not screw up the flowsheet again, but I'll test it out now...

MarcusHolly: I will do a line-by-line review of the new translator file after this, which I just haven't had the time to do yet

adam-a-a:
@luohezhiming is still trying to get bioP=True to solve with electroN-P included, but tests for other configurations should be passing. We just need those local changes to be pushed, and from there, we expect tests to pass and can consider temporarily skipping bioP=True with electronP.

adam-a-a:
Related to issue #1460

watertap/flowsheets/full_water_resource_recovery_facility/BSM2_P_extension_ui.py
```diff
@@ -3848,28 +3852,23 @@ def build_flowsheet(build_options=None, **kwargs):
     """
     Builds the initial flowsheet.
     """
-    m = build()
-
+    m = build(bio_P=False)
```
MarcusHolly: @adam-a-a @luohezhiming This GUI now only works when bio_P=False

watertap/flowsheets/full_water_resource_recovery_facility/test/test_BSM2_P_extension.py
```diff
@@ -31,7 +31,7 @@
 )
 
 
-class TestFullFlowsheet:
+class TestFullFlowsheetBioPFalse:
```
MarcusHolly, line 37: Only bio_P=False passes on GitHub

adam-a-a (approved):
Still work to do, but at least we have this passing checks without costing!

adam-a-a:
@luohezhiming let's merge this through once all checks pass and I will continue troubleshooting on PR #1467 while you work on modified ADM1 flowsheet

luohezhiming (commented):
The `BSM2_electroNP.py needs` to be deleted!

luohezhiming (approved):

PR #1438, opened 2024-06-15
Delete ZO Files

MarcusHolly:
## Fixes/Resolves:

Issue #1248 

- Deletes ZO flowsheets, which will be archived in a separate repo
- Changes flowsheet directory from watertap/examples/flowsheets to watertap/flowsheets
- Updates various imports and hard-coded file locations

### Legal Acknowledgement


adam-a-a (changes requested):
I can accept deleting the flowsheets, though I see it as unnecessary without bringing any positive outcomes.

However, I disagree with removing docs, yamls, and unit models associated with IEDO work. Why would we do this and yet keep all of the other ZO work associated with the WaterTAP3 migration?

TimBartholomew (approved):

adam-a-a (approved):

PR #1458, opened 2024-06-28
Enable concentration factor survey in OLI API

adam-a-a:
## Summary/Motivation:
One way to streamline a composition survey in OLI API, where we want to simulate the concentrating of solutes in solution as water is removed from that solution is to conduct a survey on H2O itself (as opposed to applying concentration factors to each solute in composition via a survey).

However, as our oli_api/client and oli_api/flash are configured now, this is not possible. This PR makes some adjustments to set the stage for conducting a concentration factor survey in a future PR by dealing with units as OLI would expect them to be.

- add "inflows" to FixedKeysDict input_unit_set
- alter the Flash code so that it propagates units from "inflows" allowing the user to change "molecularConcentration" without unintended consequences
- subtle cleanup and testing

### Legal Acknowledgement


watertap/tools/oli_api/flash.py
```diff
@@ -328,15 +328,15 @@ def configure_water_analysis(
...
+        conc_unit = self.input_unit_set["inflows"]["oli_unit"]
         _logger.info(f"Using {conc_unit} for inflows input")
         for k, v in inflows.items():
             charge = get_charge(k)
             input_list.append(
                 {
                     "group": get_charge_group(charge),
                     "name": get_oli_name(k),
-                    "unit": self.input_unit_set["molecularConcentration"]["oli_unit"],
+                    "unit": conc_unit,
```
adam-a-a, line 339: most important changes of the PR. Simple fix

watertap/tools/oli_api/flash.py
```diff
@@ -573,7 +573,13 @@ def configure_flash_analysis(
...
                 )
 
-        input_dict["inflows"] = inflows
+        if ("unit" in inflows.keys()) and ("values" in inflows.keys()):
+            input_dict["inflows"] = inflows
+        # otherwise, assume a solute dictionary with {solute_name: concentration_value}
+        else:
+            unit = self.input_unit_set["molecularConcentration"]["oli_unit"]
+            inflows_oli_names = {get_oli_name(k): v for k, v in inflows.items()}
+            input_dict["inflows"] = {"unit": unit, "values": inflows_oli_names}
```
adam-a-a, line 582: Another important change to allow for the user to just supply a solute:concentration dict to the `configure_flash_analysis` method for setting up isothermal flash calculations. 

This is what we do for our `configure_water_analysis` method, and this `configure_flash_analysis` seems to assume it'll be getting output from `get_apparent_species_from_true`; i.e., without this PR, the code assumes that the user is running a water analysis first and then extracting apparent species output as input to this method; either that, or the user needs to manually supply a dict that doesn't match the input dict format that goes into `configure_water_analysis`. 

This standardizes things to make things less confusing for the user.

watertap/tools/oli_api/flash.py
```diff
@@ -808,9 +814,12 @@ def _add_to_json(
         else:
             json_input["params"] = input_data
 
+        output_unit_set_info = {}
+        for k, v in self.output_unit_set.items():
+            output_unit_set_info[k] = v["oli_unit"]
```
adam-a-a, line 819: I turned output_unit_set into a nested FixedKeysDict, which basically means that (1) users can't overwrite keys, which makes sense because these keys correspond to specific OLI output property names, and (2) setting the `oli_unit` or the `pyomo_unit` of a given property will sync with both `oli_unit` and `pyomo_unit` values. See #1388 for more background if that doesn't make sense.

avdudchenko (approved):
LGMT

OOAmusat (approved):
I think this is fine.

Issue #1380, opened 2024-05-08
Spiral-wound module type issues with OARO


zacharybinger:
### Description

So there is a issue with applying the spiral-wound module type to the OARO unit model or any future osmotically-driven membrane process that operates with two fluids on each side of the membrane. To operate with two streams on a spiral-wound module, a baffle needs to be added to the "permeate side" of the envelope. I attached a figure to visualize this. I will try coming up with a simple solution to this issue, but wanted to create an issue to highlight the possibility of removing the spiral-wound module_type from OARO.
![1-s2 0-S0011916419323653-gr5_lrg](https://github.com/watertap-org/watertap/assets/23547346/d8a73f31-1a7d-4023-9c19-7fd2fbc20f0f)

### Expected Behavior

- Error in permeate recovery prediction

### Environment

- OS:
- Python:
- WaterTAP:

adam-a-a:
One approach is to undo the changes made that affected OARO in #1272 (we want to keep the changes that affected RO though). FYI @MarcusHolly @ElmiraShamlou @luohezhiming

PR #1461, opened 2024-07-02
OARO remove spiral-wound ModuleType

zacharybinger:
## Fixes/Resolves:
#1380 

## Summary/Motivation:
Implementation of OARO with a spiral-wound module is more complex than the initial implementation represented. Enabling spiral-wound operation of OARO would likely require another PR with more creative/extensive changes that I don't think are possible with our current capabilities. More precisely spiral-wound OARO would require a 2D model.

- Removing ModuleType.spiral_wound support for OARO models

### Legal Acknowledgement


MarcusHolly (approved):
LGTM

adam-a-a:
>  Enabling spiral-wound operation of OARO would likely require another PR with more creative/extensive changes that I don't think are possible with our current capabilities

Just want to clarify that the underlying _capabilities_ are there (e.g., IDAES has at least one 2d model that I can recall off the top of my head)--it is just that we haven't implemented any 2d models in WaterTAP thus far. Although such an effort would require a lot of time and effort, I just want to make sure that anyone reading through this PR understands that such a model is possible in WaterTAP and would be facilitated by some of the building blocks in IDAES.

adam-a-a (changes requested):
LGTM- the only necessary changes are updating the documentation for OARO 0D and 1D, wherever spiral-wound modules are mentioned.

ElmiraShamlou:
@zacharybinger @adam-a-a How about keeping the spiral wound option for the OARO 0D model instead of fully removing it? For velocity calculations in the permeate channel, we can redefine the cross-sectional area as follows:

Velocity (permeate channel) = Qavg / (Porosity * (Module Length / 2) * Channel Height)
where:
Module Length: Length of the module (used instead of width to reflect the perpendicular flow direction)
Division by 2: Accounts for the effects of baffles in the permeate channel

adam-a-a:
> @zacharybinger @adam-a-a How about keeping the spiral wound option for the OARO 0D model instead of fully removing it? For velocity calculations in the permeate channel, we can redefine the cross-sectional area as follows:
> 
> Velocity (permeate channel) = Qavg / (Porosity * (Module Length / 2) * Channel Height) where: Module Length: Length of the module (used instead of width to reflect the perpendicular flow direction) Division by 2: Accounts for the effects of baffles in the permeate channel

Hmm.. maybe that _could_ work. Does anyone know if we have any peer-reviewed literature on spiral wound OARO performance to test against?

zacharybinger:
> .. maybe that _could_ work. Does anyone know if we have any peer-reviewed literature on spiral wound OARO performance to test against?

Not to my knowledge, nor in the quick literature search I just did

adam-a-a:
@zacharybinger - just a reminder that this PR should be good to go soon--just waiting on the corresponding updates to the documentation to reflect removal of SW module type from OARO. We can revisit @ElmiraShamlou 's suggestion via a subsequent issue/PR as needed.

zacharybinger:
@adam-a-a Ok, once these checks pass it should be good to go then. I removed any mention of spiral-wound from the docs, there were only a few lines

adam-a-a (approved):
LGTM

PR #1465, opened 2024-07-12
Adding volume to MembraneChannel0D

adam-a-a:
## Summary/Motivation:
To enable dynamics with the RO model, one initial thing we need to do is add the volume variable into the 0D model. From this, the material accumulation constraint used in IDAES' 0D control volume can be applied, at least for the feed side of the membrane. 

- add volume variable and constraint
- bring back some tests for RO0D
- add volume var and constraint to documentation for RO0D

### Legal Acknowledgement


watertap/core/membrane_channel_base.py
```diff
@@ -658,14 +658,15 @@ def _add_calculated_pressure_change_mass_transfer_components(self):
...
             )
-
-        self.channel_height = Var(
-            initialize=1e-3,
-            bounds=(1e-4, 5e-3),
-            domain=NonNegativeReals,
-            units=units_meta("length"),
-            doc="membrane-channel height",
-        )
+        # may already be constructed under add_geometry() in MembraneChannel0D
```
```suggestion
        # in case channel_height is already constructed
```

watertap/core/membrane_channel_base.py
```diff
@@ -639,7 +639,7 @@ def _add_calculated_pressure_change_mass_transfer_components(self):
         # `_add_calculated_pressure_change` *and/or*
         # `_add_calculated_mass_transfer_coefficient`.
         # Therefore, we add this simple guard against it being called twice.
-        if hasattr(self, "channel_height"):
+        if hasattr(self, "dh"):
```
kurbansitterley, line 642: Can you explain to me why ``dh`` replaces ``channel_height`` in this check?

docs/technical_reference/unit_models/reverse_osmosis_0D.rst
fahim831, line 1: In line 181 (RHS file), where is the 0.3 exponent applied? The `\frac` looks like it's taking 6.23 and Re as its args so it will appear as 6.23/Re. So does it become (6.23/Re)^0.3 or 6.23/Re^0.3?

Is there a reason to move line 182 (LHS) to 163 (RHS)?

docs/technical_reference/unit_models/reverse_osmosis_0D.rst
```diff
@@ -159,6 +160,7 @@ Equations
    "Solvent flux across membrane (SKK)", ":math:`J_{solvent} = \rho_{solvent} A(P_{f} - P_p - \sigma(\pi_{f}-\pi_{p}))`"
    "Solute flux across membrane (solution-diffusion)", ":math:`J_{solute} = B(C_{f} - C_{p})`"
    "Solute flux across membrane (SKK)", ":math:`J_{solute} = B(C_{f} - C_{p}) + (1 - \sigma)\frac{J_{solvent}}{\rho_{solvent}}C_{f}`"
+   "Alpha (SKK)", ":math:`\alpha = \frac{1 - \sigma}{B}`"
```
adam-a-a, line 163: @fahim831 I moved alpha here just to have it follow other rows having to do with the Spiegler-Kedem-Katchalsky model (SKK).

docs/technical_reference/unit_models/reverse_osmosis_0D.rst
```diff
@@ -174,12 +176,13 @@ Equations
    "Membrane area (spiral-wound)",":math:`A_m = 2LW`"
    "Pressure drop",":math:`ΔP = (\frac{ΔP}{Δx})_{avg}L`"
    "Feed-channel velocity",":math:`v_f = Q_f/A_c`"
-   "Friction factor",":math:`f = 0.42+\frac{189.3}{Re}`"
+   "Feed-channel volume", ":math:`V_{f} = LWh_{ch}`"
+   "Friction factor (flat-plate)",":math:`f = 0.42+\frac{189.3}{Re}`"
+   "Friction factor (spiral-wound)",":math:`f = \frac{6.23}{Re^{0.3}}`"
```
adam-a-a, line 181: @fahim831 good catch- I revised this and the exponent should show up as anticipated now in the documentation.

adam-a-a: Here's what it looks like now:
image

adam-a-a: Note-- this isn't related to the mini-step I took towards enabling a dynamic RO model. I just realized that this was missing from our RO documentation in the process of adding channel volume to the documentation.

kurbansitterley (approved):

watertap/core/membrane_channel_base.py
ElmiraShamlou: I think it's because if hasattr(self, "channel_height"): return was preventing the _add_calculated_pressure_change_mass_heat_transfer_components from being called twice. But now that the channel height is added to the geometry method of the channel control volume, it would be created and incorrectly prevent the _add_calculated_pressure_change_mass_heat_transfer_components from being called. So the safeguard variable has changed to dh.

ElmiraShamlou (approved):
Nice job!

watertap/core/membrane_channel_base.py
kurbansitterley: Ah ok, thank you for sorting the logic there!

watertap/core/membrane_channel0d.py
```diff
@@ -64,6 +65,32 @@ def add_geometry(
...
+            if not hasattr(self, "channel_height"):
+                self.channel_height = Var(
+                    initialize=1e-3,
+                    bounds=(1e-4, 5e-3),
+                    domain=NonNegativeReals,
+                    units=units_meta("length"),
+                    doc="membrane-channel height",
+                )
+
+            # # TODO: negating spacer volume for now as it can be assumed negligible (Park et al., 2020: https://doi.org/10.1016/j.desal.2020.114625). Revisit.
```
zacharybinger, line 86: I don't think this is a valid assumption. The paper says "spacer volume is shown to be unimportant" but they are speaking in the context of SEC (0.5% increase). But this PR is trying to include volume for the purpose of dynamics which could be impacted by changes in channel volume. In the SI of this paper they state for their system, the spacers take up 10-20% of the channel volume. This could be significant.

fahim831 (approved):

watertap/core/membrane_channel0d.py
```diff
@@ -64,6 +65,32 @@ def add_geometry(
...
+            # # Note: considering that this volume relationship holds for both flat_sheet and spiral_wound types
+            @self.Constraint(
+                self.flowsheet().config.time, doc="Membrane-channel volume"
+            )
+            def eq_volume(b, t):
+                return b.volume[t] == b.length * b.width * b.channel_height
```
zacharybinger, line 92:
```suggestion
                return b.volume[t] == (b.length * b.width * b.channel_height)*spacer_porosity
```
But since spacer porosity isn't a var on this level, maybe we take care of this in the RO unit model?

zacharybinger (approved):
I think the spacer volume does need to get addressed eventually, but not critical for now since those changes should probably go in the RO unit model to keep the membrane_channel model more generalized

watertap/core/membrane_channel_base.py
adam-a-a: Exactly @ElmiraShamlou ! I did have some internal debate with myself about where to add the volume constraint, but this is what I arrived at. I assume we will find out later if this was a good or bad idea as we progress with efforts on implementing the dynamic RO unit.

watertap/core/membrane_channel0d.py
adam-a-a: Yes, I noticed that and perhaps wasn't descriptive enough in my comment. I also agree with you.

However, since this PR is just a mini-step in the right direction (there will be subsequent mini-PRs to help bring in the dynamic capability for RO), I put this as to-do item to revisit later. 

Rather than accounting for the loss in volume attributed to spacer volume right now, which will be more important later when we're looking for accuracy, the focus right now is just making sure new additions don't break existing code and that the _functionality works_. I don't need spacer volume to test the functionality.

zacharybinger: Sounds good and makes sense to me!

watertap/core/membrane_channel0d.py
adam-a-a: This is partly why I am pushing this for a later PR. I want to first get RO working as anticipated with dynamics and has_holdup set to True, making any _necessary changes_ to model structure as we go. 

Then, once that is set, we can come back to this and incorporate spacer volume. Also, when we _do get back to this_, spacer_porosity seems like an approximation. Do you think we should settle for that and assume negligible difference in comparison with true spacer volume?

PR #1471, opened 2024-07-24
Enable RO0D to be dynamic

fahim831:
## Fixes/Resolves: Membrane and RO modules being forced to be steady-state.

## Summary/Motivation:
At the moment, `dynamic` and `has_holdup` terms are hardcoded to be `False` in the membrane channel base class and RO 0D class. I have changed them to be allowed to be `Bool` and inherit from the model declaration.

- Changed `dynamic` and `has_holdup` domains to `Bool` in `watertap/core/membrane_channel_base.py`
- Changed `dynamic` and `has_holdup` in `_add_feed_side_membrane_channel_and_geometry` method of `watertap/unit_models/reverse_osmosis_0D.py` to inherit from `self.config`

### Legal Acknowledgement


lbianchi-lbl:
As per @adam-a-a, we should wait to merge this until a prototype of the relevant dynamic model(s) has been implemented.

adam-a-a:
It appears that our unit inconsistency issue is a known issue on the Pyomo repo: https://github.com/Pyomo/pyomo/issues/1790

adam-a-a:
I should also note that we are pointing our idaes dependency to a PR on the idaes repo to make this PR work: https://github.com/IDAES/idaes-pse/pull/1460

adam-a-a:
@andrewlee94 wondering if the error we see in the mac check is something to flat out ignore due to the typical instability we see in mac checks, or if the error we see is due to some changes in the latest version of IDAES:
`FAILED watertap/unit_models/tests/test_cstr.py::TestInitializers::test_general_hierarchical - RuntimeError: A feasible solution was not found, so no solution can be loaded.Please set opt.config.load_solutions=False to bypass this error.`

Notably, we haven't touched anything related to the cstr, but we are pointing to a PR that I put up on IDAES here. From the message, it looks like we can just set load_solutions=True but I wanted to check to see if you're familiar with this error at all. I am guessing it has something to do with the new version of ipopt being used, but haven't dug into this.

andrewlee94:
@adam-a-a A few things: first, that error message is misleading and I have been pushing the Pyomo team to replace it with more useful warnings; short answer is do not read much into that error message.

Secondly, this is probably happening due to the change to the new solver writer, and is quite likely due to scaling issues. Unfortunately there is a bug/edge case in the new writer where if IPOPT does not return an optimal solution, the writer will not load the state from the final iteration like the old one did (which results in the error message above). I am pushing to get this fixed, as it precludes debugging the infeasible solution.

Finally, you could try turning off the linear pre-solve to see if that helps - you should be able to add this to the `solver.solve` command:

```
solver.solve(m, writer_config={"linear_presolve": False})
```

setup.py
```diff
@@ -21,14 +21,15 @@
 long_description = (cwd / "README.md").read_text()
 
 SPECIAL_DEPENDENCIES_FOR_RELEASE = [
-    "idaes-pse>=2.5.0,<2.6.0rc0",  # from PyPI
+    # "idaes-pse>=2.5.0,<2.6.0rc0",  # from PyPI
+    "idaes-pse @ https://github.com/IDAES/idaes-pse/archive/refs/heads/dynamic_cv0d.zip",
```
lbianchi-lbl: This can (and should) be reverted since `SPECIAL_DEPENDENCIES_FOR_RELEASE` is only used for releases

setup.py
```diff
@@ -21,14 +21,15 @@
...
 ]
 
 SPECIAL_DEPENDENCIES_FOR_PRERELEASE = [
     # update with a tag from the nawi-hub/idaes-pse
     # when a version of IDAES newer than the latest stable release from PyPI
     # will become needed for the watertap development
-    "idaes-pse==2.5.0",
+    "idaes-pse @ https://github.com/IDAES/idaes-pse/archive/refs/heads/dynamic_cv0d.zip",
```
```suggestion
    "idaes-pse @ git+https://github.com/IDAES/idaes-pse@dynamic_cv0d",
```

setup.py
```diff
@@ -88,6 +89,7 @@
         "pydantic >= 2, <3",
         "numpy",
         "importlib-resources",
+        "idaes-pse @ https://github.com/IDAES/idaes-pse/archive/refs/heads/dynamic_cv0d.zip",
```
lbianchi-lbl: This can be removed

```suggestion
```

setup.py
adam-a-a:
```suggestion
    "idaes-pse>=2.5.0,<2.6.0rc0",  # from PyPI
```

setup.py
```diff
@@ -28,7 +28,7 @@
     # update with a tag from the nawi-hub/idaes-pse
     # when a version of IDAES newer than the latest stable release from PyPI
     # will become needed for the watertap development
-    "idaes-pse==2.5.0",
+    "idaes-pse @ git+https://github.com/IDAES/idaes-pse@dynamic_cv0d",
```
lbianchi-lbl: Is this branch supposed to exist? The only thing I could find is IDAES/idaes-pse#1460, which has already been merged, so using `main` should be enough:

```suggestion
    "idaes-pse @ git+https://github.com/IDAES/idaes-pse@main",
```

More in general, if you want to use the _name_ of a PR's head branch, you have to use the corresponding fork rather than `idaes-pse`. Otherwise, you can use the PR number by using the "special" ref GitHub creates.

In other words, the following should be equivalent (assuming for the sake of example that IDAES/idaes-pse#1460 is still open and the branch still exists on your fork):

```
# option 1: using the fork and the branch name
idaes-pse @ git+https://github.com/adam-a-a/idaes-pse@dynamic_cv0d

# option 2: using the upstream repo and the PR number
idaes-pse @ git+https://github.com/IDAES/idaes-pse@refs/pull/1460/merge
```

adam-a-a:
@bknueven @andrewlee94 @lbianchi-lbl We are currently pointing to IDAES main in this PR to incorporate the changes that were merged in IDAES PR #[1460](https://github.com/IDAES/idaes-pse/pulls?q=is%3Apr+is%3Aclosed) 

I think that as a result of this, we are seeing a test failure related to cyipopt, and I was hoping one of you might be familiar with this issue or what might be causing it:
`FAILED watertap/core/plugins/tests/test_solvers.py::TestIpoptWaterTAP::test_cyipopt_bound_relax_small
`
If there is a quick fix for this, we could try resolving in a separate PR. Otherwise, my plan was to create an earlier tag that would incorporate the needed, aformentioned IDAES changes, and hopefully those changes went in before whatever is causing the cyipopt discrepancy. Thoughts?

lbianchi-lbl:
> @bknueven @andrewlee94 @lbianchi-lbl We are currently pointing to IDAES main in this PR to incorporate the changes that were merged in IDAES PR #[1460](https://github.com/IDAES/idaes-pse/pulls?q=is%3Apr+is%3Aclosed)
> 
> I think that as a result of this, we are seeing a test failure related to cyipopt, and I was hoping one of you might be familiar with this issue or what might be causing it: `FAILED watertap/core/plugins/tests/test_solvers.py::TestIpoptWaterTAP::test_cyipopt_bound_relax_small ` If there is a quick fix for this, we could try resolving in a separate PR. Otherwise, my plan was to create an earlier tag that would incorporate the needed, aformentioned IDAES changes, and hopefully those changes went in before whatever is causing the cyipopt discrepancy. Thoughts?

I've encountered this in #1483, specifically https://github.com/watertap-org/watertap/pull/1483/commits/ecad89c8260014a7d183336a8c08bae72c39ed3c. As I understand it was due to that test being marked `xfail` (with `strict=True`, i.e. causing a failure if the test unexpectedly passes), because of a feature being implemented in a Pyomo PR (Pyomo/pyomo#3295) that ended up being in the release. In other words: the test was expected to fail before 6.8.0, and expected to pass after 6.8.0.

The change is already in #1483, but I should probably open a separate dedicated PR so that it can be merged ASAP.

adam-a-a:
Thanks for the quick and clear response, @lbianchi-lbl !

adam-a-a:
We should merge #1477 before we merge this

adam-a-a (approved):

watertap/unit_models/tests/test_reverse_osmosis_0D.py
```diff
@@ -982,3 +993,104 @@ def configure(self):
...
+        concentration_polarization_type=ConcentrationPolarizationType.calculated,
+        mass_transfer_coefficient=MassTransferCoefficient.calculated,
+        pressure_change_type=PressureChangeType.calculated,
+        module_type=ModuleType.spiral_wound,
+    )
+
+    time_nfe = len(m.fs.time) - 1
+    TransformationFactory("dae.finite_difference").apply_to(
+        m.fs, nfe=time_nfe, wrt=m.fs.time, scheme="BACKWARD"
+    )
```
luohezhiming, line 1026: It might be better to add this as a config option in future.

luohezhiming (approved):
Looks good to me, I have a minor suggestion that can be addressed in future

PR #1477, opened 2024-08-13
Add 0D Property Helper

bknueven:
## Summary/Motivation:
The `properties` attribute on the MembraneChannel0D is constructed at initialization, which is incompatible with dynamic models which transform time. This PR would make the `properties` attribute a "Mapping" type which looks up the appropriate values in `properties_in`/`properties_out` dynamically.

- Add `_0DPropertyHelper` class
- Use an instance of `_0DPropertyHelper` in `MembraneChannel0D`
- Add tests for `_0DPropertyHelper`.

### Legal Acknowledgement

### DIFF
--- a/watertap/core/membrane_channel0d.py
+++ b/watertap/core/membrane_channel0d.py
@@ -45,19 +45,19 @@ class Not0Or1Error(KeyError):
 
 class _0DPropertyHelper(collections.abc.Mapping):
     """
-    Class to make obj.properties_in and obj.properties_out
-    like obj.properties from a 1D model.
+    Class to make blk.properties_in and blk.properties_out
+    like blk.properties from a 1D model.
     """
 
-    def __init__(self, obj, reverse=False):
-        self.obj = weakref.ref(obj)
+    def __init__(self, blk, reverse=False):
+        self._blk_ref = weakref.ref(blk)
         if reverse:
             self._get_property_x = self._get_property_x_backward
         else:
             self._get_property_x = self._get_property_x_forward
 
     def __len__(self):
-        return len(self.obj().properties_in) + len(self.obj().properties_out)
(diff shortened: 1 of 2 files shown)


adam-a-a:
My initial thoughts are -- what is the tradeoff in efficiency/computational time between this approach and #1476 in reality? Is the tradeoff significant enough to merit the changes here? At the surface, it just seems like these changes reduce readability of the code, especially for new devs who are trying to make sense of, say, our RO model. 

I will try to pull these changes into our branch for dynamic RO testing and dev and see what happens. First, I will check that it works after time discretization. Assuming that checks out, then it is a matter of checking computational time I suppose.

bknueven:
> My initial thoughts are -- what is the tradeoff in efficiency/computational time between this approach and #1476 in reality? Is the tradeoff significant enough to merit the changes here? At the surface, it just seems like these changes reduce readability of the code, especially for new devs who are trying to make sense of, say, our RO model.
> 
> I will try to pull these changes into our branch for dynamic RO testing and dev and see what happens. First, I will check that it works after time discretization. Assuming that checks out, then it is a matter of checking computational time I suppose.

Well, part of the complexity is I tried to anticipate other potential issues with the old method. This should quack exactly like the `properties` Pyomo object. Looking like a dictionary instead would be less code.

The new code has 100% coverage, so I would encourage you and any other reviewer to look at the tests first to see how it is used. The implementation could perhaps be a bit simpler, but there are a lot of cases to cover.

In general, I do worry "rebuild the dictionary every time you access an element" will not scale. Maybe it doesn't matter in this case.

lbianchi-lbl (approved):
I don't feel I'm qualified to comment on the logic, but purely from the Python coding point of view, this looks good.

kurbansitterley (approved):

PR #1482, opened 2024-08-20
add crystallizer flowsheet

ElmiraShamlou:
add three crystallizer flowsheets

1) MVC with crystallizer
2) Crystallizer with chiller and condenser (no heat recovery)
3) TVC with crystallizer

### Legal Acknowledgement


watertap/flowsheets/crystallization/crystallizer_MVR.py
```diff
@@ -0,0 +1,391 @@
...
+def build():
+
+    # flowsheet set up
+    m = ConcreteModel()
+    m.fs = FlowsheetBlock(dynamic=False)
+
+    # properties
+    m.fs.properties_vapor = props_w.WaterParameterBlock()
+    m.fs.properties = props.NaClParameterBlock()
+    m.fs.properties_nacl = props_nacl.NaClParameterBlock()
```
adam-a-a, line 82: Why use two separate NaCl property models? Any reason not to use NaCl with temp dependence by itself?

watertap/flowsheets/crystallization/Crystallizer_MVR.py
```diff
@@ -0,0 +1,391 @@
...
+from idaes.core import FlowsheetBlock
+from idaes.core.util.model_statistics import degrees_of_freedom
+
+import idaes.core.util.scaling as iscale
+import idaes.logger as idaeslog
+from watertap.core.solvers import get_solver
+from idaes.core import UnitModelCostingBlock
+
+from watertap.property_models.unit_specific import cryst_prop_pack as props
+from watertap.unit_models.Crystallizer_revised import Crystallization
```
adam-a-a: I am interested to see how the revised form differs from the original. For one, I would prefer to have the revised version condensed within the original, accessible through config options, assuming there is considerable overlap in code. Waiting to see this unit model file to determine that.

watertap/flowsheets/crystallization/crystallizer_MVR.py
ElmiraShamlou: The key difference between the two property packages is that the crystallizer-specific package includes additional solid and vapor phases, which are not present in the other unit models.

watertap/costing/unit_models/crystallizer.py
```diff
@@ -93,7 +93,7 @@ def build_crystallizer_cost_param_block(blk):
     )
 
     costing = blk.parent_block()
-    costing.register_flow_type("steam", blk.steam_cost)
+    #costing.register_flow_type("steam", blk.steam_cost)
```
adam-a-a: delete commented code

watertap/costing/unit_models/crystallizer.py
```diff
@@ -120,27 +120,6 @@ def cost_crystallizer(blk, cost_type=CrystallizerCostType.default):
...
-        "electricity",
-    )
-
-    blk.costing_package.cost_flow(
-        pyo.units.convert(
-            (blk.unit_model.work_mechanical[0] / _compute_steam_properties(blk)),
-            to_units=pyo.units.m**3 / pyo.units.s,
-        ),
-        "steam",
-    )
```
adam-a-a, line 143: Curious about why these lines are being deleted. I probably didn't look closely enough, but have they been moved elsewhere or completely removed? And why?

watertap/costing/unit_models/crystallizer.py
```diff
@@ -215,72 +194,3 @@ def cost_crystallizer_by_volume(blk):
         )
     )
     _cost_crystallizer_flows(blk)
-
-
-def _compute_steam_properties(blk):
```
adam-a-a, line 220: Same question here, except it makes more sense immediately as to why we wouldn't want/need this in a costing module

watertap/flowsheets/crystallization/crystallizer_MVR.py
adam-a-a: Makes sense--I should've double-checked the imports.

watertap/flowsheets/crystallization/crystallizer_MVR.py
```diff
@@ -0,0 +1,444 @@
...
+
+    # Control volume flow blocks
+    m.fs.feed = Feed(property_package= m.fs.properties_nacl)
+    m.fs.distillate = Product(property_package= m.fs.properties_vapor)
+
+
+    # unit models: steam heater, mixer, pump, crystalizer, compressor, separator
```
adam-a-a, line 88:
```suggestion
    # unit models: steam heater, mixer, pump, crystallizer, compressor, separator
```

watertap/unit_models/steam_heater_0D.py
```diff
@@ -108,8 +110,8 @@ def initialize_build(self, *args, **kwargs):
             self.hot_side_inlet.fix()
             self.cold_side_inlet.fix()
             self.hot_side_outlet.unfix()
-            self.cold_side_outlet.unfix()
-            self.area.fix()
+            #self.cold_side_outlet.unfix()
+            #self.area.fix()
```
adam-a-a: Why did you have to comment these out? Did the initialization routine cause issues when connected in a flowsheet?

watertap/unit_models/steam_heater_0D.py
```diff
@@ -139,30 +141,51 @@ def initialize_build(self, *args, **kwargs):
...
+
+            self.outlet_liquid_mass_balance.activate()
+            self.outlet_pressure_sat.activate()
 
             with idaeslog.solver_log(solve_log, idaeslog.DEBUG) as slc:
                 res = opt.solve(self, tee=slc.tee)
             init_log.info("Initialization Complete {}".format(idaeslog.condition(res)))
 
         elif self.config.mode == Mode.CONDENSER and self.config.estimate_cooling_water:
+            print(f"DOF 1: {degrees_of_freedom(self)}")
```
adam-a-a, line 164: Should probably remove. If anything, can use a logger to convey desired output

watertap/unit_models/steam_heater_0D.py
```diff
@@ -139,30 +141,51 @@ def initialize_build(self, *args, **kwargs):
...
             cold_side_outlet_temperature = self.cold_side_outlet.temperature[0].value
             self.hot_side_inlet.fix()
             self.cold_side_inlet.fix()
-            self.cold_side_outlet.unfix()
+            #self.hot_side_outlet.unfix()
+            self.cold_side_outlet.temperature.unfix()
+            # condenser mode without cooling water estimation
+            self.outlet_liquid_mass_balance.deactivate()
+            self.outlet_pressure_sat.deactivate()
+            print(f"DOF 2: {degrees_of_freedom(self)}")
```
adam-a-a, line 175: same here

watertap/unit_models/steam_heater_0D.py
```diff
@@ -177,19 +200,21 @@ def initialize_build(self, *args, **kwargs):
...
-
-            with idaeslog.solver_log(solve_log, idaeslog.DEBUG) as slc:
-                res = opt.solve(self, tee=slc.tee)
-            init_log.info(
-                "Initialization Complete (w/ cooling water estimation): {}".format(
-                    idaeslog.condition(res)
+            #for j in self.cold_side.config.property_package.solute_set:
+             #   self.cold_side.properties_in[0].mass_frac_phase_comp["Liq", j].unfix()
+              #  self.cold_side.properties_in[0].flow_mass_phase_comp["Liq", j].fix()
+            print(f"DOF 3: {degrees_of_freedom(self)}")
```
adam-a-a, line 209: same here

watertap/unit_models/crystallizer.py
```diff
@@ -186,14 +187,14 @@ def build(self):
 
         self.pressure_operating = Var(
             initialize=1e3,
-            bounds=(0.001, 1e6),
+            bounds=(0.001, 2e6),
```
adam-a-a, line 190: I wonder if we should just set the upper bound to None.

watertap/unit_models/crystallizer.py
```diff
@@ -228,15 +229,18 @@ def build(self):
         )
 
         self.height_crystallizer = Var(
-            initialize=3, bounds=(0, 25), units=pyunits.m, doc="Crystallizer height"
+            initialize=3, bounds=(0, 30), units=pyunits.m, doc="Crystallizer height"
```
adam-a-a, line 232: Appears to be a very subtle and targeted increase in the upper bound. Is this informed by typical upper bound in practice, arbitrary, or something else?

watertap/unit_models/crystallizer.py
```diff
@@ -228,15 +229,18 @@ def build(self):
...
-            units=pyunits.m**3 / pyunits.s,
-            doc="Minimum circulation flow rate through crystallizer heat exchanger",
-        )
+
+        """
+            self.magma_circulation_flow_vol = Var(
+                initialize=1,
+                bounds=(0, 100),
+                units=pyunits.m**3 / pyunits.s,
+                doc="Minimum circulation flow rate through crystallizer heat exchanger",
```
adam-a-a, line 240: I imagine we don't need this anymore since you are essentially disaggregating parts of the crystallizer that no longer need to be lumped in, but can you explain?

watertap/unit_models/Crystallizer_revised.py
```diff
@@ -0,0 +1,827 @@
...
+                    - b.properties_vapor[0].flow_mass_phase_comp["Vap", "H2O"]
+                )
+
+                return (
+                    b.relative_supersaturation[j] * b.properties_out[0].solubility_mass_frac_phase_comp["Liq", j] * total_mass_after_evap
+                    == (
+                        mass_after_evap
+                        - b.properties_out[0].solubility_mass_frac_phase_comp["Liq", j] * total_mass_after_evap
+                    ))
+"""
```
adam-a-a, line 402: Have these all been commented out?

watertap/unit_models/Crystallizer_revised.py
```diff
@@ -0,0 +1,827 @@
...
+        def eq_crystallizer_height_constraint(b):
+            # Height is max(). Manual smooth max implementation used here: max(a,b) = 0.5(a + b + |a-b|)
+            a = b.eq_vapor_space_height + b.height_slurry
+            b = b.eq_minimum_height_diameter_ratio
+            eps = 1e-20 * pyunits.m
+            return self.height_crystallizer == 0.5 * (
+                a + b + ((a - b) ** 2 + eps**2) ** 0.5
+            )
+
+            """
```
adam-a-a, line 572: I could be reading the diff incorrectly but looks like these are all commented too. Could you comment on why certain constraints/expressions are being commented out specifically to better understand the changes?

watertap/unit_models/crystallizer.py
```diff
@@ -204,15 +205,15 @@ def build(self):
...
-        self.work_mechanical = Var(
-            self.flowsheet().config.time,
-            initialize=1e5,
-            bounds=(-5e6, 5e6),
-            units=pyunits.kJ / pyunits.s,
-            doc="Crystallizer thermal energy requirement",
-        )
-
+        """
+            self.work_mechanical = Var(
```
adam-a-a, line 209: I'm surprised to see no tests fail for the crystallizer unit especially since no changes were made to the test file.

watertap/flowsheets/crystallization/crystallizer_live_steam_with_condenser_chiller.py
```diff
@@ -0,0 +1,508 @@
...
+
+def main():
+    solver = get_solver()
+    m = build()
+    set_operating_conditions(m)
+    initialize_system(m, solver=solver)
+
+    optimize_set_up(m)
+
+    #interval_initializer(m)
```
adam-a-a: I am guessing that an InfeasibleConstraint exception (or something like that) arises when using the interval initializer.

watertap/flowsheets/crystallization/crystallizer_live_steam_with_condenser_chiller.py
```diff
@@ -0,0 +1,508 @@
+#################################################################################
```
adam-a-a, line 1: Next step is to add (1) a test file for this flowsheet and (2) documentation for this flowsheet in this section: https://watertap.readthedocs.io/en/latest/technical_reference/flowsheets/index.html

watertap/flowsheets/crystallization/crystallizer_MVR.py
```diff
@@ -0,0 +1,444 @@
+#################################################################################
```
adam-a-a, line 1: Next step is to add (1) a test file for this flowsheet and (2) documentation for this flowsheet in this section: https://watertap.readthedocs.io/en/latest/technical_reference/flowsheets/index.html

watertap/flowsheets/crystallization/Crystallizer_MVR.py
```diff
@@ -0,0 +1,444 @@
...
+from watertap.unit_models.mvc.components import Compressor
+from watertap.property_models.unit_specific import cryst_prop_pack as props
+import watertap.property_models.water_prop_pack as props_w
+import watertap.property_models.NaCl_T_dep_prop_pack as props_nacl
+from pyomo.common.log import LoggingIntercept
+from idaes.models.unit_models.translator import Translator
+from idaes.models.unit_models.separator import SplittingType
+import logging
+#from watertap.core.util.model_debug_mode import activate
+#activate()
```
adam-a-a: delete

adam-a-a (commented):
Here are some initial comments from a first pass

watertap/flowsheets/crystallization/Crystallizer_MVR.py
adam-a-a: I suppose you can get rid of Crystallizer_revised and just use the revised original now, right?

watertap/costing/unit_models/crystallizer.py
ElmiraShamlou: @adam-a-a They are removed as we integrate the steam heater, pump, and other essential components into the full FC crystallizer flowsheet. Therefore, the flow rates of the steam and recycle streams, along with their associated costs, are incorporated within the cost calculations of the steam heater and pump unit models.

watertap/unit_models/crystallizer.py
ElmiraShamlou: The tests are failing, and I'm revising them to align with the recent changes.

watertap/costing/unit_models/crystallizer.py
OOAmusat: @ElmiraShamlou but removing this hinders whoever wants to use the combined unit we had previously.

Maybe what we need is a configuration argument that determines the mode the user wants (broken down or combined), and then adds the necessary equations for each case?

ksbeattie:
This is waiting on NAWI funds for work to continue.

ksbeattie:
@ElmiraShamlou and this one too?

kurbansitterley:
I made these changes initially because the flowsheet wasn't working and I have a student that wants to use it. But also my changes match our conventions.

I reverted @adam-a-a 's modifications for steam costing where `steam_cost` was on the main costing block (e.g., `m.fs.costing`). I don't think we want to cost steam this way. Instead, models that use `steam` have their own `build_steam_cost_param_block` that results in the steam cost being at `m.fs.costing.steam.cost`. This matches how we cost materials for other units (though I just checked and it appears electroNP uses the old convention).

In making these changes, I discovered that we have two models that use a steam cost with different units: `heat_exchanger` uses $/kg but `surrogate_crystallizer` uses $/m3. Right now, if these two units were on the same flowsheet, one or the other would run into errors. This could be fixed by either changing the expression used to cost the steam flow or have two different steam costs e.g., `steam_per_m3` and `steam_per_kg`. Note that this would still have been the case if I hadn't made these changes.

Crystallizer tests are failing but that was the case prior to my changes.

watertap/costing/unit_models/crystallizer.py
kurbansitterley: this should really be part of the property package

adam-a-a:
> I made these changes initially because the flowsheet wasn't working and I have a student that wants to use it. But also my changes match our conventions.
> 
> I reverted @adam-a-a 's modifications for steam costing where `steam_cost` was on the main costing block (e.g., `m.fs.costing`). I don't think we want to cost steam this way. Instead, models that use `steam` have their own `build_steam_cost_param_block` that results in the steam cost being at `m.fs.costing.steam.cost`. This matches how we cost materials for other units (though I just checked and it appears electroNP uses the old convention).
> 
> In making these changes, I discovered that we have two models that use a steam cost with different units: `heat_exchanger` uses $/kg but `surrogate_crystallizer` uses $/m3. Right now, if these two units were on the same flowsheet, one or the other would run into errors. This could be fixed by either changing the expression used to cost the steam flow or have two different steam costs e.g., `steam_per_m3` and `steam_per_kg`. Note that this would still have been the case if I hadn't made these changes.
> 
> Crystallizer tests are failing but that was the case prior to my changes.

@kurbansitterley my main question here is what will currently happen when more than one unit on the flowsheet has its own `build_steam_cost_param_block`? What protects the user from unintentionally having different steam costs for different units when there should be a single, global steam cost (not considering the case where the user might want more than one steam cost on a flowsheet)?

kurbansitterley:
> > I made these changes initially because the flowsheet wasn't working and I have a student that wants to use it. But also my changes match our conventions.
> > I reverted @adam-a-a 's modifications for steam costing where `steam_cost` was on the main costing block (e.g., `m.fs.costing`). I don't think we want to cost steam this way. Instead, models that use `steam` have their own `build_steam_cost_param_block` that results in the steam cost being at `m.fs.costing.steam.cost`. This matches how we cost materials for other units (though I just checked and it appears electroNP uses the old convention).
> > In making these changes, I discovered that we have two models that use a steam cost with different units: `heat_exchanger` uses $/kg but `surrogate_crystallizer` uses $/m3. Right now, if these two units were on the same flowsheet, one or the other would run into errors. This could be fixed by either changing the expression used to cost the steam flow or have two different steam costs e.g., `steam_per_m3` and `steam_per_kg`. Note that this would still have been the case if I hadn't made these changes.
> > Crystallizer tests are failing but that was the case prior to my changes.
> 
> @kurbansitterley my main question here is what will currently happen when more than one unit on the flowsheet has its own `build_steam_cost_param_block`? What protects the user from unintentionally having different steam costs for different units when there should be a single, global steam cost (not considering the case where the user might want more than one steam cost on a flowsheet)?

My understanding is that `register_costing_parameter_block` will make this check and only create a new block if one doesn't exist for `steam`. I believe a critical assumption here is that the material is spelled identically. i.e., `nacl` and `NaCl` would result in two parameter blocks for NaCl. In the scenario where two units both use `steam` with two different costs, the cost used in the costing package would be that of the first unit costing model instantiated. Of course this could be modified at the flowsheet level with e.g., `m.fs.costing.steam.cost.fix(value)`

kurbansitterley:
@ElmiraShamlou added tests for the new flowsheets; only testing `main`

ksbeattie:
@ElmiraShamlou is this still going to happen?

ElmiraShamlou:
> @ElmiraShamlou is this still going to happen?

Yes, I’m picking this back up.

PR #1483, opened 2024-08-20
Update GHA runner and Python versions

lbianchi-lbl:
## Summary/Motivation

- Match Python versions currently supported by IDAES (3.9 through 3.12)
- Update GHA runner versions to more recent ones (e.g. `ubuntu-2004` -> `ubuntu-2204`)
- Add `macos` jobs using `macos-14` runners with Apple Silicon
- Remove "experimental" macOS job using non-IDAES solvers

### Legal Acknowledgement


lbianchi-lbl:
- As @bknueven points out, Python 3.12 has a non-backward-compatible [change](https://github.com/python/cpython/issues/100425) in the built-in Python `sum()` function that might be causing the Python 3.12 jobs to fail consistently across platforms
- macOS (with the new Apple Silicon runners) seems to be failing consistently across all Python versions. Most if not all failures come from the BSM2 model

ksbeattie:
Waiting for #1469

bknueven:
Let's fix the easy things to fix and mark the remainder as xfail and return to them after the next release.

bknueven:
@lbianchi-lbl I believe I got all the "false positives" in the PR which are easy to fix. I would suggest we mark the remaining as x-fail strict. The remaining models which do not pass are under active development / tweaking, so we should know if they start passing due to some change.

lbianchi-lbl:
This is on hold and will need to be reworked after #1562 has been merged in (the changesets are most likely incompatible).

ksbeattie:
We should get the changes in from this PR that we know work (like the python version updates) for the Sept release and push to the Dec release the things holding this PR up.

ksbeattie:
Resolved conflicts by preferring changes made on main.  Might still need to look into exactly what gets run on macos.

ksbeattie:
Closing this in preference to the above PRs.

PR #1485, opened 2024-09-03
Update Costing in Dye Desal Flowsheet

MarcusHolly:
## Fixes/Resolves:

Updates the dye desal flowsheet in preparation for a report

- Addresses how costing is handled for certain sets of configuration options
- Corrects minor typos in the flowsheet

### Legal Acknowledgement


watertap/flowsheets/dye_desalination/dye_desalination.py
```diff
@@ -716,11 +716,21 @@ def add_costing(m):
             ),
             doc="Specific energy consumption of the treatment train on a feed flowrate basis [kWh/m3]",
         )
-    else:
+    elif hasattr(m.fs, "dewaterer") or hasattr(m.fs, "gac"):
+        m.fs.ro_costing.electricity_cost = value(m.fs.zo_costing.electricity_cost)
+        m.fs.ro_costing.base_currency = pyunits.USD_2020
```
adam-a-a: Why do these lines fall under this conditional that checks for dewatering unit or gac? Setting electricity cost and base currency seem like something that should happen so long as m.fs.ro_costing exists. Maybe there is more to the logic here that I am missing.

MarcusHolly: Yes, you are correct

watertap/flowsheets/dye_desalination/dye_desalination.py
```diff
@@ -716,27 +725,51 @@ def add_costing(m):
...
-                * (
-                    m.fs.zo_costing.waste_disposal_cost
-                    * pyunits.convert(
-                        m.fs.brine.properties[0].flow_vol,
-                        to_units=pyunits.m**3 / m.fs.zo_costing.base_period,
+        if brine_revenue:
+            m.fs.brine_cost = Expression(
+                expr=(
+                    -1
+                    * m.fs.zo_costing.utilization_factor
```
luohezhiming, line 753: Is the zero order costing utilization factor same as ro.costing? Can you tell me some background why we need to use zero order costing in this flowsheet?

MarcusHolly: Yes, both utilization factors are set to 0.85. Zero order costing is needed to cost the ZO nanofiltration and pump models The `dye_desalination_global_costing.yaml` file is also linked to the zero order costing and provides values like `dye_disposal_cost`, `recovered_water_cost`, `electricity_cost`, etc.

adam-a-a:
Just as a reminder for both of us--should double-check capital recovery factor values for both ZO and "detailed" costing packages and ensure that they are equivalent and line up with what was used for analyses.

MarcusHolly:
> Just as a reminder for both of us--should double-check capital recovery factor values for both ZO and "detailed" costing packages and ensure that they are equivalent and line up with what was used for analyses.

I've set the capital recovery factor for ro_costing to 0.065, which is the value used by the zo_costing and our analyses

luohezhiming (approved):

watertap/flowsheets/dye_desalination/dye_desalination.py
```diff
@@ -627,6 +635,13 @@ def add_costing(m):
 
     m.fs.zo_costing = ZeroOrderCosting(case_study_definition=source_file)
     m.fs.ro_costing = WaterTAPCosting()
+    m.fs.ro_costing.electricity_cost = value(m.fs.zo_costing.electricity_cost)
+    m.fs.ro_costing.base_currency = pyunits.USD_2020
+    m.fs.ro_costing.utilization_factor.fix(0.85)
+    # Assume the same capital recovery factor as zo_costing
+    m.fs.ro_costing.capital_recovery_factor.fix(0.065051435)
```
adam-a-a: @MarcusHolly any reason why we don't just point to the variable itself to get the value instead of hard-coding here?

MarcusHolly: Nope - I'll update this

watertap/flowsheets/dye_desalination/dye_desalination.py
```diff
@@ -627,6 +635,14 @@ def add_costing(m):
 
     m.fs.zo_costing = ZeroOrderCosting(case_study_definition=source_file)
     m.fs.ro_costing = WaterTAPCosting()
+    m.fs.ro_costing.electricity_cost = value(m.fs.zo_costing.electricity_cost)
+    m.fs.ro_costing.base_currency = pyunits.USD_2020
+    m.fs.ro_costing.utilization_factor.fix(0.85)
+    # Assume the same capital recovery factor as zo_costing
+    zo_crf = m.fs.zo_costing.capital_recovery_factor
+    m.fs.ro_costing.capital_recovery_factor.fix(zo_crf)
```
adam-a-a:
```suggestion
    m.fs.ro_costing.capital_recovery_factor.fix(value(zo_crf))
```

watertap/flowsheets/dye_desalination/dye_desalination.py
```diff
@@ -895,29 +970,57 @@ def total_externalities(b):
...
+            if dye_revenue and brine_revenue:
+                return pyunits.convert(
+                    0 * pyunits.USD_2020 / pyunits.year,
+                    to_units=pyunits.USD_2020 / pyunits.year,
+                )
+            elif dye_revenue:
+                if value(m.fs.brine_cost) > 0:
+                    return pyunits.convert(
+                        m.fs.brine_cost,
+                        to_units=pyunits.USD_2020 / pyunits.year,
```
adam-a-a, line 1001: Something to keep in mind for subsequent PR and report modifications: would be good to move away from 2020 currency since it's more of an anomaly.

MarcusHolly: @adam-a-a So what should be used instead?

adam-a-a (approved):

watertap/flowsheets/dye_desalination/dye_desalination.py
adam-a-a: Could go with 2023

PR #1487, opened 2024-09-10
MCAS w/generic crystallizer

adam-a-a:
## Summary/Motivation:
Building on @OOAmusat 's work on a generic crystallizer by making it compatible with MCAS and also revising to match more closely with IDAES/WaterTAP standards. 

- add enthalpy and pressure_sat relationships for seawater into MCAS
- remove unnecessary inclusions related to properties or ion sets from crystallizer model and point to prop pack
- general code cleanup
- update mcas documentation

### Legal Acknowledgement


watertap/unit_models/tests/test_crystallizer_surrogate.py
```diff
@@ -0,0 +1,236 @@
...
+)
+from watertap.unit_models.surrogate_crystallizer import SurrogateCrystallizer
+
+from idaes.core import UnitModelCostingBlock
+from watertap.costing import WaterTAPCosting
+
+__author__ = "Oluwamayowa Amusat, Adam Atia"
+
+
+def add_crystallizer_rbf_model(
```
adam-a-a: I am thinking about bringing this back as a method of the unit model...

watertap/unit_models/tests/test_crystallizer_surrogate.py
```diff
@@ -0,0 +1,236 @@
...
+    #     add_crystallizer_nn_model(
+    #         m.fs.cryst,
+    #         surrogate_inputs_with_bounds=surrogate_inputs,
+    #         surrogate_outputs=surrogate_outputs,
+    #         path_to_onnx_model=r"C:\Users\OOAmusat-II\Desktop\NAWI\WATERTAP\Phreeqc\crystallization\NN_40_NL_3_sigmoid_WD_1e-06_LR_0.01_testing.onnx",
+    #         path_to_offsetscaler_json=r"C:\Users\OOAmusat-II\Desktop\NAWI\WATERTAP\Phreeqc\crystallization\scalingfile_testing.json",
+    #         surrogate_to_flowsheet_basis_ratio=1,
+    #     )
+    # else:
+    #     raise ValueError('"surrogate_model_form" must be "RBF" or "NN"')
```
adam-a-a: Can probably test the NN surrogate form as well but just trying rbf first.

watertap/unit_models/surrogate_crystallizer.py
OOAmusat, line 1: This looks good. The main changes seem to be in the way the ion list and vapor property package are handled.

The only comment I have is with your change to the mass balance constraint (``eq_mass_balance_constraints(b, j)``). You don't fix ``self.S['H2O'].fix(0.0)``, but instead, eliminate it from the water mass balance. Do you think the model will be more stable with that change?

watertap/unit_models/tests/test_crystallizer_surrogate.py
OOAmusat: I agree.

watertap/unit_models/tests/test_crystallizer_surrogate.py
OOAmusat: The question is whether we want to do that, or allow the user to define their own function entirely. 

I don't think it should be in the unit model file. However, we could support rbfs and maybe even NN models as defaults. We could have the code in a different file that the user can load as the default with optional arguments.

adam-a-a: That was what I was thinking--something along the lines of calling the method on the unit model to build some surrogate(s). And defaults could sit in our `data` folder with a new directory for `default_surrogates` or something like that. Either way, I don't love the idea of it living in this test file or in some future example flowsheet file.

watertap/unit_models/surrogate_crystallizer.py
adam-a-a: Consider that this was fixed to zero in the first place here, and that we only really need to index `S` by the ion list, I removed `S[j]` from the water mass balance.

kurbansitterley:
@adam-a-a are you only planning on making the surrogate compatible with MCAS or also the physical model?

watertap/unit_models/tests/test_crystallizer_surrogate.py
adam-a-a: I am going to exclude this for now since it might not be a priority given latest developments.

adam-a-a:
@OOAmusat I think this is ready to go. @ElmiraShamlou @carson-tucker @avdudchenko this could use a review from at least one of you so that we can merge this and subsequently bring in Mayo's additional changes

ElmiraShamlou:
> @OOAmusat I think this is ready to go. @ElmiraShamlou @carson-tucker @avdudchenko this could use a review from at least one of you so that we can merge this and subsequently bring in Mayo's additional changes

I would suggest adding saturation vapor pressure, as it directly impacts the system’s energy consumption. It is required for any Crystallizer heat recovery analysis.

adam-a-a:
> @adam-a-a are you only planning on making the surrogate compatible with MCAS or also the physical model?

@kurbansitterley for the moment we were just doing this for the surrogate crystallizer, but we can think about doing so for the physical model. I am guessing that accounting for solids in MCAS would up the ante.

adam-a-a:
> saturation vapor pressure

This sounds like something that we can address in a subsequent PR. This PR just focuses on making MCAS compatible with the surrogate crystallizer formulation. As you can see, one of the surrogates incorporated in the test case predicts vapor pressure, but it is treated as a custom function that is separate from the unit model. Thus, it would be on the user to add such surrogates (at this point in time), so I presume that we could do the same with saturation vapor pressure.

adam-a-a:
> > saturation vapor pressure
> 
> This sounds like something that we can address in a subsequent PR. This PR just focuses on making MCAS compatible with the surrogate crystallizer formulation. As you can see, one of the surrogates incorporated in the test case predicts vapor pressure, but it is treated as a custom function that is separate from the unit model. Thus, it would be on the user to add such surrogates (at this point in time), so I presume that we could do the same with saturation vapor pressure.

Unless you were implying that we should incorporate `pressure_sat` from the seawater property model, similar to how I brought in enthalpy @ElmiraShamlou

ElmiraShamlou:
> > > saturation vapor pressure
> > 
> > 
> > This sounds like something that we can address in a subsequent PR. This PR just focuses on making MCAS compatible with the surrogate crystallizer formulation. As you can see, one of the surrogates incorporated in the test case predicts vapor pressure, but it is treated as a custom function that is separate from the unit model. Thus, it would be on the user to add such surrogates (at this point in time), so I presume that we could do the same with saturation vapor pressure.
> 
> Unless you were implying that we should incorporate `pressure_sat` from the seawater property model, similar to how I brought in enthalpy @ElmiraShamlou

@adam-a-a Yes, I meant to incorporate the saturation vapor pressure into the Mcas, just as you did with the enthalpy. The test_surrogate_crystallizer.py links the crystallizer operating pressure (saturation pressure) to the surrogate vapor pressure output. However, if we need to use Mcas without access to the surrogate output, defining a property like pressure_sat ensures the model has access to it. This will expand the use of Mcas to other unit_models as well.

watertap/unit_models/tests/test_surrogate_crystallizer.py
```diff
@@ -0,0 +1,229 @@
...
+        "Vapor_Pressure",
+        "Calcite_g",
+        "Anhydrite_g",
+        "Glauberite_g",
+        "Halite_g",
+    ]
+    surrogate_directory = os.path.dirname(os.path.abspath(__file__))
+    surrogate_directory = os.path.join(
+        surrogate_directory, "surrogate_crystallizer_defaults"
+    )
```
ElmiraShamlou:
```suggestion
    surrogate_directory = Path(this_file_dir())
```

watertap/unit_models/tests/test_surrogate_crystallizer.py
```diff
@@ -0,0 +1,229 @@
...
+
+    for sm in range(0, len(filename)):
+        block_name = "crystallizer_surrogate" + "_" + filename[sm]
+        blk.add_component(block_name, SurrogateBlock(concrete=True))
+
+        surrogate_name = f"{filename[sm]}.json"
+        current_surrogate_filename = os.path.join(surrogate_directory, surrogate_name)
```
ElmiraShamlou:
```suggestion
        current_surrogate_filename =  str(surrogate_directory / surrogate_name)
```

watertap/unit_models/tests/test_surrogate_crystallizer.py
```diff
@@ -0,0 +1,229 @@
...
+        current_surrogate = PysmoSurrogate.load_from_file(
+            os.path.normpath(current_surrogate_filename)
+        )
```
ElmiraShamlou:
```suggestion
        current_surrogate = PysmoSurrogate.load_from_file(current_surrogate_filename)
```

watertap/unit_models/tests/test_surrogate_crystallizer.py
```diff
@@ -0,0 +1,229 @@
...
+from idaes.core.surrogate.surrogate_block import SurrogateBlock
+from idaes.core.surrogate.pysmo_surrogate import (
+    PysmoSurrogate,
+)
+from watertap.unit_models.surrogate_crystallizer import SurrogateCrystallizer
+
+from idaes.core import UnitModelCostingBlock
+from watertap.costing import WaterTAPCosting
+from watertap.core.solvers import get_solver
+
```
ElmiraShamlou, line 41:
```suggestion
from pyomo.common.fileutils import this_file_dir
from pathlib import Path
```

watertap/costing/unit_models/surrogate_crystallizer.py
```diff
@@ -0,0 +1,189 @@
...
+        initialize=0.7,
+        units=pyo.units.dimensionless,
+        doc="Crystallizer pump efficiency - assumed",
+    )
+
+    blk.pump_head_height = pyo.Var(
+        initialize=1,
+        units=pyo.units.m,
+        doc="Crystallizer pump head height -  assumed, unvalidated",
+    )
```
ElmiraShamlou, line 38: Just a note for everyone's awareness: The crystallizer cost for steam (including steam properties) and the pump is being removed in PR #1482 . These costs are now broken down into the steam heater (which uses the water vapor property package) and the pump unit model costs. This means steam and pump flows are calculated at the flowsheet level, with their respective costs corresponding to the unit models.

watertap/unit_models/tests/test_surrogate_crystallizer.py
ElmiraShamlou: This change might address the failing checks.

adam-a-a: I'm not sure it will since the checks fail for user mode, if the user installed through pip. I have moved the surrogates to the data folder and am trying another way.

watertap/unit_models/tests/test_surrogate_crystallizer.py
adam-a-a: The os.path.join approach essentially does this, except it accounts for the platform (e.g., windows,linux) and includes the appropriate separator (forward slash, double backslash).

watertap/unit_models/tests/test_surrogate_crystallizer.py
adam-a-a: This was my previous attempt to deal with the path issues between platforms. Replacing this now

watertap/unit_models/tests/test_surrogate_crystallizer.py
adam-a-a: Oh nice-- I didn't know about this utility! Thanks! I may not use it if my latest changes work but will try the next time I need to point to a path.

watertap/costing/unit_models/surrogate_crystallizer.py
adam-a-a: Good to know--will leave here in this PR and let your PR take care of this.

adam-a-a:
@ElmiraShamlou I can add pressure_sat to MCAS. In that case though, it sounds like we should relate pressure_sat to pressure_operating in the surrogate_crystallizer.

adam-a-a:
I suppose in this case, pressure_operating relates to the surrogate output, so perhaps not. I'd welcome feedback though. Maybe it is enough to just add pressure_sat to MCAS.

ElmiraShamlou:
> @ElmiraShamlou I can add pressure_sat to MCAS. In that case though, it sounds like we should relate pressure_sat to pressure_operating in the surrogate_crystallizer.

@adam-a-a Yes, this is necessary when the surrogate vapor pressure output is unavailable for a generic surrogate. It can be added as needed, using a custom constraint (like blk.pressure_conversion_constraints in the test file), which ties the operating pressure to the saturation pressure. Alternatively, this could be implemented as a config option, with the property package’s saturation vapor pressure set as the default when the surrogate option is not selected.

ElmiraShamlou:
> I suppose in this case, pressure_operating relates to the surrogate output, so perhaps not. I'd welcome feedback though. Maybe it is enough to just add pressure_sat to MCAS.

Yes, it is not required for this specific PR.

adam-a-a:
@ElmiraShamlou That was a good point to bring up though. @OOAmusat is going to put up a PR that adds the reaktoro capability in place of surrogates. I was about to revise the surrogate_crystallizer config options but remembered that we decided to wait on making more changes until after Mayo's upcoming PR. From there, I think we can arrive at a revision that would accommodate the different forms this model can take (e.g., using prop model pressure sat, using surrogate, using reaktoro block, etc.).

OOAmusat (approved):

ElmiraShamlou (approved):
Great job!

PR #1489, opened 2024-09-12
BSM2-P Effluent Metrics

MarcusHolly:
## Summary/Motivation:
Adds effluent concentration violations to the BSM2-P flowsheet

- Adds effluent metrics to the modified ASM2d property package
- Adds effluent violation constraints to the BSM2-P flowsheet
- Finishes To-Do item in BSM2 flowsheet related to effluent constraints

### Legal Acknowledgement


watertap/property_models/unit_specific/activated_sludge/modified_asm2d_properties.py
```diff
@@ -442,10 +522,90 @@ def _TSS(self):
         )
 
         def rule_TSS(b):
-            return b.TSS == b.VSS + b.ISS
+            return b.TSS == b.conc_mass_comp["X_H"]
```
adam-a-a: Are you sure about this change? Ideally, the test would show no change between this and previous.

watertap/property_models/unit_specific/activated_sludge/modified_asm2d_properties.py
```diff
@@ -210,6 +210,86 @@ def build(self):
             doc="ISS fractional content of biomass",
         )
 
+        # EQI parameters
```
adam-a-a: Can you add some general refs so we know where to look for these in future?

watertap/property_models/unit_specific/activated_sludge/modified_asm2d_properties.py
```diff
@@ -442,10 +522,90 @@ def _TSS(self):
...
+            ] + b.params.i_NBM * (
+                b.conc_mass_comp["X_H"]
+                + b.conc_mass_comp["X_PAO"]
+                + b.conc_mass_comp["X_AUT"]
+            )
+
+        self.eq_SNKj = pyo.Constraint(rule=rule_SNKj)
+
+        def rule_BOD5(b):
+            return b.BOD5 == 0.25 * (
```
adam-a-a: From memory, I recall that the factor (0.25) could change depending on whether this is calculated for influent or effluent. Can you double-check this? I handled this in some way for ASM1 (though it might not be ideal approach).

adam-a-a: If I am recalling correctly, I would at least opt for using a mutable Param instead of the hard-coded 0.25. Then it would be on the user to change the Param value as needed at the flowsheet level.

adam-a-a (commented):
My initial response to this PR before dropping any suggestions was "WOW--that was fast!"

Great work--I think some additional ASM2d testing on the new metrics, just as a sanity check, would be good. Most of all though, fingers crossed that the flowsheet can still solve.

watertap/property_models/unit_specific/activated_sludge/modified_asm2d_properties.py
adam-a-a: Also, if we are confident about using mass concentration for X_H directly, I would opt for connecting this to TSS using `add_objective_reference` instead of creating a constraint.

MarcusHolly: Sorry - I meant to undo this change. I had started to change it, but then realized TSS wasn't in the c-code. Also, I didn't bother looking at the tests yet since I knew I left out the optimization bit.

watertap/property_models/unit_specific/activated_sludge/modified_asm2d_properties.py
MarcusHolly: I am familiar with what you are talking about, but the c-code and the literature reference both hard-code 0.25, so I didn't see a need to do what you did in ASM1
image

adam-a-a: The problem here is if I want to apply to the influent, that value changes, right? Since this can be constructed on any state block and not only one that corresponds to treated effluent, we cannot just hard-code the value to what would correspond to the effluent.

adam-a-a: Can you show the same snapshot for BOD5i or whatever corresponds to influent in that file?

MarcusHolly: Ah yes, the influent value is 0.65
image

watertap/flowsheets/full_water_resource_recovery_facility/BSM2_P_extension.py
```diff
@@ -552,6 +558,14 @@ def scale_variables(m):
     scale_variables(m)
     iscale.calculate_scaling_factors(m)
 
+    # Touch treated properties
+    m.fs.Treated.properties[0].TSS
+    m.fs.Treated.properties[0].COD
+    m.fs.Treated.properties[0].BOD5
+    m.fs.Treated.properties[0].SNKj
+    m.fs.Treated.properties[0].SP_organic
+    m.fs.Treated.properties[0].SP_inorganic
```
adam-a-a: You should touch these before the solve.

MarcusHolly: These are touched before the solve. I placed it at the end of `set_operating_conditions`. Do you mean these should be touched right before the solve? How would that functionally be any different than the current placement?

adam-a-a: My mistake- I misread the GH diff and it looked as if these lines were after the solve. That said, I wonder if it makes more sense to touch before applying scaling. Right now, I think you made each an expression so it might not matter. If they were reverted back to vars, then it would matter.

watertap/flowsheets/full_water_resource_recovery_facility/BSM2.py
```diff
@@ -526,35 +526,34 @@ def setup_optimization(m, reactor_volume_equalities=False):
     add_effluent_violations(m)
 
 
-def add_effluent_violations(m):
-    # TODO: update "m" to blk; change ref to m.fs.Treated instead of CL1 effluent
```
MarcusHolly, line 530: I haven't been able to figure out why, but this change is breaking the BSM2 GUI, so I've reverted it in #1491

MarcusHolly:
Replaced by #1503

PR #1492, opened 2024-09-18
BSM2-P Effluent Metrics w/o Flowsheet Constraints

MarcusHolly:
## Fixes/Resolves:

This PR is similar to https://github.com/watertap-org/watertap/pull/1489, but it does not attempt to implement the effluent violation constraints in the BSM2_P_extension flowsheet - this will be handled in a separate PR. The constraints related to the effluent metrics are derived from the work of Flores-Alsina (specifically perf_plant_ss.m).

## Summary/Motivation:
Define the effluent metrics for the BSM2_P_extension flowsheet

- Adds effluent metrics to the modified ASM2d property package
- Updates the BSM2 GUI flowsheet images
- Updates unit model tests affected by the changes made to the modified ASM2d prop pack

### Legal Acknowledgement


watertap/property_models/unit_specific/activated_sludge/tests/test_modified_asm2d_thermo.py
```diff
@@ -204,13 +264,13 @@ def test_build(self, model):
...
-                        "Property {v_name} is an on-demand property, but was found "
-                        "on the stateblock without being demanded".format(v_name=v.name)
-                    )
+        # for v in metadata.list_supported_properties():
+        #     if metadata[v.name].method is not None:
+        #         if model.props[1].is_property_constructed(v.name):
+        #             raise PropertyAttributeError(
+        #                 "Property {v_name} is an on-demand property, but was found "
+        #                 "on the stateblock without being demanded".format(v_name=v.name)
+        #             )
```
adam-a-a: Why comment these out?

MarcusHolly: I think this test was written to check that the constraints for VSS, ISS, TSS, COD, etc. were only built on-demand, but I've changed them from constraints to expressions, so I don't think this test applies anymore. On that note, I don't think I need to be touching the effluent properties in the flowsheet anymore either.

watertap/flowsheets/full_water_resource_recovery_facility/BSM2_P_extension.py
```diff
@@ -872,9 +872,82 @@ def display_performance_metrics(m):
...
+        pyo.units.get_units(m.fs.FeedWater.properties[0].TSS),
+    )
+    print(
+        "Feed COD concentration",
+        pyo.value(m.fs.FeedWater.properties[0].COD),
+        pyo.units.get_units(m.fs.FeedWater.properties[0].COD),
+    )
+    print(
+        "BOD5 concentration",
+        pyo.value(m.fs.FeedWater.properties[0].BOD5["effluent"]),
```
adam-a-a: Should the index be influent instead of effluent?

watertap/property_models/unit_specific/activated_sludge/modified_asm2d_properties.py
```diff
@@ -512,4 +675,28 @@ def calculate_scaling_factors(self):
...
+            if iscale.get_scaling_factor(self.SNOX) is None:
+                iscale.set_scaling_factor(self.SNOX, 1e3)
+
+        if self.is_property_constructed("SP_organic"):
+            if iscale.get_scaling_factor(self.SP_organic) is None:
+                iscale.set_scaling_factor(self.SP_organic, 1e2)
+
+        if self.is_property_constructed("SP_inorganic"):
+            if iscale.get_scaling_factor(self.SP_inorganic) is None:
+                iscale.set_scaling_factor(self.SP_inorganic, 1e3)
```
adam-a-a: Do we need to scale the expressions? Have you tried without this added scaling?

watertap/flowsheets/full_water_resource_recovery_facility/BSM2_P_extension.py
MarcusHolly: Yes - although technically it should be "raw". I honestly forgot I even printed the influent metrics already, but this is just something I added without putting too much thought in. I should verify that the equations for influent really are identical to the equations for effluents. We know, at the very least, that the equation for BOD5 differs slightly, which is why we have this config option implemented

watertap/property_models/unit_specific/activated_sludge/modified_asm2d_properties.py
MarcusHolly: Just tried removing scaling in the latest commit - it still fails

docs/technical_reference/flowsheets/extended_BSM2.rst
```diff
@@ -204,6 +207,11 @@ Additional Constraints
    "Reactor 5 mass transfer", ":math:`injection_{R5, S_{O2}} = KLa_{R5} * V_{R5} * (S_{O, eq} - S_{O, out})`"
    "Reactor 6 mass transfer", ":math:`injection_{R6, S_{O2}} = KLa_{R6} * V_{R6} * (S_{O, eq} - S_{O, out})`"
    "Reactor 7 mass transfer", ":math:`injection_{R7, S_{O2}} = KLa_{R7} * V_{R7} * (S_{O, eq} - S_{O, out})`"
+   "Total suspended solids concentration", ":math:`TSS_{out} <= TSS_{max}`"
+   "Chemical oxygen demand", ":math:`COD_{out} <= COD_{max}`"
+   "Total phosphorus concentration", ":math:`P_{out} <= P_{max}`"
+   "Total nitrogen concentration", ":math:`N_{out} <= N_{max}`"
+   "5-day biological oxygen demand", ":math:`BOD5_{out} <= BOD5_{max}`"
```
adam-a-a: Should this be removed until we are able to incorporate as constraints?

MarcusHolly: Sure - I'll get rid of the change in this PR and leave it in the follow-up PR (1503)

watertap/property_models/unit_specific/activated_sludge/tests/test_modified_asm2d_thermo.py
```diff
@@ -203,24 +262,6 @@ def test_build(self, model):
...
-                        "on the stateblock without being demanded".format(v_name=v.name)
-                    )
-
-        # check that properties are built if demanded
-        for v in metadata.list_supported_properties():
-            if metadata[v.name].method is not None:
-                if not hasattr(model.props[1], v.name):
-                    raise PropertyAttributeError(
-                        "Property {v_name} is an on-demand property, but was not built "
-                        "when demanded".format(v_name=v.name)
```
adam-a-a, line 221: can you remind me why this has to be removed?

MarcusHolly: Theses tests were originally written to check that the constraints for VSS, ISS, and TSS (which were present before this PR) were only built on-demand, but I've changed them from constraints to expressions, so I think these tests don't apply anymore. If we were to keep them, the test that checks that properties are not built if not demanded would fail

adam-a-a (approved):
Nice work! I am curious how scaling became an issue after including expressions. I wonder if it had something to do with taking away variables/constraints, which had scaling, and replacing with the expressions. In effect, model scaling was slightly altered to a less stable form.

watertap/property_models/unit_specific/activated_sludge/modified_asm2d_properties.py
```diff
@@ -383,68 +486,127 @@ def build(self):
...
+                * (
+                    self.conc_mass_comp["X_H"]
+                    + self.conc_mass_comp["X_PAO"]
+                    + self.conc_mass_comp["X_AUT"]
+                )
+            )
+            return snkj
 
-        self.eq_ISS = pyo.Constraint(rule=rule_ISS)
+        self.SNKj = pyo.Expression(rule=_SNKj, doc="Kjeldahl nitrogen")
```
luohezhiming: I am curious why SNKj is short for "Kjeldahl nitrogen", I remember generally we called "total Kjeldahl nitrogen" as TKN

watertap/flowsheets/full_water_resource_recovery_facility/BSM2_P_extension.py
```diff
@@ -872,9 +876,82 @@ def display_performance_metrics(m):
...
+        pyo.units.get_units(m.fs.FeedWater.properties[0].BOD5["raw"]),
+    )
+    print(
+        "SNKj concentration",
+        pyo.value(m.fs.FeedWater.properties[0].SNKj),
+        pyo.units.get_units(m.fs.FeedWater.properties[0].SNKj),
+    )
+    print(
+        "SNOX concentration",
+        pyo.value(m.fs.Treated.properties[0].SNOX),
```
luohezhiming:
```suggestion
        pyo.value(m.fs.FeedWater.properties[0].SNOX),
```

watertap/flowsheets/full_water_resource_recovery_facility/BSM2_P_extension.py
```diff
@@ -872,9 +876,82 @@ def display_performance_metrics(m):
...
+        pyo.units.get_units(m.fs.Treated.properties[0].SNOX),
```
luohezhiming:
```suggestion
        pyo.units.get_units(m.fs.FeedWater.properties[0].SNOX),
```

luohezhiming (changes requested):
Some minor change needs to be addressed and a few comments

watertap/property_models/unit_specific/activated_sludge/modified_asm2d_properties.py
MarcusHolly: The Flores-Alsina files are inconsistent when naming this variable. Sometimes they simply refer to it as Kjeldahl nitrogen and other times they refer to it as total Kjeldahl nitrogen. I opted to use the former, but I agree with you that we should just refer to this as TKN, even though Flores-Alsina always denotes it as SNKj

luohezhiming (approved):

PR #1493, opened 2024-09-21
Add steam properties to water and crystallizer property package

kurbansitterley:
## Fixes/Resolves:

Specific volume of heating steam for crystallizer is calculated externally in costing package currently. This PR moves that to the water and crystallizer property packages. It also adds the saturation temperature of vapor (``temperature_sat_solvent``) to the water prop pack.

Also, three properties (```solubility_mass_phase_comp```,  ```solubility_mass_frac_phase_comp```, and ```dh_crystallization_mass_comp``` are raising annoying IDAES warnings because they are not standard property names but their metadata is in ```add_properties()```. This PR moves them to ```define_custom_properties()```.

## Summary/Motivation:

Add needed steam property to prop packs and move property metadata to eliminate IDAES warning generation.

~NOTE: Changes needed to _use_ this property are not in this PR; we will use for the multi effect crystallizer we are adding to REFLO watertap-org/watertap-reflo#121~

- add ```specific_vol_phase```  and ```specific_vol_sat_phase``` property to pure water and crystallization property package, indexed to ```"Vap"``` 
- add ```temperature_sat_solvent``` property to pure water property package
- add to tests
- update docs for water prop pack
- add ref to crystallizer prop pack
- move some property metadata to ```define_custom_properties()``` in crystallizer prop pack

### Legal Acknowledgement


ksbeattie:
@kurbansitterley, what's the news here?

ksbeattie:
@kurbansitterley is this still a draft/work-in-progress or is it ready for review?

watertap/property_models/unit_specific/cryst_prop_pack.py
```diff
@@ -797,6 +797,45 @@ def build(self):
             doc="Solution enthalpy parameter E5",
         )
 
+        # Specific volume parameters, computed from Affandi expression (Eq 5)
```
adam-a-a, line 800: @kurbansitterley can you add some description for this in the associated documentation? And include this reference?

kurbansitterley:
These changes were made in the wrong property package.

kurbansitterley:
Reopened with changes made in water prop pack.

watertap/property_models/unit_specific/cryst_prop_pack.py
kurbansitterley: yes, found the reference

avdudchenko (approved):
This looks good to me.

watertap/property_models/water_prop_pack.py
```diff
@@ -1071,6 +1137,52 @@ def rule_therm_cond_phase(b, p):
...
+    def _temperature_sat_solvent(self):
+        self.temperature_sat_solvent = Var(
+            initialize=298.15,
+            bounds=(273.15, 1000.15),
+            units=pyunits.K,
+            doc="Vapour (saturation) temperature of pure solvent at boiling pressure",
+        )
+
+        def rule_temperature_sat_solvent(b):
+            psat = pyunits.convert(b.pressure_sat, to_units=pyunits.kPa)
```
ElmiraShamlou: This rule needs to be fixed-- as written it is circular. We should use `pressure` instead of `pressure_sat`; using `pressure_sat` makes the rule echo the state's `T` regardless of the actual pressure, when it should return `T_sat` at the state's pressure.

ElmiraShamlou (changes requested):

watertap/property_models/tests/test_water_prop_pack.py
```diff
@@ -97,6 +99,8 @@ def configure(self):
             ("cp_mass_phase", "Vap"): 1.871e3,
             ("visc_d_phase", "Liq"): 5.466e-4,
             ("therm_cond_phase", "Liq"): 0.6404,
+            ("specific_vol_phase", "Vap"): 12.024,
+            ("temperature_sat_solvent", None): 323.118,
```
ElmiraShamlou: This is the proof for my previous comment about fixing the `temperature_sat_solvent` rule. the test returns the input temperature (323k) instead of saturation temperature at the state's pressure. True T_sat at 2 bar is around 393K (120 C), much higher than the asserted value of 323K (50 C).

ElmiraShamlou (changes requested):

docs/technical_reference/property_models/water.rst
```diff
@@ -114,19 +118,41 @@ The default scaling factors are as follows:
...
+| T.V. Bartholomew, M.S. Mauter, 
+| Computational framework for modeling membrane processes without process and solution property simplifications, 
+| Journal of Membrane Science. 573 (2019) 682–693. 
+| https://doi.org/10.1016/j.memsci.2018.11.067.
+
+| Water Gas Phase Thermochemistry Data, National Institute of Standards and Technology, 2021, 
+| https://webbook.nist.gov/cgi/cbook.cgi?ID=C7732185&amp;Mask=1#Thermo-Gas.
+
+| M. Affandi, N. Mamat, S. N. A. M. Kanafiah and N. S. Khalid. 
+| Simplified Equations for Saturated Steam Properties for Simulation Purpose.
```
ElmiraShamlou, line 155: Since this source provides saturated steam properties **only**, this limitation should be reflected in `specific_vol_phase` rule and it does not apply outside the saturation curve. Otherwise, the result will contradict `dens_mass_phase[vap]` which is based on the ideal gas law and evaluates the density based on actual temperature and pressure.

ElmiraShamlou (changes requested):

docs/technical_reference/property_models/water.rst
kurbansitterley: I have addressed this by renaming this property to `saturation_vol_sat_phase` and then having `saturation_vol_phase = 1 / dens_mass_phase[vap]`

adam-a-a:
Merge this first and then steamheater #1760 which will change based on this PR (want to change sat pressure to sat temp)

adam-a-a (approved):

watertap/property_models/water_prop_pack.py
```diff
@@ -1071,6 +1140,66 @@ def rule_therm_cond_phase(b, p):
...
+            )
+
+        self.eq_specific_vol_sat_phase = Constraint(
+            ["Vap"], rule=rule_specific_vol_sat_phase
+        )
+
+    def _specific_vol_phase(self):
+        self.specific_vol_phase = Var(
+            ["Vap"],
+            initialize=0.001,
```
ElmiraShamlou: `0.001` seems closer to liquid water than vapor; maybe we should use an initial value closer to the superheated steam state.

watertap/property_models/water_prop_pack.py
```diff
@@ -1071,6 +1140,66 @@ def rule_therm_cond_phase(b, p):
 
         self.eq_therm_cond_phase = Constraint(["Liq"], rule=rule_therm_cond_phase)
 
+    def _specific_vol_sat_phase(self):
+        self.specific_vol_sat_phase = Var(
+            ["Vap"],
+            initialize=100,
```
ElmiraShamlou: `100` feels more like a deep vacuum case rather than typical HP/LP steam

watertap/property_models/water_prop_pack.py
```diff
@@ -1071,6 +1140,66 @@ def rule_therm_cond_phase(b, p):
 
         self.eq_therm_cond_phase = Constraint(["Liq"], rule=rule_therm_cond_phase)
 
+    def _specific_vol_sat_phase(self):
+        self.specific_vol_sat_phase = Var(
+            ["Vap"],
+            initialize=1e5,
```
ElmiraShamlou:
```suggestion
            initialize=0.2,
```

ElmiraShamlou (approved):
LGTM

PR #1496, opened 2024-09-24
Base class for handling the spacer correlations and process block cla…

charansamineni:
…ss to create a RO 1D unit model with user choice for spacer type, sherwood correlation, and pressure drop calculation. Original constraints are not deleted just deactivated

## Fixes/Resolves:

-
-

### Legal Acknowledgement

### DIFF
--- a/watertap/core/custom_spacer_base.py
+++ b/watertap/core/custom_spacer_base.py
@@ -9,7 +9,6 @@
 # information, respectively. These files are also available online at the URL
 # "https://github.com/watertap-org/watertap/"
 #################################################################################
-import numpy as np  # Required for tracking the spacer performance variables
 from enum import Enum, auto
 from pyomo.core.expr.numeric_expr import sin as pysin
 from pyomo.core.expr.numeric_expr import cos as pycos
--- a/watertap/core/custom_spacer_base.py
+++ b/watertap/core/custom_spacer_base.py
@@ -1,0 +1,719 @@
+#################################################################################
+# WaterTAP Copyright (c) 2020-2024, The Regents of the University of California,
+# through Lawrence Berkeley National Laboratory, Oak Ridge National Laboratory,
+# National Renewable Energy Laboratory, and National Energy Technology
+# Laboratory (subject to receipt of any required approvals from the U.S. Dept.
+# of Energy). All rights reserved.
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and license
+# information, respectively. These files are also available online at the URL
+# "https://github.com/watertap-org/watertap/"
+#################################################################################
+import numpy as np  # Required for tracking the spacer performance variables
+from enum import Enum, auto
+from pyomo.core.expr.numeric_expr import sin as pysin
+from pyomo.core.expr.numeric_expr import cos as pycos
+from math import pi
+from idaes.core.util.env_info import __author__
+from pyomo.common.config import ConfigValue, In
+from pyomo.core import NegativeReals, value
+from pyomo.core import Var
+from pyomo.environ import PositiveReals, NonNegativeReals, Reals
+from pyomo.environ import units as pyunits
+from pyomo.util.calc_var_value import calculate_variable_from_constraint
+from watertap.core.membrane_channel1d import MembraneChannel1DBlock
+
+__author__ == "Laxmicharan Samineni"
+
+
+def _add_spacer_config(CONFIG):
+    CONFIG.declare(
+        "spacer_type",
+        ConfigValue(
+            default=SpacerType.net_spacer,
+            domain=In(SpacerType),
+            description="Argument for selecting the type of feed spacer",
+            doc="""This argument changes the type of feed spacer by adding the required geometry variables along with 
+            the corresponding mass transfer and pressure drop calculations.
+    **default** - SpacerType.net_spacer.
+    **Valid values:** {
+        no_spacer: Converts RO unit model to a case with no spacer
+        net_spacer: Adds necessary correlations for net_spacers
+        printed_spacer: Adds 3D printed spacer and necessary correlations}""",
+        ),
+    )
+
+    CONFIG.declare(
+        "sherwood_correlation",
+        ConfigValue(
+            default=SherwoodCorrelation.guillen,
+            domain=In(SherwoodCorrelation),
+            description="Argument for selecting the Sherwood correlation to be used for mass transfer calculation",
+            doc="""This argument selects the Sherwood correlation to be used in the feed side mass transfer coefficient
+            calculation. This will effect the concentration polarization modulus.
+    **default** - SherwoodCorrelation.guillen.
+    **Valid values:** {
+        graetz: Graetz-Leveque correlation is used for mass transfer correlation. Exclusive for no spacer case
+        sieder: Sieder-Tate correlation is used for mass transfer correlation. Exclusive for no spacer case
+        guillen: Guillen & Hoek correlation is used for mass transfer correlation. Standard case in WaterTAP model
+        schock: Schock & Miquel correlation is used for mass transfer correlation.
+        da_costa: DaCosta correlation is used for mass transfer correlation.
+        parameterized: Parameterizes the Sherwood number
+        References: https://doi.org/10.1016/j.cej.2008.10.030 - Guillen & Hoek
+                    Da Costa, Andre. Fluid flow and mass transfer in spacer-filled
+                    channels for ultrafiltration. Diss. UNSW Sydney, 1993. - DaCosta and no_spacer correlations
+                    1) Table 5.1 for Graetz-Leveque and Sieder-Tate correlations
+                    2) Equation 5.20 for DaCosta Sherwood correlation}""",
+        ),
+    )
+
+    CONFIG.declare(
+        "pressure_correlation",
+        ConfigValue(
+            default=PressureCorrelation.guillen,
+            domain=In(PressureCorrelation),
+            description="Argument for selecting the feed side axial pressure drop equation",
+            doc="""This argument changes the correlation used to calculate the axial pressure drop on the feed side.
+    **default** - PressureCorrelation.guillen.
+    **Valid values:** {
+    no_spacer: Hagen-Poiseuille correlation is used for pressure drop correlation.
+    guillen: Guillen & Hoek correlation is used for pressure drop calculation. Standard case in WaterTAP model
+    schock: Schock & Miquel correlation is used for pressure drop calculation.
+    da_costa: DaCosta correlation is used for pressure drop calculation.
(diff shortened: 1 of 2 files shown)


ksbeattie:
@charansamineni might you be able to provide some tests and documentation updates for this?

charansamineni:
@ksbeattie, I am working on getting it ready. I have to add some more functionality based on feedback from reviewers.

watertap/unit_models/ro_1D_custom_spacer.py
```diff
@@ -0,0 +1,224 @@
...
+from idaes.core.util.units_of_measurement import report_quantity
+
+__author__ == "Laxmicharan Samineni"
+
+# Set up logger
+_log = idaeslog.getLogger(__name__)
+_log.setLevel(idaeslog.DEBUG)
+
+
+@declare_process_block_class("ReverseOsmosis1DCustomSpacer")
```
adam-a-a, line 48: The first thought that comes to mind is that this should not be a separate unit model. Although it could be, this would add redundancy that we typically try to avoid. Have you tried bringing your changes into the existing RO1D model? If we were to implement these changes, we'd likely want to incorporate into the membrane_channel and RO classes, where relevant. This makes the implementation a little trickier but is ultimately what should be merged into the main codebase.

watertap/unit_models/ro_1D_custom_spacer.py
```diff
@@ -0,0 +1,224 @@
...
+class ReverseOsmosis1DCustomSpacerData(ReverseOsmosisBaseData, CustomSpacer):
+    CONFIG = CONFIG_Template
+    _add_has_full_reporting(config_obj=CONFIG)
+    _add_spacer_config(CONFIG=CONFIG)
+    _add_geometry_config(CONFIG=CONFIG)
+
+    def build(self):
+        _log.warning("CUSTOM SPACER RO MODEL is being used for the build")
```
adam-a-a, line 56: This would be unnecessary once incorporating into our existing RO unit, but I do like the idea of logging what the user ultimately defined in the RO model configuration! No need to pursue that in this PR, but given the increasing complexity of config options, it might not be a bad idea for the user to get a snapshot of what config settings they ultimately are working with upon instantiation, since much of this can be opaque to new and experienced users alike.

watertap/unit_models/ro_1D_custom_spacer.py
```diff
@@ -0,0 +1,224 @@
...
+                "Discretization scheme was "
+                "not specified for the "
+                "reverse osmosis module."
+                "Defaulting to backward finite "
+                "difference."
+            )
+            self.config.transformation_scheme = "BACKWARD"
+            return
+
+    def _check_spacer_config(self):
```
adam-a-a, line 82: This method is going to require quite a bit of testing to ensure that intended combinations of config options work the way they should, and unintended, incorrect combinations actually raise the exception that you would expect.

watertap/core/custom_spacer_base.py
```diff
@@ -0,0 +1,718 @@
...
+        solute_set = self.feed_side.config.property_package.solute_set
+
+        @self.feed_side.Constraint(
+            self.flowsheet().config.time,
+            self.length_domain,
+            solute_set,
+            doc="Sherwood number equation modified to "
+            + str(self.config.sherwood_correlation),
+        )
+        def eq_N_Sh_comp_modified(b, t, x, j):
```
adam-a-a, line 198: Initially, I would think that the `N_sh` constraint should keep the original name rather creating a new constraint name, for the purpose of maintaining symmetry in the model when possible.

watertap/core/custom_spacer_base.py
```diff
@@ -0,0 +1,718 @@
...
+                    )
+                elif (
+                    self.config.sherwood_correlation
+                    == SherwoodCorrelation.parameterized
+                ):
+                    return b.N_Sh_comp[t, x, j] == b.Sh_param
+
+        # ===============================================================================================================
+        self.feed_side.eq_N_Sh_comp.deactivate()
+        self.feed_side.eq_N_Sh_comp_modified.activate()
```
adam-a-a, line 243: I would prefer this not to be here. Instead, it'd be great if we just kept `eq_N_Sh_comp`. This works as a bandaid for doing analysis locally, but in terms of merging to main and documenting model additions, having the one original constraint makes more sense.

watertap/core/custom_spacer_base.py
```diff
@@ -0,0 +1,718 @@
...
+
+    def _add_pressure_drop_correlations(self):
+        # Modifying the Pressure drop correlations based on the user input
+        if self.config.pressure_correlation != PressureCorrelation.da_costa:
+            self._modify_friction_factor()
+            self.feed_side.eq_friction_factor.deactivate()
+            self.feed_side.eq_friction_factor_modified.activate()
```
adam-a-a, line 250: same thing here---rather deactivate the original constraint and activate a new one, we should try to just use the original constraint and make appropriate changes to do so with overwriting the constraint, which Pyomo will throw up a warning for.

watertap/core/custom_spacer_base.py
```diff
@@ -0,0 +1,718 @@
...
+
+            self.feed_side.spacer_height = Var(
+                initialize=1e-3,
+                units=pyunits.m,
+                within=PositiveReals,
+                doc="spacer height",
+            )
+
+        if self.config.sherwood_correlation == SherwoodCorrelation.da_costa:
+            self.feed_side.Sh_corr_factor_da_costa = Var(
```
adam-a-a, line 340: These are intended for the feed_side of RO, which indicates that this should be in membrane_channel or the RO class. Can any of the correlations be applied to the permeate side, say, in OARO? If yes, then putting in one of the membrane_channel cases might make sense. If limited to the feed side for RO, then I wonder whether this should live here. In essence, custom_space_base hinges upon RO, but the fact that its introduced as a separate base class could indicate otherwise. Does it make more sense to _not_ have this separate base class? I say this only after a relatively quick scan--it could be that having this separate class is beneficial. Will be able to tell more after testing is introduced.

watertap/core/custom_spacer_base.py
```diff
@@ -0,0 +1,718 @@
...
+        # Fixes the spacer porosity to 1
+        self.feed_side.spacer_porosity.fix(1)
+
+    def _add_net_spacer_constraints(self):
+        # Add a constraint to calculate the spacer porosity based on the spacer geometry
+        @self.feed_side.Constraint(doc="Spacer porosity")
+        # Porosity calculation from Equation 3.7 in Da Costa, Andre. Fluid flow and mass transfer in spacer-filled
+        # channels for ultrafiltration. Diss. UNSW Sydney, 1993.
+        def eq_spacer_porosity(b):
+            return b.spacer_porosity == 1 - pi * b.filament_dia**2 / (
```
adam-a-a, line 429: FYI, we normally import constants like `pi` from `idaes.core.util.constants`. In this case, idaes just uses the same math import  for pi anyway, so doesn't matter here.

watertap/core/custom_spacer_base.py
```diff
@@ -0,0 +1,718 @@
...
+                2 * b.mesh_size * 2 * b.filament_dia * pysin(b.spacer_angle_radians)
+            )
+
+        # Constraint for spacer height
+        @self.feed_side.Constraint(doc="Spacer height calculation for net spacer")
+        def eq_spacer_height(b):
+            return b.spacer_height == 2 * b.filament_dia
```
adam-a-a, line 436: Should it always be the case that spacer height is 2*D? If not, should we add a mutable Param initialized to 2 instead? Also, are we introducing redundant constraints by doing this since spacer height=channel height?

adam-a-a: I guess the question is--is there ever a case when spacer height does _not_ equal channel height? If there is no such exception, I would suggest only maintaining channel_height. And if you really are tied to referring to spacer height, you could always create an Expression, `spacer_height`, which would just point to `channel_height`

watertap/core/custom_spacer_base.py
```diff
@@ -0,0 +1,718 @@
...
+                1) Table 5.1 for Graetz-Leveque and Sieder-Tate correlations
+                2) Equation 5.20 for DaCosta Sherwood correlation
+    """
+
+    graetz = auto()
+    sieder = auto()
+    guillen = auto()
+    schock = auto()
+    da_costa = auto()
+    parameterized = auto()
```
adam-a-a, line 128: Given the breadth of additions in this PR, we'll definitely want to add documentation to readthedocs to capture all the additions. That can come at the end of this though, once satisfied with the core code changes.

adam-a-a (changes requested):
Thank you for putting up these proposed changes, and congratulations on putting up your first PR here! I have long wanted to incorporate more detailed correlations related to spacer geometry and am glad that you've taken the step to do so. 

There is a bit of work to do here before we can merge these changes. After my first pass, here are my overall, initial suggestions:

1. Add a test file for ro_1D_custom_spacer. Here, you can test different configuration setups for RO, which would show reviewers more clearly how the changes would channel through to the user. Moreover, this would give you the chance to identify low-hanging bugs in the existing code. Perhaps you shouldn't add every test under the sun in relation to the correlations in `custom_spacer_base` because, if we do keep the separate class, I would think there would be a dedicated test file for that class. That said, it won't hurt to do so minimal level of testing and move tests later. From here, we should do another pass of reviews.
2.  After that, once you have some basic tests that probe the various ways of configuring RO1D, now you can switch gears and try merging changes from ro_1D_custom_spacer directly into the existing classes for RO and/or membrane channel. You would also add your new tests for this custom RO1d to the existing test file for RO1d. Many things will break, and all that breakage will need repair until tests pass. 

Try doing step 1, and let's revisit the custom_spacer_base class after that.

ksbeattie:
@charansamineni any news on this?

charansamineni:
Hey Keith,

Sorry for the delay in getting back to you. I was out last week. I am changing my code organization based on the suggestions I received from Adam Atia. Do you think it is better for me to do that as a separate PR and close this one?

Charan

From: Keith Beattie ***@***.***>
Date: Thursday, November 14, 2024 at 1:55 PM
To: watertap-org/watertap ***@***.***>
Cc: Laxmicharan Samineni ***@***.***>, Mention ***@***.***>
Subject: Re: [watertap-org/watertap] Base class for handling the spacer correlations and process block cla… (PR #1496)

@charansamineni<https://github.com/charansamineni> any news on this?

ksbeattie:
> Do you think it is better for me to do that as a separate PR and close this one?

Either is fine, whichever is easier for you is what I'd suggest.  @adam-a-a will be reaching out to you to help.

charansamineni:
@ksbeattie: I closed this for now. I will discuss with Adam once and reopen based on the feedback.

PR #1501, opened 2024-10-02
Add initialize function to fsapi

MichaelPesce:
## Fixes/Resolves:

- adds a hook for flowsheets to provide an optional initialize function to the UI

## Summary/Motivation:

The parameter sweep tool (see [UI PR#103](https://github.com/watertap-org/watertap-ui/pull/103)) with parallel capabilities allows for an optional initialize function. in the case that there is no initialize function, it runs the build function multiple times. This takes a long time. If the UI can access a models initialize function, then parameter sweep run times can be decreased.

- update fsapi to allow for a `do_initailize` function

### Legal Acknowledgement


watertap/ui/fsapi.py
```diff
@@ -639,11 +641,14 @@ def __init__(
             (do_export, "export"),
             (do_build, "build"),
             (do_solve, "solve"),
+            (do_initialize, "initialize"),
         ):
             if arg:
                 if not callable(arg):
                     raise TypeError(f"'do_{name}' argument must be callable")
                 self.add_action(getattr(Actions, name), arg)
+            elif name == "initialize":
```
dangunter, line 650: this shouldn't be necessary, since if do_initialize is set, the previous 'if' will be activated and add the action

watertap/ui/fsapi.py
```diff
@@ -703,6 +708,20 @@ def get_diagram(self, **kwargs):
         else:
             return None
 
+    def initialize(self, *args, **kwargs):
```
dangunter, line 711: this should be a try/except like in 'build' instead of like 'get_diagram'

dangunter (commented):
I think it makes sense with the suggested changes. I need to look it over one more time though, maybe the `initialize` internal method can be dropped entirely, not sure.

watertap/ui/fsapi.py
MichaelPesce: @dangunter I threw this in there because we want the initialize function to be optional. In the case that `arg` is `None`, the `else` returns an error, because the other 3 actions are all required. By putting this here we add the initialize function (as `None`) even in the case that it is not provided, rather than erroring out

dangunter (approved):
LGTM

adam-a-a (approved):

PR #1503, opened 2024-10-02
BSM2-P Effluent Metrics w/ Flowsheet Constraints

MarcusHolly:
## Summary/Motivation:
An extension of #1492 that adds the effluent violation constraints into the flowsheet

- Adds effluent violation constraints into BSM2-P

### Legal Acknowledgement


lbianchi-lbl:
@MarcusHolly to make debugging faster by avoiding to run the entire test suite, you can search the `.github/workflows/checks.yml` file for `pytest --pyargs watertap` (there should be 2 places when it appears) and replace it with `pytest --pyargs watertap -k test_BSM2_P_extension` (i.e. `-k` followed by a space and then the name of the test file that you want to run by itself without the `.py` suffix. More info https://docs.pytest.org/en/stable/how-to/usage.html)

adam-a-a:
@lbianchi-lbl I should also note that when I run tests locally on this PR, the same tests that fail here pass on my machine. I would expect the tests to fail locally if the workflow were truly aligned. Is there a possibility that something is out-of-sync? I wonder if it has anything to do with the ipopt-watertap solver and whether that is incorporated in workflows (to start). Although we know this is a troublesome flowsheet, I think we should still expect some alignment between tests that run locally vs. workflows, so I am wondering if this warrants further investigation.

lbianchi-lbl:
The current conflicts are due to the fact that this PR is pretty old.

adam-a-a:
@luohezhiming any reason why we can't put up a pull request with your version of BSM2-P with effluent metric constraints?

luohezhiming:
> @luohezhiming any reason why we can't put up a pull request with your version of BSM2-P with effluent metric constraints?

@adam-a-a , mine version has electroNP included and optimize with cathodic potential and area volume ratio, and does not optimize with reactor volumes.

adam-a-a:
> > @luohezhiming any reason why we can't put up a pull request with your version of BSM2-P with effluent metric constraints?
> 
> @adam-a-a , mine version has electroNP included and optimize with cathodic potential and area volume ratio, and does not optimize with reactor volumes.

so there is no option to exclude electronp, right? would it be difficult to incorporate that flexibility?

lbianchi-lbl:
This is still something we want. We should wait until the scaling PRs have been merged before going back to this.

ksbeattie:
@MarcusHolly any thoughts on this now?

MarcusHolly:
I should revisit this soon, but I won't have time to work on this until after the IDAES+ meeting

PR #1512, opened 2024-10-30
Repair OLIAPI

adam-a-a:
## Fixes/Resolves:
OLI API stopped working. 

## Summary/Motivation:
Users were no longer able to get back OLI calculations. We were able to log in, generate a DBS file, and that was about it. The issue came down to an additional forward slash used in one of the URLs, which previously worked.

EDIT: Subsequently, CI checks were failing as well, while local tests were passing. This PR also addresses those issues.

- revise url
- one other catch for handling results
- revise test code to prevent CI check failure, including use of refresh token after oli login, and running checks sequentially in ci repo (thanks @lbianchi-lbl for editing the workflow)

### Legal Acknowledgement


lbianchi-lbl:
The OLI API CI failures are not entirely reproducible. @adam-a-a will reach out to OLI to investigate this.

ElmiraShamlou (approved):
LGTM

watertap/tools/oli_api/credentials.py
```diff
@@ -102,6 +103,7 @@ def __init__(
         :param access_keys: list of access keys generated by user
         :param test: bool switch for automation during tests
         :param interactive_mode: bool to switch level of logging display from info to debug only
+        :param refresh: bool to use refresh token in login method
```
MarcusHolly, line 106: What does this refresh actually do?

MarcusHolly (commented):
LGTM, but CodeCov is failing mostly as a result of the new logger and error messages being untested

watertap/tools/oli_api/credentials.py
adam-a-a: From my recollection, it enables usage of a refresh token so that we don't log out prematurely.

watertap/tools/oli_api/client.py
```diff
@@ -103,13 +105,18 @@ def __enter__(self):
...
+        _logger.info(
+            f"Exiting: deleting {len(self.session_dbs_files)} remaining DBS files created during the session that were not marked by keep_file=True."
+        )
         self.dbs_file_cleanup(self.session_dbs_files)
         return False
 
     def _prompt(self, msg, default=""):
         if self.interactive_mode:
+            msg = msg + "Enter [y]/n to proceed."
             return input(msg)
```
MarcusHolly, line 116: Does the interactive mode functionality work in Jupyter? It doesn't seem to be the case, and, if so, perhaps it should be set to False in the tutorial

MarcusHolly (approved):
LGTM - the tutorial runs fine on my end now. I just have one small comment...

PR #1519, opened 2024-11-06
ASM1 Property Package Re-Scaling

MarcusHolly:
## Summary/Motivation:
Reworks how the ASM1 property package is scaled according to the new IDAES scaling routine

- Updates ASM1 property package scaling
- Updates the corresponding test files

### Legal Acknowledgement


lbianchi-lbl:
- This needs a pre-release version of IDAES
- We might want to wait until IDAES/idaes-pse#1523 is merged to create a tag in watertap-org/idaes-pse so that the latest changes in Pyomo are also included

watertap/property_models/unit_specific/activated_sludge/asm1_properties.py
```diff
@@ -513,35 +542,3 @@ def define_display_vars(self):
...
-
-            iscale.set_scaling_factor(self.material_flow_expression[j], sf_F * sf_C)
-            iscale.set_scaling_factor(self.material_density_expression[j], sf_C)
-
-        # Enthalpy and energy terms
-        sf_rho_cp = pyo.value(1 / (self.params.dens_mass * self.params.cp_mass))
-        iscale.set_scaling_factor(
-            self.enthalpy_flow_expression, sf_F * sf_rho_cp * sf_T
-        )
-        iscale.set_scaling_factor(self.energy_density_expression, sf_rho_cp * sf_T)
```
adam-a-a: Do we need to remove this at this point? We can just leave it in until we completely do away with use of `calculate_scaling_factors`. For our purposes, I would think that we just wouldn't call calculate_scaling_factors on the model.

I guess the only reason we might consider temporarily deleting this is so calculate_scaling_factors isn't called here when we call it at the flowsheet level (for the time being).

MarcusHolly: I think we can keep these for now. I had deleted them before we discussed on Monday and just haven't added them back in yet.

adam-a-a (commented):
Unsure of current status on this, but I still would like to see things come in one at a time. My reasoning is that, ideally, if we only add a Scaler for ASM1, it might be easier to troubleshoot failed tests. Additionally, you could focus specifically on ASM1 tests and perhaps particular unit model tests that use ASM1. Moreover, adding a test for ASM1 Scaler could be helpful to gauge whether your default scaler class works better or worse with different settings. From there, we could move on to ADM1 and do the same for that.

adam-a-a:
Right now, for example, you might not be able to tell if ASM1, ADM1 liquid or ADM1 vapor properties have problematic scaling that cause checks to fail.

MarcusHolly:
> Right now, for example, you might not be able to tell if ASM1, ADM1 liquid or ADM1 vapor properties have problematic scaling that cause checks to fail.

I agree - I just haven't touched this since we had our last meeting since we're still waiting on https://github.com/IDAES/idaes-pse/pull/1523 to go through. I'll make sure only one prop pack is changed at a time.

MarcusHolly:
> * This needs a pre-release version of IDAES
> * We might want to wait until [Bumping the IDAES tag of Pyomo to check for issues IDAES/idaes-pse#1523](https://github.com/IDAES/idaes-pse/pull/1523) is merged to create a tag in watertap-org/idaes-pse so that the latest changes in Pyomo are also included

@lbianchi-lbl It seems like [Bumping the IDAES tag of Pyomo to check for issues IDAES/idaes-pse#1523](https://github.com/IDAES/idaes-pse/pull/1523) is close to being merged, so just sending a reminder that this PR needs a tag for a pre-release version of IDAES after the PR goes through. If the IDAES PR just needs one more rubber stamp, I could do that.

watertap/property_models/unit_specific/activated_sludge/asm1_properties.py
MarcusHolly: I've added these back, and I don't think there will be any problems with keeping the old `calculate_scaling_factors` around

tutorials/BSM2.ipynb
agarciadiego, line 1: Are these changes just to match the tutorial with the flowsheet.py?

MarcusHolly: Correct, although I didn't add costing

luohezhiming (approved):
LGTM!

watertap/property_models/unit_specific/activated_sludge/tests/test_asm1_reaction.py
```diff
@@ -161,6 +167,121 @@ def check_units(self, model):
         assert_units_consistent(model)
 
 
+class TestASM1ReactionScaler(object):
```
adam-a-a, line 170: One thought that comes to mind--- I am hoping we can at least track some metric, even it doesn't adequately represent the shift in model states, to keep some record to quantify the performance of a scaler class (as well as the old calculate_scaling_factors method). 

Perhaps the easiest thing to do is to use the DiagnosticToolbox, check the condition number and assert that value for the new Scaler, the old method-based approach, and any new Scaler classes (or revisions to existing ones) that we bring in. This way, we can get some idea of how impactful scaler additions/revisions are as we progress.

adam-a-a (approved):
LGTM, and I am ready to approve. @MarcusHolly  The _one thing_ I think we should consider adding is some assertions to check condition number before/after applying scaler. Ideally we'd do this for the old calculate_scaling_factors approach too. This way, we can be mindful of model improvements as we go, rather than just depending on the model to solve, which doesn't necessarily tell us whether one scaler is better than the next (which might be a fool's errand in general anyway).

watertap/property_models/unit_specific/activated_sludge/tests/test_asm1_reaction.py
MarcusHolly: @adam-a-a I should be able to add this relatively easily. Can you take a look at the AD tests I added here for comparing the old and new scaling methods: https://github.com/watertap-org/watertap/blob/2e9aaafa0e020e8afa8a7beaf9e3e6ec37e59a00/watertap/unit_models/tests/test_anaerobic_digester.py#L854

That PR is still a WIP, but all I need to do is refine the scaling for the bio_P=False case in the BSM2_P flowsheet

MarcusHolly:
> LGTM, and I am ready to approve. @MarcusHolly The _one thing_ I think we should consider adding is some assertions to check condition number before/after applying scaler. Ideally we'd do this for the old calculate_scaling_factors approach too. This way, we can be mindful of model improvements as we go, rather than just depending on the model to solve, which doesn't necessarily tell us whether one scaler is better than the next (which might be a fool's errand in general anyway).

This discussion was continued offline - these tests will be updated in future PRs as we add more UnitModelScalers to WaterTAP. We will also consider the possibility of using the IPOPTConvergenceAnalysis to compare metrics from the old scaling routine to those of the new scaling routine.

PR #1520, opened 2024-11-07
Update BSM2-P CSTR Model

MarcusHolly:
## Summary/Motivation:
The BSM2-P flowsheet used the IDAES CSTR, but it should be using the WaterTAP CSTR, which adds a constraint for hydraulic retention time.

- Substitutes IDAES CSTR for WaterTAP CSTR in BSM2-P flowsheet
- Updates the flowsheet's scaling factor

### Legal Acknowledgement

### DIFF
--- a/watertap/flowsheets/full_water_resource_recovery_facility/BSM2_P_extension.py
+++ b/watertap/flowsheets/full_water_resource_recovery_facility/BSM2_P_extension.py
@@ -742,6 +742,11 @@ def add_costing(m):
         if isinstance(block, UnitModelBlockData) and hasattr(block, "costing"):
             iscale.set_scaling_factor(block.costing.capital_cost, 1e-5)
 
+    iscale.constraint_scaling_transform(m.fs.AD.costing.capital_cost_constraint, 1e-6)
+    iscale.constraint_scaling_transform(
+        m.fs.dewater.costing.capital_cost_constraint, 1e-6
+    )
+
 
 def display_costing(m):
     print("Levelized cost of water: %.2f $/m3" % pyo.value(m.fs.costing.LCOW))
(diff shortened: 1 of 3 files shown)


watertap/flowsheets/full_water_resource_recovery_facility/BSM2_P_extension.py
```diff
@@ -537,6 +537,10 @@ def scale_variables(m):
             if "conc_mass_comp" in var.name:
                 iscale.set_scaling_factor(var, 1e1)
 
+    for unit in ("R1", "R2", "R3", "R4"):
+        block = getattr(m.fs, unit)
+        iscale.set_scaling_factor(block.hydraulic_retention_time, 1e-3)
```
adam-a-a, line 542: Since we are fixing volume, does it make sense to scale this by inverse of volume/flowrate, perhaps using plant influent flowrate as a start?

MarcusHolly: I've just tried doing this, but none of these new scaling factors work (i.e. the system fails to solve)

luohezhiming: what if we don't provide the scaling facotrs for the HRT as old version?

MarcusHolly: Same thing - if we don't provide any scaling factors for HRT, the system fails to solve

adam-a-a (approved):
LGTM- checks are passing so I am assuming these changes to make the model any more unstable than it already is.

luohezhiming (approved):

agarciadiego (approved):

PR #1521, opened 2024-11-07
Exclude flexparser 0.4 as workaround for IDAES/idaes-pse#1524

lbianchi-lbl:
## Summary/Motivation:
- Work around failures due to IDAES/idaes-pse#1524 while the issue is dealt with upstream

- Add requirement to exclude flexparser 0.4 

### Legal Acknowledgement


adam-a-a (approved):
Approving out of anticipation of checks passing

lbianchi-lbl:
Since the failure would occur when `idaes get-extensions` is called, and the checks are all currently past that stage, it's safe to merge this in even before the rest of the checks have completed.

lbianchi-lbl:
This change should ideally be reverted as the issue has been fixed upstream with Pint 0.24.4.

PR #1524, opened 2024-11-13
Update looptool documentation

johnson12742:
## Fixes/Resolves:

Update the loop_tool documentation

## Summary/Motivation:

Looptool location has not yet been updated in the docs. Also, requirement for looptool inputs have not been mentioned. This PR updates the documentation to address these issues. 

- Direct users to the new location of loop_tool
- Let users know that loop_tool inputs must be literals

### Legal Acknowledgement


johnson12742:
@lbianchi-lbl One of the checks is failing (though it is not a required). I re-ran it (in case it was a one off) but that didn't solve anything.
I only changed two lines in the looptool documentation.

lbianchi-lbl:
This looks like a server-side error on the Codecov side, so unfortunately it's unlikely we can do anything to fix it if it's not addressed by them. I couldn't find any related issue so I opened one: https://github.com/codecov/feedback/issues/577

docs/how_to_guides/how_to_use_loopTool_to_explore_flowsheets.rst
```diff
@@ -143,7 +143,9 @@ The sweep_param_loop parameters will be iterated over one by one, to sweep over
             param: key on the flowsheet that can be found using m.find_component (e.g., fs.costing.reverse_osmosis.membrane_cost)
             lower_limit: lower value for sampling
             upper_limit: upper value for sampling
-            num_samples: number of samples to run 
+            num_samples: number of samples to run
+
+Please note that .yaml only supports literals. Inputs such as 150 * 0.8 will be parsed as strings and throw error.
```
lbianchi-lbl:
```suggestion
Please note that YAML only supports literals for numeric data types. Expression-like values such as ``150 * 0.8`` will be parsed as strings and cause errors once the YAML inputs have been loaded.
```

lbianchi-lbl:
A couple of thoughts:

- There are more occurrences of the outdated import path: https://github.com/search?q=repo%3Awatertap-org%2Fwatertap+watertap.tools.analysis_tools.loop_tool&type=code
- The reason why this wasn't caught earlier is that the code snippets in the Loop tool guide use `.. code-block` instead of `.. testcode` directives, causing them to be ignored by the doctests: compare e.g.
  https://github.com/watertap-org/watertap/blob/532ce63868ca0dc949731866a7caa4e9d8c468ac/docs/how_to_guides/how_to_use_loopTool_to_explore_flowsheets.rst?plain=1#L44-L62
  https://github.com/watertap-org/watertap/blob/532ce63868ca0dc949731866a7caa4e9d8c468ac/docs/how_to_guides/how_to_run_differential_parameter_sweep.rst?plain=1#L12-L33
- Ideally, we should replace the `.. code-block` directives with `.. testcode` so that these code blocks are tested as part of our CI checks, reducing the chances for future issues with outdated code to go unnoticed

johnson12742:
@lbianchi-lbl So, what would you suggest? I am happy to close this PR if you think there is a better way to update the docs (compared to what I am doing).

lbianchi-lbl:
> @lbianchi-lbl So, what would you suggest? I am happy to close this PR if you think there is a better way to update the docs (compared to what I am doing).

@johnson12742 sorry for the confusion. To clarify, I think the changes in this PR are good (and, BTW, thanks for making them). The items I listed in my previous message are things that I think we should consider doing _in addition_ to the changes you've made.

adam-a-a (approved):

adam-a-a (commented):
Sorry- approved and then saw @lbianchi-lbl 's comments. If we want to address those in this PR, I will take my review back

johnson12742:
@lbianchi-lbl I didn't see any other instances where the documentation needs to be updated. I think we are good to go.

lbianchi-lbl:
@johnson12742 these are the lines that (as of d80a9932) still have a reference to the invalid import path:

https://github.com/johnson12742/watertap/blob/d80a99328ebb9cec8f45a44d769380204e7ae376/docs/how_to_guides/how_to_use_loopTool_to_explore_flowsheets.rst?plain=1#L1-L8

johnson12742:
@lbianchi-lbl Thanks for catching that. I don't know why it didn't show up when I searched

lbianchi-lbl (approved):

PR #1529, opened 2024-11-21
Changing ED-1D model variables to a function of length (x)

hunterbarber:
## Fixes/Resolves:
Some of the fluid dynamics variables were calculated based on conditions at the inlet. These variables are changed to appropriately be a function of length with other 1D predictions.

## Summary/Motivation:
Added fidelity of ED modeling.

- Make variables a function of length, referencing properties as a function of length
- Create diluate and concentrate flow channel sets

### Legal Acknowledgement


watertap/unit_models/electrodialysis_1D.py
```diff
@@ -713,20 +721,22 @@ def _make_performance(self):
...
-                self.flowsheet().time,
-                self.diluate.length_domain,
-                doc="Express deltaP_term by the calculated pressure drop data, concentrate.",
+                self.difference_elements,
+                doc="Express deltaP_term by the calculated pressure drop data.",
             )
-            def eq_deltaP_concentrate(self, t, x):
-                return self.concentrate.deltaP[t, x] == -self.pressure_drop[t]
+            def eq_deltaP(self, c, t, x):
+                if x not in self.difference_elements:
```
hunterbarber, line 730: I want to double-check if this is necessary, may be old code that is no longer relevant.

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -670,17 +683,12 @@ def build(self):
...
-        self.velocity_diluate = Var(
-            self.flowsheet().time,
-            self.diluate.length_domain,
-            initialize=0.01,
-            units=pyunits.meter * pyunits.second**-1,
-            doc="Linear velocity of flow",
-        )
-        self.velocity_concentrate = Var(
+        self.velocity = Var(
+            self.flow_channel_set,
```
adam-a-a, line 687: Subtle thing, but maybe the typical order should remain in place instead of putting the channel set first (i.e., time first, then space). I'd put flow_channel_set index after time, either between time and length domain, or right after length domain.

adam-a-a: Really more of a "consistency across models" issue but don't want to hold this up if there's good reason for this

watertap/unit_models/electrodialysis_1D.py
lbibl: @hunterbarber , I would advise against manually altering the discretization transformation in this model. The original 1-D model fully utilized Pyomo's DAE capabilities and was designed to support other discretization techniques besides finite differences, such as the orthogonal collocation scheme. Direct manipulation like this compromises the robustness of the DAE package.

watertap/unit_models/electrodialysis_1D.py
lbibl: @hunterbarber , if you do not like the original naming/treatment, a better design would be put this var under each of diluate and concentrate block rather than introducing an extra 2-member set that is only used once.

lbibl:
Another general suggestion: This PR seems to account for velocity changes along the length and their impact on pressure changes. This was previously unaddressed because, in typical or conventional ED systems, velocity changes along the length are widely considered negligible due to minimal water transport, as noted in most prior studies. However, this assumption might not hold for high-salinity ED systems, where water transport is significantly larger. I recommend encapsulating this length-dependent velocity and pressure calculation in a separate configuration to preserve the original design. In typical ED systems, such calculations are unnecessary and only add redundant computational complexity, complicating downstream optimization and analysis. The original tests are likely to fail as a result, so please consider making this functionality optional, triggered only when warranted, such as in high-salinity applications."

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -1908,33 +1947,46 @@ def eq_pressure_drop(self, t):
...
         )
-        def eq_pressure_drop_total(self, t):
-            return (
-                self.pressure_drop_total[t] == self.pressure_drop[t] * self.cell_length
+        def eq_pressure_drop_total(self, c, t):
+            return self.pressure_drop_total[c, t] == sum(
+                self.pressure_dx[c, t, x]
+                * self.cell_length
+                / len(self.difference_elements)
+                for x in self.difference_elements
```
johnson12742, line 1989: Is there not an integration scheme in Pyomo?

In my experience simple integration schemes like this are prone numerical error. My suggestion would be a Gauss Quadrature (or trapezoidal rule). 
Though depending on the intended system it may not be a big difference in accuracy.

hunterbarber:
> Another general suggestion: This PR seems to account for velocity changes along the length and their impact on pressure changes. This was previously unaddressed because, in typical or conventional ED systems, velocity changes along the length are widely considered negligible due to minimal water transport, as noted in most prior studies. However, this assumption might not hold for high-salinity ED systems, where water transport is significantly larger. I recommend encapsulating this length-dependent velocity and pressure calculation in a separate configuration to preserve the original design. In typical ED systems, such calculations are unnecessary and only add redundant computational complexity, complicating downstream optimization and analysis. The original tests are likely to fail as a result, so please consider making this functionality optional, triggered only when warranted, such as in high-salinity applications."

My counterargument to this is that including a rigorous pressure drop calculation is already optional. These calculations are employed when pressure drop is already configured to be calculated through `Darcy_Weisbach`. If the computational expense of a pressure drop calculation is not desired, that is already satisfied through `none` or `experimental` options. By tests failings on prior flowsheets, this actually confirms that this calculation is relevant to have as a function of velocity(x), Sh(x), Re(x), etc. even for conventional applications with any amount of water transport.

watertap/unit_models/electrodialysis_1D.py
hunterbarber: I think this is taken directly from the RO 1D models currently in WaterTAP, so I wasn't reinventing the approach of our unit models here. I would have to double check.

watertap/unit_models/electrodialysis_1D.py
hunterbarber: I think this was done because there was a discrepancy between the actual pressure drop supplied and the pressure drop in the model under the old scheme. I noted that I wanted to double check this to confirm, but was battling my environment and ipopt being broke in the meantime.

watertap/unit_models/electrodialysis_1D.py
johnson12742: Ah, got it.

Following up on that. Is the expected variation in pressure profile more or less 'complex' than in RO 1D? I saw in another thread that one of the goals of this PR is to get accurate pressure drops. A better integration scheme could be critical to achieve it.

lbibl:
The 1-d control volume and dae weren’t set up exactly the same between ED and RO by my last checking.

lbibl:
My emphasis was for the choice of treating velocity and pressure as a partial differential equation or not, and setting differential equations means extra significant constraint complexity that would not be necessary for a typical ED system and would add optimization difficulties.   Practically, if you are going to implement this change globally, you’ll most likely need to redo all ED flowsheets and tests in WT.

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -713,20 +721,22 @@ def _make_performance(self):
             self._pressure_drop_calculation()
 
             @self.Constraint(
+                self.flow_channel_set,
                 self.flowsheet().time,
-                self.diluate.length_domain,
-                doc="Express deltaP_term by the calculated pressure drop data, diluate.",
-            )
-            def eq_deltaP_diluate(self, t, x):
-                return self.diluate.deltaP[t, x] == -self.pressure_drop[t]
```
adam-a-a, line 721: Just so I understand correctly, does the existing model just assume `deltaP` (dP/dX) is constant, i.e., the same for each element? And now, with @hunterbarber 's modifications, we'd be calculating dP/dX per element rather than having a fixed dP/dx per element---is that correct? If not, please clarify.

hunterbarber: Effectively yes. In the old model, the limiting current density and pressure drop constraints were calculated using values at the inlet for variables such as velocity, friction factor, Sh, Re, etc. This indexes those variables for truer 1D predictions.

watertap/unit_models/electrodialysis_1D.py
adam-a-a: @hunterbarber is correct. This aligns with what is done for the 1DRO model. From the code, it appears that the preexisting formulation assumed constant dP/dx across each element, whereas @hunterbarber 's changes allow for varying dP/dx across the discretized length.

I need to refresh my memory on what each PressureDropMethod config option corresponds to (i.e., experimental vs Darcy_Weisbach), but if my interpretation is correct about dP/dx assumed as constant/uniform across elements, then I would opt for keeping that preexisting formulation as its own PressureDropMethod option, and @hunterbarber 's revised formulation would be a separate option. For example, in the RO model, we have an option that assumed that dP/dX is fixed and constant, another that just takes a pressure drop across the whole module as the fixed variable, and another that does a more detailed calculation (like the one shown in this PR).

To avoid widespread test failures and fear of increasing model complexity, we could maintain the preexisting formulation for pressure drop calculations as the default config option, and users would have to manually set this more detailed option for pressure drop calculations.

adam-a-a: @johnson12742 I think if you wanted to explore alternative integration schemes to improve accuracy and discuss more with IDAES/Pyomo teams, that'd be fine---but I would think that would lead to a separate PR if anything.

watertap/unit_models/tests/test_electrodialysis_1D.py
```diff
@@ -1895,12 +1899,12 @@ def test_deltaP_various_methods(self, ed_m0, ed_m1, ed_m2, ed_m3, ed_m4, ed_m5):
         assert value(ed_m[5].fs.unit.outlet_diluate.pressure[0]) == pytest.approx(
             178659.214, rel=1e-3
         )
-        assert value(ed_m[5].fs.unit.pressure_drop[0]) == pytest.approx(
+        assert value(ed_m[5].fs.unit.pressure_dx["diluate", 0, 0]) == pytest.approx(
```
johnson12742, line 1902: Are there plans to add pressure_dx["concentrate", 0, 0] and pressure_drop_total["concentrate", 0] to the unit tests?

watertap/unit_models/electrodialysis_1D.py
```diff
@@ -713,20 +721,22 @@ def _make_performance(self):
...
+                if x not in self.difference_elements:
+                    self.concentrate.deltaP[t, x].fix(0)
+                    self.diluate.deltaP[t, x].fix(0)
+                else:
+                    if c == "concentrate":
+                        return (
+                            self.concentrate.deltaP[t, x] == -self.pressure_dx[c, t, x]
+                        )
+                    elif c == "diluate":
+                        return self.diluate.deltaP[t, x] == -self.pressure_dx[c, t, x]
```
adam-a-a, line 739: For 1D, deltaP is essentially dP_dx, as is pressure_dx. It seems redundant to do this, but maybe I am missing something. Should `pressure_dx` instead just be a reference of `deltaP`?

watertap/unit_models/electrodialysis_1D.py
adam-a-a: For reference, this is what happens within 1DRO at the moment (note, we referenced dP_dX to 1d control volume's deltaP variable, then set a deltaP variable at the unit level as opposed to control volume level to represent overall pressure drop):
  ```
def _add_deltaP(self):
      self.deltaP = Var(
          self.flowsheet().config.time,
          initialize=-1e5,
          bounds=(-1e6, 0),
          domain=NegativeReals,
          units=units_meta("pressure"),
          doc="Pressure drop across unit",
      )
      if self.config.pressure_change_type == PressureChangeType.fixed_per_stage:

          @self.Constraint(
              self.flowsheet().config.time,
              self.length_domain,
              doc="Fixed pressure drop across unit",
          )
          def eq_pressure_drop(b, t, x):
              return b.deltaP[t] == b.length * b.feed_side.dP_dx[t, x]

      else:

          @self.Constraint(
              self.flowsheet().config.time, doc="Pressure drop across unit"
          )
          def eq_pressure_drop(b, t):
              return b.deltaP[t] == sum(
                  b.feed_side.dP_dx[t, x] * b.length / b.nfe
                  for x in b.difference_elements
              )
```

adam-a-a: Will `if x not in self.difference_elements:` ever be true since your new index is in terms of the difference_elements set? It appears not, but maybe I am misinterpreting.

lbibl:
@hunterbarber : we discussed on the PR in the last dev meeting and reached two potential options: (1) consider redoing this PR in the future in a way that insulates the added feature from the current setup for a typical ED, likely through a CONFIG;  (2) as these features are currently specialized for the brine concentrator case, we may consider setting up a separate repository to host the code supporting your paper in preparation and open-sourcing it alongside your future publication.  Following  either of the options could allow us to close this PR for now.   We will hold off on taking action for a short time to see if you have any input.

PR #1530, opened 2024-11-22
ADM1 Property Package Re-Scaling

MarcusHolly:
## Summary/Motivation:
Reworks how the ADM1 property package is scaled according to the new IDAES scaling routine

- Updates ADM1 property package scaling
- Updates the corresponding test files

### Legal Acknowledgement


watertap/property_models/unit_specific/anaerobic_digestion/adm1_properties_vapor.py
```diff
@@ -217,13 +218,53 @@ def release_state(self, flags, outlvl=idaeslog.NOTSET):
         init_log.info("State Released.")
 
 
+class ADM1_vaporPropertiesScaler(CustomScalerBase):
```
MarcusHolly: This may not actually be usable since the vapor prop pack doesn't have a control volume. When testing this scaler in #1535 , I get the following error: `AttributeError: '_ScalarADM1_vaporParameterBlock' object has no attribute 'temperature'`. If this is true, we should probably just undo these changes to the vapor prop pack. 

EDIT: I've undone these changes since it was causing issues with other flowsheets (properties not existing)

agarciadiego (approved):
LGTM

adam-a-a (approved):

PR #1535, opened 2024-11-29
Anaerobic Digester Re-Scaling

MarcusHolly:
## Summary/Motivation:
Adds the new IDAES scaling routine (scaler objects) to the anaerobic digester unit model. This should be merged after #1519 and #1530 are merged in.

- Updates the scaling routine of the anaerobic digester
- Updates testing

### Legal Acknowledgement


watertap/property_models/unit_specific/anaerobic_digestion/tests/test_modified_adm1_reaction.py
```diff
@@ -940,6 +940,9 @@ def model(self):
         m.fs.unit.liquid_phase.properties_in[0].TSS
         m.fs.unit.liquid_phase.properties_out[0].TSS
 
+        iscale.calculate_scaling_factors(m)
+        iscale.set_scaling_factor(m.fs.unit.liquid_phase.heat, 1e-6)
```
adam-a-a: WHy did you need this change?

watertap/unit_models/anaerobic_digester.py
```diff
@@ -836,6 +927,21 @@ def calculate_scaling_factors(self):
...
+        iscale.set_scaling_factor(self.KH_h2, 1e2)
+        iscale.set_scaling_factor(self.hydraulic_retention_time, 1e-6)
+        iscale.set_scaling_factor(self.volume_AD, 1e-2)
+        iscale.set_scaling_factor(self.volume_vapor, 1e-2)
+        iscale.set_scaling_factor(self.liquid_phase.rate_reaction_generation, 1e4)
+        iscale.set_scaling_factor(self.liquid_phase.mass_transfer_term, 1e2)
+        iscale.set_scaling_factor(self.liquid_phase.rate_reaction_extent, 1e4)
+        iscale.set_scaling_factor(self.liquid_phase.enthalpy_transfer, 1e0)
+        iscale.set_scaling_factor(self.liquid_phase.volume, 1e-2)
+        iscale.set_scaling_factor(self.electricity_consumption, 1e0)
```
adam-a-a, line 1015: Shouldn't these be handled in the scaler class?

watertap/flowsheets/full_water_resource_recovery_facility/test/test_BSM2_P_extension.py
```diff
@@ -56,71 +56,71 @@ def test_solve(self, system_frame):
             0.24219, rel=1e-3
         )
```
adam-a-a: Would it make sense to check the condition number at the flowsheet level between old scaling approach and new?

adam-a-a: Then, every time another PR impacts scaling at this flowsheet level, we can track whether condition number improves or not as we go.

watertap/property_models/unit_specific/anaerobic_digestion/tests/test_modified_adm1_reaction.py
MarcusHolly: This change is not needed. It may have been necessary at some point while I was working on this PR, but I'll undo it

watertap/unit_models/anaerobic_digester.py
MarcusHolly: Andrew mentioned that we should not mix and match the new and old scaling tools. In other words, the scaler class should not make use of iscale. This is the reason why I've moved all of the `iscale.set_scaling_factor` to the `calculate_scaling_factors` function (in this PR and in others), which won't be called when exclusively using the new scaling tools.

If your question is why don't I set these default scaling factors in the ADScaler class, it's because I figured these would make more sense as user-defined scaling factors. I know I did do some testing with having default scaling factors for these variables in the new scaler class, but I forget how thoroughly I played around with this. Currently, the only variable scaled by default in the new scaler class is volume.

MarcusHolly:
Noting that the ScalingProfiler cannot currently be applied to the AD unit model since the ADM1 and modified ADM1 reaction packages have potential divisions by zero (see `def rule_I_pH_aa(self)`).

adam-a-a:
> Noting that the ScalingProfiler cannot currently be applied to the AD unit model since the ADM1 and modified ADM1 reaction packages have potential divisions by zero (see `def rule_I_pH_aa(self)`).

Interesting note. What happens exactly? Does ScalingProfiler do some sort of check for potential division by zero beforehand and raise an exception, or do you get back some exception that occurred having to do with division by zero? It’d be good to elaborate on this.

MarcusHolly:
> > Noting that the ScalingProfiler cannot currently be applied to the AD unit model since the ADM1 and modified ADM1 reaction packages have potential divisions by zero (see `def rule_I_pH_aa(self)`).
> 
> Interesting note. What happens exactly? Does ScalingProfiler do some sort of check for potential division by zero beforehand and raise an exception, or do you get back some exception that occurred having to do with division by zero? It’d be good to elaborate on this.

This is the traceback:

```
  File "C:\GitHub\watertap\watertap\watertap\unit_models\tests\profiler_sandbox_AD.py", line 187, in <module>
    output = main()
             ^^^^^^
  File "C:\GitHub\watertap\watertap\watertap\unit_models\tests\profiler_sandbox_AD.py", line 181, in main
    results = sp.report_scaling_profiles()
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\mholl\miniconda3\envs\watertap\Lib\site-packages\idaes\core\scaling\scaler_profiling.py", line 166, in report_scaling_profiles
    results = self.profile_scaling_methods()
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\mholl\miniconda3\envs\watertap\Lib\site-packages\idaes\core\scaling\scaler_profiling.py", line 152, in profile_scaling_methods
    results[case] = self.run_case(meth, **margs)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\mholl\miniconda3\envs\watertap\Lib\site-packages\idaes\core\scaling\scaler_profiling.py", line 246, in run_case
    manual = self._run_scenario(
             ^^^^^^^^^^^^^^^^^^^
  File "C:\Users\mholl\miniconda3\envs\watertap\Lib\site-packages\idaes\core\scaling\scaler_profiling.py", line 289, in _run_scenario
    self._apply_scaling(m, scaling_method, block_based=block_based, **kwargs)
  File "C:\Users\mholl\miniconda3\envs\watertap\Lib\site-packages\idaes\core\scaling\scaler_profiling.py", line 281, in _apply_scaling
    scaling_method(c, **kwargs)
  File "C:\Users\mholl\miniconda3\envs\watertap\Lib\site-packages\idaes\core\scaling\custom_scaler_base.py", line 505, in scale_constraint_by_nominal_derivative_norm
    value(
  File "C:\Users\mholl\miniconda3\envs\watertap\Lib\site-packages\pyomo\common\numeric_types.py", line 382, in value
    tmp = obj(exception=True)
          ^^^^^^^^^^^^^^^^^^^
  File "C:\Users\mholl\miniconda3\envs\watertap\Lib\site-packages\pyomo\core\expr\base.py", line 118, in __call__
    return visitor.evaluate_expression(self, exception)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\mholl\miniconda3\envs\watertap\Lib\site-packages\pyomo\core\expr\visitor.py", line 1301, in evaluate_expression
    ans = visitor.dfs_postorder_stack(exp)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\mholl\miniconda3\envs\watertap\Lib\site-packages\pyomo\core\expr\visitor.py", line 936, in dfs_postorder_stack
    ans = self.visit(_obj, _result)
          ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\mholl\miniconda3\envs\watertap\Lib\site-packages\pyomo\core\expr\visitor.py", line 1187, in visit
    return node._apply_operation(values)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\mholl\miniconda3\envs\watertap\Lib\site-packages\pyomo\core\expr\numeric_expr.py", line 1124, in _apply_operation
    return result[0] / result[1]
           ~~~~~~~~~~^~~~~~~~~~~
ZeroDivisionError: float division by zero

```

dallan-keylogic (approved):
LGTM

agarciadiego (approved):

watertap/unit_models/anaerobic_digester.py
```diff
@@ -836,6 +1001,21 @@ def calculate_scaling_factors(self):
...
+        for i, c in self.ad_performance_eqn.items():
+            iscale.constraint_scaling_transform(c, 1e2)
+
         # TODO: improve this later; for now, this resolved some scaling issues for modified adm1 test file
         if "S_IP" in self.config.liquid_property_package.component_list:
```
adam-a-a, line 1020: we should revisit these (small scaling factors applied to S_IP and S_IN) at some point. Could be possible that these default factors render the scaled values small enough to satisfy constraint residual tolerances erroneously; i.e., solver reports optimal convergence but the real constraint is not satisfied within tolerance.

Not saying that this IS the case but that it could be a possibility that we should rule out.

watertap/unit_models/tests/test_anaerobic_digester.py
```diff
@@ -238,3 +250,365 @@ def configure(self):
...
+        m.fs.unit.liquid_outlet.temperature.fix(308.15)
+
+        iscale.calculate_scaling_factors(m.fs.unit)
+
+        # Check condition number to confirm scaling
+        sm = TransformationFactory("core.scale_model").create_using(m, rename=False)
+        jac, _ = get_jacobian(sm, scaled=False)
+        assert (jacobian_cond(jac=jac, scaled=False)) == pytest.approx(
+            2.36919186521693e14, rel=1e-3
+        )
```
adam-a-a, line 464: Any reason why we don't just add this bit to the older test in the beginning of the file? Here you are rebuilding the same model and specifying all over again

watertap/unit_models/anaerobic_digester.py
MarcusHolly: Sure, but just to provide some additional context, this particular line is from the old scaling routine. Rather than just remove the iscale functionality, I figured I'd leave it intact so that we can toggle between the old (iscale) and new (scaler objects) scaling routines when updating the BSM2 flowsheets

watertap/unit_models/tests/test_anaerobic_digester.py
```diff
@@ -238,3 +250,365 @@ def configure(self):
...
+        m.fs.unit.inlet.anions[0].fix(0.02)
+
+        m.fs.unit.volume_liquid.fix(3400)
+        m.fs.unit.volume_vapor.fix(300)
+        m.fs.unit.liquid_outlet.temperature.fix(308.15)
+
+        iscale.calculate_scaling_factors(m.fs.unit)
+
+        # Check condition number to confirm scaling
+        sm = TransformationFactory("core.scale_model").create_using(m, rename=False)
```
adam-a-a, line 460: I thought this transformation was only needed for the new scaler approach, not the old one. Is there a reason for this, or is this just an unintended copy/paste that can be removed?

watertap/unit_models/anaerobic_digester.py
```diff
@@ -67,18 +67,199 @@
 from idaes.core.util.constants import Constants
 from idaes.core.util.exceptions import ConfigurationError, InitializationError
 from idaes.core.util.tables import create_stream_table_dataframe
+from idaes.core.scaling import CustomScalerBase, ConstraintScalingScheme
 
 from watertap.costing.unit_models.anaerobic_digester import cost_anaerobic_digester
 
 __author__ = "Alejandro Garciadiego, Andrew Lee, Xinhong Liu"
```
adam-a-a:
```suggestion
__author__ = "Alejandro Garciadiego, Andrew Lee, Marcus Holly, Xinhong Liu"
```

watertap/unit_models/tests/test_anaerobic_digester.py
MarcusHolly: I wanted to compare the condition number in both cases with just the default scaling (no user-specified scaling)

watertap/unit_models/tests/test_anaerobic_digester.py
MarcusHolly: Technically it's not "needed" in any approach. It's just used to clearly distinguish the scaled from the unscaled model. I just added it here for the sake of consistency to make the comparison as one-to-one as possible. TLDR: I think it should either be present in both cases or in neither of them.

watertap/unit_models/anaerobic_digester.py
```diff
@@ -67,18 +67,199 @@
...
+            source_state=model.liquid_phase.properties_in,
+            overwrite=overwrite,
+        )
+
+        self.call_submodel_scaler_method(
+            submodel=model.liquid_phase.properties_out,
+            method="variable_scaling_routine",
+            submodel_scalers=submodel_scalers,
+            overwrite=overwrite,
+        )
```
adam-a-a, line 124: I assume that this uses the ASM2D scaler class and would sidestep that questionable default scaling factors for S_IP and S_IN that I flagged (later on in PR).

watertap/flowsheets/full_water_resource_recovery_facility/test/test_BSM2_P_extension.py
```diff
@@ -234,3 +248,13 @@ def test_costing(self, system_frame):
         assert value(m.fs.costing.total_operating_cost) == pytest.approx(
             836020.408, rel=1e-3
         )
+
+    @pytest.mark.component
+    def test_condition_number(self, system_frame):
```
adam-a-a, line 253: I am pretty sure this is the case but will ask anyway- we can't mix and match application of the "old" scaling with the "new" scaling within a flowsheet, right? That is, you can't apply the ADscaler to the AD and calculate_scaling_factors everywhere else, right?  If you could, then we could apply the new scaler class to the AD and see its impact on the flowsheet condition number.

adam-a-a (commented):
Just a few comments to respond to before I approve

watertap/unit_models/anaerobic_digester.py
MarcusHolly: It uses whatever scaler class the user specifies. For example, in my draft PR where I started implementing the scalers into BSM2, this is what it would look like. The questionable scaling factor you pointed out would only be used if they called iscale.calculate_scaling_factors
image

watertap/unit_models/tests/test_anaerobic_digester.py
adam-a-a: Hmmm... I do think it is a necessity for the new scaling approach, based on my recollection from what @andrewlee94 said when he gave a workshop on this. I could be wrong though.

watertap/unit_models/tests/test_anaerobic_digester.py
adam-a-a: can you add a check on condition number to the other case that includes user-specified scaling? would just be a good benchmark to have for improving upon later

watertap/flowsheets/full_water_resource_recovery_facility/test/test_BSM2_P_extension.py
MarcusHolly: I forget the logic, but I recall Andrew mentioning that we should not mix and match the scaling tools. That being said, I can look into this when I have the time just to be sure.

watertap/unit_models/tests/test_anaerobic_digester.py
MarcusHolly: In my current draft PR, I do not perform this transformation with the new scaling tools. In other versions of the updated BSM2 flowsheet that I have stored locally, I have tried using this implementation. Imo, it makes it more difficult to diagnose issues because some variables are coming up as badly scaled when they shouldn't be. For example, the following would come up as a poorly scaled variable: "value = 1.5, scaling_factor = 1000", where the scaling factor was applied before the transformation and the value is the scaled value of that variable. Hopefully I explained that well enough. Regardless, I may consider using this transformation when I have the flowsheet in a good-enough place, but it doesn't inherently help the system solve easier (I've actually had more trouble getting the flowsheet to converge when doing this transformation), and it makes it more difficult to look at the model diagnostics

MarcusHolly: Just as an FYI, after some discussions with Doug offline, I'll transition to using the flowsheet that does apply this transformation.

watertap/unit_models/tests/test_anaerobic_digester.py
```diff
@@ -238,3 +257,365 @@ def configure(self):
...
+            m.fs.unit,
+            submodel_scalers={
+                m.fs.unit.liquid_phase.properties_in: ADM1PropertiesScaler,
+                m.fs.unit.liquid_phase.properties_out: ADM1PropertiesScaler,
+                m.fs.unit.liquid_phase.reactions: ADM1ReactionScaler,
+            },
+        )
+
+        # Check condition number to confirm scaling
+        sm = TransformationFactory("core.scale_model").create_using(m, rename=False)
```
adam-a-a, line 532: This seems like a more painful API for scaling...the more I see it in action.

watertap/unit_models/tests/test_anaerobic_digester.py
```diff
@@ -238,3 +257,365 @@ def configure(self):
...
+        # Apply scaling to unscaled variables
+        for var in m.fs.component_data_objects(Var, descend_into=True):
+            if "conc_mass_comp" in var.name:
+                sb.set_variable_scaling_factor(var, 1e1)
+            if "conc_mol" in var.name:
+                sb.set_variable_scaling_factor(var, 1e2)
+            if "reaction_rate" in var.name:
+                sb.set_variable_scaling_factor(var, 1e6)
+
+            sb.set_variable_scaling_factor(m.fs.unit.hydraulic_retention_time[0], 1e-6)
```
adam-a-a, line 604: Why do this? Is this just the "new" equivalent to what we were doing with the old way of scaling (i.e., create a ScalerBase and set variable scaling as desired to certain variables before applying the unit's Scaler)?

adam-a-a (approved):
LGTM

watertap/unit_models/tests/test_anaerobic_digester.py
MarcusHolly: Pretty much, although this should probably be applied after the unit's Scaler since we want the latter to take priority. The way things are currently set up, the unit scaler's config option for overwrite is set to False, so it will not overwrite any pre-existing scaling factors.

PR #1536, opened 2024-12-02
Simple OD Nanofiltration model

andrewlee94:
## Summary/Motivation:
The existing Nanofiltration_ZO model does not preserve electoneutrality in the outlet streams.

Additionally, the MCAS property package does not have a new type `Initializer` or `Scaler`.

- New `Nanofiltration0D` unit model with user provided recovery fractions and option to maintain electoneutrality.
- Update MCAS properties to use `BlockTriangularizationInitializer` by default
- Adds new `MCASScaler` for use the new scaling API.

### Legal Acknowledgement


watertap/unit_models/nanofiltration0D.py
```diff
@@ -0,0 +1,654 @@
...
+        for c in model.component_data_objects(Constraint, descend_into=False):
+            self.scale_constraint_by_nominal_value(
+                c,
+                scheme=ConstraintScalingScheme.inverseMaximum,
+                overwrite=overwrite,
+            )
+
+
+@declare_process_block_class("Nanofiltration0D")
+class Nanofiltration0DData(UnitModelBlockData):
```
adam-a-a: My high-level question before diving in more deeply is-- why do we need another NF model class? Could we instead just modify the existing NF_ZO model and potentially rename if needed? Or could we delete NF_ZO and replace with this? 

My concern is that all of the different NF files that we have now can already be confusing, and adding another separate NF file will add to that confusion. So it'd be best if we could consolidate.

andrewlee94: After the fact, I am asking myself if it would have been better to merge them. In part I did this to ensure backward compatibility as this model makes a few additional assumptions (both to simplify and to enforce electroneutrality).

docs/technical_reference/unit_models/nanofiltration_0D.rst
```diff
@@ -0,0 +1,72 @@
...
+.. currentmodule:: watertap.unit_models.nanofiltration_0D
+
+Degrees of Freedom
+------------------
+The ``Nanofiltration0D`` model has the following degrees of freedom
+
+   * inlet state
+   * permeate pressure
+   * solvent recovery fraction (``solvent_recovery``)
+   * solute recovery fractions (``solute recovery``) for all solutes EXCEPT the one identified as as the electroneutrality species.
```
adam-a-a:
```suggestion
   * solute recovery fractions (``solute recovery``) for all solutes EXCEPT the one identified as the electroneutrality species.
```

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
```diff
@@ -2485,3 +2507,625 @@ def test_automatic_charge_mw_population():
...
+            defined_state=True, tol=1e-2, adjust_by_ion="Cl_-"
+        )
+
+        metadata = m.fs.properties.get_metadata().properties
+        for v in metadata.list_supported_properties():
+            getattr(stream[0], v.name)
+            assert stream[0].is_property_constructed(v.name)
+
+        initializer = stream.default_initializer()
+        initializer.initialize(stream)
```
adam-a-a, line 2581: What are your thoughts on asserting a particular record of the initializer? For example, we could use some of the Diagnostic Toolbox functionality to record numerical issues after having applied the initializer. The main idea being that we can leave some form of a benchmark behind in the test, and if we attempt to refine the initializer (or compare with another initializer) in the future, we'd have some sort of metric to compare against and try to improve on. 

The first thing I'd want to know is how does the outcome of the initializer compare with the old `initialize` method?

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
```diff
@@ -2485,3 +2507,625 @@ def test_automatic_charge_mw_population():
...
+        scaler = stream.default_scaler()
+
+        set_scaling_factor(stream[0].flow_mass_phase_comp["Liq", "H2O"], 1)
+        set_scaling_factor(stream[0].flow_mass_phase_comp["Liq", "Ca_2+"], 1e4)
+        set_scaling_factor(stream[0].flow_mass_phase_comp["Liq", "SO4_2-"], 1e3)
+        set_scaling_factor(stream[0].flow_mass_phase_comp["Liq", "Na_+"], 1e2)
+        set_scaling_factor(stream[0].flow_mass_phase_comp["Liq", "Cl_-"], 1e2)
+        set_scaling_factor(stream[0].flow_mass_phase_comp["Liq", "Mg_2+"], 1e3)
+
+        scaler.scale_model(stream[0])
```
adam-a-a, line 2666: Similar idea here but probably easier to do-- I am unsure what metric(s) would be best to "benchmark" against, but perhaps asserting the condition number, for example, would be one way of comparing against the performance of the former scaling approach as well as any new scalers (or changes to this one) that might be introduced later.

I am mainly looking for a way to avoid overlooking any potential degradation in model performance/stability as we continue to introduce these scaler classes. Maybe the ScalingProfiler can be used to some extent for this, but I don't think that could apply to comparisons against the old method-based approach for scaling.

watertap/unit_models/nanofiltration_0D.py
```diff
@@ -0,0 +1,675 @@
...
+                c,
+                scheme=ConstraintScalingScheme.inverseMaximum,
+                overwrite=overwrite,
+            )
+
+
+@declare_process_block_class("Nanofiltration0D")
+class Nanofiltration0DData(UnitModelBlockData):
+    """
+    Zero order nanofiltration model.
```
adam-a-a:
```suggestion
    Zero dimensional nanofiltration model.
```
This brings me to my main comment on this PR: we should only have one NF file or the other between the ZO and 0D forms. So, I think it would make sense to get rid of the ZO formulation and incorporate anything from it that is not included in the 0d formulation here.

watertap/unit_models/nanofiltration_0D.py
```diff
@@ -0,0 +1,675 @@
...
+    CONFIG.declare(
+        "include_temperature_equality",
+        ConfigValue(
+            default=True,
+            domain=Bool,
+            description="Whether to include temperature equality constraints",
+            doc="""Argument indicating whether temperature equality should be
+            inlcluded. If this is False, no energy balance constraint will be
+            written.""",
+        ),
```
adam-a-a: these config options deviate from our typical "standard" and I suppose that is partly because you aren't using a ControlVolume0D in the model. Regardless, I would think that you could at least maintain options like "has_pressure_change". With regard to isothermal constraints, we have an isothermal ControlVolume with config options that allow for doing what "include_temperature_equality" enables.

watertap/unit_models/nanofiltration_0D.py
adam-a-a: Actually, it may be better to just incorporate additions made in NF_0D here to the existing NF_ZO formulation (and rename ZO to 0D). I say this after observing deviations in config options and the relation of those deviations to the omission of a ControlVolume.

watertap/unit_models/nanofiltration_0D.py
```diff
@@ -0,0 +1,675 @@
...
+            )
+
+        # Check that the electroneutrality ion is a valid monovalent ion
+        if self.config.electroneutrality_ion is not None:
+            try:
+                bal_ion = prop_params.get_component(self.config.electroneutrality_ion)
+                # Check that balancing ion is monovalent
+                if (
+                    not hasattr(bal_ion.config, "charge")
+                    or abs(bal_ion.config.charge) > 1
```
adam-a-a, line 546: Why must we be restricted to a monovalent ion as the ion of choice?

watertap/unit_models/nanofiltration_0D.py
```diff
@@ -0,0 +1,675 @@
...
+            domain=str,
+            description="Balancing ion to use to maintain electroneutrality",
+            doc="""Name of ion to use to maintain electroneutrality in
+            permeate stream. If an ion is provided, the split fraction
+            for this ion will be solved implicitly to ensure electroneutrality
+            is maintained in the permeate stream. If None, user must provide
+            spl;it fractions for all monovalent ions and electroneutrality is
+            not guaranteed..""",
+        ),
+    )
```
adam-a-a: I am thinking that you do not need this option because MCAS will handle electroneutrality for you with the "assert_electroneutrality" method, where you can choose your target ion for rebalancing if needed.

adam-a-a: I think I see why you need this now. If one wants to fix split fractions for each ion, one would need to ensure electroneutrality is maintained in the permeate, and choosing the adjustable ion would allow for that to occur. Is that the main intention?

docs/technical_reference/unit_models/nanofiltration_0D.rst
adam-a-a: What happens when a user sets a recovery fraction for the ion selected as the one to balance electroneutrality?

docs/technical_reference/unit_models/nanofiltration_0D.rst
```diff
@@ -0,0 +1,72 @@
...
+
+Variables
+---------
+
+.. csv-table::
+   :header: "Description", "Symbol", "Variable Name", "Index", "Units"
+
+   "Solvent recovery", ":math:`R_{solvent}`", "solvent_recovery", None, ":math:`\text{dimensionless}`"
+   "Solute recovery", ":math:`R_j`", "solute_recovery", [j], ":math:`\text{dimensionless}`"
+   "Multivalent ion recovery", ":math:`R_{multi}`", "multivalent_recovery", None, ":math:`\text{dimensionless}`"
```
adam-a-a: I think solute rejection is more commonly used in practice. The ZO formulation includes rejection, and I think you could keep both when merging the two models.

watertap/unit_models/tests/test_nanofiltration_0D.py
```diff
@@ -0,0 +1,1030 @@
...
+        m.fs.unit.inlet.flow_mol_phase_comp[0, "Liq", "Ca_2+"].fix(0.00955)
+        m.fs.unit.inlet.flow_mol_phase_comp[0, "Liq", "Mg_2+"].fix(0.5808)
+        m.fs.unit.inlet.flow_mol_phase_comp[0, "Liq", "SO4_2-"].fix(0.02225)
+        m.fs.unit.inlet.flow_mol_phase_comp[0, "Liq", "Cl_-"].fix(1.61977)
+        m.fs.unit.inlet.flow_mol_phase_comp[0, "Liq", "Na_+"].fix(0.48357)
+        m.fs.unit.inlet.temperature[0].fix(298.15)
+        m.fs.unit.inlet.pressure[0].fix(4e5)
+
+        m.fs.unit.solvent_recovery.fix()
+        m.fs.unit.solute_recovery.fix()
```
adam-a-a: I think there be a test to check what happens when a user tries fixing solute_recovery for the electroneutrality balancing ion. Preferably, an exception stating that recovery cannot be fixed for the electroneutrality ion would arise.

watertap/unit_models/tests/test_nanofiltration_0D.py
```diff
@@ -0,0 +1,1030 @@
...
+        ) == pytest.approx(0.9 * 0.48357, rel=1e-5)
+        assert value(mcas_case.fs.unit.permeate.temperature[0]) == pytest.approx(
+            298.15, rel=1e-5
+        )
+        assert value(mcas_case.fs.unit.permeate.pressure[0]) == pytest.approx(
+            101325, rel=1e-5
+        )
+
+    @pytest.mark.component
+    def test_connservation(self, mcas_case):
```
adam-a-a:
```suggestion
    def test_conservation(self, mcas_case):
```

watertap/unit_models/nanofiltration_0D.py
adam-a-a: Another thought on this--suppose I have some performance data for my membrane, and I know what the rejection rate (or alternatively solute recovery) is for each ion that I am interested in and have characterized in my solution composition. There could be the case that electroneutrality is not satisfied due to some unmeasured constituent's contribution.

Is your solution to this case to essentially not define any electroneutrality ion, thereby disregarding electroneutrality all together in the permeate? In that case, I wonder if we should log a warning for the user, where we'd check if electroneutrality would be satisfied in the permeate, and provide the warning in the case the electroneutrality would be violated. This way, a user who is already aware of the violation and expects it can wave off the warning, while the naive user who is actually making an unintentional error and would eventually come to the realization that they should balance electroneutrality would get some early notice.

Perhaps this could be injected into the initialization routine (or ignored because we don't know when the user might decide to fix/unfix split fractions.)

adam-a-a (dismissed):
The main thing I think we need to do here is consolidate the two NF models (ZO and this new 0D). After thinking it over, I think the way we should do this is incorporate new changes to the former ZO model (and rename ZO to be 0D) but omit some new config options that deviate from "the standard".

I also had some questions and comments concerning electroneutrality.

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
andrewlee94: I am not quite sure what you would be asserting in this case. What is the outcome of an initialization in the general sense? One might say that it is converging the model to a given tolerance, but that is not quite true as it only needs to get it to the point that a full solver can take over (although many of our routines include that as the final step).

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
andrewlee94: For that you want to look at this: https://github.com/IDAES/idaes-compatibility/tree/main/src/idaes_compatibility/robustness

Rather than look at some numerical characteristic of the model, you want to look at how well the solver performed at solving it. Additionally, it cannot be a single point test, but needs to cover a range of conditions.

watertap/unit_models/nanofiltration_0D.py
andrewlee94: I did debate this whilst writing the model, and decided to not be overly constrained by standards and instead focus on what made sense for the case at hand. Config option names are easily changed at least, but my initial feeling is that a control volume is unnecessary complications in this case.

watertap/unit_models/nanofiltration_0D.py
andrewlee94: Mainly because this model was written with the assumption that higher valance species would be rejected (and thus are not suitable for charge balancing).

watertap/unit_models/nanofiltration_0D.py
andrewlee94: Actually, my response in that case is that you need to do data reconciliation to get an electroneutral solution, otherwise your data is not correct.

The model does support `None` for the electroneutrality species to allow you to forgo that constraint (and replaces it with a standard split fraction) for cases where the user wants to do that. As for doing a check on electroneutrality, to me that is a post-check the user should be doing, as there are too many edge cases that could occur.

docs/technical_reference/unit_models/nanofiltration_0D.rst
andrewlee94: There is no split fraction for the electroneutrality ion - the model does not even create one so there is no risk of over-specifying the model

docs/technical_reference/unit_models/nanofiltration_0D.rst
andrewlee94: I decided to only include one of rejection or recovery to avoid any chance of over-specifying the problem - it would be easy to switch to rejection if that would be a better choice. I would generally discourage having both, as someone will attempt to fix both of them and wonder why the model fails to solve.

watertap/unit_models/tests/test_nanofiltration_0D.py
andrewlee94: As mentioned above, it is not possible to do this

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
adam-a-a: TBH- this thought was geared more towards the Scaler originally, but then I figured that we should think about doing some form of a check on Initializers although it is not apparent what the best way to go about it would be, or if it would be worthwhile for something like a property block.  

One idea- even though it may not be the best approach- is to check the approximate number of iterations before convergence, which could be a pain between the different runners. Something like that, or we can set a low number on max_iter and check the state of the model. But again--now that I think about it- this probably makes sense for a flowsheet or a complex model and not so much for a more basic model.

Overall though, the main question that I wanted to answer was -- how can we compare two initializers/initialization methods at this level to decide on whether one is superior (for the tested use case)?

andrewlee94: That is not easy to define. For one, determining the number of iterations is difficult to start with (as the solvers do not report it), and sometime more iterations are acceptable if it results in greater robustness across a range of conditions. That perhaps is the bigger point - it is hard to mark one initializer as superior to another without considering what they were designed to do (and this is ultimately why IDAES moved to class-based initializers to that we could have multiple options if required).

One thing I can say however is that the number of iterations should be fairly consistent across runners if you have a well scaled problem (but bets are off is the scaling is not good).

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
adam-a-a: That would be perfect! I didn't know about this IPOPTConvergenceAnalysis class. Didn't explore its full functionality yet, but if I am understanding correctly, then I would say we should test a range of conditions to check how the scaler affects convergence.

I would even say that we can think about a way to apply that to probing how "good" an initializer is as well, granted we'd have to follow up the initialization with a solve if it wasn't baked in already.

watertap/unit_models/nanofiltration_0D.py
adam-a-a: If we didn't already have the NF_ZO model, then I would agree. Since we do though, I think the merging of the two should resolve this, and we can just keep the "standard" config options from the ZO formulation while adding your electroneutrality option.

docs/technical_reference/unit_models/nanofiltration_0D.rst
adam-a-a: In that case, we should move to rejection.

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
andrewlee94: I will note the plan is/was to move IPOPTConvergenceAnalysis to use the parameter sweep tool once it is ready (I tried to make it mostly compatible).

docs/technical_reference/unit_models/nanofiltration_0D.rst
adam-a-a: Rejection as in `1- permeate_concentration/feed_concentration`. This should already be in the ZO model.

watertap/unit_models/tests/test_nanofiltration_0D.py
adam-a-a: Makes sense.

docs/technical_reference/unit_models/nanofiltration_0D.rst
```diff
@@ -0,0 +1,70 @@
...
+.. currentmodule:: watertap.unit_models.nanofiltration_0D
+
+Degrees of Freedom
+------------------
+The ``Nanofiltration0D`` model has the following degrees of freedom
+
+   * inlet state
+   * permeate pressure
+   * solvent recovery fraction (``solvent_recovery``)
+   * solute rejection fractions (``rekection_comp``) for all solutes EXCEPT the one identified as the electroneutrality species.
```
MarcusHolly:
```suggestion
   * solute rejection fractions (``rejection_comp``) for all solutes EXCEPT the one identified as the electroneutrality species.
```

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -120,6 +122,272 @@ class TransportNumberCalculation(Enum):
...
+class MCASScaler(CustomScalerBase):
+    DEFAULT_SCALING_FACTORS = {
+        "act_coeff_phase_comp": 1,
+        "debye_huckel_constant": 100,
+        "dens_mass_phase": 1e-3,
+        "dens_mass_solvent": 1e-3,
+        "diffus_phase_comp": 1e10,
+        "enth_mass_phase": 1e-5,
+        "flow_mass_phase_comp": 1e3,
+        "flow_mol_phase_comp": 1e2,
```
MarcusHolly, line 134: For Vars such as mass and molar flows that can vary widely in magnitude depending on the specific application, would you recommend setting default scaling factors and have the users manually override the scaling factors themselves, if necessary (as done here), or is would it be better to not set a default scaling factor at all for such variables. One of my concerns is that a user may not know (or simply forget) that they would need to set `overwrite=True` when trying to adjust these scaling factors themselves (in a flowsheet, for instance). For property packages, I've usually just been applying default scaling factors to flow_vol, temperature, and pressure

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -120,6 +122,272 @@ class TransportNumberCalculation(Enum):
...
+
+    def constraint_scaling_routine(
+        self, model, overwrite: bool = False, submodel_scalers: dict = None
+    ):
+        for c in model.component_data_objects(Constraint, descend_into=True):
+            self.scale_constraint_by_nominal_value(
+                c,
+                scheme=ConstraintScalingScheme.inverseMaximum,
+                overwrite=overwrite,
+            )
```
MarcusHolly, line 388: Are there cases where you wouldn't want to scale all of the constraints like this? When trying this for the anaerobic digester unit model (prob the most finicky model in BSM2), the flowsheet would fail to solve when doing this, so I ended up just scaling none of the constraints by default. I believe I also tried scaling a small subset of the constraints, but still ran into issues.

watertap/property_models/multicomp_aq_sol_prop_pack.py
andrewlee94: Users should always set manual scaling factors _before_ calling automated routines, as those routines will benefit from having better scaling guesses. I.e. they should not be overwriting defaults as they should have set the scaling factors before the defaults exist. Default values should always be seen as a last-ditch effort to provide scaling when the user cannot or does not provide information.

What is needed is good documentation of the tools and expected workflows that users can follow. If they fail to understand how the tools work and use different workflows, then any issues are on the user (we cannot anticipate and protect against every possible course of action).

watertap/property_models/multicomp_aq_sol_prop_pack.py
andrewlee94: In general, every constraint should be scaled however you should determine the correct approach to use for each constraint. I.e. you need to understand each constraint and what is the appropriate way to scale it. In this case, using inverse maximum is sufficient for all the constraints, but more complex models may require a mix of scaling schemes.

andrewlee94: Also, be careful not to conflate poor scaling with poorly posed constraints. Poorly posed constraints are inherently poorly scaled and cannot be fixed by just applying a scaling factor - the latest diagnostics tools have some new checks for poorly posed constraints.

adam-a-a: @andrewlee94 just to be sure, can you elaborate on which new checks you are referring to?

watertap/unit_models/nanofiltration_ZO.py
```diff
@@ -46,6 +47,12 @@
 _log = idaeslog.getLogger(__name__)
 
 
+@deprecated(
+    "The NanofiltrationZO model has been deprecated in favor of the Nanofiltration0D model. "
+    "The Nanofiltation0D model has most of the capabilities of the NanofiltrationZO model "
```
adam-a-a, line 52: Thanks for this! You say the new model has "most of the capabilities". Can you remind me what the new model might _not_ have that the deprecated version _does_? It's been a while since I reviewed this PR.

watertap/property_models/multicomp_aq_sol_prop_pack.py
andrewlee94: `display_constraints_with_mismatched_terms` and `display_constraints_with_canceling_terms`, along with `display_problematic_constraint_terms` for details on what is wrong in each constraint. These are shown as Cautions in the numerical issues report.

watertap/unit_models/nanofiltration_ZO.py
andrewlee94: Off the top of my head, it lacks the water/solvent flux and recovery calculations. That might be all however.

watertap/property_models/multicomp_aq_sol_prop_pack.py
MarcusHolly: Just as an FYI, none of the BSM2 constraints are poorly posed. But I will definitely re-visit the constraint scaling for BSM2 when I get the time.

watertap/unit_models/nanofiltration_0D.py
```diff
@@ -0,0 +1,732 @@
...
+        self.properties_permeate = self.config.property_package.build_state_block(
+            self.flowsheet().time, doc="Material properties at inlet", **tmp_dict_out
+        )
+
+        self.add_port("inlet", self.properties_in, doc="Inlet Port")
+        self.add_port("retentate", self.properties_retentate, doc="Retentate Port")
+        self.add_port("permeate", self.properties_permeate, doc="Permeate Port")
+
+        # NF separation variables
+        self.solvent_recovery = Var(self.flowsheet().time, initialize=0.8)
```
adam-a-a: This should have a doc string. Also, `recovery_vol` would be more in line with our naming convention (although we have used `recovery_vol_phase` indexed by `Liq`; saving effort of having to add an unnecessary index). Unless this is meant to be mass recovery and not volumetric, in which case `recovery_mass_solvent` is more in line with existing naming conventions. Can also consider `recovery_mass_comp` which would give mass recovery of solvent and solutes.

watertap/unit_models/nanofiltration_0D.py
```diff
@@ -0,0 +1,732 @@
...
+                    b.properties_in[t].temperature
+                    == b.properties_retentate[t].temperature
+                )
+
+            @self.Constraint(self.flowsheet().time, doc="Permeate temperature equality")
+            def permeate_temperature_equality(b, t):
+                return (
+                    b.properties_in[t].temperature
+                    == b.properties_permeate[t].temperature
+                )
```
adam-a-a, line 740: Any reason not to just apply the modified control volume? I was assuming that it was written in a way that one wouldn't need to write out the equality constraints (but been a while since last reviewed new control volume).

andrewlee94: Mainly because this model does not use any control volumes (but does require the `EnergyBalanceType.isothermal` option in the Enum). In short, as this is basically a specialised separator a control volume does not add much as we would need to write at least half the constraints manually anyway.

watertap/unit_models/nanofiltration_0D.py
andrewlee94: So, I agree the order of should be reversed to match standards but the basis is a little more vague. As a recovery fraction for a species as a whole, mass, mole and volume basis are equivalent, so I am not sure we should qualify it. Thus, my suggestion would be `recovery_solvent`.

adam-a-a: OK-- that works.

watertap/unit_models/nanofiltration_ZO.py
adam-a-a: If we are to replace the deprecated version with this one eventually (and remove the deprecated later), we should probably preserve solvent flux, recovery, and anything else that might be missing.

watertap/unit_models/nanofiltration_0D.py
andrewlee94: Fixed.

watertap/unit_models/nanofiltration_ZO.py
andrewlee94: The first catch I see there is that to have solvent flux requires additional user input that is not required for the simplest version (at a minimum, membrane area). So, it does open the question of whether we want to have that as a default inclusion, or if it should be an optional add-on.

Also, I am not sure how much time I will have to work on this right now. Someone else is welcome to take over if we decide it is important to add the other components.

ksbeattie:
@andrewlee94 any update on this?

andrewlee94:
@ksbeattie I am waiting on a response to one of @adam-a-a's comment above.

setup.py
```diff
@@ -67,7 +67,7 @@
     python_requires=">=3.9",
     install_requires=[
         # primary requirements for unit and property models
-        "idaes-pse >=2.7.0,<2.8.0rc0",
+        "idaes-pse @ https://github.com/watertap-org/idaes-pse/archive/2.7.0.dev.watertap.2025.01.21.zip",
```
adam-a-a: Need to update this

adam-a-a: @lbianchi-lbl reminder to check this out and ensure update to latest idaes release works here

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
adam-a-a: @andrewlee94 which parameter sweep tool? The original one or the one that came out of WaterTAP?

watertap/unit_models/nanofiltration_ZO.py
adam-a-a: I think that we should include flux, area, and other basic "essentials" that are in the existing NF model. 

We can have someone else ensure those additions are made and then merge this in.

kurbansitterley:
@adam-a-a @lbianchi-lbl I could not duplicate the macOS failure locally. 

All tests pass for me locally except `TestNanofiltration0DInitializer::test_initialize` where I get:

`FAILED watertap/unit_models/tests/test_nanofiltration_0D.py::TestNanofiltration0DInitializer::test_initialize - DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses`

Then, when I try to duplicate _that_ error outside of pytest, I get:

`2025-03-06 15:23:38 [INFO] idaes.init.fs.unit: Initialization Completed, optimal - <undefined>`

Prior to running this, I re-installed my environment with the `setup.py` on this PR.

Probably relevant information:
- MacBook Pro, 14-inch, 2021
- Apple M1 Pro
- macOS 14.7.1

@lbianchi-lbl do you know if the test is run with an Intel chip (vs. M1)?

🤷‍♂️

lbianchi-lbl:
@kurbansitterley  the check is indeed running on macOS Intel, specifically using IPOPT from `conda-forge` (as opposed to the HSL-enabled build available via `idaes get-extensions`. So it's not unexpected that this test might exhibit different behavior on an Apple Silicon machine (such as yours); however, it's still informative since we don't yet have Apple Silicon running in CI. If anything, this is a further confirmation that we should retire this CI check for good and replace with Apple Silicon CI jobs as per #1306.

kurbansitterley:
> @kurbansitterley the check is indeed running on macOS Intel, specifically using IPOPT from `conda-forge` (as opposed to the HSL-enabled build available via `idaes get-extensions`. So it's not unexpected that this test might exhibit different behavior on an Apple Silicon machine (such as yours); however, it's still informative since we don't yet have Apple Silicon running in CI. If anything, this is a further confirmation that we should retire this CI check for good and replace with Apple Silicon CI jobs as per #1306.

FWIW I don't know of any Mac-using WaterTAP devs with an Intel processor. So I am not sure if we are able to troubleshoot this issue.

M1s came out in 2020, one year after Python 3.8 was released, which we also stopped testing against, sooo... I support the retirement of this experimental test.

watertap/unit_models/nanofiltration_ZO.py
kurbansitterley: @adam-a-a do we also want to remove/deprecate the _other_ ZO NF model (i.e.,  `unit_models/zero_order/nanofiltration_zo.py`)? Once this is merged in, we would have four NF models, and two with the same name.

adam-a-a: @kurbansitterley because of other model dependencies, we planned to keep the other ZO formulation that you referred to and deprecate the duplicate one that the current PR's NF model is meant to replace here. Ultimately, we'd end up with one ZO model, this simple 0d formulation, and then the multicomponent DSPMDE model--also 0d but more complex than this formulation.

Does that make sense?

kurbansitterley: Yes, that makes sense. I assumed the other ZO model is used in some flowsheets somewhere but didn't check.

kurbansitterley:
@andrewlee94 @adam-a-a I added `area` as a Var and `flux_vol_solvent`, `recovery_mass_phase_comp`, and `recovery_vol_phase` as Expressions.

I tried to do as much of a direct translation of these from `nanofiltration_ZO` as I could. Since this model doesn't use a `ControlVolume0D`, there is not mass transfer term, so the actual implementation is a bit different. Also, as @andrewlee94 said, it required adding `area` as a Var. I think as a user if I was confronted with having to provide area as an input to a 0D model, I would be a bit confused. Especially since it serves only to report some performance metrics, I would support having it as an option rather than a required input to the model. Or maybe just as a Param.

@lbianchi-lbl I also added `requires_idaes_solver` to some of the previously failing tests, but I am happy to revert if we want to let them fail.

setup.py
kurbansitterley: All the commits I just made starting with `35d51f8` were made with idaes-pse=2.8.0.dev0 and these tests passing locally.

ksbeattie:
@andrewlee94, we're about to create a March release, how important to you that it make this release?

andrewlee94:
@ksbeattie I am not using this model right now, so this is more of an enhancement offering that you can accept (or reject) whenever you want.

docs/technical_reference/unit_models/nanofiltration_0D.rst
```diff
@@ -0,0 +1,75 @@
...
+
+.. _NF0D_variables:
+
+Variables
+---------
+
+.. csv-table::
+   :header: "Description", "Symbol", "Variable Name", "Index", "Units"
+
+   "Solvent recovery", ":math:`Q_{solvent}`", "solvent_recovery", None, ":math:`\text{dimensionless}`"
```
adam-a-a: Here's what the other NF model (proposed for deprecation) includes under variables:
```
   "Mass density of solvent", ":math:`\rho_{solvent}`", "dens_solvent", "[p]", ":math:`\text{kg/}\text{m}^3`"
   "Solvent volumetric flux across membrane", ":math:`J_{solv}`", "flux_vol_solvent", "[t, j]", ":math:`\text{m}^3\text{/m}^2\text{/s}`"
   "Membrane area", ":math:`A_m`", "area", "None", ":math:`\text{m}^2`"
   "Component recovery rate", ":math:`R_j`", "recovery_mass_phase_comp", "[t, p, j]", ":math:`\text{dimensionless}`"
   "Volumetric recovery rate", ":math:`R_{vol}`", "recovery_vol_phase", "[t, p]", ":math:`\text{dimensionless}`"
   "Observed solute rejection", ":math:`r_j`", "rejection_phase_comp", "[t, p, j]", ":math:`\text{dimensionless}`"
   "Mass transfer to permeate", ":math:`M_p`", "mass_transfer_phase_comp", "[t, p, j]", ":math:`\text{kg/s}`"
```
One of my main reasons for holding back on approving this was because I wanted this new model to reflect the candidate for deprecation, where logical and warranted. Since published work uses the candidate for deprecation, my hope was that we'd maintain compatibility as much as possible.

docs/technical_reference/unit_models/nanofiltration_0D.rst
```diff
@@ -0,0 +1,75 @@
...
+
+Equations
+---------
+
+Here :math:`F` represents component flowrate and :math:`Z_j` is the charge on species j.
+
+.. csv-table::
+   :header: "Description", "Equation"
+
+   "Component material balances", ":math:`F_{in, j} = F_{retentate, j} + F_{permeate, j}`"
```
adam-a-a: As a reference, the candidate for deprecation includes the following equations in its RST file:
```
   "Solvent mass transfer", ":math:`M_{p, solv} = A_m J_{solv} \rho_{solvent}`"
   "Component recovery rate",":math:`R_j = \frac{M_{p,j}}{M_{f,in,j}}`"
   "Volumetric recovery rate",":math:`R_{vol} = \frac{Q_{p}}{Q_{f,in}}`"
   "Observed solute rejection", ":math:`r_j = 1 - \frac{C_{p}}{C_{f,in}}`"
```

docs/technical_reference/unit_models/nanofiltration_0D.rst
kurbansitterley: Just for the record:

`dens_solvent` - comes from property package in 0D model
`flux_vol_solvent` - added as Expression in 0D model
`area` - added as Var in 0D model
`recovery_mass_phase_comp` - added as Expression in 0D model
`recovery_vol_phase` - added as Expression in 0D model
`rejection_phase_comp` - present as `rejection_comp` Var in 0D model
`mass_transfer_phase_comp` - not present in 0D model since ControlVolume is not used

lbianchi-lbl:
Waiting for @kurbansitterley to apply some changes and @adam-a-a to review.

watertap/unit_models/nanofiltration_0D.py
```diff
@@ -0,0 +1,788 @@
...
+                # TODO- come up with better way to handle different locations of mw_comp in property models (generic vs simple ion prop model)
+                if hasattr(b.properties_in[t].params, "mw_comp"):
+                    mw_comp = b.properties_in[t].params.mw_comp[j]
+                elif hasattr(b.properties_in[t], "mw_comp"):
+                    mw_comp = b.properties_in[t].mw_comp[j]
+                else:
+                    raise ConfigurationError("mw_comp was not found.")
+                return (
+                    -b.properties_permeate[t].flow_mass_phase_comp[p, j] * mw_comp
+                ) / (b.properties_in[t].dens_mass_phase[p] * b.area)
```
adam-a-a, line 765: Noting this to do for an issue

PR #1542, opened 2024-12-19
CSTR Re-Scaling

MarcusHolly:
## Summary/Motivation:
Adds the new IDAES scaling routine (scaler objects) to the CSTR & CSTR w/Injection unit models.

- Updates the scaling routine of the CSTR unit models
- Updates the associated testing files

### Legal Acknowledgement


MarcusHolly:
@TimBartholomew @adam-a-a @dallan-keylogic

Based on the scaling discussion yesterday, the latest commit added some preliminary testing using the Scaling Profiler for these three unit models using the new scaling tools (Scaler objects) and the old scaling tools (iscale). I think it might make the most sense for this and similar PRs of mine to get all the Scalers and tests in place while subsequent PRs can refine the tests and scaler objects as necessary (e.g. adjusting unit model variable/constraint_scaling_routine, adjusting property model variable/constraint_scaling_routine, solution pertubation when using the ScalingProfiler, etc.)

Any advice/comments on this and does it still make sense to keep tests making the overly simplistic comparison between the Jacobian condition numbers for the old and new scaling tools?

 If you all are still planning on meeting to discuss this topic "soon", I'd be happy to tag along as well.

adam-a-a:
> @TimBartholomew @adam-a-a @dallan-keylogic
> 
> Based on the scaling discussion yesterday, the latest commit added some preliminary testing using the Scaling Profiler for these three unit models using the new scaling tools (Scaler objects) and the old scaling tools (iscale). I think it might make the most sense for this and similar PRs of mine to get all the Scalers and tests in place while subsequent PRs can refine the tests and scaler objects as necessary (e.g. adjusting unit model variable/constraint_scaling_routine, adjusting property model variable/constraint_scaling_routine, solution pertubation when using the ScalingProfiler, etc.)
> 
> Any advice/comments on this and does it still make sense to keep tests making the overly simplistic comparison between the Jacobian condition numbers for the old and new scaling tools?
> 
> If you all are still planning on meeting to discuss this topic "soon", I'd be happy to tag along as well.

Great job on putting those changes so quickly! I agree with your thought on refining as needed in subsequent PRs. As for the comparison of condition numbers, I don't see any harm in keeping them for now since they give us some sort of benchmark.

watertap/unit_models/cstr_injection.py
```diff
@@ -64,12 +63,132 @@ class ElectricityConsumption(Enum):
...
+            submodel_scalers=submodel_scalers,
+            overwrite=overwrite,
+        )
+        self.call_submodel_scaler_method(
+            submodel=model.control_volume.reactions,
+            method="variable_scaling_routine",
+            submodel_scalers=submodel_scalers,
+            overwrite=overwrite,
+        )
+
```
adam-a-a, line 122: Is `injection` another variable that should be covered here? I was thinking this Scaler looks almost the same (if not the same) as that of the CSTR, but injection is what sets this model apart from the regular cstr.

watertap/unit_models/tests/test_aeration_tank.py
```diff
@@ -103,7 +119,6 @@ def test_config():
     assert m.fs.unit.config.property_package is m.fs.properties
     assert m.fs.unit.config.reaction_package is m.fs.reactions
     assert m.fs.unit.config.electricity_consumption == ElectricityConsumption.calculated
-    assert m.fs.unit.config.has_aeration
```
adam-a-a: Why was this line deleted?

watertap/unit_models/tests/test_cstr_injection.py
```diff
@@ -467,3 +482,642 @@ def test_error_without_oxygen():
...
+        m.fs.unit.inlet.conc_mass_comp[0, "S_O"].fix(0 * units.g / units.m**3)
+        m.fs.unit.inlet.conc_mass_comp[0, "S_NO"].fix(0 * units.g / units.m**3)
+        m.fs.unit.inlet.conc_mass_comp[0, "S_NH"].fix(23 * units.g / units.m**3)
+        m.fs.unit.inlet.conc_mass_comp[0, "S_ND"].fix(5 * units.g / units.m**3)
+        m.fs.unit.inlet.conc_mass_comp[0, "X_ND"].fix(16 * units.g / units.m**3)
+        m.fs.unit.inlet.alkalinity.fix(7 * units.mol / units.m**3)
+
+        m.fs.unit.volume.fix(500)
+        m.fs.unit.injection.fix(0)
+        m.fs.unit.injection[0, "Liq", "S_O"].fix(2e-3)
```
adam-a-a, line 655: perturbing injection rate of oxygen would be useful when using scaling profiler

watertap/unit_models/cstr_injection.py
MarcusHolly: For this to work, I would need to add something like the following code into the CSTR_InjectionScaler class, but I'm not sure how to get this working (how to get it to read the has_aeration config option and whether or not S_O or S_O2 exist)

```
        if self.config.has_aeration:
            # Assuming that the supplied property package doesn't have both S_O and SO2 to represent oxygen.
            if "S_O" in self.config.property_package.component_list:
                self.scale_variable_by_default(
                    model.control_volume.mass_transfer_term[0, "Liq", "S_O"], overwrite=overwrite
                )
            elif "S_O2" in self.config.property_package.component_list:
                self.scale_variable_by_default(
                    model.control_volume.mass_transfer_term[0, "Liq", "S_O2"], overwrite=overwrite
                )
            else:
                pass
```

MarcusHolly: From a theoretical perspective, I've only been adding default scaling for variables that don't usually vary by orders of magnitude. For instance, volume for these units will usually be on the order of 1e3 (based on historical usage). However, I'm not as confident with injection. Should we expect this variable to consistently be on the order of 1e-2 or will it vary depending on the application. If the latter, we should probably avoid applying a default scaling factor here.

watertap/unit_models/tests/test_aeration_tank.py
MarcusHolly: Must've been deleted unintentionally somehow. I've added it back

watertap/unit_models/tests/test_cstr_injection.py
MarcusHolly: Addressed

agarciadiego (approved):

watertap/unit_models/aeration_tank.py
```diff
@@ -30,12 +32,147 @@
...
+
+    This Scaler relies on the associated property and reaction packages,
+    either through user provided options (submodel_scalers argument) or by default
+    Scalers assigned to the packages.
+    """
+
+    DEFAULT_SCALING_FACTORS = {
+        "volume": 1e-3,
+        "hydraulic_retention_time": 1e-3,
+        "KLa": 1e-1,
```
adam-a-a, line 47: any reason why we wouldn't use the inverse of the variable value (assuming it was fixed)?

MarcusHolly: So basically why don't we use the autoscaling functions for these variables? In general, the autoscaling tools haven't been working great for me when applying them in a wide-sweeping sense like this. They're also just used not used very much in general. I haven't seen any examples of the new Scaler Objects using the autoscaling tools

dallan-keylogic: > any reason why we wouldn't use the inverse of the variable value (assuming it was fixed)?

Fixed variables are not visible to IPOPT at all. The only reason you might need a scaling factor for them is if some other quantity depended on it.

adam-a-a: That's true...forgot about that. What I really meant was "assuming the inverse of the variable value would be tolerable".

MarcusHolly: I don't think we're aiming for "perfect scaling". I'm rusty on the mathematical reasons why we'd want to avoid it, but I know it was discussed in one of Andrew's scaling documents. Does anybody happen to have/know where that up-to-date scaling document is?

adam-a-a: Unsure, but my only concern is that volume would be tied to define flowrate, which can be whatever the user wants to specify. So for a very large system, volumes would be larger, and the scaling factor might not be small enough (for example).

adam-a-a: Additionally, volume divided by volumetric flowrate equals HRT, so it would make more sense to set a scaling factor for volume that is based on the scaling factors for HRT and volumetric flowrate (for example).

Actually, since volumetric flowrate is a state variable in ASM/ADM models, the user should always provide scaling for that. Now, I think we can safely assume some default for HRT, based on typical HRT values. From that, we can compute the scaling factor for volume as scaling_factor_of_HRT * scaling_factor_of_volumetric_flowrate

MarcusHolly: Thsoe are valid concerns... the user would just have to update these scaling factors manually for a large system. To give some background on why I chose these values, 1e-3 was a good scaling factor for HRT and volume for the CSTRs in the BSM models. This implies that the magnitude of the flow would be on the scale of 1 (equivalent to no scaling factor), which checks out after looking at 2 CSTRs in the BSM2 flowsheet. For the sake of transparency, I could add a scaling factor of 1 to the list of default scaling factors for volumetric flow.

TLDR: I'm not opposed to what you're suggesting, but I'm not sure how I would implement it, especially given Doug's comment

dallan-keylogic: > I don't think we're aiming for "perfect scaling".

What Andrew meant by "perfect scaling" was scaling every variable using a scaling factor of `1/abs(value(var))`. We don't want to do that because it typically screws up the Jacobian scaling. The best example is something like enthalpy, where the variable value can pass through zero. If you scale based on an initial condition for which the enthalpy just happens to be a number close to zero, that causes major problems.

Variables that pass through zero aren't the only case where naive scaling like that causes problems, but they're the most obvious.

adam-a-a, line 47: Let's document these default scaling factors an issue in case they ever become suspects in a "failure to converge" crime scene. I won't hold this up anymore!

adam-a-a (approved):

PR #1543, opened 2024-12-19
Clarifier Re-Scaling

MarcusHolly:
## Summary/Motivation:
Adds the new IDAES scaling routine (scaler objects) to the clarifier unit model.

- Updates the scaling routine of the clarifier unit model
- Updates the associated testing files

### Legal Acknowledgement


luohezhiming (approved):

agarciadiego (approved):

watertap/unit_models/tests/test_clarifier.py
```diff
@@ -112,7 +121,14 @@ def build():
     iscale.set_scaling_factor(m.fs.unit.effluent_state[0.0].conc_mass_comp["S_O"], 1e7)
     iscale.set_scaling_factor(m.fs.unit.effluent_state[0.0].conc_mass_comp["S_NO"], 1e7)
 
-    iscale.calculate_scaling_factors(m.fs.unit)
+    # iscale.calculate_scaling_factors(m.fs.unit)
```
adam-a-a: delete
```suggestion
```

watertap/unit_models/tests/test_clarifier.py
```diff
@@ -283,3 +299,445 @@ def configure(self):
...
+            m.fs.unit.effluent_state[0.0].conc_mass_comp["S_NO"], 1e7
+        )
+
+        iscale.calculate_scaling_factors(m.fs.unit)
+
+        # Check condition number to confirm scaling
+        sm = TransformationFactory("core.scale_model").create_using(m, rename=False)
+        jac, _ = get_jacobian(sm, scaled=False)
+        assert (jacobian_cond(jac=jac, scaled=False)) == pytest.approx(
+            2.955746851e9, rel=1e-3
```
adam-a-a: This condition number actually appears to be better than the result from using the Scaler above--unless I am misinterpreting.

watertap/unit_models/tests/test_clarifier.py
```diff
@@ -283,3 +299,445 @@ def configure(self):
...
+                m.fs.unit.underflow_state: ASM1PropertiesScaler,
+                m.fs.unit.effluent_state: ASM1PropertiesScaler,
+            },
+        )
+
+        # Check condition number to confirm scaling
+        sm = TransformationFactory("core.scale_model").create_using(m, rename=False)
+        jac, _ = get_jacobian(sm, scaled=False)
+        assert (jacobian_cond(jac=jac, scaled=False)) == pytest.approx(
+            2.0028333e4, rel=1e-3
```
adam-a-a, line 494: Ok, this is clearly better. I am unsure what the first added condition number corresponds to. I will go and check.

watertap/unit_models/tests/test_clarifier.py
```diff
@@ -112,7 +121,14 @@ def build():
...
     iscale.set_scaling_factor(m.fs.unit.effluent_state[0.0].conc_mass_comp["S_NO"], 1e7)
 
-    iscale.calculate_scaling_factors(m.fs.unit)
+    # iscale.calculate_scaling_factors(m.fs.unit)
+
+    # Check condition number to confirm scaling
+    sm = TransformationFactory("core.scale_model").create_using(m, rename=False)
+    jac, _ = get_jacobian(sm, scaled=False)
+    assert (jacobian_cond(jac=jac, scaled=False)) == pytest.approx(
+        2.82857097e12, rel=1e-3
```
adam-a-a: What's the difference between this and the other condition number from when you used the calculate_scaling_factors approach? I assume this goes back to discussion with @dallan-keylogic, but it isn't totally clear to me.

adam-a-a: To clarify, I don't understand why applying the transformation here, versus using calculate_scaling_factors, results in different condition numbers, with this one at three orders of magnitude than the other result when calculate_scaling_factors is applied.

watertap/unit_models/tests/test_clarifier.py
```diff
@@ -283,3 +299,445 @@ def configure(self):
...
+Vars Only                || 1.359E+10 | Solved 3   || 1.356E+14 | Solved 1  
+Harmonic                 || 1.359E+10 | Solved 3   || 1.228E+02 | Solved 1  
+Inverse Sum              || 1.359E+10 | Solved 3   || 6.636E+03 | Solved 1  
+Inverse Root Sum Squares || 1.359E+10 | Solved 3   || 6.633E+03 | Solved 1  
+Inverse Maximum          || 1.359E+10 | Solved 3   || 6.636E+03 | Solved 1  
+Inverse Minimum          || 1.359E+10 | Solved 3   || 9.327E+01 | Solved 1  
+Nominal L1 Norm          || 1.359E+10 | Solved 3   || 1.817E+03 | Solved 1  
+Nominal L2 Norm          || 1.359E+10 | Solved 3   || 3.354E+03 | Solved 1  
+Actual L1 Norm           || 1.359E+10 | Solved 3   || 9.450E+01 | Solved 1  
+Actual L2 Norm           || 1.359E+10 | Solved 3   || 8.480E+01 | Solved 1
```
adam-a-a, line 618: remind me again--are the values here condition numbers?

adam-a-a: along with number of iterations to solve?

watertap/unit_models/tests/test_clarifier.py
MarcusHolly: The reason they differ by 3 orders of magnitude is because the e12 value does not use iscale.calculate_scaling_factors. If you were to call iscale.calculate_scaling_factors, it would be equivalent to the e9 value. I forget my rationale for leaving both of these tests in, but I assume I meant to delete one and just forgot. I'll delete one of them now to avoid redundancies since we should be calling calculate_scaling_factors when using iscale

watertap/unit_models/tests/test_clarifier.py
MarcusHolly: Yes and yes

watertap/unit_models/tests/test_clarifier.py
adam-a-a: Ah, so comparing unscaled with scaled I suppose.

MarcusHolly: Yea, but I don't think it's worth leaving the unscaled test in, right?

watertap/unit_models/tests/test_clarifier.py
adam-a-a: @MarcusHolly - it sort of seems like perfect scaling with inverse minimum, actual L1 norm, and actual L2 norm yield the lowest condition numbers. Something to keep in mind for future refinements to the Scaler

adam-a-a (approved):
LGTM - I had a comment based on the scaling profiler reports which highlight other constraint scaling approaches which reduce the condition number. Something to consider for future enhancements of Scaler classes.

PR #1548, opened 2025-01-03
Adding surrogate option for CPHSDM calculations

### LINKED ISSUE #1556: Merging GAC CPHSDM Model with Surrogates
In #1548, @hunterbarber wants to add the CPHSDM surrogates he has been using for a while to the GAC implementation as an option via CONFIG. This is a good idea. This issue memorializes some possibilities that were (briefly) discussed in the WaterTAP dev call on Feb 6, 2025.

The question (and the reason for the issue) is first, when to do this and second, how to handle the IX implementation, which is very nearly identical (if not totally identical, I would need to check) to the GAC implementation with surrogates. To avoid duplication of effort, the question is how to handle this, given this model could be used for either IX or GAC (though this would be a less conventional approach to modeling IX, though still valid).

A few suggestions were made :

- make a base class that the IX and GAC CPHSDM models would inherit from to accommodate any differences in implementation (if any).
- rename the model to something like `AdsorptionCPHSDM` to reflect the model's utility for two different technologies
- status quo; keep it as `GAC` and maybe note in the docs that it could be used for IX as well with the right parameters

(Note - there is nothing unique about adding the surrogates that makes the current implementation usable for an IX process provided the model is parameterized correctly and the IX operating conditions meet the assumptions for the CPHSDM model. But that is also the case for using CPHSDM for a GAC process).

Lastly, while this is not a pressing issue at the moment, the timetable for merging these changes should be such that we can point to this model (and the surrogates and the documentation) when the PFAS paper is eventually submitted and accepted.

hunterbarber:
## Fixes/Resolves:
Provide a surrogate to optionally bypass user input of empirical parameters for the GAC CPHSDM.
#1556

## Summary/Motivation:
Novel implementation for CPHSDM corresponding to pending publication.
I think the plan will be to write this for GAC, and then @kurbansitterley can translate the code over to IX.

- Add surrogates into the repo
- Create config option for input or surrogate use

### Legal Acknowledgement

TODO:

- [x] add details about surrogate option to documentation


kurbansitterley:
I think this a great idea (though I assume the actual implementation would be more involved since this is a draft).

@hunterbarber the IX implementation is nearly identical to the IX implementation - I had a thought that rather than having two nearly identical implementations for IX/GAC, we "merge" them into the same unit model called like `AdsorptionCPHSDM` or something like that. I am also happy to keep them separate if it is more clear for users (and it very well might be).

hunterbarber:
> I think this a great idea (though I assume the actual implementation would be more involved since this is a draft).
> 
> @hunterbarber the IX implementation is nearly identical to the IX implementation - I had a thought that rather than having two nearly identical implementations for IX/GAC, we "merge" them into the same unit model called like `AdsorptionCPHSDM` or something like that. I am also happy to keep them separate if it is more clear for users (and it very well might be).

@kurbansitterley Yes, I agree with a shared CPHSDM file. I was saying separate because I guessed it would be an easier way to roll it out. Basically making it work for either GAC or IX first, and then copying the method over to the other unit and generalizing the location of the model. If you have any suggestions of an easier way though, that also works.

kurbansitterley:
@adam-a-a surrogate functionality is added to GAC.

What are your thoughts on including the data and script used to create the surrogates? We don't have to test it, but it could be useful for users.

kurbansitterley (approved):
LGTM but I am not sure if this review counts.

ksbeattie:
@MarcusHolly & @avdudchenko will review this for Kurby/Hunter

MarcusHolly (approved):
LGTM

kurbansitterley (approved):

adam-a-a (approved):

avdudchenko (approved):
This looks good to me overall.

PR #1551, opened 2025-01-16
Create CODEOWNERS file

lbianchi-lbl:
### Legal Acknowledgement

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -47,8 +47,8 @@ environment*.yml @lbianchi-lbl
 /watertap/unit_models/MD/ @ElmiraShamlou @adam-a-a
 /watertap/unit_models/mvc/ @carson-tucker
 /watertap/unit_models/translators/ @MarcusHolly @luohezhiming @adam-a-a
-
 /watertap/unit_models/zero_order/ @kurbansitterley @adam-a-a @luohezhiming @MarcusHolly
+
 /watertap/unit_models/aeration_tank.py @adam-a-a @MarcusHolly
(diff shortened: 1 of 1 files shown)


MarcusHolly:
A few comments:

1.) Can we add Maojian (Morgan88888888) to the list of WaterTAP members? He's already in the IDAES repo.

2.) Some of us had a discussion offline about how comprehensive the lists should be. Ultimately we landed on including any relevant names. This will help with allocating responsibilities when people are offboarding and also provides some insight into areas of the repo where we could potentially use more support (if there's only 1-2 names listed). It's also just easier to trim down a list of excess names rather than trying to dig for more names. We're open to feedback on this tho.

andrewlee94:
@MarcusHolly When we did this for the IDAES side we decided to have a short list of names with the intent that these were the first points of contact. It was not expected that the code owners would fix everything that came up, but would rather know who would be able to fix the issue and help find the right person to do the job.

MarcusHolly:
> @MarcusHolly When we did this for the IDAES side we decided to have a short list of names with the intent that these were the first points of contact. It was not expected that the code owners would fix everything that came up, but would rather know who would be able to fix the issue and help find the right person to do the job.

Thanks for adding that context. Is there any inherent downside to including a more comprehensive list of names though? For what it's worth, the names are listed from most involved/knowledgeable to least involved/knowledgeable (from left to right). So I could easily trim down the lists if we want to stick to 1-3 names.

andrewlee94:
The biggest issue is dilution of responsibility - if you get an email saying a certain thing needs to be fixed and it only has your name and one other person, you are much more likely to take ownership than if there is a longer list.

ksbeattie (approved):
LGTM

PR #1553, opened 2025-01-30
Resolving Potential Evaluation Errors

### LINKED ISSUE #1547: Poorly formulated constraint in MCAS model
### Description

The MCAS property model has the following constraint formulation (line 1464):

```
def rule_mass_frac_phase_comp(b, p, j):
    return b.mass_frac_phase_comp[p, j] == b.flow_mass_phase_comp[p, j] / sum(
        b.flow_mass_phase_comp[p, j] for j in self.params.component_list
    )
```

This leads to a potential division by zero as flow goes to zero and thus would be better reformulated as:

```
def rule_mass_frac_phase_comp(b, p, j):
    return b.flow_mass_phase_comp[p, j] == b.mass_frac_phase_comp[p, j]  * sum(
        b.flow_mass_phase_comp[p, j] for j in self.params.component_list
    )
```

However, doing this results in issues with the scaling methods, primarily due to the `transform_property_constraints` function which assumes that all constraints are written in the form of `property = function()`. I suspect that this might have further reaching effects than just this.

### Motivation

Generally, constraints should be formulated in ways to avoid potential singularities, such as avoiding divisions where possible (unless the denominator is a simple constant). Doing so helps avoid numerical issues during solving and will generally make models more robust (and often better scaled).

### Possible Implementation

This issue effectively has two parts:

1. Looking for poorly formulated constraints and rewriting them (relatively easy, especially with the diagnostics tools), and
2. Updating the scaling routines to accommodate these changes (somewhat more effort, especially with the above mentioned `transform_property_constraints` function). However, given the current push to using the new `Scaler` classes this hopefully becomes much easier (or at least can be done at the same time).

### Additional Context

_No response_

MarcusHolly:
## Fixes/Resolves:
Issue #1547 

## Summary/Motivation:
Resolves potential evaluation errors across the repo by removing division from constraints and ensuring that we're taking the log of positive numbers

- Rewrites various constraints to not use division signs, avoiding potential divisions by 0
- Adds epsilon values to various constraints to ensure we're taking the log of positive numbers

### Legal Acknowledgement


lbianchi-lbl:
This PR resolves a lot of warnings that were being emitted by the IDAES diagnostics toolbox and were being ignored until now.

watertap/property_models/unit_specific/anaerobic_digestion/adm1_reactions.py
```diff
@@ -1459,7 +1461,7 @@ def Dissociation_rule(self, t):
 
         # Equation from [2]
         def CO2_acid_base_equilibrium_rule(self, t):
-            return pyo.log(10**-self.pK_a_co2) == (
+            return pyo.log(10**-self.pK_a_co2 + eps) == (
```
andrewlee94: This constraint looks really odd to me - why not do `-pK_a_CO2 == -6.35*exp(7646/R*(1/Tref-T))` (or something to that effect)? At the moment you are taking the log of an exponent which seems to be redundant.

Also, is the `eps` really needed? If you are trying to avoid evaluation errors then bounding the terms in the logs and divisions would generally be sufficient.

watertap/unit_models/thickener.py
```diff
@@ -221,11 +221,15 @@ def TSS_in(blk, t):
...
-            return blk.p_thick * (10 / (blk.TSS_in[t]))
+            eps = 1e-30 * pyunits.kg / pyunits.m**3
+            return blk.p_thick * (10 / (blk.TSS_in[t] + eps))
 
         @self.Expression(self.flowsheet().time, doc="Remove factor")
         def f_q_du(blk, t):
-            return blk.TSS_rem / (pyunits.kg / pyunits.m**3) / 100 / blk.f_thick[t]
+            eps = 1e-30 * pyunits.m**3 / pyunits.kg
+            return (
+                blk.TSS_rem / (pyunits.kg / pyunits.m**3) / 100 / (blk.f_thick[t] + eps)
```
andrewlee94: Would bounding `f_thick` resolve this issue? As it stands you are adding an effective bound of 1e-30 (which I will note is effectively meaningless unless you have some massive scaling factors applied).

watertap/property_models/unit_specific/anaerobic_digestion/adm1_reactions.py
MarcusHolly: We could simplify it to `log(10) * -pK_a_CO2 == ...` in which case the `eps `probably wouldn't be needed. The LHS wouldn't just be `pK_a_co2` because `log` is actually the natural log, right?

`pK_a_co2` currently is bounded in the domain of `PositiveReals`, but it was still getting a potential evaluation error

watertap/unit_models/thickener.py
MarcusHolly: `f_thick` is an Expression, so I can't apply a bound to it directly. And adding a constraint to bound it seems like it could be more trouble than it's worth

andrewlee94: Well, that touches on the heart of the issue. To avoid evaluation errors you need to reformulate issues at the root cause. So here your options are to reformulate the constraint that is using this Expression such that `f_thick` is moved to the other side of the equation if possible, or you really do need some sort of intermediate variable to keep things bounded (or decide it is an acceptable issue which is sometimes unavoidable).

MarcusHolly: Alright, in this situation I can just rearrange it such that f_thick is moved to the other side, but in cases where it may not be possible to do this, is it really better to deem it an "acceptable issue" rather than just adding an `eps` term? Ig what I'm getting at is when should you add an `eps` term?

watertap/property_models/unit_specific/anaerobic_digestion/adm1_reactions.py
andrewlee94: You are correct that I was missing something: the correct constraint would be something like `-pK_a_CO2 == -6.35 + log10(exp(7646/R*(1/Tref-T)))`. However, I think change of base rules tells us that `log10(exp(x)) = C*x` where `C` is a base conversion factor.

watertap/unit_models/thickener.py
andrewlee94: My suggestion is that you should generally not add the `eps` as it could potentially cause results to change. If your only solution is the add an `eps` then you have basically already concluded the issue is unavoidable and are just making an arbitrary change to satisfy the diagnostics check. In that case, you would just be better leaving a note that these are known, unavoidable issues.

lbianchi-lbl:
@MarcusHolly will resume working on this in the coming days.

lbianchi-lbl:
@MarcusHolly will get back to this after the scaling PRs have been merged.

ksbeattie:
@MarcusHolly is this still needed?

MarcusHolly:
Yea, it's just a low priority issue. It's basically just making less warnings appear when you run the DiagnosticToolbox on various flowsheets. I won't have time to revisit this until after IDAES+

MarcusHolly:
Closing this as making the changes described here is causing more of a headache than it's worth

PR #1555, opened 2025-01-31
Ensure `interval_initializer` always restores bounds

bknueven:
## Fixes/Resolves:
If FBBT raises an exception which is caught, the interval initializer will fail to restore the bounds. This PR ensures the interval initializer will always restore the bounds even if an exception is caught.

## Summary/Motivation:
See above

- Add failing test
- Wrap FBBT and value setting in try / except block to ensure we always execute the bounds restoration.

### Legal Acknowledgement

### DIFF
--- a/watertap/core/util/initialization.py
+++ b/watertap/core/util/initialization.py
@@ -162,31 +162,38 @@ def interval_initializer(
     for v in blk.component_data_objects(Var, active=True, descend_into=True):
         bound_cache[v] = v.bounds
 
-    fbbt(blk, feasibility_tol=feasibility_tol, deactivate_satisfied_constraints=False)
+    try:
+        fbbt(
+            blk, feasibility_tol=feasibility_tol, deactivate_satisfied_constraints=False
+        )
 
-    for v, bounds in bound_cache.items():
-        if v.value is None:
-            logger.info(
-                f"variable {v.name} has no initial value: setting to {default_initial_value}"
-            )
-            v.set_value(default_initial_value, skip_validation=True)
-        if v.lb is not None:
-            if v.lb == v.ub:
-                logger.debug(f"setting {v.name} to derived value {v.value}")
-                v.set_value(v.lb, skip_validation=True)
-                continue
-            if v.value < v.lb:
-                logger.debug(
-                    f"projecting {v.name} at value {v.value} onto derived lower bound {v.lb}"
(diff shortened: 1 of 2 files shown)


watertap/core/util/initialization.py
```diff
@@ -162,31 +162,38 @@ def interval_initializer(
...
+                    )
+                    v.set_value(v.lb, skip_validation=True)
+            if v.ub is not None:
+                if v.value > v.ub:
+                    logger.debug(
+                        f"projecting {v.name} at value {v.value} onto derived upper bound {v.ub}"
+                    )
+                    v.set_value(v.ub, skip_validation=True)
+
+    except:
```
kurbansitterley, line 193: Is it worth logging something in the `except` block so the user knows something went wrong?

watertap/core/util/initialization.py
```diff
@@ -162,31 +162,38 @@ def interval_initializer(
...
-            if v.value > v.ub:
-                logger.debug(
-                    f"projecting {v.name} at value {v.value} onto derived upper bound {v.ub}"
-                )
-                v.set_value(v.ub, skip_validation=True)
-
-    for v, bounds in bound_cache.items():
-        # restore bounds to original
-        v.bounds = bounds
+                v.set_value(default_initial_value, skip_validation=True)
```
adam-a-a, line 175: What would `skip_validation=False` actually do? Not sure on what sort of validation would otherwise be conducted.

adam-a-a (approved):
LGTM

kurbansitterley (approved):
LGTM

watertap/core/util/initialization.py
bknueven: That is what the `raise` below will do -- the `Exception` will be raised as-is.

watertap/core/util/initialization.py
bknueven: `skip_validation=True` ensures the value is "valid", e.g., in the bounds. But this value was already provided by the user, so there's no need to validate it. Even if there was an issue, this isn't the place to raise a fuss about it.

watertap/core/util/initialization.py
bknueven: But, critically, the bounds will be restored before the `Exception` is raised. Up to the caller to do something with it.

Issue #1505, opened 2024-10-07
Revise UF_ZO costing defaults


adam-a-a:
Costs assumed for ultrafiltration account for "total project costs" used in the source rather than direct equipment costs. We should revise to reflect the [reduced] cost relationship attributed to direct equipment costs.

Thus, we need to edit the `ultra_filtration.yaml` to reflect the desired costing. Tagging @kurbansitterley @zacharybinger @MuktaHardikar   who were involved in this discussion.

kurbansitterley:
This specific issue came from costing ```Softening > UF > RO``` treatment train for the SETO project. UF CAPEX came out to ~$10.5M while the RO+Pump came to ~$2.5M. Naturally, this raised some eyebrows.

This is an exemplar of the mismatch we have in CAPEX/OPEX "scope" (I don't know the proper term) for some unit models.

In the case of UF, the [reference](https://www.twdb.texas.gov/publications/reports/contracted_reports/doc/1348321632_manual.pdf) (Texas Water Board, 2016; Table 3.20 & 3.21) has two costing relationships for UF: 
- Table 3.21 has "Planning Level Membrane Equipment Costs" ```CAPEX = $0.5M * (flow_mgd)```. I would argue this is more aligned with how we cost RO (i.e., only considering membrane-related costs and assuming a cost factor to account for the rest of the capital costs)
- Table 3.20 applies a factor of 5 to Table 3.21 to account for inter-stage pumping, tanks, plumbing, and the building for the UF process. I would argue this is a deviation from the "scope" of the RO costing approach. This is the current approach used to cost UF in WaterTAP.

The fix here is to use the value from Table 3.21 for UF ZO as the default parameter and apply a TIC factor to be aligned with the RO costing approach. 

However, there are likely other instances of such misaligned scope for our costing approaches that need to be considered. I imagine this is not something that will be addressed in the near future, but should be a primary issue moving forward when thinking about modifications to WaterTAP's costing framework/scope/etc. 

Tagging @TimBartholomew for visibility.

PR #1557, opened 2025-02-13
Revise UF_ZO costing defaults

zacharybinger:
## Fixes/Resolves:
#1505 

## Summary/Motivation:
In the case of UF, the [reference](https://www.twdb.texas.gov/publications/reports/contracted_reports/doc/1348321632_manual.pdf) (Texas Water Board, 2016; Table 3.20 & 3.21) has two costing relationships for UF:

    Table 3.21 has "Planning Level Membrane Equipment Costs" CAPEX = $0.5M * (flow_mgd). I would argue this is more aligned with how we cost RO (i.e., only considering membrane-related costs and assuming a cost factor to account for the rest of the capital costs)
    Table 3.20 applies a factor of 5 to Table 3.21 to account for inter-stage pumping, tanks, plumbing, and the building for the UF process. I would argue this is a deviation from the "scope" of the RO costing approach. This is the current approach used to cost UF in WaterTAP.

The fix here is to use the value from Table 3.21 for UF ZO as the default parameter and apply a TIC factor to be aligned with the RO costing approach.

- Change capital_a_factor

### Legal Acknowledgement


kurbansitterley (approved):
LGTM, we have been using this change for a while on REFLO.

adam-a-a (approved):

lbianchi-lbl:
The current failure is due to the same issue with `actions/setup-miniconda` described IDAES/idaes-pse#1569 and are unrelated to the changes in this PR. The less-than-ideal workaround is rerunning the failed job(s) until they pass.

PR #1559, opened 2025-03-05
Update BSM2-P KLa Values

MarcusHolly:
## Summary/Motivation:
The values used for KLa were fixed with incorrect units.

- Updates KLa values based on the attached references
- Updates various flowsheet test files

### Legal Acknowledgement


watertap/flowsheets/full_water_resource_recovery_facility/BSM2_P_extension.py
```diff
@@ -113,9 +119,10 @@ def main(bio_P=False):
     results = solve(m)
 
     # Switch to fixed KLa in R5, R6, and R7 (S_O concentration is controlled in R5)
-    m.fs.R5.KLa.fix(240)
-    m.fs.R6.KLa.fix(240)
-    m.fs.R7.KLa.fix(84)
+    # KLa for R5 and R6 taken from [1], and KLa for R7 taken from [2]
+    m.fs.R5.KLa.fix(240 / 24)
+    m.fs.R6.KLa.fix(240 / 24)
+    m.fs.R7.KLa.fix(84 / 24)
```
adam-a-a, line 125: Can you update these in our other flowsheets?

MarcusHolly: Will do

adam-a-a:
To add clarification to the PR description, unintended values were fixed for KLa, oxygen mass transfer cofficient. That is, the KLa variable was assigned units of 1/hour (instead of 1/day as observed in literature). However, while the literature included a value of 240/day, for example, we were using 240/hr. 

Thus, @MarcusHolly is updating those values to match what they should be in terms of 1/hr (but not changing the units of the KLa variable itself to 1/day).

adam-a-a (commented):
@MarcusHolly - this looks good, but there are three things I'd like to have included in this PR:
1) add corresponding changes to other BSM (or activated sludge process) related flowsheets that might also assume unintended values for KLa.
2) Add a test, primarily for BSM2-P and conventional BSM, that checks the value of aeration energy. Given that we were using KLa values 24x higher than intended, I want to see how this affects our aeration energy estimates.
3) double-check documentation specifically for flowsheet specifications that may need updating.

watertap/flowsheets/full_water_resource_recovery_facility/BSM2_P_extension.py
```diff
@@ -374,9 +381,13 @@ def build(bio_P=False):
...
-    for R in m.fs.aerobic_reactors:
-        iscale.set_scaling_factor(R.KLa, 1e-2)
-        iscale.set_scaling_factor(R.hydraulic_retention_time[0], 1e-3)
+    if bio_P:
+        for R in m.fs.aerobic_reactors:
+            iscale.set_scaling_factor(R.KLa, 1e-1)
+            iscale.set_scaling_factor(R.hydraulic_retention_time[0], 1e-2)
+    else:
+        for R in m.fs.aerobic_reactors:
+            iscale.set_scaling_factor(R.KLa, 1e-1)
```
luohezhiming, line 390: Might need to revisit if we could stabalize the system with only set scaling factors for KLa

luohezhiming (approved):
Looks good!

watertap/flowsheets/full_water_resource_recovery_facility/BSM2_P_extension.py
MarcusHolly: Possibly - for now the flowsheet won't solve if we don't scale HRT for bio_P=True

adam-a-a (approved):

PR #1563, opened 2025-03-25
Reformulate Expr_Ifs in ASM2d-ADM1 Translators

MarcusHolly:
## Fixes/Resolves:

This replaces the remaining Expr_ifs in the ASM2d-ADM1 translator block with the smooth Heaviside function. 

In summary:
```
if a < b:
    return c
else:
    return d
```
goes to

`(c-d)*smooth_heaviside(b-a, k) + d`

## Summary/Motivation:
Expr_ifs introduce discontinuities into the model and the smooth heaviside functions approximate continuous versions of the Expr_ifs.

- Replaces the Expr_ifs in the ASM2d-ADM1 translator
-

### Legal Acknowledgement


watertap/unit_models/translators/translator_asm2d_adm1.py
```diff
@@ -191,6 +190,15 @@ def build(self):
             mutable=True,
             doc="Smoothing factor",
         )
+        self.heaviside_k = Param(
+            initialize=1e1,
+            units=pyunits.dimensionless,
+            mutable=True,
+            doc="Smooth heaviside k parameter",
+        )
```
adam-a-a: Should this be a single value or can there be cases where we might need a separate k value for a separate smooth heaviside relationship?

watertap/unit_models/translators/translator_asm2d_adm1.py
```diff
@@ -191,6 +190,15 @@ def build(self):
...
+
+        def smooth_heaviside(x, k):
+            return 1 / (1 + exp(-2 * k * x))
```
adam-a-a: I am thinking that this belongs somewhere in watertap.core.util. Then we can import where needed. Ultimately, we probably could/should propagate this up to IDAES, but that's not our priority in the interim.

adam-a-a (dismissed):
Looks good, but I recommend moving the smooth_heaviside to utilities and ask whether we might need more than one tunable k parameter; this would make sense if one instance of the smooth_heaviside implementation called for one k value, while another instance in the model might call for a different k value.

watertap/unit_models/translators/translator_asm2d_adm1.py
MarcusHolly: In this translator block, I think we can get away with one value for k. In the ASM1-ADM1, we may need multiple k values.

watertap/unit_models/translators/translator_asm2d_adm1.py
MarcusHolly: Agreed - I'll move it to util

watertap/unit_models/translators/translator_asm2d_adm1.py
```diff
@@ -191,6 +190,15 @@ def build(self):
             mutable=True,
             doc="Smoothing factor",
         )
+        self.heaviside_k = Param(
+            initialize=1e1,
```
Morgan88888888: Shall we provide better initial values for the users, maybe as the default? Or are the current initial values the preferred ones from other references?

Morgan88888888 (changes requested):
The idea of using a smooth function sounds great, but I am wondering how much improvement was shown in solving the flowsheet model. Could you please comment on it?

watertap/unit_models/translators/translator_asm2d_adm1.py
MarcusHolly: Ironically there was no improvement. This is likely because the flowsheet only operates under one of the conditionals in the expr_if, so it's almost like the other conditional doesn't even exist. So when we switched to this smooth function, there's no noticeable improvement. That's my rudimentary understanding, at least.

MarcusHolly: This k value was not derived from any literature/references - just trial and error.

adam-a-a: What if we perturb the model? That is, you can write a test either for the translator itself or on BSM2-P where you vary inlet concentrations, for example (ideally ones that correspond to the components dealt with in expression ifs/smooth heaviside approximations) and see what the differences are

It adds on work, and we could otherwise just cross our fingers and hope that this works better as we move forward---but then we may have harder time later if things continue to go wrong. Would be good to determine what impact this limited modification has thus far before proceed--by doing a parameter sweep (or testing out that IPOPTConvergenceAnalysis capability?)

adam-a-a: Note-- my comment above was in response to @Morgan88888888 's comment-- " I am wondering how much improvement was shown in solving the flowsheet model. Could you please comment on it?"

Morgan88888888: @adam-a-a  Thanks, and I like your idea, and testing extreme cases is a good way to estimate potential improvement.

MarcusHolly: I've added some IPOPTConvergenceAnalysis tests. The results are also identical to the version of the translator in main

MarcusHolly: @adam-a-a @Morgan88888888 I've figured out why this PR doesn't seem to have any effect on the model diagnostics. It's because, in the original expr_if, a is ALWAYS larger than b, regardless of the feed composition. So it's as if the first conditional doesn't even exist. For b to be larger than A, we would need to change the parameter values themselves (specifically `Ni["X_pr"]` or `i_NXS`)

TLDR: These expr_ifs aren't causing any problems. If there were problematic expr_ifs, they've already been addressed in the previous PR where I replaced them with smooth_min or smooth_max.

watertap/unit_models/translators/tests/test_translator_asm2d_adm1.py
```diff
@@ -634,3 +642,77 @@ def test_conservation(self, asmadm):
...
+
+        assert ca.results[1]["success"]
+        assert ca.results[1]["results"]["iters"] == 3
+        assert ca.results[1]["results"]["iters_in_restoration"] == 0
+        assert ca.results[1]["results"]["iters_w_regularization"] == 0
+        assert ca.results[1]["results"]["numerical_issues"]
+
+        assert not ca.results[2]["success"]
+
+        assert not ca.results[3]["success"]
```
agarciadiego: I like the inclusion of this test. However, we should be careful about how adding more of this test would affect testing times. Additionally, I believe that for this test (not in this case, as Marcus demonstrated in a previous comment) to effectively show whether convergence is improved, the tests should be conducted over the flowsheet rather than on the specific translators.

watertap/core/util/misc.py
```diff
@@ -29,3 +30,16 @@ def is_constant_up_to_units(expr):
             continue
         return False
     return True
+
+
+def smooth_heaviside(x, k):
```
avdudchenko, line 36: How is this different then smooth_min, smooth_max in ideas? Looks pretty similar in operation, just configured to default out to 0 or the value you want?

MarcusHolly: `Smooth_min `and `smooth_max `have also been implemented in the BSM2 translator blocks, but not all of the Expr_Ifs can be rewritten as `smooth_min `or `smooth_max`, so `smooth_heaviside `is being applied in these cases

avdudchenko: Makes sense, it might make sense to put a PR up on IDAES to add this function to their math file as I imagine this has broad utility and it would make sense to merge it out of WaterTAP to IDAES.

MarcusHolly: Put up a PR in IDAES here: https://github.com/IDAES/idaes-pse/pull/1603

Morgan88888888 (approved):
As we discussed yesterday, it seems the BSM2 model will always pick one option, and this happens for all BSM2 implementations, not just ours. I don’t have any further feedback—hopefully, we can find some benefit from it somewhere in the future.

watertap/unit_models/translators/tests/test_translator_asm2d_adm1.py
adam-a-a: > the tests should be conducted over the flowsheet

This is a great point, and I think we should take one more step here to evaluate effects on at least a subset of the plant-wide flowsheet (or the whole). What you did here @MarcusHolly was exactly what I was hoping for. We just now want to make sure we get a sense of how this actually impacts flowsheet optimization. You may have mentioned on a call that flowsheet stability hasn't been affected, but I am unsure.

MarcusHolly: There didn't appear to be any change in the flowsheet's model diagnostics, but I didn't do a parameter sweep, so I'll work on that soon (tomorrow?)

MarcusHolly:
@andrewlee94 I've been trying to apply some of the IPOPTConvergenceAnalysis tools to this PR to see how replacing the Expr_ifs with the heavidside function affects the model diagnostics of the BSM2 flowsheet. However, I'm running into an issue where tests are passing locally, but failing across the board here. Could the issue be that IPOPTConvergenceAnalysis is using the ipopt solver rather than ipopt-watertap? If so, can I just add a config option where the user can specify the solver? See this line: https://github.com/IDAES/idaes-pse/blob/6e72af390ace4a2883036db394b5ea20cf42ff65/idaes/core/util/model_diagnostics.py#L3578

andrewlee94:
@MarcusHolly That is quite possible. I cannot recall what the WaterTAP solver config does but it was important for getting things to solve. You should be able to modify the convergence analysis tool to use the WaterTAP solver however - the only reason IPOPT was hard coded was because it only knows how to parse IPOPT outputs, but you should be able to use different configurations.

MarcusHolly:
> @MarcusHolly That is quite possible. I cannot recall what the WaterTAP solver config does but it was important for getting things to solve. You should be able to modify the convergence analysis tool to use the WaterTAP solver however - the only reason IPOPT was hard coded was because it only knows how to parse IPOPT outputs, but you should be able to use different configurations.

Will I need to open a PR in IDAES to do so? I'm not sure how I would get it to use the ipopt-watertap solver w/o directly modifying the class in IDAES

andrewlee94:
You won't want to make a PR, as IDAES cannot know about the WaterTAP solver. Start by making changes locally.

lbianchi-lbl:
@MarcusHolly to test the changes from IDAES/idaes-pse#1606 in this PR, you can change the `idaes-pse` requirement in `setup.py`:

https://github.com/watertap-org/watertap/blob/dbc7e78d9d4a31c0300110dc20a6ae63184a4436/setup.py#L72

to `idaes-pse @ git+https://github.com/IDAES/idaes-pse@6291747dabaa1a791e55b19d34957203bb5e2ee4`

MarcusHolly:
@lbianchi-lbl Even after implementing the changes from the above IDAES PR (which allow us to use the ipopt-watertap solver rather than the ipopt solver), there are still discrepancies between my local testing and the tests here on GitHub. That is, all the tests are passing for me locally, but some of the assert statements are failing across the board (on Windows and Linux) here. Any ideas for why this might be the case besides BSM2 doing BSM2 things? Even when I install a fresh environment, everything still passes for me locally.

andrewlee94:
@MarcusHolly Generally, you cannot expect different platforms to give the same performance (although you might hope that they are similar). For IDAES, we only run these sorts of tests on one OS and Python version.

From a very quick look at the results, you are seeing different numbers of iterations on different platforms which is not unexpected (although the difference is a bit larger than you might hope for (~10%?)).

MarcusHolly:
> @MarcusHolly Generally, you cannot expect different platforms to give the same performance (although you might hope that they are similar). For IDAES, we only run these sorts of tests on one OS and Python version.
> 
> From a very quick look at the results, you are seeing different numbers of iterations on different platforms which is not unexpected (although the difference is a bit larger than you might hope for (~10%?)).

Thanks for the context @andrewlee94! Yes, I'm seeing different number of iterations on different platforms. I assumed adding absolute tolerance would allow for some wiggle room, but as you mentioned, some of the differences are quite significant.

Regardless, it seems like this tool isn't intended to be applied across every OS and Python version, so these tests probably aren't the way to go about this. Adam and I had been exploring how we could compare the BSM2 flowsheet diagnostics before and after this change on the translator blocks. Is there a test or tool that would allow us to efficiently do so... it seems like there may not be a silver bullet for this.

andrewlee94:
Your options are to either generate a baseline for each platform, or to do one platform and trust that it applies equally to all.

You might be better off looking at the DiagnosticsToolbox, and check things like making sure the condition number does not increase.

MarcusHolly:
The condition number is exactly the same before and after this change. There are small increases in the number of iterations (~10%), so maybe this is a good-enough metric to validate these changes.

MarcusHolly:
@lbianchi-lbl Revisiting a discussion from the last dev call, how would I go about adding a test that only applies to 1 OS and python version?

lbianchi-lbl:
@MarcusHolly Adding something like this to the test file should work:

```py
import platform


is_reference_platform = platform.system() == "Windows" and platform.python_version_tuple()[:2] == ("3", "11")
reference_platform_only = pytest.mark.xfail(
    condition=(not is_reference_platform),
    run=True,
    strict=False,
    reason="These tests are expected to pass only on the reference platform (Python 3.11 on Windows)",
)

# ... rest of the file

@reference_platform_only
class TestAllFragile:
    def test_fragile_1(self):
         ...
    def test_fragile_2(self):
        ...

class TestOnlyPartiallyFragile:
    def test_non_fragile_part(self):
        ...
    @reference_platform_only
    def test_fragile_part(self):
        ...

@reference_platform_only
def test_my_fragile_func():
   ...
```

**UPDATE**: using `xfail` instead of `skipif` means the code is still run, minimizing the difference in code coverage between reference and non-reference cases.

MarcusHolly:
@lbianchi-lbl Thanks Ludovico! Can you remind me if there's any risk associated with updating the setup.py file as done in this PR? If we don't want to merge this change in, then I'll have to remove the platform-only tests, which isn't a big deal.

MarcusHolly:
Decided to just remove the IPOPTConvergenceAnalysis tests since I don't think it's worth the headache it was causing

lbianchi-lbl (approved):

PR #1570, opened 2025-04-10
Add In-Line References to Modified ASM2d

### LINKED ISSUE #1166: Ensure references added for conversion factors
Maybe we should mention the source for these parameters, it seems not from the original ASM2D properties model

_Originally posted by @luohezhiming in https://github.com/watertap-org/watertap/pull/1078#discussion_r1266908966_
      
The comment refers to the modified ASM2D property/reaction model.

MarcusHolly:
## Fixes/Resolves:

Issue #1166 

- Adds in-line references regarding where parameters values come from in the literature

### Legal Acknowledgement

### DIFF
--- a/watertap/property_models/unit_specific/activated_sludge/modified_asm2d_properties.py
+++ b/watertap/property_models/unit_specific/activated_sludge/modified_asm2d_properties.py
@@ -176,7 +176,7 @@ class ModifiedASM2dParameterData(PhysicalParameterBlock):
             units=pyo.units.K,
         )
 
-        # COD to VSS coefficients
+        # COD to VSS coefficients [1]
         self.CODtoVSS_XI = pyo.Var(
             initialize=1.5686,
             units=pyo.units.dimensionless,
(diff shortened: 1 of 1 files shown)


agarciadiego (approved):

Morgan88888888 (commented):
Just to confirm, reference 1 is used in two sections, as you indicated, and the rest rely on reference 2 — is that correct?

MarcusHolly:
> Just to confirm, reference 1 is used in two sections, as you indicated, and the rest rely on reference 2 — is that correct?

Technically both reference 1 and reference 2 point to the same resource - this GitHub page: https://github.com/wwtmodels/Plant-Wide-Models, so they could be used interchangeably as far as the in-line citations are concerned

Morgan88888888 (approved):
Thanks! Not sure if we want to throw that GitHub link into the references too, but it looks good to me already.

MarcusHolly:
> Thanks! Not sure if we want to throw that GitHub link into the references too, but it looks good to me already.

Rather than pointing to the GitHub (which has a bunch of extraneous information), it'd be better if we just directly pointed to the c file where we got these parameter values from. And if we do that here, we should probably do that elsewhere in the BSM2 code... which might take some time. I could get to it eventually if we think it's worth the effort. I just figured I'd resolve this one issue since I had a spare few minutes yesterday.

PR #1573, opened 2025-04-23
Add Electrocoagulation 0D Model

kurbansitterley:
## Fixes/Resolves:

Adds EC 0D model, costing, test file, and documentation.

## Summary/Motivation:

This PR adds a 0D electrocoagulation model compatible with MCAS that builds off of the existing EC ZO model. In particular, this model adds the ability to estimate the overpotential required with Nernst and Tafel equations.

The model has been validated against results in both the Gu (2009) paper and the Dubrawski (2014) paper. These are provided as test cases in addition to a costing test case developed as part of the SETO project in collaboration with NMSU.

TODO:

- [x] add documentation

- Add EC 0D model, costing, test file, and documentation.

### Legal Acknowledgement


watertap/unit_models/electrocoagulation.py
```diff
@@ -0,0 +1,976 @@
...
+                    * log10(
+                        pyunits.convert(
+                            b.current_density / b.cathodic_exchange_current_density,
+                            to_units=pyunits.dimensionless,
+                        )
+                    )
+                )
+
+            @self.Constraint(doc="Overpotential calculation")
+            def eq_nernst_overpotential(b):
```
MuktaHardikar: Just double checking how you got this. Could you reference the equation number from the source?

kurbansitterley: added some equation numbers as they appear in the references

docs/technical_reference/unit_models/electrocoagulation.rst
```diff
@@ -0,0 +1,371 @@
+.. _EC_0D:
+
+Electrocoagulation (0D)
```
adam-a-a, line 3: Should filename across the board be `electrocoagulation_0D`, aligning with naming for other units?

watertap/costing/unit_models/electrocoagulation.py
```diff
@@ -0,0 +1,286 @@
...
+
+def build_iron_cost_param_block(blk):
+    costing = blk.parent_block()
+    blk.cost = pyo.Param(
+        initialize=3.41,
+        units=pyo.units.USD_2021 / pyo.units.kg,  # Anuf et al., 2022
+        mutable=True,
+        doc="Cost of iron plate per kg",
+    )
+    costing.register_flow_type("iron", blk.cost)
```
adam-a-a, line 62: I wonder if there should be a custom electrode material in case the user wants to use a different material and not have to settle for the misnomer of aluminum or iron. How about something like `custom_electrode`?

watertap/costing/unit_models/electrocoagulation.py
```diff
@@ -0,0 +1,286 @@
...
+
+
+@register_costing_parameter_block(
+    build_rule=build_aluminum_cost_param_block,
+    parameter_block_name="aluminum",
+)
+@register_costing_parameter_block(
+    build_rule=build_iron_cost_param_block,
+    parameter_block_name="iron",
+)
```
adam-a-a, line 147: Why register both? Should we just register one depending on whether the use specified aluminum or iron?

watertap/unit_models/electrocoagulation.py
```diff
@@ -0,0 +1,991 @@
...
+            self.flowsheet().config.time, doc="Material properties of waste", **tmp_dict
+        )
+
+        prop_in = self.properties_in[0]
+        prop_out = self.properties_out[0]
+        prop_waste = self.properties_waste[0]
+
+        self.add_port(name="inlet", block=self.properties_in)
+        self.add_port(name="outlet", block=self.properties_out)
+        self.add_port(name="waste", block=self.properties_waste)
```
adam-a-a: SHould this be termed "properties_waste"? What if you are recovering a resource as the "waste" stream? Would `properties_byproduct` be more general and not cause major problems for dependent analyses?

watertap/unit_models/electrocoagulation.py
```diff
@@ -0,0 +1,991 @@
...
+        self.cell_volume = Var(
+            initialize=1,
+            bounds=(0, None),
+            units=pyunits.m**3,
+            doc="Volume of electrolytic cell",
+        )
+
+        self.current_density = Var(
+            initialize=1,
+            bounds=(0, 2000),
```
adam-a-a, line 487: Should we remove the upper bound here?

watertap/unit_models/electrocoagulation.py
```diff
@@ -0,0 +1,991 @@
...
+            units=pyunits.volt,
+            doc="Overpotential",
+        )
+
+        @self.Expression(doc="Conductivity")
+        def conductivity(b):
+            return pyunits.convert(
+                prop_in.conc_mass_phase_comp["Liq", "TDS"] / b.tds_to_cond_conversion,
+                to_units=pyunits.S / pyunits.m,
+            )
```
adam-a-a, line 539: Do you think we should eventually add this  (or similar) to MCAS?

watertap/costing/unit_models/electrocoagulation.py
kurbansitterley: We have been using the "register everything" approach for a while (see e.g., the costing method for IX). This was an improvment from the initial approach to material costs, where every unit model had its own cost for specific materials and wasn't (necessarily) consistent at the flowsheet level.

I am happy to change this. Plopping the decorator in the `if/else` block for electrode material below was not cooperative, though.

watertap/unit_models/electrocoagulation.py
```diff
@@ -0,0 +1,991 @@
...
+                prop_out.temperature
+                == b.frac_increase_temperature * prop_in.temperature
+            )
+
+        @self.Constraint(doc="Water recovery")
+        def eq_water_recovery(b):
+            return (
+                prop_out.flow_mass_phase_comp["Liq", "H2O"]
+                == prop_in.flow_mass_phase_comp["Liq", "H2O"]
+                * b.recovery_frac_mass_water
```
adam-a-a: Just checked repo for `recovery_frac_mass_water` and I don't think this naming has been used anywhere. 

We do have the following for water in particular:

- `recovery_frac_mass_H2O` in ZO models (from TAP3)
- `recovery_mass_H2O` in ED_0D and ED_1D
- `recovery_mass_phase_comp` -- but this accounts for other components recoveries too
- `recovery_mass` in MD

watertap/unit_models/electrocoagulation.py
```diff
@@ -0,0 +1,991 @@
...
+                "Liquid Outlet": self.outlet,
+                "Waste Outlet": self.waste,
+            },
+            time_point=time_point,
+        )
+
+    def _get_performance_contents(self, time_point=0):
+
+        # TODO
+        var_dict = {}
```
adam-a-a: I see this is a todo. Would be good to populate some minimal set of bare essentials.

adam-a-a (commented):
Nice job -- most of my comments are trivial but might be important for consistency and enabling users to use more widely (or more easily when interconnecting with other units).

watertap/costing/unit_models/electrocoagulation.py
adam-a-a: Ah, right. I was just thinking more about how it might confusing to see anything related to iron costs if you were only considering aluminum, but I suppose it can't hurt considered we've been using the "register everything" approach.

adam-a-a:
RTD check failing possibly because of missing/misplaced comma:
image

adam-a-a:
@luohezhiming I thought it might be good for you to review this as it might relate to your work on electroN-P

watertap/unit_models/electrocoagulation.py
kurbansitterley: Done! Github copilot to the rescue.

watertap/unit_models/electrocoagulation.py
kurbansitterley: went with `recovery_mass_H2O`

watertap/unit_models/electrocoagulation.py
kurbansitterley: Yep. If I was an experimentalist with a cond > TDS parameter that I knew worked for my system, I would want to have this option.

watertap/unit_models/electrocoagulation.py
kurbansitterley: I am pretty sure this is an astronomically high current density upper bound for EC, no one should ever hit this.

Though it is true that I sort of added upper bounds randomly across this unit model.

watertap/unit_models/electrocoagulation.py
kurbansitterley: Done

watertap/costing/unit_models/electrocoagulation.py
kurbansitterley: It is a good idea and is certainly implementable, but a user could do this with the existing framework by just changing the proper parameter values manually at the flowsheet level.

Actually, that would be essentially how I would want to implement it. I'm not sure how else you could implement a customizable config option without requiring the user to change the materials-specific parameters anyway.

docs/technical_reference/unit_models/electrocoagulation.rst
kurbansitterley: I thought about this and I am happy to make the change. Though after surveying other unit model names, I don't really think this is an established convention in our unit model naming to a significant extent (e.g.,  gac, electrolyzer, clarifier, etc.).

For that matter, maybe I should remove the `(0D)` from these docs title?

kurbansitterley:
@lbianchi-lbl FYI I added myself and @MuktaHardikar as CODEOWNERS for the model in this PR. Should we be doing that or let you primo devs handle that sort of thing?

lbianchi-lbl:
@MuktaHardikar will take a look.

kurbansitterley:
@srikanthallu can you review if you get a chance

MuktaHardikar (approved):
LGTM!

watertap/costing/unit_models/electrocoagulation.py
avdudchenko: Yeah, we don't support simple costing changes today in watertap, the user can inherit the costing and modify the blocks as well.

watertap/unit_models/electrocoagulation.py
```diff
@@ -0,0 +1,1069 @@
...
+        )
+
+        self.frac_increase_temperature = Param(
+            initialize=1.05,
+            mutable=True,
+            units=pyunits.dimensionless,
+            doc="Fractional increase in water temperature from inlet to outlet",
+        )
+
+        if self.config.overpotential_calculation == OverpotentialCalculation.detailed:
```
avdudchenko, line 282: I do not believe you documented this in the documentation.

watertap/unit_models/electrocoagulation.py
avdudchenko: I suggest removing upper bound as it might make the model more stable in general. IPOPT does not enjoy bounds.

watertap/unit_models/electrocoagulation.py
avdudchenko: That might make sense, but it gets complicated as the conversion is not necessary linear in many cases - some time there is an offset or non-linearity to it... so not sure best way to do it - maybe have an option for a constant and a surrogate...but thats a lot of configs to include just for a few models that use this (e.g. ED and EC where we need conductivity for resistance calcs...)

watertap/unit_models/electrocoagulation.py
```diff
@@ -0,0 +1,1069 @@
...
+
+            for k in state_dict.keys():
+                if state_dict[k].is_indexed():
+                    state_args[k] = {}
+                    for m in state_dict[k].keys():
+                        state_args[k][m] = state_dict[k][m].value
+                else:
+                    state_args[k] = state_dict[k].value
+
+        self.state_args_out = state_args_out = deepcopy(state_args)
```
avdudchenko, line 866: why are we creating a local and a class level var? Should just make one or the other so its clear.

watertap/unit_models/tests/test_electrocoagulation.py
```diff
@@ -0,0 +1,651 @@
...
+    m = ConcreteModel()
+    m.fs = FlowsheetBlock(dynamic=False)
+
+    m.fs.properties = MCASParameterBlock(**ec_feed)
+    m.fs.unit = ec = Electrocoagulation(
+        property_package=m.fs.properties,
+        electrode_material="iron",
+        overpotential_calculation="detailed",
+    )
+    set_scaling_factor(ec.properties_in[0].flow_mass_phase_comp["Liq", "H2O"], 1e2)
```
avdudchenko, line 148: These should really be handled through set.default scaling factor for tests. (e.g. m.fs.properties.set_default_scaling(prop,idx,scale) this way calcualte_scaling_factors works correctly

watertap/unit_models/tests/test_electrocoagulation.py
```diff
@@ -0,0 +1,651 @@
...
+    m = ConcreteModel()
+    m.fs = FlowsheetBlock(dynamic=False)
+
+    m.fs.properties = MCASParameterBlock(**ec_feed)
+    m.fs.unit = ec = Electrocoagulation(
+        property_package=m.fs.properties,
+        electrode_material="aluminum",
+        overpotential_calculation="regression",
+    )
+    set_scaling_factor(ec.properties_in[0].flow_mass_phase_comp["Liq", "H2O"], 1e2)
```
avdudchenko, line 225: Same here, can we use the "default_scaling" method on props instead.

watertap/unit_models/tests/test_electrocoagulation.py
```diff
@@ -0,0 +1,651 @@
...
+    set_scaling_factor(ec.properties_in[0].flow_mass_phase_comp["Liq", "TDS"], 1e5)
+
+    set_scaling_factor(ec.properties_out[0].flow_mass_phase_comp["Liq", "H2O"], 1e2)
+    set_scaling_factor(ec.properties_out[0].flow_mass_phase_comp["Liq", "TDS"], 1e5)
+
+    set_scaling_factor(
+        ec.properties_byproduct[0].flow_mass_phase_comp["Liq", "H2O"], 1e5
```
avdudchenko, line 232: we should not really ever need to do this as it can lead to solver errors (e.g. you are solving mass flow to tolerance of 1e2 on feed side, but trying to solve it to 1e5 on the by_products side... that might be an issue)

watertap/unit_models/tests/test_electrocoagulation.py
```diff
@@ -0,0 +1,651 @@
...
+    m = ConcreteModel()
+    m.fs = FlowsheetBlock(dynamic=False)
+
+    m.fs.properties = MCASParameterBlock(**ec_feed)
+    m.fs.unit = ec = Electrocoagulation(
+        property_package=m.fs.properties,
+        electrode_material="iron",
+        overpotential_calculation="regression",
+    )
+    set_scaling_factor(ec.properties_in[0].flow_mass_phase_comp["Liq", "H2O"], 1e2)
```
avdudchenko, line 300: Same comments about scaling as above.

docs/technical_reference/costing/electrocoagulation.rst
avdudchenko, line 1: I think this is missing documentation on overpopulation and other functions T.T

avdudchenko (changes requested):
Overall looks great. But we should expand docs on the over potential calcs on so on. We should also use standard scaling methods in tests to very they work as expected (e.g. prop.set_default_scale)

avdudchenko (approved):
LGMT - looks good.

PR #1577, opened 2025-04-30
WaterTAP Project Template CLI

zacharybinger:
## Fixes/Resolves:

Adds a CLI utility for initializing new WaterTAP project templates using Cookiecutter. If no issue exists, this PR introduces a new developer feature for rapidly scaffolding standardized project directories.

## Summary/Motivation:
To improve the developer experience and promote consistency across new WaterTAP projects, this PR introduces a watertap project command-line interface. It allows users to generate project folders from a GitHub-hosted Cookiecutter template with automatic fallback to a local template when offline. 

A predesigned project template can be automatically created by running the following in the terminal:
`watertap project`

Additions to the project structure can be added here:
https://github.com/watertap-org/watertap_project_template

- Adds watertap_cli.py with a project subcommand
- Integrates GitHub template fetching with fallback to a local template directory
- Automatically deletes the cached Cookiecutter template to avoid prompts
- Supports both interactive and --no-input project generation
- Allows --extra-context injection for scripting or automation

### Legal Acknowledgement


adam-a-a:
@zacharybinger this looks like an interesting PR. I can imagine what the intention is here, but I am unclear on the motivation behind the PR, how it arose, and what the vision for its usage would be. Would be good to discuss on the dev call

adam-a-a:
@zacharybinger I would also add that if we are to bring this in, we should add documentation on this along with perhaps a brief how-to to clarify the value of using it

adam-a-a:
one thought that comes to mind is that perhaps a feature like this should be reflected in the UI (though separate from this PR). 

That is, I should be able to click on "Create new project" in the UI, which would then create some of the folder structure shown here. Then, while setting up/generating results in the UI, the user could save data/outputs on the go, which would point to relevant folders for saving

adam-a-a:
@dangunter @MichaelPesce do you think we can make the case for including this in the UI work for PrOMMiS (and other projects)? I think this is something that any project in the IDAES+ world could use, and I am curious to hear your thoughts on how we could implement this so that any IDAES+ project could use this--via CLI as well as the UI.

ksbeattie:
Closing this in preference to issue #1617

PR #1578, opened 2025-05-02
ZO Flowsheet Pointer

### LINKED ISSUE #1445: Archive Zero Order Flowsheets
### Description

All of the zero-order flowsheets were deleted in #1438, so a separate repository should be made that archives these now-deleted files.

### Motivation

We still want these flowsheets to be accessible if necessary, but it's not worth the effort to constantly maintain them.

### To do

- [ ] ~Create repository e.g. `watertap-org/zero-order-flowsheets` (@lbianchi-lbl)~
- [ ] Figure out the most appropriate structure and a concrete list of what flowsheets should be there (@lbianchi-lbl, @MarcusHolly)

MarcusHolly:
## Fixes/Resolves:
Issue #1438 

- Adds a pointer inside the flowsheet index to point users to the archived zero-order flowsheets
-

### Legal Acknowledgement

### DIFF
--- a/docs/technical_reference/flowsheets/index.rst
+++ b/docs/technical_reference/flowsheets/index.rst
@@ -25,4 +25,4 @@ The flowsheets below can be accessed at the
    seawater_RO_desalination
 
 Several zero-order flowsheets have been archived but can still be accessed via the WaterTAP
-`1.0 Release <https://github.com/watertap-org/watertap/releases/tag/1.0.0>`_.
+`0.12 Release <https://github.com/watertap-org/watertap/tree/0.12.0/watertap/examples/flowsheets/case_studies/wastewater_resource_recovery>`_.
--- a/docs/technical_reference/flowsheets/index.rst
+++ b/docs/technical_reference/flowsheets/index.rst
@@ -23,3 +23,6 @@ The flowsheets below can be accessed at the
    mvc
    RO_with_energy_recovery
    seawater_RO_desalination
+
+    Several zero-order flowsheets have been archived but can still be accessed via the WaterTAP
+    `1.0 Release <https://github.com/watertap-org/watertap/releases/tag/1.0.0>`_
--- a/docs/technical_reference/flowsheets/index.rst
+++ b/docs/technical_reference/flowsheets/index.rst
@@ -24,5 +24,5 @@ The flowsheets below can be accessed at the
    RO_with_energy_recovery
    seawater_RO_desalination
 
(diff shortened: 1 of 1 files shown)


luohezhiming (approved):

agarciadiego (approved):
LGTM

adam-a-a (changes requested):
@MarcusHolly I wonder if this should/could be more direct. When I go to the release tag, since I know what to look for, I am able to navigate to the relevant folder, but I cannot find the flowsheets we're intending to point to. I think they may have been removed for the 1.0 but I could be wrong. Should we instead point to the release before 1.0?

I think it would be better to point something like the actual folder's permalink, e.g.: https://github.com/watertap-org/watertap/tree/a5103f312bea4ea535a9abc401c4fe893e733766/watertap/examples/flowsheets/case_studies/wastewater_resource_recovery

MarcusHolly:
Thanks for double checking this Adam. The files were deleted in 1.0, so I should've pointed to the prior release (0.12). I've changed the link to point to the 0.12 release since I think the URL looks cleaner, and it's easier for users to understand exactly where these files exist compared to the permalink

adam-a-a:
> Thanks for double checking this Adam. The files were deleted in 1.0, so I should've pointed to the prior release (0.12). I've changed the link to point to the 0.12 release since I think the URL looks cleaner, and it's easier for users to understand exactly where these files exist compared to the permalink

The URL definitely is cleaner. I wonder though if we'd want to display that clean URL and use its associated permalink behind the scenes, just because it'd be more resilient to any potential changes in the future (like a maintainer moving/deleting tags). Probably an unlikely scenario, but just throwing it out there.

MarcusHolly:
> > Thanks for double checking this Adam. The files were deleted in 1.0, so I should've pointed to the prior release (0.12). I've changed the link to point to the 0.12 release since I think the URL looks cleaner, and it's easier for users to understand exactly where these files exist compared to the permalink
> 
> The URL definitely is cleaner. I wonder though if we'd want to display that clean URL and use its associated permalink behind the scenes, just because it'd be more resilient to any potential changes in the future (like a maintainer moving/deleting tags). Probably an unlikely scenario, but just throwing it out there.

We could do that... although it might confuse users to have the 0.12 link displayed but be directed to the permalink. We can just bring this up in the dev call, but I would also assume that old tags are unlikely to be changed/deleted.

adam-a-a (approved):
LGTM - we can still ask during dev call even though we think it is low risk.

Issue #1579, opened 2025-05-07
Broken URL for Parmeter Estimation Tutorial


adam-a-a:
### Description

I was reviewing our tutorial landing page and discovered that our "Conducting a Parameter Estimation" tutorial has a broken link"

![Image](https://github.com/user-attachments/assets/fd0270fa-9865-4510-a561-7f82003262fa)

### Expected Behavior

Expect to be taken to the tutorial

### Steps to Reproduce

1. Go to the [tutorial landing page](https://nbviewer.org/github/watertap-org/watertap/blob/main/tutorials/00-index.ipynb)
2. Click on "Conducting a Parameter Estimation" tutorial under the "Advanced" section.

### Environment

- OS:
- Python:
- WaterTAP:

PR #1581, opened 2025-05-07
flush nbviewer cache

adam-a-a:
## Fixes/Resolves:

#1579 
-
-

### Legal Acknowledgement

### DIFF
--- a/docs/tutorials/index.rst
+++ b/docs/tutorials/index.rst
@@ -7,7 +7,7 @@ Browsing the tutorials
 The tutorials included with WaterTAP are available in the "tutorials" directory
 of the WaterTAP GitHub repository: `<https://github.com/watertap-org/watertap/tree/main/tutorials/>`_.
 
-To view a static version of the notebooks online, you can use the following link: `<https://nbviewer.org/github/watertap-org/watertap/blob/main/tutorials/00-index.ipynb>?flush_cache=true`_.
+To view a static version of the notebooks online, you can use the following link: `<https://nbviewer.org/github/watertap-org/watertap/blob/main/tutorials/00-index.ipynb?flush_cache=true>`_.
 
 Running the notebooks
 ---------------------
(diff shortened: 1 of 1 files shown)


MarcusHolly (approved):
Approving since it seems straightforward, but what does flushing the cache do?

zacharybinger (approved):

docs/tutorials/index.rst
```diff
@@ -7,7 +7,7 @@ Browsing the tutorials
 The tutorials included with WaterTAP are available in the "tutorials" directory
 of the WaterTAP GitHub repository: `<https://github.com/watertap-org/watertap/tree/main/tutorials/>`_.
 
-To view a static version of the notebooks online, you can use the following link: `<https://nbviewer.org/github/watertap-org/watertap/blob/main/tutorials/00-index.ipynb>`_.
+To view a static version of the notebooks online, you can use the following link: `<https://nbviewer.org/github/watertap-org/watertap/blob/main/tutorials/00-index.ipynb>?flush_cache=true`_.
```
lbianchi-lbl:
```suggestion
To view a static version of the notebooks online, you can use the following link: `<https://nbviewer.org/github/watertap-org/watertap/blob/main/tutorials/00-index.ipynb?flush_cache=true>`_.
```

I'm not familiar with this option, but if `flush_cache=true` is supposed to be a URL parameter, then I believe it should be part of the URL in the ReST hyperlink.

lbianchi-lbl (changes requested):

docs/tutorials/index.rst
adam-a-a: Thanks, @lbianchi-lbl . I realized this after the fact and committed your suggestion.

adam-a-a: Notably, the issue has been resolved (seems you just have to wait a few minutes for the changes to kick in for nbviewer), so this PR might not be necessary.

adam-a-a:
> Approving since it seems straightforward, but what does flushing the cache do?

It's supposed to flush the cache of the Jupyter notebook viewer, when there is some delay and the viewer is still displaying old content or pointing to an old url, from what I understand. However, the issue resolved itself, and currently, this change is somehow breaking the RTD check, so I think I'll just close this.

Issue #1544, opened 2024-12-20
Implementing IDAES Scaler Objects into BSM2


MarcusHolly:
### Description

BSM2 Flowsheets

- [x] BSM2 Flowsheet #1585 
- [x] BSM2-P Flowsheet #1596 

BSM2 Unit Models
- [x] Anaerobic Digester #1535
- [x] CSTR #1542 
- [x] CSTR w/Injection #1542 
- [x] Clarifier #1543 
- [x] Thickener #1545 
- [x] Dewaterer #1545 
- [x] Translator Blocks #1546 

BSM2 Property Models
- [x] ADM1 #1530 
- [x] Modified ADM1 #1537 
- [x] ASM1 #1519 
- [x] Modified ASM2d #1533 

### Motivation

These scaler objects should replace the existing scaling routine that makes use of iscale. Eventually, all WaterTAP unit and property models should have these Scalers. However, there are no plans to deprecate iscale at the moment.

### Possible Implementation

See example PRs above.

### Additional Context

IDAES PR that introduced the new scaler objects: https://github.com/IDAES/idaes-pse/pull/1429

MarcusHolly:
The next step for this scaling effort should be to focus on one of the the above "Unit Model PRs" at a time, rather than trying to refine all of them at once. This will make it easier for reviewers and ensure we don't have to constantly go back and forth between these PRs while we make improvements. As for the improvements that need to be made in each PR, I think we should consider the following:

1.) Assign default scaling factors for more variables and make sure manual scaling factors in the tests are set before calling the automated routines (may already be the case). See the thread here for additional context: https://github.com/watertap-org/watertap/pull/1536#discussion_r1934259780

2.) Ensure every constraint is scaled in the models. See the thread here for additional context: https://github.com/watertap-org/watertap/pull/1536#discussion_r1934280919

3.) Use the IPOPTConvergenceAnalysis tool to help compare whether the new scaling tools are performing better than the old method. See an example of an IDAES test using this tool here: https://github.com/IDAES/idaes-pse/blob/6cea367e53f7cae5615eb9f627708f35379ce1e3/idaes/core/util/tests/test_model_diagnostics.py#L2861

MarcusHolly:
@adam-a-a @luohezhiming I think Chenyu and I can work together on this. Is there anything I missed in the summary above or does anything not make sense?

PR #1550, opened 2025-01-14
Applying Scalers to BSM2 Flowsheets

MarcusHolly:
## Summary/Motivation:
Applies the scaler objects for various unit model and property models into the BSM2 flowsheet. Many of the "files changed" will disappear as other PRs are merged in.

## Changes proposed in this PR:
- Adds config option to toggle between the new and old scaling methods
-

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #1585, opened 2025-05-12
Apply Scalers to Conventional BSM2

MarcusHolly:
## Fixes/Resolves:

#1544 - This PR merges all the changes from the individual scaler PRs in the aforementioned issue

## Summary/Motivation:
This applies the IDAES Scaler Objects developed for various property and unit models into the conventional BSM2 flowsheet

- Minor changes to some of the Scaler objects
- Implements new IDAES scaling routine into the BSM2 flowsheet

### Legal Acknowledgement


MarcusHolly:
Replaces #1550

watertap/flowsheets/full_water_resource_recovery_facility/BSM2.py
```diff
@@ -55,52 +72,136 @@
...
-    pyo.assert_optimal_termination(results)
-    print("\n\n=============OPTIMIZATION RESULTS=============\n\n")
-    # display_results(m)
-    display_costing(m)
-    display_performance_metrics(m)
+
+    rescale_system(m)
+
+    rescaling = pyo.TransformationFactory("core.scale_model")
+    rescaled_model = rescaling.create_using(m, rename=False)
```
MarcusHolly: @dallan-keylogic Before optimization I was able to reduce the Jacobian Condition number of the solve from 5.502e19 to  4.366e9. However, I'm not exactly sure how scaling should be handled when we after we `setup_optimization`. I imagine we will need to rescale the system since the optimization fails to solve without additional scaling. Am I going about this the right way (creating a `rescale_system` that calls the initial scaling routine and that aims to build on top of it) or is there a better way to do this? For instance, is it possible to update/recommended to just update the initial scaling routine with rather than doing a second scaling transformation?

watertap/flowsheets/full_water_resource_recovery_facility/BSM2.py
```diff
@@ -589,11 +1440,61 @@ def Vol_3(self, t):
         return m.fs.R5.volume[0] >= m.fs.R4.volume[0] * 0.5
 
 
-def solve(blk, solver=None, tee=False):
+def rescale_system(m, reactor_volume_equalities=True):
+    m.scaling_factor = pyo.Suffix(direction=pyo.Suffix.EXPORT)
```
adam-a-a: Was this Suffix already constructed in scale_system under `m.scaling_factor`? Should we add a check first to check for existence of m.scaling_factor first? I don't know what we'd gain from this, except that it's generally discouraged to overwrite attributes.

watertap/flowsheets/full_water_resource_recovery_facility/BSM2.py
```diff
@@ -589,11 +1440,61 @@ def Vol_3(self, t):
...
+    sb = ScalerBase()
+    csb = CustomScalerBase()
+
+    sb.set_variable_scaling_factor(
+        m.fs.DU.overflow_state[0].conc_mass_comp["X_BH"], 1e5, overwrite=True
+    )
+    sb.set_variable_scaling_factor(
+        m.fs.DU.overflow_state[0].conc_mass_comp["X_BA"], 1e5, overwrite=True
+    )
```
adam-a-a: seems fairly specific--what made you apply these scaling factors for these particular variables on DU?

watertap/flowsheets/full_water_resource_recovery_facility/BSM2.py
```diff
@@ -385,20 +401,838 @@ def set_operating_conditions(m, reactor_volume_equalities=False):
...
+    sb.set_variable_scaling_factor(
+        m.fs.RADM.liquid_phase.reactions[0].reaction_rate["R1"],
+        1e5,
+        overwrite=True,
+    )
+    sb.set_variable_scaling_factor(
+        m.fs.RADM.liquid_phase.reactions[0].reaction_rate["R11"],
+        1e5,
+        overwrite=True,
+    )
```
adam-a-a: I'm unclear on why we would us `sb`, a separate ScalerBase instance, to apply manual scaling factors to a unit as opposed to using the instance of ADScaler to set all of the variable scale factors manually.

watertap/flowsheets/full_water_resource_recovery_facility/BSM2.py
```diff
@@ -385,20 +401,838 @@ def set_operating_conditions(m, reactor_volume_equalities=False):
...
+    cstr_scaler = CSTRScaler()
+    for unit in cstr_list:
+        cstr_scaler.scale_model(
+            unit,
+            submodel_scalers={
+                unit.control_volume.properties_in: ASM1PropertiesScaler,
+                unit.control_volume.properties_out: ASM1PropertiesScaler,
+                unit.control_volume.reactions: ASM1ReactionScaler,
+            },
+        )
```
adam-a-a: One thought I am getting from seeing the new scaler api in action a bit more: it could be easier and cleaner to implement something like the highlighted code under the CSTR's calculate_scaling_factors() method, where an instance is created, and the scale_model method is called on the instance. The calculate_scaling_factors() method could have an arg that'd allow you to choose either the old or new approach to scaling the unit. 

To clarify my motivation--I am not immediately sold on the idea of having to build a complex flowsheet, then have to create an instance of a unique scaler class for each unit, then call scale_model and have to specify submodel_scalers.  Just feel like more of this could be tucked under the hood rather than having to do this at the flowsheet level.

I am not making this comment as a suggested change for this PR. It is more a question to you and maybe @andrewlee94  and @dallan-keylogic , as well as something for us to think about before we roll this out across whole repos (here, IDAES, PrOMMiS)

watertap/flowsheets/full_water_resource_recovery_facility/BSM2.py
```diff
@@ -526,18 +1375,20 @@ def setup_optimization(m, reactor_volume_equalities=False):
...
 
     m.fs.R5.outlet.conc_mass_comp[:, "S_O"].unfix()
-    m.fs.R5.outlet.conc_mass_comp[:, "S_O"].setub(8e-3)
+    m.fs.R5.outlet.conc_mass_comp[:, "S_O"].setub(8.2e-3)
 
     # Unfix fraction of outflow from reactor 5 that goes to recycle
     m.fs.SP5.split_fraction[:, "underflow"].unfix()
-    # m.fs.SP5.split_fraction[:, "underflow"].setlb(0.45)
+    m.fs.SP5.split_fraction[:, "underflow"].setlb(0.45)
+    m.fs.SP5.split_fraction[:, "overflow"].setlb(0.45)
```
adam-a-a: Are `underflow` and `overflow` the only two outlets here? If so, this seems like we're limiting both split fractions to be between 45-55%.

watertap/flowsheets/full_water_resource_recovery_facility/BSM2.py
andrewlee94: Note that you should only need to provide the submodel scalers if you want to use something other than the default - those options are there to give the user more control **IF** they need it. However, if you just want to use the default scaler for the property package than that should happen automatically.

Also, in general you should not need to create a new custom Scaler when working on a flowsheet - these should be created when the unit model is developed and (if done right) should be applicable to most use cases.

dallan-keylogic: Also, shouldn't this be a `ComponentMap`, not a dictionary?

andrewlee94: In this case I do not think it is strictly necessary, as `Scalers` are not Pyomo components. That said, it might make some sense for consistency and to train users to make use of ComponentMaps.

dallan-keylogic: I thought the problem was using Pyomo components as keys, not values. Actually, I'm surprised we're not getting a warning about them being unhashable.

watertap/flowsheets/full_water_resource_recovery_facility/BSM2.py
andrewlee94: If you do want to fully re-scale the system, you should make use of the utility functions to delete existing suffixes (of just use `component.del_component`).

MarcusHolly: What if we want to keep most of the original scaling, but just tweak a few specific values? I assume in this case we don't need to delete the existing Suffix? If so, then I should delete the reconstruction of the Suffix Adam pointed out here

andrewlee94: Firstly, you do not want to completely replace the existing Suffix then. After that, it really depends on exactly what you are trying to do as there are many ways to go about it. The problem with scaling is that it is all interconnected, so unless you are working with either only one value or all values, you have to be very careful about how you do it.

If you are tweaking individual values, then just change the individual values as needed.

dallan-keylogic: The problem is that he's using a suffix on the root model. He might actually be creating suffixes containing different scaling factors in different places, from which it's unclear which version Pyomo will use. That reminds me, we should strongly discourage users from setting values on a suffix directly. He should instead use `idaes.core.scaling.set_scaling_factor`.

watertap/flowsheets/full_water_resource_recovery_facility/BSM2.py
MarcusHolly: I don't think these are actually necessary. I think these initially came up as having extreme Jacobians, but the underlying issue was probably that the new constraints created by setup_optimization were not being scaled. If I remove these lines, it solves marginally better (20 less iterations and fewer variables close to their bounds)

watertap/flowsheets/full_water_resource_recovery_facility/BSM2.py
andrewlee94: There is a standard for which suffix Pyomo will use (I know because I wrote it), but I am not sure how well documented it is. I _think_ the rules are:

1. First look for a suffix entry for the specific component data object (i.e. indexed value) closest to the model being solve (i.e. the Suffix on the ConcreteModel takes precedence over one in a sub-Block).
2. If not suffix entry is found for the component data object, look for a Suffix value for the parent component starting from the parent Block and moving upwards (i.e. a value in a sub-Block takes precedence over one in a parent).

To give an example:

```
m = ConcreteModel()
m.b = Block()
m.b.v = Var([1,2])

m.scaling_factor = Suffix(
    {m.b.v[1]: 1, m.b.v: 2}
)
m.b.scaling_factor = Suffix(
    {m.b.v[1]: 3, m.b.v: 4}
)
```

When solving this block, the scaling factors used are:

* m.b.v[1] = 1 - for a component data object the scaling factor on the parent takes precedence
* m.b.v[2] = 4 - when a scaling factor is not set for a given index, the scaling factor for the parent component closest to where it is defined is used.

andrewlee94: Also, in regards to @dallan-keylogic's comment - yes, we need to discourage setting scaling factors manually as it otherwise becomes next to impossible to keep track of where scaling factors are set (and allows for cases when multiple scaling factors could be set for the same component). Advanced users can make use of this if they want, but for general use tools we should always set the scaling factor on the parent block of the component for sanity's sake.

watertap/flowsheets/full_water_resource_recovery_facility/BSM2.py
andrewlee94: @MarcusHolly Note that number of iterations is not a great metric to use. Fewer iterations is not better if tolerance/precision is lost in the process. Condition number is generally a better first metric to use.

MarcusHolly: Condition number is unchanged

watertap/flowsheets/full_water_resource_recovery_facility/BSM2.py
MarcusHolly: So just to be clear, is the suggestion that instead of `sb.set_variable_scaling_factor(m.fs.RADM.liquid_phase.heat[0], 1e-3, overwrite=True)`, we should use `set_scaling_factor(m.fs.RADM.liquid_phase.heat[0], 1e-3)`. If so, why was `set_variable_scaling_factor` created in the ScalerBase in the first place?

dallan-keylogic: You should use `set_scaling_factor` over altering the suffix directly. The difference for `sb.set_variable_scaling_factor` is that it limits the scaling factor set between `sb.CONFIG.max_variable_scaling_factor` and `sb.CONFIG.min_variable_scaling_factor`. There are similar max and min values for constraint scaling factors.

I think this is annoying feature that 1) catches unexperienced users unaware and 2) attempts to hide users' bad scaling from them instead of teaching them how to scale better. Pretty much everywhere we go in PrOMMiS these limits need to be overridden because of how extreme the model's values are in SI units.

watertap/flowsheets/full_water_resource_recovery_facility/BSM2.py
MarcusHolly: That's correct. The system can solve without either of these lower bounds, but the condition number will be 3.862E+12 as opposed to 1.892E+11. The post-optimization condition number of the version in main is 6.317E+12.

andrewlee94: That is pointing at an underlying issue you should probably understand. For one, a condition number of 2e11 is still very large and ideally should be reduced (if possible). You should be asking why having these arbitrary bounds is giving an apparent improvement in the condition number of the system - my first guess is that you probably have some variables where you have not set a scaling factor and it is falling back on the variable bounds for its best guess as to the nominal value.

@adam-a-a is correct that those values appear to be arbitrary and are greatly limiting the scope of system operation.

MarcusHolly: Just to offer some reference points, in this PR:

- Pre-optimization condition number reduced from 5.502e19 to 1.228e10
- Post-optimization condition number reduced from 6.317E+12 to 1.896E+11

There is definitely room for improvement with the post-optimization scaling, but I've been working on this PR locally for a while and just wanted to push up what I had atm to get some feedback. So an open question is if this PR is good enough for now (since we also need to get around to making these changes for the BSM2-P flowsheet), or if there are some more refinements that should be made before merging this in. It seems like we should just remove these bounds for now. I will note that I haven't spent much time trying to improve the post-optimization solve

andrewlee94: I suggest removing those bounds, but also look at how the scaling factor is being set for those variables. I suspect you are not setting a value and that the tools are falling back to using the bounds as a best guess, in which case you should be setting a scaling factor in your flowsheet scaling routine.

watertap/flowsheets/full_water_resource_recovery_facility/BSM2.py
MarcusHolly: Short answer is that I'm unaware of a way to apply scaling factors directly to the instance of ADScaler or if that's even a good idea in the first place.

andrewlee94: For application specific scaling factors, you should just set the scaling factors directly without using the base class. You should generally not be writing case-specific scaling classes.

Scalers are meant to help automate the common tasks, not handle the case-specific scaling. For those cases, use things like `set_scaling_factor` from the utilities directly (this should mostly only involve scaling variables, as constraints can generally be automated).

PR #1588, opened 2025-05-14
Dye Desal Base Currency

MarcusHolly:
## Summary/Motivation:
Changes the base currency specified in the associated yaml from USD_2020 to USD_2023. 

- Update the base_currency in the dye desal yaml

### Legal Acknowledgement


adam-a-a:
@MarcusHolly Note that for our report, I had copied the dye_desal flowsheet file and added surrogates for the NF unit there, separately, to avoid mucking up the flowsheet we had, in case things went wrong. When I revisited my alternative flowsheet (months later), I had to merge in all updates to the original flowsheet from main and resolve conflicts. Observing how most changes that I needed were really an update to the currency year, I added those changes to my flowsheet with NF surrogates. 

After doing that, I tried solving the flowsheet, and because there is an assert_units_consistent check somewhere in there, an exception was raised on the total_externalities constraint, showing that 2020 and 2023 currencies were somehow factored into that constraint. I then went to the parameters yaml file, updated all 2020 values to 2023, and the issue was resolved. 

Thus, I would expect there to be _some_ effect at least reflected via total externalties equation, but not sure. Just want to add this detail for your awareness.

watertap/flowsheets/dye_desalination/dye_desalination_global_costing.yaml
```diff
@@ -1,4 +1,4 @@
-base_currency: USD_2020
+base_currency: USD_2023
 base_period: year
 defined_flows:
   electricity:
```
adam-a-a, line 4: Actually, I updated all of these 2020 values to 2023 (just by updating units, not the actual values)

adam-a-a: And that should probably be done for the analysis results too (unfortunately). The other approach would be to just document what these parameter costs are in 2023 dollars to report _those_ values in our report (rather than report the 2020 values).

MarcusHolly: Oh ok, I see your point. These units should've been in 2023 dollars in the first place, so it's not as simple as converting from USD_2020 to USD_2023. In other words, the values are correct, and only the units need to change - which should have some effect on the results

adam-a-a (approved):

kurbansitterley (approved):
LGTM, lets get this in before changing `electricity_cost` units

PR #1603, opened 2025-07-14
Remove kla from bsm2

adam-a-a:
## Summary/Motivation:
KLa, the oxygen mass transfer coefficient, and associated constraints were being constructed on the BSM2-P flowsheet when the AerationTank (CSTR_with_injection) already constructs these vars and constraints at the unit level. I am removing the redundant var and constraints.

- remove KLa and associated param/cosntraints
- revise tests
- also modified Mixer mixingtype config to do away with deactivating unwanted pressure equality constraints

### Legal Acknowledgement

### DIFF
--- a/watertap/flowsheets/full_water_resource_recovery_facility/BSM2.py
+++ b/watertap/flowsheets/full_water_resource_recovery_facility/BSM2.py
@@ -64,6 +64,7 @@ from idaes.models.unit_models import (
     PressureChanger,
     Product,
 )
+import idaes.logger as idaeslog
 
 from watertap.unit_models.aeration_tank import (
     AerationTank,
@@ -81,6 +82,9 @@ from watertap.costing.unit_models.clarifier import (
     cost_circular_clarifier,
     cost_primary_clarifier,
 )
+
+# Set up logger
+_log = idaeslog.getLogger(__name__)
 
 
 def main(reactor_volume_equalities=True):
@@ -113,7 +117,7 @@ def main(reactor_volume_equalities=True):
     display_costing(m)
     display_performance_metrics(m)
 
-    return m, results
+    return m, results, rescaled_model
 
 
 def build():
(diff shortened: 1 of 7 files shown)


adam-a-a:
@MarcusHolly one thing I noticed that isn't directly related to the issue here is that the scaling factor for flow_vol, while seemingly set to 1e2, shows up as being set to 1e1 when I run report_scaling_factors. Do you know why this would be the case?

adam-a-a:
I merged in PR #1609 to try to troubleshoot (and ended up not using those changes). We can merge #1609 first.

luohezhiming (approved):
Looks good, results does not change and much easier to follow!

watertap/flowsheets/full_water_resource_recovery_facility/BSM2.py
```diff
@@ -1040,14 +1044,10 @@ def initialize_system(m):
     initializer = BlockTriangularizationInitializer()
 
     def function(unit):
-        initializer.initialize(unit)
+        initializer.initialize(unit, output_level=_log.debug)
```
MarcusHolly, line 1047: Do we still want this output level?

watertap/flowsheets/full_water_resource_recovery_facility/BSM2_P_extension.py
```diff
@@ -761,9 +723,9 @@ def initialize_system(m, bio_P=False, solver=None):
     def function(unit):
         # TODO: Resolve why bio_P=True does not work with the BTInitializer
         if bio_P:
-            unit.initialize(outlvl=idaeslog.INFO, solver="ipopt-watertap")
+            unit.initialize(outlvl=idaeslog.DEBUG)
         else:
-            initializer.initialize(unit)
+            initializer.initialize(unit, output_level=_log.debug)
```
MarcusHolly, line 728: Same question here about the output levels

MarcusHolly (approved):
LGTM

PR #1605, opened 2025-07-17
ASM2d-N2O Model Development

MarcusHolly:
## Summary/Motivation:
Develops the activated sludge process model no.2d, which accounts for phosphorus (P), sulfur (S), iron (Fe), and greenhouse gas emission (N2O), but this preliminary model only accounts for P and GHG, so it'll be referred to as the ASM2d-N2O model.

- Adds property and reaction package for ASM2d-N2O
- Adds testing for the new property and reaction packages
- Identifies a mistake in the modified ASM2d implementation

### Legal Acknowledgement


watertap/property_models/unit_specific/activated_sludge/asm2d_n2o_properties.py
```diff
@@ -0,0 +1,708 @@
...
+        self.ISS = pyo.Expression(rule=_ISS, doc="Inorganic suspended solids")
+
+        def _TSS(self):
+            tss = self.VSS + self.ISS
+            return tss
+
+        self.TSS = pyo.Expression(rule=_TSS, doc="Total suspended solids")
+
+        # TODO: Double check these effluent metrics
+
```
adam-a-a: Marking this todo- should address or be sure to track in an issue if we decide to handle later.

watertap/property_models/unit_specific/activated_sludge/asm2d_n2o_reactions.py
```diff
@@ -0,0 +1,3088 @@
...
+# National Renewable Energy Laboratory, and National Energy Technology
+# Laboratory (subject to receipt of any required approvals from the U.S. Dept.
+# of Energy). All rights reserved.
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and license
+# information, respectively. These files are also available online at the URL
+# "https://github.com/watertap-org/watertap/"
+#################################################################################
+"""
+ASM2d-PSFe-GHG reaction package.
```
adam-a-a:
```suggestion
ASM2d-N2O reaction package.
```

watertap/property_models/unit_specific/activated_sludge/asm2d_n2o_reactions.py
```diff
@@ -0,0 +1,3088 @@
...
+
```
adam-a-a, line 15: Can add comment about what we are currently incorporating in the model and excluding

watertap/property_models/unit_specific/activated_sludge/asm2d_n2o_properties.py
MarcusHolly: These should be handled later as they were not mentioned in the c files. I can make educated guesses as to what they might be based on my understanding of the changes I made, but we should find a literature reference to confirm this.

MarcusHolly:
This PR doesn't have any of the errors identified in the previous BSM2 implementations, but we should keep those modifications in mind when developing the ADM model that is compatible with ASM2d-N2O. The only thing potentially withholding this PR from being merged would be if we want to add conservation tests.

PR #1609, opened 2025-07-25
fail flag arg for interval_initializer

adam-a-a:
## Summary/Motivation:
The `interval_initializer` utility function was previously implemented to improve model initialization, which it achieves for the most part. However, there are select cases where `InfeasibleConstraintException` will be raised, halting the initialization and subsequent optimization attempts (e.g., for a flowsheet where you might be ok with an imperfect initialization that gets the model to just the right point where it can be solved successfully). 

Here, I simply add a `fail_flag` arg to interval_initializer, set to False as default, so that it will log a warning if an exception is raised during interval_initializer calls. If the user wants to debug a troubled model, or just be more restrictive, the user can set fail_flag=True, which will log the error then raise the exception.

- add fail_flag as arg to interval_initializer
- add tests

### Legal Acknowledgement

### DIFF
--- a/watertap/core/util/initialization.py
+++ b/watertap/core/util/initialization.py
@@ -14,7 +14,7 @@ This module contains utility functions for initialization of WaterTAP models.
 """
 
 
-__author__ = "Adam Atia"
+__author__ = "Adam Atia, Ben Knueven"
 
 from pyomo.environ import check_optimal_termination, ComponentMap, Var
 from pyomo.contrib.fbbt.fbbt import fbbt
@@ -151,7 +151,8 @@ def interval_initializer(
         feasibility_tol : tolerance to use for FBBT (default: 1e-6)
         default_initial_value: set uninitialized variables to this value (default: 0.0)
         logger : logger to use (default: watertap.core.util.initialization)
-        fail_flag: boolean that raises exception if True or logs warning when an exception occurs
+        fail_flag : Boolean argument to specify error or warning (Default: fail_flag=False produces logger warning.
+                        set fail_flag=True to raise an error and when interval initialization encounters exception.)
     Returns:
         None
 
--- a/watertap/core/util/initialization.py
+++ b/watertap/core/util/initialization.py
@@ -141,7 +141,7 @@ def assert_no_degrees_of_freedom(blk):
 
 
 def interval_initializer(
-    blk, feasibility_tol=1e-6, default_initial_value=0.0, logger=_log
+    blk, feasibility_tol=1e-6, default_initial_value=0.0, logger=_log, fail_flag=False
 ):
     """
     Improve the initialization of ``blk`` utilizing interval arithmetic.
@@ -151,7 +151,7 @@ def interval_initializer(
         feasibility_tol : tolerance to use for FBBT (default: 1e-6)
         default_initial_value: set uninitialized variables to this value (default: 0.0)
         logger : logger to use (default: watertap.core.util.initialization)
-
+        fail_flag: boolean that raises exception if True or logs warning when an exception occurs 
     Returns:
         None
 
@@ -190,9 +190,12 @@ def interval_initializer(
                     )
                     v.set_value(v.ub, skip_validation=True)
 
-    except:
-        raise
-
+    except Exception as e:
+        if fail_flag:
+            _log.error(f"Interval initializer failed for {blk} because of the following: {e}")
+            raise e
+        else:
+            _log.warning(f"Interval initializer failed for {blk} because of the following: {e}")
     finally:
         # restore the bounds before leaving this function
         for v, bounds in bound_cache.items():
--- a/watertap/core/util/initialization.py
+++ b/watertap/core/util/initialization.py
@@ -151,7 +151,7 @@ def interval_initializer(
         feasibility_tol : tolerance to use for FBBT (default: 1e-6)
         default_initial_value: set uninitialized variables to this value (default: 0.0)
         logger : logger to use (default: watertap.core.util.initialization)
-        fail_flag: boolean that raises exception if True or logs warning when an exception occurs 
+        fail_flag: boolean that raises exception if True or logs warning when an exception occurs
     Returns:
         None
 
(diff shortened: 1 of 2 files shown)


watertap/core/util/initialization.py
```diff
@@ -151,7 +151,7 @@ def interval_initializer(
         feasibility_tol : tolerance to use for FBBT (default: 1e-6)
         default_initial_value: set uninitialized variables to this value (default: 0.0)
         logger : logger to use (default: watertap.core.util.initialization)
-
+        fail_flag: boolean that raises exception if True or logs warning when an exception occurs
```
kurbansitterley: this docstring is a bit cryptic imo... why not use the same dosctring for `fail_flag` as in the other functions in the file?

```
fail_flag : Boolean argument to specify error or warning
(Default: fail_flag=False produces logger warning. Set fail_flag=True to raise an error and stop
the initialization routine.)
```

watertap/core/util/tests/test_initialization.py
```diff
@@ -11,9 +11,9 @@
 #################################################################################
 
 import pytest
-
+from pyomo.common.errors import InfeasibleConstraintException
 from pyomo.environ import ConcreteModel, Var, Constraint, ConstraintList
-
+import re
```
kurbansitterley: standard library imports should be at the top of the file

watertap/core/util/tests/test_initialization.py
```diff
@@ -11,9 +11,9 @@
...
 from watertap.core.solvers import get_solver
```
kurbansitterley: I think when we got our own solver, someone did a ctrl+F, "replace all" resulting in this out-of-order import for the solver (i.e., grouped with `idaes` imports rather than `watertap` imports).

Could be corrected in this file or not for this PR, but should be a global change we make at some point (does pylint not check this sort of thing???)

kurbansitterley (changes requested):
great addition, a few easy changes (you could convince me on the docstring comment)

watertap/core/util/initialization.py
adam-a-a: good point

watertap/core/util/tests/test_initialization.py
adam-a-a: I think that this check may be in place on the IDAES repo but for some reason does not come into play here. Never delved into the pylint workflow so unsure whether there is something different between pylint configurations on IDAES and this repo.

adam-a-a: THat said, it isn't high on our list of priorities but would be nice to have.

adam-a-a:
@kurbansitterley addressed the changes

watertap/core/util/initialization.py
MarcusHolly: Should False really be the default? Imo, most users are unlikely to even know that these config options exist, so we should assume the default will take effect more often than not. So I think it's better to make the user manually relax the error if they're confident that they've encountered a "select case where InfeasibleConstraintException will be raised" but the model can still get to the right point. If False, is the default, it allows unaware users or users that have "warning fatigue" (e.g. there are some flowsheets that display a lot of warnings, so this one may get lost in the mess) to ignore the warning when it ought to be addressed

MarcusHolly (commented):
LGTM - just one small comment

adam-a-a:
@MarcusHolly I considered that, but interval_initializer was something that was incorporated after the fact across most models. For some models, its implementation was unsuccessful from what I recall, so interval initializer just wasn't included in the initialization routine of those models. So considering that this was meant to be more of a performance enhancer for overall initialization routines--and that the initialization routine itself can still raise an initializationerror after interval initializer fails and is bypassed, I thought that this would be fine.

That's the first point. The counterargument to the fact that the user may overlook an InfeasibleConstraintException warning is that even if they were to get the exception raised, a new user likely wouldn't know what to do and wouldn't realize that they can set fail_flag=False and try again. Additionally, new and seasoned users alike can find it challenging to figure out the root of the issue even when the InfeasibleConstraintException is raised, which makes me think that the ideal default should be one that reduces barriers and yields a path to success more easily. In the select use case that brought this to my attention, before interval initializer was introduced to that flowsheet, I could easily sweep across a range of concentrations and recoveries to solve the model. Now, with interval initializer _enforced_, I cannot. Thus, I personally would opt for interval initializer to be optional by default--it'll still run under the hood during initialization, and if successful, then great. If overall initialization fails, then it'll take a more advanced user to dig through logger messages anyhow because the initialization issue likely wouldn't be trivial.

MarcusHolly (approved):
Still don't agree 100%, but it's clear you've given this some thought, so LGTM. I don't think it's a huge deal either way, but I agree setting the default to True would cause a few more headaches.

kurbansitterley:
> a new user likely wouldn't know what to do and wouldn't realize that they can set fail_flag=False and try again

I am sorta 50/50 about the default being `True` or `False` but this tips me more toward `False` as the default.

watertap/core/util/tests/test_initialization.py
```diff
@@ -1,5 +1,5 @@
 #################################################################################
-# WaterTAP Copyright (c) 2020-2024, The Regents of the University of California,
+# WaterTAP Copyright (c) 2020-2025, The Regents of the University of California,
```
kurbansitterley, line 2: @adam-a-a I have been curious about when we were going to address this issue in main (I have been doing it in REFLO all year). Should it just be a "change as you go" type thing?

doesn't really matter for this PR but the change caught my eye.

kurbansitterley (approved):
LGTM

adam-a-a:
For the record and for future consideration, let me put it another way. When this was rolled out, folks did not test whether its usage was successful across a range of conditions. Admittedly, those ranges of conditions might also be untested for various models. Putting that aside, when interval_initializer was successful for a given model—I.e., the model didn’t fail during or after initialization for whatever standard test cases were in place, it was included in the initialization routine of said model.

LSRRO is one flowsheet that incorporates interval_initializer through the RO base class. In several cases where I change the feed concentration and feed flow rate to ones that deviate from our default feed conditions, the flowsheet will fail during initialization at the interval_initializer step. However, if I pass over the associated exception, the flowsheet continues with the rest of its original initialization routine. If an InitializationError arises, it is raised. If SequentialDecomposition fails to converge, the routine continues on to attempt to solve the square problem, and then the optimization problem. The square problem may fail to converge, but the optimization will succeed. This is the way that flowsheet has always been used, so there is nothing new there.

I’d rather set the default so that initialization would NOT be haulted at the interval initialization step for all models. The alternative would mean that I’d need to patch in a specific deviation for RO base, or add an arg to its initialization_build method that will allow me to toggle the interval initializer’s enforcement, which is more of an effort to roll out if doing that across the board for all models. If we wanted something like that, I’d recommend we log that as an issue and deal with it in a future PR, and accept this PR as an intermediate remedy.

PR #1634, opened 2025-08-28
BSM2 GUI Fix

MarcusHolly:
## Fixes/Resolves:

Fixes an issue in the UI where the user had to modify the default values for the ASP reactor volumes in order for the GUI to solve

- Updates the build_flowsheet function in the bsm2_ui.py file

### Legal Acknowledgement


MarcusHolly:
Not sure how to address these Linux failures... I see that the one-click native installer doesn't even have a Linux download option. Is the GUI operational on Linux? If not, can we ignore these failures?

adam-a-a (approved):

PR #1645, opened 2025-09-18
Adds option to relax DOFs during initlization in RO

avdudchenko:
## Fixes/Resolves:

Many of our models can be difficult to initialize with 0DOF's, however, almost all models require that there are0 DOFs at start of initialization and do not give a user option to relax the system. 

This PR adds an option to RO to allow the user to specify the expected number of DOFs during initialization. 

## Summary/Motivation:
Expands options for initialization configuration of RO, but ideal other PR's will expand this to other models in the future. 

- adds initialization_degrees_of_freedom option to RO intilization function
- adds example and tests to RO_with energy recovery.

### Legal Acknowledgement


dallan-keylogic (approved):
These changes look reasonable. Note that relaxing the degrees of freedom will be much harder using the new initialization API, but we can worry about that later.

watertap/flowsheets/RO_with_energy_recovery/tests/test_RO_with_energy_recovery_simulation.py
```diff
@@ -225,6 +226,43 @@ def test_initialize_system(self, system_frame):
             6.885e6, rel=1e-3
         )
 
+    @pytest.mark.component
+    def test_initialize_system_with_relaxation(self, system_frame):
+        m = build(erd_type="pressure_exchanger")
+
+        set_operating_conditions(m)
+        initialize_system(m, solver=solver, relaxed_initialization=True)
+        m.fs.RO.area_objective.deactivate()
```
adam-a-a: isn't this deactivated in initialize_system? Should we instead verify deactivation? What about deleting the objective entirely after the initialization routine?

watertap/flowsheets/RO_with_energy_recovery/tests/test_RO_with_energy_recovery_simulation.py
```diff
@@ -225,6 +226,43 @@ def test_initialize_system(self, system_frame):
...
+        # check results across pressure exchanger, proxy for both upstream and downstream of RO
```
adam-a-a, line 237: Should add a check here to assert expected degrees of freedom.

watertap/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
```diff
@@ -372,14 +364,28 @@ def solve(blk, solver=None, tee=False, check_termination=True):
...
+            expr=(m.fs.RO.recovery_mass_phase_comp[0, "Liq", "H2O"] - target_recovery)
+            ** 2
+        )
 
+        expected_DOFs = 1
+        m.fs.RO.initialize(
+            optarg=optarg, initialization_degrees_of_freedom=expected_DOFs
+        )
+        m.fs.RO.recovery_mass_phase_comp[0, "Liq", "H2O"].fix(target_recovery)
+        m.fs.RO.area_objective.deactivate()
```
adam-a-a: Might be good to log what's happening here so that the user can optionally track. Something that gives visibility to what is happening when I choose relaxed initialization

watertap/flowsheets/RO_with_energy_recovery/tests/test_RO_with_energy_recovery_simulation.py
avdudchenko: Yup, removed it.

watertap/flowsheets/RO_with_energy_recovery/tests/test_RO_with_energy_recovery_simulation.py
avdudchenko: Yeap done.

watertap/flowsheets/RO_with_energy_recovery/tests/test_RO_with_energy_recovery_simulation.py
avdudchenko: Regarding deleting. We can, but I found some odd behaviors when doing that. So I think its safer to just deactivate and leave it alone. In practice, you would really use this with like LSRRO, or more complex flowsheet, so you would do cycles of reactivate/activate.

watertap/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
avdudchenko: Added some print statements,as we don't have a formal logger on the flowsheet.

kurbansitterley (approved):

ksbeattie:
@adam-a-a and/or @bknueven would like to take another review of this.

watertap/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
```diff
@@ -372,14 +364,36 @@ def solve(blk, solver=None, tee=False, check_termination=True):
...
 
+        expected_DOFs = 1
+        print(
+            f"Unfixed membrane area, and added objective to achieve target recovery of {target_recovery}"
+        )
+        m.fs.RO.initialize(
+            optarg=optarg, initialization_degrees_of_freedom=expected_DOFs
+        )
+        m.fs.RO.recovery_mass_phase_comp[0, "Liq", "H2O"].fix(target_recovery)
+        m.fs.RO.area_objective.deactivate()
```
bknueven: We should probably just delete the objective instead of deactivating it, and check that the deletion is successful.

watertap/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery.py
```diff
@@ -372,14 +364,36 @@ def solve(blk, solver=None, tee=False, check_termination=True):
...
-def initialize_system(m, solver=None):
+def initialize_system(m, solver=None, relaxed_initialization=False):
     if solver is None:
         solver = get_solver()
     optarg = solver.options
 
     # ---initialize RO---
-    m.fs.RO.initialize(optarg=optarg)
+    if relaxed_initialization:
+        print("Initializing RO with relaxed recovery specification")
```
bknueven: These print statements should probably be logging statements.

bknueven (changes requested):

bknueven (approved):

adam-a-a (approved):

Issue #1679, opened 2025-11-01
Update pylint version to match pyomo


sufikaur:
The Pyomo 6.9.5 release broke pylint testing in IDAES which is pinned to a very old version of pylint, initiating an upgrade to version 3.3.9. Check if this causes downstream effects in prommis.

See [PR 1684](https://github.com/watertap-org/watertap/pull/1684)

ksbeattie:
Should be addressed by #1684

PR #1684, opened 2025-11-10
Update pylint version to match pyomo (v 3.3.9)

sufikaur:
## Fixes/Resolves:

[Issue 1679](https://github.com/watertap-org/watertap/issues/1679) 
See  [idaes PR 1683](https://github.com/IDAES/idaes-pse/pull/1683) 

-
-

### Legal Acknowledgement


sufikaur:
After looking at each test, I think that the warnings can be disabled globally.
There are 2 tests that I would like others to confirm will not run into issues.

Currently, I have disabled on each line, but after reviewing, they can be generalized

watertap/unit_models/zero_order/coag_and_floc_zo.py
```diff
@@ -208,7 +208,8 @@ def rule_chem_flow(blk, t, j):
             elif j == "polymer":
                 chemical_dosage = blk.polymer_dose[t]
             return blk.chemical_flow_mass[t, j] == pyunits.convert(
-                chemical_dosage * blk.properties[t].flow_vol,
+                chemical_dosage
+                * blk.properties[t].flow_vol,  # don't fully understand this test
```
sufikaur: I would appreciate someone else looking at this test. @bknueven do you recall this model?

bknueven: This is fine, if clunky.

That said, I think in many, many places, instead of suppressing `pylint` we should instead add an additional `else` clause which raises an exception. WaterTAP is big enough that we should have our own exceptions anyways.

E.g., in this case the `else` clause could raise a `WaterTAPDeveloperError`, or if we want to be cute, a `FronzenPipes` error. It doesn't even need a message if we know we'll never hit the `else` clause upon inspection.

sufikaur: Thank you for taking a look and sounds good-- suppressing was a temporary step and throwing a dev error makes sense.

watertap/tools/oli_api/util/watertap_to_oli_helper_functions.py
```diff
@@ -78,6 +78,7 @@ def get_oli_name(watertap_name: str) -> str:
     components = watertap_name.split("_")
     if len(components) == 0:
         raise IOError(f" Unable to parse solute '{watertap_name}'.")
+    molecule = ""
     if len(components) == 1:
         molecule = components[0]
     elif len(components) == 2:
```
bknueven: I think this is another case where we should raise an exception if neither condition is satisfied.

requirements-dev.txt
```diff
@@ -5,8 +5,8 @@ pre-commit
 pytest-cov
 
 # TODO pin exact version for pylint and astroid
```
MarcusHolly: Can resolve this TODO now

watertap/core/util/chemistry.py
```diff
@@ -122,8 +122,11 @@ def get_charge_group(charge: int) -> str:
         group = "Neutrals"
     elif charge > 0:
         group = "Cations"
-    elif charge < 0:
+    elif charge < 0:  # could change this to an else if we are confiedent charge is int
```
MarcusHolly: I think we can change this to an else w/o any issues

watertap/core/util/initialization.py
```diff
@@ -87,6 +87,8 @@ def check_dof(blk, fail_flag=False, logger=_log, expected_dof=0):
 
     """
     if degrees_of_freedom(blk) != expected_dof:
+        # can fix by initizalizing msg or can disable message
+        msg = "Message failed to initialize."
```
MarcusHolly: I would probably just replace the else clause with a generic WaterTAPDeveloperError because the issue is that the else clause is calling a message that hasn't necessarily been initialized yet, right?

watertap/costing/unit_models/gac.py
```diff
@@ -101,6 +103,11 @@ def cost_gac_gravity(blk):
 def _build_gac_cost_param_block(blk, contactor_type):
 
     adsorbent_unit_cost_coeff_data = {0: 4.58342, 1: -1.25311e-5}
+    # contactor_cost_coeff_data, other_cost_param_data, energy_consumption_coeff_data = (
+    #     {},
+    #     {},
+    #     {},
+    # )  # other option is to initialize as empty dicts and fill based on contactor type
```
MarcusHolly: I think this solution is fine

watertap/unit_models/uv_aop.py
```diff
@@ -862,13 +862,17 @@ def calculate_scaling_factors(self):
         # these variables should have user input, if not there will be a warning
         if iscale.get_scaling_factor(self.uv_intensity) is None:
             sf = iscale.get_scaling_factor(self.uv_intensity, default=0.1, warning=True)
-        iscale.set_scaling_factor(self.uv_intensity, sf)
-
+        # take a look at this test together
+        iscale.set_scaling_factor(
+            self.uv_intensity, sf  # pylint: disable=used-before-assignment
+        )
```
MarcusHolly: Not sure what you mean by this comment

watertap/unit_models/uv_aop.py
```diff
@@ -862,13 +862,17 @@ def calculate_scaling_factors(self):
...
         if iscale.get_scaling_factor(self.exposure_time) is None:
             sf = iscale.get_scaling_factor(
                 self.exposure_time, default=1e-2, warning=True
             )
-            iscale.set_scaling_factor(self.exposure_time, sf)
+            iscale.set_scaling_factor(
+                self.exposure_time, sf
+            )  # why is this one set inside the if but none of the rest?
```
MarcusHolly: Good catch - this should not be inside the if statement

requirements-dev.txt
kurbansitterley: and be sure to remove the comment 😁

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -2641,8 +2641,12 @@ def assert_electroneutrality(
 
         # touch this var since it is required for this method
         self.conc_mol_phase_comp
-
+        """this could also be defined 'state_var["Liq", adjust_by_ion].fix(ion_adjusted)'
```
kurbansitterley: what is this doc string / comment???

watertap/costing/unit_models/ion_exchange.py
```diff
@@ -268,6 +270,10 @@ def cost_ion_exchange(blk):
 
     elif ix_type == "anion":
         resin_cost = ion_exchange_params.anion_exchange_resin_cost
+    else:
+        raise WaterTapDeveloperError(
```
kurbansitterley: This gets checked in the unit model directly via checking of the charge on a component. It should be essentially impossible to get to this point.

... but maybe that is the point of a developer error?

watertap/core/util/initialization.py
kurbansitterley: I remember there was a reason we did a msg here and not an error or log ... I think

requirements-dev.txt
sufikaur: Haha the comment has been here for 2 years! Changed it to more definite statement

watertap/costing/unit_models/gac.py
sufikaur: Just confirming, this solution meaning the initializing empty dicts

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -2641,8 +2641,12 @@ def assert_electroneutrality(
 
         # touch this var since it is required for this method
         self.conc_mol_phase_comp
-
+        """this could also be defined 'state_var["Liq", adjust_by_ion].fix(ion_adjusted)' 
+            on line 2606 but isn't necessary because the same check on 2675 does define it and 
+            adjust_by_ion will either be none or defined since it is a parameter to the method
+            """
```
sufikaur: Any thoughts on this? @MarcusHolly 
Is disabling the best course of action?

watertap/unit_models/uv_aop.py
sufikaur: Mostly referring to the comment about the if statement below. Needed more eyes in particular on this test

watertap/unit_models/uv_aop.py
sufikaur: Shouldn't all be inside the if statement? I think the fact that they are outside is causing the sf possibly used before assignment error

I simply disabled for now since the comment above says that an error would be thrown elsewhere

watertap/costing/unit_models/gac.py
MarcusHolly: Yes

watertap/property_models/multicomp_aq_sol_prop_pack.py
MarcusHolly: I think it's fine as is

watertap/custom_exceptions.py
```diff
@@ -0,0 +1,19 @@
+class FrozenPipes(Exception):
+    """
+    Custom exception for WaterTap developer errors.
```
MarcusHolly:
```suggestion
    Custom exception for WaterTAP developer errors.
```

MarcusHolly (approved):

watertap/unit_models/uv_aop.py
```diff
@@ -862,18 +862,14 @@ def calculate_scaling_factors(self):
...
         iscale.set_scaling_factor(
             self.uv_intensity, sf  # pylint: disable=used-before-assignment
         )
         if iscale.get_scaling_factor(self.exposure_time) is None:
             sf = iscale.get_scaling_factor(
                 self.exposure_time, default=1e-2, warning=True
             )
-            iscale.set_scaling_factor(
-                self.exposure_time, sf
-            )  # why is this one set inside the if but none of the rest?
```
bknueven: I think the answer here is this one is correct, and the other are wrong. The code doesn't make sense otherwise, right?

watertap/unit_models/uv_aop.py
```diff
@@ -862,14 +862,14 @@ def calculate_scaling_factors(self):
...
+        iscale.set_scaling_factor(
+            self.uv_intensity, sf  # pylint: disable=used-before-assignment
+        )
         if iscale.get_scaling_factor(self.exposure_time) is None:
             sf = iscale.get_scaling_factor(
                 self.exposure_time, default=1e-2, warning=True
             )
-            iscale.set_scaling_factor(self.exposure_time, sf)
-
+        iscale.set_scaling_factor(self.exposure_time, sf)
```
bknueven: Just to reiterate -- I think all the calls to `set_scaling_factor` in this method should be *inside* the associated `if` statement. That's on the only way this code makes sense.

watertap/costing/unit_models/gac.py
```diff
@@ -101,6 +102,11 @@ def cost_gac_gravity(blk):
 def _build_gac_cost_param_block(blk, contactor_type):
 
     adsorbent_unit_cost_coeff_data = {0: 4.58342, 1: -1.25311e-5}
+    contactor_cost_coeff_data, other_cost_param_data, energy_consumption_coeff_data = (
+        {},
+        {},
+        {},
+    )
```
bknueven: I think this is another one of those times we should raise an exception in an `else`. Otherwise we will get an error elsewhere and it'll be difficult to trace back here.

watertap/property_models/water_prop_pack.py
```diff
@@ -1059,9 +1060,17 @@ def rule_therm_cond_phase(b, p):
...
                     1 / 3
                 )
+            else:
+                raise FrozenPipes(
+                    f"Index '{p}' is not valid for indexed component 'fs.stream[0].therm_cond_phase"
+                )
             return (
                 b.therm_cond_phase[p]
-                == 10**log10_kw * 1e-3 * pyunits.W / pyunits.m / pyunits.K
+                == 10**log10_kw  # pylint: disable=possibly-used-before-assignment
```
bknueven: This seems like a failing of `pylint` if we really need this here -- this variable is either set or an exception is raised.

bknueven (commented):
Thanks @sufikaur. I found a few more small things, but I think this is about ready.

kurbansitterley (approved):

watertap/costing/unit_models/ion_exchange.py
bknueven: Yes. If somebody changes something upstream which breaks that assumption, we'll catch it here instead of just charging forward.

watertap/property_models/water_prop_pack.py
sufikaur: hmm any thoughts then on what do do? @bknueven

watertap/costing/unit_models/gac.py
sufikaur: changed this to an else

watertap/property_models/multicomp_aq_sol_prop_pack.py
sufikaur: removed

bknueven (approved):

PR #1685, opened 2025-11-11
Add WaterTAP Flowsheet block for simpler flowsheet assembly.

avdudchenko:
## Summary/Motivation:
Currently, we spend alot of time building custom flowsheets that require full re-parametrization of each unit model. This makes the units non-portable, and results in duplicate code. We also use low-level API for building flowsheets, where all units have to be manually connected via arcs, and propagated. This can make sense in a few scenarios (Where a flowsheet has a very complex connection structure), but in most cases, the connections and their flow is mostly linear (even in complex processes like LSRRO). 

The WaterTAPFlowsheetBlock brings a layer of abstraction for units with focus on providing a standard building platform for flowsheets where unit models become portable between flowsheets, and where time spent on configuring each unit model is minimized. It defines standard functions necessary to fully define a unit model from fixed operation variables to scaling factors, and additionally provides a method for standard connection management. A full example of benefits/implementation can be found in [this branch and flowsheet](https://github.com/avdudchenko/watertap/blob/adding_reaktoro_flowsheets_units/watertap/flowsheets/reaktoro_enabled_flowsheets/flowsheets/softening_acid_ro.py)

An example workflow with unit models build on watertapflowsheet block may look as follows:

```
# Define our flowsheet and properties
m.fs = FlowsheetBlock()
m.fs.costing = WaterTAPCosting()
m.fs.properties = MCASParameterBlock(**mcas_props)

# Watertap flowsheet units 
# Here feed specs would be a dict that contains information about feed composition.
m.fs.feed = MultiCompFeed(
        default_property_package=m.fs.properties,
        **feed_specs,
    )

#Here softening specs would be a dict that defines reactions and solids that would form
m.fs.softening_unit = PrecipitationUnit(
    default_property_package=m.fs.properties,
    default_costing_package=m.fs.costing,
    **softeining_specs)

# standard IDAES units
m.fs.product= Product(property_package=m.fs.seawater_props)
m.fs.solids_waste =  Product(property_package=m.fs.seawater_props)

# this builds all the arcs for us and adds tracking for thier propagation
m.fs.feed.outlet.connect_to(m.fs_softening_unit.inlet)
m.fs.fs_softening_unit.treated.connect_to(m.product.inlet) # connects to normal IDAES units!
m.fs.fs_softening_unit.waste.connect_to(m.solids_waste.inlet) # connects to normal IDAES units!
# standard arc expansion!
TransformationFactory("network.expand_arcs").apply_to(m)

m.fs.feed.fix_and_scale() # fixes operation and scales the unit model

m.fs.feed.initialize() # intialzies and propagates all outlet connections
m.fs.softening_unit.initialzie() #initlaizes and propages all outlet connections

# the outlets from softening were propagated by the WaterTAP flowsheet block, so we can just initialize these!
m.fs.product.initialize()
m.fs.solid_waste.initalize()

... do a box solve if wanted 

# use built-in optimization config function to do our optimization. 
m.fs.softening.set_optimization_operation()

.... go on to do optimization etc. 
```

- Add watertapflowsheet block
- Add connection utilities that manage arcs and equality constraints for unit connections
- Add reporting utility for building custom reports for flowsheet units. 

### Legal Acknowledgement


watertap/core/tests/test_watertap_flowsheet_block.py
```diff
@@ -0,0 +1,218 @@
...
+        ConfigValue(
+            doc="Feed flow rate (m^3/h)",
+            default=0.001 * pyunits.m**3 / pyunits.s,
+        ),
+    )
+
+    def build(self):
+        super().build()
+        self.feed = Feed(property_package=self.config.default_property_package)
+        self.solute_type = list(self.config.default_property_package.solute_set)[0]
```
kurbansitterley, line 67: Do you have examples of this for other property models?

watertap/core/watertap_flowsheet_block.py
```diff
@@ -0,0 +1,177 @@
...
+from watertap.core.util.report import (
+    build_report_table,
+)
+
+
+@declare_process_block_class("WaterTapFlowsheetBlock")
+class WaterTapFlowsheetBlockData(FlowsheetBlockData):
+    CONFIG = FlowsheetBlockData.CONFIG()
+    CONFIG.declare(
+        "default_costing_package",
```
kurbansitterley, line 35: I am guessing you adopted this prepending of `default_` nomenclature from `FlowsheetBlock` config, do you know why they do this? It is strange. Like, why not just `costing_package` or `property_package`. There could be still be defaults.

watertap/core/util/connections.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def __init__(self, name, port, var_dict, unit_block_reference):
+        self.name = name
+        self.port = port
+        self.unit_block_reference = unit_block_reference
+        if var_dict != None and isinstance(var_dict, dict) == False:
+            raise TypeError(
+                "Var dict must be a dictionary with structure {'variable name': pyomo.var}"
+            )
+        self.var_dict = var_dict
```
kurbansitterley: outside of the use case for `ph` is there another reason to use the `var_dict`?

watertap/core/watertap_flowsheet_block.py
```diff
@@ -0,0 +1,177 @@
...
+
+    def build(self):
+        self.outlet_connections = []
+        super().build()
+
+    def fix_and_scale(self):
+        self.set_fixed_operation()
+        self.scale_before_initialization()
+
+    def initialize_unit(self, **kwargs):
```
kurbansitterley: Should have the methods that are required raise a `NotImplementedError` in this base

kurbansitterley:
I would be interested in seeing other more complex unit implementations if that is somewhere

watertap/core/tests/test_watertap_flowsheet_block.py
avdudchenko: This should work for nay property model, the implementation here assume a single solute, but can be extended to work with MCAS as well.  (I know it works with Seawater, NaCl and MCAS prop pacakges at least..).

watertap/core/watertap_flowsheet_block.py
avdudchenko: The based on idaes flowsheet block. The idea is that you can have more then one property block and maybe even costing blocks on a flowsheet model. 

Example is using RO with MCAS, the default property package is "MCAS" and ro_property_pacakges is NaCl prop pack or SeawaterPropPack, etc. Its just very explicit in the supplied prop pack is "default" acting prop pack that this flowsheet unit can work with.

avdudchenko:
> I would be interested in seeing other more complex unit implementations if that is somewhere

@kurbansitterley  If you read the PR text, I link directly to [reaktoro enabled flowsheet examples](https://github.com/avdudchenko/watertap/tree/adding_reaktoro_flowsheets_units/watertap/flowsheets/reaktoro_enabled_flowsheets) that use this stucture...(You should also have access to [hpro_analysis repo ](https://github.com/avdudchenko/hpro_analysis) which is private if anyone else try to acces (ping me on slack and I can invite you)).

watertap/core/watertap_flowsheet_block.py
```diff
@@ -0,0 +1,177 @@
...
+        """Developer should implement a routine to unfix variables for unit to perform optimization"""
+
+    def scale_before_initialization(self, **kwargs):
+        """Developer should implement scaling function to scale unit using
+        default values before initialization routine is ran"""
+
+    def scale_post_initialization(self, **kwargs):
+        """Developer should implement scaling function to scale unit after initialization routine is ran"""
+
+    def register_port(self, name, port=None, var_list=None):
```
adam-a-a, line 89: Should we enforce that any of these be developed if subclassing from this? If so, which ones should be deemed "required"? The reason to do so would really be to help the user to make sure to implement methods that provide the most value and make life easier.

adam-a-a: ANd this question goes for any of the methods that are placeholders right now

avdudchenko: It depends on the model, but we need to include scale_before_initialize and set_fixed_operation for initialization. Ideally, some default optimization configuration as well

watertap/core/util/tests/test_connections.py
```diff
@@ -0,0 +1,168 @@
...
+        port=m.fs.feed.outlet,
+        var_dict={"pH": m.fs.feed.ph},
+        unit_block_reference=m.fs.feed,
+    )
+    m.fs.inlet_port_container = PortContainer(
+        name="inlet2",
+        port=m.fs.product.inlet,
+        var_dict={"pH": m.fs.product.ph},
+        unit_block_reference=m.fs.product,
+    )
```
dallan-keylogic, line 60: Is it common to create a completely new variable and attach it to a property block? Why not just add it to the properties?

watertap/core/util/tests/test_connections.py
```diff
@@ -0,0 +1,168 @@
...
+def test_connect_with_port_registration(build_test_model):
+    m = build_test_model
+
+    # Connect the two PortContainers
+    def register_outlet_connection(connection):
+        m.fs.feed.registered_connections = []
+        m.fs.feed.registered_connections.append(connection)
+        print(m.fs.feed.registered_connections)
+
+    m.fs.feed.register_outlet_connection = register_outlet_connection
```
dallan-keylogic, line 89: Why are you creating a new function in a test and attaching it to the unit model? If the user is sophisticated enough to write such a function, they're sophisticated enough to use the low level API.

watertap/core/tests/test_watertap_flowsheet_block.py
```diff
@@ -0,0 +1,218 @@
...
+            index=("Liq", "H2O"),
+        )
+        self.config.default_property_package.set_default_scaling(
+            "flow_mass_phase_comp",
+            1
+            / self.feed.properties[0]
+            .flow_mass_phase_comp["Liq", self.solute_type]
+            .value,
+            index=("Liq", self.solute_type),
+        )
```
dallan-keylogic, line 120: I don't want to create new functions based on the old scaling API.

watertap/core/tests/test_watertap_flowsheet_block.py
```diff
@@ -0,0 +1,218 @@
...
+from pyomo.environ import (
+    assert_optimal_termination,
+)
+import pytest
+
+from io import StringIO
+
+
+@declare_process_block_class("FlowsheetFeed")
+class FlowsheetFeedData(WaterTapFlowsheetBlockData):
```
dallan-keylogic, line 45: Why not just subclass `Feed`? Why create a new object one level higher than `Feed`?

watertap/core/util/connections.py
```diff
@@ -0,0 +1,186 @@
...
+        if self.var_dict != None:
+            for var, obj in self.var_dict.items():
+                obj.fix()
+
+    def unfix(self):
+        """this will unfix the port and all variables in the var_dict"""
+        self.port.unfix()
+        if self.var_dict != None:
+            for var, obj in self.var_dict.items():
+                obj.unfix()
```
dallan-keylogic, line 65: We already have a `fix_initialization_states` method for the `UnitModelBlock`. Is there functionality being added here that isn't already possible from that method?

watertap/core/util/connections.py
```diff
@@ -0,0 +1,186 @@
...
+        arc = Arc(source=self.get_port(outlet), destination=self.get_port(inlet))
+
+        def get_safe_name(name):
+            return name.replace(".", "_").replace("-", "_")
+
+        arc_name = f"{get_safe_name(outlet.name)}_to_{get_safe_name(inlet.name)}"
+        outlet.unit_block_reference.add_component(
+            arc_name,
+            arc,
+        )
```
dallan-keylogic, line 113: Has this been tested when one port extends another port? I know that the `name` attribute behaves in unexpected ways when you create a `Reference` to a `Var` (instead of getting, e.g., `fs.unit.outlet.flow_mol_phase_comp`, you get `IndexedVar` instead).

watertap/core/util/connections.py
```diff
@@ -0,0 +1,186 @@
...
+
+        # find it, if it was not created correctly, we will get an error
+        self.registered_arc = outlet.unit_block_reference.find_component(arc_name)
+        if self.registered_arc is None:
+            raise ValueError(f"Arc was not created correctly for {arc_name}")
```
dallan-keylogic, line 118: Do we know of cases in which this `ValueError` is going to happen? If not, it should be changed to `BurntToast` (or at least `RuntimeError`).

watertap/core/util/connections.py
```diff
@@ -0,0 +1,186 @@
...
+                            "Created arc constraint: %s",
+                            f"eq_{outlet_key}_{outlet.name}_to_{inlet.name}",
+                        )
+                        # scale the constraint
+                        sf = iscale.get_scaling_factor(outlet.var_dict[outlet_key])
+                        if sf != None:
+                            iscale.constraint_scaling_transform(
+                                constraint,
+                                sf,
+                            )
```
dallan-keylogic, line 170: [This should happen automatically](https://github.com/IDAES/idaes-pse/blob/e7e13b89ed22d3d0c33ce9c902a35bff5a947558/idaes/core/util/scaling.py#L213) for all variables on arc constraints when you run `iscale.calculate_scaling_factors` on the entire flowsheet.

watertap/core/watertap_flowsheet_block.py
```diff
@@ -0,0 +1,177 @@
...
+    PortContainer,
+    ConnectionContainer,
+)
+from watertap.core.util.report import (
+    build_report_table,
+)
+
+
+@declare_process_block_class("WaterTapFlowsheetBlock")
+class WaterTapFlowsheetBlockData(FlowsheetBlockData):
```
dallan-keylogic, line 32: Is this supposed to be an entire flowsheet or just a unit model? It's subclassing the flowsheet class, but the function docstrings seem to hint this is supposed to contain a single unit model. If it contains a single unit model, it should either be a `UnitModeBlock` or be a subclass of the unit model which it contains.

watertap/core/watertap_flowsheet_block.py
```diff
@@ -0,0 +1,177 @@
...
+        self.set_fixed_operation()
+        self.scale_before_initialization()
+
+    def initialize_unit(self, **kwargs):
+        """Developer should implement an initialize routine for their flowsheet model"""
+
+    def initialize(self, **kwargs):
+        """routine to initialize a unit and propagate its connections"""
+        self.initialize_unit()
+        self.propagate_outlets()
```
dallan-keylogic, line 69: This workflow is not compatible with the new method of initialization, in which a separate initializer class is created and attached to the unit/flowsheet as the `default_initializer` attribute.

watertap/core/watertap_flowsheet_block.py
```diff
@@ -0,0 +1,177 @@
...
+
+    def set_optimization_operation(self, **kwargs):
+        """Developer should implement a routine to unfix variables for unit to perform optimization"""
+
+    def scale_before_initialization(self, **kwargs):
+        """Developer should implement scaling function to scale unit using
+        default values before initialization routine is ran"""
+
+    def scale_post_initialization(self, **kwargs):
+        """Developer should implement scaling function to scale unit after initialization routine is ran"""
```
dallan-keylogic, line 87: Similarly, instead of a method, scaling should be done via a class derived from the `CustomScalerBase` and attached as the `default_scaler `attribute.

dallan-keylogic (commented):
This PR seems somewhat confused. Wrappers are created for the Pyomo `Port` class, the Pyomo `Arc` class, and the IDAES `UnitModelBlock` class, but with different method names. Much of this functionality could be accomplished using existing features for those classes.

I think it might be good to add an additional method to the `UnitModelBlock` to fix some of its geometry and performance variables to predetermined values (to represent, for example, a single mass-produced style of unit being used in multiple places) but most of the other functionality should be redirected into the existing `UnitModelBlock` functions.

watertap/core/util/tests/test_connections.py
avdudchenko: Yes, so I guess I could just inherit MCAS and add pH as on-demand property. This is just an example. 

The key issue is that regardless of how we add this new Var to our blocks, it still needs to be passed between unit models. From my understanding of Arc creation, only state props are passed into Arcs (Mass/mol flows, temp, pressure), so we would need to add pH or pE to them as well, but most unit models to do not work with these vars, and we just want to expose them to relevant parts of the unit model (e.g. Reaktoro-PSE or a surrogate model) with out having to create a bunch of new iso_pH constraints for example.  

What is not clear to me is how, with current Arc and on-demand props, or state props, we can selectively create vars like pH and pass only through certain arcs we designate. (e.g. lets say we have the new unit_block and I built a new unit model, I would want to crate the pH wars only on inlet and outlet, and then specify internally how to pass it, in many cases I just want to have a single pH var and pass its reference to the next model... not sure how we do that (e.g. imagine a splitter - it has a single pH var, we don't need a new pH var for each splitter outlet they are all the same...).

In my current workflows, pH and pE is a separate a layer from state props, as they are only handled by Reakter-PSE in select locations.  - This is one reason for why I created container port structure - the other reason is to simplify connecting unit models - but I agree this functionality could be moved to unit_block - and probably should be there as it makes it so much easier to build flowsheet.  I just did not want to try and push it into IDAES till people gave a thumbs up to the concept. If you and others think that adding automatic arc creation functionality is useful, I would be happy to rip this out and add it as a PR to IDAES -> or should it be even in Pyomo?

watertap/core/util/tests/test_connections.py
avdudchenko: This is just to test the functionality works correctly.

watertap/core/tests/test_watertap_flowsheet_block.py
avdudchenko: And most models still use old scaling methods. this is also still compatible with new scaling methods - as you would just drop in the scaling class here and call it.

watertap/core/tests/test_watertap_flowsheet_block.py
avdudchenko: Not sure I understand the question. This creates an example of how to use WaterTAP flowsheet block - we do not want to inherit Feed here as it does not give us functionality from WaterTap flowsheet block.

watertap/core/util/connections.py
avdudchenko: Sure, if the port containers become part of UntiModelBlock structure.

watertap/core/util/connections.py
avdudchenko: hmm not really sure I encountered that in watertap models yet. so not sure. but I should probably write a test for this scenario depending on where we end up putting this code.

watertap/core/util/connections.py
avdudchenko: Yeah, this is more of a BurntToast error I guess. We can update.

watertap/core/util/connections.py
avdudchenko: This is not done for the arc variables, but for the custom vars user added to the port container.

watertap/core/watertap_flowsheet_block.py
avdudchenko: This can be a single unit model or multiple unit models - (E.g. pump + RO), or it could be different functionality (E.g. RO + ReaktoroBlock). I originally envisioned it containing multiple units but its really being used to do both things. It's not quite clean.

watertap/core/watertap_flowsheet_block.py
avdudchenko: Is there an example of it, I have yet to encounter this or see it being used. (And what are the advantages of this approach?

watertap/core/watertap_flowsheet_block.py
avdudchenko: Right, but this would contain a set of scaling classes for each unit model and execute them. 

I still assume the new scaling tools require you to first build the model and then call the scaling class?

watertap/core/util/tests/test_connections.py
dallan-keylogic: > From my understanding of Arc creation, only state props are passed into Arcs

Those are the only properties added to ports when the `UnitModelBlock.add_inlet_port` and `UnitModelBlock.add_outlet_port` are used, but after calling those methods, you can add additional variables to the ports with the `Port.add` method.

> I just want to have a single pH var and pass its reference to the next model... not sure how we do that (e.g. imagine a splitter - it has a single pH var, we don't need a new pH var for each splitter outlet they are all the same...).

Now that we have the linear presolve working, duplicate pH linked via simple equality constraints can be eliminated automatically. But for cases like the splitter, you can just add the same `Var` to both the inlet and outlet ports (like what is done in the [StateJunction](https://github.com/IDAES/idaes-pse/blob/e7e13b89ed22d3d0c33ce9c902a35bff5a947558/idaes/models/unit_models/statejunction.py#L117)).

> This is one reason for why I created container port structure - the other reason is to simplify connecting unit models - but I agree this functionality could be moved to unit_block - and probably should be there as it makes it so much easier to build flowsheet. I just did not want to try and push it into IDAES till people gave a thumbs up to the concept. If you and others think that adding automatic arc creation functionality is useful, I would be happy to rip this out and add it as a PR to IDAES -> or should it be even in Pyomo?

We'll probably determine that on a case-by-case basis. After discussing UI integration with @dangunter , I think we want to track which ports are inlets and which ports are outlets on the `UnitModelBlock` class (see IDAES/idaes-pse#1722). The `connect_to` method might become obsolete with more sophisticated Gen AI-based autocomplete. Automatic propagation of variables along arcs can be done with the [sequential decomposition](https://pyomo.readthedocs.io/en/stable/api/pyomo.network.decomposition.SequentialDecomposition.html) like is done [in the UKy flowsheet.](https://github.com/prommis/prommis/blob/32068be572c4f33d85eb20b8dbf8d9b8db7374f7/src/prommis/uky/uky_flowsheet.py#L1110)

avdudchenko: The Port.add method causes a number of issues when you try to use any of the IDAES visualize tools - and I had issues if you had variable added from multiple blocks to same port for some reason. 

Regarding linear pre-solve - that's great if you can use normal ipopt, and if you are not using GrayBox integration and cyipopot, as well as want your model to solve... in my tests so far linear pre-solve results in less stable models for what ever reason. 

I think if we want to have UI specific capacity thatch standard, we need a decorator that helps developer/user define which ports are inlets or outlet - in this framework, we simply pass the port objects from the unit models, so it still works as we are not creating a new port objects at all. 

I am not sure if `connect_to` being passed over for Gen-AI is a strong argument, I do not want to entrust slop machines in creasing my arcs, or putting simple model creation API behind a paywall.

I am also not sure why you are mentioning automatic propagation, we already do it with propagate method for arcs, here in we just now package it into the unit model so user does not have to handle it using trivial logic, and not having to rely on a tool that might not work with all cases, for example does sequential decomposition work with cyipopt and graybox models? Does it work on all flowsheet configurations? 

I do not mind convenient complex tools like linear-resolve or sequential decomposition being available, but I am not sure why they supersede developing simple and convenient API's.

avdudchenko:
Closing this as there seems to be no interest in adding this directly, and rather the interest is in using AGI and the existing API instead.

adam-a-a:
Reopening this

PR #1686, opened 2025-11-12
Print properties contained in Seawater prop model

adam-a-a:
## Fixes/Resolves:
NA
## Summary/Motivation:
While preparing materials for Academy, realized it isn't very easy to list all properties on a property model.

One would need to dive into source code or technical documentation. 

This PR attempts to add `list_properties` and `print_properties` methods to the SeawaterParameterBlock, using new utility helper functions, `get_property_metadata` and print_property_metadata`.

This also required defining a separate PropertySet for Seawater--otherwise, you'll get back a long list of all supported IDAES properties that aren't necessarily supported in the seawater prop model.

Can extend this to other property models once we cover this PR. 

**Example implementation:**
```
m = ConcreteModel()
m.fs.seawater_properties = SeawaterParameterBlock()
m.fs.seawater_properties.print_properties()
```
**Output:**
```
Property Description         Model Attribute          Units          
---------------------------------------------------------------------
Boiling point elevation      boiling_point_elevation  K              
Mass concentration           conc_mass                kg*m**(-3)     
Specific heat capacity       cp_mass                  J/(kg*K)       
Mass density of solution     dens_mass                kg*m**(-3)     
Mass density of pure water   dens_mass_solvent        kg*m**(-3)     
Latent heat of vaporization  dh_vap_mass              J*kg**(-1)     
Diffusivity                  diffus                   m**2/s         
Enthalpy flow                enth_flow                J/s            
Specific enthalpy            enth_mass                J*kg**(-1)     
Mass flow rate               flow_mass                kg/s           
Molar flowrate               flow_mol_phase_comp      mol/s          
Total volumetric flow rate   flow_vol                 m**3/s         
Mass fraction                mass_frac                dimensionless  
Molality                     molality                 mol/kg         
Mole fraction                mole_frac                dimensionless  
Osmotic coefficient          osm_coeff                dimensionless  
Pressure                     pressure                 Pa             
Osmotic pressure             pressure_osm             Pa             
Vapor pressure               pressure_sat             Pa             
Temperature                  temperature              K              
Thermal conductivity         therm_cond               W/(m*K)        
Dynamic viscosity            visc_d                   Pa*s      
```

- add helper functions to print or return list of dicts of properties 
- add methods to seawater parameterblock that leverage the helper function
- add tests for helper function

### Legal Acknowledgement


watertap/core/util/property_helpers.py
```diff
@@ -0,0 +1,50 @@
...
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and license
+# information, respectively. These files are also available online at the URL
+# "https://github.com/watertap-org/watertap/"
+#################################################################################
+
+import pandas as pd
+
+
+def print_property_metadata(prop_pkg, return_df=False):
```
dangunter: I think a simpler interface would be get_property_metadata. Wrapping that in a print statement is very little burden on the caller and then you don't need this keyword.

watertap/core/util/property_helpers.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+    Print all supported properties from a WaterTAP/IDAES property package.
+
+    Args:
+        prop_pkg: The property model ParameterBlock (e.g., m.fs.properties)
+        return_df: If True, returns a Pandas DataFrame instead of printing.
+    """
+    metadata = prop_pkg.get_metadata()
```
dangunter, line 23: can this fail?

watertap/core/util/property_helpers.py
```diff
@@ -0,0 +1,50 @@
...
+    vars, units, docs = [], [], []
+
+    for v in metadata.properties:
+        vars.append(v._name)
+        units.append(str(v._units))
+        docs.append(v._doc)
+
+    if return_df:
+        return pd.DataFrame(
+            {"Property Description": docs, "Model Attribute": vars, "Units": units}
```
dangunter: I would suggest column headers with no spaces. "Name", "Description", "Units". The prefix "Property" is implied in all cases.

watertap/core/util/property_helpers.py
```diff
@@ -0,0 +1,50 @@
...
+        )
+
+    # Pretty-print
```
dangunter: so again, I would just return the data frame. People know how to print a dataframe in a basic way in a terminal (i.e. just pass obj to `print()`) and a notebook will do it for you. If you really must print for the user, I'd suggest a separate method `print_property_metadata` that consists of basically : 
```
pprint.pprint(get_property_metadata(*args).to_string(index=False))
```

watertap/core/util/tests/test_property_helpers.py
```diff
@@ -0,0 +1,64 @@
...
+    expected_cols = ["Property Description", "Model Attribute", "Units"]
+    assert list(df.columns) == expected_cols
+
+    # Check content
+    assert "flow_mass" in df["Model Attribute"].values
+    assert "temperature" in df["Model Attribute"].values
+    assert "kg/s" in df["Units"].values
+
+
+def test_print_property_metadata_pretty_print(capsys):
```
dangunter: I would just test that there are, say, between 100 and 1000 characters in the output.

watertap/property_models/seawater_prop_pack.py
```diff
@@ -736,9 +737,22 @@ def build(self):
         self.set_default_scaling("diffus_phase_comp", 1e9)
         self.set_default_scaling("boiling_point_elevation_phase", 1e0, index="Liq")
 
+    def list_properties(self, return_df=False):
```
dangunter: these seem completely generic, so should be on a base class

watertap/property_models/seawater_prop_pack.py
```diff
@@ -1789,3 +1798,127 @@ def calculate_scaling_factors(self):
 
         # transforming constraints
         transform_property_constraints(self)
+
```
dangunter, line 1803: I realize this is normal in IDAES but I find this way of creating data very verbose.

watertap/property_models/tests/test_seawater_prop_pack.py
```diff
@@ -332,3 +334,45 @@ def configure(self):
...
+def test_list_properties(capsys):
+    m = ConcreteModel()
+    m.fs = FlowsheetBlock(dynamic=False)
+    m.fs.props = props.SeawaterParameterBlock()
+
+    # clear any existing captured output, call list_properties, then capture its output
+    capsys.readouterr()
+    m.fs.props.list_properties()
+    captured = capsys.readouterr().out
+    expected_output = """
```
dangunter: tets for string matches are not a good idea imho. at a minimum use a regex to look for some things least likely to change.

watertap/property_models/seawater_prop_pack.py
adam-a-a: yes, but I started with this first before attempting to add to a base class (which would probably be best on IDAES repo but would slow things down for us)

watertap/property_models/tests/test_seawater_prop_pack.py
adam-a-a: I was expecting this sort of response! I was aiming for a quick test to make sure my seawater prop model was listing the correct properties. Can reconsider this, but I was hoping to get something merged asap and revise in subsequent PR(s)

MarcusHolly (approved):
If the intent is just to get something merged in preparation for Friday, then I think this is good as is. None of Dan's comments strike me as urgent, so these (and other) refinements can be made in future PRs.

watertap/core/util/property_helpers.py
adam-a-a: do you mean make it a method on parameterblock so that I don't need prop_pkg keyword? Or did you mean remove the need for returning df in this function and focus on getting the metadata?

watertap/core/util/property_helpers.py
adam-a-a: Is there an issue with the way it's printing now or any advantage to doing it differently?

watertap/core/util/property_helpers.py
adam-a-a: ok will do!

watertap/core/util/property_helpers.py
adam-a-a: Here's the method (below). I suppose we can add a check for the `_metadata` attribute and raise an exception if it doesn't exist. Do you know of a better check to ensure the property model is a property model? Another idea is to check whether the prop_pkg input is a ParameterBlock.
```
    @classmethod
    def get_metadata(cls):
        """Get property parameter metadata.

        If the metadata is not defined, this will instantiate a new
        metadata object and call `define_metadata()` to set it up.

        If the metadata is already defined, it will be simply returned.

        Returns:
            PropertyClassMetadata: The metadata
        """
        if cls._metadata is None:
            pcm = PropertyClassMetadata()
            cls.define_metadata(pcm)
            cls._metadata = pcm

            # Check that the metadata was actually populated
            # Check requires looking at private attributes
            # pylint: disable-next=protected-access
            if pcm._properties is None or pcm._default_units is None:
                raise PropertyPackageError(
                    "Property package did not populate all expected metadata."
                )
        return cls._metadata
```

watertap/property_models/seawater_prop_pack.py
adam-a-a: Me too, but given the current IDAES functionality and the way metadata is handled and reported back, this was the cleaner way of producing the list of property names actually supported on this property model, without the noise. Open to suggestions if there's a slicker approach, but this is meant a quick fix, with planned refinement in subsequent PRs (this quarter).

watertap/core/util/property_helpers.py
kurbansitterley: I think he means rename this `get_property_metadata` and return the df and then the user can do what they want?

I like the idea of this being a method on ParameterBlock. But could you not just use this function and then `@classmethod` to avoid adding this code to every property model?

watertap/core/util/property_helpers.py
avdudchenko: Does it need to be a dataframe? Why not just a dict?

watertap/core/util/property_helpers.py
avdudchenko: I would argue we should not return a dataframe in the first place. 

The reason is we don't want to maintain code specific to pandas for printing stuff. 

I think we we do want to display/print information we should use native python display tools otherwise we might run into system quirks ... (I wounder if pyomo table printer would work for this - also can look in my PR for WaterTAP flowsheet block for displaying a table - it uses native pyomo printer... ?)

watertap/property_models/seawater_prop_pack.py
avdudchenko: I feel like this is redundant, don't the pyomo vars in the property block contain all of the same information? Would it not be easier to make a list of the "vars" you want to display and simply grab it?

avdudchenko (approved):
Overall looks good, I am not 100% sure about relying on pandas data frames for data storage and printing. 

My biggest question is can't we pull most of the information here from already build pyomo vars defining the supported properties?

I almost wounder if this just need a registration class that lets you track vars you want to report and provide ways to both interact with?

watertap/core/util/property_helpers.py
adam-a-a: I figured such a change would need to be made in IDAES.

ksbeattie:
@dangunter will take a look at this.

dallan-keylogic:
When opening #1797, I discovered that you are missing `energy_density_phase` as a property. It should be added to either the property set here or the one in IDAES. (Also, shouldn't it be `energy_dens_phase` to be in line with our naming convention for density?)

adam-a-a:
> When opening #1797, I discovered that you are missing `energy_density_phase` as a property. It should be added to either the property set here or the one in IDAES. (Also, shouldn't it be `energy_dens_phase` to be in line with our naming convention for density?)

I think this was missed as it might have been added in after I had put this PR up. The PR used that naming convention but we can change: https://github.com/watertap-org/watertap/commit/e50196ee105d4d288e9d1eb4c968a32e3c8526e9#diff-4adce7bb9b34523615766eb320cb2f1e70bc9c4ba105576b1afc3c80e652a7aaR757

adam-a-a:
@dangunter did you get a chance to look at the PropertyMetadata bit in IDAES?

Issue #1682, opened 2025-11-07
module_type config not documented in RO tech ref


adam-a-a:
module_type config option is not documented in the RO model's technical reference. Should also check for any other missing segments in RO documentation while we're at it.

PR #1688, opened 2025-11-14
Add Autodocs for RO Config Args

MarcusHolly:
## Fixes/Resolves:

#1682 

## Summary/Motivation:
Adds the autogenerated docs describing the configuration arguments for 0D RO

### Legal Acknowledgement

### DIFF
--- a/docs/technical_reference/unit_models/reverse_osmosis_0D.rst
+++ b/docs/technical_reference/unit_models/reverse_osmosis_0D.rst
@@ -194,7 +194,9 @@ Equations
 Class Documentation
 -------------------
 
-* :mod:`watertap.unit_models.reverse_osmosis_0D`
+.. automodule:: watertap.unit_models.reverse_osmosis_0D
+    :members:
+    :noindex:
 
 References
 ----------


MarcusHolly:
Is there a reason we don't have the config args displaying like this for every unit model? Just wanted to clarify before I went and made the changes on the rest of the documentation.

kurbansitterley:
> Is there a reason we don't have the config args displaying like this for every unit model? Just wanted to clarify before I went and made the changes on the rest of the documentation.

my guess is because the doc string quality is all over the place

MarcusHolly:
> > Is there a reason we don't have the config args displaying like this for every unit model? Just wanted to clarify before I went and made the changes on the rest of the documentation.
> 
> my guess is because the doc string quality is all over the place

Ok, maybe this PR can just update the RO doc since Academy people may be looking at this soon. Can I just get a second pair of eyes to review this so that we can merge it in the next hour.

kurbansitterley (approved):
LGTM 

imo it would be better to have some kind of table or something that presented the configuration arguments, options, and descriptions clearly because the autdoc results are sort of cryptic, but for a later time.

MuktaHardikar (approved):

MarcusHolly:
Merging this now as a temporary solution, but should still revisit a more permanent solution for this unit model as well as the other with regards to properly documenting the configuration arguments.

PR #1689, opened 2025-11-17
Adding method for fitting spec sheets to standard RO model

avdudchenko:
## Fixes/Resolves:
Currently, we do not provide simple scripts to enable fitting of spec sheet data to our unit models. This creates a folder within flowsheets to store such methods, and kicks it off by adding a script for fitting A/B values to standard RO spec sheets. 

## Summary/Motivation:
We need to provide a simple way for users to fit model parameters from spec sheets. This creates a simple example for fitting RO module to typical spec sheet information for RO modules.  Ideally this will start us on a discussion of what more generalized approach might look like and add other unit models to the list. 

-
-

### Legal Acknowledgement


watertap/flowsheets/specsheet_fitting_tools/ro_module_fitter.py
```diff
@@ -0,0 +1,314 @@
...
+    """
+
+    solver = get_solver()
+
+    m = ConcreteModel()
+    m.fs = FlowsheetBlock(dynamic=False)
+
+    m.fs.properties = props.NaClParameterBlock()
+    # seem feed and ro model
+    m.fs.feed = Feed(property_package=m.fs.properties)
```
adam-a-a: WHy not eliminate Feed entirely and only use the RO instance?

watertap/flowsheets/specsheet_fitting_tools/ro_module_fitter.py
```diff
@@ -0,0 +1,314 @@
...
+
+    m.fs.feed_to_ro = Arc(source=m.fs.feed.outlet, destination=m.fs.RO.inlet)
+
+    TransformationFactory("network.expand_arcs").apply_to(m)
+
+    # specify feed conditions
+    m.fs.feed.properties[0].flow_vol_phase["Liq"].fix(water_production_rate / recovery)
+    m.fs.feed.properties[0].conc_mass_phase_comp["Liq", "NaCl"].fix(feed_conc)
+    m.fs.feed.properties[0].temperature.fix(temperature + 273.15)
+    m.fs.feed.properties[0].pressure.fix(pressure)
```
adam-a-a: eliminating Feed, we would just do 
m.fs.RO.feed_side.properties[0,0].flow
```suggestion
    m.fs.RO.feed_side.properties[0,0].flow_vol_phase["Liq"].fix(water_production_rate / recovery)
    m.fs.RO.feed_side.properties[0,0].conc_mass_phase_comp["Liq", "NaCl"].fix(feed_conc)
    m.fs.RO.feed_side.properties[0,0].temperature.fix(temperature + 273.15)
    m.fs.RO.feed_side.properties[0,0].pressure.fix(pressure)
```

watertap/flowsheets/specsheet_fitting_tools/ro_module_fitter.py
```diff
@@ -0,0 +1,314 @@
...
+
+    print("Degrees of freedom before initialization: ", degrees_of_freedom(m.fs.feed))
+    assert degrees_of_freedom(m.fs.feed) == 0
+    # Solve the feed to get mass flows and concentrations through out
+    result = solver.solve(m.fs.feed)
```
adam-a-a:
```suggestion
    result = solver.solve(m.fs.RO.feed_side.properties[0,0])
```

watertap/flowsheets/specsheet_fitting_tools/ro_module_fitter.py
```diff
@@ -0,0 +1,314 @@
...
+    assert_optimal_termination(result)
+    propagate_state(m.fs.feed_to_ro)
```
adam-a-a: no need to propagate without feed

watertap/flowsheets/specsheet_fitting_tools/ro_module_fitter.py
```diff
@@ -0,0 +1,314 @@
...
+    membrane_name="RO_module",
+    water_production_rate=38 * pyunits.m**3 / pyunits.day,
+    nacl_rejection=99.5 * pyunits.percent,  # percent
+    feed_conc=1500 * pyunits.mg / pyunits.liter,
+    recovery=0.15,
+    module_length=1 * pyunits.m,
+    pressure=150 * pyunits.psi,
+    membrane_area=40 * pyunits.m**2,
+    channel_height=1e-3 * pyunits.m,
+    spacer_porosity=0.95,
```
adam-a-a: spacer porosity is rarely (if ever) provided. I think 0.85 would be a more reasonable default (where 0.95 is a bit closer to that of an open channel).

adam-a-a (commented):
Took a quick first pass at the main utility function. I think this is definitely the type of thing we need to be adding to WaterTAP, and this sort of method in particular has been a long time coming. 

I have some initial comments, but the main question I have is--would it be better to add this as a method to the RO unit? What would be the pros and cons? Without having thought it through thoroughly, my initial reaction is that this would be better as a method on the RO class.

watertap/flowsheets/specsheet_fitting_tools/ro_module_fitter.py
avdudchenko: It was a more transparent process for me as, working with RO feed block is not quiet as straightforward as you can't access flow_vol via the ro inlet port and need to directly interact with feed side properties, which are also connected to other constraints on the model. 

Honestly, I dont think it matters it was just faster for me to setup the feed then spend time debuging the ro feed-side properties block. 

The other reason is you get the error below as it expects flow vol and concentration to be unfixed... before initialization, it's simple to fix but I remember there being issues like this.

"idaes.core.util.exceptions.PropertyPackageError:
While initializing fs.RO.feed_side.properties, the degrees of freedom are -2, when zero is required.
Initialization assumes that the state variables should be fixed and that no other variables are fixed.
If other properties have a predetermined value, use the calculate_state method before using initialize to determine the values for the state variables and avoid fixing the property variables."

watertap/flowsheets/specsheet_fitting_tools/ro_module_fitter.py
avdudchenko: Thats just a random default option for real modules I use 0.9, changed it 0.85.

avdudchenko:
> Took a quick first pass at the main utility function. I think this is definitely the type of thing we need to be adding to WaterTAP, and this sort of method in particular has been a long time coming.
> 
> I have some initial comments, but the main question I have is--would it be better to add this as a method to the RO unit? What would be the pros and cons? Without having thought it through thoroughly, my initial reaction is that this would be better as a method on the RO class.

I removed the feed as per the suggestion. 

Adding it as an explicit method would add unnecessary support code to RO that will need to make a lot of assumptions. I am not really sure what is the best way to go. 

I really think we want to first figure out a generic way to store parameters/specs for unit models so we can create a way to generate standard output files and have unit models injest them to fix the operation. 

It really might be better to create a separate tool that handles model data management 
1) A tool that defines a standard method for storing default unit specs
2) This same tool can then be used to fit data for a given unit model using similar spec files - e.g. a file that defines known parameters and defines what variables to fit. 


There are a few complications -> not all spec sheets are the same or data is the same, some users will need to modify the fitting scripts, some users might have strong opinions on what parameters to fit - > for example, in RO for pressure drop we can fit spare porosity or friction factor correlation values etc. 

I really think, we first should build a number of such "fitting" scripts for different unit models, and then see if its possible to create generalized methods and what the requirements would be, as creating a new set of tools or adding more functionality to RO/etc. will potentially drive us into a corner. Although  creating a data storage method might be the first step.

watertap/flowsheets/specsheet_fitting_tools/tests/expected_ro_results/BW30 PRO-4040.yaml
```diff
@@ -0,0 +1,30 @@
+A:
```
adam-a-a, line 1: I think these yamls should be moved to `watertap/data`, with a specific subdirectory for membrane_element_specs (or whatever we want to name that subdirectory).

kurbansitterley:
> > Took a quick first pass at the main utility function. I think this is definitely the type of thing we need to be adding to WaterTAP, and this sort of method in particular has been a long time coming.
> > I have some initial comments, but the main question I have is--would it be better to add this as a method to the RO unit? What would be the pros and cons? Without having thought it through thoroughly, my initial reaction is that this would be better as a method on the RO class.
> 
> I removed the feed as per the suggestion.
> 
> Adding it as an explicit method would add unnecessary support code to RO that will need to make a lot of assumptions. I am not really sure what is the best way to go.
> 
> I really think we want to first figure out a generic way to store parameters/specs for unit models so we can create a way to generate standard output files and have unit models injest them to fix the operation.
> 
> It really might be better to create a separate tool that handles model data management
> 
> 1. A tool that defines a standard method for storing default unit specs
> 2. This same tool can then be used to fit data for a given unit model using similar spec files - e.g. a file that defines known parameters and defines what variables to fit.
> 
> There are a few complications -> not all spec sheets are the same or data is the same, some users will need to modify the fitting scripts, some users might have strong opinions on what parameters to fit - > for example, in RO for pressure drop we can fit spare porosity or friction factor correlation values etc.
> 
> I really think, we first should build a number of such "fitting" scripts for different unit models, and then see if its possible to create generalized methods and what the requirements would be, as creating a new set of tools or adding more functionality to RO/etc. will potentially drive us into a corner. Although creating a data storage method might be the first step.

I think I could go either way on making it a class method or standalone - but what are thoughts on putting it in a more general location like `watertap/core/utils` or just straight up `watertap/tools` (which actually has no useful tools to anyone without an OLI license atm).

avdudchenko:
> > > Took a quick first pass at the main utility function. I think this is definitely the type of thing we need to be adding to WaterTAP, and this sort of method in particular has been a long time coming.
> > > I have some initial comments, but the main question I have is--would it be better to add this as a method to the RO unit? What would be the pros and cons? Without having thought it through thoroughly, my initial reaction is that this would be better as a method on the RO class.
> > 
> > 
> > I removed the feed as per the suggestion.
> > Adding it as an explicit method would add unnecessary support code to RO that will need to make a lot of assumptions. I am not really sure what is the best way to go.
> > I really think we want to first figure out a generic way to store parameters/specs for unit models so we can create a way to generate standard output files and have unit models injest them to fix the operation.
> > It really might be better to create a separate tool that handles model data management
> > 
> > 1. A tool that defines a standard method for storing default unit specs
> > 2. This same tool can then be used to fit data for a given unit model using similar spec files - e.g. a file that defines known parameters and defines what variables to fit.
> > 
> > There are a few complications -> not all spec sheets are the same or data is the same, some users will need to modify the fitting scripts, some users might have strong opinions on what parameters to fit - > for example, in RO for pressure drop we can fit spare porosity or friction factor correlation values etc.
> > I really think, we first should build a number of such "fitting" scripts for different unit models, and then see if its possible to create generalized methods and what the requirements would be, as creating a new set of tools or adding more functionality to RO/etc. will potentially drive us into a corner. Although creating a data storage method might be the first step.
> 
> I think I could go either way on making it a class method or standalone - but what are thoughts on putting it in a more general location like `watertap/core/utils` or just straight up `watertap/tools` (which actually has no useful tools to anyone without an OLI license atm).

@kurbansitterley Obviously a general tool would live there, but were you suggesting in the interim moving this code (ro_fitter) to tools? Watertap/tools make sense, but i do not think core/utils is a good place. 

@adam-a-a I was not planning to store fitted data and let users generate the data by running this code, but I don't see why not. I will update it.

watertap/flowsheets/specsheet_fitting_tools/ro_module_fitter.py
```diff
@@ -0,0 +1,322 @@
...
+        transformation_scheme="BACKWARD",
+        transformation_method="dae.finite_difference",
+        module_type="spiral_wound",
+        finite_elements=10,
+        has_full_reporting=True,
+    )
+
+    # m.fs.feed_to_ro = Arc(source=m.fs.feed.outlet, destination=m.fs.RO.inlet)
+
+    TransformationFactory("network.expand_arcs").apply_to(m)
```
kurbansitterley, line 183: I suppose this isn't needed anymore

kurbansitterley:
> > > > Took a quick first pass at the main utility function. I think this is definitely the type of thing we need to be adding to WaterTAP, and this sort of method in particular has been a long time coming.
> > > > I have some initial comments, but the main question I have is--would it be better to add this as a method to the RO unit? What would be the pros and cons? Without having thought it through thoroughly, my initial reaction is that this would be better as a method on the RO class.
> > > 
> > > 
> > > I removed the feed as per the suggestion.
> > > Adding it as an explicit method would add unnecessary support code to RO that will need to make a lot of assumptions. I am not really sure what is the best way to go.
> > > I really think we want to first figure out a generic way to store parameters/specs for unit models so we can create a way to generate standard output files and have unit models injest them to fix the operation.
> > > It really might be better to create a separate tool that handles model data management
> > > 
> > > 1. A tool that defines a standard method for storing default unit specs
> > > 2. This same tool can then be used to fit data for a given unit model using similar spec files - e.g. a file that defines known parameters and defines what variables to fit.
> > > 
> > > There are a few complications -> not all spec sheets are the same or data is the same, some users will need to modify the fitting scripts, some users might have strong opinions on what parameters to fit - > for example, in RO for pressure drop we can fit spare porosity or friction factor correlation values etc.
> > > I really think, we first should build a number of such "fitting" scripts for different unit models, and then see if its possible to create generalized methods and what the requirements would be, as creating a new set of tools or adding more functionality to RO/etc. will potentially drive us into a corner. Although creating a data storage method might be the first step.
> > 
> > 
> > I think I could go either way on making it a class method or standalone - but what are thoughts on putting it in a more general location like `watertap/core/utils` or just straight up `watertap/tools` (which actually has no useful tools to anyone without an OLI license atm).
> 
> @kurbansitterley Obviously a general tool would live there, but were you suggesting in the interim moving this code (ro_fitter) to tools? Watertap/tools make sense, but i do not think core/utils is a good place.
> 
> @adam-a-a I was not planning to store fitted data and let users generate the data by running this code, but I don't see why not. I will update it.

Yes my suggestion was more just to move it to somewhere else rather than having it in flowsheets. `watertap/tools` seems like a natural place, maybe a new sub directory called `unit_models` or something. 

I am curious if you see a tool like this being part of a future `WaterTapFlowsheetBlock` RO model. Or more generally - having these  parameter estimation routines for any unit model as part of that framework.

avdudchenko:
> > > > > Took a quick first pass at the main utility function. I think this is definitely the type of thing we need to be adding to WaterTAP, and this sort of method in particular has been a long time coming.
> > > > > I have some initial comments, but the main question I have is--would it be better to add this as a method to the RO unit? What would be the pros and cons? Without having thought it through thoroughly, my initial reaction is that this would be better as a method on the RO class.
> > > > 
> > > > 
> > > > I removed the feed as per the suggestion.
> > > > Adding it as an explicit method would add unnecessary support code to RO that will need to make a lot of assumptions. I am not really sure what is the best way to go.
> > > > I really think we want to first figure out a generic way to store parameters/specs for unit models so we can create a way to generate standard output files and have unit models injest them to fix the operation.
> > > > It really might be better to create a separate tool that handles model data management
> > > > 
> > > > 1. A tool that defines a standard method for storing default unit specs
> > > > 2. This same tool can then be used to fit data for a given unit model using similar spec files - e.g. a file that defines known parameters and defines what variables to fit.
> > > > 
> > > > There are a few complications -> not all spec sheets are the same or data is the same, some users will need to modify the fitting scripts, some users might have strong opinions on what parameters to fit - > for example, in RO for pressure drop we can fit spare porosity or friction factor correlation values etc.
> > > > I really think, we first should build a number of such "fitting" scripts for different unit models, and then see if its possible to create generalized methods and what the requirements would be, as creating a new set of tools or adding more functionality to RO/etc. will potentially drive us into a corner. Although creating a data storage method might be the first step.
> > > 
> > > 
> > > I think I could go either way on making it a class method or standalone - but what are thoughts on putting it in a more general location like `watertap/core/utils` or just straight up `watertap/tools` (which actually has no useful tools to anyone without an OLI license atm).
> > 
> > 
> > @kurbansitterley Obviously a general tool would live there, but were you suggesting in the interim moving this code (ro_fitter) to tools? Watertap/tools make sense, but i do not think core/utils is a good place.
> > @adam-a-a I was not planning to store fitted data and let users generate the data by running this code, but I don't see why not. I will update it.
> 
> Yes my suggestion was more just to move it to somewhere else rather than having it in flowsheets. `watertap/tools` seems like a natural place, maybe a new sub directory called `unit_models` or something.
> 
> I am curious if you see a tool like this being part of a future `WaterTapFlowsheetBlock` RO model. Or more generally - having these parameter estimation routines for any unit model as part of that framework.

@kurbansitterley we are on same page with tool location. For the general tooling it would be a separate tools for wrapping around unit models and configuring them for fitting data I would envision - and then generating standardized data files that could be loaded into the model.  The data files could be yaml/json format, or what ever else. 

I think we can discuss the tool and how its structured in a different place as there are many ways to do it and list of cons/pros for each approach can get a tiny bit long....

kurbansitterley:
@avdudchenko see #1699 for possible location for this (and other) tools

avdudchenko:
This is stale now.

PR #1691, opened 2025-11-20
Make MCAS compatible with MVC

adam-a-a:
## Summary/Motivation:
Enabling compatibility between MCAS and MVC would eliminate need for translators and could potentially facilitate saturation index predictions later on.

- first get MCAS with only TDS as solute to work with MVC flowsheet
- [TODO] test with more than one solute
- [TODO] add tests

### Legal Acknowledgement


watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -1376,15 +1376,15 @@ def build(self):
...
+            bounds=(273.15, 1000),
             domain=NonNegativeReals,
             units=pyunits.K,
             doc="State temperature",
         )
 
         self.pressure = Var(
             initialize=101325,
-            bounds=(1e5, None),
+            bounds=(1e3, 50000000),
```
adam-a-a, line 1387: most important change- reducing the lower bound on pressure in MCAS! Otherwise, the evaporator cannot initialize and will remain infeasible, due to the equality constraints equating brine pressure and vapor pressure to saturation pressure.

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -2545,7 +2553,7 @@ def default_material_balance_type(self):
         return MaterialBalanceType.componentTotal
 
     def default_energy_balance_type(self):
-        return EnergyBalanceType.none
+        return EnergyBalanceType.enthalpyTotal
```
adam-a-a, line 2556: another critical change for compatibility with heatexchanger. Potentially can set the energybalancetype directly in HX, but chose to update MCAS.

watertap/flowsheets/mvc/mvc_single_stage_mcas.py
```diff
@@ -0,0 +1,1089 @@
...
+    m.fs.costing.add_specific_energy_consumption(m.fs.distillate.properties[0].flow_vol)
+    m.fs.costing.base_currency = pyo.units.USD_2020
+
+
+def set_electroneutral_ion_fractions(m, tds=0.1):
+    """
+    Set ion mass fractions to achieve target TDS and electroneutrality for any ionic composition.
+
+    Algorithm:
+      1. Separate ions into cations (charge > 0) and anions (charge < 0)
```
adam-a-a, line 410: MCAS already does this separation.

watertap/flowsheets/mvc/mvc_single_stage_mcas.py
```diff
@@ -0,0 +1,1089 @@
...
+      2. Distribute TDS equally among all ions initially
```
adam-a-a, line 411: MCAS also has a total_dissolved_solids property method that isn't error-proof (i.e., there are some known issues), but it should calculate TDS based on ions in the set.

watertap/flowsheets/mvc/mvc_single_stage_mcas.py
```diff
@@ -0,0 +1,1089 @@
...
+    )
+
+    m.fs.costing.cost_process()
+    m.fs.costing.add_annual_water_production(m.fs.distillate.properties[0].flow_vol)
+    m.fs.costing.add_LCOW(m.fs.distillate.properties[0].flow_vol)
+    m.fs.costing.add_specific_energy_consumption(m.fs.distillate.properties[0].flow_vol)
+    m.fs.costing.base_currency = pyo.units.USD_2020
+
+
+def set_electroneutral_ion_fractions(m, tds=0.1):
```
adam-a-a, line 405: @tristantc unsure why we need this. It could be that there is an exceptional case here. If not, I would think that MCAS can already handle this itself without the need for an additional helper function such as this

tristantc: Thanks for bringing that up @adam-a-a! I'm already aware of the `total_dissolved_solids` [mg/L] variable and `assert_electroneutrality` method used to figure out ion concentrations from a given TDS in MCAS. I just included the helper function to ensure the feed TDS is 0.1 kg/kg and electroneutral for specific ions, but it's not actually required; I only added it to help reproduce the results. Anyway, I'm happy to try those out instead.

watertap/property_models/multicomp_aq_sol_prop_pack.py
avdudchenko: I belive this is expected behavior - e.g. the property package informs the model that it can do the energy balance, rather than we force the unit model to use an unsported type by accident.

watertap/property_models/multicomp_aq_sol_prop_pack.py
avdudchenko: Is there a reason to not set lower bound to ~1 Pa, 1000 seems arbitary - I am not 100% sure its possible for MVC to go lower -or where we get there but I do see pressures drop to below 10,000 Pa with crystalizer for example. This seems abit too close to home.

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -2530,9 +2538,9 @@ def get_material_flow_terms(self, p, j):
             return self.flow_mass_phase_comp[p, j]
 
     # TODO: add enthalpy terms later
-    # def get_enthalpy_flow_terms(self, p):
-    #     """Create enthalpy flow terms."""
-    #     return self.enth_flow
+    def get_enthalpy_flow_terms(self, p):
```
avdudchenko, line 2541: I would think its better to enable this if user specifies an option for type of enthalpy to return - e.g. seawater, nacl - I am nto sure what equation we actually use here - same as is done for osmotic pressure?

watertap/flowsheets/mvc/mvc_single_stage_mcas.py
avdudchenko, line 1: Why are we creating a whole new model, that is basically identical to the standard mvc model except for the property package. This seems to add a-lot of extra code to maintain. I believe we should simply add an option to standard MVC model to either build with Seawater/NaCl prop package or MCAS and feed composition and test it. We should be able to add options through out the flowsheet to catch ions/tds type reporting based on the property package. - or at least separate out functions/code that is property package specific.  - this is sort of done in standard MVC model -but it uses just TDS to imitated Seawater prop package. Instead it should accept an potential composition and use that instead.

watertap/flowsheets/mvc/mvc_single_stage_mcas.py
avdudchenko: This should really be part of "specifying the feed' if we want to be complete.

avdudchenko: Also - this looks to be a non-standard approach - you are basically force re balancing ratio of anions to cations. This is never done in charge neutarlity or feed water reconciliation. We always charge balance on specified ion - and as @adam-a-a  mentions MCAS I belive has a built in method - it might be embded in NF-DSPM-DE model - and in that case should be moved to MCAS.

watertap/flowsheets/mvc/mvc_single_stage.py
```diff
@@ -67,31 +76,47 @@ def main():
...
     # flowsheet set up
     m = ConcreteModel()
     m.fs = FlowsheetBlock(dynamic=False)
 
     # Properties
-    m.fs.properties_feed = props_sw.SeawaterParameterBlock()
+    if feed_properties is None or feed_properties == "seawater":
+        m.fs.properties_feed = props_sw.SeawaterParameterBlock()
+    elif feed_properties == "MCAS":
+        m.fs.properties_feed = MCASParameterBlock(
```
avdudchenko, line 108: We should simply add an option to supply a dict with ion composition and data to configure MCAS if its being used - so all of this could could be reused for multi-comp feeds as well. and update rest of standard single stage mvc to handle single species or multi-specie feeds based on property package - that should not be that difficult to do.

avdudchenko (commented):
Overall seems like on the right track - we should really avoid creating copy pasted code. Instead of creating a nearly identical MVC model, we should modify the existing MVC model to handle single and multi-component feeds.  

We also should add tests to verify that enthalpy is available and works correctly in MCAS tests.

watertap/flowsheets/mvc/mvc_single_stage_mcas.py
adam-a-a: You're right. This was never meant to be merged in and should be removed. Instead, this was merely draft code for exploratory purposes as we collaborated on the PR. If this PR moves forward, this file will be removed, and we'd handle by making the existing MVC flowsheet work with different prop models rather than being hard-coded to one.

ksbeattie:
@adam-a-a PrOMMiS does need this, not urgently as they have a work-around, but they do want this.  Can you, or someone you designate, get to this before June?

sufikaur:
This work was separated into multiple PRs
#1771 #1765 #1767

PR #1699, opened 2025-11-24
Move `calculate_operating_pressure` to utils and add docs

kurbansitterley:
## Fixes/Resolves:

Moves `calculate_operating_pressure` helper function from the RO w ERD flowsheet to a new file for unit model utility functions.

## Summary/Motivation:

The `calculate_operating_pressure` helper function in the RO w ERD flowsheet is useful outside of that flowsheet. This PR moves that function to the newly created `watertap/core/utils/unit_models.py` where we could put this and other useful tools for specific unit models.

`calculate_operating_pressure` was modified from the current form to be able to accept:
- seawater property model state blocks
- NaCl property model state blocks
- NaCl w temp dependence model state blocks
- RO membrane channel 0D and 1D control volumes

It also provides lower and upper bounds on acceptable arguments for `water_recovery_mass` and `salt_passage`. 

- move and genericize `calculate_operating_pressure` helper function from the RO w ERD flowsheet
- modify RO w ERD flowsheet to use this function
- add test for this function
- add documentation for `calculate_operating_pressure` in a new section Technical Reference/Unit Models/Unit Model Utilities

## TO DO:
- [x] documentation

### Legal Acknowledgement


watertap/core/util/unit_models.py
```diff
@@ -0,0 +1,113 @@
+#################################################################################
```
adam-a-a, line 1: One note about location--since I noticed the discussion on #1689: the tools folder originally had parameter sweep, in addition to oli_api and some associated oli utilities, the latter of which were then moved to utilities. 

To me, this fits more closely with what we have under utilities. The only exception is that this a utility in the context of unit models. I suppose this is subjective, but if you are shooting for consistency alongside historical context, there it is.

watertap/tools/unit_models/reverse_osmosis.py
```diff
@@ -0,0 +1,113 @@
...
+
+    if not 1e-3 < water_recovery_mass < 0.999:
+        raise ValueError("water_recovery_mass argument must be between 0.001 and 0.999")
+
+    if isinstance(feed_state_block, NaClStateBlockData) or isinstance(
+        feed_state_block, NaClTDepStateBlockData
+    ):
+        comp = "NaCl"
+    if isinstance(feed_state_block, SeawaterStateBlockData):
+        comp = "TDS"
```
adam-a-a: Why do we need these conditionals? Can't we just point to the component on component_list, as we do in unit models? A little complexity arises with MCAS, which we can add logic for, but can happen in a separate PR. 

Instead, I'd propose a conditional check on the supported stateblockdata classes and raise an exception if it is not one of the anticipated prop model stateblocks.

watertap/core/util/unit_models.py
kurbansitterley: do you mean `watertap.core.util`? I suppose it is subjective and also dependent on what the difference between a "tool" and a "utility" function is. 

To me, `from watertap.tools.unit_models import ____` (or even `from watertap.tools import _____`) is more obvious and accessible. I would moderately fight for this hill but not die on it.

watertap/tools/unit_models/reverse_osmosis.py
```diff
@@ -0,0 +1,113 @@
...
+)
+from watertap.core import MembraneChannel0DBlock, MembraneChannel1DBlock
+from watertap.core.solvers import get_solver
+
+
+__all__ = ["calculate_operating_pressure"]
+
+
+def calculate_operating_pressure(
+    feed_state_block=None,
```
adam-a-a:
```suggestion
    state_block=None,
```

watertap/tools/unit_models/reverse_osmosis.py
```diff
@@ -0,0 +1,113 @@
...
+    over_pressure=0.15,
```
adam-a-a:
```suggestion
    over_pressure_factor=1.15,
```
I think it'd be more intuitive for the user if this were a factor representing what factor to multiply the brine osm press by, rather than the fraction that we have now. 

Then you'd adjust the equation below and eliminate need to add 1 to the factor.

watertap/tools/unit_models/reverse_osmosis.py
```diff
@@ -0,0 +1,113 @@
...
+    water_recovery_mass=0.5,
+    salt_passage=0.01,
```
adam-a-a:
```suggestion
    salt_passage_fraction=0.01
```
Any reason not to have the default as 0? This is more in line with the typical rule of thumb calculation you'd make to estimate brine osmotic pressure.

watertap/core/util/unit_models.py
```diff
@@ -0,0 +1,113 @@
...
+    if isinstance(feed_state_block, SeawaterStateBlockData):
+        comp = "TDS"
+
+    if solver is None:
+        solver = get_solver()
+
+    tmp = ConcreteModel()  # create temporary model
+    prop = feed_state_block.config.parameters
+
+    tmp.feed = prop.build_state_block([0])
```
adam-a-a, line 88: This is nice. I wonder though if you could just clone the stateblock as your tmp model?

```suggestion
    tmp = feed_state_block.clone()
```

watertap/tools/unit_models/reverse_osmosis.py
```diff
@@ -0,0 +1,113 @@
...
+    tmp.feed[0].flow_mass_phase_comp["Liq", "H2O"].fix(
+        value(feed_state_block.flow_mass_phase_comp["Liq", "H2O"])
+        * (1 - water_recovery_mass)
+    )
+    tmp.feed[0].flow_mass_phase_comp["Liq", comp].fix(
+        value(feed_state_block.flow_mass_phase_comp["Liq", comp]) * (1 - salt_passage)
+    )
+    tmp.feed[0].pressure.fix(
+        101325
+    )  # valid when osmotic pressure is independent of hydraulic pressure
```
adam-a-a:
```suggestion
    # valid when osmotic pressure is independent of hydraulic pressure
    tmp.feed[0].pressure.fix(
        101325
    )  
```
so formatting doesn't get screwed up

watertap/core/util/unit_models.py
kurbansitterley: hmm let me try

update: doesn't work.

watertap/tools/unit_models/reverse_osmosis.py
kurbansitterley: updated to 0

watertap/tools/unit_models/reverse_osmosis.py
```diff
@@ -0,0 +1,122 @@
...
+        raise ValueError("water_recovery_mass argument must be between 0.001 and 0.999")
+
+    if not over_pressure_factor >= 1.0:
+        raise ValueError(
+            "over_pressure_factor argument must be greater than or equal to 1.0"
+        )
+
+    print(over_pressure_factor, water_recovery_mass, salt_passage)
+
+    comp = [c for c in state_block.component_list if c != "H2O"][0]
```
adam-a-a:
```suggestion
    comp = state_block.params.solute_set
```

watertap/core/util/unit_models.py
```diff
@@ -0,0 +1,116 @@
...
+
+    if not 1e-3 < water_recovery_mass < 0.999:
+        raise ValueError("water_recovery_mass argument must be between 0.001 and 0.999")
+
+    if not over_pressure_factor >= 1.0:
+        raise ValueError(
+            "over_pressure_factor argument must be greater than or equal to 1.0"
+        )
+
+    comp = state_block.params.solute_set.first()
```
adam-a-a, line 75: Assuming that we are limiting this to single solute prop models right now, I’d recommend just checking the solute_set had a length no greater than 1. If greater than 1, raise exception since not supported yet

kurbansitterley: With the state block type check above, is it possible to get to this point with a state block where `len(solute_set) > 1`?

avdudchenko: I would extend it a bit.
1) Check if there is osmotic_pressure available
2) Check if there is single solute, if there is more then one, check if there is total_disolved_solids -> this will get MCAS to work although, it needs to have "seawater or NACL" properties for osmotic pressure enabled and density I think as well. This should cover most of our prop packs then.

watertap/core/util/unit_models.py
kurbansitterley: moved to `watertap.core.util` but couldn't really decide on where or what to call the file. Right now it is in `util/unit_models.py` but could make an argument that it should be in `misc.py` or even `initialization.py`

sufikaur:
Merge in #1757 first.

docs/technical_reference/unit_models/unit_model_utilities.rst
```diff
@@ -0,0 +1,63 @@
...
+
+This function will estimate the operating pressure of a unit model based on the osmotic pressure of the inlet stream. It can accept the following arguments:
+
+    * ``state_block``: The state block of the RO feed that has the non-pressure state variables set to desired values. Can accept state blocks from NaCl or seawater property models.
+    * ``over_pressure_factor``: The amount of operating pressure above the brine osmotic pressure represented as a fraction (default=1.15)
+    * ``water_recovery_mass``: The mass-based fraction of inlet H2O that becomes permeate (default=0.5)
+    * ``salt_passage``: The mass-based fraction of inlet salt that becomes permeate (default=0)
+    * ``solver``: Solver object to be used (default=None)
+
+An separate model of the provided ``state_block`` is created and the operating pressure is calculated based on the provided ``water_recovery_mass`` and ``salt_passage`` values.
```
MarcusHolly:
```suggestion
A separate model of the provided ``state_block`` is created and the operating pressure is calculated based on the provided ``water_recovery_mass`` and ``salt_passage`` values. 
```

MarcusHolly (approved):
LGTM - just one minor typo

watertap/core/util/tests/test_unit_models.py
```diff
@@ -0,0 +1,220 @@
...
+
+    with pytest.raises(
+        TypeError,
+        match="state_block must be created with SeawaterParameterBlock, NaClParameterBlock, or NaClTDepParameterBlock",
+    ):
+        m = ConcreteModel()
+        m.fs = FlowsheetBlock(dynamic=False)
+        m.fs.properties = MCASParameterBlock(solute_list=["Na_+"])
+        m.fs.feed = Feed(property_package=m.fs.properties)
+        calculate_operating_pressure(state_block=m.fs.feed.properties[0])
```
adam-a-a, line 188: Should we make an issue to think about how to make this applicable to MCAS?

adam-a-a (approved):
LGTM - just one comment on whether to make an issue for considering the utility's usage for MCAS down the line.

PR #1695, opened 2025-11-21
Bring WaterTap in line with changes in IDAES scaling

dallan-keylogic:
## Fixes/Resolves:
Changes in IDAES caused some test failures in WaterTap (mostly involving scaling). This PR fixes those failures.

This requires https://github.com/IDAES/idaes-pse/pull/1711 from idaes-pse.

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #1704, opened 2025-12-02
RO Scaling improvements

dallan-keylogic:
## Summary/Motivation:
Add additional scaling for various unit models, mostly involving RO. Depends on #1695.

Also, the unit test harness is slightly modified. The check for "badly scaled variables" can be bypassed, and an optional check for the model condition number is added instead.

### Legal Acknowledgement

### DIFF
--- a/watertap/core/membrane_channel1d.py
+++ b/watertap/core/membrane_channel1d.py
@@ -242,14 +242,14 @@ class MembraneChannel1DBlockData(MembraneChannelMixin, ControlVolume1DBlockData)
             self.release_state(source_flags, outlvl)
 
     def calculate_scaling_factors(self):
+        if iscale.get_scaling_factor(self.area) is None:
+            iscale.set_scaling_factor(self.area, 100)
+
         super().calculate_scaling_factors()
 
         # setting scaling factors for variables
 
         # will not override if the user provides the scaling factor
-        ## default of 1 set by ControlVolume1D
-        if iscale.get_scaling_factor(self.area) == 1:
-            iscale.set_scaling_factor(self.area, 100)
 
         if hasattr(self, "pressure_change_total"):
             for v in self.pressure_change_total.values():
--- a/watertap/core/membrane_channel_base.py
+++ b/watertap/core/membrane_channel_base.py
@@ -19,6 +19,7 @@ from pyomo.environ import (
     Expression,
     Param,
     NonNegativeReals,
+    value,
     Var,
     units as pyunits,
 )
@@ -996,7 +997,11 @@ class MembraneChannelMixin:
         if hasattr(self, "velocity"):
             for v in self.velocity.values():
                 if iscale.get_scaling_factor(v) is None:
-                    iscale.set_scaling_factor(v, 1)
+                    sf_A = iscale.get_scaling_factor(self.area, default=1)
+                    sf_F_vol = iscale.get_scaling_factor(
+                        self.properties[t, x].flow_vol_phase["Liq"], default=1
+                    )
+                    iscale.set_scaling_factor(v, sf_F_vol / sf_A, 1)
 
         if hasattr(self, "friction_factor_darcy"):
             for v in self.friction_factor_darcy.values():
@@ -1007,6 +1012,71 @@ class MembraneChannelMixin:
             for v in self.cp_modulus.values():
                 if iscale.get_scaling_factor(v) is None:
                     iscale.set_scaling_factor(v, 1)
+
+        # TODO why did this file have zero scaling factors for constraints
+        if hasattr(self, "eq_dh"):
+            sf_dh = iscale.get_scaling_factor(self.dh, default=1)
+            iscale.constraint_scaling_transform(self.eq_dh, sf_dh)
+
+        if hasattr(self, "eq_area"):
+            sf_A = iscale.get_scaling_factor(self.area, default=1)
+            iscale.constraint_scaling_transform(self.eq_area, sf_A)
+
+        if hasattr(self, "eq_velocity"):
+            for (t, x), condata in self.eq_velocity.items():
+                sf_F_vol = iscale.get_scaling_factor(
+                    self.properties[t, x].flow_vol_phase["Liq"], default=1
+                )
+                iscale.constraint_scaling_transform(condata, sf_F_vol)
+
+        if hasattr(self, "eq_equal_flow_vol_interface"):
+            for (t, x), condata in self.eq_equal_flow_vol_interface.items():
+                sf_F_vol = min(
+                    iscale.get_scaling_factor(
+                        self.properties[t, x].flow_vol_phase["Liq"], default=1
+                    ),
+                    iscale.get_scaling_factor(
+                        self.properties_interface[t, x].flow_vol_phase["Liq"], default=1
+                    ),
+                )
+                iscale.constraint_scaling_transform(condata, sf_F_vol)
+
+        if hasattr(self, "eq_equal_pressure_interface"):
+            for (t, x), condata in self.eq_equal_pressure_interface.items():
+                sf_P = min(
+                    iscale.get_scaling_factor(
+                        self.properties_interface[t, x].pressure, default=1
+                    ),
+                    iscale.get_scaling_factor(
+                        self.properties[t, x].pressure, default=1
+                    ),
+                )
+                iscale.constraint_scaling_transform(condata, sf_P)
+
+        if hasattr(self, "eq_K"):
+            for (t, x, j), condata in self.eq_K.items():
+                sf_K = iscale.get_scaling_factor(self.K[t, x, j])
+                # Hopefully h has been calculated before this
+                sf_h = 1 / value(self.dh)
+                iscale.constraint_scaling_transform(condata, sf_K * sf_h)
+
+        if hasattr(self, "eq_N_Re"):
+            for (t, x), condata in self.eq_N_Re.items():
+                sf_Re = iscale.get_scaling_factor(self.N_Re[t, x], default=1)
+                sf_A = iscale.get_scaling_factor(
+                    self.area, default=1 / value(self.area)
+                )
+                sf_visc = iscale.get_scaling_factor(
+                    self.properties[t, x].visc_d_phase["Liq"], default=1
+                )
+                iscale.constraint_scaling_transform(condata, sf_Re * sf_A * sf_visc)
+
+        if hasattr(self, "eq_N_Sc_comp"):
+            for (t, x, j), condata in self.eq_N_Sc_comp.items():
+                sf_visc = iscale.get_scaling_factor(
+                    self.properties[t, x].visc_d_phase["Liq"], default=1
+                )
+                iscale.constraint_scaling_transform(condata, sf_visc)
 
 
 # helper for validating configuration arguments for this CV
--- a/watertap/flowsheets/full_water_resource_recovery_facility/BSM2.py
+++ b/watertap/flowsheets/full_water_resource_recovery_facility/BSM2.py
@@ -1042,7 +1042,9 @@ def initialize_system(m):
     seq.set_guesses_for(m.fs.R1.inlet, tear_guesses1)
     seq.set_guesses_for(m.fs.asm_adm.inlet, tear_guesses2)
 
-    initializer = BlockTriangularizationInitializer()
+    initializer = BlockTriangularizationInitializer(
+        calculate_variable_options={"eps": 2e-8}, skip_final_solve=True
+    )
 
     def function(unit):
         try:
--- a/watertap/flowsheets/full_water_resource_recovery_facility/BSM2_P_extension.py
+++ b/watertap/flowsheets/full_water_resource_recovery_facility/BSM2_P_extension.py
@@ -502,14 +502,20 @@ def scale_system(m, bio_P=False):
     ad_scaler = ADScaler()
     ad_scaler.scale_model(m.fs.AD)
 
-    set_scaling_factor(m.fs.AD.KH_co2, 1e1)
-    set_scaling_factor(m.fs.AD.KH_ch4, 1e1)
-    set_scaling_factor(m.fs.AD.KH_h2, 1e2)
-    set_scaling_factor(m.fs.AD.liquid_phase.heat, 1e1)
+    for vardata in m.fs.AD.KH_co2.values():
+        set_scaling_factor(vardata, 1e1)
+    for vardata in m.fs.AD.KH_ch4.values():
+        set_scaling_factor(vardata, 1e1)
+    for vardata in m.fs.AD.KH_h2.values():
+        set_scaling_factor(vardata, 1e2)
+    for vardata in m.fs.AD.liquid_phase.heat.values():
+        set_scaling_factor(vardata, 1e1)
     if bio_P:
-        set_scaling_factor(m.fs.AD.liquid_phase.reactions[0].S_H, 1e1)
+        for blkdata in m.fs.AD.liquid_phase.reactions.values():
+            set_scaling_factor(blkdata.S_H, 1e1)
     else:
-        set_scaling_factor(m.fs.AD.liquid_phase.reactions[0].S_H, 1e2)
+        for blkdata in m.fs.AD.liquid_phase.reactions.values():
+            set_scaling_factor(blkdata.S_H, 1e2)
 
     cstr_list = [m.fs.R1, m.fs.R2, m.fs.R3, m.fs.R4]
     cstr_scaler = CSTRScaler()
@@ -517,7 +523,8 @@ def scale_system(m, bio_P=False):
         cstr_scaler.scale_model(unit)
 
     for unit in cstr_list:
-        set_scaling_factor(unit.hydraulic_retention_time, 1e-3)
+        for vardata in unit.hydraulic_retention_time.values():
+            set_scaling_factor(vardata, 1e-3)
 
     aeration_list = [m.fs.R5, m.fs.R6, m.fs.R7]
     aeration_scaler = AerationTankScaler()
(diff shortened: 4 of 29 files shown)


watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -1407,8 +1407,8 @@ def _flow_mol_phase_comp(self):
 
             def rule_flow_mol_phase_comp(b, p, j):
                 return (
-                    b.flow_mass_phase_comp[p, j]
-                    == b.flow_mol_phase_comp[p, j] * b.params.mw_comp[j]
+                    b.flow_mass_phase_comp[p, j] / b.params.mw_comp[j]
+                    == b.flow_mol_phase_comp[p, j]
```
dallan-keylogic, line 1411: This is changed due to how the WaterTap scaling utilities work. Because the constraint is named `eq_flow_mol_phase_comp`, it is automatically scaled using the scaling factor for `flow_mol_phase_comp`. However, as originally written, it ought to be scaled instead like `flow_mass_phase_comp`. This slight modification of the constraint is easier than messing with that workflow.

watertap/core/membrane_channel1d.py
```diff
@@ -242,14 +242,14 @@ def initialize(
             self.release_state(source_flags, outlvl)
 
     def calculate_scaling_factors(self):
+        if iscale.get_scaling_factor(self.area) is None:
+            iscale.set_scaling_factor(self.area, 100)
```
adam-a-a, line 246: Should we also display a warning to the user that we are doing this since so scaling factor was provided?

dallan-keylogic: Do the old scaling tools really need additional warnings to spam the user's logs?

In the new scaling tools, I'm going to raise an exception instead, so they can no longer ignore it.

watertap/core/membrane_channel_base.py
```diff
@@ -1008,6 +1013,71 @@ def calculate_scaling_factors(self):
                 if iscale.get_scaling_factor(v) is None:
                     iscale.set_scaling_factor(v, 1)
 
+        # TODO why did this file have zero scaling factors for constraints
```
adam-a-a, line 1016: @bknueven can probably give a more informative response, but the line of thinking was that scaling all variables should suffice in most cases without the need to scale constraints. However, we know that this isn't always the case, as we've experienced particularly when dealing with getting "complex" models to converge.

watertap/flowsheets/full_water_resource_recovery_facility/test/test_BSM2_P_extension.py
```diff
@@ -75,17 +75,9 @@ def test_structural_issues(self, system_frame):
     @pytest.mark.component
     def test_numerical_issues(self, system_frame):
         dt = DiagnosticsToolbox(system_frame)
-        warnings, next_steps = dt._collect_numerical_warnings()
-        assert len(warnings) == 3
+        warnings, _ = dt._collect_numerical_warnings()
+        assert len(warnings) == 1
```
adam-a-a, line 79: nice

watertap/flowsheets/full_water_resource_recovery_facility/BSM2.py
```diff
@@ -1042,7 +1042,9 @@ def initialize_system(m):
     seq.set_guesses_for(m.fs.R1.inlet, tear_guesses1)
     seq.set_guesses_for(m.fs.asm_adm.inlet, tear_guesses2)
 
-    initializer = BlockTriangularizationInitializer()
+    initializer = BlockTriangularizationInitializer(
+        calculate_variable_options={"eps": 2e-8}, skip_final_solve=True
```
adam-a-a, line 1046: how is `eps `applied through `calculate_variable_options`?

watertap/flowsheets/full_water_resource_recovery_facility/BSM2_P_extension.py
```diff
@@ -502,52 +502,69 @@ def scale_system(m, bio_P=False):
...
-                0, "Liq", "H2O"
-            ],
-            1e-6,
-        )
+        for t in m.fs.time:
+            set_scaling_factor(
+                m.fs.R5.control_volume.rate_reaction_generation[t, "Liq", "S_I"], 1e-3
+            )
+            # TODO is it intentional that we're setting scaling factors for some constraints
+            # but doing constraint scaling transforms for others?
```
adam-a-a, line 561: Not intentional--traditionally we applied constraint scaling transforms.

watertap/property_models/multicomp_aq_sol_prop_pack.py
adam-a-a, line 1411: Why opt for this form?

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
```diff
@@ -3095,19 +3098,19 @@ def test_scaler(self):
...
-        ] == pytest.approx(19320000, rel=1e-3)
+        ] == pytest.approx(8.617e07, rel=1e-3)
         assert m.fs.stream[0].scaling_factor[
             m.fs.stream[0].eq_elec_mobility_phase_comp["Liq", "Cl_-"]
-        ] == pytest.approx(12660000, rel=1e-3)
+        ] == pytest.approx(8.617e07, rel=1e-3)
         assert m.fs.stream[0].scaling_factor[
             m.fs.stream[0].eq_elec_mobility_phase_comp["Liq", "Mg_2+"]
-        ] == pytest.approx(18200000, rel=1e-3)
+        ] == pytest.approx(4.309e07, rel=1e-3)
```
adam-a-a, line 3113: Unclear why we are getting relatively significant changes in these test values

watertap/unit_models/tests/test_nanofiltration_0D.py
```diff
@@ -1143,13 +1143,13 @@ def test_scaling(self, mcas_case):
...
-        ] == pytest.approx(0.0000025, rel=1e-3)
+        ] == pytest.approx(1e-5, rel=1e-3)
         assert mcas_case.fs.unit.scaling_factor[
             mcas_case.fs.unit.retentate_temperature_equality[0.0]
-        ] == pytest.approx(0.003354, rel=1e-3)
+        ] == pytest.approx(1e-2, rel=1e-3)
         assert mcas_case.fs.unit.scaling_factor[
             mcas_case.fs.unit.permeate_temperature_equality[0.0]
-        ] == pytest.approx(0.003354, rel=1e-3)
+        ] == pytest.approx(1e-2, rel=1e-3)
```
adam-a-a, line 1152: These changes seem significant

watertap/unit_models/tests/test_reverse_osmosis_1D.py
```diff
@@ -154,6 +154,9 @@ def configure(self):
             },
         }
 
+        self.skip_badly_scaled_vars = True
+        self.condition_number = 3.9140627e8  # Was previously 4.8157167e15
```
adam-a-a, line 158: these seem like great improvements!

watertap/flowsheets/full_water_resource_recovery_facility/BSM2.py
dallan-keylogic: It is the tolerance used to check whether the constraint residual is small enough to consider the constraint "satisfied". When solving the flowsheet, I encountered a problem where it couldn't reduce the residual of some enthalpy constraint (i.e., a constraint that needs a scaling factor <<1)  to less than 1e-8. Due to Pyomo/pyomo#3785 , it does not take the constraint's scaling factor into account when calculating this residual. Using a `constraint_scaling_transformation` would allow scaling to be taken into account, but we wanted to move away from that in the new scaling tools.

watertap/property_models/multicomp_aq_sol_prop_pack.py
dallan-keylogic: This is changed due to how the WaterTap scaling utilities work. Because the constraint is named eq_flow_mol_phase_comp, it is automatically scaled using the scaling factor for flow_mol_phase_comp. However, as originally written, it ought to be scaled instead like flow_mass_phase_comp. This slight modification of the constraint is easier than messing with that workflow.

watertap/unit_models/tests/test_nanofiltration_0D.py
dallan-keylogic: These changes weren't a result of this PR, but instead #1695. They're due to a change in how the nominal value expression walker works. Previously it used `1/value(var)` if `var` was fixed, but now it uses `1/scaling_factor[var]` regardless of whether or not `var` is fixed.

watertap/flowsheets/full_water_resource_recovery_facility/BSM2_P_extension.py
kurbansitterley: this appears to contradict @MarcusHolly in #1695

watertap/core/membrane_channel1d.py
adam-a-a: @dallan-keylogic in the new scaling tools, could you think about perhaps a setting where the user could choose to have exceptions raised _or_ bypass (with warnings if you prefer)? The default would be to raise exceptions, but it'd be nice to have an escape route for devs (for whatever exploratory reasons someone might have).

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
adam-a-a: It's clearer now.

watertap/core/membrane_channel1d.py
avdudchenko: The default scale should be 1/100 - typical RO area is 1-100 m2 in our cases since our feed is about 1 kg/s. actual a scaling factor of 1 or 0.1 works as well.

watertap/unit_models/reverse_osmosis_0D.py
```diff
@@ -189,3 +189,15 @@ def calculate_scaling_factors(self):
         if hasattr(self, "width"):
             if iscale.get_scaling_factor(self.width) is None:
                 iscale.set_scaling_factor(self.width, 1)
+
+        for (t, x, j), condata in self.eq_mass_frac_permeate.items():
+            sf = iscale.get_scaling_factor(
+                self.flux_mass_phase_comp[t, x, "Liq", j], default=1
```
avdudchenko, line 195: This might cause more issues since as the fluxes can get very small for salt. I don't remember what exact scaling factor but it should related to actual component scaling factor otherwise this will cause issues.

watertap/unit_models/reverse_osmosis_0D.py
```diff
@@ -189,3 +189,15 @@ def calculate_scaling_factors(self):
...
+            )
+            iscale.constraint_scaling_transform(condata, sf)
+
+        for (t, p, j), condata in self.eq_permeate_production.items():
+            sf = iscale.get_scaling_factor(
+                self.mixed_permeate[t].get_material_flow_terms(p, j), 1
```
avdudchenko, line 201: Same here. Not sure useing default value of 1 is safe.

watertap/unit_models/reverse_osmosis_1D.py
```diff
@@ -242,3 +242,59 @@ def calculate_scaling_factors(self):
...
                     iscale.set_scaling_factor(v, 1e-4)
+
+        if hasattr(self, "eq_permeate_outlet_isobaric"):
+            for (t, x), condata in self.eq_permeate_outlet_isobaric.items():
+                sf_P = min(
+                    iscale.get_scaling_factor(
+                        self.permeate_side[t, x].pressure, default=1
+                    ),
+                    iscale.get_scaling_factor(
+                        self.mixed_permeate[t].pressure, default=1
```
avdudchenko, line 253: should we not default to 1e-5 since rpessure is always in Pa?

watertap/unit_models/reverse_osmosis_1D.py
```diff
@@ -242,3 +242,59 @@ def calculate_scaling_factors(self):
...
+        for (t, p, j), condata in self.eq_permeate_production.items():
+            sf = iscale.get_scaling_factor(
+                self.mixed_permeate[t].get_material_flow_terms(p, j), default=1
+            )
+            iscale.constraint_scaling_transform(condata, sf)
+
+        for (t, x, p, j), condata in self.eq_mass_transfer_term.items():
+            sf = min(
+                iscale.get_scaling_factor(
+                    self.mass_transfer_phase_comp[t, x, p, j], default=1
```
avdudchenko, line 275: Here is well, settting these to 1 might be problematic as these terms can be very small.

avdudchenko (commented):
We should be carefuly setting constraint scaling factors to 1 for many of the RO expressions. Some of the flux and mass transfer terms can be 1e-6 or so depending on system being solved. I almost rather leave these to be auto scaled  by Watertap solver rather then assume their scale and let user refine these

watertap/core/membrane_channel1d.py
dallan-keylogic: This isn't the membrane mass transfer surface area---it's the cross-sectional area for fluid flow.

watertap/unit_models/reverse_osmosis_0D.py
dallan-keylogic: This change is designed to take advantage of a scaling factor for `flux_mass_phase_comp` if the user has set it but revert to the previous default behavior if it hasn't been set.  Using `constraint_scaling_transform` with a default value of `1` is usually tantamount to not transforming the variable in the first place.  (The exception would be if `get_constraint_transform_applied_scaling_factor` is used to check for previous scaling. However, in `watertap` that function is used only in tests, and it does not appear in `watertap-solvers` at all.)

watertap/unit_models/reverse_osmosis_0D.py
dallan-keylogic: See [this comment.](https://github.com/watertap-org/watertap/pull/1704/changes/BASE..725d09e060b858a1e0c8555169590c8a42551e44#r2975260841)

watertap/unit_models/reverse_osmosis_1D.py
dallan-keylogic: If the user hasn't applied a scaling factor for pressure, then using a default value of `1e-5` can result in bad conditioning (although IPOPT's gradient-based scaling would probably take care of it). All WaterTAP property packages should also have default scaling factors for pressure set, while a non-WaterTAP property package might do something silly like using PSI for pressure.

watertap/unit_models/reverse_osmosis_1D.py
dallan-keylogic: See [this comment.](https://github.com/watertap-org/watertap/pull/1704/changes/BASE..725d09e060b858a1e0c8555169590c8a42551e44#r2975260841)

dallan-keylogic:
> Member



> We should be carefuly setting constraint scaling factors to 1 for many of the RO expressions. Some of the flux and mass transfer terms can be 1e-6 or so depending on system being solved. I almost rather leave these to be auto scaled by Watertap solver rather then assume their scale and let user refine these

The WaterTAP solver does not look for scaling factors applied using `constraint_scaling_transform` when performing autoscaling. It behaves in the same fashion whether the constraint was transformed after the fact or whether a scaling factor was hardcoded in the constraint body. Therefore, as far as the WaterTAP solver is concerned, performing a `constraint_scaling_transform` with a scaling factor of `1` is equivalent to not applying a scaling factor at all.

watertap/core/membrane_channel1d.py
avdudchenko: Oh, that naming convention is confusing. Then the 100 scaling factor is correct. Ignore me.

watertap/unit_models/reverse_osmosis_0D.py
avdudchenko: I actually misunderstood how water tap solver was checking scaling, not sure if we used to check for application of scaling transform. But you are correct. Right now it does not check for application of scaling transform, only scaling factors. So this does not impact the scaling, which was my concern.

watertap/unit_models/reverse_osmosis_1D.py
avdudchenko: Fair enough - I was under impression that in IDAES we always use Pa as our base unit, but thats fine.

avdudchenko:
> > Member
> 
> > We should be carefuly setting constraint scaling factors to 1 for many of the RO expressions. Some of the flux and mass transfer terms can be 1e-6 or so depending on system being solved. I almost rather leave these to be auto scaled by Watertap solver rather then assume their scale and let user refine these
> 
> The WaterTAP solver does not look for scaling factors applied using `constraint_scaling_transform` when performing autoscaling. It behaves in the same fashion whether the constraint was transformed after the fact or whether a scaling factor was hardcoded in the constraint body. Therefore, as far as the WaterTAP solver is concerned, performing a `constraint_scaling_transform` with a scaling factor of `1` is equivalent to not applying a scaling factor at all.

This is correct, I was mistaken when reading the solver implementation [here ](https://github.com/watertap-org/watertap-solvers/blob/ff32328bc9ba8f65108924ce7df63066827cf5c8/src/watertap_solvers/_base.py#L365) and thought it would account for scaling transforms, but it ignores constraint scaling transforms being applied. (which I thought we did, either I was mistaken or  we changed the behavior at one point.. maybe @bknueven can chime in here.).  This brings a good question to consider moving forward into ipoptv2 era, where we use scaling factors - the use of scaling factors on constraints will cause our automatic Jacobin scaling to stop applying automatic scaling factors (which might explain the inconsistencies i observed when using Ipoptv2, although I have no idea where in water-tap we actually apply scaling factors to constraints.. This is unrelated to this PR and is a problem for another day)

PR #1707, opened 2025-12-05
Updating copyright header for 2026

ksbeattie:
## Fixes/Resolves:

The copyright header having the wrong year in all the files

## Summary/Motivation:

We should have the correct year in the copyright header of each file

- Pretty much every .py file
-

### Legal Acknowledgement

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,5 +1,5 @@
 #################################################################################
-# WaterTAP Copyright (c) 2020-2024, The Regents of the University of California,
+# WaterTAP Copyright (c) 2020-2026, The Regents of the University of California,
 # through Lawrence Berkeley National Laboratory, Oak Ridge National Laboratory,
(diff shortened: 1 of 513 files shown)


MarcusHolly (approved):

johnson12742 (approved):

bknueven:
We should do this, but we should also have on our radar updating the name of a particular laboratory, or just changing it to `Alliance for Sustainable Energy, LLC`, which is what is preferred anyhow.

sufikaur (approved):

johnson12742 (approved):

ksbeattie:
> We should do this, but we should also have on our radar updating the name of a particular laboratory, or just changing it to `Alliance for Sustainable Energy, LLC`, which is what is preferred anyhow.

I agree but I fear a change like that will need to involve an update to the IPMP.

PR #1710, opened 2025-12-08
Enhance GUI Output Descriptions

MarcusHolly:
## Summary/Motivation:
When you try to sweep variables on the GUI, it can be hard to decipher what is meant by "Volumetric flow rate", so this PR now specifies "Feed volumetric flow rate" in these scenarios.

- Adds better descriptions for outputs in the GUI

### Legal Acknowledgement


watertap/flowsheets/dye_desalination/dye_desalination_ui.py
```diff
@@ -141,7 +130,7 @@ def export_variables(flowsheet=None, exports=None, build_options=None, **kwargs)
     )
     exports.add(
         obj=fs.dye_separation.nanofiltration.water_permeability_coefficient[0],
-        name="NF Water permeability Coefficient, A",
+        name="NF water permeability coefficient, A",
```
kurbansitterley: Should we be consistent in the UI for variable naming? i.e., here we have "NF water permeability coefficient, A" but in the RO w ERD UI export we have "RO water permeability coefficient"

I guess I am saying consider dropping the `, A` (or adding it above). Same for other variable names.

watertap/flowsheets/dye_desalination/dye_desalination_ui.py
```diff
@@ -423,7 +412,7 @@ def export_variables(flowsheet=None, exports=None, build_options=None, **kwargs)
     )
     exports.add(
         obj=fs.concentrated_dye.properties[0].conc_mass_comp["tds"],
-        name="NF retentate concentration, tds",
+        name="NF retentate concentration of tds",
```
kurbansitterley: If this is just for labeling, I think it is ok to use `TDS` rather than `tds`

kurbansitterley (changes requested):
Looking at some of the existing labels, some of these changes were _badly_ needed. 

I made two minor suggestions.

watertap/flowsheets/dye_desalination/dye_desalination_ui.py
MarcusHolly: I addressed this in the cases I found, but lmk if I missed anything

kurbansitterley (approved):

PR #1716, opened 2025-12-18
Update Costing Documentation

MarcusHolly:
## Summary/Motivation:
The cost methods are missing from most of the documentation

- Adds costing method to documentation
- Modifies names of costing variables that are too ambiguous

### Legal Acknowledgement


watertap/costing/unit_models/mixer.py
```diff
@@ -133,7 +133,7 @@ def cost_naocl_mixer(blk, dosing_rate):
 
 
 def build_caoh2_cost_param_block(blk):
-    blk.cost = pyo.Param(
+    blk.CaOH2_cost = pyo.Param(
```
adam-a-a: Curious what inspired these changes to make `cost` more specific

MarcusHolly: Mainly just personal preference - I thought it was a bit too ambiguous. There also wasn't any consistency as to what `cost` meant. In some scenarios, it referred to unit model costs and in others it referred to chemical costs. Evidently, this name change has resulted in some test failures that I need to fix. Now, unit costs should be consistently called `unit_cost` and chemical costs called `chemical_cost`

MarcusHolly: In reality, it probably wasn't too problematic since I believe it will usually be paired with the corresponding docstring

watertap/flowsheets/oaro/oaro_multi.py
```diff
@@ -251,7 +251,7 @@ def build(number_of_stages, erd_type=ERDtype.pump_as_turbine):
     m.fs.costing.reverse_osmosis.factor_membrane_replacement.fix(0.15)
     m.fs.costing.reverse_osmosis.membrane_cost.fix(30)
     m.fs.costing.reverse_osmosis.high_pressure_membrane_cost.fix(50)
-    m.fs.costing.high_pressure_pump.cost.fix(53 / 1e5 * 3600)
+    m.fs.costing.high_pressure_pump.unit_cost.fix(53 / 1e5 * 3600)
```
MarcusHolly, line 254: @adam-a-a The best argument for updating the `cost` names is to make code more intuitive when flowsheets are updating the values of these variables.

watertap/costing/unit_models/mixer.py
adam-a-a: @MarcusHolly I wonder if we should just use `unit_cost` across the board. For chemicals,  $ / kg cost is technically a unit cost. Any cost per unit of quantity can be rendered a unit_cost. For pumps, unit_cost would be something like $/W, and so on.

watertap/flowsheets/oaro/oaro_multi.py
adam-a-a: I propose `unit_cost` across the board so that is clear that the cost is in $ per some quantity, as opposed to some absolute cost.

watertap/costing/unit_models/mixer.py
MarcusHolly: I agree with this - thanks for the suggestion

docs/technical_reference/costing/membrane_distillation.rst
```diff
@@ -10,7 +10,7 @@ The following parameters are constructed for the unit on the FlowsheetCostingBlo
    :header: "Description", "Symbol", "Parameter Name", "Default Value", "Units"
 
    "Membrane replacement factor", ":math:`f_{replace}`", "``factor_membrane_replacement``", "0.2", ":math:`\text{yr}^{-1}`"
-   "Membrane cost", ":math:`C_{mem}`", "``membrane_unit_cost``", "56", ":math:`\text{USD}_{2018}\text{/m}^2`"
+   "Membrane cost", ":math:`C_{mem}`", "``membrane_cost``", "56", ":math:`\text{USD}_{2018}\text{/m}^2`"
```
MarcusHolly, line 13: Should instances of membrane_cost also be replaced with unit_cost?

docs/technical_reference/costing/mixer.rst
```diff
@@ -9,17 +9,17 @@ The following parameters are constructed for the unit on the FlowsheetCostingBlo
...
+   "Mixer unit cost", ":math:`C_{mix, NaOCl}`", "``unit_cost``", "5.08", ":math:`\text{USD}_{2018}\text{/m}^{3}\text{/day}`"
+   "NaOCl cost", ":math:`C_{NaOCl}`", "``NaOCl_cost``", "0.23", ":math:`\text{USD}_{2018}\text{/kg}`"
    "NaOCl purity", ":math:`p_{NaOCl}`", "``purity``", "0.15", ":math:`\text{dimensionless}`"
 
-   "**CaOH2 mixer**"
-   "Mixer unit cost", ":math:`C_{mix, CaOH2}`", "``cost``", "873.911", ":math:`\text{USD}_{2018}\text{/kg/day}`"
-   "CaOH2 cost", ":math:`C_{CaOH2}`", "``unit_cost``", "0.12", ":math:`\text{USD}_{2018}\text{/kg}`"
+   "**CaOH2 mixer** (cost method = `cost_caoh2_mixer`)"
+   "Mixer unit cost", ":math:`C_{mix, CaOH2}`", "``unit_cost``", "873.911", ":math:`\text{USD}_{2018}\text{/kg/day}`"
+   "CaOH2 cost", ":math:`C_{CaOH2}`", "``CaOH2_cost``", "0.12", ":math:`\text{USD}_{2018}\text{/kg}`"
```
MarcusHolly: How should we handle this? Call the unit model cost and the chemical cost `unit_COST ?`

adam-a-a: Ugh-- another alternative can be `unit_cost_CaOH2` or basically `unit_cost` appended or prepended with the specific item name (e.g., `unit_cost_pump`, `unit_cost_antiscalant`). 

We should also probably check IDAES costing and try to be as consistent as we can across projects. For example, if IDAES already uses some convention for these sorts of costs (and the naming isn't too cryptic), we can use that convention.

MarcusHolly: From a quick skim, I don't think there's an IDAES convention for these sorts of costs.

watertap/costing/unit_models/ion_exchange.py
```diff
@@ -22,7 +22,7 @@
 
 def build_hcl_cost_param_block(blk):
 
-    blk.cost = pyo.Param(
+    blk.unit_cost_HCl = pyo.Param(
```
kurbansitterley: I like modifying cost param names to universally be `unit_cost` (where appropriate) but I am less on board for these changes only because `m.fs.costing.HCl.unit_cost_HCl` seems redundant. I say apply the same `unit_cost` naming convention here.

Or am I missing the reasoning for this deviation?

kurbansitterley (commented):
only requested change would be to apply the `unit_cost` standard to chemical costs as well. 

There are also a few other places where this standard could be applied (just going through the costing files): 
- steam and NaCl recovery cost for Crystallizer
- MgCl and phosphorous costs in electroNP
- steam cost for HX
- heater unit cost for heater_chiller (?) 
- steam cost for steam ejector
- steam cost for surrogate crystallizer

Though you could argue that for some of those, it doesn't quite fit the pattern here (e.g., steam costs are commonly costed without a separate flow parameter block `m.fs.costing.steam` but rather registered directly via instantiation of the costing parameter block for the unit model; see crystallizer costing model for example); those changes might be better suited for a separate PR.

kurbansitterley:
@MarcusHolly did you look to apply this standard to any of the costing for ZO models?

watertap/costing/unit_models/ion_exchange.py
MarcusHolly: Fair point - I will make all of these `unit_cost`

MarcusHolly:
> @MarcusHolly did you look to apply this standard to any of the costing for ZO models?

I don't think the ZO costing needs any changes. The YAMLs look pretty consistent to me in terms of the nomenclature they use

watertap/costing/unit_models/mixer.py
```diff
@@ -87,7 +87,7 @@ def build_naocl_mixer_cost_param_block(blk):
 
 def build_naocl_cost_param_block(blk):
 
-    blk.cost = pyo.Param(
+    blk.unit_cost_NaOCl = pyo.Param(
```
kurbansitterley: should also be `unit_cost` based on convention in other files?

watertap/costing/unit_models/mixer.py
```diff
@@ -133,7 +133,7 @@ def cost_naocl_mixer(blk, dosing_rate):
 
 
 def build_caoh2_cost_param_block(blk):
-    blk.cost = pyo.Param(
+    blk.unit_cost_CaOH2 = pyo.Param(
```
kurbansitterley: should be `unit_cost`? unless I am missing something

docs/technical_reference/costing/membrane_distillation.rst
kurbansitterley: This is marked as resolved (un-resolving just for visibility) and you kept it as membrane_cost but I had the same dilemma when I was making my last review.
I think a reasonable standard for applying the unit_cost convention would be:

- it is a cost per something (flow rate, kg, etc.). As you pointed out though, this could be used for lots of things in our costing pkgs. So the second point is also needed
- the costing parameter block that it is on is exclusively for that item (e.g., `m.fs.costing.nacl` is only for NaCl costs). This is contrasted with membrane costs, which is part of the RO costing block; i.e., if you saw `m.fs.costing.reverse_osmosis.unit_cost `it is not clear what it is a unit cost for.

in short, I think it is preferred to keep it as `membrane_cost` here (and other places) imo

kurbansitterley (changes requested):
two more changes for chemical `unit_cost`

kurbansitterley (approved):

adam-a-a (approved):
LGTM - thanks for your patience

PR #1717, opened 2025-12-31
Scaling, fix minor MCAS issue, and set up Seawater properties for dynamics

dallan-keylogic:
## Summary/Motivation:
As part of working on dynamics in WaterTap, I added material and energy density terms to the Seawater property package and made some improvements to unit model scaling.

Additionally, I found and fixed an issue when the MCAS property package was used 1) in a `Feed` block (and probably also `Product` block) when 2) using a time set with more than one element. The issue seems to have been caused because the state variables (either `flow_mol_phase_comp` or `flow_mass_phase_comp` depending on the configuration) were being created through the on-demand property framework, instead of creating them when initializing the state block. The fact that those state variables were not entirely initialized when the `Feed` block tried to create `References` to them at the unit-model level (moving the time index set from living on the state block to living on the state variable) caused an obscure Pyomo error.

### Legal Acknowledgement


watertap/core/membrane_channel1d.py
```diff
@@ -242,14 +242,14 @@ def initialize(
             self.release_state(source_flags, outlvl)
 
     def calculate_scaling_factors(self):
+        if iscale.get_scaling_factor(self.area) is None:
+            iscale.set_scaling_factor(self.area, 100)
```
MarcusHolly, line 246: Was the change replacing `== 1` with` is None` intentional?

watertap/flowsheets/full_water_resource_recovery_facility/test/test_full_WRRF_with_ASM1_ADM1.py
```diff
@@ -230,7 +230,8 @@ def test_optimization_windows(self, optimized_system_frame):
             # Python 3.9 and 3.10
             cond == pytest.approx(1.95367e11, rel=1e-2)
             # Python 3.11 and 3.12
-            or cond == pytest.approx(3.44132e11, rel=1e-2)
+            # or cond == pytest.approx(3.44132e11, rel=1e-2)
```
MarcusHolly, line 233: Is there a reason to keep this comment around? I think it's helpful for reviewers to see that the number decreased, but other than that, I don't think there's much use for an old condition number

watertap/flowsheets/full_water_resource_recovery_facility/test/test_full_WRRF_with_ASM1_ADM1.py
```diff
@@ -245,5 +246,7 @@ def test_optimization_linux(self, optimized_system_frame):
         # Check condition number to confirm scaling
         jac, _ = get_jacobian(m.rescaled_model, scaled=False)
         assert (jacobian_cond(jac=jac, scaled=False)) == pytest.approx(
-            3.44152e11, rel=1e-3
+            3.44152e11,
+            # 2.71713e11,
```
MarcusHolly, line 250: Same here

watertap/unit_models/tests/test_reverse_osmosis_1D.py
```diff
@@ -154,6 +154,9 @@ def configure(self):
             },
         }
 
+        self.skip_badly_scaled_vars = True
+        self.condition_number = 3.9140627e8  # Was previously 4.8157167e15
```
MarcusHolly, line 158: Same comment here (and below) - consider deleting after getting approvals

MarcusHolly (approved):
Minor comments, but overall LGTM

adam-a-a (approved):

watertap/core/membrane_channel1d.py
dallan-keylogic: Yes. Previously this scaling was happening after the `ControlVolume1D` scaled things, which resulted in a default scaling factor of 1. Now that it's happening before the `ControlVolume1D` scales things, we check to see if the user provided a scaling factor.

adam-a-a: @MarcusHolly I think this because @dallan-keylogic  runs this check _before_  `super().calculate_scaling_factors()`, which avoids the default value of 1 that the ControleVolume would set.

PR #1720, opened 2026-01-08
Update Dye Desal NF Pressure

MarcusHolly:
## Summary/Motivation:
Updates the NF applied pressure based on feedback from our collaborators

- Update applied pressure
- Fix minor typo in testing

### Legal Acknowledgement

### DIFF
--- a/watertap/data/techno_economic/nanofiltration.yaml
+++ b/watertap/data/techno_economic/nanofiltration.yaml
@@ -129,7 +129,7 @@ rHGO_dye_rejection:
       units: dimensionless
       constituent_longform: Total Dissolved Solids (TDS)
   applied_pressure:
-    value: 7   # corresponds to default of 100 psi
+    value: 2   # corresponds to default of 29 psi
     units: bar
   water_permeability_coefficient:
     value: 100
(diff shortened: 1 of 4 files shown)


watertap/data/techno_economic/nanofiltration.yaml
```diff
@@ -129,7 +129,7 @@ rHGO_dye_rejection:
       units: dimensionless
       constituent_longform: Total Dissolved Solids (TDS)
   applied_pressure:
-    value: 7   # corresponds to default of 100 psi
+    value: 2   # corresponds to default of 29 psi
```
adam-a-a, line 132: We shouldn't be changing the value here. These are defaults that will break other tests when changed. 

Instead, we should only make the change within the flowsheet of interest itself (or possibly not include in version on main at all).

MarcusHolly: @adam-a-a This value in the yaml is specifically for the rHGO_dye_rejection subclass, so shouldn't this just reflect the actual value. The 2 bar number was an initial assumption we've made, but the 7 bar is Tina's updated  number. Also, I've already updated the tests that this change broke. The current failures are a result of the repo-wide issue from 2-ish weeks ago. Pulling from main should resolve all the test failures.

adam-a-a: Good point that I overlooked. Since this is attributed to the subtype, this should be fine.

adam-a-a (approved):

luohezhiming (approved):

Issue #1730, opened 2026-01-28
Stoichiometric Reactor- report() method failure


adam-a-a:
### Description

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[3], [line 1](vscode-notebook-cell:?execution_count=3&line=1)
----> [1](vscode-notebook-cell:?execution_count=3&line=1) m.fs.unit.report()

File c:\Users\Adam\.conda\envs\academy-reaktoro\Lib\site-packages\idaes\core\base\process_base.py:337, in ProcessBlockData.report(self, time_point, dof, ostream, prefix)
    334 performance = self._get_performance_contents(time_point=time_point)
    336 # Get stream table
--> [337](file:///C:/Users/Adam/.conda/envs/academy-reaktoro/Lib/site-packages/idaes/core/base/process_base.py:337) stream_table = self._get_stream_table_contents(time_point=time_point)
    339 # Set model type output
    340 if hasattr(self, "is_flowsheet") and self.is_flowsheet:

File ~\my-nawi-hub\watertap\watertap\unit_models\stoichiometric_reactor.py:800, in StoichiometricReactorData._get_stream_table_contents(self, time_point)
    796 def _get_stream_table_contents(self, time_point=0):
    797     return create_stream_table_dataframe(
    798         {
    799             "Inlet": self.inlet,
--> [800](https://file+.vscode-resource.vscode-cdn.net/c%3A/Users/Adam/my-nawi-hub/watertap_academy_fall_2025/week8/~/my-nawi-hub/watertap/watertap/unit_models/stoichiometric_reactor.py:800)             "Reactor outlet": self.reactor_outlet,
    801             "Outlet": self.outlet,
    802             "Waste": self.waste,
    803         },
    804         time_point=time_point,
    805     )

File c:\Users\Adam\.conda\envs\academy-reaktoro\Lib\site-packages\pyomo\core\base\block.py:552, in BlockData.__getattr__(self, val)
    549     return _component_decorator(self, ModelComponentFactory.get_class(val))
    550 # Since the base classes don't support getattr, we can just
    551 # throw the "normal" AttributeError
--> [552](file:///C:/Users/Adam/.conda/envs/academy-reaktoro/Lib/site-packages/pyomo/core/base/block.py:552) raise AttributeError(
    553     "'%s' object has no attribute '%s'" % (self.__class__.__name__, val)
    554 )

[...]

PR #1734, opened 2026-02-06
Resolve report issue of stoichiometric reactor

luohezhiming:
## Fixes/Resolves:

Fixed issue #1730 

-
-

### Legal Acknowledgement

### DIFF
--- a/watertap/unit_models/stoichiometric_reactor.py
+++ b/watertap/unit_models/stoichiometric_reactor.py
@@ -788,9 +788,10 @@ class StoichiometricReactorData(UnitModelBlockData):
                 var_dict["Reagent dose - " + r] = self.reagent_dose[r]
                 var_dict["Reagent flow - " + r] = self.flow_mass_reagent[r]
 
-        for p in self.precipitate_list:
-            var_dict["Precipitate flow - " + p] = self.flow_mass_precipitate[p]
-            var_dict["Precipitate conc - " + p] = self.conc_mass_precipitate[p]
+        if len(self.config.precipitate.keys()):
+            for p in self.precipitate_list:
+                var_dict["Precipitate flow - " + p] = self.flow_mass_precipitate[p]
+                var_dict["Precipitate conc - " + p] = self.conc_mass_precipitate[p]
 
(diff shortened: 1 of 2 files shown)


MarcusHolly (approved):

MarcusHolly (changes requested):
Approved before looking at the tests... seems there's an issue with precipitate_list

watertap/unit_models/stoichiometric_reactor.py
```diff
@@ -783,9 +783,10 @@ def _get_performance_contents(self, time_point=0):
...
-        for r in self.reagent_list:
-            var_dict["Reagent dose - " + r] = self.reagent_dose[r]
-            var_dict["Reagent flow - " + r] = self.flow_mass_reagent[r]
+        if len(self.config.reagent.keys()):
+            for r in self.reagent_list:
+                var_dict["Reagent dose - " + r] = self.reagent_dose[r]
+                var_dict["Reagent flow - " + r] = self.flow_mass_reagent[r]
 
         for p in self.precipitate_list:
             var_dict["Precipitate flow - " + p] = self.flow_mass_precipitate[p]
```
adam-a-a: without looking at this thoroughly, you'll need to recheck all the logic here to make sure the report method will still work regardless of the configuration chosen for stoich reactor. If you do not choose precipitation reactor option and only consider dissolution, for example, I don't think there will be a precipitate_list. 

Good job on taking the initiative to fix this. Now we just need to make sure that report works for every configuration option.

adam-a-a (changes requested):
Need to test all configurations of the reactor and make sure report works in all modes

watertap/unit_models/stoichiometric_reactor.py
luohezhiming: @adam-a-a I resolved the reporting conflicts, it now works correctly for all configuration options, but one of the user mode (py3.11/linux) check failed, not sure about the reason, seem a github issue

avdudchenko (approved):
LGTM

adam-a-a (approved):

MarcusHolly (approved):

Issue #1657, opened 2025-10-02
Add support for python 3.12 and 3.13


sufikaur:
Add testing to workflows and update code to support python 3.12 and 3.13 (#1739 )
Address runner image issues from watertap  [PR 1483](https://github.com/watertap-org/watertap/pull/1483)

See idaes-pse upgrades for reference [PR 1669]( https://github.com/IDAES/idaes-pse/pull/1669)

PR #1739, opened 2026-02-27
Add python 3.12 and 3.13 to workflow

sufikaur:
## Fixes/Resolves:
[Issue #1657](https://github.com/watertap-org/watertap/issues/1657)

## Summary/Motivation:
Add support for newer python versions

- adds additional pytest jobs with python versions 3.12 and 3.13
-

### Legal Acknowledgement

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -182,6 +182,8 @@ jobs:
         python-version:
           - '3.10'
           - '3.11'
+          - '3.12'
+          - '3.13'
         os:
           - linux
           - win64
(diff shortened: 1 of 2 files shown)


MarcusHolly:
During the previous dev call, we had discussed replacing the assertions of the condition numbers (for the BSM2-P test) with assertions of the log of the condition number. However, if we were to do so, we would also have to update the assertions that weren't failing in this test file and other flowsheet test files for the sake of consistency. Rather than going through this, I opted to add additional markers on the tests such that certain tests are only run on specific operating systems (Windows or Linux) and Python versions (3.10 - 3.13). If these tests continue to cause issues in the future, we should transition all the condition number assertions to the aforementioned log alternative.

ksbeattie (approved):
LGTM

dangunter:
all things considered, looks fine

PR #1741, opened 2026-03-04
2-Stage LSRRO GUI Fix

MarcusHolly:
## Fixes/Resolves:
A member of the Academy was running the LSRRO flowsheet in the GUI and found out that the 2-stage system couldn't be built.

## Summary/Motivation:
Adds an if statement such that the total membrane area variable isn't built for a 2-stage or 1-stage system

- Updates the LSRRO UI file

### Legal Acknowledgement

### DIFF
--- a/watertap/flowsheets/lsrro/lsrro_ui.py
+++ b/watertap/flowsheets/lsrro/lsrro_ui.py
@@ -599,19 +599,17 @@ def export_variables(flowsheet=None, exports=None, build_options=None, **kwargs)
         is_output=True,
         output_category="System metrics",
     )
-    if len(fs.Stages) > 1:
-        total_area = sum(fs.ROUnits[i].area for i in fs.Stages)
-        exports.add(
-            obj=total_area,
-            name="Total membrane area",
-            ui_units=pyunits.m**2,
-            display_units="m^2",
-            rounding=2,
-            description="Total membrane area",
-            is_input=False,
-            is_output=True,
-            output_category="System metrics",
-        )
+    exports.add(
+        obj=fs.total_membrane_area,
+        name="Total membrane area",
+        ui_units=pyunits.m**2,
+        display_units="m^2",
+        rounding=2,
+        description="Total membrane area",
+        is_input=False,
+        is_output=True,
+        output_category="System metrics",
+    )
     exports.add(
         obj=fs.annual_feed,
         name="Annual feed flow",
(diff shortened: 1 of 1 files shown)


watertap/flowsheets/lsrro/lsrro_ui.py
```diff
@@ -599,18 +599,19 @@ def export_variables(flowsheet=None, exports=None, build_options=None, **kwargs)
         is_output=True,
         output_category="System metrics",
     )
-    total_area = sum(fs.ROUnits[i].area for i in fs.NonFinalStages)
```
adam-a-a, line 602: This original logic is wrong as it only adds up membrane for stages 1 through N-1 instead of stages 1 through N

watertap/flowsheets/lsrro/lsrro_ui.py
```diff
@@ -599,18 +599,19 @@ def export_variables(flowsheet=None, exports=None, build_options=None, **kwargs)
...
-        ui_units=pyunits.m**2,
-        display_units="m^2",
-        rounding=2,
-        description="Total membrane area",
-        is_input=False,
-        is_output=True,
-        output_category="System metrics",
-    )
+    if len(fs.NonFinalStages) > 1:
+        total_area = sum(fs.ROUnits[i].area for i in fs.NonFinalStages)
```
adam-a-a: This would extend the erroneous total membrane area calculation. Here's the fix:
```suggestion
    total_area = sum(fs.ROUnits[i].area for i in fs.Stages)
```

adam-a-a (changes requested):
We need to correct the total area calculation.

watertap/flowsheets/lsrro/lsrro_ui.py
adam-a-a: Now, someone MIGHT want to experiment and choose 1 stage, in which case you'd just get a single conventional RO stage. I think the suggested change would work either way, but worth a check

MarcusHolly: 1 stage worked for the original fix, but I'm double checking now that it works when switching over to fs.Stages

adam-a-a: In fact, @dangunter reminded me of something--this total area is already calculated as an expression on the flowsheet, so we just need to reference that: `m.fs.total_membrane_area `

MarcusHolly: Addressed

adam-a-a (approved):

adam-a-a:
This should be good to merge. We just need one more approval. Additionally, we should notify @MichaelPesce so that we can incorporate into the next GUI release

MichaelPesce (approved):
tested on the GUI, works for me

PR #1747, opened 2026-03-17
Reaktoro environment

sufikaur:
## Fixes/Resolves:

-
-

### Legal Acknowledgement


environment.yml
```diff
@@ -0,0 +1,62 @@
+name: watertap-reaktoro
+
+channels:
+  - conda-forge
+
+dependencies:
+  # core language
+  - python>=3.10,<3.12
+
+  # conda only packages
```
avdudchenko: I guess why are we manually managing all sub dependencies - it looks like most of these are reaktoro based dependencies and cyipopt - we should let those packages manage their own dependency tree (.e.g conda will solve that for us and their recipes include correct dependencies.

environment.yml
```diff
@@ -0,0 +1,62 @@
...
+  - mumps-seq=5.8.2
+  - eigen=3.4.0
+  - pybind11=2.13.6
+  - nlohmann_json=3.12.0
+  - fmt=12.1.0
+  - spdlog=1.17.0
+  - yaml-cpp=0.8.0
+  - multipledispatch=0.6.0
+
+  # Jupyter environment
```
avdudchenko: We should let watertap pip install manage these requirements separately - and once watertap is part of conda-forge we can move this out I guess... but the watertap conda recepie will handles all the requirements so this is redundunt.

avdudchenko (commented):
This looks good, but seems we are tracking sub-dependencies that are normally managed by the conda-forge recopies and toml files. So it seems a tad redundant?

ksbeattie:
This is on hold until the conda packaging of WaterTAP is complete and also waiting on a decision on if we really want to support so many environment files.

PR #1751, opened 2026-03-27
RO w/ERD GUI Improvements

### LINKED ISSUE #1706: RO w/ERD in GUI needs major improvements
Certain variables that should be exposed as inputs are not, and certain variables that should not be exposed as inputs are. 

Capital annualization factor is exposed, but plant lifetime and discount factor are not. Fixing annualization factor will result in error because of overspecification. Thus, we should expose plant lifetime and discount factor too. 

Membrane area is not exposed in inputs, so I cannot choose to change area as I please (neither is length). 

This is just the start.

MarcusHolly:
## Fixes/Resolves:

https://github.com/watertap-org/watertap/issues/1706

## Summary/Motivation:
Exposes more variables in the GUI for the reverse osmosis with energy recovery device flowsheet

- update list of exposed variables
-

### Legal Acknowledgement


watertap/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery_ui.py
```diff
@@ -149,6 +162,17 @@ def export_variables(flowsheet=None, exports=None, build_options=None, **kwargs)
...
+    exports.add(
+        obj=fs.RO.length,
+        name="RO stage length",
+        ui_units=pyunits.m,
+        display_units="m",
+        rounding=2,
+        description="Stage length",
+        is_input=True,
+        input_category="Reverse Osmosis",
+        is_output=False,
```
MuktaHardikar: Do we want to update this to be `is_output = True`?

MarcusHolly: I've set this to true for width and length now. Lmk if any other variables should be exposed or hidden

MuktaHardikar:
Variable that would be useful in the output tab I think:
1. Is the pressure drop per stage 
2. Total pressure drop- Easier to compare with experimental data. 
4. Total power along with SEC.

watertap/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery_ui.py
```diff
@@ -253,6 +277,28 @@ def export_variables(flowsheet=None, exports=None, build_options=None, **kwargs)
         input_category="System costing",
```
MuktaHardikar, line 437: This is a minor thing but I would the move system costing section after product and disposal sections

watertap/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery_ui.py
MuktaHardikar: Actually could we rename this to total membrane length. Because its not length per stage right? Its no of stages x stage per length

MarcusHolly:
I've also added config options for the ERD_type since that was missing.

watertap/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery_ui.py
```diff
@@ -80,8 +88,43 @@ def export_variables(flowsheet=None, exports=None, build_options=None, **kwargs)
...
+        obj=fs.RO.area,
+        name="RO membrane area",
+        ui_units=pyunits.m**2,
+        display_units="m2",
+        rounding=2,
+        description="RO membrane area",
+        is_input=True,
+        input_category="Reverse Osmosis",
+        is_output=True,
+        output_category="System metrics",
```
MuktaHardikar:
```suggestion
        output_category="Reverse Osmosis",
```

watertap/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery_ui.py
```diff
@@ -163,27 +217,170 @@ def export_variables(flowsheet=None, exports=None, build_options=None, **kwargs)
     )
```
MuktaHardikar:
```suggestion

```

watertap/flowsheets/RO_with_energy_recovery/RO_with_energy_recovery_ui.py
```diff
@@ -254,101 +451,40 @@ def export_variables(flowsheet=None, exports=None, build_options=None, **kwargs)
...
-        is_input=False,
-        is_output=True,
-        output_category="Disposal",
+        description="Electricity cost",
+        is_input=True,
+        input_category="System costing",
+        is_output=False,
     )
 
     # System metrics
```
MuktaHardikar, line 474:
```suggestion
    # System metrics
    exports.add(
        obj=fs.RO.recovery_mass_phase_comp[0, "Liq", "H2O"],
        name="RO water mass recovery",
        ui_units=pyunits.dimensionless,
        display_units="fraction",
        rounding=2,
        description="Water mass recovery of RO unit",
        is_input=True,
        input_category="Reverse Osmosis",
        is_output=True,
        output_category="System metrics",
    )
```

MuktaHardikar (changes requested):
Made some suggestions just to move all system levels outputs at the end. I think everything else looks good.

MuktaHardikar (approved):

kurbansitterley (approved):

Issue #1752, opened 2026-04-07
Improve Platform Usability


MarcusHolly:
### Description

Complete the following action items to improve the new user experience:

- [ ] **Create a baseline, end-to-end seawater RO flowsheet that is accessible in the GUI as a SWRO full plant template, from intake to outfall** #1763 
- [x] Create a utility function that lists all ports on a unit model #1757 
- [ ] Create a utility function that lists all properties on a property model #1686 
- [ ] Create a utility function that lists all the model parameters the user must fix
- [ ] Create a better workflow for specifying feed conditions on a unit model
- [ ] Improve documentation references and add more visualization (e.g., m.fs.vsiualize() & IDAES Connectivity Tool)

### Motivation

These developments were identified by stakeholders in the WaterTAP Academy. See https://app.box.com/file/2124448237821 for the full report on these roadmapping efforts.

adam-a-a:
Note- I added a task here: "Create a baseline, end-to-end seawater RO flowsheet that is accessible in the GUI as a SWRO full plant template." This is something that was directly requested by at least one of our Academy participants and has come up in discussions with other stakeholders. Thus, I am marking it here is one of our higher priority tasks to improving usability, as we already have a seawater_RO flowsheet that is representative of a flowsheet. Now, we just need to reevaluate and see if there are improvements to be made that would enable users to customize it, then port that to the GUI.

kurbansitterley:
> Create a utility function that lists all the model parameters the user must fix

@adam-a-a Do you have  any vision for the implementation of this? My initial thought is that this could vary widely depending on the data you have and the type of analysis you are trying to do.

PR #1757, opened 2026-04-08
Utility Function for Listing Unit Model Ports

MarcusHolly:
## Fixes/Resolves:

First item in #1752 

## Summary/Motivation:
Addresses advice from stakeholders to make it easier to see the ports on a unit model

- Adds a utility function that prints all the ports on a unit model
- Adds tests

### Legal Acknowledgement


watertap/core/util/misc.py
```diff
@@ -44,3 +47,26 @@ def smooth_heaviside(x, k):
     """
     function = 1 / (1 + exp(-k * x))
     return function
+
+
+def list_ports(unit):
```
adam-a-a: It would be useful if this could descend into a flowsheet and optionally list all ports on unit models in that flowsheet. That is, this should be applicable to a single unit model, but also to a flowsheet of several models that perhaps have not been connected yet; the user would seek to connect them and would need to know port names.

watertap/core/util/misc.py
```diff
@@ -44,3 +47,26 @@ def smooth_heaviside(x, k):
...
+    if not isinstance(unit, UnitModelBlockData):
+        raise TypeError(
+            f"Expected a UnitModelBlockData instance, but "
+            f"got {type(unit).__name__!r}."
+        )
+
+    ports = dict(unit.component_map(Port))
+
+    for name, port in ports.items():
+        print(f" The {name} port is {port}")
```
adam-a-a: It'd be nice to generate a table with Unit | Port Name | Description, formatted in a similar fashion to #1686

adam-a-a: Rather than printing a string, having something that can be read or written easily (e.g., export result to csv, etc.). I think that was what I was going for in #1686

MarcusHolly:
This is what the output looks like for an individual model and a flowsheet (BSM2-P)
image

kurbansitterley:
for information - I also created this same file with a test in #1699

watertap/core/util/unit_models.py
```diff
@@ -0,0 +1,61 @@
...
+        units = [block]
+
+    for unit in units:
+        ports = dict(unit.component_map(Port))
+        for name, port in ports.items():
+            rows.append(
+                {
+                    "Unit Model": type(unit).__name__.removeprefix("_Scalar"),
+                    "Port Name": name,
+                    "Port": port.name,
```
MarcusHolly, line 81: Add columns for any streams connected at the frontend and backend

watertap/core/util/unit_models.py
```diff
@@ -0,0 +1,61 @@
...
+                }
+            )
+
+    df = pd.DataFrame(rows)
+
+    pd.set_option("display.max_rows", None)
```
MarcusHolly: Can remove this and set back to the default

adam-a-a:
We'll eventually migrate this to IDAES to be used across any IDAES flowsheet or unit model. We should also consider descend_into functionality (@dallan-keylogic recommends NOT descending by default if applied at unit model level). 

For now, implementing as we are now for WaterTAP is a sufficient stepping stone in line with our milestone.

watertap/core/util/flowsheet.py
```diff
@@ -0,0 +1,61 @@
...
+        units = [block]
+
+    for unit in units:
+        ports = dict(unit.component_map(Port))
+        for name, port in ports.items():
+            rows.append(
+                {
+                    "Unit Model": type(unit).__name__.removeprefix("_Scalar"),
+                    "Port Name": name,
+                    "Port": port.name,
```
adam-a-a: Right- so a "Connection" column that either reads "Unconnected" if not yet connected to anything, versus the name of the model and port it is connected to.

watertap/core/util/flowsheet.py
```diff
@@ -0,0 +1,61 @@
...
+            f"Expected a UnitModelBlockData or FlowsheetBlockData instance, but "
+            f"got {type(block).__name__!r}."
+        )
+
+    rows = []
+
+    if isinstance(block, FlowsheetBlockData):
+        units = [
+            u
+            for u in block.component_objects(Block)
```
adam-a-a, line 73:
```suggestion
            for u in block.component_objects(Block, descend_into=descend_into)
```
I think we may want to allow the user to specify descend_into in list_ports, and we should experiment with setting the default to either True or False before settling on default. @dallan-keylogic recommended False.

kurbansitterley (commented):
Where did we land on a `descend_into` kwarg?

Also, do we want documentation for this here or a future PR?

MarcusHolly:
> Where did we land on a `descend_into` kwarg?
> 
> Also, do we want documentation for this here or a future PR?

Regarding descend_into, I forget if we wanted to apply it on the flowsheet-level or not. For documentation, do we have a designated spot for utility functions? Since we plan on adding a bunch of utility functions in the near future, perhaps we should have a spot for them under technical references. I think documentation can be handled in a subsequent PR tho

kurbansitterley:
> For documentation, do we have a designated spot for utility functions? Since we plan on adding a bunch of utility functions in the near future, perhaps we should have a spot for them under technical references. I think documentation can be handled in a subsequent PR tho

I started a docs page for unit model utilities in #1699 but I could just as easily drop it there and we could do all docs for these utilities in a separate PR.

MarcusHolly:
> > For documentation, do we have a designated spot for utility functions? Since we plan on adding a bunch of utility functions in the near future, perhaps we should have a spot for them under technical references. I think documentation can be handled in a subsequent PR tho
> 
> I started a docs page for unit model utilities in #1699 but I could just as easily drop it there and we could do all docs for these utilities in a separate PR.

We can just bring this up in the dev call tomorrow

dallan-keylogic:
You should test the cases when:
1. There are multiple arcs connected to a single port 
2. Some or all arcs are deactivated. (Note: this means that the `arc_expanded` block is deactivated, not that the `Arc` object itself is deactivated. I think in this case you should mention that the arc exists, but also include "(deactivated)" as a postscript.
3. Specifically mention if a port has any variables set up as `Extensive` instead of `Equality` (which is the default). [Pyomo allows the user to split variables between different streams](https://pyomo.readthedocs.io/en/stable/explanation/modeling/network.html#arc-expansion-transformation) using the `Extensive` arc expansion rule. This rule is declared on the port with certain variables. As far as I know, we never use that rule in any IDAES+ projects, but perhaps users might do so in a custom unit model.

Cases (1) and (2) are important to the Batch RO work that I'm doing. To avoid state blocks with zero flows, I have two arcs connected to a port. When we're recirculating brine, I activate the recirculation `arc_expanded` block and deactivate the purge `arc_expanded` block. When in purge mode, I deactivate the recirculation arc and activate the purge arc.

MarcusHolly:
> You should test the cases when:
> 
> 1. There are multiple arcs connected to a single port
> 2. Some or all arcs are deactivated. (Note: this means that the `arc_expanded` block is deactivated, not that the `Arc` object itself is deactivated. I think in this case you should mention that the arc exists, but also include "(deactivated)" as a postscript.
> 3. Specifically mention if a port has any variables set up as `Extensive` instead of `Equality` (which is the default). [Pyomo allows the user to split variables between different streams](https://pyomo.readthedocs.io/en/stable/explanation/modeling/network.html#arc-expansion-transformation) using the `Extensive` arc expansion rule. This rule is declared on the port with certain variables. As far as I know, we never use that rule in any IDAES+ projects, but perhaps users might do so in a custom unit model.
> 
> Cases (1) and (2) are important to the Batch RO work that I'm doing. To avoid state blocks with zero flows, I have two arcs connected to a port. When we're recirculating brine, I activate the recirculation `arc_expanded` block and deactivate the purge `arc_expanded` block. When in purge mode, I deactivate the recirculation arc and activate the purge arc.

Addressed 1 and 2, although you should check the implementation for 2 to make sure it suits your needs. As for 3, I think we can leave that off for now. I also don't think it's been used in any of the IDAES+ projects. Plus, if someone is tech-savvy enough to use Extensive arc expansions, they won't be struggling to find ports on a unit model.

dallan-keylogic:
One more thing: it would probably be good to add a column with the arc names in addition to the source and destination ports. That would probably make @dangunter 's life easier when trying to grab information for the GUI.

dallan-keylogic (approved):
LGTM

adam-a-a (approved):
LGTM

PR #1760, opened 2026-04-17
Fixes redundunt constraint in SteamHeater0D

avdudchenko:
## Fixes/Resolves:
The steam heater was building outlet_liquid_mass_balance, which is redundant because the heat exchanger already builds a standard material balance constraint on the hot side, which performs the same operation.  This error resulted in the tests being incorrectly written, as they left the area of the heat exchanger unfixed, even though to fully specify the system, besides supplying the heat transfer rate, the area or heat duty must be fixed. The tests did not specify the heat duty or area, and due to extra constraint passed the 0DOF test. 

This resulted in the tests being incorrectly written 
## Summary/Motivation:
The SteamHetear0D was unstable due to a duplicate constraint. This should improve stability. 

- removed duplicate outlet_liquid_mass_balance constraint
- removed heater mode, as it does not do anything. 
- fixed test to be proper 0DOF tests
- resolved issue where the tests were passing, but the actual IPOPT output showed the solves did not succeed due to multiplier errors, which was a result of duplicate errors
- reworked initialization routine to be more in line with standard approaches

### Legal Acknowledgement


ElmiraShamlou (commented):
The capability to represent liquid/vapor equilibrium through mole fractions in the property package directly affects how this unit should be modeled and how the DOF should be closed. The PR correctly removes the redundant constraint, but it then replaces it with two non-ideal ways of handling the extra DOF: (1) it fixes the area, which indirectly fixes the heat duty, and fixes the vapor outlet content. This only works if the selected area is large enough to bring the system to a fully saturated or subcooled liquid state; otherwise, the model becomes infeasible. In addition, this approach gives you no real control over whether the outlet is exactly saturated or how much subcooling occurs--you simply observe whatever state happens to be feasible. (2) It removes the heater mode and its associated test, even though that mode was specifically designed for cases where there is no upstream unit connected on the hot side and the steam extraction rate needs to be estimated. The unit-specific initialization was added precisely to make that case work with the IDAES heat exchanger initialization routine.

avdudchenko:
> The capability to represent liquid/vapor equilibrium through mole fractions in the property package directly affects how this unit should be modeled and how the DOF should be closed. The PR correctly removes the redundant constraint, but it then replaces it with two non-ideal ways of handling the extra DOF: (1) it fixes the area, which indirectly fixes the heat duty, and fixes the vapor outlet content. This only works if the selected area is large enough to bring the system to a fully saturated or subcooled liquid state; otherwise, the model becomes infeasible. In addition, this approach gives you no real control over whether the outlet is exactly saturated or how much subcooling occurs--you simply observe whatever state happens to be feasible. (2) It removes the heater mode and its associated test, even though that mode was specifically designed for cases where there is no upstream unit connected on the hot side and the steam extraction rate needs to be estimated. The unit-specific initialization was added precisely to make that case work with the IDAES heat exchanger initialization routine.

I am not sure what you means that I added 2 non-ideal ways to handle the DOF.

I do not fix any additional variables in the model to address the DOF - the user must specify it

1) The initialization routines are updates so they comply with our standard practice of being agnostic to integration in the flowsheet.
2) The user can deactivate constraints before initialization, and reinitialize with them after the fact
3) If the user wants to not have any subcooling ,they can simply deactivate sat constraint and add constraint that requires outlet pressure equals saturation pressure (or we should supply that as an option through a config...) I am not really sure why you are discussing this as the changes in this PR do not change the behavior of the unit in anyway. And initialization routines should not set the unit behavior, as its expected that after initialization the model will be resolved, and some variables will be unfixed to enable optimization. Neither heater nor condenser mode set any additional constraints at build to configure unit operation.

For the heater - condenser modes:
This only changed how the model initialized.

1) For Heater mode, the area was fixed but not estimated in any way, and the outlet sat pressure, and redundant constraints were deactivated to allow the heater to initialize
- it is not clear to me how this helped if the unit was not connected to anything or configured anyway, as if the outlet pressure was still above sat it would still fail to initialize and fail to solve at the next step.
2) For the Condenser, the model was just initialized as is.

Basically, the two initialization routines arrive at the same state, and will fail for the same reasons.

Our initialization routines in general should not fix or unfix variables for the user except inlet state vars:
1) The user should decide how to configure the unit model for initialization. (e.g. which variables to fix)
- Both modes resulted in total condensation of vapor.
- If we want a special initialization option, it should be passed as such (e.g solve_with_deactivated_pressure_sat_first).
- I am actually in favor of removing estimate_cooling_water, but that is a bit more clear than "heater or condenser modes" and what they do.  At least it does not fix variables (I removed its prior behavior of unfixing area).

ElmiraShamlou:
The main motivation behind this unit model was to serve as a true steam heater capable of estimating the steam extraction rate, not as a simple condenser. I think the source of our disagreement is that this particular use case may not have come up in the use cases you have considered so far. The crystallizer_MVR flowsheet works fine with the simplest condenser model because it does not need to estimate an extraction rate; the steam flow there is already pinned by the upstream compressor. A true steam heater, like the kind used in a live-steam preheater, is a different problem: the steam extraction rate is the unknown the unit must determine from the heat duty on the cold side.

To see why the current PR does not cover this case, I would recommend two reads: the IDAES FeedwaterHeater0D model and the IDAES HeatExchanger initialization routine. More importantly, I would recommend testing this PR directly on an extraction-rate case with hot-side T and P fixed, Liq_in = 0, Vap_in left as the unknown, and area plus U fixed. What you will see is that the unit sits at DOF = +1. The P_out >= P_sat inequality is slack and does not pin anything on the model side. The only way to close that DOF is to promote the saturation condition from an inequality to an equality, meaning the hot outlet must sit exactly on the saturation line.

Even once the user does that and the unit-level DOF is clean, the next issue is that IDAES’s stock HX initialization routine fails with “too few DOF” at Stage 2. Briefly, Stage 2 deactivates heat_transfer_equation, fixes cold_side.heat, and runs a decoupled solve. A unit-level saturation equality added on top of that stage’s own fixing and deactivation breaks its DOF accounting. This is exactly why FeedwaterHeater0D.initialize_build takes a snapshot, deactivates its extraction closure, fixes inlets plus area plus U, unfixes both outlets, calls the parent initialization, and then restores everything. The fix/unfix sequence during initialization is not decorative; it is what allows a unit with a closure that only makes sense at the unit level to survive a multi-stage HX initialization.

On the second point, I think we should also be honest that the existing tests, even for the simple condenser case, pass by luck rather than by design. We pin Vap_out = 0 at build time and choose an area/U pair that happens to push the outlet to or below saturation for the specific flows in the test. That works for those specific numbers, but it is not a design; it is a coincidence that the hard-coded constraint happens to be consistent with the solution. If we had a real VLE mechanism, the outlet would determine its own phase split and we would not have to guess. I see there is another PR adding steam-side properties to water_prop_pack. If, as part of that, we also add temperature_sat(P) and enth_mol_sat_phase["Liq"] / enth_mol_sat_phase["Vap"], we would be able to model this unit much more cleanly and make its behavior reflect real physics instead of a fortunate parameter choice.

avdudchenko:
> The main motivation behind this unit model was to serve as a true steam heater capable of estimating the steam extraction rate, not as a simple condenser. I think the source of our disagreement is that this particular use case may not have come up in the use cases you have considered so far. The crystallizer_MVR flowsheet works fine with the simplest condenser model because it does not need to estimate an extraction rate; the steam flow there is already pinned by the upstream compressor. A true steam heater, like the kind used in a live-steam preheater, is a different problem: the steam extraction rate is the unknown the unit must determine from the heat duty on the cold side.
> 
> To see why the current PR does not cover this case, I would recommend two reads: the IDAES FeedwaterHeater0D model and the IDAES HeatExchanger initialization routine. More importantly, I would recommend testing this PR directly on an extraction-rate case with hot-side T and P fixed, Liq_in = 0, Vap_in left as the unknown, and area plus U fixed. What you will see is that the unit sits at DOF = +1. The P_out >= P_sat inequality is slack and does not pin anything on the model side. The only way to close that DOF is to promote the saturation condition from an inequality to an equality, meaning the hot outlet must sit exactly on the saturation line.
> 
> Even once the user does that and the unit-level DOF is clean, the next issue is that IDAES’s stock HX initialization routine fails with “too few DOF” at Stage 2. Briefly, Stage 2 deactivates heat_transfer_equation, fixes cold_side.heat, and runs a decoupled solve. A unit-level saturation equality added on top of that stage’s own fixing and deactivation breaks its DOF accounting. This is exactly why FeedwaterHeater0D.initialize_build takes a snapshot, deactivates its extraction closure, fixes inlets plus area plus U, unfixes both outlets, calls the parent initialization, and then restores everything. The fix/unfix sequence during initialization is not decorative; it is what allows a unit with a closure that only makes sense at the unit level to survive a multi-stage HX initialization.
> 
> On the second point, I think we should also be honest that the existing tests, even for the simple condenser case, pass by luck rather than by design. We pin Vap_out = 0 at build time and choose an area/U pair that happens to push the outlet to or below saturation for the specific flows in the test. That works for those specific numbers, but it is not a design; it is a coincidence that the hard-coded constraint happens to be consistent with the solution. If we had a real VLE mechanism, the outlet would determine its own phase split and we would not have to guess. I see there is another PR adding steam-side properties to water_prop_pack. If, as part of that, we also add temperature_sat(P) and enth_mol_sat_phase["Liq"] / enth_mol_sat_phase["Vap"], we would be able to model this unit much more cleanly and make its behavior reflect real physics instead of a fortunate parameter choice.

I am still really confused. The current model does not do what you are expecting it to do.

First of the steam heater does not enforce `P_out==P_sat`  when configured as a Steam Heater. So I am a bit lost as to why it is expected that initialization routines are written for a non-existent behavior.

If you want that behavior in the unit, it should explicitly enforce it when the heater mode is provided (e.g. set up `P_out==P_sat` constraint) and then perform the initialization approach you describe.

Regarding point two, if the user supplies bad guess the current initialization will fail anyway, this in general is true regardless of how you write the initialization routine.

The purpose of the PR is to fix prior incorrect implementation and assumptions, but what you are suggesting seems to be:
- Add new behavior that was not previously there (P_sat=P_out)
- Add rigorous initialization methods to support this behavior
- Include more rigorous tests, as prior tests (Which remained mostly unchanged) are insufficient. 

I would envision this range of work to be outside the scope of this PR as an enhancement that goes after this PR if anyone has motivation and time to do so ( as I do not have either).

In the current form the unit model does not really achieve any of the objectives your comments are focused on and it will lead to unstable models due to the re-undulant constraint.


Finally, the user can achieve the described behavior if they do the following:

```python
   
    # build the model
    m = ConcreteModel()
    m.fs.steam_properties = props_w.WaterParameterBlock()
    m.fs.cold_side_properties = props_sw.SeawaterParameterBlock()

    m.fs.unit = SteamHeater0D(
        hot_side_name="hot",
        cold_side_name="cold",
        hot={
            "property_package": m.fs.steam_properties,
        },
        cold={
            "property_package": m.fs.cold_side_properties,
        },
        delta_temperature_callback=delta_temperature_chen_callback,
        flow_pattern=HeatExchangerFlowPattern.countercurrent,
    )
    m.fs.cold_side_properties.set_default_scaling(
        "flow_mass_phase_comp",
        1,
        index=("Liq", "H2O"),
    )
    m.fs.cold_side_properties.set_default_scaling(
        "flow_mass_phase_comp",
        100,
        index=("Liq", "TDS"),
    )
    m.fs.steam_properties.set_default_scaling(
        "flow_mass_phase_comp",
        1,
        index=("Liq", "H2O"),
    )
    m.fs.steam_properties.set_default_scaling(
        "flow_mass_phase_comp",
        1,
        index=("Vap", "H2O"),
    )
    # setup the cold side
    m.fs.unit.cold_side_inlet.pressure.fix(101325)
    m.fs.unit.cold_side_inlet.temperature.fix(273.15 + 25)
    total_flow_mass = 10
    TDS_mass_frac = 0.035
    m.fs.unit.cold_side_inlet.flow_mass_phase_comp[0, "Liq", "TDS"].fix(
        TDS_mass_frac * total_flow_mass
    )
    m.fs.unit.cold_side_inlet.flow_mass_phase_comp[0, "Liq", "H2O"].fix(
        (1 - TDS_mass_frac) * total_flow_mass
    )
    
    # guess an initial area
    m.fs.unit.area.fix(10)
    m.fs.unit.overall_heat_transfer_coefficient.fix(2e3)
    iscale.calculate_scaling_factors(m.fs.unit)

    # unfix vapor pressure
    m.fs.unit.hot_side_inlet.flow_mass_phase_comp[0, "Vap", "H2O"].unfix()
    m.fs.unit.hot_side_inlet.flow_mass_phase_comp[0, "Liq", "H2O"].fix(0)

    # fix pressure and temperature for hot side inlet
    m.fs.unit.hot_side_inlet.temperature.fix(273.15 + 140)
    m.fs.unit.hot_side_inlet.pressure[0].fix(201325)

    # Deactivate loose constraint on unit for sat pressure
    m.fs.unit.outlet_pressure_sat.deactivate()
    
    # add constraint for sat equality
    m.fs.unit.eq_pressure_set = Constraint(
        expr=m.fs.unit.hot_side.properties_out[0].pressure
        == m.fs.unit.hot_side.properties_out[0].pressure_sat
    )
    # deactivate so we can initialize
    m.fs.unit.eq_pressure_set.deactivate()
    m.fs.unit.initialize()

    # activate the pressure sat constraint 
    m.fs.unit.eq_pressure_set.activate()
    # re-solve with new system 
```

avdudchenko:
This now uses same steam heater as in the PR here: #1482

ksbeattie:
Blocked by #1493

ElmiraShamlou (approved):
LGTM. We can change `pressure_sat` to `temperature_sat` in a separate PR once the steam properties PR is merged.

adam-a-a:
@ElmiraShamlou has some more updates that she will push to #1482

adam-a-a (approved):
LGTM - we should eventually make sure this model is properly documented.

PR #1761, opened 2026-04-18
Fixes issues in surrogate crystalzier formulation with MW, missing water, scaling, factors, and formulations.

avdudchenko:
## Fixes/Resolves:
This superceeds #1758 

- Fixes original formulation that did not track water when supplying solids
- Assumed MW was based on apparent species, which resulted in correct mol estimates
- Updates removal to be calculated using mol basis, might lead to better stability, but is also easier to read
- Updates solids_ion_dict to use the same format as stoichometeric reactor for consistency 
- Updates scaling, especially for solids formed based on limiting solid ion (works with either mass/mol basis or should).

-
-

### Legal Acknowledgement

3. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.


watertap/unit_models/surrogate_crystallizer.py
```diff
@@ -309,7 +375,11 @@ def eq_P_con1(b):
 
         @self.Constraint()
         def eq_P_con2(b):
-            return self.pressure_operating == b.properties_out_vapor[0].pressure
+            return self.vapor_pressure == b.properties_out_vapor[0].pressure
```
OOAmusat, line 378: Why do we want an additional variable and constraint for this?

watertap/unit_models/surrogate_crystallizer.py
```diff
@@ -514,11 +609,47 @@ def calculate_scaling_factors(self):
...
+                )
+            return sf
+
+        if iscale.get_scaling_factor(self.temperature_operating) is None:
+            iscale.set_scaling_factor(
+                self.temperature_operating,
+                iscale.get_scaling_factor(self.properties_in[0].temperature),
+            )
+        if iscale.get_scaling_factor(self.pressure_operating) is None:
+            iscale.set_scaling_factor(self.pressure_operating, 1e-5)
```
OOAmusat, line 644:
```suggestion
            iscale.set_scaling_factor(self.pressure_operating, 1e-3)
```

watertap/unit_models/surrogate_crystallizer.py
```diff
@@ -514,11 +609,47 @@ def calculate_scaling_factors(self):
...
+
+        if iscale.get_scaling_factor(self.vapor_pressure) is None:
+            iscale.set_scaling_factor(self.vapor_pressure, 1e-5)
```
OOAmusat, line 647:
```suggestion
            iscale.set_scaling_factor(self.vapor_pressure, 1e-3)
```

watertap/unit_models/surrogate_crystallizer.py
```diff
@@ -514,11 +609,47 @@ def calculate_scaling_factors(self):
...
+
+        iscale.set_scaling_factor(self.evaporation_percent, 1e-2)
```
OOAmusat:
```suggestion
        iscale.set_scaling_factor(self.evaporation_percent, 1e-1)
```

watertap/unit_models/surrogate_crystallizer.py
OOAmusat, line 644: Pressures in crystallizer systems tend to be in the 10^3-10^4 range, hence the suggested change in scaling.

watertap/unit_models/surrogate_crystallizer.py
```diff
@@ -528,6 +659,84 @@ def calculate_scaling_factors(self):
...
+        if (
+            iscale.get_constraint_transform_applied_scaling_factor(
+                self.eq_total_vapor_constraint
+            )
+            is None
+        ):
+            sf = iscale.get_scaling_factor(self.flow_mass_vap_total)
+            iscale.constraint_scaling_transform(self.eq_total_vapor_constraint, sf)
+
+        for j in self.flow_mass_sol_comp_true:
```
OOAmusat, line 683: So the mass scales based on the mol, and mol based on mass. What happens when both are None? It looks like it will scale on the default of ``flow_mass_phase_comp`` from the property package - is that correct?

OOAmusat (commented):
The crystallizer changes look good to me.

The main thing to discuss is the issue of streamlined steam costs. One thing that did occur to me about unifying the cost of steam is that it that assumes that all the steam-using units require steam of the same quality (i.e., pressure). That may not always be the case, which may be why having it per unit may not be a bad idea. Something to think about.

Once we decide on what to do there, the rest looks fine to me.

watertap/unit_models/surrogate_crystallizer.py
avdudchenko: I added this so that one could decouple the vapor pressure from the outlet pressure on the unit (for example we can assume that the outlet pressure of the surrogate is always at ambient conditions. (you need this to make MVR- crystalzier work.

watertap/unit_models/surrogate_crystallizer.py
avdudchenko: Our pressure is normally in Pa, so 1e-5 is more accurate for tihs to be consistent with other system pressures.

watertap/unit_models/surrogate_crystallizer.py
avdudchenko: same as above.

watertap/unit_models/surrogate_crystallizer.py
avdudchenko: It is generally better to maintain unit scaling consistency unless there are very large discrepancies in my experience.

watertap/unit_models/surrogate_crystallizer.py
avdudchenko: Umm, that should never happen... since that means the property package was never scaled. We should raise an error, but we don't do that in other places.

OOAmusat (approved):
LGTM

watertap/unit_models/tests/test_surrogate_crystallizer.py
```diff
@@ -221,11 +235,42 @@ def test_rbf_surrogate():
...
+    results_dict_mol = {
+        "Calcite_g": 1.2168759070266173e-06,
+        "Anhydrite_g": 0.005257278639031265,
+        "Glauberite_g": 0.2205018242772385,
+        "Halite_g": 25.1569597870144,
+    }
+    for key, obj in m.fs.cryst.flow_mol_sol_comp_apparent.items():
+        # results_dict_mol[key] = obj.value
+        assert pytest.approx(obj.value, rel=1e-6) == results_dict_mol[key]
+    # print(results_dict_mass)
```
adam-a-a:
```suggestion
```

watertap/unit_models/surrogate_crystallizer.py
```diff
@@ -514,11 +609,47 @@ def calculate_scaling_factors(self):
...
 
-        iscale.set_scaling_factor(
-            self.temperature_operating,
-            iscale.get_scaling_factor(self.properties_in[0].temperature),
+        def get_mol_scale(prop_type, j):
+            sf = iscale.get_scaling_factor(prop_type.flow_mol_phase_comp["Liq", j])
+            if sf is None:
+                sf = iscale.get_scaling_factor(
+                    prop_type.flow_mass_phase_comp["Liq", j]
+                ) / value(
```
ElmiraShamlou: This should be `* MW` 
`sf_mol= sf_mass * MW`

watertap/unit_models/surrogate_crystallizer.py
```diff
@@ -514,11 +609,47 @@ def calculate_scaling_factors(self):
...
+                    )
+                )
+            return sf
+
+        def get_mass_scale(prop_type, j):
+            sf = iscale.get_scaling_factor(prop_type.flow_mass_phase_comp["Liq", j])
+            if sf is None:
+                sf = iscale.get_scaling_factor(
+                    prop_type.flow_mol_phase_comp["Liq", j]
+                ) * value(
```
ElmiraShamlou: This should be `/MW`
`sf_mass = Sf_mol / MW`

ElmiraShamlou (changes requested):

watertap/unit_models/surrogate_crystallizer.py
avdudchenko: Done

watertap/unit_models/surrogate_crystallizer.py
avdudchenko: Done

ksbeattie:
@ElmiraShamlou, can you review after your suggested changes?

watertap/unit_models/surrogate_crystallizer.py
```diff
@@ -168,58 +211,81 @@ def build(self):
...
 
         self.heat_required = Var(
             initialize=1e5,
-            domain=NonNegativeReals,
+            domain=Reals,
             units=pyunits.kW,
             bounds=(-5e6, 5e6),
             doc="Heat requirement for crystallizer (kW)",
         )
+        self.vapor_pressure = Var(
```
ElmiraShamlou, line 282: The new `vapor_pressure` variable and accompanying `eq_P_con3` feel redundant and makes it hard to follow the pressure logic inside the crystallizer. The crystallizer's `operating_pressure` is really the chamber pressure, the pressure inside the vessel where the brine and vapor coexist, and at steady state this is the same as the brine's vapor pressure. Any downstream unit that takes the vapor stream sees it at the chamber pressure. So we're effectively carrying two names for the same thing.

If the goal is to decouple a pressure from `pressure_operating`, that decoupling should happen on the liquid outlet side, not the vapor side. That would actually reflect real plant operation, where the pump receiving liquid downstream of the chamber is typically mounted below it, and the liquid pressure at the pump suction is boosted by the static head of the brine column above it. So, we could just deactivate `eq_P_con1` and let the user fix the liquid outlet pressure directly or keep the constraint and add a static head term to it to provide the positive head the pump needs.

But I don't think we should decouple vapor pressure from the chamber operating pressure, they're the same physical quantity, and adding `vapor_pressure` as a separate variable seems to add complexity to the pressure logic without adding any extra capability.

ElmiraShamlou (changes requested):

watertap/unit_models/surrogate_crystallizer.py
ElmiraShamlou, line 338: Now that water is correctly tracked as part of the solid bookkeeping, it may also be worth reflecting that in the H₂O mass balance by adding `flow_mass_sol_comp_true` to the water balance. Otherwise, for hydrate solids, total mass in will not equal total mass out.

ksbeattie:
@avdudchenko any update on this (Elmira's comments)?

ksbeattie:
@ElmiraShamlou @avdudchenko, moving this to the Sept release board.  Please object if that's incorrect.

ElmiraShamlou (approved):

PR #1762, opened 2026-04-18
Unifies steam costing across steam using units

avdudchenko:
## Fixes/Resolves:
This fixes collision when using multiple units with steam costs. This also enables user to use different steam costs by default by providing units with a steam cost name. When costing a unit the steam cost is first checked for on the costing block, and if not it will be added.  This enables user to easily do things like

```
crystalzier_a.costing = UnitModelCostingBlock(
            flowsheet_costing_block=m.fs.costing,
            costing_method_arguments={"cost_steam_flow": True, "steam_type": "low_pressure_steam"},
        )
crystalzier_b.costing = UnitModelCostingBlock(
            flowsheet_costing_block=m.fs.costing,
            costing_method_arguments={"cost_steam_flow": True, "steam_type": "high_pressure_steam"},
        )
```
Additionally, the cost_steam_flow includes a unified set of functions to cost steam based on mass, heat duty, or volume flow and utilizes standard methods to estimate steam cost based on volume by providing appropriate conversions on the fly. 

## Summary/Motivation:
- Our steam costing was fundamentally incorrectly implemented, resulting in the inability to build multiple units that use steam on the same flow sheet due to name collisions. 

- add cost_steam_flow
- update crystalzier ,surrogate crystalizer, heat exchager, and steam ejector to use new steam costing method. 

### Legal Acknowledgement


watertap/costing/unit_models/crystallizer.py
```diff
@@ -119,7 +117,7 @@ def cost_crystallizer(blk, cost_type=CrystallizerCostType.default):
         )
 
 
-def _cost_crystallizer_flows(blk):
+def _cost_crystallizer_flows(blk, steam_type="steam"):
```
adam-a-a, line 120: Why is `steam_type` needed? Also "steam" as `steam_type` seems redundant. What are the expected options for `steam_type`?

kurbansitterley:
I encountered a similar issue in a flowsheet for REFLO. @avdudchenko did you consider adding the necessary steam properties to the property package (rather than this separate helper function)?

watertap/costing/tests/test_util.py
```diff
@@ -185,3 +186,69 @@ def test_rectifier_costing():
...
+    )
+
+    cost_steam_flow(
+        costing_package=m.fs.costing,
+        steam_cost_type="steam_heat_duty_basis",
+        steam_heat_duty=m.fs.heat_duty_for_steam,
+        steam_pressure=m.fs.steam_pressure,
+    )
+
+    cost_steam_flow(
```
adam-a-a, line 234: Note: We'll want to add a test to see if expected behavior occurs when "steam_cost" is already constructed, and the user calls cost_steam_flow again with the same `steam_cost_type`.

kurbansitterley: Does `register_costing_parameter_block` not make that check? See our conversation starting [here](https://github.com/watertap-org/watertap/pull/1482#issuecomment-3286761138)

avdudchenko: The issue is it build steam_cost as var on the block and registered it which did not check for redundant additions some how.  and I think @adam-a-a  comment is we do not have any explicit tests in current untis/flowsheets that validate multiple units sharing same steam cost.

avdudchenko:
> I encountered a similar issue in a flowsheet for REFLO. @avdudchenko did you consider adding the necessary steam properties to the property package (rather than this separate helper function)?

This is probably a good idea.. but outside the scope of this PR and would not work with all unit models:

- Crystalizer and Surrogate Crystalizer calculate heat duty and use that to estimate steam-cost 
- Steam heater does have "steam" flow of sort, where a prop pack could be used 
- Steam ejector might as well, but not 100% sure. 

I think we should have a steam prop pack, and we check if steam prop pack is available to compute steam mass, otherwise use the functions in this PR get relevant conversions... (ideally steam prop pack exposes conversions for use in computing steam props used for costs).

 This would be a different PR in scope though but @OOAmusat @ElmiraShamlou  should chime in, I think @ElmiraShamlou has brought the issue of WaterTAP property packages being quiet bad for steam/vapor/liquid modeling, and lack of steam prop package is an issue.

watertap/costing/tests/test_util.py
adam-a-a: Right, so in either of these cases, adding testing around this would help answer these questions. @kurbansitterley for example, adding a check that the expected exception to be thrown by `register_costing_parameter_block` is raised is one. And also explicit testing as @avdudchenko suggested.

@ElmiraShamlou can you add a couple of tests on this?

ElmiraShamlou:
The root issue with this PR is that steam is priced on a volumetric basis ($/m³), and every other problem follows from that single choice, including a fair amount of redundancy (notably _compute_steam_properties) that exists only to prop up the volumetric price unit:
The correct bases for steam pricing are `$/ton` or `$/GJ`. Specific volume varies by 5–10× with pressure, so using $/m³ bakes thermodynamics into the price unit and produces nonsensical economic behavior. 
For example, using a volumetric basis can incorrectly invert the cost per kg between LP and HP steam. At a fixed volumetric price, mass flow is converted via mass_flow × sp_vol. Low-pressure steam ends up more expensive per kg because each kg occupies more m³; high-pressure steam ends up cheaper per kg for the same reason. That is the opposite of industrial reality, where independently generated LP steam is cheaper per kg. Also, trying to compensate for this by tuning the $/m³ values per steam_type is non-intuitive and error-prone. 

I recommend following changes:
1) Change the default pricing basis from `$/m³` to `$/ton` or `$/GJ`; both are industry-standard and should be offered as independent options. Don't force conversion between them; price whatever the unit reports (mass flow to `$/ton`, heat duty to `$/GJ`). Optional conversion: heat-basis to mass-basis, if a unit reports heat duty but the user wants `$/ton` pricing.

2) Move steam properties to a Water property package. They will be more useful there to implement saturation conditions for thermal processes.

kurbansitterley:
> > I encountered a similar issue in a flowsheet for REFLO. @avdudchenko did you consider adding the necessary steam properties to the property package (rather than this separate helper function)?
> 
> This is probably a good idea.. but outside the scope of this PR and would not work with all unit models:
> 
> * Crystalizer and Surrogate Crystalizer calculate heat duty and use that to estimate steam-cost
> * Steam heater does have "steam" flow of sort, where a prop pack could be used
> * Steam ejector might as well, but not 100% sure.
> 
> I think we should have a steam prop pack, and we check if steam prop pack is available to compute steam mass, otherwise use the functions in this PR get relevant conversions... (ideally steam prop pack exposes conversions for use in computing steam props used for costs).
> 
> This would be a different PR in scope though but @OOAmusat @ElmiraShamlou should chime in, I think @ElmiraShamlou has brought the issue of WaterTAP property packages being quiet bad for steam/vapor/liquid modeling, and lack of steam prop package is an issue.

resurrected #1493 and added some props to the water property package, though it wouldn't fix the costing issue.

adam-a-a:
> > > I encountered a similar issue in a flowsheet for REFLO. @avdudchenko did you consider adding the necessary steam properties to the property package (rather than this separate helper function)?
> > 
> > 
> > This is probably a good idea.. but outside the scope of this PR and would not work with all unit models:
> > 
> > * Crystalizer and Surrogate Crystalizer calculate heat duty and use that to estimate steam-cost
> > * Steam heater does have "steam" flow of sort, where a prop pack could be used
> > * Steam ejector might as well, but not 100% sure.
> > 
> > I think we should have a steam prop pack, and we check if steam prop pack is available to compute steam mass, otherwise use the functions in this PR get relevant conversions... (ideally steam prop pack exposes conversions for use in computing steam props used for costs).
> > This would be a different PR in scope though but @OOAmusat @ElmiraShamlou should chime in, I think @ElmiraShamlou has brought the issue of WaterTAP property packages being quiet bad for steam/vapor/liquid modeling, and lack of steam prop package is an issue.
> 
> resurrected #1493 and added some props to the water property package, though it wouldn't fix the costing issue.

I wanted to remind everyone that when it comes to pure water and steam, IDAES has IAPWS (though I assume there'd be incompatibility issues to remedy preventing its immediate usage in WaterTAP). We already have the pure water property model that was originally made for MVC. So I think @kurbansitterley 's changes in #1493 should be enough. 

I don't think we should go to the extent of spinning out any new property models for this purpose if somehow #1493 isn't enough. If it came to that, I would think revisiting the IAPWS implementation would be the next step.

avdudchenko:
> > > I encountered a similar issue in a flowsheet for REFLO. @avdudchenko did you consider adding the necessary steam properties to the property package (rather than this separate helper function)?
> > 
> > 
> > This is probably a good idea.. but outside the scope of this PR and would not work with all unit models:
> > 
> > * Crystalizer and Surrogate Crystalizer calculate heat duty and use that to estimate steam-cost
> > * Steam heater does have "steam" flow of sort, where a prop pack could be used
> > * Steam ejector might as well, but not 100% sure.
> > 
> > I think we should have a steam prop pack, and we check if steam prop pack is available to compute steam mass, otherwise use the functions in this PR get relevant conversions... (ideally steam prop pack exposes conversions for use in computing steam props used for costs).
> > This would be a different PR in scope though but @OOAmusat @ElmiraShamlou should chime in, I think @ElmiraShamlou has brought the issue of WaterTAP property packages being quiet bad for steam/vapor/liquid modeling, and lack of steam prop package is an issue.
> 
> resurrected #1493 and added some props to the water property package, though it wouldn't fix the costing issue.

I updated the costing to now use mass basis. 

One key thing is that for some units, we need to convert heat duty to actual steam cost bass on mass and volume (for example users might want to specify steam pressure to get steam energy content and use explicit cost per mass basis. 

So I do not see how we get rid of the need to convert between volume and energy and mass of steam from different unit model costing methods and formulation, not all units take a steam a  prop pack simply to convert between heat duty and mass/energy content need.

adam-a-a:
- `surrogate_crystallizer` - doesn't have steam mass flow
- `crystallizer` - doesn't have steam mass flow
would need a remedy for these if using mass basis

avdudchenko:
> * `surrogate_crystallizer` - doesn't have steam mass flow
> 
>     * `crystallizer` - doesn't have steam mass flow
>       would need a remedy for these if using mass basis

Yeah, so a solution that might simply things is to cost team based on energy content as suggested by @ElmiraShamlou before as both simply use heat_duty to estimate mass or flow volume of steam for costing. 

My only concern is getting steam costs on kJ basis isntead of kg and providing simple comparison in other literature. Regardless I am partial to either per kg or kJ methods.

ElmiraShamlou:
> > * `surrogate_crystallizer` - doesn't have steam mass flow
> >   
> >   * `crystallizer` - doesn't have steam mass flow
> >     would need a remedy for these if using mass basis
> 
> Yeah, so a solution that might simply things is to cost team based on energy content as suggested by @ElmiraShamlou before as both simply use heat_duty to estimate mass or flow volume of steam for costing.
> 
> My only concern is getting steam costs on kJ basis isntead of kg and providing simple comparison in other literature. Regardless I am partial to either per kg or kJ methods.

One thing worth thinking about: when a unit model like `surrogate_crystallizer` or `crystallizer` doesn't actually have a steam flow, does it really make sense to cost it as steam? It might be cleaner to add a universal energy flow -- something like `energy`, `thermal_energy`, or `heat`, in the same spirit as `electricity`. So any unit that just needs to pay for energy can use it directly, without caring where the energy comes from. WaterTap is missing this, and I think it would be nicer to expose it as its own flow rather than keep tucking it inside steam.

avdudchenko:
> > > * `surrogate_crystallizer` - doesn't have steam mass flow
> > >   
> > >   * `crystallizer` - doesn't have steam mass flow
> > >     would need a remedy for these if using mass basis
> > 
> > 
> > Yeah, so a solution that might simply things is to cost team based on energy content as suggested by @ElmiraShamlou before as both simply use heat_duty to estimate mass or flow volume of steam for costing.
> > My only concern is getting steam costs on kJ basis isntead of kg and providing simple comparison in other literature. Regardless I am partial to either per kg or kJ methods.
> 
> One thing worth thinking about: when a unit model like `surrogate_crystallizer` or `crystallizer` doesn't actually have a steam flow, does it really make sense to cost it as steam? It might be cleaner to add a universal energy flow -- something like `energy`, `thermal_energy`, or `heat`, in the same spirit as `electricity`. So any unit that just needs to pay for energy can use it directly, without caring where the energy comes from. WaterTap is missing this, and I think it would be nicer to expose it as its own flow rather than keep tucking it inside steam.

I think the issue is we do not have a simple way of costing energy besides electricity on our units in WaterTAP costing (e.g. we hard code the energy source). 

I also don't think adding `thermal_energy` is the best way to go as its ambiguous (e.g. we might have future instances where units using different types of fuel end up using same energy cost value due to using `thermal_energy` )

Instead we need to be more explicit in energy naming creation for example, `surrogate` and `nacl crystallize` use 3 bar steam, so they should simply cost things based on 3 bar steam - and use the right cost basis for it.

I wounder if the solution to all of this to change the utility `cost_steam_flow ` to instead `cost_energy_flow ` which can be used on units to add energy cost and create shared costs based on actual unit needs (e.g. allow units to cost their energy based on kg or kJ or m3 basis or what ever. 

The utility would handle adding the cost to the costing block and ensure there is no collisions while letting units share costs for same energy types. 

This is probably a different PR as it would also absorb electricity into it as well. 

We just need to ensure units can accept names for energy costs in case a user wants to create different cost basis.

ksbeattie:
@avdudchenko, any news here?

ksbeattie:
@avdudchenko , @ElmiraShamlou, moving this to the Sept release, object if it shouldn't be.

ElmiraShamlou (approved):

adam-a-a:
We can merge as is now. In a subsequent PR, we can move towards something like $/kWh_thermal for heat costs while also maintaining ability to have different prices for different qualities of heat.

kurbansitterley (approved):
I am looking back on the comments in this PR now, and there was talk of adding steam properties to property models, which we did in #1493. I think those changes moot `cost_steam_flow` but it would be a mess to un-moot, so can do that in a future PR.

PR #1765, opened 2026-04-27
Add enthalpy to MCAS

adam-a-a:
## Summary/Motivation:
These changes were peeled off of PR #1691, originally to make MVC compatible with MCAS. This PR isolates the changes made in PR #1691 so that only the changes to MCAS are included. 

Additionally, some minor fixes (but not all inclusive) were made to calculations tied to total_dissolved_solids.

- add enthalpy 
- minor fixes to TDS related calculations

### Legal Acknowledgement

### DIFF
--- a/docs/how_to_guides/how_to_use_MCAS_property_model.rst
+++ b/docs/how_to_guides/how_to_use_MCAS_property_model.rst
@@ -95,8 +95,8 @@ A portion of the displayed output is shown below.
             None : 273.15 : 298.15 :  1000 :  True :  True : NonNegativeReals
         pressure : State pressure
             Size=1, Index=None, Units=Pa
-            Key  : Lower  : Value     : Upper    : Fixed : Stale : Domain
-            None : 1000.0 : 5000000.0 : 50000000 :  True :  True : NonNegativeReals
+            Key  : Lower    : Value     : Upper : Fixed : Stale : Domain
+            None : 100000.0 : 5000000.0 :  None :  True :  True : NonNegativeReals
         flow_mol_phase_comp : Component molar flow rate
             Size=3, Index=fs.properties.phase_list*fs.properties.component_list, Units=mol/s
             Key             : Lower : Value : Upper : Fixed : Stale : Domain
--- a/docs/how_to_guides/how_to_use_MCAS_property_model.rst
+++ b/docs/how_to_guides/how_to_use_MCAS_property_model.rst
@@ -96,7 +96,7 @@ A portion of the displayed output is shown below.
         pressure : State pressure
             Size=1, Index=None, Units=Pa
             Key  : Lower  : Value     : Upper : Fixed : Stale : Domain
-            None :    1.0 : 5000000.0 :  None :  True :  True : NonNegativeReals
+            None :     1  : 5000000.0 :  None :  True :  True : NonNegativeReals
         flow_mol_phase_comp : Component molar flow rate
             Size=3, Index=fs.properties.phase_list*fs.properties.component_list, Units=mol/s
             Key             : Lower : Value : Upper : Fixed : Stale : Domain
--- a/docs/how_to_guides/how_to_use_MCAS_property_model.rst
+++ b/docs/how_to_guides/how_to_use_MCAS_property_model.rst
@@ -96,7 +96,7 @@ A portion of the displayed output is shown below.
         pressure : State pressure
             Size=1, Index=None, Units=Pa
             Key  : Lower  : Value     : Upper : Fixed : Stale : Domain
-            None : 1000.0 : 5000000.0 :  None :  True :  True : NonNegativeReals
+            None :    1.0 : 5000000.0 :  None :  True :  True : NonNegativeReals
         flow_mol_phase_comp : Component molar flow rate
             Size=3, Index=fs.properties.phase_list*fs.properties.component_list, Units=mol/s
             Key             : Lower : Value : Upper : Fixed : Stale : Domain
--- a/docs/how_to_guides/how_to_use_MCAS_property_model.rst
+++ b/docs/how_to_guides/how_to_use_MCAS_property_model.rst
@@ -86,29 +86,29 @@ A portion of the displayed output is shown below.
 
 .. testoutput::
 
-   Block fs.state_block[0]
+    Block fs.state_block[0]
 
-     Variables:
-       temperature : State temperature
-           Size=1, Index=None, Units=K
-           Key  : Lower  : Value  : Upper  : Fixed : Stale : Domain
-           None : 273.15 : 298.15 : 373.15 :  True :  True : NonNegativeReals
-       pressure : State pressure
-           Size=1, Index=None, Units=Pa
-           Key  : Lower    : Value     : Upper : Fixed : Stale : Domain
-           None : 100000.0 : 5000000.0 :  None :  True :  True : NonNegativeReals
-       flow_mol_phase_comp : Component molar flow rate
-           Size=3, Index=fs.properties.phase_list*fs.properties.component_list, Units=mol/s
-           Key             : Lower : Value : Upper : Fixed : Stale : Domain
-           ('Liq', 'Cl_-') :     0 : 0.483 :  None :  True :  True : NonNegativeReals
-            ('Liq', 'H2O') :     0 :  53.8 :  None :  True :  True : NonNegativeReals
-           ('Liq', 'Na_+') :     0 : 0.483 :  None :  True :  True : NonNegativeReals
-       flow_mass_phase_comp : Component Mass flowrate
-           Size=3, Index=fs.properties.phase_list*fs.properties.component_list, Units=kg/s
-           Key             : Lower : Value                : Upper : Fixed : Stale : Domain
-           ('Liq', 'Cl_-') :     0 :             0.016905 :  None : False : False :  Reals
-            ('Liq', 'H2O') :     0 :   0.9683999999999999 :  None : False : False :  Reals
-           ('Liq', 'Na_+') :     0 : 0.011108999999999999 :  None : False : False :  Reals
+      Variables:
+        temperature : State temperature
+            Size=1, Index=None, Units=K
+            Key  : Lower  : Value  : Upper : Fixed : Stale : Domain
+            None : 273.15 : 298.15 :  1000 :  True :  True : NonNegativeReals
+        pressure : State pressure
+            Size=1, Index=None, Units=Pa
+            Key  : Lower  : Value     : Upper    : Fixed : Stale : Domain
+            None : 1000.0 : 5000000.0 : 50000000 :  True :  True : NonNegativeReals
+        flow_mol_phase_comp : Component molar flow rate
+            Size=3, Index=fs.properties.phase_list*fs.properties.component_list, Units=mol/s
+            Key             : Lower : Value : Upper : Fixed : Stale : Domain
+            ('Liq', 'Cl_-') :     0 : 0.483 :  None :  True :  True : NonNegativeReals
+             ('Liq', 'H2O') :     0 :  53.8 :  None :  True :  True : NonNegativeReals
+            ('Liq', 'Na_+') :     0 : 0.483 :  None :  True :  True : NonNegativeReals
+        flow_mass_phase_comp : Component Mass flowrate
+            Size=3, Index=fs.properties.phase_list*fs.properties.component_list, Units=kg/s
+            Key             : Lower : Value                : Upper : Fixed : Stale : Domain
+            ('Liq', 'Cl_-') :     0 :             0.016905 :  None : False : False :  Reals
+             ('Liq', 'H2O') :     0 :   0.9683999999999999 :  None : False : False :  Reals
+            ('Liq', 'Na_+') :     0 : 0.011108999999999999 :  None : False : False :  Reals
        ...
 
 The default material flow basis (i.e., state variable) for the MCAS property model is component molar flowrate. 
--- a/docs/how_to_guides/how_to_use_MCAS_property_model.rst
+++ b/docs/how_to_guides/how_to_use_MCAS_property_model.rst
@@ -95,8 +95,8 @@ A portion of the displayed output is shown below.
             None : 273.15 : 298.15 :  1000 :  True :  True : NonNegativeReals
         pressure : State pressure
             Size=1, Index=None, Units=Pa
-            Key  : Lower    : Value     : Upper : Fixed : Stale : Domain
-            None : 100000.0 : 5000000.0 :  None :  True :  True : NonNegativeReals
+            Key  : Lower  : Value     : Upper : Fixed : Stale : Domain
+            None : 1000.0 : 5000000.0 :  None :  True :  True : NonNegativeReals
         flow_mol_phase_comp : Component molar flow rate
             Size=3, Index=fs.properties.phase_list*fs.properties.component_list, Units=mol/s
             Key             : Lower : Value : Upper : Fixed : Stale : Domain
--- a/docs/how_to_guides/how_to_use_MCAS_property_model.rst
+++ b/docs/how_to_guides/how_to_use_MCAS_property_model.rst
@@ -95,8 +95,8 @@ A portion of the displayed output is shown below.
             None : 273.15 : 298.15 :  1000 :  True :  True : NonNegativeReals
         pressure : State pressure
             Size=1, Index=None, Units=Pa
-            Key  : Lower  : Value     : Upper : Fixed : Stale : Domain
-            None :     1  : 5000000.0 :  None :  True :  True : NonNegativeReals
+            Key  : Lower : Value     : Upper : Fixed : Stale : Domain
+            None :     1 : 5000000.0 :  None :  True :  True : NonNegativeReals
         flow_mol_phase_comp : Component molar flow rate
             Size=3, Index=fs.properties.phase_list*fs.properties.component_list, Units=mol/s
             Key             : Lower : Value : Upper : Fixed : Stale : Domain
--- a/docs/how_to_guides/how_to_use_MCAS_property_model.rst
+++ b/docs/how_to_guides/how_to_use_MCAS_property_model.rst
@@ -10,7 +10,7 @@ users to model the chemical and physical properties of simple systems without th
 
    # quiet idaes logs
    import idaes.logger as idaeslogger
-   idaeslogger.getLogger('ideas.core').setLevel('CRITICAL')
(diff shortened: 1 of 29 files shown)


watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -2371,12 +2357,15 @@ def _enth_mass_phase(self):
         def rule_enth_mass_phase(b, p):
             # temperature in degC, but pyunits in K
             t = b.temperature - 273.15 * pyunits.K
```
avdudchenko, line 2355: I assume this is sea water properties? 

I know for osmotic pressure, we have an option for SeaWater or NaCl. We should probably do the same here. 

The other thing that I have not looked at, is if these specific properties should be centralized and inherited by different property packages to remove duplicate code. 

We basically have same enthalpy, vapor pressure, and osmotic pressure equations in 2 different property packages - and possibly others (I never use anything beyond these three). 

It seems to me it would be useful to be able to pull in these properties from a centralized location and use them in different places. 

This would also provide a way for users to supply custom properties with out having to modify the property package directly. 

Basically I am thinking of a property_utility file that stores  rules to build different properties (starting with Enthalpy, Osmotic Pressure, Vapor Pressure) that are then used by the three property packages. This could also be set up to let a user supply a custom property based on a config options. 

I think osmotic pressure, vapor, and enthalpy are common things users (industrial users as well) will want to customize frequently based on feed water. So making it trivial to do via config in MCAS would make a lot of sense.

avdudchenko (approved):
A small suggestion for making MCAS a bit more flexible.  Although, that's probably a separate PR.

watertap/property_models/multicomp_aq_sol_prop_pack.py
adam-a-a: > I assume this is sea water properties?
> 
> I know for osmotic pressure, we have an option for SeaWater or NaCl. We should probably do the same here.

Yes, from the MIT thermodynamic seawater property library originally. Actually, I have only brought in seawater property relationships (for density, now enthalpy and pressure_sat), not NaCl--the reason for this being the relatively wide range of TDS that the MIT property library is meant to cover. Maybe something from NaCl is used in MCAS and I forgot, but that's what I recall without an in-depth check.

> It seems to me it would be useful to be able to pull in these properties from a centralized location and use them in different places.

I agree with that and have thought about something like that before (pulling common expressions for properties from a centralized location). I think it'd be good to discuss options before anyone were to spend time on that. For now, I plan to first get the minimal set of changes needed to render MCAS compatible with unit models that use enthalpy.

kurbansitterley: I agree. Many (most?) of our prop packs are identical with just a few additional properties tacked on.

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -2177,43 +2171,63 @@ def rule_total_hardness(b):
...
+        # def rule_total_hardness(b):
+        #     return b.total_hardness == pyunits.convert(
+        #         sum(
+        #             self.conc_mol_phase_comp["Liq", j]
+        #             * float(value(self.params.charge_comp[j]))
+        #             for j in polyvalent
+        #         )
+        #         * caco3_eq_weight,
+        #         to_units=pyunits.mg / pyunits.L,
+        #     )
```
adam-a-a: For the record:trying to refactor this method for improved readability--and also switching to conc_mol_phase_comp rather than flow_mol_phase_comp/flow_vol_phase in the constrraint--caused a lot of headaches with this PR. I believe that scaling of the property is adversely affected when making this switch and somehow becomes extremely problematic for NF DSPMDE flowsheet examples and/or tutorial. Didn't take the time to dig into differences in scaling to confirm, but this is my guess.

docs/how_to_guides/how_to_use_MCAS_property_model.rst
```diff
@@ -86,29 +86,29 @@ A portion of the displayed output is shown below.
...
+
+      Variables:
+        temperature : State temperature
+            Size=1, Index=None, Units=K
+            Key  : Lower  : Value  : Upper : Fixed : Stale : Domain
+            None : 273.15 : 298.15 :  1000 :  True :  True : NonNegativeReals
+        pressure : State pressure
+            Size=1, Index=None, Units=Pa
+            Key  : Lower  : Value     : Upper : Fixed : Stale : Domain
+            None :    1.0 : 5000000.0 :  None :  True :  True : NonNegativeReals
```
adam-a-a:
```suggestion
            None :    1   : 5000000.0 :  None :  True :  True : NonNegativeReals
```

watertap/unit_models/electrodialysis_bipolar_1D.py
```diff
@@ -1118,21 +1118,20 @@ def eq_get_total_areal_resistance_x(self, t, x):
...
-            doc="Calculate current density from the electrical input",
+            doc="Calculate total current generated via catalyst action",
         )
-        def eq_get_current_density(self, t, x):
-
-            @self.Constraint(
-                self.flowsheet().time,
-                self.diluate.length_domain,
-                doc="Calculate total current generated via catalyst action",
+        def eq_current_relationship(self, t, x):
```
adam-a-a, line 1123: Noticed a mistake in bipolar ED when tests failed due to MCAS changes. That's what these changes are for-- a naive, partial repair of BMED to get my checks passing on this PR.

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -2153,67 +2146,56 @@ def _total_hardness(self):
...
-        try:
-            total_dissolved_solids_temp = pyunits.convert(
-                sum(self.conc_mass_phase_comp["Liq", j] for j in self.params.ion_set),
-                to_units=pyunits.mg / pyunits.L,
-            )
 
-            def rule_total_dissolved_solids(b):
-                return b.total_dissolved_solids == total_dissolved_solids_temp
+        total_dissolved_solids_temp = pyunits.convert(
+            sum(self.conc_mass_phase_comp["Liq", j] for j in self.params.solute_set),
```
adam-a-a, line 2191: Updated TDS to sum over solute_set instead of only ion_set

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -2153,67 +2146,56 @@ def _total_hardness(self):
             domain=NonNegativeReals,
             bounds=(0, None),
             units=pyunits.mg / pyunits.L,
-            doc="total hardness as CaCO3",
+            doc="Total hardness as CaCO3 equivalent",
```
adam-a-a, line 2149: Refactored total_hardness simply to check for the presence of polyvalent cations, which are needed to compute total hardness.

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -1463,9 +1450,15 @@ def rule_flow_mass_phase_comp(b, p, j):
             )
 
     def _flow_mass_comp(self):
-        @self.Expression(self.params.component_list)
-        def flow_mass_comp(b, j):
-            return b.flow_mass_phase_comp["Liq", j]
+        add_object_reference(
+            self,
+            "flow_mass_comp",
```
adam-a-a, line 1455: will need flow_mass_comp for units related to MVC from what I recall (i.e., needed later if using MCAS with MVC)

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -1376,15 +1363,15 @@ def build(self):
...
             domain=NonNegativeReals,
             units=pyunits.K,
             doc="State temperature",
         )
 
         self.pressure = Var(
             initialize=101325,
-            bounds=(1e5, None),
+            bounds=(1, None),
             domain=NonNegativeReals,
```
adam-a-a, line 1375: Updating bounds to enable enthalpy calculations and lower pressure values expected in MVC

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
```diff
@@ -1701,17 +1700,7 @@ def test_compatibility_with_mixer():
         charge={"Na_+": 1, "Cl_-": -1},
     )
 
-    m.fs.mixer1 = Mixer(
-        property_package=m.fs.properties, energy_mixing_type=MixingType.none
-    )
-
-    with pytest.raises(
-        NotImplementedError,
-        match="property package has not implemented the get_enthalpy_flow_terms method. Please contact the property package developer.",
```
adam-a-a, line 1710: No longer relevant since MCAS now calculates enthalpy

watertap/flowsheets/nf_dspmde/nf_with_bypass.py
```diff
@@ -157,7 +157,7 @@ def fix_init_vars(m):
     # fix initial guess for splitter
     m.fs.by_pass_splitter.split_fraction[0, "bypass"].fix(0.9)
     m.fs.by_pass_splitter.split_fraction[0, "bypass"].setlb(0.05)
-    m.fs.by_pass_splitter.split_fraction[0, "bypass"].setub(None)
+    m.fs.by_pass_splitter.split_fraction[0, "bypass"].setub(1.001)
```
adam-a-a, line 160: Added this bound to improve stability. Associated test failed after MCAS updates, and this remedied the issue

watertap/flowsheets/nf_dspmde/nf.py
```diff
@@ -257,7 +256,7 @@ def fix_init_vars(m):
     m.fs.NF.nfUnit.mixed_permeate[0].pressure.fix(101325)
     # NF membrane props for NF270
     m.fs.NF.nfUnit.radius_pore.fix(0.5e-9)
-    m.fs.NF.nfUnit.membrane_thickness_effective.fix(8.598945196055952e-07)
+    m.fs.NF.nfUnit.membrane_thickness_effective.fix(8.5989e-07)
```
adam-a-a, line 259: While troubleshooting test failure, I figured it'd be better not to have so many decimal places, in case it contributed to instability in any way

watertap/flowsheets/nf_dspmde/nf.py
```diff
@@ -279,6 +278,7 @@ def unfix_opt_vars(m):
     # Touch total_hardness (on-demand property) at feed and disposal for reporting
     m.fs.feed.properties[0].total_hardness
     m.fs.disposal.properties[0].total_hardness
+    iscale.calculate_scaling_factors(m)
```
adam-a-a, line 281: Rescaling after touching total_hardness

docs/how_to_guides/how_to_use_MCAS_property_model.rst
```diff
@@ -86,29 +86,29 @@ A portion of the displayed output is shown below.
...
+
+      Variables:
+        temperature : State temperature
+            Size=1, Index=None, Units=K
+            Key  : Lower  : Value  : Upper : Fixed : Stale : Domain
+            None : 273.15 : 298.15 :  1000 :  True :  True : NonNegativeReals
+        pressure : State pressure
+            Size=1, Index=None, Units=Pa
+            Key  : Lower : Value     : Upper : Fixed : Stale : Domain
+            None :     1 : 5000000.0 :  None :  True :  True : NonNegativeReals
```
adam-a-a, line 99: bounds were changed in MCAS for pressure and temperature, so the corresponding how-to, where output is checked, had to be updated correspondingly

docs/how_to_guides/how_to_use_MCAS_property_model.rst
```diff
@@ -10,7 +10,7 @@ users to model the chemical and physical properties of simple systems without th
 
    # quiet idaes logs
    import idaes.logger as idaeslogger
-   idaeslogger.getLogger('ideas.core').setLevel('CRITICAL')
+   idaeslogger.getLogger('idaes.core').setLevel('CRITICAL')
```
adam-a-a, line 13: typo in idaes fixed

avdudchenko (approved):
My primary concern of pressure was addressed. LGMT

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -201,18 +200,6 @@ def variable_scaling_routine(
...
-                if j == "H2O":
-                    self.set_variable_scaling_factor(v, 1, overwrite=overwrite)
-                else:
-                    sf = self.get_scaling_factor(
-                        model.flow_mass_phase_comp[p, j]
-                    ) / self.get_scaling_factor(
-                        model.flow_mass_phase_comp["Liq", "H2O"]
-                    )
-                    self.set_variable_scaling_factor(v, sf, overwrite=overwrite)
-
```
dallan-keylogic, line 215: Is the idea that the user will set this scaling factor manually?

kurbansitterley (approved):
LGTM. I have a comment in a dependent PR but this one is 👍🏻

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -2153,67 +2146,56 @@ def _total_hardness(self):
...
-            def rule_total_hardness(b):
-                return b.total_hardness == total_hardness_temp
-
-            self.eq_total_hardness = Constraint(rule=rule_total_hardness)
-
-        except InconsistentUnitsError:
+        polyvalent = [
+            j for j in self.params.cation_set if value(self.params.charge_comp[j]) >= 2
+        ]
+        if not polyvalent:
```
dallan-keylogic, line 2154:
```suggestion
        if len(polyvalent) == 0:
```
Personally, I'm not fond of implicitly casting things like lists to booleans. I think explicitly checking the list's length makes code more readable.

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -2153,67 +2146,56 @@ def _total_hardness(self):
...
+                for j in self.params.cation_set
+                if value(self.charge_comp[j]) > 1
+            ),
+            to_units=pyunits.mg / pyunits.L,
+        )
+
+        def rule_total_hardness(b):
+            return b.total_hardness == total_hardness_temp
+
+        self.eq_total_hardness = Constraint(rule=rule_total_hardness)
```
dallan-keylogic, line 2179: Is there a way to move the property calculation entirely into `rule_total_hardness`. I've run into some obscure bugs when using non-local variables in Pyomo `Constraint` rules. (Any necessary nonlocal variables can be passed to the constraint using `functools.partial`.)

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -2153,67 +2146,56 @@ def _total_hardness(self):
...
+        total_dissolved_solids_temp = pyunits.convert(
+            sum(self.conc_mass_phase_comp["Liq", j] for j in self.params.solute_set),
+            to_units=pyunits.mg / pyunits.L,
+        )
 
-            self.eq_total_dissolved_solids = Constraint(
-                rule=rule_total_dissolved_solids
-            )
+        def rule_total_dissolved_solids(b):
+            return b.total_dissolved_solids == total_dissolved_solids_temp
```
dallan-keylogic, line 2196: Nonlocal variable in constraint.

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -2371,12 +2353,7 @@ def _enth_mass_phase(self):
...
             t = b.temperature - 273.15 * pyunits.K
-            S_kg_kg = (
-                pyunits.convert(
-                    b.total_dissolved_solids, to_units=pyunits.kg / pyunits.m**3
-                )
-                / b.dens_mass_phase[p]
-            )
+            S_kg_kg = sum(b.mass_frac_phase_comp[p, j] for j in b.params.solute_set)
             S_g_kg = S_kg_kg * 1000
             P = b.pressure - 101325 * pyunits.Pa
```
dallan-keylogic, line 2358: Can you create a `pressure_ref` `Param` to use here instead?

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -2371,12 +2353,7 @@ def _enth_mass_phase(self):
         def rule_enth_mass_phase(b, p):
             # temperature in degC, but pyunits in K
             t = b.temperature - 273.15 * pyunits.K
```
dallan-keylogic, line 2355: Could you create a `temperature_ref` `Param` to use here?

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -2535,10 +2507,9 @@ def get_material_flow_terms(self, p, j):
         elif self.params.config.material_flow_basis == MaterialFlowBasis.mass:
             return self.flow_mass_phase_comp[p, j]
 
-    # TODO: add enthalpy terms later
-    # def get_enthalpy_flow_terms(self, p):
-    #     """Create enthalpy flow terms."""
-    #     return self.enth_flow
+    def get_enthalpy_flow_terms(self, p):
+        """Create enthalpy flow terms."""
+        return self.enth_flow
```
dallan-keylogic, line 2512: Could you create a phase-indexed version of `enth_flow` to use here instead?

dallan-keylogic (commented):
Some minor comments.

I am very cautious about using nonlocal variables in constraint rules because it ended up causing a very obscure bug as part of the SOC dynamics work. The value of an outside variable is not frozen when a function is created, even for immutable types like strings. So if you do something like
```
for p in ["Liq", "Vap"]:
    def rule_density_phase(b):
        return b.dens_phase[p] == sum(b.conc_phase_comp[p, j] for j in b.component_set)
    self.add_component(
        "eq_dens_" + p,
        Constraint(rule=rule_dens_phase)
    )
```
the value of `p` inside the rule changes as its value in the outside context changes. 

Normally, this isn't an issue, but if the constraint rule ever has to be called again, then the resulting constraint will be incorrect. One place where we end up calling a constraint rule additional times after creating it is when discretizing a `ContinuousSet` using PyomoDAE. If the discretization method needs to create additional time points or length points (such as when discretizing using collocation), then the rule will get called again at those new points.  So we would have `properties[x_new].eq_dens_Liq` containing the expression `b.dens_phase["Vap"] == sum(b.conc_phase_comp["Vap", j] for j in b.component_set)`.

If this isn't clear, we can set up a meeting to talk about it.

Issue #1754, opened 2026-04-07
Streamline Connectivity and Flexibility


MarcusHolly:
### Description

Complete the following actions to simplify the process of making custom models and flowsheets:

- [ ] Add a compatibility table mapping unit models to compatible property models
- [ ] Decouple costing functions from flowsheet build functions in all flowsheets #1768 
- [ ] Add configuration option to toggle between 0D and 1D unit models where applicable
- [ ] Create an in-depth tutorial on building a custom unit model

### Motivation

These developments were identified by stakeholders in the WaterTAP Academy. See https://app.box.com/file/2124448237821 for the full report on these roadmapping efforts.

PR #1768, opened 2026-04-30
Decouple Costing from Build Functions

MarcusHolly:
## Fixes/Resolves:

2nd bullet point in #1754

## Summary/Motivation:
Isolate out the costing models so that users can easily swap it out for an alternative costing method.

- Creates stand-alone costing functions
- Test results are expected to remain the same

### Legal Acknowledgement

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -60,6 +60,7 @@ environment*.yml @sufikaur
 /watertap/unit_models/cstr.py @adam-a-a @MarcusHolly
 /watertap/unit_models/cstr_injection.py @adam-a-a @MarcusHolly
 /watertap/unit_models/dewatering.py @adam-a-a @agarciadiego
+/watertap/unit_models/electrocoagulation.py @kurbansitterley @MuktaHardikar
 /watertap/unit_models/electrodialysis_0D.py @lbibl
 /watertap/unit_models/electrodialysis_1D.py @lbibl
 /watertap/unit_models/electrolyzer.py @adam-a-a @lbibl
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -21,7 +21,7 @@ Create a Conda environment (in this example, named ``watertap``) where WaterTAP 
 
 .. code-block:: shell
 
-   conda create --name watertap --yes python=3.11
+   conda create --name watertap --yes -c conda-forge python=3.11
 
 Activate the ``watertap`` environment using the command given below. If the environment was activated successfully, the environment's name will be displayed in the terminal prompt such as ``(watertap) project-directory $``.
 
@@ -58,7 +58,7 @@ Create a Conda environment (in this example, named ``watertap``) where WaterTAP 
 
 .. code-block:: shell
 
-   conda create --name watertap --yes python=3.11
+   conda create --name watertap --yes -c conda-forge python=3.11
 
 Activate the ``watertap`` environment using the command given below. If the environment was activated successfully, the environment's name will be displayed in the terminal prompt such as ``(watertap) project-directory $``.
 
@@ -139,11 +139,11 @@ For WaterTAP developers
 
 This section is for developers who plan to modify or contribute to WaterTAP's codebase. Typically, *contributing to WaterTAP* will involve opening a Pull Request (PR) in WaterTAP's repository. For more information, refer to :ref:`developer-guide`.
 
-Create a Conda environment (in this example, named ``watertap-dev``) where WaterTAP and all dependendencies needed for development will be installed, then activate it:
+Create a Conda environment (in this example, named ``watertap-dev``) where WaterTAP and all development dependencies will be installed, then activate it:
 
 .. code-block:: shell
 
-   conda create --name watertap-dev --yes python=3.11 && conda activate watertap-dev
+   conda create --name watertap-dev --yes -c conda-forge python=3.11 cyipopt reaktoro && conda activate watertap-dev
 
 Clone the WaterTAP repository to your local development machine using ``git clone``, then enter the newly created ``watertap`` subdirectory:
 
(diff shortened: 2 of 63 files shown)


MarcusHolly:
Not sure how to go about resolving the Python 3.11 Linux failures...

dallan-keylogic (approved):
LGTM

kurbansitterley (commented):
do we want to enforce a naming convention of ``add_costing``? I see an ``add_costs`` in the MVC flowsheet. 

Otherwise,  LGTM and I will approve.

MarcusHolly:
> do we want to enforce a naming convention of `add_costing`? I see an `add_costs` in the MVC flowsheet.
> 
> Otherwise, LGTM and I will approve.

I think we should try to stick with the naming convention unless there's a specific reason not to.

kurbansitterley (approved):

adam-a-a (commented):
Is there a tradeoff to incorporating add_costing outside of build functions? I could be mistaken, but it looks like add_costing is still called from within build functions. I was originally thinking we'd call add_costing outside and after build calls.

I will also note that I think this will serve as a stepping stone for when/if we refactor all of our flowsheet models to take the same form as unit model blocks, where we set these things through config options (e.g., choice of costing package). PR #1685 is one example approach although there are relevant changes to be made in that example. Not for this PR though

adam-a-a (changes requested):
holding this PR until placement of add_costing is discussed/resolved

adam-a-a:
The main reason I can think of for separating add_costing from the build function completely:
In select cases, it is possible that model convergence can be improved/debugged more easily when separating the solve attempt without and subsequently with costing.

However, this may complicate things for the GUI or other toolboxes that may expect a model with costing to be returned.

MarcusHolly:
> Is there a tradeoff to incorporating add_costing outside of build functions? I could be mistaken, but it looks like add_costing is still called from within build functions. I was originally thinking we'd call add_costing outside and after build calls.
> 
> I will also note that I think this will serve as a stepping stone for when/if we refactor all of our flowsheet models to take the same form as unit model blocks, where we set these things through config options (e.g., choice of costing package). PR #1685 is one example approach although there are relevant changes to be made in that example. Not for this PR though

@adam-a-a You are right, add_costing is still called from within the build functions. If you look at some earlier commits, I had actually started out by calling add_costing outside of the build function. In most cases, it worked fine. But in some cases (forgetting off the top of my head), flowsheets/tests were unable to solve if costing was created after the build. It's possible this could have been resolved with additional flowsheet reformulation but rather than making more changes to the flowsheets than necessary, I decided to call add_costing from within the build to preserve the original pecking order as closely as possible. 

As you mentioned in your most recent comment, entirely separating add_costing from the build function would require updating all of the GUI files. I didn't run into any GUI-related issues when doing so, but I will admit that I did not thoroughly investigate this. 

In summary, I went with this approach to change as little as possible about the flowsheets/GUIs while still meeting the spirit of decoupling costing from the build. And most notably, this approach did not run into any of the earlier failures related to rearranging the order of operations.

adam-a-a:
> > Is there a tradeoff to incorporating add_costing outside of build functions? I could be mistaken, but it looks like add_costing is still called from within build functions. I was originally thinking we'd call add_costing outside and after build calls.
> > I will also note that I think this will serve as a stepping stone for when/if we refactor all of our flowsheet models to take the same form as unit model blocks, where we set these things through config options (e.g., choice of costing package). PR #1685 is one example approach although there are relevant changes to be made in that example. Not for this PR though
> 
> @adam-a-a You are right, add_costing is still called from within the build functions. If you look at some earlier commits, I had actually started out by calling add_costing outside of the build function. In most cases, it worked fine. But in some cases (forgetting off the top of my head), flowsheets/tests were unable to solve if costing was created after the build. It's possible this could have been resolved with additional flowsheet reformulation but rather than making more changes to the flowsheets than necessary, I decided to call add_costing from within the build to preserve the original pecking order as closely as possible.
> 
> As you mentioned in your most recent comment, entirely separating add_costing from the build function would require updating all of the GUI files. I didn't run into any GUI-related issues when doing so, but I will admit that I did not thoroughly investigate this.
> 
> In summary, I went with this approach to change as little as possible about the flowsheets/GUIs while still meeting the spirit of decoupling costing from the build. And most notably, this approach did not run into any of the earlier failures related to rearranging the order of operations.

Thanks for the quick response, @MarcusHolly. On that note, can you double-check that m.fs.costing.initialize() is invoked in all of these modules? Looks like it isn't at least for the crystallizer example. 

That could be one potential reason why separating costing might have previously been problematic in select cases. Whatever the case, I think it is good practice to ensure the costing block is initialized, and we'd want developers to realize this if ever they were looking upon existing flowsheet examples to implement costing in a new flowsheet.

Given your response, I think we can settle with your implementation here and merge after ensuring initialization of costing blocks.

MarcusHolly:
> > > Is there a tradeoff to incorporating add_costing outside of build functions? I could be mistaken, but it looks like add_costing is still called from within build functions. I was originally thinking we'd call add_costing outside and after build calls.
> > > I will also note that I think this will serve as a stepping stone for when/if we refactor all of our flowsheet models to take the same form as unit model blocks, where we set these things through config options (e.g., choice of costing package). PR #1685 is one example approach although there are relevant changes to be made in that example. Not for this PR though
> > 
> > 
> > @adam-a-a You are right, add_costing is still called from within the build functions. If you look at some earlier commits, I had actually started out by calling add_costing outside of the build function. In most cases, it worked fine. But in some cases (forgetting off the top of my head), flowsheets/tests were unable to solve if costing was created after the build. It's possible this could have been resolved with additional flowsheet reformulation but rather than making more changes to the flowsheets than necessary, I decided to call add_costing from within the build to preserve the original pecking order as closely as possible.
> > As you mentioned in your most recent comment, entirely separating add_costing from the build function would require updating all of the GUI files. I didn't run into any GUI-related issues when doing so, but I will admit that I did not thoroughly investigate this.
> > In summary, I went with this approach to change as little as possible about the flowsheets/GUIs while still meeting the spirit of decoupling costing from the build. And most notably, this approach did not run into any of the earlier failures related to rearranging the order of operations.
> 
> Thanks for the quick response, @MarcusHolly. On that note, can you double-check that m.fs.costing.initialize() is invoked in all of these modules? Looks like it isn't at least for the crystallizer example.
> 
> That could be one potential reason why separating costing might have previously been problematic in select cases. Whatever the case, I think it is good practice to ensure the costing block is initialized, and we'd want developers to realize this if ever they were looking upon existing flowsheet examples to implement costing in a new flowsheet.
> 
> Given your response, I think we can settle with your implementation here and merge after ensuring initialization of costing blocks.

I will double check this now. For the crystallizer example, the costing was not initialized in the original version (the one in main), but I will add it now and look for other examples where it may be missing.

adam-a-a (approved):
LGTM

PR #1771, opened 2026-05-05
Mcas evaporator check

adam-a-a:
## Summary/Motivation:
Test MCAS with enthalpy against evaporator. This breaks down the attempt to make MCAS compatible with MVC in PR #1691 

- add test for evaporator with mcas 

### Legal Acknowledgement

### DIFF
--- a/watertap/property_models/multicomp_aq_sol_prop_pack.py
+++ b/watertap/property_models/multicomp_aq_sol_prop_pack.py
@@ -11,10 +11,10 @@
 #################################################################################
 """
 This property package computes a multi-component aqueous solution that can
-contain ionic and/or neutral solute species. It supports basic calculation 
-of component quantities and some physical, chemical and electrical properties. 
+contain ionic and/or neutral solute species. It supports basic calculation
+of component quantities and some physical, chemical and electrical properties.
 
-This property package was formerly named the "ion_DSPMDE_prop_pack" and was originally 
+This property package was formerly named the "ion_DSPMDE_prop_pack" and was originally
 designed for use with the Donnan Steric Pore Model with Dielectric Exclusion (DSPMDE) for
 nanofiltration.
 """
@@ -1835,15 +1835,11 @@ class MCASStateBlockData(StateBlockData):
                 == ElectricalMobilityCalculation.none
             ):
                 if (p, j) not in self.params.config.elec_mobility_data.keys():
-                    raise ConfigurationError(
-                        """ 
+                    raise ConfigurationError(""" 
                         Missing the "elec_mobility_data" configuration to build the elec_mobility_phase_comp 
                         and/or its derived variables for {} in {}. 
                         Provide this configuration or use ElectricalMobilityCalculation.EinsteinRelation.
-                        """.format(
-                            j, self.name
-                        )
-                    )
+                        """.format(j, self.name))
                 else:
                     return (
                         b.elec_mobility_phase_comp[p, j]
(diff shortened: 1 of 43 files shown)


adam-a-a:
This will just be one file change once PR #1765 is merged.

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -1376,15 +1364,15 @@ def build(self):
...
+            bounds=(273.15, 1000),
             domain=NonNegativeReals,
             units=pyunits.K,
             doc="State temperature",
         )
 
         self.pressure = Var(
             initialize=101325,
-            bounds=(1e5, None),
+            bounds=(1e3, None),
```
avdudchenko: really worried this is a bit to close to lower bound. I am pretty sure i seen reaktoro predict vapor pressures at 1e3 scale...I wounder if we should put this to 1e2 as lower bound (or 0?)

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -2153,10 +2147,10 @@ def _total_hardness(self):
             domain=NonNegativeReals,
             bounds=(0, None),
             units=pyunits.mg / pyunits.L,
-            doc="total hardness as CaCO3",
+            doc="Total hardness as CaCO3 equivalent",
         )
         # add try/except to handle case without multivalent cations,
-        # which would return 0 and result in Inconsitentunits error due to conversion of dimensionless to mg/L
+        # which would return 0 and result in InconsistentUnits error due to conversion of dimensionless to mg/L
         try:
```
avdudchenko: I wounder if we should simply do an if check if there are divalent ions or not, instead of relying on a try/except to build this. My concern if the constraints gets partially build and then it total hardness gets fixed. Sort of a really odd edge case, but never know how things change in the future and and an explicit if check would be reliable.

avdudchenko (approved):
Minor edit requests (Only key one is pressure bounds on sat pressure... really concerned this will cause issues in processes where pressure does drop very low and its easier to change a lower bound specific unit, rather then changing it through out a flowsheet...

watertap/property_models/multicomp_aq_sol_prop_pack.py
adam-a-a: Good point--I will try to address this on PR #1765. I ended up reverting my changes because stability issues came up, but I don't think they were caused by the check for polyvalent ions, so I can bring that back in.

watertap/property_models/multicomp_aq_sol_prop_pack.py
adam-a-a: OK, I will reduce further in PR #1765 and make sure nothing breaks. Note--this PR builds off of #1765 , so the only relevant changed file in this PR will end up being the test file for the evaporator. Please let me if you think we should go the extra mile on testing evap with MCAS. For now, I started with the simplest case: sodium and chloride ions only. Anys suggestions on a test composition beyond that would be welcome.

avdudchenko: Ahh, make sense. Thanks!  I think useing Na Cl is fine, we just need to check that MCAS in mole flow basis does not break the evaporator as it uses mass basis to do initialization routines.

watertap/property_models/multicomp_aq_sol_prop_pack.py
avdudchenko: Sounds good - looks like my suggestion caused issues - my bad haha.

watertap/property_models/multicomp_aq_sol_prop_pack.py
adam-a-a: THis has now been addressed in #1765 (reduced lower bound to 1 Pa)

tristantc:
LGTM

tristantc (approved):
LGTM

PR #1773, opened 2026-05-06
Davies activity coefficient fix

adam-a-a:
## Summary/Motivation:
Fix activity coefficient calculation under Davies model. Currently, the Debye Huckel constant, A, is underestimated, leading to activity coefficients close to 1 and resulting in not much difference from assuming ideality.

These changes will alter rejection and flux predictions in NF DSPMDE when using Davies as the activity coefficient model. Other models that use the Davies config should also be impacted.

- incorporate sqrt(density_solvent) into the debye_huckel_constant calculation
- change default value for dielectric constant parameter
- revise activity coefficient constraint to be on log10 basis instead of natural log
- update associated tests

### Legal Acknowledgement


avdudchenko (approved):
I wanted to levave comments in file, but git hub gave me errors. 

Looks good (Assuming it passeses test). There is some newly added commented code in test files for NF model that should be cleaned up, and some print lines (like line 331 in test_nf_dspmde) file)

watertap/property_models/multicomp_aq_sol_prop_pack.py
```diff
@@ -1916,10 +1917,10 @@ def rule_act_coeff_phase_comp(b, p, j):
                 == ActivityCoefficientModel.davies
             ):
                 I = b.ionic_strength_molal
-                return log(
+                return log10(
                     b.act_coeff_phase_comp[p, j]
                 ) == -b.debye_huckel_constant * b.charge_comp[j] ** 2 * (
-                    I**0.5 / (1 * pyunits.mole**0.5 / pyunits.kg**0.5 + I**0.5)
+                    I**0.5 / (1 + pyunits.kg**0.5 * pyunits.mol**-0.5 * I**0.5)
                     - b.params.debye_huckel_b * I
```
MuktaHardikar, line 1924: Just double checking with https://reaktoro.org/api/PageActivityModelDebyeHuckel.html, should it `+b.params.debye_huckel_b * I`?

adam-a-a: The denominator needs to work out to be dimensionless. I spent quite a bit of time trying to figure this out. It is one of those equations where units don't work as cleanly as usual (e.g., like empirical formulae that don't always abide by units fully). Thus, the way it is implemented here, the units I added will cancel out the units on molal ionic strength. 

I will double check the signage on the `b*I` term.

adam-a-a: Thanks for bringing this up @MuktaHardikar. The link you provided is useful but for Debye Huckel activity model. What I implement here is for the Davies activity model, which has this corresponding ref verifying the `-bI` term: https://reaktoro.org/api/PageActivityModelDavies.html

adam-a-a: Actually, both are equivalent in the sense that the negative sign outside of the parentheses gets distributed to `-bI` which would result in `+bI`. The reference you have shows the terms factored out, while my ref doesn't.

MuktaHardikar:
LGTM! Need to fix the tests failing in the NF test. A notebook test is failing too.

MuktaHardikar (approved):
LGTM!

watertap/unit_models/tests/test_nanofiltration_DSPMDE_0D.py
```diff
@@ -1801,6 +1789,10 @@ def test_pressure_recovery_step_2_ions():
     assert_optimal_termination(results)
 
     pressure_steps = np.linspace(1.8e5, 20e5, 10)
+    from watertap.core.util.model_diagnostics.infeasible import (
```
kurbansitterley: do we want to keep this here?

watertap/unit_models/tests/test_nanofiltration_DSPMDE_0D.py
```diff
@@ -62,7 +61,9 @@
 from idaes.models.unit_models import (
     Feed,
 )
-import idaes.logger as idaeslog
+
+# Uncomment the following line to enable logging during testing (e.g., initialization_tester(m, outlvl=idaeslog.DEBUG))
```
kurbansitterley: kept here intentionally?

tutorials/modeling_multicomponent_rejection.ipynb
```diff
@@ -1028,12 +1028,12 @@
     "    expr=m.fs.unit.mixed_permeate[0].total_hardness <= m.fs.max_hardness\n",
     ")\n",
     "# Scale the new constraint\n",
-    "set_scaling_factor(\n",
-    "    m.fs.unit.mixed_permeate[0].total_hardness,\n",
-    "    1 / value(m.fs.unit.mixed_permeate[0].total_hardness),\n",
-    ")\n",
-    "set_scaling_factor(m.fs.max_hardness, 1 / value(m.fs.max_hardness))\n",
-    "constraint_scaling_transform(m.fs.max_hardness_constraint, 1 / value(m.fs.max_hardness))"
+    "# set_scaling_factor(\n",
```
kurbansitterley: uncomment?

watertap/property_models/tests/test_multicomp_aq_sol_prop_pack.py
```diff
@@ -2072,7 +2119,7 @@ def test_seawater_data_with_flow_mass_basis():
         0.07820, rel=1e-3
     )
 
-    assert value(stream[0].debye_huckel_constant) == pytest.approx(0.01554, rel=1e-3)
+    assert value(stream[0].debye_huckel_constant) == pytest.approx(0.510, rel=5e-3)
```
kurbansitterley, line 2122: rel value intentional (it is fine with me)?

tutorials/modeling_multicomponent_rejection.ipynb
MuktaHardikar: Removing these scaling steps resolved the test fail for the notebook.

watertap/unit_models/tests/test_nanofiltration_DSPMDE_0D.py
MuktaHardikar: This can be deleted. There were some print statements in the test file that were removed so this is also a redundant import.

kurbansitterley (approved):
LGTM

Issue #1780, opened 2026-05-17
Adding References and Updating ZO Models


kurbansitterley:
I have been going through all of the ZO models to determine the bounds for the costing relationships. In the process found many errors or inconsistencies, or otherwise just general issues with some of the parameters in the yamls.

This issue is to track the changes made to these ZO unit models. I have already made the necessary changes in a PR on my fork (https://github.com/kurbansitterley/watertap/pull/15) and will be adding PRs to main. I focused on the models that I was most familiar with, which were those that originated with WaterTAP3 and not AMO models.

The changes made to ZO models fall into the following categories:
- simply adding references to the yaml
- modifying costing parameters. This could be updating the costing year to match the reference or changing the parameter values themselves. I made these changes when I could not track down the reference for the existing parameters, but I found a regression/data/reference for new parameters. 
- adding bounds to ZO models to ensure they fell in the limits of the costing method

I was able to find references for the following ZO models:
- aeration_basin #1792 
- air_flotation #1782 
- anaerobic_digestion_oxidation #1792 
- backwash solids handling #1792 
- bio active filtration #1792 
- bioreactor #1792 
- cartridge filtration #1792 
- chemical addition #1783 
- chlorination #1792 
- clarifier #1792 
- coag and floc  #1791 
- cooling supply #1792 
- decarbonator #1792 
- deep well injection #1784 
- dissolved air flotation #1792 
- dual media filtration #1792 
- deep well injection #1784 
- electrocoagulation #1792 
- electrodialysis reversal #1792 
- evaporation pond #1785 
- filter press #1786 
- fixed bed #1787 
- ion exchange #1792 
- iron and manganese removal #1789 
- media filtration #1792 
- microfiltration #1792 
- microscreen filtration #1793 
- muncipal drinking #1792 
- nanofiltration #1792 
- ozonation #1790 
- ozone aop #1790 
- primary separator #1792 
- pump #1792 
- screen #1793 
- sedimentation #1792 
- settling pond #1792 
- static mixer #1792 
- storage tank #1792 
- surface discharge #1792 
- sw onshore intake #1792 
- tri media filtration #1792 
- ultra filtration #1792 
- uv #1792 
- walnut shell filter #1792 
- water pumping station #1792 
- well field #1792 

In a few cases, I made minor changes to the unit model themselves, which will be detailed in the individual PRs.

[...]

PR #1782, opened 2026-05-17
Remove Air Flotation ZO

kurbansitterley:
## Fixes/Resolves:

#1780 

## Summary/Motivation:

The costing relationship for air_flotation_zo is a duplicate of the raw water pumping station ZO model. Without a suitable alternative, this model should be removed because it does not represent an air flotation unit.

- Remove everything related to AirFlotationZO from the repo
-

### Legal Acknowledgement


MarcusHolly (approved):

watertap/data/techno_economic/air_flotation.yaml
```diff
@@ -5,12 +5,13 @@ default:
...
+    reference: Fit to data from McGivney & Kawamura Figure 5.5.36b
     reference_state:
-      value: 15772.5491
+      value: 15772.5491 # 100 MGD
       units: m^3/hr
     capital_a_parameter:
-      value: 4.488923561e6
-      units: USD_2015
+      value: 4.488923561e6 # 24348.0 * 100 * 3.14 (TPEC)
+      units: USD_2014
```
adam-a-a: If this is from McGivney & Kawamura, why not use USD_2007? Revisiting the reference, 2007 dollars seems to be the assumption.

watertap/data/techno_economic/air_flotation.yaml
```diff
@@ -5,12 +5,13 @@ default:
   capital_cost:
     basis: flow_vol
     cost_factor: None
+    reference: Fit to data from McGivney & Kawamura Figure 5.5.36b
```
adam-a-a: Fig 5.5.36b seems to be for raw water pumping rather than air flotation. 

Did you mean 5.7.4? While we're at it, I think we should include valid range of flowrate based on the cost curve. For example, the stated range for fig 5.7.4 is 10 - 100 MGD.

adam-a-a (changes requested):
double-check a couple things

watertap/data/techno_economic/air_flotation.yaml
```diff
@@ -1,19 +1,27 @@
 default:
   energy_electric_flow_vol_inlet:
     value: 0.3
-    units: kWh/m^3
+    units: kWh/m^3 # default is raw
```
adam-a-a: Do you mean m3 of raw influent?

adam-a-a (approved):

adam-a-a (changes requested):
Noticed that the old fig is referenced. Should we update?

kurbansitterley:
> Noticed that the old fig is referenced. Should we update?

The reference is still correct, I just changed the basis.

adam-a-a (approved):
LGTM- the plan is to merge this as soon as the cost relationship is swapped with the correct one.

kurbansitterley:
> LGTM- the plan is to merge this as soon as the cost relationship is swapped with the correct one.

So 5.7.4 is for a full DAF _plant_ (not just the unit process), the cost breakdown of what is included in that curve is in Appendix A4b in McGivney & Kawamura and includes chemical additions, rapid mix, clarifier, storage etc etc. I don't think we want all of that included in one of our ZO models. I did a cursory search for a DAF cost curve and could not find one. 

Given that, I will modify this PR to remove this unit model.

(btw, this air flotation model is separate from the dissolved air flotation ZO model; that model uses a costing relationship that was developed from a clarifier)

MarcusHolly (approved):

PR #1783, opened 2026-05-17
Update Chemical Addition ZO

kurbansitterley:
## Fixes/Resolves:

#1780 

- updating yaml with references
- adding proper basis for each chemical subtype
- add costing tests

NOTE: I discovered that the capex basis parameters that we are using _already have TPEC applied_ but we are applying it again with our `cost_factor`. I decided not to make that change here since it would be sort of a mess to untangle but should be done in a future PR.

### Legal Acknowledgement


MarcusHolly (approved):

watertap/unit_models/zero_order/chemical_addition_zo.py
```diff
@@ -71,24 +69,39 @@ def build(self):
...
                 to_units=pyo.units.m**3 / pyo.units.s,
             )
 
-        self.chemical_flow_constraint = pyo.Constraint(
-            self.flowsheet().time, rule=rule_chem_flow
-        )
+        self.chemical_flow_constraint = pyo.Constraint(rule=rule_chem_flow_vol)
+
+        def rule_chem_flow_mass(blk):
+            return blk.chemical_flow_mass == pyo.units.convert(
```
jakechurchi, line 87: Would it be more clear to have mass flow constraint first, then `rule_chem_flow_vol` be `chemical_flow_mass / solution_density`?

watertap/unit_models/zero_order/chemical_addition_zo.py
jakechurchi, line 1: `self.chemical_flow_vol` still has the self.flowsheet.time() but no other variable does?

watertap/unit_models/zero_order/chemical_addition_zo.py
```diff
@@ -109,15 +122,15 @@ def cost_chemical_addition(blk, number_of_parallel_units=1):
...
-        t0 = blk.flowsheet().time.first()
-        chem_flow_mass = (
-            blk.unit_model.chemical_dosage[t0]
-            * blk.unit_model.properties[t0].flow_vol
-            / blk.unit_model.ratio_in_solution
-        )
-        sizing_term = blk.unit_model.chemical_flow_vol[t0] / (
-            pyo.units.gal / pyo.units.day
-        )
+        if chem_name in ["alum"]:
```
jakechurchi, line 123: Would it be worth the effort for this cost equation to be reworked so the sizing term is gal/day for consistency?

watertap/data/techno_economic/chemical_addition.yaml
```diff
@@ -20,8 +28,15 @@ alum:
...
   capital_cost:
-    basis: chemical_flow_mass
     cost_factor: TPEC
+    reference: McGivney & Kawamura, Figure 5.5.6, Cost Estimating Manual for Water Treatment Facilities
+    validity_range:
+      lower_bound:
+        value: 2
+        units: gal/hr
+      upper_bound:
+        value: 10000
```
jakechurchi: Graph in reference only goes to 1,000 gal/hr

watertap/data/techno_economic/chemical_addition.yaml
jakechurchi, line 1: Didn't double check the solution density values

watertap/data/techno_economic/chemical_addition.yaml
```diff
@@ -20,8 +28,15 @@ alum:
     value: 0.5
     units: dimensionless
   capital_cost:
-    basis: chemical_flow_mass
     cost_factor: TPEC
+    reference: McGivney & Kawamura, Figure 5.5.6, Cost Estimating Manual for Water Treatment Facilities
+    validity_range:
+      lower_bound:
+        value: 2
```
jakechurchi, line 36: Probably a lower bound of 40 should be used since that was the lowest flowrate used to find the power law fits?

watertap/data/techno_economic/chemical_addition.yaml
```diff
@@ -78,33 +134,21 @@ caustic_soda:
     value: 0.5
     units: dimensionless
   capital_cost:
-    basis: chemical_flow_mass
     cost_factor: TPEC
+    reference: McGivney & Kawamura, Figure 5.5.12b, Cost Estimating Manual for Water Treatment Facilities
+    validity_range:
```
jakechurchi, line 141: Valid range from data is 360 to 1600 gal/day

watertap/data/techno_economic/chemical_addition.yaml
```diff
@@ -39,14 +54,47 @@ ammonia:
     value: 0.3
```
jakechurchi: 29% in reference doc

watertap/data/techno_economic/chemical_addition.yaml
```diff
@@ -154,8 +213,15 @@ lime:
...
   capital_cost:
-    basis: chemical_flow_mass
     cost_factor: TPEC
+    reference: McGivney & Kawamura, Figure 5.5.9, Cost Estimating Manual for Water Treatment Facilities
+    validity_range:
+      lower_bound:
+        value: 10
+        units: lb/day
+      upper_bound:
+        value: 700
```
jakechurchi: 660 lb/day in the reference figure

watertap/data/techno_economic/chemical_addition.yaml
```diff
@@ -192,11 +266,44 @@ sulfuric_acid:
     value: 1
```
jakechurchi: 93% in reference document

watertap/data/techno_economic/chemical_addition.yaml
```diff
@@ -192,11 +266,44 @@ sulfuric_acid:
...
     units: dimensionless
   capital_cost:
-    basis: chemical_flow_mass
     cost_factor: TPEC
+    reference: McGivney & Kawamura, Figure 5.5.11, Cost Estimating Manual for Water Treatment Facilities
+    validity_range:
+      lower_bound:
+        value: 11
```
jakechurchi, line 301: Lower bound should be 795 gal.day based on data used for power law. Significant deviation from the original fit @ 11 gal/day.

jakechurchi: This also applies to anti-scalant, hydrochloric acid, and sodium bisulfite

watertap/data/techno_economic/chemical_addition.yaml
```diff
@@ -39,14 +54,47 @@ ammonia:
...
+    units: dimensionless
+  capital_cost:
+    cost_factor: TPEC
+    reference: McGivney & Kawamura, Figure 5.5.14, Cost Estimating Manual for Water Treatment Facilities
+    validity_range:
+      lower_bound:
+        value: 200
+        units: lb/day
+      upper_bound:
+        value: 5000
```
jakechurchi: Goes up to 5100 gal/day

watertap/data/techno_economic/chemical_addition.yaml
```diff
@@ -192,11 +266,44 @@ sulfuric_acid:
...
+      upper_bound:
+        value: 5300
+        units: gal/day
     capital_a_parameter:
       value: 900.97
       units: USD_2007
     capital_b_parameter:
       value: 0.6179
+      units: dimensionless
+chlorine:
```
jakechurchi: Only one not in alphabetical order?

watertap/data/techno_economic/default_case_study.yaml
```diff
@@ -37,21 +44,24 @@ defined_flows:
...
-    value: 0.34
-    units: USD_2020/kg
+    value: 0.44
+    units: USD_2023/kg
     purity: 1
+    reference: Chlorine, US Gulf, domestic, quarter contract; CatCost v 1.1.1
   ferric_chloride:
     value: 0.88
     units: USD_2020/kg
     purity: 1
```
jakechurchi, line 59: Is the idea to have a reference for every chemical? Quite a few are still missing.

watertap/data/techno_economic/default_case_study.yaml
```diff
@@ -16,13 +16,20 @@ defined_flows:
...
   alum:
-    value: 0.69
-    units: USD_2020/kg
+    value: 0.67
+    units: USD_2023/kg
     purity: 0.17
+    reference: Aluminum sulphate, 5-lb. bgs., c.l., works, frt. equald., 17% Al203, W. Coast; CatCost v 1.1.1
```
jakechurchi, line 22: I don't have experience with CatCost tool, so I didn't double check these costs

jakechurchi (commented):
Some changes to validity range on the capex equations should be made. Seems like they were pulled from the original source, but many of those relations were linear and the power law equations used here should have different bounds.

Also- I can review again if density and unit costs need to be double checked.

watertap/unit_models/zero_order/chemical_addition_zo.py
kurbansitterley: done

watertap/unit_models/zero_order/chemical_addition_zo.py
kurbansitterley: I would argue that this is consistent with the reference. The user will never know the difference, we just need to make sure it is correctly implemented.

watertap/unit_models/zero_order/chemical_addition_zo.py
kurbansitterley: The `pump_electricity` function requires the flow rate being costed is indexed to time.

watertap/data/techno_economic/chemical_addition.yaml
kurbansitterley: did you check this against the data or the original equation?

kurbansitterley: sulfuric-acid-fits
I didn't make the original fit but I can see why they did it the way they did. If you fit power law to the full set you end up significantly under predicting at higher flows. We want to capture that behavior at upper flows

watertap/data/techno_economic/chemical_addition.yaml
kurbansitterley: actually upper bound is 10000; see Table 5.2.1 in McGivney & Kawamura

watertap/data/techno_economic/chemical_addition.yaml
kurbansitterley: see Table 5.2.1

watertap/data/techno_economic/chemical_addition.yaml
kurbansitterley: I think we go with the range of the original equation here (Table 5.2.1)

watertap/data/techno_economic/default_case_study.yaml
kurbansitterley: I updated costs and added the reference if I could find one for each chemical. Some of the cost values are holdovers from WT3 and we don't have a reference for it.

watertap/data/techno_economic/default_case_study.yaml
kurbansitterley: https://catcost.chemcatbio.org/

go to WebApp (at top)

watertap/data/techno_economic/chemical_addition.yaml
```diff
@@ -9,6 +9,14 @@ default:
...
     units: dimensionless
+  capital_cost:
+    cost_factor: TPEC
+    capital_a_parameter:
+      value: 10000
+      units: USD_2007
+    capital_b_parameter:
+      value: 0.7
+      units: dimensionless
 alum:
```
MuktaHardikar, line 20: Reference for density if needed: https://www.pentachemicals.eu/data/Bezpecnostni_listy/en/Aluminium%20sulfate%20solution%2050%20in%20water_1459_1.0.pdf

kurbansitterley: I don't think we need references for densitys but I added it.

MuktaHardikar (approved):
LGTM!

MuktaHardikar (approved):
LGTM!

watertap/data/techno_economic/chemical_addition.yaml
jakechurchi: Yeah sorry- didn't see that table, I was basing it off the figures

jakechurchi (approved):
LGTM

PR #1785, opened 2026-05-17
Update Evaporation Pond ZO

kurbansitterley:
## Fixes/Resolves:

#1780 

## Summary/Motivation:

In addition to #1780, this model was not usable with costing since a `cost_factor` as never added to the costing method once we made that a requirement. This will remedy that.

- add references
- update units to be consistent
- test costing against reference

### Legal Acknowledgement

### DIFF
--- a/watertap/data/techno_economic/evaporation_pond.yaml
+++ b/watertap/data/techno_economic/evaporation_pond.yaml
@@ -1,4 +1,6 @@
 default:
+  # reference_evaporation_rate: M. E. Jensen and H. R. Haise, "Estimating Evapotranspiration from Solar Irradiation". (1963).
+  # reference_evaporation_pond: Section 10. Membrane Concentrate Disposal: Practices and Regulation (2006)
   energy_electric_flow_vol_inlet:
     value: 0.0
     units: kWh/m^3
@@ -7,16 +9,16 @@ default:
     units: kelvin
   solar_radiation:
     value: 25
-    units: mJ/m^2
+    units: MJ/m^2/day
   evap_rate_calc_a_parameter:
     value: 0.41
-    units: mm/day
+    units: (mm*m^2)/MJ
   evap_rate_calc_b_parameter:
     value: 0.025
-    units: m^2/mJ
+    units: 1/kelvin
   evap_rate_calc_c_parameter:
     value: 0.078
-    units: m^2/mJ
+    units: dimensionless
   evaporation_rate_adj_factor:
     value: 0.7
     units: dimensionless
@@ -32,16 +34,17 @@ default:
   dike_height:
     value: 8
     units: ft
-  capital_cost:
+  capital_cost: 
     cost_factor: None
+    # reference: Section 10.6 Membrane Concentrate Disposal: Practices and Regulation (2006)
     liner_thickness:
-      value: 50
+      value: 60
       units: mil
     land_cost:
       value: 5000
       units: acre^-1 
     land_clearing_cost:
-      value: 1000
+      value: 4000
       units: acre^-1     
     cost_per_acre_a_parameter:
       value: 5406
--- a/watertap/data/techno_economic/evaporation_pond.yaml
+++ b/watertap/data/techno_economic/evaporation_pond.yaml
@@ -48,7 +48,7 @@ default:
     land_cost:
       value: 5000
       units: acre^-1 
-    land_clearing_cost:
+    land_clearing_cost: # default is for medium wooded land type
       value: 4000
       units: acre^-1     
     cost_per_acre_a_parameter:
--- a/watertap/data/techno_economic/evaporation_pond.yaml
+++ b/watertap/data/techno_economic/evaporation_pond.yaml
@@ -1,6 +1,6 @@
 default:
-  # reference_evaporation_rate: M. E. Jensen and H. R. Haise, "Estimating Evapotranspiration from Solar Irradiation". (1963).
-  # reference_evaporation_pond: Section 10. Membrane Concentrate Disposal: Practices and Regulation (2006)
+  reference_evaporation_rate:
+    value: M. E. Jensen and H. R. Haise, "Estimating Evapotranspiration from Solar Irradiation". (1963).
   energy_electric_flow_vol_inlet:
     value: 0.0
     units: kWh/m^3
@@ -8,8 +8,9 @@ default:
     value: 298
     units: kelvin
   solar_radiation:
-    value: 25
+    value: 21.3
     units: MJ/m^2/day
+    reference: Tucson, AZ; GHI average for TMY 2024, data from https://nsrdb.nlr.gov/data-viewer
   evap_rate_calc_a_parameter:
     value: 0.41
     units: (mm*m^2)/MJ
@@ -18,9 +19,6 @@ default:
     units: 1/kelvin
   evap_rate_calc_c_parameter:
     value: 0.078
-    units: dimensionless
-  evaporation_rate_adj_factor:
-    value: 0.7
     units: dimensionless
   adj_area_calc_a_parameter:
     value: 1.2
@@ -36,7 +34,14 @@ default:
     units: ft
   capital_cost: 
     cost_factor: None
-    # reference: Section 10.6 Membrane Concentrate Disposal: Practices and Regulation (2006)
+    reference: BLM, Membrane Concentrate Disposal - Practices and Regulation (2006), Section 10.6 
+    validity_range:
+      lower_bound:
+        value: 10
+        units: acre
+      upper_bound:
+        value: 100
+        units: acre
     liner_thickness:
       value: 60
       units: mil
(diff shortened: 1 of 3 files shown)


MarcusHolly (approved):

watertap/data/techno_economic/evaporation_pond.yaml
```diff
@@ -1,24 +1,24 @@
...
   solar_radiation:
-    value: 25
-    units: mJ/m^2
+    value: 21.3
+    units: MJ/m^2/day
+    reference: Tucson, AZ; GHI average for TMY 2024, data from https://nsrdb.nlr.gov/data-viewer
   evap_rate_calc_a_parameter:
     value: 0.41
-    units: mm/day
+    units: (mm*m^2)/MJ
```
jakechurchi, line 16: Idk if there's any value to giving the parameters more descriptive names, but physically this is the reciprocal of latent heat of vaporization of pure water.

watertap/data/techno_economic/evaporation_pond.yaml
```diff
@@ -1,24 +1,24 @@
...
-    units: m^2/mJ
+    units: 1/kelvin
   evap_rate_calc_c_parameter:
     value: 0.078
-    units: m^2/mJ
-  evaporation_rate_adj_factor:
-    value: 0.7
     units: dimensionless
   adj_area_calc_a_parameter:
     value: 1.2
```
jakechurchi, line 24: This is a contingency factor, name could reflect that

watertap/data/techno_economic/evaporation_pond.yaml
jakechurchi, line 26: Name could reflect that this is extra surface area for dike

watertap/data/techno_economic/evaporation_pond.yaml
```diff
@@ -32,16 +32,24 @@ default:
...
-      value: 50
+      value: 60
       units: mil
     land_cost:
       value: 5000
       units: acre^-1 
     land_clearing_cost:
-      value: 1000
+      value: 4000
       units: acre^-1
```
jakechurchi, line 53: Why don't these have currency units like the parameters below?

watertap/data/techno_economic/evaporation_pond.yaml
```diff
@@ -32,16 +32,24 @@ default:
...
+        value: 100
+        units: acre
     liner_thickness:
-      value: 50
+      value: 60
       units: mil
     land_cost:
       value: 5000
       units: acre^-1 
     land_clearing_cost:
```
jakechurchi: Could note this cost corresponds to "Medium Wooded" land type

watertap/unit_models/zero_order/evaporation_pond_zo.py
```diff
@@ -228,20 +212,24 @@ def cost_evaporation_pond(blk):
...
-                + E * blk.unit_model.dike_height[t0]
+                + E * blk.unit_model.dike_height
             ),
             to_units=blk.config.flowsheet_costing_block.base_currency,
         )
+        factor = parameter_dict["capital_cost"]["cost_factor"]
+        blk.costing_package.add_cost_factor(blk, factor)
 
-        blk.capital_cost_constraint = pyo.Constraint(expr=blk.capital_cost == expr)
+        blk.capital_cost_constraint = pyo.Constraint(
```
jakechurchi, line 228: `cost_factor` is set to 'None' in the yaml

watertap/unit_models/zero_order/tests/test_evaporation_pond_zo.py
jakechurchi, line 51: These solutes are never referenced in the model or costing function (except TDS). Is that intended?

watertap/data/techno_economic/evaporation_pond.yaml
jakechurchi: Perhaps `cost_per_acre_c_parameter `and `cost_per_acre_d_parameter `should be dimensionless so that these can have units of $/acre.

watertap/unit_models/zero_order/tests/test_evaporation_pond_zo.py
```diff
@@ -265,27 +261,51 @@ def test_report(self, model):
...
+    m.fs.unit.inlet.flow_mass_comp[0, "H2O"].fix(flow_mass)
+    m.fs.unit.inlet.flow_mass_comp[0, "tds"].fix(flow_conc)
     m.fs.unit.load_parameters_from_database(use_default_removal=True)
 
     m.fs.unit.costing = UnitModelCostingBlock(flowsheet_costing_block=m.fs.costing)
+    # Test same costing conditions as reference
+    m.fs.unit.dike_height.fix(8 * pyunits.ft)
+    m.fs.costing.evaporation_pond.liner_thickness.fix(60 * pyunits.mil)
+    m.fs.costing.evaporation_pond.land_cost.fix(5000 * pyunits.acre**-1)
+    m.fs.costing.evaporation_pond.land_clearing_cost.fix(4000 * pyunits.acre**-1)
```
jakechurchi, line 300: See other comment about changing units is $/acre

watertap/unit_models/zero_order/tests/test_evaporation_pond_zo.py
```diff
@@ -299,6 +319,8 @@ def test_costing():
...
     assert isinstance(m.fs.unit.costing.capital_cost, Var)
     assert isinstance(m.fs.unit.costing.capital_cost_constraint, Constraint)
-
-    assert_units_consistent(m.fs)
-    assert degrees_of_freedom(m.fs.unit) == 0
+    assert pytest.approx(value(m.fs.unit.area), rel=1e-3) == 9.99437
+    assert pytest.approx(value(m.fs.unit.adj_area), rel=1e-3) == 16.6973
+    assert pytest.approx(value(m.fs.costing.LCOW), rel=1e-3) == 0.96967
+    # Total capital cost = $743,376 from worksheet in reference
+    assert pytest.approx(value(m.fs.costing.total_capital_cost), rel=1e-3) == 774628.72
```
jakechurchi: I believe this number should exactly match the calculated total capital cost of $737,045 if the same formula is being used.

watertap/unit_models/zero_order/evaporation_pond_zo.py
jakechurchi, line 228: I'm also surprised the reference worksheet doesn't include excavation costs- I thought that might be significant.

watertap/unit_models/zero_order/tests/test_evaporation_pond_zo.py
```diff
@@ -265,27 +261,51 @@ def test_report(self, model):
...
     m.fs.unit = EvaporationPondZO(property_package=m.fs.params, database=m.db)
 
-    m.fs.unit.inlet.flow_mass_comp[0, "H2O"].fix(10)
-    m.fs.unit.inlet.flow_mass_comp[0, "tds"].fix(123)
-    m.fs.unit.inlet.flow_mass_comp[0, "magnesium"].fix(456)
-    m.fs.unit.inlet.flow_mass_comp[0, "calcium"].fix(789)
-    m.fs.unit.inlet.flow_mass_comp[0, "nitrate"].fix(10)
-    m.fs.unit.inlet.flow_mass_comp[0, "sulfate"].fix(11)
-    m.fs.unit.inlet.flow_mass_comp[0, "tss"].fix(12)
+    flow_vol = 0.0416 * pyunits.Mgallons / pyunits.day
```
jakechurchi: Area is calculated from the flow_vol --> consider lowering to give exactly 10 Acre pond size to match the reference example.

watertap/unit_models/zero_order/tests/test_evaporation_pond_zo.py
jakechurchi, line 58: Units?

jakechurchi (commented):
Suggesting an adjustment to some units and maybe variable names. Also changing the test value to exactly match example from reference.

watertap/unit_models/zero_order/tests/test_evaporation_pond_zo.py
jakechurchi: `assert pytest.approx(value(m.fs.unit.costing.capital_cost), rel=1e-3) == 737045` was added. This is the value to compare against for checking the equation is implemented correctly. The equation is a regression, so noting $743,376 as the true comparison is still valid.

jakechurchi: Also- I noticed `costing.total_capital_cost` assumes a land factor cost of 0.15%, but land costs have already been included in `unit.costing.capital_cost`. This has a very tiny impact on cost... but probably shouldn't be there.

watertap/data/techno_economic/evaporation_pond.yaml
kurbansitterley: I like keeping costing units in the `cost_per_acre` params, but yes they could go either place

watertap/unit_models/zero_order/tests/test_evaporation_pond_zo.py
kurbansitterley: > `assert pytest.approx(value(m.fs.unit.costing.capital_cost), rel=1e-3) == 737045` was added. This is the value to compare against for checking the equation is implemented correctly. The equation is a regression, so noting $743,376 as the true comparison is still valid.

I modified this to test actual value of `736841.37`.

kurbansitterley: > Also- I noticed `costing.total_capital_cost` assumes a land factor cost of 0.15%, but land costs have already been included in `unit.costing.capital_cost`. This has a very tiny impact on cost... but probably shouldn't be there.

Yes this is also why testing `total_capital_cost` is a good idea. We can't _not_ apply that land cost factor selectively to different unit models unfortunately (we could in a flowsheet that contained only a single unit because it would be changed globally)

watertap/unit_models/zero_order/tests/test_evaporation_pond_zo.py
kurbansitterley: the values aren't important in these tests

watertap/unit_models/zero_order/tests/test_evaporation_pond_zo.py
kurbansitterley: these are just dummy values, but they would assume the var units of kg/s

watertap/unit_models/zero_order/evaporation_pond_zo.py
kurbansitterley: I am guessing they are lumped in with the dike height

watertap/data/techno_economic/evaporation_pond.yaml
kurbansitterley: I am not sure that is the case

jakechurchi (approved):
LGTM

PR #1788, opened 2026-05-17
Update GAC ZO

kurbansitterley:
## Fixes/Resolves:

#1780 

-
-

### Legal Acknowledgement

### DIFF
--- a/watertap/data/techno_economic/gac.yaml
+++ b/watertap/data/techno_economic/gac.yaml
@@ -29,6 +29,14 @@ default: # TODO: confirm any differentiation in removal across subtypes
     units: kg/m^3
   capital_cost:
     cost_factor: None
+    reference: United States Environmental Protection Agency. (2021). Work Breakdown Structure-Based Cost Model for Granular Activated Carbon Drinking Water Treatment.
+    validity_range:
+      lower_bound:
+        value: 0.124
+        units: Mgallons/day
+      upper_bound:
+        value: 75.072
+        units: Mgallons/day
     contactor_cost_coeff_0:
       value: 10010.9
       units: USD_2020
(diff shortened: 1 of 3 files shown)


MarcusHolly (approved):

kurbansitterley:
I need to fix the relationship for gravity basin costing

watertap/data/techno_economic/gac.yaml
jakechurchi, line 8: Add specific study name and year?

jakechurchi, line 4: Reference? I don't see a kW/m3 value in the costing model spreadsheet.

jakechurchi, line 22: 80% of costing model examples have 7.5 min as the EBCT

jakechurchi, line 69: Examples in costing spreadsheet have 100x lower energy between pressure vessel and gravity technologies. The values here are only 2x different.

jakechurchi, line 129: This is a more descriptive naming convention for power law cost coefficients. In other models, it's just "a" and "b"

jakechurchi, line 3: The cost spreadsheet gives 3 energy coefficients (for a poly fit). But only a constant electricity parameter seems to be implemented.

watertap/unit_models/zero_order/tests/test_gac_zo.py
```diff
@@ -337,19 +351,38 @@ def test_costing(subtype):
     m.fs.params = WaterParameterBlock(solute_list=["sulfur", "toc", "tss"])
 
     m.fs.costing = ZeroOrderCosting()
+    m.fs.costing.base_currency = pyunits.USD_2017
```
jakechurchi, line 354: The reference costing year is 2020, unless this specific comparison is different

watertap/unit_models/zero_order/tests/test_gac_zo.py
```diff
@@ -322,13 +319,30 @@ def test_load_parameters(self, model, subtype):
...
+    "pressure_vessel": 0.200446,
+    "gravity": 0.19947,
+}
+sec_dict = {
+    "default": 0.013288,
+    "pressure_vessel": 0.00652,
+    "gravity": 0.01328,
+}
+capex_dict = {
+    "default": 1219387.58,  # ~$1.348M from reference
```
jakechurchi, line 333: I kind of thought default was the exact same as gravity. There is no information on default in the cost spreadsheet.

jakechurchi (commented):
Only major change is the energy intensity could use a polynomial fit instead of a constant value. 
The polynomial coefficients have already been calculated in the cost spreadsheet.

PR #1791, opened 2026-05-17
Update Coag and Floc ZO

kurbansitterley:
## Fixes/Resolves:

#1780 

- use one type of polymer `polymer` and add do `default_case_study`
- add variables/constraints necessary to cost the flow of alum and polymer
-

### Legal Acknowledgement


watertap/data/techno_economic/coag_and_floc.yaml
```diff
@@ -2,8 +2,6 @@ default:
   recovery_frac_mass_H2O:
     value: 0.9
     units: dimensionless
-    reference: Based on costs for Aluminum Sulfate (Liquid Alum Feed) from Cost Estimating
-      Manual for Water Treatment Facilities (McGivney/Kawamura)
```
MarcusHolly, line 6: Removed but not replaced... assuming the data shown in this reference did not match our implementation

MarcusHolly (approved):
LGTM

watertap/data/techno_economic/coag_and_floc.yaml
kurbansitterley: This is a reference to Figure 5.5.6 in McGivney/Kawamura, which is included in the reference I added below.

kurbansitterley: Throughout all these yaml changes I tried to be consistent in putting the ``reference`` under the proper subheading (in this case, under `capital_cost`)

watertap/data/techno_economic/coag_and_floc.yaml
```diff
@@ -50,6 +48,37 @@ default:
...
+        upper_bound:
+          value: 220
+          units: lb/day 
+      flocculator:
+        lower_bound:
+          value: 0.015
+          units: Mgallons
+        upper_bound:
+          value: 7
+          units: gallons
```
jakechurchi: Should be 7.5 Mgallons (units are wrong)

watertap/unit_models/zero_order/tests/test_coag_and_floc_zo.py
```diff
@@ -244,29 +242,44 @@ def test_costing():
...
+    m.fs.costing.base_currency = pyunits.USD_2007
+
+    m.fs.unit = CoagulationFlocculationZO(property_package=m.fs.params, database=m.db)
+    rho = 1000 * pyunits.kg / pyunits.m**3
+    flow_vol = 26736 * pyunits.gallon / pyunits.minute
+    flow_mass = flow_vol * rho
+    m.fs.unit.inlet.flow_mass_comp[0, "H2O"].fix(flow_mass)
+    m.fs.unit.inlet.flow_mass_comp[0, "tss"].fix(0.1)
+    m.fs.unit.load_parameters_from_database(use_default_removal=True)
+    m.fs.unit.alum_dose.fix(10)
```
jakechurchi, line 258: Shouldn't there also be a line for `polymer_dose.fix()`?

jakechurchi (changes requested):

jakechurchi (changes requested):
An incorrect unit in the yaml. Otherwise good.

watertap/data/techno_economic/coag_and_floc.yaml
kurbansitterley: hmmm... actually I think it would be 7.4 Mgallons from the figure, but Table 5.2.1 says 0.02-7 Mgallons... I say we go with the figure for upper bound and table for lower bound?

watertap/unit_models/zero_order/tests/test_coag_and_floc_zo.py
kurbansitterley: this is set by yaml

watertap/unit_models/zero_order/coag_and_floc_zo.py
jakechurchi, line 1: Changes since last review-
- The chemicals are now specified by dosage and ratio in solution, requiring additional variables. 
- change from cation/anion to alum/polymer

Ok- seems good to me

jakechurchi (approved):
LGTM

PR #1792, opened 2026-05-17
Update ZO References/yamls; Test Some ZO Costing Results

kurbansitterley:
## Fixes/Resolves:

Remainder of #1780 

- add references, modify parameter and costing year in a few cases
- test some costing results

### Legal Acknowledgement


watertap/data/techno_economic/dissolved_air_flotation.yaml
```diff
@@ -17,9 +17,6 @@ default:
   recovery_frac_mass_H2O:
     value: 0.9999
     units: dimensionless
-    reference: (1) https://www.engineeringtoolbox.com/water-pumping-costs-d_1527.html,
-      (2) https://onlinelibrary.wiley.com/doi/pdf/10.1002/9780470260036.ch5, (3) Cost
-      Estimating Manual for Water Treatment Facilities (McGivney/Kawamura)
```
MarcusHolly, line 22: Why were these deleted? Did the information from these resources not match our implementation?

watertap/data/techno_economic/surface_discharge.yaml
```diff
@@ -13,6 +13,7 @@ default:
     pipe_cost_basis:
       value: 35000
       units: USD_2020/(miles*inches)
+      reference: Personal communication from WT3 member for produced water case studies
```
MarcusHolly: Where is this number from? Might be more professional to say something like "Anecdotal data from XYZ (person or case study) or simply just "produced water case studies"

watertap/data/techno_economic/surface_discharge.yaml
```diff
@@ -38,6 +39,7 @@ emwd:
...
```
MarcusHolly: Same here

watertap/data/techno_economic/uv.yaml
```diff
@@ -1,5 +1,5 @@
 default:
-  capital_cost:  #Cost parameters retrieved by fitting TWDB data in ALAMO (MADp modeler, max terms=4)
```
MarcusHolly, line 2: This was deleted but not replaced... was this intentional?

watertap/data/techno_economic/uv_aop.yaml
```diff
@@ -1,5 +1,5 @@
-default:         # default is for hydrogen peroxide
-  capital_cost:  #Cost parameters retrieved by fitting TWDB data in ALAMO (MADp modeler, max terms=4)
```
MarcusHolly: Same here

watertap/data/techno_economic/uv_aop.yaml
```diff
@@ -56,7 +56,7 @@ default:         # default is for hydrogen peroxide
     units: kWh/m^3
 
 hydrogen_peroxide:
-  capital_cost: #Cost parameters retrieved by fitting TWDB data in ALAMO (MADp modeler, max terms=4)
+  capital_cost:
     cost_factor: None
```
MarcusHolly, line 65: Same here

watertap/data/techno_economic/well_field.yaml
```diff
@@ -13,6 +13,7 @@ default:
     pipe_cost_basis:
       value: 35000
       units: USD_2018/(miles*inches)
+      reference: Personal communication from WT3 member for produced water case studies
```
MarcusHolly: See comment above about being more "professional" here

MarcusHolly (approved):
Minor comments, but looks good otherwise

watertap/data/techno_economic/surface_discharge.yaml
kurbansitterley: Yes I wasn't sure how to phrase this; I like just "Produced water case studies"

watertap/data/techno_economic/uv.yaml
kurbansitterley: I am working on tracking down the reference for the UV costing parameters, but I know that it is not from fitting TWDB data so will update in a separate PR.

watertap/data/techno_economic/dissolved_air_flotation.yaml
kurbansitterley: Correct, I could not track down the reference for this model. I couldn't even track down the model in WT3. The closest I could get is that the exponent _might_ be from the sedimentation model in McGivney/Kawamura 😅.

That said, there is a DAF model in M&K, but this is not it.

watertap/data/techno_economic/deep_well_injection.yaml
```diff
@@ -1,18 +1,13 @@
...
   capital_cost:
     cost_factor: None
     well_pump_cost:
       value: 16.9e6 # this is wells/pumps fixed capital AFTER applying TIC factor
       units: USD_2011
     pipe_cost_basis:
       value: 35000
-      units: USD_2011/(miles*inches)
+      units: USD_2020/(miles*inches)
+      reference: Produced water case studies
```
jakechurchi, line 11: No tab for DWI on the cost reference spreadsheet, so can't verify these cost numbers.

watertap/unit_models/zero_order/tests/test_deep_well_injection_zo.py
```diff
@@ -10,7 +10,7 @@
 # "https://github.com/watertap-org/watertap/"
 #################################################################################
 """
-Tests for zero-order well field model
```
jakechurchi, line 13: This file hasn't been updated like the other tests.

watertap/data/techno_economic/well_field.yaml
```diff
@@ -1,6 +1,14 @@
 default:
   capital_cost:
     cost_factor: None
+    reference: Voutchkov, N. (2018). Desalination Project Cost Estimating and Management, Table 4.7; 50 m deep wells
+    validity_range:
+      lower_bound:
+        value: 1000
+        units: m^3/day
+      upper_bound:
+        value: 300000
```
jakechurchi: 1250 m3/hr = 30,000 m3/day

watertap/data/techno_economic/well_field.yaml
```diff
@@ -32,12 +41,14 @@ emwd:
     capital_a_parameter:
```
jakechurchi, line 41: The way this fit was created is strange to me because it should just have 1000 m3/day intervals between all the points instead of generating most of the points from the final equation. But probably it's fine.

jakechurchi: This might be what the comment about using a "subset of data is refering to"?

watertap/unit_models/zero_order/deep_well_injection_zo.py
```diff
@@ -81,30 +80,32 @@ def cost_deep_well_injection(blk, number_of_parallel_units=1):
         )
 
         # Get costing parameter sub-block for this technology
-        A, B, C = blk.unit_model._get_tech_parameters(
+        A, pipe_cost_basis, C = blk.unit_model._get_tech_parameters(
```
jakechurchi: I would change this to look like well field, with A and B parameters, then pipe_cost_basis. Wierd to have A and C with no B.

watertap/unit_models/zero_order/chlorination_zo.py
```diff
@@ -65,16 +65,23 @@ def build(self):
         self._fixed_perf_vars.append(self.chlorine_decay_rate)
 
         self.chlorine_dose = Var(
-            self.flowsheet().time, units=pyunits.mg / pyunits.L, doc="Chlorine dose"
+            self.flowsheet().time,
+            # costing correlation is only valid for doses between 1 and 25 mg/L
+            bounds=(1, 25),
```
jakechurchi, line 70: Could add an assertion of these bounds in the test file for chlorination (this was present for some of the other tests.)

watertap/data/techno_economic/ion_exchange.yaml
```diff
@@ -45,14 +65,36 @@ anion_exchange:  # TODO: add some differentiation between default, anion, and ca
       value: 0
```
jakechurchi: Should this be 0.9 like with the default?

watertap/unit_models/zero_order/tests/test_ion_exchange_zo.py
```diff
@@ -121,16 +118,16 @@ def test_solve(self, model):
     @pytest.mark.skipif(solver is None, reason="Solver not available")
     @pytest.mark.component
     def test_solution(self, model):
-        assert pytest.approx(10.005, rel=1e-5) == value(
+        assert pytest.approx(10.0041, rel=1e-5) == value(
```
jakechurchi, line 122: I think most other tests use rel=1e-3 now

watertap/unit_models/zero_order/tests/test_ion_exchange_zo.py
```diff
@@ -302,6 +299,23 @@ def test_load_parameters(self, model, subtype):
...
+    "anion_exchange": 0.00542,
+}
+capex_dict = {
+    "default": 5047528.78,
+    "cation_exchange": 6274349.35,  # $5885981 from reference
+    "anion_exchange": 5047528.78,  # $4237262 from reference
+}
+
+
 @pytest.mark.parametrize("subtype", [k for k in params.keys() if k != "clinoptilolite"])
```
jakechurchi, line 322: Why bypass the test for clinoptilolite? I guess because there is no source (comment below)

watertap/data/techno_economic/ion_exchange.yaml
jakechurchi, line 1: clinoptilolite capital costs lack a reference (and don't have a tab in the costing spreadsheet)

watertap/data/techno_economic/backwash_solids_handling.yaml
```diff
@@ -2,10 +2,24 @@ default:
...
     units: dimensionless
+  capital_cost:
+    basis: flow_vol
+    cost_factor: TIC
+    reference: Based on costs for subprocesses in Table 5.7.1 (Equations 29, 30, 31, 32, 33, 37, 38, 39) in Cost Estimating Manual for Water Treatment Facilities (McGivney/Kawamura)
+    reference_state:
+      value: 100
+      units: Mgallons/day
+    capital_a_parameter:
+      value: 2870106
```
jakechurchi: This value doesn't match the TCI fits on the spreadsheet. b param matches though.

jakechurchi: There is an A value matching this on the spreadsheet that I believe corresponds to the FCI fit? Because 2870106 * (1/100)^0.9183 = 41812 (the FCI coefficient)

watertap/data/techno_economic/uv.yaml
jakechurchi, line 1: Source for the cost values?

watertap/data/techno_economic/uv_aop.yaml
jakechurchi, line 1: Source for cost values?

watertap/data/techno_economic/anaerobic_digestion_oxidation.yaml
```diff
@@ -5,9 +5,11 @@ default:
   capital_cost:
     basis: flow_vol
     cost_factor: None
+    reference: NREL Waste-to-Energy Model (WESyS)
```
jakechurchi, line 8: NLR

watertap/data/techno_economic/anaerobic_digestion_oxidation.yaml
```diff
@@ -5,9 +5,11 @@ default:
...
     reference_state:
-      value: 911.054
+      value: 911.054 # 5.77 MGD
       units: m^3/hr
+      reference: WT3 Beef Processing case study
```
jakechurchi, line 12: When I google "WT3 Beef Processing case study", this doesn't come up. If we want this reference to be findable, maybe include a link or actual name of study?

watertap/data/techno_economic/cartridge_filtration.yaml
```diff
@@ -1,8 +1,30 @@
 default:
+  energy_electric_flow_vol_inlet:
+    value: 0.0002  # todo: revisit this value later on
```
jakechurchi, line 3: This seems super low, like below the energy just to pump water from one location to another. Could use a source.

watertap/data/techno_economic/clarifier.yaml
```diff
@@ -5,22 +5,31 @@ default:
   capital_cost:
     cost_method: cost_power_law_flow
     basis: flow_vol
-    cost_factor: None
+    cost_factor: TPEC
+    reference: McGivney & Kawamura, Figure 5.5.24c, Cost Estimating Manual for Water Treatment Facilities
+    validity_range:
+      # validity range presented here for completeness,
+      # but WaterTAP model is a function of flow rate, not basin area, so cannot determine 
+      # validity range for model
```
jakechurchi, line 13: This comment is relevant to other technologies!

watertap/data/techno_economic/clarifier.yaml
```diff
@@ -5,22 +5,31 @@ default:
...
+          units: ft^2 # of basin surface area
+        upper_bound:
+          value: 160000
+          units: ft^2 # of basin surface area
     reference_state:
       value: 5450.0
       units: m^3/hr
     capital_a_parameter:
-      value: 6.03553383e5
+      value: 177516
```
jakechurchi, line 25: Double checking that surface area scales proportionally to volumetric flowrate, which I guess is an assumption of using this curve.

watertap/unit_models/zero_order/tests/test_conventional_activated_sludge_zo.py
```diff
@@ -214,17 +228,34 @@ def test_costing(subtype):
...
+    m.fs.costing.cost_process()
+    m.fs.costing.add_LCOW(m.fs.unit.properties_in[0].flow_vol)
+    m.fs.costing.add_specific_energy_consumption(
+        m.fs.unit.properties_in[0].flow_vol, name="SEC"
+    )
+
+    m.fs.unit.initialize()
+
+    results = solver.solve(m)
+    assert check_optimal_termination(results)
```
jakechurchi, line 260: Why not just use `assert_optimal_termination`?

watertap/unit_models/zero_order/tests/test_uv_zo.py
```diff
@@ -17,6 +17,7 @@
 
 
 from pyomo.environ import (
+    assert_optimal_termination,
     check_optimal_termination,
```
jakechurchi: Seem reduntant to have both these functions

watertap/unit_models/zero_order/tests/test_media_filtration_zo.py
```diff
@@ -20,9 +20,11 @@
     Block,
     ConcreteModel,
     Constraint,
+    check_optimal_termination,
```
jakechurchi: Redundant if we have `assert_optimal_termination`?

watertap/data/techno_economic/decarbonator.yaml
```diff
@@ -5,6 +5,8 @@ default:
   capital_cost:
     basis: flow_vol
     cost_factor: None
+    reference: Survey of High-Recovery and Zero Liquid Discharge Technologies for Water Utilities (2008). Table A2.2.
+    # validity range cannot be determined because this is the only data point for decabonator in Table A2.2
```
jakechurchi, line 9: I know this is just one data point, but should `capital_b_parameter` really be 1? I thought something like 0.7 was more standard.

watertap/data/techno_economic/electrocoagulation.yaml
```diff
@@ -72,7 +91,22 @@ default:
     capital_floc_a_parameter:
       value: 952902
       units: USD_2007/Mgallons
+      reference: McGivney & Kawamura, Figure 5.5.22, Cost Estimating Manual for Water Treatment Facilities
+      validity_range:
+        lower_bound:
+          value: 0
+          units: Mgallons
+        upper_bound:
+          value: 7
```
jakechurchi: Figure goes up to 7.4 MG

watertap/data/techno_economic/electrocoagulation.yaml
```diff
@@ -72,7 +91,22 @@ default:
...
       value: 177335
       units: USD_2007
-
+      reference: McGivney & Kawamura, Figure 5.5.22, Cost Estimating Manual for Water Treatment Facilities
+      validity_range:
+        lower_bound:
+          value: 0
+          units: Mgallons
+        upper_bound:
+          value: 7
```
jakechurchi, line 115: Figure goes up to 7.4 MG

watertap/data/techno_economic/electrocoagulation.yaml
jakechurchi, line 1: Is the source for the default values the same as for the costs?

watertap/data/techno_economic/electrodialysis_reversal.yaml
```diff
@@ -2,9 +2,25 @@ default:
   elec_coeff_1:
     value: 0.2534
     units: kWh/m^3
+    reference: Membrane Technology and Applications; R. W. Baker (2012) DOI- 10.1002/9781118359686. Fit to data from Figure 10.11
```
jakechurchi: Should be Figure 10.15

watertap/data/techno_economic/electrodialysis_reversal.yaml
```diff
@@ -2,9 +2,25 @@ default:
...
+      lower_bound:
+        value: 0
+        units: mg/L
+      upper_bound:
+        value: 20000
+        units: mg/L
   elec_coeff_2:
     value: 5.149E-4
     units: L/mg*kWh/m^3
+    reference: Membrane Technology and Applications; R. W. Baker (2012) DOI- 10.1002/9781118359686. Fit to data from Figure 10.11
```
jakechurchi: Figure 10.15

watertap/data/techno_economic/electrodialysis_reversal.yaml
```diff
@@ -14,13 +30,14 @@ default:
     capital_a_parameter:
       value: 31e6
       units: USD_2016
+      reference: Based on costs from Fort Irwin / Irwin Water Works case study
     capital_b_parameter:
       value: 1.0
```
jakechurchi, line 35: Assuming this curve is based on one data point- I thought like 0.7 was a more common default exponent?

watertap/data/techno_economic/primary_separator.yaml
```diff
@@ -5,12 +5,24 @@ default:
   capital_cost:
     basis: flow_vol
     cost_factor: TIC
+    reference: McGivney & Kawamura, Figure 5.5.24c, Cost Estimating Manual for Water Treatment Facilities
```
jakechurchi, line 8: Could note this is for a rectangular clarifier

watertap/data/techno_economic/primary_separator.yaml
```diff
@@ -5,12 +5,24 @@ default:
...
+          value: 0
+          units: ft^2 # of basin surface area
+        upper_bound:
+          value: 160000
+          units: ft^2 # of basin surface area
     reference_state:
       value: 5450.0
       units: m^3/hr
     capital_a_parameter:
       value: 6.03553383e5
```
jakechurchi: I think this value might be wrong... but there is no costing test in 'test_primary_separator_zo.py` to compare this against an expected cost value.

watertap/data/techno_economic/water_pumping_station.yaml
```diff
@@ -43,15 +60,24 @@ treated:
...
-      value: 0.8667
+      value: 0.8657
       units: dimensionless
     reference_state:
       value: 1
       units: Mgallons/day
   electricity:
-    value: 186.4  # default for treated in hrsd case (recharge) (converted from 250 hp)
+    value: 186.4
     units: kW
```
jakechurchi, line 82: Shouldn't electricity be a kWh/m3 value because it increases with flowrate? Certaintly not valid over same range as capital costs above.

watertap/data/techno_economic/storage_tank.yaml
jakechurchi, line 1: Confused about the default capital cost numbers. Don't see a reference and b > 1 ?

watertap/data/techno_economic/storage_tank.yaml
```diff
@@ -28,9 +28,17 @@ treated_storage:
...
     cost_factor: None
+    reference: Process Equipment Cost Estimation Final Report (2002) "Cone Roof Tank"
+    validity_range:
+      lower_bound:
+        value: 191.2
+        units: m^3
+      upper_bound:
+        value: 37854.1
+        units: m^3
     capital_a_parameter:
```
jakechurchi, line 39: Looking at the data, I wonder if a linear fit of all data would be better than power law fit of two data points.

watertap/data/techno_economic/sw_onshore_intake.yaml
jakechurchi, line 1: No tab for this in the cost spreadsheet.
I don't know how the fit was calculated, but it seems to consistiently overpredict Figure 4.2. I don't know how Figure 4.4 was used.

watertap/data/techno_economic/sw_onshore_intake.yaml
```diff
@@ -2,12 +2,20 @@ default:
   capital_cost:
     basis: flow_vol
     cost_factor: None
+    reference: Voutchkov, N. (2018). Desalination Project Cost Estimating and Management, Figure 4.2 + Figure 4.4
```
jakechurchi, line 5: If Figure 4.4 is being used, does this assume a dry well or wet well?

watertap/data/techno_economic/ultra_filtration.yaml
jakechurchi, line 1: What is the reference for the energy intensity?

watertap/data/techno_economic/pump.yaml
```diff
@@ -5,69 +5,102 @@ default:
...
+    reference_state:
+      value: 1
+      units: Mgallons/day
 recharge_well:
   energy_electric_flow_vol_inlet:
     value: 0.076798998
     units: kWh/m^3
   capital_cost:
     basis: flow_vol
+    reference: Aquifer Replenishment System Concept Cost Evaluation Prepared for Hampton Roads Sanitation District. Power consumption based on 75-ft pump head (34 psi delta P).
```
jakechurchi, line 103: Do any of the other pump subtypes have references for the power consumption?

watertap/data/techno_economic/pump.yaml
```diff
@@ -85,6 +118,7 @@ ultrafiltration_feed:
   capital_cost:
     basis: flow_vol
     cost_factor: None
+    reference: McGivney & Kawamura Figure 5.5.35a, Cost Estimating Manual for Water Treatment Facilities
```
jakechurchi: Surprised this fit appears to be based off one data point, but the reference is a figure (not in cost spreadsheet).

watertap/data/techno_economic/sedimentation.yaml
```diff
@@ -4,6 +4,7 @@ default:
     units: kWh/m^3
   capital_cost:
     cost_factor: TPEC
+    reference: McGivney & Kawamura, Figure 5.5.24c, Cost Estimating Manual for Water Treatment Facilities
```
jakechurchi: This figure is for the rectangular clarifier, but the scaling variable is basin surface area, which I thought was an issue (has to be flowrate)

jakechurchi: Compare this to settling_pond.yaml, which cites same figure, but has a validity range and different a_parameter

watertap/data/techno_economic/surface_discharge.yaml
jakechurchi, line 1: No tab in spreadsheet for this technology

jakechurchi (changes requested):
Mostly minor comments, but a few changes needed. There are unit models that don't have an associated tab in the cost spreadsheet, and I noted where I wasn't able to double check the values.

watertap/data/techno_economic/anaerobic_digestion_oxidation.yaml
kurbansitterley: it was NREL when it was made

watertap/data/techno_economic/anaerobic_digestion_oxidation.yaml
kurbansitterley: this was conducted in WT3 as a baselining exercise, it is not available in a public document but is the only ref we have. Not ideal but it is what we have.

watertap/data/techno_economic/cartridge_filtration.yaml
kurbansitterley: This is another instance where this came from WT3 without a reference. This is the case for any value in these yamls without a reference. Filling in these gaps is the whole idea behind these PRs where possible.

watertap/data/techno_economic/clarifier.yaml
kurbansitterley: That is the assumption yes.

watertap/data/techno_economic/decarbonator.yaml
kurbansitterley: This is another WT3 model. I did not want to make parameter value changes unless I had a reference to point to. These kinds of changes around changing the costing relationship without a reference (developing a new model) could be done in a separate PR.

watertap/data/techno_economic/electrocoagulation.yaml
kurbansitterley: yes this is another example where the table and figure have different bounds

watertap/data/techno_economic/electrocoagulation.yaml
kurbansitterley: same as above.

watertap/data/techno_economic/electrodialysis_reversal.yaml
kurbansitterley: see decarbonator comment

watertap/data/techno_economic/ion_exchange.yaml
kurbansitterley: I did not modify that modify subtype because I am not familiar with it, @MarcusHolly might know bounds for it

watertap/data/techno_economic/ion_exchange.yaml
kurbansitterley: good catch

watertap/data/techno_economic/primary_separator.yaml
kurbansitterley: you are correct, this is the value _after_ TPEC is applied, which we apply afterwards

watertap/data/techno_economic/pump.yaml
kurbansitterley: nope

kurbansitterley: imo this model (and others) should just be removed for this reason, but "that's for a different PR"

watertap/data/techno_economic/pump.yaml
kurbansitterley: same comment

watertap/data/techno_economic/storage_tank.yaml
kurbansitterley: yeah this was the best fit for the data

watertap/data/techno_economic/storage_tank.yaml
kurbansitterley: maybe, but in a future PR

watertap/data/techno_economic/ultra_filtration.yaml
kurbansitterley: could not locate the reference

watertap/data/techno_economic/uv.yaml
kurbansitterley: this was in a separate PR

watertap/data/techno_economic/uv_aop.yaml
kurbansitterley: this was in a separate PR

watertap/data/techno_economic/well_field.yaml
kurbansitterley: I am confused what you mean here

watertap/data/techno_economic/well_field.yaml
kurbansitterley: I am confused by how a lot of these were created but I believe you are correct

watertap/unit_models/zero_order/tests/test_conventional_activated_sludge_zo.py
kurbansitterley: the effect is the same

watertap/unit_models/zero_order/tests/test_ion_exchange_zo.py
kurbansitterley: yeah 1e-5 is probably overkill but I think both are ok, not worth going and changing all tests

watertap/data/techno_economic/ion_exchange.yaml
MarcusHolly: I am just as clueless

watertap/unit_models/zero_order/tests/test_deep_well_injection_zo.py
kurbansitterley: I mostly only updated tests for files that had a cost curve that was a function of flow.

watertap/data/techno_economic/water_pumping_station.yaml
kurbansitterley: There is a config option for this model where you can set the pumping power directly, which is how this would be used.

watertap/data/techno_economic/sedimentation.yaml
kurbansitterley: Let me check into this

watertap/data/techno_economic/well_field.yaml
jakechurchi: The current bound is 300,000 m3/day but it should be 30,000 m3/day

watertap/unit_models/zero_order/deep_well_injection_zo.py
kurbansitterley: I changed them to be their parameter names in the yaml

watertap/data/techno_economic/electrocoagulation.yaml
kurbansitterley: this is a model that benefits from user input for design parameters, so these are more just reasonable values

watertap/data/techno_economic/well_field.yaml
kurbansitterley: You are correct. This relationship is created from the equations in Table 4.7

watertap/data/techno_economic/well_field.yaml
kurbansitterley: good catch!

watertap/data/techno_economic/deep_well_injection.yaml
kurbansitterley: Let me check into this

watertap/data/techno_economic/sedimentation.yaml
kurbansitterley: > This figure is for the rectangular clarifier, but the scaling variable is basin surface area, which I thought was an issue (has to be flowrate)

There is no reason the scaling variable has to be flowrate, any design variable is fine.

watertap/data/techno_economic/deep_well_injection.yaml
kurbansitterley: added to spreadsheet

watertap/data/techno_economic/pump.yaml
kurbansitterley: > Surprised this fit appears to be based off one data point, but the reference is a figure (not in cost spreadsheet).

What makes you think this is based off one data point? You are correct that this is a curve. It is the treated water pumping station and is a subtype of `water_pumping_station_zo`

watertap/data/techno_economic/surface_discharge.yaml
kurbansitterley: added tab

watertap/data/techno_economic/sedimentation.yaml
kurbansitterley: > Compare this to settling_pond.yaml, which cites same figure, but has a validity range and different a_parameter

The difference between settling_pond + clarifier and sedimentation is that sedimentation calculates basin area (which is the actual independent variable in the costing equation) and settling_pond + clarifier scale to a reference flow from an assumed surface area. They are not perfect relationships.

watertap/unit_models/zero_order/deep_well_injection_zo.py
```diff
@@ -81,30 +80,35 @@ def cost_deep_well_injection(blk, number_of_parallel_units=1):
...
+                * blk.unit_model.pipe_distance[t0]
+                * blk.unit_model.pipe_diameter[t0],
+                to_units=blk.config.flowsheet_costing_block.base_currency,
+            )
         )
-
-        cost_total = pyo.units.convert(
-            cost_well_pump + cost_pipe,
-            to_units=blk.config.flowsheet_costing_block.base_currency,
+        blk.base_cost = pyo.Expression(
```
jakechurchi, line 107: Unclear if the pipe cost should be included in the base_cost. Spreadsheet seems to imply the power law only applies to the Well+pump CAPEX. Also, weird to me that the pipe costs would scale with flowrate via the power law if they already scale with pipe diameter.

watertap/data/techno_economic/well_field.yaml
```diff
@@ -32,12 +41,14 @@ emwd:
...
       units: USD_2018
+      reference: Voutchkov, N. (2018). Desalination Project Cost Estimating and Management, Table 4.7; 50 m deep wells
     capital_b_parameter:
       value: 0.9196
       units: dimensionless
     pipe_cost_basis:
       value: 82600
-      units: USD_2018/(miles*inches)
+      units: USD_2020/(miles*inches)
+      reference: Brackish water case studies
```
jakechurchi, line 51: There a significant difference between the pipe cost referenced here and the one in deep_well_injection. 
(35000 vs. 86200). What accounts for that?

watertap/unit_models/zero_order/well_field_zo.py
```diff
@@ -88,18 +113,8 @@ def cost_well_field(blk):
...
-            to_units=pyo.units.dimensionless,
-        )
         expr = pyo.units.convert(
-            A * Q**B
-            + (
-                pipe_cost_basis
-                * blk.unit_model.pipe_distance[t0]
-                * blk.unit_model.pipe_diameter[t0]
-            ),
+            blk.well_cost + blk.pipe_cost,
```
jakechurchi, line 117: Double checking the pipe cost is supposed to scale with the flowrate (same comment on DWI)

watertap/data/techno_economic/cartridge_filtration.yaml
jakechurchi: So should this be left as a TODO, or be done now? 

Here's a source for 0.05 kWh/m3 - https://www.mdpi.com/1996-1073/18/5/1086 
in paragraph after Figure 7. 
Not sure I trust though b/c Figure 7 is a scuffed bar chart...

watertap/unit_models/zero_order/tests/test_conventional_activated_sludge_zo.py
jakechurchi: I put some other comments calling this out in other files, but it's not really a problem

watertap/data/techno_economic/decarbonator.yaml
jakechurchi: Should an issue be raised requesting those improvements?

watertap/data/techno_economic/pump.yaml
jakechurchi: Because the refernce state is not 1 and the exponent is 0.7 (a default value), I assumed this was based off one point w/ reference flow and CAPEX = a_parameter.

If this is supposed to be the treated water station, then the cost parameters don't match the curve.

watertap/data/techno_economic/sedimentation.yaml
jakechurchi: Ok I see - though if the scaling variable can be area, I don't understand why the clarifier scales to a reference flow if there is an existing cost equation in terms of basin area. Is there no clear relationship between area and flowrate?

watertap/unit_models/zero_order/surface_discharge_zo.py
```diff
@@ -74,7 +73,7 @@ def cost_surface_discharge(blk):
         )
 
         # Get costing parameter sub-block for this technology
-        A, B, pipe_cost_basis, ref_state = blk.unit_model._get_tech_parameters(
+        A, B, pipe_cost_basis, reference_state = blk.unit_model._get_tech_parameters(
```
jakechurchi, line 76: In other some of the other units, A and B were replaced with more descriptive names

watertap/unit_models/zero_order/surface_discharge_zo.py
```diff
@@ -85,7 +84,32 @@ def cost_surface_discharge(blk):
                 "reference_state",
             ],
         )
+        # Have to convert to base currency here
+        pipe_cost_basis = pyo.units.convert(
```
jakechurchi, line 88: Why is this conversion needed? Shouldn't that be handled in the expression below?

watertap/unit_models/zero_order/surface_discharge_zo.py
```diff
@@ -85,7 +84,32 @@ def cost_surface_discharge(blk):
...
+        blk.pipe_cost = pyo.Expression(
+            expr=pyo.units.convert(
+                pipe_cost_basis
+                * blk.unit_model.pipe_distance[t0]
+                * blk.unit_model.pipe_diameter[t0],
+                to_units=blk.config.flowsheet_costing_block.base_currency,
+            )
+        )
 
+        blk.near_shore_discharge = pyo.Expression(
```
jakechurchi, line 102: You should be able to use the `_general_power_law_form` method here I believe, like in 'deep_well_injection.py'.

jakechurchi (commented):
There are a handful of remaining comments to address

watertap/unit_models/zero_order/deep_well_injection_zo.py
kurbansitterley: The original model included both in the base cost, so it remains this way in this mod

watertap/data/techno_economic/well_field.yaml
kurbansitterley: They are different because they came from different case studies.

watertap/unit_models/zero_order/well_field_zo.py
kurbansitterley: Yes, that was the previous implementation so it remains.

watertap/data/techno_economic/cartridge_filtration.yaml
kurbansitterley: left as a TODO. I am also skeptical of that Figure because it says "Filtration" only. It is ok to leave this as is, users are able to easily modify these values.

watertap/data/techno_economic/decarbonator.yaml
kurbansitterley: I don't think anyone would ever work on that issue. Or, I'd rather have an issue that presents all the issues with all the ZO models... If someone needs a decarbonator model, we can revisit. If I needed a decarbonator model, I would probably just make a new one.

watertap/data/techno_economic/pump.yaml
kurbansitterley: oops I thought that the exponent was 0.8671, I removed the reference (I don't know what it is)

watertap/unit_models/zero_order/surface_discharge_zo.py
kurbansitterley: I am going to leave these as-is

watertap/data/techno_economic/sedimentation.yaml
kurbansitterley: Because that's how the clarifier model was implemented originally. If you want to make a model that has basin area as the x-axis for costing, then you have to create a model that will calculate basin area. So you are suddenly confronted with a design problem. What are the needed design variables? What are their bounds? How are they calculated? It is something that wasn't done with the clarifier that @adam-a-a chose to do with the sedimentation model. It is a real undertaking of model development that I didn't want to go in this PR. Ultimately, if we wanted a better clarifer model, we could go down that road.

In the clarifer, we took a basin area that was associated with a flow rate and assumed that the basin area would scale with that flow rate. It's not the worst way to go about it but there are better ways; I would be ok with including a cost from a model with that approach in my flowsheet. This PR was really about adding references and making changes where there were very obvious mistakes and this is not one of those.

watertap/unit_models/zero_order/surface_discharge_zo.py
kurbansitterley: I am honestly not a huge fan of that method and prefer to be explicit so someone who is viewing the file can see what is going on right here (e.g., in this file) rather than chasing down `_general_power_law_form`.

kurbansitterley: Oh, in addition - `_general_power_law_form` can't be used because of the added `pipe_cost` (yes I know it is treated differently here than in well field and DWI).

watertap/unit_models/zero_order/surface_discharge_zo.py
kurbansitterley: Sure, most of the time. But [explicit is better than implicit](https://peps.python.org/pep-0020/#the-zen-of-python). 

And also, it has to be done this way here because pyunits doesn't always like unit coversions for sum expressions (e.g., `pyunits.convert(A + B)`), so you sometimes have to make the unit conversions separately (I have not figured out when it does and doesn't work).

watertap/data/techno_economic/sw_onshore_intake.yaml
kurbansitterley: Added a tab.

watertap/data/techno_economic/sw_onshore_intake.yaml
kurbansitterley: This curve is made by first fitting data in 4.2 and 4.4 (dry well) to power law. And then combining both of those and fitting that data to power law.

watertap/data/techno_economic/backwash_solids_handling.yaml
kurbansitterley: This value is the correct A parameter if 100 MGD is used as reference state.

jakechurchi: Sorry- my point was that the cost factor here should be labeled FCI instead of TCI to match spreadsheet, not that the A value should change.

watertap/data/techno_economic/backwash_solids_handling.yaml
```diff
@@ -11,10 +11,10 @@ default:
     cost_factor: TIC
```
jakechurchi, line 11: This should be FCI not TIC

jakechurchi (approved):
Last comment below. Otherwise LGTM!

watertap/data/techno_economic/backwash_solids_handling.yaml
kurbansitterley: sorry maybe my last comment didn't go through - there is no FCI cost factor in WaterTAP. Only TPEC or TIC. TIC is correct here.

PR #1793, opened 2026-05-17
Update Screen ZO

kurbansitterley:
## Fixes/Resolves:

#1780 

## Summary/Motivation:

Having two screen ZO models is unnecessary. So this PR also will combine `MicroscreenZO` into `ScreenZO` and add additional screen subtypes for the ZO model.

- wipe MicroscreenZO model from the face of the earth
- add microscreen subtype to screen ZO model
- add additional screen subtypes
- test costing results for all subtypes
- add references

### Legal Acknowledgement


watertap/data/techno_economic/screen.yaml
```diff
@@ -1,34 +1,112 @@
...
     units: kWh/m^3
   capital_cost:
     basis: flow_vol
     cost_factor: None
+    reference: Voutchkov, N. (2018). Desalination Project Cost Estimating and Management. Figure 4.6
     reference_state:
-      value: 16666.66667
-      units: m^3/hr
+      value: 1
+      units: m^3/day
```
MarcusHolly, line 18: Big difference... just want to to double check that this was intentional

MarcusHolly (approved):
LGTM, just one thing to double check

watertap/data/techno_economic/screen.yaml
kurbansitterley: It is correct. You will notice the A parameters changed by a factor of ~1e5 also. The reference uses m3/day so this is keeping it consistent with that.

watertap/data/techno_economic/screen.yaml
```diff
@@ -1,34 +1,140 @@
...
+        value: 10000 # estimated
+        units: m^3/day
+      upper_bound:
+        value: 400000
+        units: m^3/day
+    reference_state:
+      value: 1
+      units: m^3/day
+    capital_a_parameter:
+      value: 13.184
```
jakechurchi: In the cost spreadsheet, this value is 12.969

watertap/data/techno_economic/screen.yaml
```diff
@@ -1,34 +1,140 @@
...
+    reference: Voutchkov, N. (2018). Desalination Project Cost Estimating and Management. Figure 4.5
+    validity_range:
+      lower_bound:
+        value: 10000 # estimated
+        units: m^3/day
+      upper_bound:
+        value: 400000
+        units: m^3/day
+    reference_state:
+      value: 1
```
jakechurchi, line 52: What is a reference state in this context? It's totally outside the validity range of the of the cost EQs.

watertap/data/techno_economic/screen.yaml
```diff
@@ -1,34 +1,140 @@
...
+      units: m^3/day
+    capital_a_parameter:
+      value: 15.598
```
jakechurchi: In spreadsheet, value is 5.3181

watertap/data/techno_economic/screen.yaml
```diff
@@ -1,34 +1,140 @@
...
+      units: USD_2018
+    capital_b_parameter:
+      value: 0.9586
```
jakechurchi: In spreadsheet, value is 1.0479.

watertap/data/techno_economic/screen.yaml
```diff
@@ -1,34 +1,140 @@
...
+        value: 400000
+        units: m^3/day
     reference_state:
-      value: 16666.66667
-      units: m^3/hr
+      value: 1
+      units: m^3/day
     capital_a_parameter:
-      value: 3.25e6
+      value: 10.526
```
jakechurchi: In spreadsheet, value is 4.1127

watertap/data/techno_economic/screen.yaml
```diff
@@ -1,34 +1,140 @@
...
       units: USD_2018
     capital_b_parameter:
-      value: 0.9
+      value: 0.9653
```
jakechurchi: In spreadsheet, value is 1.043

jakechurchi (changes requested):
Several of the capital cost parameters don't match the values in the zo_costing_model_data spreadsheet

watertap/data/techno_economic/screen.yaml
kurbansitterley: man I really got my subtypes mixed up here

watertap/data/techno_economic/screen.yaml
kurbansitterley: check out `cost_power_law_flow` in ZO base class

basically, one way that you can develop/create a cost curve is if you know the cost at a "reference state" (e.g., at a particular flow rate) you can then scale that cost to other flows with

`C = A*(Q/Q_ref)**b`

where `A` is the cost at `Q_ref`, and `b` can be assumed from typical TEA factors (0.7). So this allows you to create a relationship from a single data point. But that is not really ideal, it's better to have a full blown cost curve with a custom fit than to take that approach. 

In this case, we have a cost curve as a function of flow rate, so setting the `reference_state` to 1 allows you use the flow rate directly. It is a quirk of how the equation is written, but it is useful also because `reference_state` also dicates the units that your flow rate is in. We convert `Q/Q_ref` to be dimensionless in the units of the `reference_state`.

jakechurchi: Ok, right. That makes sense why the "reference state" is 1 here!

jakechurchi (approved):
LGTM

Issue #1806, opened 2026-06-03
Flex RO Tutorial Improvement


jakechurchi:
### Description

Additional functionalities are needed for using Pricetaker for flexible RO on NAWI 3.25 (Water Replenishment District). The specific improvements to the problem formulation are:
- RO energy surrogate function of two variables (water recovery and feed flowrate) 
- Penalizing solutions that change the flowrate excessively (i.e. preference for on-on-off-off rather than on-off-on-off)
- Replacing the generic pretreatment unit model with UF 
- Additional flow based costs (feed, brine, chemicals)
- Delayed shut-downs (This is distinct from ramping rates, which does exist in Pricetaker already)
- Constraint ensuring operational states at beginning and end of period are consistent
- Could also integrate a surrogate for brine / permeate quality in terms of the feed flowrate and water recovery.
 
Analysis also involves calculating numbers after the model solves
- Flexibility metrics from doi.org/10.1038%2Fs44221-024-00316-4
- Replacement costs as a function of flexibility
- A total cost as a sum of capex and opex

And few other pieces where the functionality is present, but tutorial would benefit from having an example usage
- Demand Response  
- Solar PV and battery
- An improved output plot

### Motivation

These functions will enable analyses that can fit the specifics of a wider range of treatment plants. Additions to the tutorial will improve usability.

### Possible Implementation

A second tutorial notebook will be create an example based on WRD. The additional unit models will be added into existing files. The structure would similar to the existing tutorials, but would be couched within a comparison of different rate structures (including one with demand response) to highlight how the tool could be used to inform admin level decisions in additional to operations.

Alternatively, the existing tutorial could potentially be expanded in a “what if you also wanted to…” and add complexity by adding these functions one by one. This would have the benefit of reducing overlap between the tutorials. But it would make it quite long, conflate the Santa Barbra and WRD plants, and require two solves.

### Additional Context

Previous changes to Pricetaker tutorial was included here #1592

PR #1822, opened 2026-06-21
Update pricetaker tutorial

jakechurchi:
## Fixes/Resolves:

Addresses issue #1806

## Summary/Motivation:
The current pricetaker tutorial was focused on one facility. This PR provides a more complicated example.

- Adding a second tutorial notebook that has energy consumption as a function of both flowrate and recovery
- Adds unit models and functions that allow modeling a facility for NAWI project 3.25

### Legal Acknowledgement


watertap/flowsheets/flex_desal/tests/test_flex_flow.py
```diff
@@ -0,0 +1,418 @@
...
+                "bound_push": 1e-8,
+                "bound_frac": 1e-8,
+                "tol": 1e-6,
+                "max_iter": 5000,
+            }
+        )
+        results = solver.solve(m, tee=True)
+
+        # This is line that is causing the failure on Linux. But I don't know why.
+        # results = solver.solve(m)
```
jakechurchi, line 340: This line is causing failure on Linux, but is fine on Windows. Unsure how to resolve the problem as the error output log gives almost no information

watertap/flowsheets/flex_desal/tests/test_pricetaker_WRD.py
```diff
@@ -0,0 +1,418 @@
...
+from idaes.apps.grid_integration import PriceTakerModel
+
+from watertap.flowsheets.flex_desal import wrd_ro_flowsheet as fs
+from watertap.flowsheets.flex_desal import utils
+from watertap.flowsheets.flex_desal.params import FlexDesalParams
+from watertap.core.solvers import get_solver
+
+
+@pytest.mark.unit
+def test_installed_idaes_pse_version_minimum():
```
jakechurchi: Likely an unneeded test?

watertap/flowsheets/flex_desal/tests/test_flex_flow.py
```diff
@@ -0,0 +1,418 @@
...
+        fs.add_maximum_shutdowns(m)
+
+        # Add the slow shutdown constraint
+        fs.add_delayed_shutdown_constraints(m)
+
+        assert hasattr(m, "max_shutdowns_per_24h_window")
+        assert hasattr(m, "posttreatment_unit_commitment_shutdown")
+
+        # Limit hours when start-ups and shutdowns can occur to reflect when operators are on-site
+        fs.add_working_hours_constraint(m)
```
jakechurchi, line 223: Difficult to do assertions for this function because it only exists for a subset of the hours

watertap/flowsheets/flex_desal/tests/test_flex_flow.py
```diff
@@ -0,0 +1,418 @@
...
+                assert ro_skid.startup.fixed
+                assert ro_skid.startup() == 0
+                assert ro_skid.shutdown.fixed
+                assert ro_skid.shutdown() == 0
+
+    @pytest.mark.unit
+    def test_add_expressions(self, system_frame):
+        m, price_data, peak_hours = system_frame
+
+        fs.add_useful_expressions(m)
```
jakechurchi, line 239: I feel like this function should either be scrapped, or some of the functions below should be included.

watertap/flowsheets/flex_desal/tests/test_flex_flow.py
```diff
@@ -0,0 +1,418 @@
...
+        assert isinstance(m.total_customer_cost, pyo.Expression)
+        assert isinstance(m.total_op_cost, pyo.Expression)
+        assert isinstance(m.fixed_cost, pyo.Expression)
+        assert isinstance(m.total_cost, pyo.Expression)
+        assert isinstance(m.LCOW, pyo.Expression)
+
+    @pytest.mark.unit
+    def test_fixing_operations(self, system_frame):
+        m, price_data, peak_hours = system_frame
+        fs.fix_operations_for_first_four_days(m, peak_hours=peak_hours)
```
jakechurchi, line 285: Is it worth adding assertions for this? Feel like I'm missing the point of those assert statements. If there is no error on calling the function, that means those expressions/vars/constraints were successfully created, right?

tutorials/flex_ro/WRD_schematic.png
jakechurchi, line 1: This is a pretty rough diagram, but I want to establish the treatment steps, the parallel ro trains, and the costed flows.

watertap/flowsheets/flex_desal/tests/test_flex_flow.py
adam-a-a: If there is a failure to converge to a solution, then it is usually because of an issue with initialization or scaling. It could also potentially have to do with problems in the formulation, but since you say the solve passes on Windows, I'd assume the former.

jakechurchi: Did some testing with Kurby- It's a problem with the watertap solver, which just rejects the problem because it includes binary variables. Going to change to the IDAES solver. 
The point of this "solve" is not to generate a real solution, but to allow the post_solve functions to be tested.

tutorials/flex_ro/flex_flow.ipynb
```diff
@@ -0,0 +1,654 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def run_model(result_name=\"wrd_result\"):\n",
+    "    m = build_model()\n",
+    "    add_constraints(m)\n",
+    "    add_expressions(m)\n",
+    "    add_flow_changes_penalty(m)\n",
+    "    add_objective(m)\n",
+    "    # TODO: Address how we want to handle testing the notebook without an appropriate solver.\n",
```
jakechurchi, line 535: For both Pricetaker tutorial notebooks, the default solver is gurobi, which will not work with testing. So instead I have it loading the ipopt solver as a back-up. 
Don't like this being in the tutorial, but at least it will run without any other solvers.

tutorials/flex_ro/flex_flow.ipynb
```diff
@@ -0,0 +1,654 @@
...
+    "    # TODO: Double check these baseline numbers\n",
+    "    fs.calculate_flexibility_metrics(\n",
+    "            m,\n",
+    "            baseline_power=1080,\n",
+    "            baseline_electricity_cost=50843,  # $\n",
+    "            baseline_replacement_cost=992,\n",
+    "        ),\n",
+    "    \n",
+    "    design_var_values = m.get_design_var_values()\n",
+    "    # The flowrate change penalty variables are added to m, but are not design variables!\n",
```
jakechurchi, line 480: Is there a cleaner way to implement this by not putting the flowchange penalty variables directly on m?

tutorials/flex_ro/flex_flow.ipynb
```diff
@@ -0,0 +1,654 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "1b791b94",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def solve_PT_model(m):\n",
+    "    # These could also be inputs to the function instead\n",
+    "    solver_name = \"gurobi\"\n",
```
jakechurchi, line 417: I am working on testing these files with gurobi, but not the other solvers. Should that be done, or just delete them as options?

watertap/flowsheets/flex_desal/params.py
```diff
@@ -265,6 +269,11 @@ class FlexDesalParams:
     include_battery: bool = False
     include_onsite_solar: bool = False
     onsite_capacity: float = 0
+    # Other parameters not used in tutorial, but have related functions in wrd_flowsheet.py
+    nonworking_hours: list[int] = field(default_factory=list)
+    rainy_days: int = None
```
jakechurchi, line 274: This parameter might be is specific to the WRD case, but I'm not using it in the tutorial. I will keep it in for now though.

watertap/flowsheets/flex_desal/params.py
```diff
@@ -273,6 +282,9 @@ def __post_init__(self):
         self.posttreatment = PosttreatmentParams()
         self.brinedischarge = BrineDischargeParams()
         self.battery = Battery()
+        # Only used for second tutorial
+        self.wrd_uf = WRD_UFParams()
+        self.wrd_ro = WRD_ROParams()
```
jakechurchi, line 287: Documentation note- Adding these here would be an extra step to implement a new unit model

watertap/flowsheets/flex_desal/params.py
```diff
@@ -288,3 +300,94 @@ def __post_init__(self):
...
+        """Returs the coefficients of the surrogate model as a dictionary"""
+        return {
+            "a": self.surrogate_a,
+            "b": self.surrogate_b,
+            "c": self.surrogate_c,
+        }
+
+    def get_energy_intensity(self, flowrate):
+        """Returns the energy intensity for a given flowrate"""
+        # Placeholder implementation
```
jakechurchi: Probably should test if the surrogate_type with quadratic energy intensity works... could be an additional test in the testing file. Or it could be deleted if that's a cleaner solution.
Only allowing the PYSMO surrogate option?

watertap/flowsheets/flex_desal/params.py
```diff
@@ -288,3 +300,94 @@ def __post_init__(self):
...
+        coeffs = self.surrogate_coeffs
+        if self.surrogate_type == "quadratic_energy_intensity":
+            return coeffs["a"] + coeffs["b"] * flowrate + coeffs["c"] * flowrate**2
+
+        return None
```
jakechurchi: In contrast to the other RO model, finding the minimum energy intensity and the bounds on the energy intensity is not performed in this unit model. Because the flowrate can go to 0, those functions cause issues. Also, very unclear to me why they are there in the first place? To provide bounds on the SEC? Doubt that improves solve time much.

watertap/flowsheets/flex_desal/unit_models.py
```diff
@@ -76,6 +77,20 @@ def intake_operation_model(blk, params: um_params.IntakeParams):
     _add_required_variables(blk)
     blk.recovery.fix(params.get_recovery)
     blk.energy_intensity.fix(params.energy_intensity)
+    blk.feed_cost = Var(within=NonNegativeReals, doc="Cost of feed water")
+    if params.feed_cost is not None:
```
jakechurchi, line 83: These costs are optional inputs

watertap/flowsheets/flex_desal/wrd_unit_models.py
```diff
@@ -0,0 +1,313 @@
...
+            expr=blk.reject_flowrate
+            >= blk.op_mode * (3 * pyunits.m**3 / pyunits.hr * 15),
+            doc="Minimum reject flowrate when operating (linearised via reject_flowrate variable)",
+        )
+
+    else:
+        raise ValueError("Unrecognized surrogate type")
+
+
+def wrd_reverse_osmosis_operation_model(blk, params: um_params.WRD_ROParams):
```
jakechurchi, line 92: These unit models could potentially be moved back into `unit_models.py`, but are in a separate file to emphasize they are for the WRD case study. Arguably, though, the functions are more general because they allow flexibility in terms of recovery and flowrate.

watertap/flowsheets/flex_desal/wrd_unit_models.py
```diff
@@ -0,0 +1,313 @@
...
+    )
+
+    # symmetry breaking for >1 skid. Skids can only operate if the previous skid is on
+    @blk.Constraint(blk.set_ro_skids)
+    def symmetry_breaking_cuts_ro(b, index):
+        if index == 1:
+            return Constraint.Skip
+        return b.ro_skid[index].op_mode <= b.ro_skid[index - 1].op_mode
+
+    # Also add one for the flowrate itself
```
jakechurchi, line 160: IRL it doesn't have to be the same 2 RO trains on. But it's identical in terms of energy usage.

watertap/flowsheets/flex_desal/wrd_unit_models.py
```diff
@@ -0,0 +1,313 @@
...
+
+    @blk.Constraint(blk.set_min_operating_skids)
+    def minimum_ro_skids_shutdown(b, index):
+        return b.ro_skid[index].shutdown == b.ro_skid[1].shutdown
+
+    for skid in blk.set_ro_skids:
+        # Bounds on recovery
+        blk.ro_skid[skid].recovery.setlb(params.minimum_recovery)
+        blk.ro_skid[skid].recovery.setub(params.maximum_recovery)
+        # Add an upper bound for flow to help solver
```
jakechurchi, line 186: No lower bound because that would be 0 (when train is off)

watertap/flowsheets/flex_desal/wrd_unit_models.py
```diff
@@ -0,0 +1,313 @@
...
+    blk.coeffs = Param(["a", "b", "c"], initialize=params.surrogate_coeffs)
+
+    blk.op_flow_limits_lower = Constraint(
+        expr=blk.feed_flowrate >= blk.op_mode * params.minimum_flowrate,
+        doc="Enforce minimum flowrate when operating",
+    )
+    blk.op_flow_limits_upper = Constraint(
+        expr=blk.feed_flowrate <= blk.op_mode * params.maximum_flowrate,
+        doc="Enforce maximum flowrate when operating",
+    )
```
jakechurchi, line 53: These constraints are a super key difference between this and the previous ro skid model

watertap/flowsheets/flex_desal/wrd_unit_models.py
```diff
@@ -0,0 +1,313 @@
...
+        blk.energy_surrogate = SurrogateBlock()
+        blk.energy_surrogate.build_model(
+            energy_surrogate,
+            input_vars=[blk.recovery, blk.feed_flowrate],  # RR,
+            output_vars=[blk.energy_intensity],
+        )
+
+        # reject_flowrate = feed - product = feed*(1-recovery), already a linear
+        # variable via mass_balance.  Using it here avoids the bilinear product
+        # feed*(1-recovery)
```
jakechurchi, line 81: This should be included in the documentation somewhere probably

watertap/flowsheets/flex_desal/wrd_ro_flowsheet.py
```diff
@@ -0,0 +1,857 @@
...
+
+    if params.include_onsite_solar:
+        blk.power_generation = OperationModel(
+            model_func=um.power_generation_operation_model,
+            model_args={"params": params},
+        )
+        blk.net_power_consumption += -blk.power_generation.power_utilized
+
+        # IF BATTERY is included, NOT SURE THIS VARIABLE IS NEEDED!
+        blk.excess_solar_power = Var(
```
jakechurchi, line 203: I had to add in this variable- otherwise there was a problem with the energy balance (presence of solar energy would prevent shutdowns)

watertap/flowsheets/flex_desal/wrd_ro_flowsheet.py
```diff
@@ -0,0 +1,857 @@
...
+            * blk.period[d, t].power_from_grid
+            * params.num_months
+        )
+
+    m.calculate_fixed_monthly_cost = Constraint(
+        expr=m.fixed_monthly_cost == params.fixed_monthly_cost * params.num_months
+    )
+
+
+def add_flow_costs(m):
```
jakechurchi, line 334: New function, prime for documentation

watertap/flowsheets/flex_desal/wrd_ro_flowsheet.py
```diff
@@ -0,0 +1,857 @@
...
+        expr=m.params.timestep_hours
+        * (
+            sum(m.period[:, :].intake.chemical_cost)
+            + sum(m.period[:, :].posttreatment.chemical_cost)
+        ),
+        doc="Total cost of chemicals over the time horizon ($)",
+    )
+
+
+def add_flow_changes_penalty_binary(m):
```
jakechurchi: New- This is probably the most complicated addition, both in implementation and in justification

watertap/flowsheets/flex_desal/wrd_ro_flowsheet.py
```diff
@@ -0,0 +1,857 @@
...
+                    for d in m.set_days
+                )
+                for i in range(1, m.params.wrd_uf.num_uf_pumps + 1)
+            )
+        )
+    )
+
+
+def calculate_replacement_costs(m):
+    """Adds expression for the replacement cost"""
```
jakechurchi, line 468: New, def needs documentation

watertap/flowsheets/flex_desal/wrd_ro_flowsheet.py
```diff
@@ -0,0 +1,857 @@
...
+                    * m.params.num_months
+                    / 12
+                    for i, replacement_type in enumerate(params.replacement_types)
+                )
+            ),
+            doc="Total replacement costs annualized over the time horizon",
+        )
+
+
+def calculate_flexibility_metrics(
```
jakechurchi, line 517: New, needs to be included in the documentation

watertap/flowsheets/flex_desal/wrd_ro_flowsheet.py
```diff
@@ -0,0 +1,857 @@
...
+    print(f"Levelized Cost of Flexibility ($/kWh): {LVOF:.2f}")
+    # Levelized Cost of Flexibility. Only flexibility costs are the replacement costs
+
+    m.maximum_power = Var(initialize=maximum_power)
+    m.energy_capacity = Var(initialize=discharge_energy_capacity)
+    m.power_capacity = Var(initialize=discharge_power_capacity)
+    m.LVOF = Var(initialize=LVOF)
+
+
+def add_useful_expressions(m):
```
jakechurchi, line 595: Highly consider expanding or deleting this

watertap/flowsheets/flex_desal/wrd_ro_flowsheet.py
```diff
@@ -0,0 +1,857 @@
...
+
+    @m.Constraint()
+    def match_train_1_at_start_and_end(blk):
+        return (
+            blk.period[first_point].reverse_osmosis.ro_skid[1].op_mode
+            == blk.period[last_point].reverse_osmosis.ro_skid[1].op_mode
+        )
+
+
+def fix_operations_for_first_four_days(m, peak_hours=None):
```
jakechurchi, line 623: Also not used in tutorial example? Review the comments on this function

watertap/flowsheets/flex_desal/wrd_ro_flowsheet.py
```diff
@@ -0,0 +1,857 @@
...
+                    0
+                )  # 4th skid off during peak hours. If 0 flex skids, forces this train off. But that should be ok for cases we are looking at.
+            else:
+                # Just ensure plant is on during the non-peak hours
+                m.period[d, p].reverse_osmosis.ro_skid[1].op_mode.fix(
+                    1
+                )  # Plant must be on
+
+
+### NOT USED IN THE TUTORIAL EXAMPLE - That might mean they aren't being tested? ###
```
jakechurchi: Revisit these functions after drafting the documentation and having some working tests

jakechurchi (commented):
Commenting to guide changes + make future reviewing easier

watertap/flowsheets/flex_desal/tests/test_flex_flow.py
```diff
@@ -21,7 +21,7 @@
 from watertap.flowsheets.flex_desal import wrd_ro_flowsheet as fs
 from watertap.flowsheets.flex_desal import utils
 from watertap.flowsheets.flex_desal.params import FlexDesalParams
-from watertap.core.solvers import get_solver
+from idaes.core.solvers import get_solver
```
jakechurchi, line 24: Important to note WATERTAP `get_solver` will fail on Mac and Linux!

watertap/flowsheets/flex_desal/wrd_unit_models.py
```diff
@@ -0,0 +1,313 @@
+from idaes.apps.grid_integration import OperationModel
```
kurbansitterley: Need to add header. And might as well group imports properly while you're at it.

watertap/flowsheets/flex_desal/tests/test_pricetaker_WRD.py
kurbansitterley: why did you add it?

watertap/flowsheets/flex_desal/tests/test_flex_flow.py
kurbansitterley: Is there a reason to piece the build out like this rather than just having a single build function that does all the `.add_{stuff}` and returns the fully constructed model?

watertap/flowsheets/flex_desal/tests/test_flex_flow.py
kurbansitterley: these notebooks aren't being tested yet but this will be an issue once they are

kurbansitterley (commented):
This is a totally new tutorial and flowsheet rather than an update to the existing tutorial, correct?

Because this PR would add two flex desal flowsheets and tutorials in the repository, before I go much further I'd like to see some changes to organization.
- the two flowsheets/tutorials should be in their own directories within `*/flex_ro` directories with their own assets/utility files
- within their own directories any data/jsons/pngs should be in an assets folder or something (I know that other tutorials don't follow this convention but I think they should in a future mod)
- in general I am not wild about `wrd_` being prepended to all the WRD-inspired additions. It is not really descriptive to what the files are and will be lost on a new user approaching WaterTAP tutorials. I don't have any good suggestions atm, but - what is the primary difference between the two flowsheets? 
- our other tutorials have more descriptive names (sorta the same point as the last bullet) so I'd like to see `mp_model` and `wrd_model` changed at minimum

I know we are sorta on a time crunch so interested in what @adam-a-a thinks here.

jakechurchi:
> * in general I am not wild about `wrd_` being prepended to all the WRD-inspired additions. It is not really descriptive to what the files are and will be lost on a new user approaching WaterTAP tutorials. I don't have any good suggestions atm, but - what is the primary difference between the two flowsheets?
> * our other tutorials have more descriptive names (sorta the same point as the last bullet) so I'd like to see `mp_model` and `wrd_model` changed at minimum

This is a good point. I would say the biggest difference is that energy intensity is a function of flowrate in the WRD case. So naming convention could be "flex recovery" (for original tutorial) and "flex flow" for the WRD one? 
Though WRD is flexible w.r.t. both flowrate and recovery.

Will wait for feedback before renaming everything though.

watertap/flowsheets/flex_desal/tests/test_pricetaker_WRD.py
jakechurchi: When I first started using pricetaker, the watertap-dev requirements were still on idaes 2.8 an I had to manually upgrade. IDK what the default is now.

watertap/flowsheets/flex_desal/tests/test_flex_flow.py
jakechurchi: For testing purposes this division it helps identify where issues would arise. Also, some of the add functions are optional (ex. `add_working_hours_constraint`) so I wouldn't want that in the build function.

watertap/flowsheets/flex_desal/tests/test_pricetaker_WRD.py
jakechurchi: After installing a new environment, it has updated idaes, so this test can be deleted

watertap/flowsheets/flex_desal/tests/test_flex_flow.py
jakechurchi: The ones that could be included in the build would probably be `add_demand_and_fixed_costs`, `add_useful_expressions`, and `constrain_water_production`.

watertap/flowsheets/flex_desal/utils.py
jakechurchi, line 31: If my understanding is correct, this test can be deleted because gurobi can deal with general nonlinear constraints?

jakechurchi, line 55: This would need to be like `m.params.ro.surrogate_type`

tutorials/flex_ro/flex_flow.ipynb
jakechurchi: I have successfully tested with gurobi and the xfails pass

watertap/flowsheets/flex_desal/tests/test_flex_flow.py
jakechurchi: We will want #1824 to be merged before this. And potentially we should address #1835 in this PR? Or rename this "adding second pricetaker tutorial". Then a later PR can address the issue in the existing notebook

watertap/flowsheets/flex_desal/params.py
```diff
@@ -90,6 +90,8 @@ class IntakeParams(UnitParams):
     minimum_flowrate: float = 1063.5
     nominal_flowrate: float = 1063.5
     maximum_flowrate: float = 1063.5
+    feed_cost: float = None  # in $/m3
+    chemical_cost: float = None  # in $/m3
```
MuktaHardikar, line 94: I think we should add chemical cost to pretreatment as well as a general change

watertap/flowsheets/flex_desal/params.py
MuktaHardikar, line 274: I would name this as something general like 'shutdown_days'. Could you clarify what CAPEX_yr is and why its needed?

watertap/flowsheets/flex_desal/params.py
```diff
@@ -288,3 +300,76 @@ def __post_init__(self):
         self.num_hours = total_num_seconds / 3600
         self.num_days = self.num_hours / 24
         self.num_months = self.num_days / 31
+
+
+@dataclass
+class WRD_ROParams(UnitParams):
```
MuktaHardikar, line 306: I think it makes sense to create a `flex_flow_and_recovery_params` that inherits from FlexDesalParams and overrides the __post_init__ based on the case study? That way we have case study wise param files.

watertap/flowsheets/flex_desal/utils.py
```diff
@@ -136,3 +156,254 @@ def get_baseline_model(m):
         bm.fix_operation_var("battery.power_discharge", 0)
 
     return bm
+
+
+def wrd_plot_function(m, n_time_points, output_stem=None, peak_hours=None):
```
MuktaHardikar, line 161: I suggest a case wise utils file as well. This is not really a general plot function and there are also case study specific constraints

MuktaHardikar:
I think the added tutorial looks good overall. There are a couple of typos and the link to the other tutorial needs to be updated
High level suggestions:
- I think we should remove 'wrd' from the filenames and the files should be named `flex_recovery` for the original tutorial and `flex_flow_and_recovery` for the new tutorial. Be sure to make these changes even in the watertap/flowsheets folder
- I agree with @kurbansitterley that each tutorial should live in their own folder within `flex_desal` and have their respective datafiles also stored within another `assets` folder.
- Within the watertap/flowsheets folder, I think the changes made to `flowsheet.py` are good. But having a case study specific `params` and `utils` file might make more sense. We could inherit from the general structure of these files in their case study specific version instead of directly modifying it.

watertap/flowsheets/flex_desal/wrd_ro_flowsheet.py
```diff
@@ -0,0 +1,807 @@
...
+            within=NonNegativeReals,
+            units=pyunits.kW,
+            doc="Excess solar power that is generated but not utilized",
+        )
+        blk.net_power_consumption += (
+            blk.excess_solar_power
+        )  # Slack variable so that when plant power is zero, the solar has somewhere to go.
+
+    if params.include_battery:
+        blk.battery = OperationModel()
```
MuktaHardikar, line 213: Do you need to pass a model_func here?

watertap/flowsheets/flex_desal/wrd_ro_flowsheet.py
```diff
@@ -0,0 +1,807 @@
...
+    blk.overall_power_balance = Constraint(
+        expr=blk.power_from_grid == blk.net_power_consumption,
+        doc="Computes the total power purchased from the grid",
+    )
+
+    # Add cost expressions
+    add_operational_cost_expressions(blk, params)
+
+
+def add_delayed_startup_constraints(m):
```
MuktaHardikar, line 231: Are these constraints relevant to WRD?

watertap/flowsheets/flex_desal/wrd_ro_flowsheet.py
```diff
@@ -0,0 +1,807 @@
...
+    @m.Constraint()
+    def match_train_1_at_start_and_end(blk):
+        return (
+            blk.period[first_point].reverse_osmosis.ro_skid[1].op_mode
+            == blk.period[last_point].reverse_osmosis.ro_skid[1].op_mode
+        )
+
+
+### NOT USED IN THE TUTORIAL EXAMPLE - That might mean they aren't being tested? ###
+def fix_operations_for_first_four_days(m, peak_hours=None):
```
MuktaHardikar, line 623: Drop this?

watertap/flowsheets/flex_desal/wrd_ro_flowsheet.py
```diff
@@ -0,0 +1,807 @@
...
+        return (
+            sum(
+                blk.period[period_points[k]].reverse_osmosis.ro_skid[1].shutdown
+                for k in range(w, w + window_steps)
+            )
+            <= params.max_daily_shutdowns
+        )
+
+
+def restrict_flexible_trains(m, num_flexible_trains):
```
MuktaHardikar, line 713: Some documentation here would be helpful

watertap/flowsheets/flex_desal/params.py
jakechurchi: Done

watertap/flowsheets/flex_desal/params.py
jakechurchi: I've created two different param files. However, they are completely separate, don't inherit from each other at the moment.

watertap/flowsheets/flex_desal/utils.py
jakechurchi: Done

watertap/flowsheets/flex_desal/wrd_ro_flowsheet.py
jakechurchi: I have documentation in a separate branch at the moment. I was waiting until for feedback before going all in on documentation. It could be included in this PR, but my original intention was to have it as a separate PR.

Issue #1823, opened 2026-06-22
Some notebooks are not being tested


kurbansitterley:
### Description

Notebooks are tested with 

```pytest --nbmake --nbmake-kernel="watertap-dev" **/*.ipynb```

This will only run notebooks that are in tutorials directory, but there are a few that are in subdirectories of tutorials. See #1822 

### Expected Behavior

We test all notebooks in the repository.

### Environment

- OS:
- Python:
- WaterTAP:

ksbeattie:
Superseded by #1835

PR #1824, opened 2026-06-22
Preliminary fix for notebook testing issue

MarcusHolly:
## Fixes/Resolves:

#1823 

## Summary/Motivation:
Ensures that all notebooks in the tutorials folder and any nested directories inside the tutorials folder

- Updates the checks.yml line relevant to .ipynb files

### Legal Acknowledgement


tutorials/flex_ro/mp_model.ipynb
```diff
@@ -33,7 +33,12 @@
     "import pyomo.environ as pyo\n",
     "from watertap.flowsheets.flex_desal import flowsheet as fs\n",
     "from watertap.flowsheets.flex_desal import utils\n",
-    "from watertap.flowsheets.flex_desal.params import FlexDesalParams"
+    "from watertap.flowsheets.flex_desal.params import FlexDesalParams\n",
+    "\n",
+    "# Skip tests if gurobi is not installed\n",
+    "if not pyo.SolverFactory(\"gurobi\").available():\n",
```
kurbansitterley: This will certainly fix the test failure. But it will solve locally for me with the idaes solver - is it worth just having it solve in the notebook? Otherwise we have a tutorial that is not solvable without a commercial solver. 

A user that is looking at a pricetaker tutorial _might_ have gurobi available, so I suppose I am ok with this... or maybe- it isn't worth solving without gurobi since it would give nonsense results?

MarcusHolly: I think defaulting to the IDAES solver would be ideal. How long did it take to solve?

MarcusHolly: The IDAES solver and WaterTAP solver both appear to give similar results, but they are noticeably different from the Gurobi results:

WaterTAP Results: 
image


Gurobi Results: 
image

kurbansitterley: yeah that makes sense if there are discrete variables in the model formulation

kurbansitterley: also the watertap solver did not work locally for me (Mac) so might run into issues with linux tests if we go with the watertap solver

adam-a-a:
@MarcusHolly I think we can either (a) skip the test that calls gurobi or (b) change to a solver that can handle mixed integers and is open source. IPOPT shouldn't be used if we have binary/integer variables.

For (a) can we inject something like the following?
```

@pytest.mark.skipif(
    not SolverFactory("gurobi").available(),
    reason="Gurobi not available"
)
```

MarcusHolly:
> @MarcusHolly I think we can either (a) skip the test that calls gurobi or (b) change to a solver that can handle mixed integers and is open source. IPOPT shouldn't be used if we have binary/integer variables.
> 
> For (a) can we inject something like the following?
> 
> ```
> 
> @pytest.mark.skipif(
>     not SolverFactory("gurobi").available(),
>     reason="Gurobi not available"
> )
> ```

That is similar to what I've tried before, but I didn't use the pytest mark. I will try this

kurbansitterley:
> @MarcusHolly I think we can either (a) skip the test that calls gurobi or (b) change to a solver that can handle mixed integers and is open source. IPOPT shouldn't be used if we have binary/integer variables.
> 
> For (a) can we inject something like the following?
> 
> ```
> 
> @pytest.mark.skipif(
>     not SolverFactory("gurobi").available(),
>     reason="Gurobi not available"
> )
> ```

This works and I think is almost exactly what @MarcusHolly had in #[da0f797](https://github.com/watertap-org/watertap/pull/1824/commits/da0f7971b5598d10d327bfcc6601f65e12f10330)

If we are going to require gurobi to use the notebook, that should probably be made clear somewhere at the top of the notebook?

adam-a-a:
> > @MarcusHolly I think we can either (a) skip the test that calls gurobi or (b) change to a solver that can handle mixed integers and is open source. IPOPT shouldn't be used if we have binary/integer variables.
> > For (a) can we inject something like the following?
> > ```
> > 
> > @pytest.mark.skipif(
> >     not SolverFactory("gurobi").available(),
> >     reason="Gurobi not available"
> > )
> > ```
> 
> This works and I think is almost exactly what @MarcusHolly had in #[da0f797](https://github.com/watertap-org/watertap/pull/1824/commits/da0f7971b5598d10d327bfcc6601f65e12f10330)
> 
> If we are going to require gurobi to use the notebook, that should probably be made clear somewhere at the top of the notebook?

I agree- we should add a markdown cell that clearly states which solvers can be used (and which are open source vs commercial). Double-check with @radhakrishnatg 

I remember we had things set up to solve with gurobi as well as barron, cplex, and scip at one point, but my memory is foggy. @radhakrishnatg would know best

tutorials/flex_ro/mp_model.ipynb
adam-a-a: If we were solving the model with only continuous modulation and not discrete shutdown, we could use ipopt. However, if we are including modular shutdown, we cannot.

MarcusHolly:
> > > @MarcusHolly I think we can either (a) skip the test that calls gurobi or (b) change to a solver that can handle mixed integers and is open source. IPOPT shouldn't be used if we have binary/integer variables.
> > > For (a) can we inject something like the following?
> > > ```
> > > 
> > > @pytest.mark.skipif(
> > >     not SolverFactory("gurobi").available(),
> > >     reason="Gurobi not available"
> > > )
> > > ```
> > 
> > 
> > This works and I think is almost exactly what @MarcusHolly had in #[da0f797](https://github.com/watertap-org/watertap/pull/1824/commits/da0f7971b5598d10d327bfcc6601f65e12f10330)
> > If we are going to require gurobi to use the notebook, that should probably be made clear somewhere at the top of the notebook?
> 
> I agree- we should add a markdown cell that clearly states which solvers can be used (and which are open source vs commercial). Double-check with @radhakrishnatg
> 
> I remember we had things set up to solve with gurobi as well as barron, cplex, and scip at one point, but my memory is foggy. @radhakrishnatg would know best

image

MarcusHolly:
> > > > @MarcusHolly I think we can either (a) skip the test that calls gurobi or (b) change to a solver that can handle mixed integers and is open source. IPOPT shouldn't be used if we have binary/integer variables.
> > > > For (a) can we inject something like the following?
> > > > ```
> > > > 
> > > > @pytest.mark.skipif(
> > > >     not SolverFactory("gurobi").available(),
> > > >     reason="Gurobi not available"
> > > > )
> > > > ```
> > > 
> > > 
> > > This works and I think is almost exactly what @MarcusHolly had in #[da0f797](https://github.com/watertap-org/watertap/pull/1824/commits/da0f7971b5598d10d327bfcc6601f65e12f10330)
> > > If we are going to require gurobi to use the notebook, that should probably be made clear somewhere at the top of the notebook?
> > 
> > 
> > I agree- we should add a markdown cell that clearly states which solvers can be used (and which are open source vs commercial). Double-check with @radhakrishnatg
> > I remember we had things set up to solve with gurobi as well as barron, cplex, and scip at one point, but my memory is foggy. @radhakrishnatg would know best
> 
> image

I added a cell at the top that recommends having a gurobi license, but I can add the other solvers (shown above) to that note as well if we think it'd be a good idea. But perhaps Radhakrishna should speak to this.

kurbansitterley (approved):

jakechurchi (approved):
I think the exception raised in #1835 could / should be addressed in #1822 so that the tutorial notebooks are actually tested. Or it could be a separate PR that gets merged before #1822.

PR #1836, opened 2026-06-26
Make RO compatible with MCAS

adam-a-a:
## Fixes/Resolves:

-
-

### Legal Acknowledgement

### DIFF
--- a/watertap/unit_models/reverse_osmosis_base.py
+++ b/watertap/unit_models/reverse_osmosis_base.py
@@ -246,6 +246,19 @@ class ReverseOsmosisBaseData(InitializationMixin, UnitModelBlockData):
                 ]
             )
 
+        # electroneutrality constraint for permeate
+        @self.Constraint(
+            self.flowsheet().config.time,
+            doc="Electroneutrality constraint for permeate",
+        )
+        def eq_electroneutrality_permeate(b, t):
+            return 0 == sum(
+                b.mixed_permeate[t].params.get_component(j).config.charge
+                * b.mixed_permeate[t].flow_mol_phase_comp[p, j]
+                for p, j in b.mixed_permeate[t].phase_component_set
+                if j in b.mixed_permeate[t].params.ion_set
+            )
+
         self._add_flux_balance()
         if self.config.has_pressure_change:
             self._add_deltaP()
--- a/watertap/unit_models/reverse_osmosis_base.py
+++ b/watertap/unit_models/reverse_osmosis_base.py
@@ -247,18 +247,19 @@ class ReverseOsmosisBaseData(InitializationMixin, UnitModelBlockData):
             )
 
         if hasattr(self.mixed_permeate[0].params, "ion_set"):
-            # electroneutrality constraint for permeate
-            @self.Constraint(
-                self.flowsheet().config.time,
-                doc="Electroneutrality constraint for permeate",
-            )
-            def eq_electroneutrality_permeate(b, t):
-                return 0 == sum(
-                    b.mixed_permeate[t].params.get_component(j).config.charge
-                    * b.mixed_permeate[t].flow_mol_phase_comp[p, j]
-                    for p, j in b.mixed_permeate[t].phase_component_set
-                    if j in b.mixed_permeate[t].params.ion_set
+            if len(self.mixed_permeate[0].params.ion_set)> 1:
+                # electroneutrality constraint for permeate
+                @self.Constraint(
+                    self.flowsheet().config.time,
+                    doc="Electroneutrality constraint for permeate",
                 )
+                def eq_electroneutrality_permeate(b, t):
+                    return 0 == sum(
+                        b.mixed_permeate[t].params.get_component(j).config.charge
+                        * b.mixed_permeate[t].flow_mol_phase_comp[p, j]
+                        for p, j in b.mixed_permeate[t].phase_component_set
+                        if j in b.mixed_permeate[t].params.ion_set
+                    )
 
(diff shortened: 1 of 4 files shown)


watertap/unit_models/tests/test_reverse_osmosis_1D.py
```diff
@@ -1087,3 +1093,179 @@ def configure(self):
...
+    m.fs.unit.B_comp.unfix()
+    m.fs.unit.rejection_phase_comp[0, "Liq", "Na_+"].fix(0.9)
+
+    for r in np.linspace(0.9, 0.99, 10):
+        m.fs.unit.rejection_phase_comp[0, "Liq", "Na_+"].fix(r)
+        results = solver.solve(m, tee=True)
+        assert_optimal_termination(results)
+
+
+# @pytest.mark.requires_idaes_solver
```
kurbansitterley: commented because it will fail? or need to revisit?

watertap/unit_models/reverse_osmosis_base.py
```diff
@@ -246,6 +246,21 @@ def eq_rejection_phase_comp(b, t, j):
                 ]
             )
 
+        if hasattr(self.mixed_permeate[0].params, "ion_set"):
```
kurbansitterley, line 249: would it be overkill here to type check `self.config.property_package` is MCAS?

watertap/unit_models/reverse_osmosis_base.py
```diff
@@ -246,6 +246,21 @@ def eq_rejection_phase_comp(b, t, j):
...
+            if len(self.mixed_permeate[0].params.ion_set) > 1:
```
kurbansitterley, line 250: would it even be possible to run RO with MCAS with a single ion?

watertap/unit_models/tests/test_reverse_osmosis_1D.py
```diff
@@ -1087,3 +1093,181 @@ def configure(self):
...
+    # feed_mass_frac = 0.03
+    feed_pressure = 50e5
+    feed_temperature = 273.15 + 25
+    membrane_area = 37
+    A = 4.2e-12
+    B = 3.5e-8
+    pressure_atmospheric = 101325
+    feed_mass_frac = {
+        "Ca_2+": 382e-6,
+        # "Mg_2+": 1394e-6,
```
kurbansitterley, line 1210: why no Mg?

watertap/unit_models/tests/test_reverse_osmosis_1D.py
```diff
@@ -1087,3 +1093,181 @@ def configure(self):
         self.condition_number = 7.502876 * 1e10  # Was previously 1.5101144e17
 
         return m
+
```
kurbansitterley, line 1096: do we need to test either 0D or 1D with MCAS with `dynamic=True`?

kurbansitterley (commented):
few comments for now

watertap/unit_models/reverse_osmosis_base.py
adam-a-a: I figured this should work with any prop model that has an ion set (MCAS is the only one at the moment). If I did explicitly check for MCAS, I'd still need to check that more than 1 ion was defined for electroneutrality to make sense. Now that I mention this, I was assuming that defining one ion would lead to an exception somewhere before this, but now I am not certain. It'd probably be good to raise an exception if someone defined only one ion.

watertap/unit_models/reverse_osmosis_base.py
adam-a-a: Ah, this gets at my response above. Ideally, MCAS should raise an exception if a single ion is defined.

watertap/unit_models/tests/test_reverse_osmosis_1D.py
adam-a-a: just wanted to make a simpler test more quickly :)

watertap/unit_models/tests/test_reverse_osmosis_1D.py
adam-a-a: Maybe 0d since 1d with dynamics hasn't been merged.

watertap/unit_models/reverse_osmosis_base.py
```diff
@@ -246,6 +246,21 @@ def eq_rejection_phase_comp(b, t, j):
...
+                # electroneutrality constraint for permeate
+                @self.Constraint(
+                    self.flowsheet().config.time,
+                    doc="Electroneutrality constraint for permeate",
+                )
+                def eq_electroneutrality_permeate(b, t):
+                    return 0 == sum(
+                        b.mixed_permeate[t].params.get_component(j).config.charge
```
MuktaHardikar, line 258: Do we need to this for the feed as well?

kurbansitterley (approved):

watertap/unit_models/reverse_osmosis_base.py
adam-a-a: I usually use `assert_electroneutrality` through MCAS, then fix the feed to ensure electroneutrality.

watertap/unit_models/tests/test_reverse_osmosis_1D.py
adam-a-a: revisited

MuktaHardikar (approved):
LGTM. I didn't review the test_file for RO 0D dynamic thoroughly because I'm not too familiar with it. But everything else looks alright to me and tests are passing.

PR #1846, opened 2026-07-06
BSM2 Translator & Scaling Revisions

MarcusHolly:
## Summary/Motivation:
There was a mistake in the implementation of X_PP in the modified ADM1 reaction package and ADM1-ASM2d translator

- Fixes implementation of X_PP
- Updates pyproject.toml to use latest IDAES scaling functionality
- Updates scaling in BSM2-P flowsheets

### Legal Acknowledgement


watertap/unit_models/anaerobic_digester.py
```diff
@@ -83,12 +83,18 @@ class ADScaler(CustomScalerBase):
     """
 
     DEFAULT_SCALING_FACTORS = {
-        "volume": 1e-2,
+        "volume": 1e-3,
+        "volume_AD": 1e-3,
+        "KH_h2": 1e4,
+        "KH_co2": 1e2,
+        "KH_ch4": 1e2,
+        "heat": 1e1,
```
adam-a-a, line 91: WHat is the order of magnitude of `heat`? I would assume it to be on the order of 1e5 - 1e6. Does this default of 1e1 make sense? It may be a relic that you copied over, but curious to hear your thoughts.

watertap/unit_models/tests/test_cstr_injection.py
```diff
@@ -762,215 +761,10 @@ def test_example_case_scaler(self):
...
-        )
+        scaler.scale_model(m.fs.unit)
 
         # Check condition number to confirm scaling
-        sm = TransformationFactory("core.scale_model").create_using(m, rename=False)
-        jac, _ = get_jacobian(sm, scaled=False)
+        jac, _ = get_jacobian(m, scaled=False)
         assert (jacobian_cond(jac=jac, scaled=False)) == pytest.approx(
-            1.1526931e7, rel=1e-3
+            4.99607739612e12, rel=1e-3
```
adam-a-a, line 774: The condition number got worse for this one. Why?

watertap/unit_models/tests/test_cstr.py
```diff
@@ -675,206 +675,10 @@ def test_example_case_scaler(self):
...
-        )
+        scaler.scale_model(m.fs.unit)
 
         # Check condition number to confirm scaling
-        sm = TransformationFactory("core.scale_model").create_using(m, rename=False)
-        jac, _ = get_jacobian(sm, scaled=False)
+        jac, _ = get_jacobian(m, scaled=False)
         assert (jacobian_cond(jac=jac, scaled=False)) == pytest.approx(
-            1.65862e5, rel=1e-3
+            1.20164823851e11, rel=1e-3
```
adam-a-a, line 689: Condition number worse. Why? Since this is in both cstr and cstr with injection, I am assuming we can make an adjustment to scaling in cstr that will cascade over into cstr_injection.

watertap/unit_models/tests/test_clarifier.py
```diff
@@ -556,181 +555,10 @@ def test_example_case_scaler(self):
...
-        )
+        scaler.scale_model(m.fs.unit)
 
         # Check condition number to confirm scaling
-        sm = TransformationFactory("core.scale_model").create_using(m, rename=False)
-        jac, _ = get_jacobian(sm, scaled=False)
+        jac, _ = get_jacobian(m, scaled=False)
         assert (jacobian_cond(jac=jac, scaled=False)) == pytest.approx(
-            2729.85, rel=1e-3
+            9.364726e3, rel=1e-3
```
adam-a-a, line 561: Hard to see why the condition number got worse here and why it improved above.

watertap/unit_models/tests/test_cstr_injection.py
MarcusHolly: I wouldn't be overly concerned with the condition number we see in the unit model testing. Previously, I had set the default scaling factors for the unit models to roughly minimize the condition number for that unit model in isolation (which is the number we previously saw here). However, I've made changes to the default scaling factors for this unit model and others to limit the number of changes we need to make to the default scaling factors in the BSM2-P flowsheets. In other words, I've changed the default scaling factors to create cleaner scaling functions in the BSM2-P flowsheets (by not having to manual adjust the defaults) rather than trying to minimize the condition number of the unit in isolation.

watertap/unit_models/tests/test_anaerobic_digester.py
```diff
@@ -117,13 +117,16 @@ def build():
     m.fs.unit.volume_vapor.fix(300)
     m.fs.unit.liquid_outlet.temperature.fix(308.15)
 
+    # Set scaling factors for badly scaled variables
+    iscale.set_scaling_factor(m.fs.unit.liquid_phase.heat[0], 1e3)
+    iscale.set_scaling_factor(m.fs.unit.liquid_phase.rate_reaction_extent, 1e4)
+    iscale.set_scaling_factor(m.fs.unit.liquid_phase.volume[0], 1e-2)
```
adam-a-a: Do these align with the defaults set later in the Scaler class? Should they or good reason to deviate? Additionally, do these scaling factors make sense? I am referring mostly to `heat`. SHould double check the order of magnitude on that variable and assign an appropriate scaling factor.

watertap/unit_models/tests/test_cstr.py
MarcusHolly: See above comment

watertap/unit_models/tests/test_aeration_tank.py
```diff
@@ -640,7 +640,7 @@ def test_example_case_scaler(self):
         sm = TransformationFactory("core.scale_model").create_using(m, rename=False)
         jac, _ = get_jacobian(sm, scaled=False)
         assert (jacobian_cond(jac=jac, scaled=False)) == pytest.approx(
-            1.1526931e7, rel=1e-3
+            5.12596365e8, rel=1e-3
```
adam-a-a, line 647: Slightly worse condition number could stem from cstr scaling.

MarcusHolly: See comment on CSTR_injection test

watertap/property_models/unit_specific/anaerobic_digestion/tests/test_modified_adm1_reaction.py
```diff
@@ -109,7 +109,7 @@ def test_build(self, model):
             ]
 
         # Expected non-zero stoichiometries
-        # Values from table 11 in reference
+        # Values from Gujer matrix in reference
```
adam-a-a: Let's be sure to specify the reference precisely for record-keeping and future troubleshooting.

watertap/unit_models/anaerobic_digester.py
MarcusHolly: Heat is -2e-3 on BSM2-P an -3e-3 on BSM2. I will try changing this scaling factor to 1e3 and see what happens

watertap/property_models/unit_specific/anaerobic_digestion/tests/test_modified_adm1_reaction.py
MarcusHolly: I think this is actually an improvement in clarity. I changed it because I wasn't sure what table 11 was referring to and I'm fairly certain we directly used the Gujer matrix Excel file rather than a table in the paper/supplementary resources

watertap/unit_models/tests/test_anaerobic_digester.py
```diff
@@ -253,363 +256,273 @@ def configure(self):
...
+#         m.fs.props_vap = ADM1_vaporParameterBlock()
+#         m.fs.rxn_props = ADM1ReactionParameterBlock(property_package=m.fs.props)
+#
+#         m.fs.unit = AD(
+#             liquid_property_package=m.fs.props,
+#             vapor_property_package=m.fs.props_vap,
+#             reaction_package=m.fs.rxn_props,
+#             has_heat_transfer=True,
+#             has_pressure_change=False,
+#         )
```
adam-a-a: Are you going to revive this commented code or delete? If deleting, why?

watertap/flowsheets/full_water_resource_recovery_facility/BSM2_P_extension.py
```diff
@@ -737,18 +648,15 @@ def initialize_system(m, bio_P=False, solver=None):
     initializer = BlockTriangularizationInitializer()
 
     def function(unit):
-        # TODO: Resolve why bio_P=True does not work with the BTInitializer
-        if bio_P:
-            unit.initialize(outlvl=idaeslog.DEBUG)
-        else:
-            initializer.initialize(unit, output_level=_log.debug)
```
adam-a-a, line 744: Is this resolved now? Does bio_P=True work with BTInitializer?

watertap/flowsheets/full_water_resource_recovery_facility/BSM2_P_extension.py
```diff
@@ -737,18 +648,15 @@ def initialize_system(m, bio_P=False, solver=None):
...
     seq.run(m, function)
 
 
 def solve(m, solver=None):
     if solver is None:
         solver = get_solver()
+        solver.options["max_iter"] = 500
     results = solver.solve(m, tee=True)
     check_solve(results, checkpoint="closing recycle", logger=_log, fail_flag=True)
     pyo.assert_optimal_termination(results)
```
adam-a-a: do we need both of these? Since fail_flag = True, I don't think the assert_optimal_termination line will ever run if there is failure to converge. Also, if we keep check_solve line, should improve the checkpoint message. Alternatively, consider removing the check_solve line and just asserting optimal termination

watertap/flowsheets/full_water_resource_recovery_facility/test/test_BSM2_P_extension.py
```diff
@@ -110,58 +111,58 @@ def test_solve(self, system_frame):
...
-        ) == pytest.approx(0.0011858, rel=1e-3)
+        ) == pytest.approx(0.0008358, rel=1e-3)
         assert value(
             m.fs.Treated.properties[0].conc_mass_comp["S_PO4"]
-        ) == pytest.approx(0.703159, rel=1e-3)
+        ) == pytest.approx(0.000421, rel=1e-3)
         assert value(m.fs.Treated.properties[0].conc_mass_comp["S_K"]) == pytest.approx(
-            0.367442, rel=1e-3
+            0.3642, rel=1e-3
         )
```
adam-a-a, line 94: A change in some results, while scaling (condition number) hasn't changed much. I assume the results are changing due to the X_PP fix.

watertap/unit_models/anaerobic_digester.py
adam-a-a: Hmmm--we really should verify against previously published results. e-3 seems like low an order of magnitude, but I could be missing something. I wonder if this has anything to do with the methane issue reported.

watertap/flowsheets/full_water_resource_recovery_facility/BSM2_P_extension.py
MarcusHolly: Nope - the BTInitializer actually doesn't work regardless of what bio_P is now. It hits iteration limits in both cases. I can add the TODO note back but I'm not sure if it's really an issue or an issue worth investigating too deeply

watertap/property_models/unit_specific/anaerobic_digestion/tests/test_modified_adm1_reaction.py
adam-a-a: Maybe the "reference" is given explicitly in the file already and it didn't show up because it wasn't relevant to the diff. Can you make sure that the reference is clearly provided, along with where one would find said Gujer matrix? I am guessing that would be supplementary material from Solon et al., 2017. We could also link their material on GH.

watertap/unit_models/anaerobic_digester.py
MarcusHolly: I plan on creating a follow-up PR to hopefully address the methane issue. It's my hope that updating the operating conditions of BSM2-P to match Flores-Alsina will magically resolve the methane issue. I will need your help with running MATLAB for this PR tho. I also plan on creating another PR at some point to update the scaling in BSM2 as well as investigating the methane issue there since the I believe those operating conditions already match the literature (but I will have to double check to be sure)

watertap/property_models/unit_specific/anaerobic_digestion/tests/test_modified_adm1_reaction.py
MarcusHolly: I'll add a link to the Box file for additional clarity tho

watertap/unit_models/anaerobic_digester.py
MarcusHolly: With this being said, I think it's fine to leave this scaling factor as it is for now and revisit this in the future PR(s) that address the methane issue. This way we're not spinning our wheels trying to diagnose an issue that might resolve itself relatively soon. I agree it'd be nice to see what published results say tho.

watertap/unit_models/tests/test_anaerobic_digester.py
MarcusHolly: They line up now. I agree that it makes sense to line them up since one of the purposes of this test is to compare the new scaling method with the old scaling method (iscale).

watertap/unit_models/tests/test_anaerobic_digester.py
MarcusHolly: I just commented this out to speed up testing locally, but forgot to add back in. I've fixed this.

watertap/flowsheets/full_water_resource_recovery_facility/BSM2_P_extension.py
MarcusHolly: Removed check_solve as well as the pyo.assert_optimal_termination(results) in the main function, which was being run (redundantly) after the solve.

watertap/property_models/unit_specific/anaerobic_digestion/tests/test_modified_adm1_reaction.py
adam-a-a: We would want to link to a published reference, so that should be the source rather than a Box link that isn't publicly accessible.

MarcusHolly: That is linked at the top of this file (before this PR)

watertap/flowsheets/electroNP/electroNP_flowsheet.py
```diff
@@ -58,6 +62,70 @@
...
+    "X_ac",
+    "X_h2",
+    "X_I",
+    "X_PHA",
+    "X_PP",
+    "X_PAO",
+    "S_K",
+    "S_Mg",
+]
+
```
luohezhiming: I think we don't need to create this comp_list. It can be directly called by the properties package `component_list ` which is already defined. If you want to exclude `H2O`, I think it is supposed to be `solute_set`

watertap/flowsheets/electroNP/BSM2_electroNP_no_bioP.py
```diff
@@ -82,16 +86,71 @@
...
+    "X_ac",
+    "X_h2",
+    "X_I",
+    "X_PHA",
+    "X_PP",
+    "X_PAO",
+    "S_K",
+    "S_Mg",
+]
+
```
luohezhiming: I think we don't need to create this comp_list. It can be directly called by the properties package `component_list`  which is already defined. If you want to exclude `H2O`, I think it is supposed to be `solute_set`

watertap/flowsheets/full_water_resource_recovery_facility/BSM2_P_extension.py
```diff
@@ -469,6 +445,7 @@ def set_operating_conditions(m, bio_P=False):
     m.fs.CL2.split_fraction[0, "effluent", "X_PP"].fix(0.00187)
     m.fs.CL2.split_fraction[0, "effluent", "X_S"].fix(0.00187)
 
+    # [1]
```
luohezhiming, line 448:
```suggestion
```

watertap/property_models/unit_specific/activated_sludge/modified_asm2d_properties.py
```diff
@@ -52,6 +52,27 @@
...
+    "S_IC",
+    "X_AUT",
+    "X_H",
+    "X_I",
+    "X_PAO",
+    "X_PHA",
+    "X_PP",
+    "X_S",
+]
+
```
luohezhiming: Same issue, the` component_list ` is supposed to be called from the properties package

watertap/property_models/unit_specific/anaerobic_digestion/modified_adm1_properties.py
```diff
@@ -44,6 +44,39 @@
...
+    "X_ac",
+    "X_h2",
+    "X_I",
+    "X_PHA",
+    "X_PP",
+    "X_PAO",
+    "S_K",
+    "S_Mg",
+]
+
```
luohezhiming: Same suggestion as above

watertap/property_models/unit_specific/anaerobic_digestion/modified_adm1_reactions.py
```diff
@@ -1764,13 +1764,31 @@ class ModifiedADM1ReactionScaler(CustomScalerBase):
...
+        "conc_mol_Mg": 1e5,
+        "conc_mol_K": 1e2,
+        "conc_mol_nh3": 1e3,
+        "conc_mol_co2": 1e3,
+        "conc_mass_va": 1e4,
+        "conc_mass_bu": 1e4,
+        "conc_mass_pro": 1e3,
+        "conc_mass_ac": 1e3,
+        "S_H": 1e8,
     }
```
luohezhiming, line 1769: I would suggest not add these scaling factors here, I am not sure why some concentrations are conc_mol and some are conc_mass, but they are all defined for a specific flowsheet, I'd suggestto add these scaling factors in the flowsheet level. Otherwise, once the influent or the diagram changes, these value makes no sense and might make another flowsheet harder to solve.

watertap/unit_models/translators/translator_adm1_asm2d.py
```diff
@@ -365,7 +365,7 @@ def XI_AD1(blk, t):
         )
         def SIP_AD1(blk, t):
             return (
-                blk.properties_in[t].conc_mass_comp["X_PP"] / mw_XPP
+                blk.properties_in[t].conc_mass_comp["X_PP"] / mw_p
```
luohezhiming, line 367: Just a remind, we can delete the redundant `mw_XPP` in the file for now.

watertap/unit_models/clarifier.py
```diff
@@ -53,6 +53,7 @@ class ClarifierScaler(CustomScalerBase):
     DEFAULT_SCALING_FACTORS = {
         "surface_area": 1e-3,
         "electricity_consumption": 1,
+        "split_fraction": 1e1,
```
luohezhiming, line 56: I think for split_fraction, we'd better simply set the scaling factor as 1 (1e0), cause it will always between 0 and 1. I know it might have some subsequent consequences if we modify this value, but I think we'd better to avoid the scaling error populate throught the calculation that you have to change scaling for other variables 1 order of magnitude smaller.

watertap/unit_models/electroNP_ZO.py
```diff
@@ -38,12 +40,128 @@
...
+    either through user provided options (submodel_scalers argument) or by default
+    Scalers assigned to the packages.
+    """
+
+    DEFAULT_SCALING_FACTORS = {
+        "magnesium_chloride_dosage": 1e1,
+        "MgCl2_flowrate": 1e-2,
+        "electricity": 1e-1,
+        "energy_electric_flow_mass": 1e2,
+        "split_fraction": 1e2,
```
luohezhiming: I think split_fraction scaling factor is supposed to be 1

luohezhiming (commented):
Great work. I have a few suggestions. Basically, we don't need to create extra component_list cause it is already defined for all properties package. And some scaling factors, if possible, I would avoid to change them in the original code, try to add them in the flowsheet level. The scaling sometimes help solving a specific flowhseet, but it can make another flowsheet even harder to solve if we defined it in the unit or properties package.

watertap/unit_models/electroNP_ZO.py
MarcusHolly: Change this to 1e1

watertap/unit_models/clarifier.py
MarcusHolly: Fine to keep this as 1e1

watertap/flowsheets/full_water_resource_recovery_facility/BSM2_P_extension.py
MarcusHolly: @luohezhiming Why should this reference be removed?

watertap/flowsheets/electroNP/BSM2_electroNP_no_bioP.py
```diff
@@ -409,16 +410,18 @@ def set_operating_conditions(m):
...
     m.fs.FeedWater.conc_mass_comp[0, "X_S"].fix(94.1 * pyo.units.g / pyo.units.m**3)
     m.fs.FeedWater.conc_mass_comp[0, "X_H"].fix(370 * pyo.units.g / pyo.units.m**3)
-    m.fs.FeedWater.conc_mass_comp[0, "X_PAO"].fix(
-        51.5262 * pyo.units.g / pyo.units.m**3
-    )
-    m.fs.FeedWater.conc_mass_comp[0, "X_PP"].fix(1e-6 * pyo.units.g / pyo.units.m**3)
+    m.fs.FeedWater.conc_mass_comp[0, "X_PAO"].fix(500 * pyo.units.g / pyo.units.m**3)
+    # m.fs.FeedWater.conc_mass_comp[0, "X_PAO"].fix(
+    #     51.5262 * pyo.units.g / pyo.units.m**3
+    # )
```
luohezhiming: With modification of PO4 to 15 mg/L, I am sure we don't need to modify X_PAO to 500 mg/L, keep it as 51.5262 will also work

luohezhiming (commented):
Looks good, the only thingis BSM2_electroNP_no_bioP flowheet, we don't need to change X_PAO to 500

watertap/flowsheets/electroNP/BSM2_electroNP_no_bioP.py
```diff
@@ -409,16 +410,18 @@ def set_operating_conditions(m):
...
     m.fs.FeedWater.conc_mass_comp[0, "X_I"].fix(84 * pyo.units.g / pyo.units.m**3)
     m.fs.FeedWater.conc_mass_comp[0, "X_S"].fix(94.1 * pyo.units.g / pyo.units.m**3)
     m.fs.FeedWater.conc_mass_comp[0, "X_H"].fix(370 * pyo.units.g / pyo.units.m**3)
+    # m.fs.FeedWater.conc_mass_comp[0, "X_PAO"].fix(500 * pyo.units.g / pyo.units.m**3)
     m.fs.FeedWater.conc_mass_comp[0, "X_PAO"].fix(
         51.5262 * pyo.units.g / pyo.units.m**3
     )
-    m.fs.FeedWater.conc_mass_comp[0, "X_PP"].fix(1e-6 * pyo.units.g / pyo.units.m**3)
+    m.fs.FeedWater.conc_mass_comp[0, "X_PP"].fix(10 * pyo.units.g / pyo.units.m**3)
+    # m.fs.FeedWater.conc_mass_comp[0, "X_PP"].fix(1e-6 * pyo.units.g / pyo.units.m**3)
```
adam-a-a:
```suggestion
```

watertap/flowsheets/electroNP/BSM2_electroNP_no_bioP.py
```diff
@@ -409,16 +410,18 @@ def set_operating_conditions(m):
...
     m.fs.FeedWater.conc_mass_comp[0, "S_NH4"].fix(26.6 * pyo.units.g / pyo.units.m**3)
     m.fs.FeedWater.conc_mass_comp[0, "S_NO3"].fix(1e-6 * pyo.units.g / pyo.units.m**3)
-    m.fs.FeedWater.conc_mass_comp[0, "S_PO4"].fix(1e-6 * pyo.units.g / pyo.units.m**3)
+    m.fs.FeedWater.conc_mass_comp[0, "S_PO4"].fix(15 * pyo.units.g / pyo.units.m**3)
     m.fs.FeedWater.conc_mass_comp[0, "S_I"].fix(57.45 * pyo.units.g / pyo.units.m**3)
     m.fs.FeedWater.conc_mass_comp[0, "S_N2"].fix(25.19 * pyo.units.g / pyo.units.m**3)
     m.fs.FeedWater.conc_mass_comp[0, "X_I"].fix(84 * pyo.units.g / pyo.units.m**3)
     m.fs.FeedWater.conc_mass_comp[0, "X_S"].fix(94.1 * pyo.units.g / pyo.units.m**3)
     m.fs.FeedWater.conc_mass_comp[0, "X_H"].fix(370 * pyo.units.g / pyo.units.m**3)
+    # m.fs.FeedWater.conc_mass_comp[0, "X_PAO"].fix(500 * pyo.units.g / pyo.units.m**3)
```
adam-a-a:
```suggestion
```

adam-a-a (approved):

PR #1849, opened 2026-07-16
MCAS in SWRO Flowsheet

adam-a-a:
## Fixes/Resolves:

-
-

### Legal Acknowledgement


adam-a-a:
@sufikaur @ksbeattie this is the other PR that could potentially make it into the patch release

adam-a-a:
@sufikaur This followup PR that I was hoping to get into the patch release is actually ready! I was able to resolve the issues. There may be a couple of trivial test failures to resolve on my end (checking now). My hope is that we can still squeeze this in once we get two reviews.

watertap/flowsheets/seawater_RO_desalination/seawater_RO_desalination.py
```diff
@@ -407,16 +408,17 @@ def set_operating_conditions(m):
     # RO unit
     desal.RO.A_comp.fix(4.2e-12)  # membrane water permeability coefficient [m/s-Pa]
     desal.RO.B_comp.fix(3.5e-8)  # membrane salt permeability coefficient [m/s]
+    desal.RO.B_comp[0, "tss"].fix(1e-10)  # membrane salt permeability coefficient [m/s]
```
avdudchenko, line 411: Nice, so we can do this even without having to provide charge.

watertap/flowsheets/seawater_RO_desalination/seawater_RO_desalination.py
```diff
@@ -353,17 +318,50 @@ def set_operating_conditions(m):
     temperature = 298 * pyunits.K
```
avdudchenko, line 318: This is a minor comment. but we should probably put this into to fix_operation_function. Not critical.

watertap/flowsheets/seawater_RO_desalination/seawater_RO_desalination.py
```diff
@@ -353,17 +318,50 @@ def set_operating_conditions(m):
...
+    )
+    iscale.set_scaling_factor(
+        m.fs.feed.properties[0].conc_mass_phase_comp["Liq", "tds"],
+        value(10 / conc_mass_tds),
+    )
+    iscale.set_scaling_factor(
+        m.fs.feed.properties[0].conc_mass_phase_comp["Liq", "tss"],
+        value(10 / conc_mass_tss),
+    )
+
```
avdudchenko, line 334: I almost think, if you make this into fix_feed_conditions function, it will make a clear example of how to set feed state using concentrations and flow volume.

watertap/flowsheets/seawater_RO_desalination/seawater_RO_desalination.py
```diff
@@ -353,17 +318,50 @@ def set_operating_conditions(m):
...
+    m.fs.feed.properties.calculate_state(
+        var_args={
+            ("conc_mass_phase_comp", ("Liq", "tds")): conc_mass_tds,
+            ("conc_mass_phase_comp", ("Liq", "tss")): conc_mass_tss,
+            ("flow_vol_phase", "Liq"): flow_vol,
+        },
+        hold_state=True,
+    )
+    m.fs.properties.set_default_scaling(
```
avdudchenko, line 343: I would also put all scaling stpes into a separate function. 

this we can call build, fix, scale, initalize etc.

watertap/flowsheets/seawater_RO_desalination/seawater_RO_desalination.py
avdudchenko, line 485: not sure why we are using prtrt/psttrt they are realy confusing and do not make it easier to read the code. I suggest we just use explicity m.fs.pretreatment etc.

watertap/flowsheets/seawater_RO_desalination/seawater_RO_desalination.py
```diff
@@ -601,8 +601,8 @@ def add_costing(m):
     m.fs.costing.base_currency = pyunits.USD_2023
```
avdudchenko, line 608: This will not wor kcorrectly. The costing is already built and using default costing units loaded at build_global_params call, which happens at costing build. That means the vars are using different base_currency here then you would expect. - e.g. any vars build at global will use base currency (I think USD_2019) and any vars down stream will use 2023. This is an issue, as it means you might be down converting some of the costs rather then up based on CPI index.

avdudchenko: Ideally, we should define base cost at our desired year, and any unit costs/capex/etc have cost defined at specific year (USD_2020 or something - in many cases they just use base_currency which is what ever is the base_currecny var in the costing package. I am not sure how that impacts all equations but its inconsistent, for example your electricity price will use USD_2018 ([set here](https://github.com/watertap-org/watertap/blob/e78ca2c776d267f1eae3c20edf9f09268fd89748/watertap/costing/watertap_costing_package.py#L53), which is called [here](https://github.com/watertap-org/watertap/blob/e78ca2c776d267f1eae3c20edf9f09268fd89748/watertap/costing/watertap_costing_package.py#L503) and build_global_params is called at build by base flowshetCostingblock [here](https://github.com/IDAES/idaes-pse/blob/a60fbf8192a697c60564b68379847ce6e9dbb3b9/idaes/core/base/costing_base.py#L177) which means all vars called after [here](https://github.com/watertap-org/watertap/blob/e78ca2c776d267f1eae3c20edf9f09268fd89748/watertap/costing/watertap_costing_package.py#L511) use USD_2018, and only anything built after build_global_parms will end up using USD_2023. which means base all aggreaget costs etc on USD_2018, build units with csoitng of USD_2023 unless they specify otherwise, and cost eelctricty on USD_2018! 

yay....

avdudchenko: Unless I can't read code... and then I am wrong and should pursue my wifes dream of running a Cat farm somewhere...

avdudchenko (approved):
Overall nice. but be aware of bugged up base_currency usage here.  (just make sure the tests pass. Most of my comments are related to clearer organization of some code but its very optional)

watertap/flowsheets/seawater_RO_desalination/seawater_RO_desalination.py
kurbansitterley: > This will not wor kcorrectly. The costing is already built and using default costing units loaded at build_global_params call, which happens at costing build. That means the vars are using different base_currency here then you would expect. - e.g. any vars build at global will use base currency (I think USD_2019) and any vars down stream will use 2023. This is an issue, as it means you might be down converting some of the costs rather then up based on CPI index.

So is it correct that doing e.g., `m.fs.costing.base_currency = pyunits.USD_2023` will produce results in USD2023, but that any units that provide costing parameters in `base_currency` will use the default?

kurbansitterley: this seems like a bigger issue honestly (I think this is what you are saying). 

Seems like the only way to define `base_currency` to cover all our bases would be to do it via CONFIG

> Ideally, we should define base cost at our desired year, and any unit costs/capex/etc have cost defined at specific year (USD_2020 or something - in many cases they just use base_currency which is what ever is the base_currecny var in the costing package. I am not sure how that impacts all equations but its inconsistent, for example your electricity price will use USD_2018 ([set here](https://github.com/watertap-org/watertap/blob/e78ca2c776d267f1eae3c20edf9f09268fd89748/watertap/costing/watertap_costing_package.py#L53), which is called [here](https://github.com/watertap-org/watertap/blob/e78ca2c776d267f1eae3c20edf9f09268fd89748/watertap/costing/watertap_costing_package.py#L503) and build_global_params is called at build by base flowshetCostingblock [here](https://github.com/IDAES/idaes-pse/blob/a60fbf8192a697c60564b68379847ce6e9dbb3b9/idaes/core/base/costing_base.py#L177) which means all vars called after [here](https://github.com/watertap-org/watertap/blob/e78ca2c776d267f1eae3c20edf9f09268fd89748/watertap/costing/watertap_costing_package.py#L511) use USD_2018, and only anything built after build_global_parms will end up using USD_2023. which means base all aggreaget costs etc on USD_2018, build units with csoitng of USD_2023 unless they specify otherwise, and cost eelctricty on USD_2018!
> 
> yay....

adam-a-a: We'll need to test this to be 100% sure and then resolve across other flowsheets that might have this implementation

kurbansitterley: @adam-a-a @avdudchenko see #1860

kurbansitterley (approved):

MuktaHardikar:
@kurbansitterley Is this ready to merge? I believe #1856 was merge in here already.

adam-a-a:
As discussed with @MuktaHardikar , we'll remove the updated base currency (and correct regression tests) from this PR for now until #1864 goes through. The one other sanity check we'll want is on the UI side, since the SWRO flowsheet was intended to go into the latest UI version, with associated fixes. We should make sure that no bugs slipped through when testing in the UI (at least none that are obvious).

watertap/flowsheets/seawater_RO_desalination/seawater_RO_desalination.py
MuktaHardikar: I was checking this again (potentially a pointless exercise but just wanted to note it) and compared a couple of costing build scenarios:
1. Update base_currency to USD_2023 after building `m.fs.costing` - as it was originally
2. Update the default_case_study yaml to have the base_currency as USD_2023
3. Use the units in the default_case_study yaml which is MUSD_2018 and then convert MUSD_2018 to USD_2023

All resulted in the same LCOW value in the units USD_2023.

If I understand it correctly, in this case because we are using ZeroOrderCosting, self.electricity_cost was provided the value MUSD_2018 as base_currency from the default_case_study yaml during build. The aggregate_cost variables are created when cost_process() is called [here.](https://github.com/IDAES/idaes-pse/blob/a60fbf8192a697c60564b68379847ce6e9dbb3b9/idaes/core/base/costing_base.py#L231) by which point base_currency was updated to USD_2023. The aggregate_flow_costs are created in cost_process() which also uses the updated base_currency [here](https://github.com/IDAES/idaes-pse/blob/a60fbf8192a697c60564b68379847ce6e9dbb3b9/idaes/core/base/costing_base.py#L472). So it didn't change the LCOW results. But if anyone uses the electricity_cost variable in a flowsheet it would be with different base_currency units.

I still agree that having it as a CONFIG makes it more consistent throughout the flowsheet.

MuktaHardikar (approved):
@MarcusHolly tested the GUI on windows but it is currently failing on Mac. Otherwise LGTM!

kurbansitterley (approved):

MuktaHardikar:
The GUI for Mac works as well.

PR #1857, opened 2026-08-18
Add element conservation test for modified ASM2d

luohezhiming:
## Fixes/Resolves:

-
-

### Legal Acknowledgement

### DIFF
--- a/watertap/property_models/unit_specific/activated_sludge/tests/test_modified_asm2d_reaction.py
+++ b/watertap/property_models/unit_specific/activated_sludge/tests/test_modified_asm2d_reaction.py
@@ -57,11 +57,6 @@ solver = get_solver()
 def total_phosphorus_flow(port, rxn_props):
     """
     Total phosphorus mass flowrate (kg P/s) through a state block port
-    (e.g. a CSTR's inlet or outlet).
-
-    S_PO4 and X_PP are tracked directly as kg P/m3; the remaining
-    P-bearing components (S_F, X_I, X_S, X_H, X_PAO, X_AUT) are converted
-    using the i_P* content parameters from the reaction parameter block.
     """
     conc = port.conc_mass_comp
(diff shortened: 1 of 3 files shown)


watertap/property_models/unit_specific/activated_sludge/tests/test_modified_asm2d_reaction.py
```diff
@@ -54,6 +54,79 @@
...
+        + value(conc[0, "S_I"])
+        - (12 / 7) * value(conc[0, "S_N2"])
+        + value(rxn_props.i_COD_NOx) * value(conc[0, "S_NO3"])
+        + value(conc[0, "X_I"])
+        + value(conc[0, "X_S"])
+        + value(conc[0, "X_H"])
+        + value(conc[0, "X_PAO"])
+        + value(conc[0, "X_PHA"])
+        + value(conc[0, "X_AUT"])
+    )
```
adam-a-a, line 127: I envisioned these being actual Expressions on the property model so that for any StateBlock, I can reference one of these Expressions and obtain the value.

adam-a-a: So- the other annoying complication is that in addition to wanting these as Expression on any state block, these should technically be on the property model rather than reaction model, which would mean moving all the `i_xx` parameters to the property model as well....

I suppose that becomes painful when doing this for all biokinetic process models. Then again, the elemental calculations will be model-specific anyway.

MarcusHolly: Should also add functionality such that total mass flow (total phosphorus, total COD, etc.) can be called on any state block rather than having a helper function as done here.

MarcusHolly (approved):

PR #1858, opened 2026-08-19
Revise ASM3 reaction package

luohezhiming:
## Fixes/Resolves:

-
-

### Legal Acknowledgement

### DIFF
--- a/watertap/property_models/unit_specific/activated_sludge/asm3_reactions.py
+++ b/watertap/property_models/unit_specific/activated_sludge/asm3_reactions.py
@@ -32,6 +32,7 @@ from idaes.core import (
     ReactionBlockDataBase,
     ReactionBlockBase,
 )
+from pyomo.common.config import ConfigValue
 from idaes.core.util.misc import add_object_reference
 from idaes.core.util.exceptions import BurntToast
 import idaes.logger as idaeslog
@@ -49,6 +50,21 @@ class ASM3ReactionParameterData(ReactionParameterBlock):
     """
     Reaction Parameter Block Class
     """
+
+    CONFIG = ReactionParameterBlock.CONFIG()
+    CONFIG.declare(
+        "calibrated_params",
+        ConfigValue(
+            default=None,
+            description="Optional dict of directly specified kinetic parameter values for mu_H and mu_A",
+            doc="""Optional dict for directly specifying mu_H and mu_A instead of computing them
+            from the two reference temperatures (10C and 20C) via Arrhenius interpolation.
+            **default** - None.
+            **Valid values:** {
+            **None** - mu_H and mu_A are temperature-indexed Vars computed via Arrhenius interpolation,
+            **dict** - mu_H and mu_A are declared as plain unindexed Vars fixed to the given values,""",
+        ),
+    )
 
     def build(self):
         """
@@ -141,7 +157,6 @@ class ASM3ReactionParameterData(ReactionParameterBlock):
         add_object_reference(self, "i_SSSTO", self.config.property_package.i_SSSTO)
 
         # Kinetic Parameters
-        # Note: ref_temp_1 is 10 Celsius degree, ref_temp_2 is 20 Celsius degree
         k_H_dict = {"ref_temp_1": 2, "ref_temp_2": 3}
         self.k_H = pyo.Var(
             k_H_dict.keys(),
@@ -197,13 +212,21 @@ class ASM3ReactionParameterData(ReactionParameterBlock):
             doc="Saturation constant for for X_STO (g-COD-X_STO / g-COD-X_H)",
         )
         mu_H_dict = {"ref_temp_1": 1, "ref_temp_2": 2}
-        self.mu_H = pyo.Var(
-            mu_H_dict.keys(),
-            domain=pyo.PositiveReals,
-            initialize=mu_H_dict,
-            units=pyo.units.day**-1,
-            doc="Heterotrophic max. growth rate of X_H (day^-1)",
-        )
+        if self.config.calibrated_params:
+            self.mu_H = pyo.Var(
+                initialize=self.config.calibrated_params["mu_H"],
+                domain=pyo.PositiveReals,
+                units=pyo.units.day**-1,
+                doc="Heterotrophic max. growth rate of X_H (day^-1)",
+            )
+        else:
+            self.mu_H = pyo.Var(
+                mu_H_dict.keys(),
+                domain=pyo.PositiveReals,
+                initialize=mu_H_dict,
+                units=pyo.units.day**-1,
+                doc="Heterotrophic max. growth rate of X_H (day^-1)",
+            )
         self.K_NH4 = pyo.Var(
             initialize=0.01e-3,
             units=pyo.units.kg / pyo.units.m**3,
@@ -251,13 +274,21 @@ class ASM3ReactionParameterData(ReactionParameterBlock):
 
         # Autotrophic organisms X_A, nitrifying activity
         mu_A_dict = {"ref_temp_1": 0.35, "ref_temp_2": 1}
-        self.mu_A = pyo.Var(
-            mu_A_dict.keys(),
-            domain=pyo.PositiveReals,
-            initialize=mu_A_dict,
-            units=pyo.units.day**-1,
-            doc="Autotrophic max. growth rate of X_A (day^-1)",
-        )
+        if self.config.calibrated_params:
+            self.mu_A = pyo.Var(
+                initialize=self.config.calibrated_params["mu_A"],
+                domain=pyo.PositiveReals,
+                units=pyo.units.day**-1,
+                doc="Autotrophic max. growth rate of X_A (day^-1)",
+            )
+        else:
+            self.mu_A = pyo.Var(
+                mu_A_dict.keys(),
+                domain=pyo.PositiveReals,
+                initialize=mu_A_dict,
+                units=pyo.units.day**-1,
+                doc="Autotrophic max. growth rate of X_A (day^-1)",
+            )
         self.K_A_NH4 = pyo.Var(
             initialize=1e-3,
             units=pyo.units.kg / pyo.units.m**3,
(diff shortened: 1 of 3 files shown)


watertap/property_models/unit_specific/activated_sludge/asm3_reactions.py
```diff
@@ -334,17 +380,17 @@ def build(self):
...
+        z2 = (y2 / 14.0) * 61.0
+        z3 = (y3 / 14.0 - x3 / 14.0) * 61.0
+        z4 = (y4 / 14.0) * 61.0
+        z5 = (y5 / 14.0 - x5 / 14.0) * 61.0
+        z6 = (y6 / 14.0) * 61.0
+        z7 = (y7 / 14.0 - x7 / 14.0) * 61.0
+        z9 = (-x9 / 14.0) * 61.0
+        z10 = (y10 / 14.0 - 1.0 / (self.Y_A * 14.0)) * 61.0
+        z11 = (y11 / 14.0) * 61.0
+        z12 = (y12 / 14.0 - x12 / 14.0) * 61.0
```
MarcusHolly: Is it worth substituting the 14s and 61s here with mw_n and mw_alk just so that it's more clear where these numbers are coming from?

watertap/property_models/unit_specific/activated_sludge/asm3_reactions.py
```diff
@@ -50,6 +51,21 @@ class ASM3ReactionParameterData(ReactionParameterBlock):
...
+            description="Optional dict of directly specified kinetic parameter values for mu_H and mu_A",
+            doc="""Optional dict for directly specifying mu_H and mu_A instead of computing them
+            from the two reference temperatures (10C and 20C) via Arrhenius interpolation.
+            **default** - None.
+            **Valid values:** {
+            **None** - mu_H and mu_A are temperature-indexed Vars computed via Arrhenius interpolation,
+            **dict** - mu_H and mu_A are declared as plain unindexed Vars fixed to the given values,""",
+        ),
+    )
+
```
MarcusHolly, line 70: I think we should add at least one test to test_asm3_reaction.py to ensure this is working as intended

watertap/property_models/unit_specific/activated_sludge/asm3_reactions.py
```diff
@@ -360,12 +408,8 @@ def build(self):
...
 
         # Reaction Stoichiometry
-        # This is the stoichiometric part the Peterson matrix in dict form
-        # Note that reaction stoichiometry is on a mass basis.
-        # For alkalinity, this requires converting the mass of nitrogen species
-        # reacted to mass of alkalinity converted using a charge balance (effectively MW_C/MW_N)
-        mw_alk = 12 * pyo.units.kg / pyo.units.kmol
-        mw_n = 14 * pyo.units.kg / pyo.units.kmol
+        # mw_alk = 61 * pyo.units.kg / pyo.units.kmol
+        # mw_n = 14 * pyo.units.kg / pyo.units.kmol
```
MarcusHolly: Delete this since you moved these to L359

MarcusHolly (approved):
Need to remove some commented code, but everything LGTM now

watertap/property_models/unit_specific/activated_sludge/asm3_reactions.py
```diff
@@ -50,6 +51,21 @@ class ASM3ReactionParameterData(ReactionParameterBlock):
...
+            default=None,
+            description="Optional dict of directly specified kinetic parameter values for mu_H and mu_A",
+            doc="""Optional dict for directly specifying mu_H and mu_A instead of computing them
+            from the two reference temperatures (10C and 20C) via Arrhenius interpolation.
+            **default** - None.
+            **Valid values:** {
+            **None** - mu_H and mu_A are temperature-indexed Vars computed via Arrhenius interpolation,
+            **dict** - mu_H and mu_A are declared as plain unindexed Vars fixed to the given values,""",
+        ),
+    )
```
adam-a-a, line 69: I understand the intent---you want to bypass the temperature adjustment for these two parameters and just provide your own values. However, `calibrated_params` reads as a generic mechanism for supplying any set of calibrated parameters to the reaction package. Here, you only allow for disabling Arrhenius interpolation for mu_H and mu_A parameters.

Since this currently only handles mu_H and mu_A, either the config should be renamed to reflect that scope, or the override logic should be generalized to work for all temperature-dependent parameters.

watertap/property_models/unit_specific/activated_sludge/asm3_reactions.py
```diff
@@ -649,199 +693,50 @@ def _rxn_rate(self):
         self.reaction_rate = pyo.Var(
             self.params.rate_reaction_idx,
             initialize=0,
+            domain=pyo.NonNegativeReals,
```
adam-a-a, line 662: Why was this domain needed, and you sure reaction rates will always be positive? Why wouldn't we expect negative reaction rates?

watertap/property_models/unit_specific/activated_sludge/asm3_reactions.py
```diff
@@ -308,20 +339,37 @@ def build(self):
...
-        x7 = (self.f_XI - 1.0) / (64.0 / 14.0 - 24.0 / 14.0)
+        x7 = (self.f_XI - 1.0) / (64.0 / mw_n - 24.0 / mw_n)
         x8 = -1.0
-        x9 = -1.0 / (64.0 / 14.0 - 24.0 / 14.0)
-        x10 = -(64.0 / 14.0) / self.Y_A + 1.0
+        x9 = -1.0 / (64.0 / mw_n - 24.0 / mw_n)
+        x10 = -(64.0 / mw_n) / self.Y_A + 1.0
         x11 = self.f_XI - 1.0
-        x12 = (self.f_XI - 1.0) / (64.0 / 14.0 - 24.0 / 14.0)
+        x12 = (self.f_XI - 1.0) / (64.0 / mw_n - 24.0 / mw_n)
```
adam-a-a: What about 64 and 24?

watertap/property_models/unit_specific/activated_sludge/asm3_reactions.py
luohezhiming: The paper Gujer 1999 directly specify rates should >= 0, "Table 2. Kinetic rate expressions ρⱼ for ASM3. All ρⱼ ≥ 0"

watertap/property_models/unit_specific/activated_sludge/asm3_reactions.py
luohezhiming: This part  is directly translated from Uconn's julia code, I am not so sure what 64 and 24 represent, and don't want to make a guess

watertap/property_models/unit_specific/activated_sludge/asm3_reactions.py
```diff
@@ -649,199 +693,50 @@ def _rxn_rate(self):
...
-                    pyo.units.convert(
-                        blk.state_ref.temperature / pyo.units.K,
-                        to_units=pyo.units.dimensionless,
-                    )
-                    - (blk.params.ref_temp_2 + 273.15)
-                )
-            )
-            return mu_H
+        k_H = _arrhenius(self.params.k_H)
+        k_STO = _arrhenius(self.params.k_STO)
```
adam-a-a: One concern with the refactor is that the temperature-corrected kinetic parameters (k_H, k_STO, mu_H, mu_A, etc.) are now local expressions used only in the rate equations rather than Expression objects attached to the model. Previously these quantities could be inspected and reported directly after solve. Is there a reason not to retain them as model Expressions and use _arrhenius() to define those Expressions? That would preserve the reporting/debugging interface while still removing the duplicated Arrhenius calculations.

adam-a-a (changes requested):
Though this fixes some issues that we want to resolve for a use case, it also introduces subtle issues (losing Expression objects for temperature-corrected parameters, for example). Please address the comments, and I will re-review when you notify.

watertap/property_models/unit_specific/activated_sludge/asm3_reactions.py
```diff
@@ -360,12 +376,8 @@ def build(self):
...
 
         # Reaction Stoichiometry
-        # This is the stoichiometric part the Peterson matrix in dict form
-        # Note that reaction stoichiometry is on a mass basis.
-        # For alkalinity, this requires converting the mass of nitrogen species
-        # reacted to mass of alkalinity converted using a charge balance (effectively MW_C/MW_N)
-        mw_alk = 12 * pyo.units.kg / pyo.units.kmol
-        mw_n = 14 * pyo.units.kg / pyo.units.kmol
+        # mw_alk = 61 * pyo.units.kg / pyo.units.kmol
+        # mw_n = 14 * pyo.units.kg / pyo.units.kmol
```
adam-a-a:
```suggestion

```

watertap/property_models/unit_specific/activated_sludge/tests/test_asm3_reaction.py
```diff
@@ -479,5 +479,128 @@ def test_solution(self, model):
...
+    @classmethod
+    def model(cls):
+        m = ConcreteModel()
+
+        m.fs = FlowsheetBlock(dynamic=False)
+
+        m.fs.props = ASM3ParameterBlock()
+        m.fs.rxn_props = ASM3ReactionParameterBlock(
+            property_package=m.fs.props,
+            calibrated_params={"mu_H": 2, "mu_A": 1},
```
adam-a-a, line 619: Let's test the other params. Let's also assert expected variable structure to double-check how some variable may or may no longer be indexed. 

Ideally, we'd want variables to maintain the same structure when possible (either they never are indexed or are always indexed, regardless of the calibrated_params setting)

watertap/property_models/unit_specific/activated_sludge/tests/test_asm3_reaction.py
```diff
@@ -479,5 +479,128 @@ def test_solution(self, model):
             3.04264, rel=1e-4
         )
         assert value(model.fs.R1.outlet.alkalinity[0]) == pytest.approx(
-            4.9608e-3, rel=1e-4
+            5.0759e-3, rel=1e-4
+        )
+
+
+class TestReactor_CalibratedParameters:
```
adam-a-a, line 608: @luohezhiming another important comment that we should make sure to address--how has ASM3 been verified against literature? You may have done this already, but we should cite the reference verified against and test values we get against values from references. We want to avoid any transcription errors (or other errors) that can cause cascading issues down the line.

watertap/property_models/unit_specific/activated_sludge/tests/test_asm3_reaction.py
luohezhiming: Added TestParamBlock_CalibratedStructure class for variable structure tests of different cases (no calibrated; full-calibrated and partially-calibrated

watertap/property_models/unit_specific/activated_sludge/tests/test_asm3_reaction.py
luohezhiming: The literature does not give a standard simulation for results validation, but the results show 0% difference from uconn's flowsheet, where the validation results will be print out if run uconn's flowsheet

Issue #1860, opened 2026-08-27
Setting `base_currency` and `base_period` is nuanced and inconsistent


kurbansitterley:
### Description

There are lots of places in WaterTAP where we designate the units for a `Var` or `Param` etc as the `base_currency` for the costing package. Currently our default `base_currency` is USD2018. Fewer places use `base_period`, but I mention it here because the same issue applies.

At the flowsheet level, we can change the `base_currency` with e.g., `m.fs.costing.base_currency = pyunits.USD_2023`. But at this point, it is too late for some variables, and maybe a lot of variables, as outlined by @avdudchenko in [this comment.](https://github.com/watertap-org/watertap/pull/1849#discussion_r3686388102). This depends on when `base_currency`/`base_period` is set at the flowsheet level.

It is common for us to designate some costing variables to be in units of `base_currency`. This makes sense, as we want them to be reported in the currency year of interest. Consider the following flowsheet and two places where these costing parameters are set: (1) immediately after instantiating the costing block and (2) after building the property, costing, and process models.

```
from pyomo.environ import ConcreteModel, units as pyunits
from idaes.core import FlowsheetBlock, UnitModelCostingBlock
from watertap.costing import WaterTAPCosting
from watertap.property_models.multicomp_aq_sol_prop_pack import MCASParameterBlock
from watertap.unit_models.gac import GAC

m = ConcreteModel()
m.fs = FlowsheetBlock(dynamic=False)
m.fs.costing = WaterTAPCosting()
# Point 1
# m.fs.costing.base_currency = pyunits.USD_2023
# m.fs.costing.base_period = pyunits.month
print(f"electricity_cost units @ 1 = {pyunits.get_units(m.fs.costing.electricity_cost)}")
print(f"plant_lifetime units @ 1 = {pyunits.get_units(m.fs.costing.plant_lifetime)}")

m.fs.properties = MCASParameterBlock(
    solute_list=["DCE"],
    mw_data={"H2O": 0.018, "DCE": 0.09896},
    ignore_neutral_charge=True,
)
m.fs.unit = GAC(property_package=m.fs.properties)

m.fs.unit.costing = UnitModelCostingBlock(flowsheet_costing_block=m.fs.costing)
print(f"unit capital cost units @ 1 = {pyunits.get_units(m.fs.unit.costing.capital_cost)}")
print(f"unit operating cost units @ 1 = {pyunits.get_units(m.fs.unit.costing.fixed_operating_cost)}")
[...]

PR #1864, opened 2026-09-02
Declare `base_currency` and `base_period` in WaterTAPCosting config

kurbansitterley:
## Fixes/Resolves:

#1860 

## Summary/Motivation:

#1860 

- declare `base_currency_year` and `base_period` via configuration argument for WaterTAP costing package
- remove `set_base_currency_base_period` 
- test this functionality
- update docs to reflect this change
- make changes throughout repo where it makes sense to set `base_currency_year` as a config where we previously did e.g., `m.fs.costing.base_currency = pyunits.USD_2020` 
- add logging statements to costing package that shows what `base_currency` and `base_period` are set to _and_ how they were set (i.e., either from the case study yaml or via the config)

from yaml: 
```
2026-09-15 12:49:12 [INFO] idaes.watertap.costing.watertap_costing_package: Setting base_currency from case study yaml: MUSD_2018
2026-09-15 12:49:12 [INFO] idaes.watertap.costing.watertap_costing_package: Setting base_period from case study yaml: year
```
from config:
```
2026-09-15 12:49:12 [INFO] idaes.watertap.costing.watertap_costing_package: Setting base_currency from config: USD_2011
2026-09-15 12:49:12 [INFO] idaes.watertap.costing.watertap_costing_package: Setting base_period from config: year
```
## NOTE
There are many ZO tests where we directly set the base currency via e.g., `m.fs.costing.base_currency = pyunits.USD_2014`. The docs will say that we discourage this behavior after this PR. The "official" way to do this would be to set base currency via the yaml and then, if it is not in the yaml, to use the `base_currency_year` in the costing package configuration. However, setting the base currency directly like this in the tests was done in these ZO model tests to check the costing calculations against the reference; to set `base_currency` the _suggested_ way (via yaml) we would need to have a separate yaml for each of these tests, which seems silly for this purpose. 

As an alternative, I could refactor this so that `base_currency` in ZO costing package will preferentially use the `base_currency_year` config rather than the entry in the yaml. But that is not the current status. I opted to maintain the current "framework" for ZO models where we allow users to set their technoeconomic parameters via the case study yaml. 

TODO:
- [x] test for ZO costing
- [x] add to documentation
- [x] modify necessary how-tos

### Legal Acknowledgement

### DIFF
--- a/watertap/costing/tests/test_util.py
+++ b/watertap/costing/tests/test_util.py
@@ -149,9 +149,7 @@ def test_rectifier_costing():
     # build generic models
     m = pyo.ConcreteModel()
     m.fs = idc.FlowsheetBlock(dynamic=False)
-    m.fs.costing = WaterTAPCosting()
-    # change base units for value assertions
-    m.fs.costing.base_currency = pyo.units.USD_2021
+    m.fs.costing = WaterTAPCosting(base_currency_year=2018)
     m.fs.unit = idc.UnitModelBlock()
 
     # build power variable required for cost_rectifier
@@ -178,12 +176,12 @@ def test_rectifier_costing():
     assert pyo.check_optimal_termination(results)
 
     # check values
-    assert pytest.approx(59320, rel=1e-3) == pyo.value(m.fs.unit.costing.capital_cost)
+    assert pytest.approx(50531, rel=1e-3) == pyo.value(m.fs.unit.costing.capital_cost)
     assert pytest.approx(111.1, rel=1e-3) == pyo.value(m.fs.unit.costing.ac_power)
     assert pytest.approx(111.1, rel=1e-3) == pyo.value(
         m.fs.costing.aggregate_flow_electricity
     )
-    assert pytest.approx(80040, rel=1e-3) == pyo.value(
+    assert pytest.approx(68180, rel=1e-3) == pyo.value(
         m.fs.costing.aggregate_flow_costs["electricity"]
     )
 
--- a/watertap/costing/tests/test_zero_order_costing.py
+++ b/watertap/costing/tests/test_zero_order_costing.py
@@ -14,6 +14,7 @@ Tests for general zero-order costing methods
 """
 
 import os
+import re
 import yaml
 import tempfile
 import pytest
@@ -40,6 +41,7 @@ from idaes.core.util.model_statistics import (
     number_unfixed_variables,
 )
 from idaes.core.util.misc import add_object_reference
+from idaes.core.util.exceptions import ConfigurationError
 
 from watertap.costing.zero_order_costing import (
     ZeroOrderCosting,
@@ -502,20 +504,20 @@ class TestWorkflow:
 
 @pytest.mark.component
 def test_watertap_costing_config_zo():
-    # Create a temporary case study file without base_currency and base_period
+
     here = os.path.dirname(os.path.abspath(__file__))
-    default_cs_path = f"{os.path.dirname(os.path.dirname(here))}/data/techno_economic/default_case_study.yaml"
 
-    with open(default_cs_path, "r", encoding="utf-8") as default_cs:
-        data = default_cs.read()
-        data = yaml.load(data, Loader=yaml.Loader)
+    m = ConcreteModel()
+    m.fs = FlowsheetBlock(dynamic=False)
+    m.fs.costing = ZeroOrderCosting()
 
+    data = _load_case_study_definition(m.fs.costing)
     data.pop("base_currency", None)
     data.pop("base_period", None)
 
-    with tempfile.NamedTemporaryFile(
-        suffix=".yaml", delete=False, mode="w", encoding="utf-8"
-    ) as tf:
+    # Create a temporary case study file without base_currency and base_period
+    temp_path = f"{here}/temp.yaml"
+    with open(temp_path, "w", encoding="utf-8") as tf:
         yaml.safe_dump(data, tf, sort_keys=False)
         temp_path = tf.name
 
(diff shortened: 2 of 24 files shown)


watertap/costing/watertap_costing_package.py
```diff
@@ -500,7 +513,19 @@ def build_global_params(self):
...
-        self.set_base_currency_base_period()
+
+        if not 1990 <= self.config.base_currency <= 2023:
+            raise ConfigurationError(
+                f"Base currency year must be between 1990 and 2023, but got {self.config.base_currency}"
+            )
+
+        self.base_currency = getattr(pyo.units, f"USD_{self.config.base_currency}")
+
+        if self.config.base_period not in ["year", "month", "day"]:
```
adam-a-a: I think we can consider some logic here that leverages pint units to accept different base period strings compatible with pint's library.

adam-a-a (dismissed):
This is great. The only suggestion is to not have the list of acceptable strings for base period and instead use Pint to determine compatibility (then raise exception if not compatible with time)

watertap/costing/watertap_costing_package.py
```diff
@@ -50,11 +69,47 @@ def register_currency_definitions(self):
...
+
+        if self.base_period is None:
+            try:
+                self.base_period = getattr(pyo.units, self.config.base_period)
+            except AttributeError:
+                raise ConfigurationError(
+                    f"{self.config.base_period} is not a valid unit."
+                )
+
+            self.base_period = getattr(pyo.units, self.config.base_period)
```
MuktaHardikar, line 107: Is this needed again?

adam-a-a:
```suggestion
```

watertap/costing/watertap_costing_package.py
```diff
@@ -50,11 +69,47 @@ def register_currency_definitions(self):
...
-        self.base_currency = pyo.units.USD_2018
-        # Set a base period for all operating costs
-        self.base_period = pyo.units.year
+    def validate_watertap_costing_config(self):
+        """
+        Validate the configuration of a WaterTAP costing block.
+        and set the base_currency and base_period attributes.
+        """
+
+        self.base_currency = None
```
adam-a-a, line 78: Do we need another check here in case someone tried calling this method on their own a subsequent time? That is, if `validate_watertap_costing_config` is called a second time, should the logic check for existence of base_currency and base_period first, and return if so?

i.e.
```
if self.base_currency is not None and self.base_period is not None:
    return
```

adam-a-a: Or raise an exception stating that one or both have been defined, and that if the user wants to alter, they should create a new instance

watertap/unit_models/tests/test_electrocoagulation.py
```diff
@@ -584,10 +583,10 @@ def test_costing(self):
             "aggregate_variable_operating_cost": 0.0,
             "aggregate_flow_electricity": 146.33,
             "aggregate_flow_aluminum": 64565.49,
-            "aggregate_flow_costs": {"electricity": 118796.73, "aluminum": 162263.39},
+            "aggregate_flow_costs": {"electricity": 89793.59, "aluminum": 162263.39},
```
MuktaHardikar, line 586: Shouldn't the `aggregate_flow_costs` remain the same for electricity too?

watertap/costing/watertap_costing_package.py
```diff
@@ -50,11 +69,47 @@ def register_currency_definitions(self):
...
+        self.base_period = None
```
MuktaHardikar, line 79: self.base_currency = None
        self.base_period = None
```suggestion
        if (
            getattr(self, "base_currency", None) is not None
            and getattr(self, "base_period", None) is not None
        ):
            raise Exception(
                f"base_currency and base_period already exists on this costing block "
                f"with values {self.base_currency} and {self.base_period}, respectively."
            )
        
        self.base_currency = None
        self.base_period = None

```

watertap/costing/watertap_costing_package.py
```diff
@@ -50,11 +69,47 @@ def register_currency_definitions(self):
...
+                raise ConfigurationError(
+                    f"{self.config.base_period} is not a valid unit."
+                )
+
+            self.base_period = getattr(pyo.units, self.config.base_period)
+
+            if not self.base_period._pint_unit.dimensionality == "[time]":
+                msg = f"base_period configuration must be a unit of time "
+                msg += f"but got {self.config.base_period} {self.base_period._pint_unit.dimensionality}."
+                raise ConfigurationError(msg)
```
MuktaHardikar, line 113:
```suggestion
        _log.debug(
            f"base_currency={self.base_currency}, base_period={self.base_period}"
        )
```

watertap/costing/watertap_costing_package.py
kurbansitterley: Good point and I hadn't considered a user calling this method on their own.

imo, we should discourage that behavior. A motivating factor for this change is that there are instances across the package that use `base_currency` as their units at instantiation and simply doing e.g. `m.fs.costing.base_currency = pyunits.USD_2020` was _not_ modifying the units after creating the costing package. Additionally, a user would have to be pretty WaterTAP-savvy to get the intended outcome of re-calling `validate_watertap_costing_config` since it does not have any arguments currently; i.e., they would first have to do e.g., `m.fs.costing.config.base_currency_year = 2020`. 

So we should think carefully about if we want to allow the user to reset the `base_currency` after instantiation. If we do, I would want to also add a method that would, at minimum, re-set all the costing-specific parameters/variables to be in the new `base_currency` and I am not sure how well this would work (e.g., would already-declared Expressions/Constraints convert to the new currency? I don't know). 

I would propose that we make `validate_watertap_costing_config` a hidden method/un-documented _and_ add the check you described if a savvy user tries to re-call the hidden method. Related: I considered making `base_currency` and `base_period` hidden attributes so that trying to do `m.fs.costing.base_currency = pyunits.USD_2020` would print/log a warning describing the intended behavior, but that would have required changing `base_currency` to `_base_currency` everywhere and I was worried about additional fallout.

adam-a-a: I don't think we need to make any substantial changes to what you have here, but I thought it might make sense to at least exit the function a have a logger message reporting the currency and base period, if somehow, some way, this method were called after being called before. Right now, it looks like the function will set base curency and period to None, then check that they're None before executing the rest of the code.

I just noticed that @MuktaHardikar suggested a fix that I agree with.

watertap/unit_models/tests/test_electrocoagulation.py
kurbansitterley: No. Prior to this PR, `electricity_cost` had units of `base_currency`, but `base_currency` was hardcoded as 2018 because that was the default. This PR corrects that issue so electricty costs will all be in the `base_currency_year` declared via config, in this case 2023.

watertap/costing/watertap_costing_package.py
adam-a-a: I think this works. Later on, we can decide if we want to be less restrictive for whatever reason.

watertap/costing/watertap_costing_package.py
```diff
@@ -50,11 +69,47 @@ def register_currency_definitions(self):
...
+        if "case_study_definition" in self.config:
+            # it is a ZeroOrderCosting block, so we preferentially
+            # use the values from the _cs_def if available
+            if "base_currency" in self._cs_def:
+                self.base_currency = getattr(pyo.units, self._cs_def["base_currency"])
+            if "base_period" in self._cs_def:
+                self.base_period = getattr(pyo.units, self._cs_def["base_period"])
+
+        if self.base_currency is None:
+            if not 1990 <= self.config.base_currency_year <= 2023:
```
adam-a-a: Should this check happen regardless of whether base_currency is None or not?

adam-a-a: That is, I think we should be checking currency year whether it was introduced through a yaml case study file or directly through a config option.

watertap/costing/tests/test_util.py
```diff
@@ -149,9 +149,7 @@ def test_rectifier_costing():
     # build generic models
     m = pyo.ConcreteModel()
     m.fs = idc.FlowsheetBlock(dynamic=False)
-    m.fs.costing = WaterTAPCosting()
-    # change base units for value assertions
-    m.fs.costing.base_currency = pyo.units.USD_2021
+    m.fs.costing = WaterTAPCosting(base_currency_year=2018)
```
adam-a-a: This is meant to be 2021

watertap/costing/tests/test_zero_order_costing.py
```diff
@@ -332,6 +336,8 @@ def model(cls):
         m.fs.params = WaterParameterBlock(solute_list=["sulfur", "toc", "tss"])
 
         m.fs.costing = ZeroOrderCosting()
+        # NOTE: setting base_currency in this way is discouraged
+        # It is recommended to set the base_currency through the case study definition file
         m.fs.costing.base_currency = pyunits.USD_2020
```
adam-a-a, line 341: Hmmm.... I think what we want to move towards is raising an exception of someone were to do this. I think this means we'd need to make some changes in IDAES's FlowsheetCostingBlock as well.

kurbansitterley: I considered doing something like this: 

```
@property
def base_currency(self):

    _log.info("Setting base_currency directly is not supported.")

    # or raise an exception

    return self.base_currency
```

But I think this would spam the logs with the same message any time `base_currency` is accessed.

watertap/costing/tests/test_zero_order_costing.py
```diff
@@ -495,3 +501,95 @@ def test_solution(self, model):
...
+    assert m.fs.costing.base_currency == pyunits.USD_2000
+    assert m.fs.costing.config.base_period == "day"
+    assert m.fs.costing.base_period == pyunits.day
+    assert pyunits.get_units(m.fs.costing.total_capital_cost) == pyunits.get_units(
+        pyunits.USD_2000
+    )
+    assert pyunits.get_units(m.fs.costing.total_operating_cost) == pyunits.get_units(
+        pyunits.USD_2000 / pyunits.day
+    )
+
```
adam-a-a, line 611: What about `salaries_percent_FCI` and the other cost components? Can we check the units on those too?

kurbansitterley: made this change, but it required changing when the inherited costing packages call `build_global_params`

PR #1865, opened 2026-09-03
Seawater properties with concentration and volumetric flow rate state variables

dallan-keylogic:
## Summary/Motivation:
This PR implements a version of the seawater properties with mass concentration and volumetric flow rate as state variables. Because WaterTAP's testing harness previously supported only phase component flow state variables, some changes had to be made to make it support these new state variables. The process of modifying the harness revealed other issues with the `fix_initialization_states` methods of multiple other property packages.

- New version of seawater properties with new state variables
- Changes to property test harness to accommodate redundant state variables.
- Correct implementation of `fix_initialization_states` for other property packages.

### Legal Acknowledgement


watertap/property_models/seawater_prop_pack_VTPC.py
```diff
@@ -0,0 +1,690 @@
...
+# National Laboratory of the Rockies, and National Energy Technology
+# Laboratory (subject to receipt of any required approvals from the U.S. Dept.
+# of Energy). All rights reserved.
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and license
+# information, respectively. These files are also available online at the URL
+# "https://github.com/watertap-org/watertap/"
+#################################################################################
+"""
+Initial property package for seawater system
```
adam-a-a, line 13: High-level comment: why can't we do this in the primary sw prop pack and allow for selection of Q and C as state vars through a config option? Do we need a duplicate prop pack to achieve this?

watertap/property_models/seawater_prop_pack_VTPC.py
```diff
@@ -0,0 +1,690 @@
...
+    SeawaterParameterData,
+    SeawaterStateBlockData,
+)
+
+# Set up logger
+_log = idaeslog.getLogger(__name__)
+
+
+@declare_process_block_class("SeawaterVTPCParameterBlock")
+class SeawaterVTPCParameterData(SeawaterParameterData):
```
adam-a-a, line 67: OK- I see now that this inherits from the SW prop pack.

watertap/property_models/seawater_prop_pack_VTPC.py
```diff
@@ -0,0 +1,690 @@
...
+    "SeawaterVTPCStateBlock", block_class=_SeawaterVTPCStateBlock
+)
+class SeawaterVTPCStateBlockData(SeawaterStateBlockData):
+    """A seawater property package."""
+
+    def build(self):
+        """Callable method for Block construction."""
+        # Do not call build from immediate parent class
+        # but instead from its parent (StateBlockData)
+        super(SeawaterStateBlockData, self).build()
```
adam-a-a, line 426: since you inherit from SeawaterStateBlockData, will its build method be invoked and result in rewriting components? Won't flow_mass_phase_comp also be created as a variable here then?

dallan-keylogic: The way the `super` function works is really confusing. This does not call the build method of `SeawaterStateBlockData`, but instead it calls the build method of `StateBlockData`.

watertap/property_models/seawater_prop_pack_VTPC.py
```diff
@@ -0,0 +1,690 @@
...
+
+        # unfix all variables fixed with var_args
+        for (k, v_name, ind), previously_fixed in flags.items():
+            if not previously_fixed:
+                var = getattr(self[k], v_name)
+                var[ind].unfix()
+
+        # fix state variables if hold_state
+        if hold_state:
+            self.fix_initialization_states(self)
```
dallan-keylogic, line 411: If I change `fix_state_vars` to `fix_initialization_states` in the original seawater properties, I can inherit this entire method.

MuktaHardikar: Since you removed the unfixing of flow_mass_phase_comp variable, we could change it `fix_initialization_states` in the initialization routine even in the seawater_prop_pack I believe.

watertap/property_models/seawater_prop_pack_VTPC.py
```diff
@@ -0,0 +1,690 @@
...
+        self.pressure = Var(
+            initialize=101325,
+            bounds=(1e3, 5e7),
+            units=pyunits.Pa,
+            doc="Pressure",
+        )
+
+        if not self.config.defined_state:
+
+            @self.Constraint(self.phase_list)
```
MuktaHardikar, line 467: Just want to double check why this constraint is needed. Without this constraint, if `defined_state` is False, `conc_mass_phase_comp` is still a degree of freedom. The `flow_mass_phase_comp` is calculated using constraint in line 503. And `mass_frac_phase_comp` is calculated in 514. Wouldn't the density be calculated using the `mass_frac_phase_comp` and the density function in the seawater package? And in that case you don't need to unfix `conc_mass_phase_comp` during initialization

MuktaHardikar: But that way `dens_mass_phase`,`flow_mass_phase_comp` and `mass_frac_phae_comp` are not created.

dallan-keylogic: When we calculate the state variables from the phase component flows (which typically come from material balance equations), we have `n_comp` constraints (`eq_flow_mass_phase_comp`) but `n_comp+1` variables (`conc_mass_phase_comp` and `flow_vol_phase`). When we calculate the mass fractions, we introduce `n_comp` equations (`eq_mass_frac_phase_comp`) and `n_comp` unknowns (`mass_frac_phase_comp`). We still have one degree of freedom, which is provided by ensuring that the concentrations add up to the density.

For IDAES property packages in which mole fraction is a state variable, we have the constraint that the sum of mole fractions equals one. That corresponds to the sum of the concentrations adding up to the density. If you divided both sides by density, it's also equivalent to the sum of the mass fractions adding up to 1.

MuktaHardikar: This makes sense. Should this constraint be satisfied even for `defined_state=True`? Otherwise it would be possible to enter multiple pairs of concentrations for TDS and H2O that results in the same mass fraction (so the density calculated using the correlation in seawater prop pack is the same) but the density would not be the same as `sum(b.conc_mass_phase_comp[p, j] for j in b.component_list)`. Its no longer `defined_state=True` in that case I suppose.

dallan-keylogic: The option `defined_state=True` exists primarily because of arc constraints. Every state variable in a unit inlet is defined by an arc constraint from the previous unit. We thus don't have any degrees of freedom left to satisfy a concentration sum constraint. Usually, the stream is coming from another state block with `defined_state=False`, so the concentration values should still add up to density. The exception is when it's coming from a feed block or the user is fixing values on the inlet. In that case, we rely on the user to use consistent concentration values, just like we do in IDAES when the user provides mole fractions.

However, while it's pretty easy for the user to provide mole fractions that add up to 1, it's much harder for the user to provide concentration values that satisfy the density constraint. I've been thinking about using some `conc_mass_solute_comp` variable as a state variable so the user can fix the TDS values (or, in other property packages, all solute concentrations) without fixing the water concentration. That's a departure from what we do in IDAES, so I've been a bit hesitant.

